PR #53, opened 2020-04-11
Enhancement to support Docker Compose 

weyert:
I have written a `Dockerfile` and a `docker-compose.yml` to allow running Plausible within Docker, the docker compose will create a `postgres`-database, and also will build the `Plausible` docker image you can try it out using;

```
docker-compose build
docker-compose up
```

After the latter command succeeds you should have access to the development build of Plausible at url http://localhost:8000. You can use the docker logs to dig up the verification url from the mail ;)

Of course, this is not the right way to run things in Production but I don't have enough experience with Elixer to make this work. I tried to make a release but that didn't work as it appears use the environment variables at build time when running the release. If anyone know how to do this that would be great :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,91 @@
+FROM elixir:alpine
+
+ARG app_name=plausible
+ARG phoenix_subdir=.
+ARG build_env=prod
+ARG AVIEN_DATABASE_URL=""
+
+# Set up build environment.
+ENV MIX_ENV=${build_env} \
+  REPLACE_OS_VARS=true \
+  TERM=xterm
+ENV AVIEN_DATABASE_URL=${AVIEN_DATABASE_URL}
+RUN set -xe
+RUN echo "MIX_ENV=$MIX_ENV"
+RUN echo "AVIEN_DATABASE_URL=$AVIEN_DATABASE_URL"
+
+# Set the build directory.
+RUN mkdir /opt/app
+WORKDIR /opt/app
+
+# Install build tools needed in addition to Elixir:
+# NodeJS is used for Webpack builds of Phoenix assets.
+# Hex and Rebar are needed to get and build dependencies.
+
+# This step installs all the build tools we'll need
+RUN apk update && \
+  apk upgrade --no-cache && \
+  apk add --no-cache \
+  postgresql-client \
+  nodejs \
+  npm \
+  git \
+  build-base && \
+  mix local.rebar --force && \
+  mix local.hex --force
+
+# Copy the application files into /opt/app.
+COPY . .
+
+# Build the application.
+RUN mix do deps.get --only $MIX_ENV, deps.compile
+
+# Build assets by running a Webpack build and Phoenix digest.
+# If you are using a different mechanism for asset builds, you may need to
+# alter these commands.
+RUN cd ${phoenix_subdir}/assets \
+  && npm install \
+  && ./node_modules/webpack/bin/webpack.js --mode production \
+  && cd .. \
+  && mix phx.digest
+
+# Create the release, and move the artifacts to well-known paths so the
+# runtime image doesn't need to know the app name. Specifically, the binary
+# is renamed to "start_server", and the entire release is moved into
+# /opt/release.
+RUN mix release \
+  && mv _build/${build_env}/rel/${app_name} /opt/release \
+  && mv /opt/release/bin/${app_name} /opt/release/bin/start_server
+
+# # Step 2: Execute the application
+# FROM alpine:latest
+
+# RUN apk update \
+#     && apk --no-cache --update add bash ca-certificates openssl-dev
+
+ENV PORT=4000 \
+  REPLACE_OS_VARS=true
+EXPOSE ${PORT}
+EXPOSE 8000
+
+# # Set the install directory. The app will run from here.
+# WORKDIR /opt/app
+
+# # Obtain the built application release from the build stage.
+# COPY --from=0 /opt/release .
+# #COPY --from=releaser app/bin/ ./bin
+
+
+# # Copy the Docker entrypoint.sh to the release docker image
+# COPY --from=0 /opt/app/entrypoint.sh .
+
+# # Start the server.
+
+# # # CMD exec /opt/app/bin/start_server start
+
+RUN chmod +x /opt/app/entrypoint.sh
+
+# CMD /opt/app/entrypoint.sh
+# CMD tail -f /dev/null
+
+ENTRYPOINT /opt/app/entrypoint.sh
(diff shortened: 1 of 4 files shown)


alexanderadam:
Nice, thank you for your work! :pray: 
This looks great. I have not enough knowledge about Elixir/Phoenix so I can't say much about those but I have some feedback anyway, if that's okay for you:

### 1. Docker images should _not_ run as `root` if possible

…(and it's nearly _always_ possible :wink:. Or [how the Docker docs put it](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user):

> If a service can run without privileges, use `USER` to change to a non-root user.

Ideally UID and GID are also arguments, so that the ids of the unprivileged user can match some ids of the volume owner in case that's necessary.

So you could add something like this:

```Dockerfile
ARG UID
ARG GID

RUN apk add --no-cache shadow && \
    if [ -z "`getent group $GID`" ]; then \
      addgroup -S -g $GID plausibleuser; \
    else \
      groupmod -n plausibleuser `getent group $GID | cut -d: -f1`; \
    fi && \
    if [ -z "`getent passwd $UID`" ]; then \
      adduser -S -u $UID -G plausibleuser -s /bin/sh plausibleuser; \
    else \
      usermod -l plausibleuser -g $GID -d /home/plausibleuser -m `getent passwd $UID | cut -d: -f1`; \
    fi



# Set the build directory.
RUN mkdir /opt/app
WORKDIR /opt/app

# copy files or ideally check out master branch with git
# install dependencies (before switching to unprivileged user)

RUN chown plausibleuser:plausibleuser /opt/app

USER plausibleuser

# everthing after the above line will be run as plausibleuser
```

There are also many posts on this topic about the why's and the hows (i.e. [1](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b), [2](https://medium.com/better-programming/running-a-container-with-a-non-root-user-e35830d1f42a), [3](https://medium.com/redbubble/running-a-docker-container-as-a-non-root-user-7d2e00f8ee15)).

### 2. Cleanup of some unneeded comments

It seems that there are some legacy comments. Right? :thinking: 

### 3. Use production environment

> I tried to make a release but that didn't work as it appears use the environment variables at build time when running the release.

I'm not an Elixir dev neither but can you explain what's happening?
I mean apart from adding env variables like `PGPASSWORD` and `POSTGRES_HOST` to the other environments as well (maybe those should be renamed to `POSTGRES_PASSWORD` and `POSTGRES_HOST` to match the variables of the postgres image?)

Also it's probably better to set/export `MIX_ENV` just once. Currently the entrypoint will set it explicitly to `dev` a few times. So I guess this might lead to errors in case someone forgets about it.

weyert:
Glad to hear you liked it @alexanderadam, yeah, fair point not to run it as root. I will have a look at that don't think that shouldn't be too difficult but we are still talking about computers ;)

`PGPASSWORD` and `POSTGRES_HOST` environment variables are the names used by PostgresSQL CLI tools. The idea was to keep waiting until the PostgresSQL database becomes available and then run your migration scripts at startup.

Yeah, regarding production environment, the main problem I had here was it didn't use my process environment variables when running your application instead it got all compiled at compile time and kept using those. I seem to have resolved this in the other branch in my repo.

Do you have experiences with `mix release`? My other branch is using that to try the runtime environment vars config working. I will keep cracking on it  :)

alexanderadam:
> Do you have experiences with `mix release`?

I'm afraid not, sorry.

steffenix:
The entrypoint file is more complex that it should be, after the release task you should not have to use mix but just start the server with:

```
 _build/prod/rel/app/bin/app start
```

In order to run the migrations before the start I do:
```
_build/prod/rel/app/bin/app eval "App.Release.migrate";
```

With this following file added to the code.

```
defmodule App.Release do
  @app :app

  def migrate do
    for repo <- repos() do
      {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :up, all: true))
    end
  end

  def rollback(repo, version) do
    {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :down, to: version))
  end

  defp repos do
    Application.load(@app)
    Application.fetch_env!(@app, :ecto_repos)
  end
end
```

weyert:
Thanks @steffenix. I will update my PR. I don't get the feeling the author of this solution appreciates this PR, though

ukutaht:
Thanks to everyone on this.

Sorry, of course I appreciate this PR and I'm very thankful. I've planned to take a look at this but just haven't got to it yet.

My main concern is that I don't run the app on docker. Not locally nor in production. So it's likely that the docker setup will break and get out of sync at some point as I develop the codebase and infrastructure.

I will test this out and merge soon. Looking at the code, I want to make sure the changes to `config/dev.ex` don't break the standard local postgres setup.

Again, sorry for the delay on this.

alexanderadam:
> My main concern is that I don't run the app on docker. Not locally nor in production. So it's likely that the docker setup will break and get out of sync at some point as I develop the codebase and infrastructure.

Users will give feedback soon enough I guess.
Furthermore [there's already a travis config](https://github.com/plausible-insights/plausible/blob/master/.travis.yml) and travis can use docker images anyway, so I guess it should be possible to get early feedback via travis, too.

> I will test this out and merge soon.

[As mentioned earlier](https://github.com/plausible-insights/plausible/pull/53#issuecomment-613292743) I think this PR is _not_ merge-ready yet :

1. This PR requires a cleanup (many legacy comments, especially in `Dockerfile`)
2. Changes for `prod.exs` and `test.exs` are still missing.
3. [`ENV` variable consistency would be nice](https://github.com/plausible-insights/plausible/pull/53#issuecomment-613394795)
4. Image is still running as `root`

tckb:
@ukutaht Really appreciate the effort you put in to open source it. I was searching for this exact solution. Although there are other  (OSS) products, either they are incomplete or have not open-sourced  the "privacy" focused part. 

I am forking the current repo in-order to "add-in"  the missing `Dockerfile` , I already have some of personal elixir apps running on my personal k8s cluster, I plan to add this one too. 

I plan to work on this one, once this is stable enough and I will (read _may_)  a PR for merging it into the upstream. Feel free to join for contributions. 

https://gitlab.com/tckb-public/plausible

weyert:
Thanks @tckb, any plans to make a PR for it? I am happy to make the changes listed by @alexanderadam

tckb:
Yes, I plan to do it but, I'm not sure of the changes are aligned with the author. As I mentioned in https://github.com/plausible-insights/plausible/issues/26#issuecomment-625121700 I have few things to do. I am tracking the issues I will work in the coming days here: https://gitlab.com/tckb-public/plausible/-/issues 

once this is finished I will create a PR

ukutaht:
Closing this in favour of #64

PR #56, opened 2020-04-20
Correctly position more links on mobile

pmhoudry:
The "more" links are absolutely positioned. It works fine on desktop when there is two columns. On mobile, the absolute sends the "more" links on top of each other. Adding a simple "relative" class on each cards that contains a "more" link ensures that they stay when they belong.


ukutaht:
Awesome, thanks for the fix!

PR #57, opened 2020-04-20
Correctly position websites list on mobile

pmhoudry:
On mobile, the website list at /sites does not have any X padding. The elements touch the screen left and right. This corrects it by wrapping the list in a div with "px-4" on it.

This could also be achieved by setting "px-4" and "md:px-0" on every element but it is unecessarily complex and if some more elements appeared on this screen, it would be easy to forget this. A container should be in charge of this, which is what is proposed here.

Only two lines are added. Two spaces are added at the beginning of all the other lines changed for better formatting of the code.


ukutaht:
Looks great! Thank you!

PR #64, opened 2020-05-15
Support for docker based self-hosting

tckb:
In this PR, I made changes for making it easy for self-hosting. 
Salient changes are listed below: 

- [x] Add default user during initial setup
- [x]  Add support for SMTP server for Mailer 
- [x]  Update base-domain according to "HOST" param 
- [x]  Cosmetic changes across website for self-hosting 

// Pending changes from review: 
- [x] Revert TZ to UTC  https://github.com/plausible-insights/plausible/pull/64#discussion_r428450420
- [x]  Integrate clickhouse and merge changes from upstream https://github.com/plausible-insights/plausible/pull/64#pullrequestreview-415885280
- [x] Resolve Conflicts


alexanderadam:
Wow, this was a lot of work but it looks great for me.

tckb:
> Wow, this obviously was a lot of work but it looks great for me.
> Thank you for your work!
the large number of changes is just a facade of formatting.  the  fact of the matter is that `mix` formatted all  files according the rules.  In fact the major changes are only limited to the `config`

config/config.exs
```diff
@@ -1,31 +1,41 @@
...
     accepts: ~w(html json)
   ],
   pubsub: [name: Plausible.PubSub, adapter: Phoenix.PubSub.PG2]
 
 config :sentry,
-  dsn: "https://0350a42aa6234a2eaf1230866788598e@sentry.io/1382353",
+  dsn: System.get_env("SENTRY_DSN"),
   included_environments: [:prod, :staging],
-  environment_name: String.to_atom(Map.get(System.get_env(), "APP_ENV", "dev")),
+  environment_name: String.to_atom(Map.get(System.get_env(), "MIX_ENV", "dev")),
```
ukutaht, line 34: Did you remove APP_ENV altogether? In my setup, there's a distinction between APP_ENV and Mix env.

The only case where it really matters is that on the staging server, I'm running `MIX_ENV=prod` and `APP_ENV=staging`. For me, the mix env says whether we should be running the code in dev, test or prod configuration even though I might run the prod configuration on different 'app environments'.

Does that make sense?

lib/plausible_migration.ex
```diff
@@ -0,0 +1,152 @@
+defmodule Plausible.Release do
```
ukutaht, line 1: Is there a good reason to not match the module name to the filename? I'd prefer for them to be aligned if possible

HOSTING.md
```diff
@@ -0,0 +1,127 @@
+# Plausible Insights
```
ukutaht, line 1: FYI, we've decided to go with 'Plausible Analytics' as the brand name. I know we're out of date in some places like the github org name but for all new assets I'd prefer to use 'Plausible Analytics'

config/config.exs
tckb: yes of course, the idea was  **APP_ENV**  would indicate the _current_  *version* of application the running in the environment (prod, staging etc).  

But after looking carefully, it was a bit confusing so I have replaced APP_ENV with _**APP_VERSION**_ https://github.com/tckb/plausible/blob/master/config/releases.exs#L31 and added it to release https://github.com/tckb/plausible/blob/master/config/releases.exs#L74

This way it will allow us to track bugs with the current version. Gives a better view of the context, track changes etc https://github.com/tckb/plausible/blob/master/config/releases.exs#L74

lib/plausible_migration.ex
tckb: no, I started it something else and ended up the current version. The module name needs to be "Plausible.Migration" although, ". Release" also makes total sense. Which one would you suggest? I would prefer release.

- [ ] Plausible.Release

- [ ] Plausible.Migration

HOSTING.md
tckb: Good choice, I will edit it in the HOSTING documentation.

config/config.exs
ukutaht: Gotcha. I think it's a good idea to tag the APP_VERSION on requests to sentry.

However, if you look at the screenshot below, I have prod and staging errors separated in Sentry at the moment. With this setup above, the MIX ENV is `prod` in both prod and staging so I would lose the separation.


Screenshot 2020-05-15 at 22 29 14

tckb:
>     1. What's the advatange of building a release over just running `MIX_ENV=prod mix phx.server` in a docker container with Elixir installed? The Mix release feels like a similar artifact as a docker image, so isn't the release adding an unnecessary extra step here?

There are several uses and advantages of releases. Here's the official version: https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#why-releases 
But, personally. I think a production environment needs to be "deployed" as an immutable artifact and shouldn't be deployed with a source code. A release artifact is a self-contained, per-compiled binary without any additional dependencies. 

>     2. The original [feature request](https://plausible.nolt.io/56) specifies Docker Hub as the registry. Can we mirror from Gitlab registry to docker hub? Can we upload straight to docker hub given that Plausible doesn't use Gitlab at all at the moment?

I am not sure if there's any direct way to mirror Gitlab registry to Docker Hub. Since it is a fork of the plausible, for my testing (and might contain unstable and / or features that might not be part of upstream yet) uploading image to Gitlab makes sense but, once the PR is merged I believe travis can actually push into Docker Hub,  the official Docker image. 

> Also, I don't know if it's too late now but I'd prefer to revert the formatting commit and add it in separately. It makes it difficult to figure out what parts of the diff are actual changes as opposed to formatting.

Ouch. Sorry about the noise. I figured the project was already formatted and I pushed my changes with formatted files in _the very first commit_.  I can try to separate it but this will take little bit more time as there are many files. 

> Once this is ready to go, I think it would be best for me to host the app using the same Docker setup. This way there's a single setup to maintain and I'll be more motivated to work on this way of hosting.

Yes that makes sense. But, since you already have a stable setup, you might want to re-align the current setup according to the changes.

config/config.exs
tckb: Here's how I view environment and stages. May be we can align on this: 


| Type                   | Environment name | Releases                              | Notes                                                                                                                                                                                 |
|------------------------|------------------|---------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| development            | dev              | dev-<commit_hash>, like "dev-1a92n2bd | here you deploy the "development" version; new tests and unstable versions should be probably go here.  In some cases, you can assume this as a "staging" version of the application  |
| production             | prod             | release-v<sem_version>                | a stable version of the application; the version should be always stable                                                                                                              |
| pre-production/staging | prod             | master-<commit_hash>                  | a stable preview of the application                                                                                                                                                   |

ukutaht: That looks good to me. 

I would still add the `type` of environment (`development`, `production`, `pre-production/staging`) as a config option and send it to Sentry here. Or we could add an option like `SENTRY_ENV` to pass in.

Using Mix env in Sentry means that both production andd pre-production stages would send `prod` as their environment, making it more difficult to separate staging and prod errors. I use Sentry to debug errors in production and seeing staging noise there would slow the task.

lib/plausible_migration.ex
ukutaht: Sure, let's go with Release

ukutaht:
> once the PR is merged I believe travis can actually push into Docker Hub, the official Docker image.

Great, that does make it easier with the current setup.

> There are several uses and advantages of releases. Here's the official version: https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#why-releases

Cool, pre-compiling modules is enough reason to do it. Analytics servers get hammered constantly so improving warmup time is a worthwhile effort.

ukutaht:
Something that came to mind as well: Currently, we get the visitor country code from the `X-Country` header which is set by Netlify's CDN.

Since we don't control how Plausible will be deployed, we can't rely on custom headers set by the CDN. We probably need some sort of local geolocation database. There's also a feature request for more accurate location (I guess that means counties/cities). I'll research what the libraries/tools are available for this.

tckb:
> Something that came to mind as well: Currently, we get the visitor country code from the `X-Country` header which is set by Netlify's CDN.
> 
> Since we don't control how Plausible will be deployed, we can't rely on custom headers set by the CDN. We probably need some sort of local geolocation database. There's also a feature request for more accurate location (I guess that means counties/cities). I'll research what the libraries/tools are available for this.

Funny I was actually on suggest MaxMind. I did use it before and does provide pretty accurate results;

ukutaht:
Awesome. I can pick this is up probably first week on June, after the Clickhouse migration is complete.

Another thing that came to mind with Netlify: I use their custom header to get the remote client ip. See the code here: https://github.com/plausible-insights/plausible/blob/master/lib/plausible_web/controllers/api/external_controller.ex#L86

It's really hacky but it works for now. A more robust solution would be using the `X-Forwarded-For` header, potentially with this plug: https://github.com/ajvondrak/remote_ip

tckb:
> Another thing that came to mind with Netlify: I use their custom header to get the remote client ip. See the code here: https://github.com/plausible-insights/plausible/blob/master/lib/plausible_web/controllers/api/external_controller.ex#L86
> It's really hacky but it works for now. A more robust solution would be using the `X-Forwarded-For` header, potentially with this plug: https://github.com/ajvondrak/remote_ip

Using custom headers isn't  necessarily  _hacky_ , with `remote_ip` you can add `x-bb-ip` as additional search headers to `:headers` 

@ukutaht I cleaned up the PR, reverted the formatting change. I will add the formatting changes to a different PR.

config/config.exs
tckb: I see. This is already possible using the `ENVIRONMENT`  environment variable.  https://github.com/tckb/plausible/blob/master/config/releases.exs#L31 

This defaults to "prod" but, you can always change it to "staging" or "pre-prod"

ukutaht: Ah cool, I missed the `EVIRONMENT` variable in a release. That's great, if I also start using releases for staging and prod then I can just use that :)

tckb: Yes!

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+RUN apt-get update && \
+    apt-get install -y bash openssl --no-install-recommends&& \
+    apt-get clean autoclean && \
+    apt-get autoremove --yes && \
+    rm -rf /var/lib/{apt,dpkg,cache,log}/
+
+COPY .gitlab/build-scripts/docker-entrypoint.sh /entrypoint.sh
+
+RUN chmod a+x /entrypoint.sh && \
+    export TZ="Europe/Berlin" && \
```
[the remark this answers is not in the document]
tckb: Ah! the TZ creeped up from my a different project of mine. I'll switch it back to UTC.

tckb:
>  I've been working with it for almost a month, I can help integrate the new database into this infra setup if needed.

I will look into the changes, it would definately require changes to Dockerfile and docker-compose. It would be helpful if you push all your changes and have code-freeze until I make the changes or have the release/changes fixed date for the next push.  This will help in saving time for ping-pong updates and fix my timeline. 

>let me know if you agree with the next steps:
>Resolve conflicts
>Integrate Clickhouse
>Merge this PR

This is on my plate, I will take care of it. Need some time as I explained above.

>I will set up Travis and DockerHub for this repository

Yes, that will be good idea.

>Deploy a staging version using DockerHub.
>Deploy a production release using DockerHub.

I can help you in setting up CI/CD pipeline. We can discuss this offline. 

>Public alpha release of self-hosted Plausible. 

Yes, a blog-post would be awesome!

>Going forward I will keep a proper Changelog/Release notes

I think this makes sense and provide more visibility and could also help in actually defining  a timeline for your [roadmap](https://plausible.nolt.io/roadmap)

>conform to the release tagging system we agreed on

I always like [semantic versions](https://semver.org)

ukutaht:
Great!

I'm pretty much in a code-freeze for a few days as I'm deploying the new database and working on other performance-related things. There may be some light changes but we should not conflict.

As for Clickhouse, here's the configuration for it:
https://github.com/plausible-insights/plausible/blob/master/config/prod.exs#L74
And the docker setup for testing:
https://github.com/plausible-insights/plausible/blob/master/.travis.yml#L7

Let me know if there's any way I can help with this. I'm available for Twitter DMs and a call if needed :)

tckb:
> Great!
> 
> I'm pretty much in a code-freeze for a few days as I'm deploying the new database and working on other performance-related things. There may be some light changes but we should not conflict.
> 
> As for Clickhouse, here's the configuration for it:
> https://github.com/plausible-insights/plausible/blob/master/config/prod.exs#L74
> And the docker setup for testing:
> https://github.com/plausible-insights/plausible/blob/master/.travis.yml#L7
> 
> Let me know if there's any way I can help with this. I'm available for Twitter DMs and a call if needed :)

tckb:
@ukutaht  All requested changes are finished and is ready for review. I have integrated clickhouse and resolved the conflicts.  There's still pending _formatting_ PR . I will raise another one right after this one.

PR #83, opened 2020-06-13
Added SCHEME to HOSTING.md

makitsune:
Not being documented caused some confusion for me. The tracker link had references to http links so events weren't being sent because https sites aren't allowed to use http links.


ukutaht:
Thanks!

Issue #88, opened 2020-06-15
Option to not require authentication at all


kpsuperplane:
Our self hosted version is behind a network-level authentication wall already, it would be nice to not require additional accounts.

An easy way perhaps is to automatically sign into the default account and disable the sign out option?

P.S. Thank you thank you so much for supporting open source efforts <3

Issue #86, opened 2020-06-14
Option to disble registration


RealOrangeOne:
It'd be good to be able to disable registration on self-hosted instances, so people can't create accounts on private instances.

tckb:
This could be a quick fix.

Issue #85, opened 2020-06-14
Don't ship marketing site as part of self-hosted package


RealOrangeOne:
Running this self hosted is nice, but it'd be nice if it didn't contain all the marketting site components with it. Having the homepage either be the login prompt, or a list of public sites (configurable?) would be so much nicer!

Specifically, things which act a bit weirdly when self hosting:

- "Live Demo" button doesn't work unless there's a site setup for 'plausible.io'
- Loading up to marketing site for Plausible feels _weird_.
- Footer links don't really work

ukutaht:
Makes total sense 👍

Issue #26, opened 2020-01-15
Official Docker container at Docker hub


alexanderadam:
Hi folks,

thank you so much for your work. :pray: 
I just thought it would be nice to have [an official :whale2: Docker container](https://hub.docker.com/search?q=plausible&type=image) for easier trying out plausible on a local machine or server.

See also [`nolt.io` issue 56](https://feedback.plausible.io/56).

ukutaht:
Thanks for opening the issue alex,

At the moment I'm working very actively on the product itself. I haven't made it easy to self-host because it's a lot of work and it would slow down development. Adding documentation for self-hosting and a Docker image is something I can see happening in the future but but I'm not going to address it immediately. Hope you understand.

alexanderadam:
Hi Uku,

awesome, thank you for your fast response!
I wasn't expecting something immediately at all, so no pressure on that.
I just thought it would probably a good idea in general, so no need to rush. :wink:

tjomk:
@alexanderadam It is fairly easy to run the app locally if you have node and Elixir installed. It's based on my setup (I prefer to run database via docker and Elixir/Node apps locally). Here's a rough step-by-step:

1. Pull postgres docker container and run it: `docker run -it --name plausible -e POSTGRES_PASSWORD=postgres -e POSTGRES_USER=postgres -e POSTGRES_DB=plausible_dev`. If you have different database settings, modify `config/dev.exs` and look for the block that starts with `config :plausible, Plausible.Repo`. It defines database connection.
2. Install Elixir dependencies: `mix deps.get`. This should be run from the root of repository.
3. Run database migrations: `mix ecto.migrate`
4. Install frontend dependencies by going to the `assets` folder in the repo root and then running `npm install`.
5. Start the app: `mix phx.server`

If everything went well, you should be able to see access the app via http://localhost:8000.

Although keep in mind that as Uku mentioned, there's more to just starting the app locally. You'll need to dig a bit into the code to configure other aspects like e.g. email, landing page, trial, etc. But that should also get you going and set up for exploration :-)

alexanderadam:
Thank you for responding @tjomk,

I agree that this is fine for a _development_ setup but a Docker container is obviously much easier for other environments. A properly tagged image allows easy rollbacks and updates. Also it wouldn't require to have all of the build dependencies locally available.

These are some reasons why other analytics alternatives can simply be tested/installed with a simple `docker pull`. They have all some official Docker containers available:

[...]

PR #89, opened 2020-06-15
Selfhosted version improvements

### LINKED ISSUE #133: email server without authentication
Hi, thank you for creating/open-sourcing plausible.io!

It's just what I wanted for personal/hobby projects and very easy to setup / migrate to it (I think it took me less than 1 hour to set it up and migrate all my personal projects from google analytics)

Some questions popped up, most of them already have issues with the `self-hosted ` label, but one I couldn't find is: how do I have to set the environment variables if my email server doesn't require auth?

I noticed in [the config](https://github.com/plausible/analytics/blob/master/config/config.exs#L146) auth is set to `:always`. It appears there's also a `:if_available` setting [documented in bamboo_smtp](https://github.com/fewlinesco/bamboo_smtp/blob/39a27e7f9879aa4f494abba2d3275b82e93f59d1/lib/bamboo/adapters/smtp_adapter.ex#L27) -- assuming it does what I think, would it be possible to use that or, probably even better, make the option configurable?

Not sure how easy it is, but including a "send me a test email" button would be very helpful

### LINKED ISSUE #198: Setting db poolsizes for postgres and clickhouse gives me a kernel dum
Setting db poolsizes for postgres and clickhouse gives me a kernel dump;

I am not sure why its DATABASE_POOL_SIZE instead of DATABASE_POOLSIZE for postgres, but either way the kernel hiccups.

DATABASE_POOL_SIZE=20
CLICKHOUSE_DATABASE_POOLSIZE=20

> {"Kernel pid terminated",application_controller,"{application_start_failure,plausible,{{shutdown,{failed_to_start_child,'Elixir.DBConnection.ConnectionPool',{{badmatch,{error,{#{'__exception__' => true,'__struct__' => 'Elixir.ArgumentError',message => <<\"ranges (first..last) expect both sides to be integers, got: 1..\\"20\\"\">>},[{'Elixir.Range',new,2,[{file,\"lib/range.ex\"},{line,63}]},{'Elixir.DBConnection.ConnectionPool.Pool',init,1,[{file,\"lib/db_connection/connection_pool/pool.ex\"},{line,18}]},{supervisor,init,1,[{file,\"supervisor.erl\"},{line,295}]},{gen_server,init_it,2,[{file,\"gen_server.erl\"},{line,374}]},{gen_server,init_it,6,[{file,\"gen_server.erl\"},{line,342}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,249}]}]}}},[{'Elixir.DBConnection.ConnectionPool',init,1,[{file,\"lib/db_connection/connection_pool.ex\"},{line,22}]},{gen_server,init_it,2,[{file,\"gen_server.erl\"},{line,374}]},{gen_server,init_it,6,[{file,\"gen_server.erl\"},{line,342}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,249}]}]}}},{'Elixir.Plausible.Application',start,[normal,[]]}}}"}
> Kernel pid terminated (application_controller) ({application_start_failure,plausible,{{shutdown,{failed_to_start_child,'Elixir.DBConnection.ConnectionPool',{{badmatch,{error,{#{'__exception__' => true

tckb:
This PR will address some of the concerns raised during the self-hosted versions:

closes https://github.com/plausible/analytics/issues/26, closes https://github.com/plausible/analytics/issues/85, closes https://github.com/plausible/analytics/issues/86, closes https://github.com/plausible/analytics/issues/88, closes #133 

- [x] cleanup docker-compose 
- [x] separate migration step 
- [x] remove/skip plausible.io landing page for self-hosting setup 
- [x] optionally enable registration page
- [x] optionally enable sign-in page
- [x] add tests
- [x] remove landing page and other unused templates
- [x]  remove `Joken` dependenc
- [x]  fix the app page to show login/register instead of "trial period"
- [x]  fix config param inconsistency issue and invalid value to param - closes #198


ukutaht:
I still don't see much reason to keep it tbh. If someone has a fork and they're adding code for a landing page, it shouldn't be a big ask to also change the `Pagecontroller.Index` handler to render the landing page instead of doing redirection.

Instead of referencing an old landing page we don't use anymore, it would be better to reference the code for our [live landing page](https://github.com/plausible/website/blob/master/index.html).

The main reason I want to remove it is because our CSS is compiled based on the classes used in the actual HTML templates. By removing the landing page we remove some of Tailwind's CSS classes from being included in the final build. This would make the whole app lighter.

tckb:
> The main reason I want to remove it is because our CSS is compiled based on the classes used in the actual HTML templates. By removing the landing page we remove some of Tailwind's CSS classes from being included in the final build. This would make the whole app lighter.

Agreed. okay, so removing the templates at https://github.com/plausible/analytics/tree/master/lib/plausible_web/templates/page would suffice right?

ukutaht:
> Okay, I agree on the size part. So removing the templates at https://github.com/plausible/analytics/tree/master/lib/plausible_web/templates/page would suffice right?

Yeah, and changing the controller to redirect based on the login state here:
https://github.com/plausible/analytics/blob/master/lib/plausible_web/controllers/page_controller.ex#L25

tckb:
> Yeah, and changing the controller to redirect based on the login state here:

This is already handled by [auto_auth_plyg](https://github.com/tckb/plausible/blob/master/lib/plausible_web/plugs/auto_auth_plug.ex). 

I will remove unnecessary assets and templates to needed for self-hosting

ukutaht:
Cool 👍 I will merge once those are removed

ukutaht:
One more thing, let's make the landing page disabled by default.

tckb:
> One more thing, let's make the landing page disabled by default.
 
If I remove the template, then there wouldn't be any landing page :P  I am not sure if I understand this correctly

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,98 +1,14 @@
...
-  end
-
-  defp sign_token!(user) do
-    claims = %{
-      id: user.id,
-      email: user.email,
-      name: user.name
-    }
-
-    signer = Joken.Signer.create("HS256", "4d1d2ae6-4595-4d0b-b98a-8ca5b1f2095a")
```
ukutaht, line 49: Can remove the `Joken` dependency, this is the only place it was used.

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,98 +1,14 @@
...
-      redirect(conn, external: "https://plausible.nolt.io/roadmap")
-    end
-  end
-
-  def contact_form(conn, _params) do
-    render(conn, "contact_form.html")
-  end
-
-  def submit_contact_form(conn, %{"text" => text, "email" => email}) do
-    PlausibleWeb.Email.feedback(email, text) |> Plausible.Mailer.send_email()
```
ukutaht, line 81: Can remove the `def feedback` definition and the associated template from `PlausibleWeb.Email`

lib/plausible_web/templates/layout/app.html.eex
```diff
@@ -52,11 +52,8 @@
                 <% true -> %>
                   <ul class="flex" x-show="!document.cookie.includes('logged_in=true')">
                     <li>
-                      <div class="inline-flex">
-                        <a href="/login" class="font-medium text-gray-500 hover:text-gray-900 focus:outline-none focus:text-gray-900 transition duration-150 ease-in-out">Login</a>
-                      </div>
                       <div class="inline-flex rounded shadow ml-6">
-                        <a href="/register" class="inline-flex items-center justify-center px-5 py-2 border border-transparent text-base leading-6 font-medium rounded-md text-white bg-indigo-600 hover:bg-indigo-500 focus:outline-none focus:shadow-outline transition duration-150 ease-in-out">Start free trial</a>
```
ukutaht, line 59: I realize that showing 'Start free trial' doesn't make a lot of sense for self-hosters, but I feel like we should keep a button that says 'Sign up' or 'Register' (if registration is enabled in config).

lib/plausible_web/controllers/page_controller.ex
tckb: cool! I will remove it.

lib/plausible_web/controllers/page_controller.ex
tckb: @ukutaht  the feedback definition from page_controller is removed.  the template `PlausibleWeb.Email`  where is it located?

lib/plausible_web/templates/layout/app.html.eex
tckb: As per the new changes: 

_Landing page_ is disabled by default so, "/" will be redirected to the login page. 



image



There the "register" button is enabled _conditionally_  and will be redirected to the signup page. These changes only apply to the landing page(which is hidden, will _always_ direct to login)

lib/plausible_web/controllers/page_controller.ex
ukutaht: Function definition: https://github.com/plausible/analytics/blob/master/lib/plausible_web/email.ex#L103

And I was wrong, there's no template for that email as the content is just passed through. The function definition can go.

lib/plausible_web/templates/layout/app.html.eex
ukutaht: > These changes only apply to the landing page

They apply anywhere the default layout`layout/app.html` is rendered which includes the live demo and all public sites.
 The `Start free trial` can be seen in the live demo here:

Screenshot 2020-06-29 at 14 50 48

I want to make sure the navbar of the live demo and public sites has a button to the registration page. This is why I'm suggesting we make this button appear conditionally as on the login page.

tckb: Arm, you're right! I didn't even realize it (I didn't have a hosted version to test it, before). Makes sense. I will change "Start free trial" to "Register" if this available or, this will be login.

Issue #71, opened 2020-06-03
Self hosting on podman


afruizc:
First of all, thank you for the work on plausible and especially the recent effort on self-hosting.

I am using podman instead of docker to run the containers. When running the main container, I get the following error: 

```
# podman run --name plausible-eph --hostname plausible --pod plausible plausible:http db migrate
Loading plausible..
Starting dependencies..
Starting repos..
Running migrations for plausible
** (Postgrex.Error) ERROR 42704 (undefined_object) type "regnamespace" does not exist
    (ecto_sql 3.4.4) lib/ecto/adapters/sql.ex:593: Ecto.Adapters.SQL.raise_sql_call_error/1
    (elixir 1.10.3) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (ecto_sql 3.4.4) lib/ecto/adapters/sql.ex:686: Ecto.Adapters.SQL.execute_ddl/4
    (ecto_sql 3.4.4) lib/ecto/migration/runner.ex:343: Ecto.Migration.Runner.log_and_execute_ddl/3
    (ecto_sql 3.4.4) lib/ecto/migration/runner.ex:117: anonymous fn/6 in Ecto.Migration.Runner.flush/0
    (elixir 1.10.3) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto_sql 3.4.4) lib/ecto/migration/runner.ex:116: Ecto.Migration.Runner.flush/0
    (stdlib 3.12.1) timer.erl:166: :timer.tc/1
```

I am using [this bash script](https://gist.github.com/afruizc/8fb076cbf58f853dc58594a6a5f0cd27) to start all the containers. I always start the main container manually to make sure the db and clickhouse are up.

Please let me know if I need to add more detail.

ukutaht:
Hey, thanks for reporting.

Got my hands full trying to get the Docker version running in production at the moment. Will take a look at podman (had never heard of it before) once everything's working with standard Docker.

tckb:
@afruizc that doesn't look like _plausible_ specific issue. Something related to postgresql, can you try it with latest version of postgresql? 

PS: I have never used podman myself but looked interesting :)

makitsune:
Using the image `postgres:12.3-alpine` works perfectly for me!

tckb:
Great! @afruizc what is the finalized config? May be you can add here the complete working config here ? It could be beneficial to others

[...]

PR #90, opened 2020-06-18
Use a working postgres version for docker-compose

kolaente:
As stated in #71, the docker-compose example does not work with postgres `9.4` but with `12` this pr updates the default version to `12`.


ukutaht:
Thanks 💚 💙 💜 💛 ❤️

PR #94, opened 2020-06-19
Add extra volume for clickhouse server events

kolaente:
Without the volume, all analytics data is lost on container restart.


ukutaht:
Looks good, unfortunately it conflicts with your previous PR so you'll have to resolve these first :)

kolaente:
Whoops, I guess I should have checked that first :upside_down_face: 

Resolved, should be good to got now.

ukutaht:
Thank you!!!

Issue #230, opened 2020-07-17
A way to change admin password (not only a reset)


RealOrangeOne:
## Prerequisites 
- [x] I have searched open and closed issues to make sure that the bug has not yet been reported.

## Bug report

**Describe the bug**

As a user, I cannot change my password

**Expected behavior**

I can change my password.

**Screenshots (If applicable)**

![image](https://user-images.githubusercontent.com/6527489/87811261-45157c00-c856-11ea-807d-13824bc43293.png)

**Environment (If applicable):**
 - OS: Linux (Docker)

tckb:
At the moment, I have not looked into changing the password. The could be added as the next feature (this not a bug and this feature was never added) but this might have implication on the cloud version. @ukutaht  / @metmarkosaric  is this supported in the cloud offering?

metmarkosaric:
This is the way we recommend cloud users to change passwords at this stage: https://docs.plausible.io/reset-password

tckb:
@RealOrangeOne  If you have configured smtp, you can reset your admin password `<base_url>/password/request-reset`

RealOrangeOne:
Using the password reset flow is an idea, never thought of that, thanks!

I think I meant this issue more as a feature request than a bug anyway!

PR #233, opened 2020-07-21
Create the admin user as a separate command

RealOrangeOne:
This means the migrations can always be run on startup, meaning the database won't get out of sync with the running application. It also means users who don't set the credential variables in their compose files don't keep creating admin users every time the application starts.

I'm hoping at some point we can get to a point where the `setup` container isn't necessary, which is how i've setup plausible myself, but that'll come as a later PR.

Looks like the diff is a bit noisy due to trailing space cleanup, sorry!

### DIFF
--- a/HOSTING.md
+++ b/HOSTING.md
@@ -35,13 +35,13 @@ The repo supplies with a [Docker Compose](./docker-compose.yml) file and the sam
     docker-compose down
     ```
 
-- After the setup, you can start plausible as -- 
+- After the setup, you can start plausible as --
     ```bash
-    docker-compose up -d plausible 
+    docker-compose up -d plausible
     ```
-     after a successful startup (can take upto 5 mins), `plausible` is available at port `80`, navigate to [`http://localhost`](http://localhost). 
- 
-- stopping plausible --  
+     after a successful startup (can take upto 5 mins), `plausible` is available at port `80`, navigate to [`http://localhost`](http://localhost).
+
+- stopping plausible --
     ```bash
     docker-compose down
     ```
@@ -51,12 +51,12 @@ The repo supplies with a [Docker Compose](./docker-compose.yml) file and the sam
     docker volume rm plausible_event-data -f
     docker volume rm plausible_db-data -f
     ```
-Note: 
+Note:
 - #1 you need to stop plausible and restart plausible  if you change the environment variables.
 - #2 With docker-compose, you need to remove the existing container and rebuild if you want your changes need to be reflected:
     ```bash
     docker rmi -f  plausible_plausible:latest
-    docker-compose up -d plausible 
+    docker-compose up -d plausible
     ```
 ### Non-docker building
 It is possible to create a release artifact by running a release.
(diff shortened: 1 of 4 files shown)


tckb:
It needs some changes. will post soon!

tckb:
Cool! good PR 👍 

@RealOrangeOne  What is the your current requirement? 

First some thoughts from my side -

> migrations can always be run on startup,

Migrations are more or less used for "Settingup" / "preparing" the db and  isn't normally required to run on every startup. 

>I'm hoping at some point we can get to a point where the setup container isn't necessary

setup container is not (read "should not be") required for a production system. I would like to emphasize that the docker-compose as [mentioned in the docs](https://github.com/plausible/analytics/blob/master/HOSTING.md#up-and-running), is only meant as a mere starting point.  

>who don't set the credential variables in their compose files don't keep creating admin users every time the application starts.

Current migrations does not create admin/default user every time. Once it is created on the first run, the  `init_admin` will skip creating another time (as it should) check  [plausible_release.ex#L32 ](https://github.com/plausible/analytics/blob/master/lib/plausible_release.ex#L32) so separating it might not work as you intended. 

Instead, I would like to see a separate command for resetting admin password, this would resolve https://github.com/plausible/analytics/issues/230  what do you guys think?

RealOrangeOne:
> Migrations are more or less used for "Settingup" / "preparing" the db 

My understanding of migrations is that they setup the DB schema. Should said schema need to change, migrations would need re-running (although they'd only run things required to get from the current state to the new state).

> setup container is not (read "should not be") required for a production system. I would like to emphasize that the docker-compose as mentioned in the docs, is only meant as a mere starting point.

I totally agree, in my setup I avoided the need for a setup container entirely, and it's worked out great, and IMO much easier to comprehend what's going on. Hoping to contribute said process back at some point.

> Current migrations does not create admin/default user every time. 

This is the crux of the issue i'm trying to resolve. If you don't setup `$ADMIN_USER_NAME` etc, the migrations *will* create you a new admin user every time. For my use case, I don't want these set in the environment, as it's more things to sync up should I ever change email / password. Also means the `docker-compose.yml` file stays lighter, which is nice.

To avoid the DB schema becoming out of sync (whether forgetfulness or auto update), it'd be handy to run `entrypoint.sh db migrate` as part of the container startup, as the migrations would always be run. I've tested running them multiple times and it seems to work fine. Unfortunately this creates an admin user every time the container restarts, which is annoying. This fix prevents that, at the cost of an additional step at install time.

> Instead, I would like to see a separate command for resetting admin password, this would resolve #230 what do you guys think?

This feels like a separate requirement, but yes I think that'd solve #230.

StudioLE:
> I'm hoping at some point we can get to a point where the `setup` container isn't necessary, which is how i've setup plausible myself, but that'll come as a later PR.

Do you have a branch somewhere showing your WIP? I'm having issues (#243) with the self-hosted setup and am curious if your revised methodology has solved them?

RealOrangeOne:
@StudioLE I don't, although I can say I never hit #243.

ukutaht:
@tckb Are you recommending changes to this PR?

I'm in favour of merging it. For the cloud setup for example we don't want any admin user at all. I would like to be able to run migrations separately without creating an admin user.

tckb:
Lets go with it for now. We can figure out any issues later.

aaromp:
Take this comment with a grain of salt because I have very little idea of what I'm doing haha

This change seems to have caused an error for me here's a screenshot of the error that brought me here. 

Screen Shot 2020-07-30 at 9 21 18 PM

I reverted this change locally and have things up and running. I'm not sure what the issue with the change is specifically, though.

Thanks a bunch for making this this self-hosted option available.

PR #237, opened 2020-07-21
Expose a default port in container

RealOrangeOne:
This allows tools like traefik to detect the running port correctly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,6 @@ RUN mix release plausible
 FROM debian:bullseye
 LABEL maintainer="tckb <tckb@tgrthi.me>"
 ENV LANG=C.UTF-8
-ENV PORT=8080
 
 RUN apt-get update && \
     apt-get install -y bash openssl --no-install-recommends&& \
@@ -83,5 +82,5 @@ COPY --from=buildcontainer /app/_build/prod/rel/plausible /app
 RUN chown -R plausibleuser:plausibleuser /app
 WORKDIR /app
 ENTRYPOINT ["/entrypoint.sh"]
-EXPOSE 8080
+EXPOSE 8000
 CMD ["run"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,6 +65,7 @@ RUN mix release plausible
 FROM debian:bullseye
 LABEL maintainer="tckb <tckb@tgrthi.me>"
 ENV LANG=C.UTF-8
+ENV PORT=8080
 
 RUN apt-get update && \
     apt-get install -y bash openssl --no-install-recommends&& \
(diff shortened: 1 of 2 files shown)


tckb:
As I understand that you are talking about https://docs.traefik.io/providers/docker/#port-detection, this would have been nice but, the problem is that port where the server is running is not (may not be) _defined_ in the Dockerfile and a part of the environment settings, https://github.com/plausible/analytics/blob/master/HOSTING.md#environment-variables 

This is a typical scenario where the port can not be known before hand and is highly dependent on the deployment environment.

RealOrangeOne:
> As I understand that you are talking about https://docs.traefik.io/providers/docker/#port-detection

I am indeed.

Whilst yes the port is configurable, it does (AFAIK) have a default value, in which case having the default implementation support port detection is beneficial.

tckb:
Although it is  recommended changing the default port, its **8000**  [releases.exs#L7](https://github.com/plausible/analytics/blob/master/config/releases.exs#L7) and exposing doesn't actual publish the port and could serve as a type of documentation.  I am fine with it now but this issue might come up later.

ukutaht:
Not a massive fan of baking the default port into the image itself. When you change the `PORT` env variable, the exposed port won't match up with the one the server is bound to.

I guess it won't hurt though and it makes things work easier by default if you're using a tool that auto-detects.

Just a thought, can't you define the exposed port at runtime and achieve the same effect? https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose

tckb:
> the exposed port won't match up with the one the server is bound to.

Yes, this is bound to happen if you are changing the server port. But exposing port doesn't publish it. 

>I guess it won't hurt though and it makes things work easier by default if you're using a tool like Traefik.

From my understanding, this is _only a convenience_ in **traefik**, which is not the case in other proxy servers. 


@RealOrangeOne  this reminds me, did you check this https://docs.docker.com/compose/compose-file/#expose ? In your docker-compose, you can add expose port and traefik should detect the port right?

RealOrangeOne:
@tckb correct. That's my current solution for now, but would be nice not to need to do that. 

I don't believe it's just a traefik thing, it's useful for anything which uses this to detect which ports a container is using, without directly publishing them to the system.

tckb:
@ukutaht  what are your thoughts on this? I am not entire convinced of hard-coding the default port in official docker image, as exposing/publishing ports of server are dependent on the environment of user and should be addressed by the user. But, I also do not any issue for now.

oliver-kriska:
@RealOrangeOne  hi, you can create own Dockerfile for example:
```Dockerfile
FROM plausible/analytics:latest 
EXPOSE 8000
```
This will do FOR YOU what you was asked. You have to just build this image. If you use Docker Hub, their building image has setup which monitor parent/base image and it trigger build when this parent/base is changed.

Issue #243, opened 2020-07-26
Self-hosted setup fails to connect to Clickhouse


StudioLE:
### Precheck

Please note that this tracker is only for bugs. Do not use the issue tracker for help or support. [Our docs](https://docs.plausible.io/) are a great place for most answers, but if you can’t find your answer there, you can [contact us](https://plausible.io/contact). Thanks!

## Prerequisites
- [x] I have searched open and closed issues to make sure that the bug has not yet been reported.

## Bug report
**Describe the bug**
Clickhouse database creation/migration fails when following the steps for [docker self-hosted setup](https://github.com/plausible/analytics/blob/master/HOSTING.md).

```
docker-compose run --rm setup
```

**Expected behavior**
Expected an error free setup but instead errors are returned (shown in red) during the setup and the webserver returns 500 errors on a variety of pages.

**Screenshots (If applicable)**

![image](https://user-images.githubusercontent.com/193290/88475542-bc1ed480-cf28-11ea-9d61-224950c3e043.png)

**Environment (If applicable):**
Docker version 19.03.12, build 48a66213fe
docker-compose version 1.25.5, build 8a1c60f6
Ubuntu Server 20.04
Using default environment variables https://github.com/plausible/analytics/blob/master/plausible-variables.sample.env

tckb:
@StudioLE  those error message are expected while creating the database. Can you test it further by  following all steps?

StudioLE:
If I continue with the setup and launch plausible the webserver seems to work. I can login with the admin credentials configured in the `.env` but when I go to add a new site on the page following the "snippet" I get a 500 error.

![image](https://user-images.githubusercontent.com/193290/88477704-23915000-cf3a-11ea-9b8d-aaaff7f07f9c.png)


 Checking the logs with `docker logs **plausible container name**` show's it's filled with connection errors. Hence why I assumed that there was an issue with the Clickhouse setup/migration.

![image](https://user-images.githubusercontent.com/193290/88477657-d1e8c580-cf39-11ea-80b4-5799a4493d53.png)

tckb:
@StudioLE are you running the stock docker-compose? Can you share your compose file ?

samuel-p:
I had the same error setting up a self-hosted instance and solved it by adding the following environment variables to the setup and plausible container.

```
CLICKHOUSE_DATABASE_USER=default
CLICKHOUSE_DATABASE_PASSWORD=
```

StudioLE:
[...]

PR #244, opened 2020-07-26
Added clickhouse user and password (#243)

StudioLE:
Revised `.env` to include the missing env variables (#243)


ukutaht:
Thanks!

Issue #219, opened 2020-07-13
Alternative geolocation database


ColdHeat:
I saw Plausible on HN awhile ago and noticed that it used Maxmind databases for ip to country mapping. I had a similar issue myself and did not want to force users to download the database and attribute IP data to Maxmind for basic country dat

I wrote an open source country geoip database that perhaps could be useful: 
https://github.com/geoacumen/geoacumen-country

I am using it in my own projects and intend to make it automatically publish new versions every so often. 

I imagine this is more of something that self-hosters could use but I figured I would mention it.

ukutaht:
Thanks for the suggestion (and sorry for my slow response).

I'm definitely interested in packaging a freely licensed country database with Plausible so people don't have to bring their own. As you probably know, I can't distribute Maxmind's own database easily.

A couple questions:
* Are there any accuracy pitfalls with iptoasn.com ?
* Is there any freely available datasets that go to city level? More granular location data has been requested many times.

ColdHeat:
From my research and discussion with people, leveraging a mapping from an ASN to their IP ranges is more reliable than using the RiR's directly published data. 

I don't know the accuracy of iptoasn or the approach in general. But I'm willing to add more data sources if they can be provided to me. 

As for a city level database, I've been trying to do that myself as well and it basically seems impossible unless you use another company's data (and deal with the licensing) or if you build the dataset yourself.

ukutaht:
Thanks!

I'm thinking maybe it's possible to package your country database with Plausible and ask people to bring their own only if they want city level data.

So I'm in favour of packaging the geoacumen database and leaving a config option open to use a custom database. I don't have much motivation to work on this at the moment as we're already running MaxMind happily. Happy to merge a PR for this, I can't give a timeline for when I can pick it up myself.

Findus23:
Hi,

Just for your information: Since the Maxmind license change at the beginning of 2020, Matomo uses the free db-ip database by default:
https://db-ip.com/db/lite.php

It works like the Maxmind database, but is still licensed under a Creative Commons Attribution 4.0 International License.

ukutaht:
Thanks @Findus23 

I like this option more as it also includes cities, which is a highly requested feature for Plausible.

PR #249, opened 2020-07-27
added maxmindinc/geoipupdate container to self-hosted docker setup

samuel-p:
This allows to automatically setup and update a geoip database.


ukutaht:
Thanks for the PR!

What happens if one doesn't fill in their MaxMind account details here?

I think it makes sense to include a note about how to use this image to the hosting doc: https://github.com/plausible/analytics/blob/master/HOSTING.md#ip-geolocation

samuel-p:
> What happens if one doesn't fill in their MaxMind account details here?

The container will exit with the following error:

```
ERROR: You must set the environment variables GEOIPUPDATE_ACCOUNT_ID, GEOIPUPDATE_LICENSE_KEY, and GEOIPUPDATE_EDITION_IDS!
```

I also can add some details how to use this image in [`HOSTING.md`](https://github.com/plausible/analytics/blob/master/HOSTING.md#ip-geolocation).

tckb:
I think this is not reasonable, ip geolocation should remain optional. As of now, the docker compose  is usable without any additional external signups. The current change forces users to signup maxmid and agree to their license.  @samuel-p  if you could make this config param optional that’s great !

samuel-p:
As far as I know it is not possible to make a service optional in one `docker-compose.yml` file. Alternatively I could move the whole setup for that container and according config to `HOSTING.md` or create a separate `docker-compose-geoip.yml` which contains the setup including geoip, as an alternative to the default `docker-compose.yml`.

ukutaht:
Yeah I think the geoip container should be optional and things should work by default if you don't have Maxmind credentials at hand.

Just a random thought, it could be nice in the future to build a downloader/updater for the database into the web server itself. So we would just need to pass the credentials to the server and it would manage the database itself. 

For now, I like the idea of providing a separate docker-compose file and highlighting it in the `HOSTING.md` doc

tckb:
> it could be nice in the future to build a downloader/updater for the database into the web server itself...

the idea seems good for folks having maxmind's db and also would add value to cloud version.  personally, I would like to move away from Maxminds db as much as possible, I am not a big fan of their  license agreement and their new changes.  I got my 👀  on https://github.com/plausible/analytics/issues/219

samuel-p:
I created the separate `docker-compose-geoip.yml` and reverted the changes in the original `docker-compose.yml`. I also updated the `HOSTING.md` with the new file. Somwhow I messed up when trying to merge the upstream into my changes, but I this it should be fixed now. Can someone please check my changes again and merge them if it all fits so far?

ukutaht:
Looks good to me!

PR #262, opened 2020-07-31
Fix init_admin and self-hosted sample vars

fkrull:


ukutaht:
Thanks for your contribution!

PR #269, opened 2020-08-04
Sort screen sizes by count

RealOrangeOne:
Sorting by a fixed list does make quick referencing slightly easier, but having them ordered looks much nicer, as the other tabs are also ordered.

This may also come with a performance boost, as the ordering is being handed by Clickhouse.


ukutaht:
Thanks!

PR #275, opened 2020-08-09
Sefhosting improvements/fixes

### LINKED ISSUE #250: Broken links in footer when self-hosted
# Prerequisites

## Bug report
I have setup a self-hosted plausible instance using the official docker image [`plausible/analytics`](https://hub.docker.com/r/plausible/analytics) and integrated it one of my websites. I have noticed that there are many links in the footer which don't work, because they are not available in my instance. Only the community links work. See https://plausible.sp-codes.de/sp-codes.de

**Expected behavior**
The links should be removed for self-hosted instances and only valid links should be visible.

**Screenshots (If applicable)**
Bildschirmfoto von 2020-07-27 18-04-17

tckb:
- Fixed broken links  (closes #250 )


ukutaht:
Thanks!

PR #308, opened 2020-09-01
Fix for GeoIP container not starting when deploying Plausible

avimanyu786:
I was facing an issue with the GeoIP container not starting when deploying Plausible and so countries would not show up on the map. I have updated the **docker-compose-geoip.yml** file accordingly and have also added the default port number 1080(documented on maxmind/geoupdate's Docker Hub). Henceforth, it got fixed. Hope this helps. Thank you for Plausible!


ukutaht:
Thanks! Looks good to me :)

PR #326, opened 2020-09-25
Add documentation to Dockerfile about exposed port 

etienneCharignon:
This information is helpfull to users of the Docker image from Dockerhub when defining containers port mapping. 

It is also used by some automatic scripting like for example docker-gen (https://github.com/jwilder/docker-gen)


etienneCharignon:
I did not see that the point was already discued in #237. I'm closing

Issue #332, opened 2020-09-26
Refactoring opportunity


preslavrachev:
Hey guys,

You may decide to ignore this issue altogether. However, I have spotted a few places, where we could work towards making the code easier to read and simpler to comprehend.

Take, for instance the `if-else` block here:
https://github.com/plausible/analytics/blob/924c96f0a563432fffcb83c6c01cb04169d1a819/lib/plausible_web/controllers/stats_controller.ex#L17

Generally, nested blocks increase complexity. What I would do instead, is focus on the happy path inside the controller action, and take care of externalities in their respective components. IMO, the check whether a user's trial has expired could perfectly fit a separate Plug (e.g. `UserBillingPlug`) that you can wrap this controller action with. With that alone, you remove one level of indentation, i.e. code complexity.

Another trick I'd use is to avid `else` as much as possible. Take the inner ìf-else` block, for example:
https://github.com/plausible/analytics/blob/924c96f0a563432fffcb83c6c01cb04169d1a819/lib/plausible_web/controllers/stats_controller.ex#L20

If you focus on taking the case of NOT having page views inside the if, you wouldn't need the `else` at all. It will  just be the default.

If you accept supporters on your journey, I would gladly offer my knowledge in helping out with such refactoring suggestions ;)

Cheers!

ukutaht:
Hi. Thanks for the suggestions. Care to open a PR for this?

preslavrachev:
Yes, I'd be glad to. I'd try to do it in the coming days

PR #337, opened 2020-09-30
#332: Refactor stats controller

preslavrachev:
Addresses the code refactoring opportunities addressed in #332. 

NOTE: I had to change my original idea a bit, since, of course Elixir does not support early returns. The preferred way here is to introduce multiple overloading functions and choose the right one with pattern matching.

@ukutaht at the end, It's a matter of taste and code style. I leave the final decision to you. Please, keep in mind that while I added one test to guarantee that the conditions are satisfied, I would still love to see being tested in some sort of integration environment.

Cheers!

### DIFF
--- a/lib/plausible_web/controllers/stats_controller.ex
+++ b/lib/plausible_web/controllers/stats_controller.ex
@@ -5,38 +5,38 @@ defmodule PlausibleWeb.StatsController do
   alias Plausible.Stats.Query
 
   plug PlausibleWeb.AuthorizeStatsPlug when action in [:stats, :csv_export]
+  plug PlausibleWeb.UpgradeBillingPlug when action in [:stats]
 
   def base_domain() do
     PlausibleWeb.Endpoint.host()
   end
 
-  def stats(conn, _params) do
-    site = conn.assigns[:site]
-    user = conn.assigns[:current_user]
+  def stats(%{assigns: %{site: site, has_pageviews: true}} = conn, _params) do
+    demo = site.domain == base_domain()
+    offer_email_report = get_session(conn, site.domain <> "_offer_email_report")
 
-    if user && Plausible.Billing.needs_to_upgrade?(conn.assigns[:current_user]) do
-      redirect(conn, to: "/settings")
+    conn
+    |> assign(:skip_plausible_tracking, !demo)
+    |> remove_email_report_banner(site)
+    |> put_resp_header("x-robots-tag", "noindex")
+    |> render("stats.html",
+      site: site,
+      has_goals: Plausible.Sites.has_goals?(site),
+      title: "Plausible · " <> site.domain,
+      offer_email_report: offer_email_report,
+      demo: demo
+    )
+  end
+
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -5,38 +5,38 @@ defmodule PlausibleWeb.StatsController do
...
+  plug PlausibleWeb.UpgradeBillingPlug when action in [:stats]
 
   def base_domain() do
     PlausibleWeb.Endpoint.host()
   end
 
-  def stats(conn, _params) do
-    site = conn.assigns[:site]
-    user = conn.assigns[:current_user]
+  def stats(%{assigns: %{site: site, has_pageviews: true}} = conn, _params) do
```
ukutaht, line 14: What's setting the `has_pageviews` assign here?

preslavrachev: Good point you're asking. It's the default function below. This way of pattern matching with function overloading is quite a standard practice in the Elixir world, but I admit that it isn't to everyone's taste.

The first time a call to this function gets performed, it won't match. Instead, the default overload below would be called, which would perform the check and set `has_pageviews` handing over to `stats` once again. The second time, the first function will match.

It's a bit more code, but IMHO, it allows the reader to focus on the happy path, instead of using nested `if-else`. From the signature, one can also easily glance what condition matched for it to be called (`has_pageviews: true`).

Does it make sense?

ukutaht: I see, thanks for explaining.

I'm not a big fan of the overloading. While clever, it was unexpected for me and took a lot longer to decode than a normal if-else.

I do think extracting the `PlausibleWeb.UpgradeBillingPlug` makes things cleaner. I'd propose we keep the plug, the new test, and remove the function overloading to have a single `if-else` in the function. Thoughts?

preslavrachev: Agreed. I actually wanted to write a blog post on this topic, and your feedback is quite valuable. It is indeed a bit quirky.

I will revert the overloading and update the PR hopefully by the end of the day

preslavrachev: @ukutaht BTW, what is your opinion on pattern matching as part of function arguments (as opposed to assigning variables inside the function body)? I personally find this quite neat:

`def stats(%{assigns: %{site: site}} = conn, _params) do`

because it tells you right away what the function expects the argument to contain. I'd propose that we keep it in the PR too.

ukutaht: yeah, deconstructing is nice I think.

preslavrachev: Done. Brought the `if-else` back ;)

ukutaht:
Thanks!

preslavrachev:
> Thanks!

I'd love to keep helping you guys. I love Plausible, and am trying to contribute and make it even better 🚀

ukutaht:
Please do, contributions very welcome!

PR #340, opened 2020-10-01
Expose tracker script as `index.js`

RealOrangeOne:
### Changes

This enables self-hosted applications to be compatible with the pattern for supporting custom domains.

Custom domains serve the script from `index.js`, which if directed to plausible itself, 404s.

https://plausible.io/js/index.js currently 404s, so this won't impact the production setup.

This also adds the slight benefit of allowing self hosted instances to use a completely disguised script URL.

### DIFF
--- a/lib/plausible_web/router.ex
+++ b/lib/plausible_web/router.ex
@@ -37,6 +37,7 @@ defmodule PlausibleWeb.Router do
   end
 
   get "/js/plausible.js", PlausibleWeb.TrackerController, :plausible
+  get "/js/index.js", PlausibleWeb.TrackerController, :plausible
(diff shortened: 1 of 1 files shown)


ukutaht:
hey sorry for the no response. The tracker implementation has completely changed and this PR would not work anymore.

Instead of allowing both names, how do you feel about making the script name configurable with an environment varibale?

ukutaht:
Closing for now until a solution is agreed.

RealOrangeOne:
I'm not sure how useful it being configurable is. Although I guess if it's not much overhead making it configurable, it can't hurt to make it so

PR #352, opened 2020-10-10
Update container image to run as non root

BobyMCbobs:
# Goal

This PR removes extra dependencies in the container image to run as a non-root user, instead defaulting all processes to run as root from the entrypoint.
This is important, as an _exec_'s into the container image will not run as root, unlike currently.

### Changes

### Tests

### Changelog

### Documentation

### DIFF
--- a/.gitlab/build-scripts/docker-entrypoint.sh
+++ b/.gitlab/build-scripts/docker-entrypoint.sh
@@ -1,16 +1,11 @@
 #!/bin/bash
 set -e
 
-chmod a+x /app/*.sh
-
 if [[ "$1" = 'run' ]]; then
-      exec gosu plausibleuser /app/bin/plausible start
+      /app/bin/plausible start
 
 elif [[ "$1" = 'db' ]]; then
-      exec gosu plausibleuser /app/"$2".sh
- else
-      exec "$@"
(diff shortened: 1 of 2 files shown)


ukutaht:
Thanks for the PR!

> defaulting all processes to run as root from the entrypoint.

Sorry, as I understand this change is defaulting all processes to run as `plausibleuser` who is **not** root, correct?

BobyMCbobs:
> Thanks for the PR!
> 
> > defaulting all processes to run as root from the entrypoint.
> 
> Sorry, as I understand this change is defaulting all processes to run as `plausibleuser` who is **not** root, correct?

### Currently
When the container is started, the _root_ user hands off running Plausible code as the _plausibleuser_ user.
This means that any commands run with `docker exec` will run as _root_ in the container instead of _plausibleuser_.

In some environments, running any container as _root_ is not allowed due to security implications.

### To clarify
In this PR: from the entrypoint of the container, any process will be run as the user specified (_plausibleuser_) and nothing will run as _root_

---

References:
- https://docs.docker.com/engine/reference/builder/#user

ukutaht:
Thanks @BobyMCbobs ! 🙏

PR #402, opened 2020-11-10
Fix hashmode when using plausible's script with hash and outbound-links at the same time

### LINKED ISSUE #401: Cannot use hashMode and outbound-links at the same time
## Prerequisites

## Bug report
**Describe the bug**
https://plausible.io/js/plausible.hash.outbound-links.js only loads `outbound-links`. It does not load `hashMode`.

**Expected behavior**
Both `outbound-links` and `hashMode` should be loaded.

**Screenshots (If applicable)**
N/A

**Details**
The problem comes from the line below:

https://github.com/plausible/analytics/blob/039480a0d14186f2f3f13da030dd7a7c39f50673/tracker/compile.js#L21

```
{hash: true, outboundLinks: true}
```
 should be
```
{hashMode: true, outboundLinks: true}
```

Maronato:
### Changes

Fixes a typo on the tracker script compiler.

Resolves #401 

### Tests

### Changelog

### Documentation

### DIFF
--- a/tracker/compile.js
+++ b/tracker/compile.js
@@ -18,6 +18,6 @@ function compilefile(input, output, templateVars = {}) {
 compilefile(relPath('src/plausible.js'), relPath('../priv/tracker/js/plausible.js'))
 compilefile(relPath('src/plausible.js'), relPath('../priv/tracker/js/plausible.hash.js'), {hashMode: true})
 compilefile(relPath('src/plausible.js'), relPath('../priv/tracker/js/plausible.outbound-links.js'), {outboundLinks: true})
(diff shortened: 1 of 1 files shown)


Maronato:
Since [`1.1.2`](https://github.com/plausible/analytics/blob/master/CHANGELOG.md#112---unreleased) is still marked as *unreleased,* I don't think any changes need to be made to the changelog, otherwise this fix would belong in `1.1.3`.

What do you think, @ukutaht ?

ukutaht:
Thanks @Maronato ! Yes, I agree about the changelog.

PR #403, opened 2020-11-10
Update contribution guidelines

birjolaxew:
### Changes

Updated the contribution guidelines to better illustrate a first time setup. The list of system dependencies is mostly useful to avoid an error (with an unhelpful error message) when compiling the `:gettext` dependency if Erlang isn't installed. Users may still encounter some errors (e.g. requiring Elixir 1.10+), but there are friendly error messages for those.

The make file is updated to start Clickhouse and PostgreSQL in detached mode. This is just a quick usability improvement so both containers are ran in the background while you continue with the other steps.

(on a sidenote, big thanks to @Eptis for adding contribution guidelines; the lack of them have been a roadblock for me in the past)

### Tests
- [X] This PR does not require tests

### Changelog
- [X] This PR does not make a user-facing change

### Documentation
- [X] Updated CONTRIBUTING.md

PR #404, opened 2020-11-10
Change Docker volume used by Clickhouse to a local directory

birjolaxew:
## Changes

Just a quick update for an issue that I noticed after submitting #403 (darn you and your quick merging ;) ). Previously the Clickhouse container would use a volume mounted in the users home directory, if created using the Makefile. Polluting the home directory is no bueno, so this moves the directory into the current directory, and makes it hidden by default. It also adds said directory to `.gitignore`.

### Tests

### Changelog

### Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ plausible-*.tar
 # If NPM crashes, it generates a log, let's ignore it too.
 npm-debug.log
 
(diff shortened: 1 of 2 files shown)


ukutaht:
Good move, thanks!

PR #408, opened 2020-11-11
Support going back in history after updating period

birjolaxew:
### Changes

Previously going from a "default" period (i.e. one where it isn't specified in the URL search parameters, but is instead loaded from localstorage) to one where it *is* specified in the URL search parameters and then going back one step in history would not reset the period correctly.

This fixes that by implementing a navigation function which updates the current history entry before navigation happens, if that navigation would update the period.

<dl>
<dt>Before</dt>
<dd>

```
View dummy.site    -->    View 7day period      -->      Go back in history
Shows 30d of data         Shows 7d of data               Shows 7d of data

History:                  History:                       History:
0. /dummy.site            -1. /dummy.site                0. /dummy.site
                          0. /dummy.site?period=7d       +1. /dummy.site?period=7d
```

</dd>
<dt>After</dt>
<dd>

```
View dummy.site    -->    View 7day period      -->      Go back in history
Shows 30d of data         Shows 7d of data               Shows 30d of data

History:                  History:                       History:
0. /dummy.site            -1. /dummy.site?period=30d     0. /dummy.site?period=30d
                          0. /dummy.site?period=7d       +1. /dummy.site?period=7d
```

</dd>
</dl>

### Tests
- [ ] Automated tests have been added *(as far as I can see there are no tests for frontend)*

### Changelog

### Documentation


ukutaht:
Ha, definitely an annoyance I've noticed. Using `history.replace` the way you've done is quite clever I think. I tested it locally and it works beautifully, well done!

I will take some time to play around with the implementation. On one hand the current `navigateToQuery` function is nice and simple but using it with `<Link>` feels a bit cumbersome to me.. especially the fact that many Link components now have a `to` prop which is ignored and instead the `onClick` is taking over. 

I wish we had a component similar to `<Link>` that embedded in it all the behaviour introduced here. What do you think?

birjolaxew:
I definitely agree that the actual implementation is a bit cumbersome. I was playing around with a couple different implementations, but eventually settled on this slightly hacky navigation interception strategy - mostly so most of the code didn't have to change much.

Good call on creating a custom `<Link>` component just to clean things up a bit. Admittedly I didn't realize that links were a problem until I went through and checked that everything worked, so that part of the code definitely isn't the cleanest ;) I'll get a `<Link>` replacement implemented sometime today, hopefully.

On a sidenote I'm not *super* happy with this implementation. Ideally we'd be able to intercept route changes in react-router so we could update them regardless of how they were triggered, but I wasn't able to find anything of the sort when I looked. There is [`history.listen`](https://github.com/ReactTraining/history/blob/master/docs/api-reference.md#history.listen), but that unfortunately triggers after the route changes.

ukutaht:
Cool, I'm happy to merge as-is but I'd like to spend a bit of time to see if there's a simple way to make it a bit more elegant. I'm not a React wizard but I have some ideas.

On thing I'd like to try as well: instead of waiting until the user actually navigates, maybe we could just add a hook when the dashboard is mounting. If the `period` query param is not set in the URL, then `history.replace` immediately. So even if you land on `/plausible.io` it would immediately be replaced with `/plausible.io?period=30d`.

One problem with this approach would be an unnecessary re-render of the whole dashboard as it's mounting. If this can be avoided, then this might be a pretty nice approach as well. If it can't be avoided, I definitely prefer replacing the history when the user navigates like in your PR.

birjolaxew:
Sure, that'd be the simpler solution for sure. It would sacrifice those nice clean URLs tho ;)

Another solution I was playing around with is using the state you can set for each history change to keep track of what the localstorage state was before the change. That way you could reinstate it on each popstate, and it wouldn't have any side effects like changing the URL. Unfortunately it ended up being way too complicated to implement it with the declarative nature of React Router, so l abandoned it fairly quickly.

birjolaxew:
I took the liberty of fixing a minor issue I stumbled upon in this PR: links that only updated the search part of the URL previously failed to specify a pathname too. This meant that they e.g. linked to `/?path=/` instead of `/dummy.site?path=/` like they should. The links would still work, since the native [`history.pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) which React Router uses to update the URL accepts partial URLs, but React Router would generate the wrong URL which could cause problems if users open the link in a new tab.

ukutaht:
@birjolaxew Sorry I totally missed your QueryLink implementation! The update got lost in my notifications somehow.

I really like the new version. I think it's a good improvement on using `<Link onClick={navigateToQuery(...)}>`. Could you rebase `CHANGELOG.md` with master so the PR can be merged? I will merge once conflicts are resolved.

birjolaxew:
@ukutaht No problem. I merged in the changelog changes so it should be mergeable now.

ukutaht:
Thanks @birjolaxew !

PR #426, opened 2020-11-18
Add Cross-Origin-Resource-Policy header to script

gzuidhof:
Hey Plausible devs,

I am trying to embed the `plausible.js` onto a page that has the [`require-corp`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy) header set, which means that for every resource that is loaded they must be clearly marked as cross-origin OK.

The tracker script response currently doesn't have that header set, so I can't load it right now. This would solve that.

### Changes

### Tests

### Changelog
Note: if anything, it's more permissive.

### Documentation


ukutaht:
Makes sense, thanks! I will deploy it to our live server today.

gzuidhof:
Amazing, thank you!

PR #432, opened 2020-11-21
WIP: Feat/multi account sites

Eptis:
### Changes

### Tests

### Changelog

### Documentation

This is a first approach for allowing multiple accounts to access a site. This adds:
- Docker compose setup for easier local development
- The concept of an owner of a site. This is reflected as an owner_id in the sites table. A migration is included that makes all current accounts linked to a site the owner. A different migration adds roles (`admin` or `viewer`) to site memberships and makes the current owner an admin.
- A 'members' section within site settings. Here you can:
   - Update the role of existing members (functional)
  - Invite other users to your site (non functional)

Unfortunately I currently don't have a time to continue working on this, but don't want it go to waste. Also, the more I work on this the larger the impact this feature has on the application and I can't make these decisions. 

#### Some things to consider when adding access of multiple accounts to a site:
- What should an 'admin' be able to do? and what about a 'viewer'?
- How does inviting a user work? Is a new user account also in trial even though the site they are invited to is already being paid for?

### What is still missing:
- [ ] Sectioning off particular sections of the application to be only accessible by members/admins of the site.
- [ ] making the invite member form functional. Currently no invite is sent and no membership is created.
- [ ] Setting up an invite flow for users that don't have an account yet, and users that already have a plausible account.
- [ ] Creating an invite system which handles invite tokens to create the membership when accepted.
- [ ] Preventing an owner from removing themselves from the site. Or an admin removing the owner.
- [ ] Lots of tests

My recommendation would also be to split up the SettingsController into different modules. A lot of logic is currently embedded in the controller which might make it difficult to test and maintain.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Development setup
 
-The easiest way to get up and running is to [install](https://docs.docker.com/get-docker/) and use Docker for running both Postgres and Clickhouse.
+The easiest way to get up and running is to [install](https://docs.docker.com/get-docker/) Docker and [install](https://docs.docker.com/compose/install/) Docker Compose for running both Postgres and Clickhouse.
 
 Make sure Docker, Elixir, Erlang and Node.js are all installed on your development machine.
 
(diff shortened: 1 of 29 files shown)


ukutaht:
Thanks @Eptis !!

I will take a look at some point but closing now since it's not ready nor being worked on.

PR #470, opened 2020-12-14
Fix overly-sticky date in query parameters

### LINKED ISSUE #439: Selecting today doesn't work when on a different day
If you select Today, then click on the horizontal arrows to change day and then select Today, it won't bring you back to the current day.

Here I'm clicking on Today, but it doesn't move me because it's a link to the 20 of november.
image

maco:
### Changes

The date param was sticking around even when not specified
resulting in Last Month -> This Month still having last month's
date.

Fixes #439

### Tests
Frontend doesn't have a test framework

### Changelog

### Documentation


ukutaht:
Works great, thanks @maco!

PR #480, opened 2020-12-16
Adds "Load More" functionality to pages modal

### LINKED ISSUE #429: Make all the pages visible in the Top Pages list
"we’re customers at Plausible – I was wondering how I can see *all* pages. In the dashboard I’m only seeing top pages, that list seems to cut off pages that have a lower page view count. I’d love to see those, too. Is that possible?"

(feedback via email)

would be great it we could add the "load more" button in pages too so you can keep scrolling (same as we have for referral sources)

Vigasaurus:
### Changes

Adds "Load More" capability to pages modal - both top-pages and entry-pages. Allows for a user to get all the way to all pages, not just the top 100. 

Fixed #429 

For clarity: I was running into an issue with this setup where I would receive all `n` pages as expected, but I'd also consistently get a seemingly random `n+1` page sent, which was a complete duplicate of a previously sent entry. I suspect it was just an issue with how I was generating data, as even without my changes it was appearing in my local if I bumped the limit on the pages API call (and not the live demo or my self-hosted instance). But I figured I'd mention it in case you could see any code reason for that happening. 

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - Collect and display browser & OS versions plausible/analytics#397
 - Simple notifications around traffic spikes plausible/analytics#453
 - Dark theme option/system setting follow plausible/analytics#467
+- "Load More" capability to pages modal plausible/analytics#480
 
 ### Changed
 - Use alpine as base image to decrease Docker image size plausible/analytics#353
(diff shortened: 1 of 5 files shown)


ukutaht:
Tested, works perfectly fine on my machine. Thank you!!

Unrelated to this PR, I noticed a regression while testing: when opening a modal popup like the pages one, the dashboard layout seems to shift to the right by a few pixels. It's annoying and it didn't used happen. The issue is on master, I'm running a git bisect to find which commit introduced the problem.

PR #483, opened 2020-12-17
disable date navigation arrows when date is invalid for site

maco:
### Changes

Change date navigation arrows to real buttons and disable them when date is invalid for site

Keyboard navigation could still leave the valid zone, but I'm not sure what the correct usability for that would be. Perhaps a nnotification banner? Keyboard nav could imply a disabled user, so it would need to have correct ARIA attributes.

### Tests
no js test framework

### Changelog
Fits in with previous changelog entry

### Documentation


ukutaht:
Thanks!

What's the benefit of them being `button` elements as opposed to `a` links?

maco:
HTML `button` elements have a disabled state available, and `a` elements don't. That was the most relevant part in this case. Also, buttons are usable with space and enter, which is nice. But generally [using an actual button when you want it to act like a button](http://web-accessibility.carnegiemuseums.org/content/buttons/#:~:text=Buttons%20vs.,key%20only%20triggers%20a%20link.) is best practice for web accessibility rather than decorating a link.

ukutaht:
> HTML `button` elements have a disabled state available, and `a` elements don't. That was the most relevant part in this case. Also, buttons are usable with space and enter, which is nice. But generally [using an actual button when you want it to act like a button](http://web-accessibility.carnegiemuseums.org/content/buttons/#:~:text=Buttons%20vs.,key%20only%20triggers%20a%20link.) is best practice for web accessibility rather than decorating a link.

Fair enough. I considered them pagination elements that provide navigation which is why they are link at the moment. According to the Buttons vs. Links section of the linked document, navigation elements should be links. But these arrows also look like buttons and trigger javascript, so according to the doc it seems it's both a link and a button. I don't really mind either way, will give this a review.

assets/js/dashboard/query.js
```diff
@@ -93,6 +94,30 @@ class QueryLink extends React.Component {
...
+
+  onClick(e) {
+    e.preventDefault()
+    navigateToQuery(this.props.history, this.props.query, this.props.to)
+    if (this.props.onClick) this.props.onClick(e)
+  }
+
+  render() {
+    const { history, query, to, ...props } = this.props
+    return <LinkButton
```
ukutaht, line 111: Pretty sure this would work just as well with just an HTML `button` here. The only thing LinkButton customizes is the `onClick` handler which is never being called anyways since the click is handled in this component.

ukutaht (changes requested):
Looks good! The only change I would request is removing `LinkButton` and using `button` instead unless there's a good reason to have a `LinkButton`.

assets/.eslintrc.json
```diff
@@ -1,4 +1,7 @@
 {
+  "env": {
+    "browser": true
+  },
```
maco, line 4: So eslint won't error about `document.get...` stuff.

assets/js/dashboard/datepicker.js
```diff
@@ -1,147 +1,244 @@
...
+  formatMonthYYYY,
+  formatISO,
+  isToday,
+  lastMonth,
+  nowForSite,
+  isSameMonth,
+  isThisMonth,
+  parseUTCDate,
+  isBefore,
+  isAfter,
```
maco, line 18: these 3 and QueryButton are new
the rest of this import stuff is just formatting

assets/js/dashboard/datepicker.js
```diff
@@ -1,147 +1,244 @@
...
+    document.addEventListener("mousedown", this.handleClick, false);
   }
 
   componentWillUnmount() {
     document.removeEventListener("keyup", this.handleKeyup);
-    document.removeEventListener('mousedown', this.handleClick, false);
+    document.removeEventListener("mousedown", this.handleClick, false);
   }
 
   handleKeyup(e) {
```
maco, line 41: changes in this function prevent the keyboard arrow keys from going past time boundaries

assets/js/dashboard/datepicker.js
```diff
@@ -1,147 +1,244 @@
...
   }
 
   handleClick(e) {
     if (this.dropDownNode && this.dropDownNode.contains(e.target)) return;
 
-    this.setState({open: false})
+    this.setState({ open: false });
   }
 
   timeFrameText() {
```
maco, line 110: the changes in this function are linting changes

assets/js/dashboard/datepicker.js
```diff
@@ -1,147 +1,244 @@
...
+          >
+            <polyline points="9 18 15 12 9 6"></polyline>
+          </svg>
+        </QueryButton>
       </div>
-    )
+    );
   }
 
   renderArrows() {
```
maco, line 200: these are linting changes

assets/js/dashboard/datepicker.js
```diff
@@ -1,147 +1,244 @@
...
-    } else if (query.period === 'realtime') {
-      return 'Realtime'
-    } else if (query.period === 'custom') {
+    } if (query.period === 'custom') {
       return `${formatDayShort(query.from)} - ${formatDayShort(query.to)}`
     }
+    return 'Realtime'
   }
 
   renderArrow(period, prevDate, nextDate) {
```
maco, line 137: **actual changes here**

assets/js/dashboard/query.js
```diff
@@ -91,28 +89,53 @@ class QueryLink extends React.Component {
...
+        to={{ pathname: window.location.pathname, search: generateQueryString(to) }}
+        onClick={this.onClick}
+      />
+)
   }
 }
 const QueryLinkWithRouter = withRouter(QueryLink)
 export { QueryLinkWithRouter as QueryLink };
 
+class QueryButton extends React.Component {
```
maco, line 104: **actual change here**

maco:
I didn't make all the changes the linter suggested, mainly because I figured the diff would become unreadable if I did it all it in one go. But the automatic semi-colons and quotes/inverted-commas and no-else-after-return things, those are in here. I figure getting things formatted is a gradual thing.

ukutaht (changes requested):
Thanks for the updates!

I think dark mode was introduced since this PR was opened. Anyways I noticed that the Datepicker in this PR does not implement dark mode:

Screenshot from 2020-12-30 10-40-14@2x

Second, I noticed that on the month view, both buttons are disabled in all cases. Is this intentional? I don't see why we would disable arrow navigation for months.

Third, this is not a showstopper for this PR but my general take on formatting is that whatever checks the formatting should also be able to automatically format the code according to its own styleguide. I've found that when humans have to make linters happy, we end up arguing over style a lot.

With elixir code we enfore `mix format` for all code and now I've added `mix format --check-formatted` to our CI so our tests fail if a PR is not formatted properly. I've also added `mix format` in my personal pre-commit hook.

This need not be part of this PR, but I'd like to devise a similar system for JS/CSS. We can use an automated formatter to format the whole project and start enforcing that every piece of code needs to pass `npm run check-format`. If your code doesn't adhere to the style, it should be automatically fixable with `npm run format`. I'm not too familiar with the JS ecosystem, I guess prettier can do this?

maco:
Prettier does auto-format when it's just formatting, but those code checkers I put in before include more than just formatting. You know how Elixir will tell you "hey, you made this variable, but then you didn't use it"? ESLint tells you that kind of stuff too. It also has other elements of style, like preferring stateless functions over classes in React. Code organization, like alphabetize your functions. Accessibility stuff, like "you forgot to include an aria-label there."

I'll get to the dark mode for the buttons.

maco:
I did the dark mode and cleaned up a bunch of the linting erros. Here's what remains on these two files:

> /Users/mackenziemorgan/code/analytics/assets/js/dashboard/datepicker.js
>   137:3   error  renderArrow should be placed after openCalendar                          react/sort-comp
>   201:3   error  Expected to return a value at the end of method 'renderArrows'           consistent-return
>   226:14  error  Arrow function should not return assignment                              no-return-assign
>   291:9   error  JSX props should not use .bind()                                         react/jsx-no-bind
>   301:3   error  Expected to return a value at the end of method 'renderDropDownContent'  consistent-return
>   365:16  error  Arrow function should not return assignment                              no-return-assign
>   367:11  error  JSX props should not use .bind()                                         react/jsx-no-bind
>   
>   /Users/mackenziemorgan/code/analytics/assets/js/dashboard/query.js
>    46:3   error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax
>    47:17  error  Unary operator '++' used                                                                                                                                                   no-plusplus
>    95:9   error  Prop spreading is forbidden                                                                                                                                                react/jsx-props-no-spreading
>   105:1   error  Component should be written as a pure function                                                                                                                             react/prefer-stateless-function
>   114:52  error  'e' is not defined                                                                                                                                                         no-undef
>   128:17  error  Expected to return a value at the end of function 'toHuman'

Some of that involves rearranging order of functions, some involves actual changes to the logic, so I'm leaving that stuff out of this PR. Might tackle it separately, though.

ukutaht (approved):

PR #484, opened 2020-12-17
Add eslint, stylelint, prettier, and pre-commit

maco:
### Changes
- When you try to commit, Pre-commit will run checks for all three, if
  installed, _only_ on staged files, and prevent committing if the linters are
  unhappy.
- Several eslint rules that generate a lot of errors are disabled for now.
  That's to allow a more gradual transition, so you won't change 1 line and be
  told to fix 100 other lines (but maybe like 20 because there are still a lot
  of rules enabled).
- Prettier is set to require pragma. That's also to allow a more gradual
  transition. As each file is tidied up, run Prettier on it, and it'll add a
  special comment that tells it that it's now responsible for keeping that one
  tidy.

Oh, and Pre-commit supports custom hooks, too, though there is already one for `mix format` and `mix test`. It's the language-agnostic choice for git pre-commit hooks.
### Tests

### Changelog

### Documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v2.2.1"
+    hooks:
+      - id: prettier
+        files: "assets/js|assets/css"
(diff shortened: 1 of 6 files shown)


ukutaht:
Thanks!

PascalPixel:
The prettier config on this is broken currently, for whatever reason, this fixes it; https://github.com/plausible/analytics/pull/2150

PR #486, opened 2020-12-17
Adds manual refresh button in historical dashboard

### LINKED ISSUE #406: refresh button
A simple refresh button so that you don't have to refresh all the page every time (I think on the right of the date range select it would fit well) - here is the icon: https://heroicons.dev/?query=refr

image

Vigasaurus:
### Changes

Title. Helps reduce page reloads/query changes just to refresh data.

This PR is build on top of #409 to help lay out the sticky dashboard coherently with this new button (and its interaction with the pinning button). This should only be merged after #409. 

Fixes #406 

### Tests

### Changelog

### Documentation

### Screenshots

image


ukutaht:
@Vigasaurus I haven't looked at this yet but we have discussed this request with Marko before and we were both against adding a refresh button. The browser already has a refresh button and we don't think it makes sense to replicate standard browser behaviour. I'd rather make sure Plausible works great with the browser refresh than implement a workaround.

I'll review some other stuff first and then come back to this one.

Vigasaurus:
That makes sense! I'll see if I can come up with a sensible way to interact with the basic refresh.

PR #490, opened 2020-12-19
fix: low contrast text in dark mode

vincentdoerig:
### Changes

Change color of link in dark mode because of low contrast.

before:
link with low contrast

after:
link with enough contrast

### Tests

### Changelog
- [x] This PR does not make a user-facing change (it is a user-facing change but it is not worth adding imo)

### Documentation


Vigasaurus:
👍 I'm sure there are actually a few more instances of this exactly. I didn't start swapping the indigo-700 to the 500 until I was about halfway through my dark mode pass, and likely missed some on my way back through.

ukutaht:
@vincentdoerig Thanks!

PR #491, opened 2020-12-20
Changes "This Month" view to Month-to-date for current month

### LINKED ISSUE #391: Change "This Month" to display Month To Date view
"This Month" currently displays the full month which is not that useful, especially in the first half of the month which leaves most of the top graph empty and the comparison wrong too. If "This Month" simply shows month to date view rather than the full month would be much more useful. 

(request from [Twitter](https://twitter.com/nbgoodall/status/1322149369378938882))

Vigasaurus:
### Changes

Updates the "This Month" view to be called "Month to Date", and properly update `query` and `prevQuery` values to reflect that. 
Fixes #391 

I tried to get the same working for "Day to hour" or whatever it'd be called, but I was unable to get the query object to properly take a `DateTime` range structure, it only seems to function with `Date.range()` - but I did leave all the progress I had there in the initial commit, if you know an easy way to make that conversion work. 

This functions in the way that Marko wanted/researched, where if today is 12/20/2020, the MTD view would compare data from 12/01/2020-12/20/2020 to 11/11/2020-11/30/2020. See screenshots below for the updated `query` and `prevQuery` objects respectively, for months December and November, respectively. 

I didn't do the graph truncation here - but if it's wanted it's pretty trivial to add either from the data sent by Clickhouse or in React directly. 

### Tests

### Changelog

### Documentation

### Screenshots

image
image


lib/plausible/stats/query.ex
```diff
@@ -6,7 +6,19 @@ defmodule Plausible.Stats.Query do
     Map.put(query, :date_range, Date.range(new_date, new_date))
   end
 
-  def shift_back(query) do
+  def shift_back(%__MODULE__{period: "month"} = query, site) do
+    {new_first, new_last} = if Timex.compare(Timex.now(site.timezone), query.date_range.last, :day) <= 0 && Timex.compare(Timex.now(site.timezone), query.date_range.first, :day) >= 0 do # if the current day is within the original query
```
ukutaht, line 10: If we assume that `query.date_range.first` is the first of the month, we can simplify this conditional:
```elixir
if Timex.compare(Timex.now(site.timezone), query.date_range.first, :month) == 0  do # Querying current month to date
```

ukutaht (changes requested):
I'd like to simplify that conditional but otherwise looks good. This will be a great improvement! Happy to leave the day-by-day comparison for a future PR, it does require more changes.

ukutaht (approved):

ukutaht:
Thanks!

PR #494, opened 2020-12-21
Allows middle-click and ctrl+click outbound clicks to properly open as user intended

### LINKED ISSUE #451: Outbound link tracking breaks shortcuts to Open in New Tab
## Prerequisites

## Bug report
**Describe the bug**
When I use the outbound link tracking version of the Plausible script, outbound links no longer open in a new tab when I Ctrl-Click or Cmd-Click them.

Instead, the current tab navigates to the destination after the event fires, the same as if I had clicked the link without any modifier keys.

**Expected behavior**
Normally, when I Ctrl-Click or Cmd-Click a link in my browser, it opens in a new tab. I expect outbound links on my site to behave this way too, even when Plausible is tracking the clicks.

**Environment (If applicable):**
 - OS: Windows and macOS
 - Browser: Firefox
 - Version: 83.0b9

Vigasaurus:
### Changes

Allows middle clicks or ctrl+click, cmd+click, shift+click to track to plausible, and open in a new tab if requested, instead of always redirecting the current tab. 

Fixes #451 

This will almost certainly cause conflicts with #489 in the compiled script, so I didn't push them in this PR at all. Once both are merged we can re-compile the merged version. 

### Tests

### Changelog

### Documentation

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -70,10 +70,9 @@
 
         // Delay navigation so that Plausible is notified of the click
         if(!link.target || link.target.match(/^_(self|parent|top)$/i)) {
-          if (middle || event.ctrlKey || event.metaKey || event.shiftKey) window.open(link.href)
-          else {
+          if (!(event.ctrlKey || event.metaKey || event.shiftKey) && click) {
             setTimeout(function() {
-              if (click) location.href = link.href;
+              location.href = link.href;
             }, 150);
             event.preventDefault();
           }
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -58,21 +58,25 @@
   {{#if outboundLinks}}
   function handleOutbound(event) {
     var link = event.target;
+    var middle = event.type == "auxclick" && event.which == 2;
+    var click = event.type == "click";
       while(link && (typeof link.tagName == 'undefined' || link.tagName.toLowerCase() != 'a' || !link.href)) {
         link = link.parentNode
       }
 
       if (link && link.href && link.host && link.host !== location.host) {
-        if ((event.type == "auxclick" && event.which == 2) || event.type == "click")
+        if (middle || click)
         plausible('Outbound Link: Click', {props: {url: link.href}})
 
         // Delay navigation so that Plausible is notified of the click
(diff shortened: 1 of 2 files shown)


tracker/src/plausible.js
```diff
@@ -56,23 +56,34 @@
   }
 
   {{#if outboundLinks}}
-  function registerOutboundLinkEvents() {
-    document.addEventListener('click', function (event) {
-      var link = event.target;
+  function handleOutbound(event) {
+    var link = event.target;
+    var middle = event.type == "auxclick" && event.which == 2;
+    var click = event.type == "click";
```
ukutaht, line 62: What's going on with the spacing here? Seems like this line is 1 space to the left of the next line which cannot be correct.

tracker/src/plausible.js
```diff
@@ -56,23 +56,34 @@
...
 
       if (link && link.href && link.host && link.host !== location.host) {
+        if (middle || click)
         plausible('Outbound Link: Click', {props: {url: link.href}})
 
         // Delay navigation so that Plausible is notified of the click
         if(!link.target || link.target.match(/^_(self|parent|top)$/i)) {
-          setTimeout(function() { location.href = link.href; }, 150);
-          event.preventDefault();
+          if (middle || event.ctrlKey || event.metaKey || event.shiftKey) window.open(link.href)
```
ukutaht, line 73: What does this line do? When someone uses middle click on `cmd+click`, I would expect we do nothing here and allow default browser behaviour.

ukutaht: Oh, the if-else statement is formatted inconsistently

ukutaht (changes requested):
Nice way to account for middle clicks as well. Looks like formatting is a bit off and I'm curious what the purpose of `window.open` is when middle click or modifier is used.

Vigasaurus:
So the reason for using window.open is that browsers interpret middle clicks as "Open in New Tab", the same as ctrl+click or cmd+click - and in the old way we'd always redirect the current tab to the new link. This way ensures that if a user does any of the actions that would normally redirect their click into a new tab (middle click, ctrl+click, cmd+click, or shift+click) - we open a new tab for the link as expected, as opposed to redirecting the current tab.

ukutaht (approved):

PR #495, opened 2020-12-21
Fixes couple of Flatpickr related issues

### LINKED ISSUE #466: Choosing same day for start and end in Flatpickr results in an empty visitors-graph
## Prerequisites
- [X] I have searched open and closed issues to make sure that the bug has not yet been reported.

## Bug report
**Describe the bug**
If a user selects the same day from Flatpickr for both the start and the end, the stats all work as expected, but the graph renders as zero width (with only one x-axis entry) - resulting in a seemingly empty graph/no data, while data does exist.

**Expected behavior**
The graph should render potentially a straight line - with the leftmost and rightmost x-axis entries being the same day, or potentially just a single dot centrally aligned for the day, as to not imply to a user that there is no graph data to show. 

**Screenshots**
image
image

**Environment:**
 - OS: [e.g. Windows]
 - Browser [e.g. Firefox]
 - Version [e.g. 84]

Vigasaurus:
### Changes

- Ensures that selecting the same day for both entries in Flatpickr doesn't result in an empty graph, and instead changes to "day" view for the single day selected. 
- Does the same as #470 but for `from` and `to` params, which persist when moving from a custom date range to any other period

Fixes #466 

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ All notable changes to this project will be documented in this file.
 - Fix URL decoding in query parameters plausible/analytics#416
 - Fix overly-sticky date in query parameters plausible/analytics/#439
 - Prevent picking dates before site insertion plausible/analtics#446
+- Fix overly-sticky from and to in query parameters plausible/analytics#495
+- Adds support for single-day date selection plausible/analytics#495
 
 ### Security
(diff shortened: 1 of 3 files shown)


ukutaht:
Thanks!

ukutaht:
This is now deployed live

PR #500, opened 2020-12-22
Improves realtime top stats and conversion rate calculation

### LINKED ISSUE #444: Realtime Conversion Rate wrong
The realtime conversion rate is strange, it calculates _current_ visitors divided by unique conversions in the last 30 minutes. 
In my opinion it should be unique users in the last 30 minutes divided by unique conversions in the last 30 minutes.
Screen Shot 2020-12-01 at 12 14 32

Vigasaurus:
### Changes

- Adds `Unique Visitors (Last 30 min)` as a top stat in realtime view to help clear up goal conversions reasoning. 
- Goal conversion numerator appropriately uses last 30 minute unique visitors instead of realtime visitors. 
- Fixes  #444

I figured the first point here would be important to show when using it as a numerator in a calculation, but if you think it's unnecessary lmk. 

### Tests

### Changelog

### Documentation

### Screenshots
image


ukutaht:
Thanks! works great

PR #503, opened 2020-12-23
Mobile/Filter Overflow UX Improvements

Vigasaurus:
### Changes

- Adds Filter dropdown when appropriate
- Fixes Flatpickr year selection UI issues (black text in dark mode)
- Fixes Flatpickr placement on desktop and mobile
- Fixes Date selection dropdown placement on mobile
- Improves mobile visibility and click-ability on all dropdowns (site switcher, filters, date selection) by bumping text size
- Reduces padding on mobile sticky nav to make it less tall
  - I tried really hard to make the one-line solution work, but the only real way to do it would be to strip everything down to just icons in mobile mode (site switcher being just a favicon, a filter icon, and a clock for date selection), making them not only harder to click, but also requiring at least one click to figure out what they really mean. In my opinion this kills a lot of the intuitive nature of the UI - but lmk if that's something you want to pursue further. 
- Actually fixes the sticky date parameters from #495 and #470
  - These were causing the issues with the `date undefined` error that Marko found (and cause a major issue on currently deployed prod where if you remove a filter with any of the three fields in the query `from, to, date` - they get removed no matter what)
- Fixes a lot of the UI not maintaining scroll position issues
  - I couldn't find the root causes of all of them so it could still happen occasionally, but it seems to be jumping then resetting back to where everything was when it finishes loading most of the time. Other times (usually when using the map container) it actually moves and stays moved. 🤷‍♂️ 
- Couple other minor UI/UX changes - mostly related to navigation/that region

### Tests

### Changelog

### Documentation


Vigasaurus:
@ukutaht if you're still around, I'd love your thoughts on the filter dropdown implementation - https://dev.vigneshjoglekar.com/plausible.io?period=30d&country=USA&browser=Chrome&source=Hacker+News&browser_version=87.0 - it's not done as far as styling/spacing on mobile goes, but the actual collapsing logic should be all done. (I do still have a chunk of refactoring etc to do, so the code is likely not close to its completed state).

ukutaht:
Hey, I'm still around today and thanks for this. I get this issue when testing:

Screenshot from 2020-12-23 11-34-53@2x

Vigasaurus:
Yeah I found some other issues, gonna poke around - will let you know when its ready again

Vigasaurus:
Alrighty - it *should* be working well now, or at least I haven't been able to break it yet.

ukutaht:
Thanks @Vigasaurus. I think it behaves correctly, no issues there.

I've been looking at the code for about 30mins and I'm still completely confused by it. Can you try to refactor the it to add more clarity? I'll probably go over it as well once you're done.

metmarkosaric:
Thanks @Vigasaurus! I tested it on my phone. I like the new filters dropdown. Takes so much less space.

In the custom range picker the font for the year is black. The custom date picker popup goes outside of the screen for me. I wonder if there's a way to center it like the rest of the app?

photo_2020-12-23_11-29-45

I get this error when I clear filters if I'm on a custom selected date range. From what I've seen, it doesn't happen when I'm not on custom range.

photo_2020-12-23_11-29-01

I wonder if there's a way to make the sticky header take one line only on mobile devices? Right now the second line is half empty only showing the date. Perhaps retitle "Active Filters" to "Filters" on mobiles save space? And maybe replace the actual dates with a calendar icon or something like that on mobiles?

photo_2020-12-23_11-28-58

What do you think?

Vigasaurus:
Yeah absolutely, I'm planning on it. Most of the code right now is pretty messy and has a lot of unnecessary/unused stuff and is lacking some pretty vital comments to make stuff clear. My biggest thing here was just to verify functionality :D

Vigasaurus:
@metmarkosaric yeah I did notice the black year there too, I'm pretty sure it's just that I didn't account for the disabled selector (I'll add it). 

As for centering the popups - I am planning on the for sure. On mobile I intend to make them generally bigger and centered to try to make them more similar to native dropdowns instead of the desktop style. 

As for the error with the date, I haven't seen that before but I'll definitely try to reproduce it/see if I can find a root cause/if I caused it. 

As for making it a one-liner, I'm unsure - because right now your view is also a bit compressed since it's missing the down arrow for site switching. I can look into it for sure, but I have a hunch that it'll 1. get very overcrowded a d condensed and 2. not play very well with long domain names, and I think that truncating there could cause problems for the with multiple similar domains.

metmarkosaric:
Ok I'll try and test it with logged in state and other domain names when it's on staging. In general the feedback was that sticky header takes too much space on mobile devices so if there are elements we can hide on small screen to make it fit in one line, it's worth considering.

Vigasaurus:
Yeah that makes sense for sure, I just fear that in an effort to save space we may inadvertently remove some of the intuitive nature of Plausible's UI. I'll definitely explore it and see what solutions I can come up with.

Vigasaurus:
@ukutaht Code refactored, rest should be all ready for review. Definitely lmk if you want more cleanup/explanation in `filters.js`.

@metmarkosaric All the things you mentioned should be fixed, check out the PR main body for details

It's still live on my dev server, so let me know if you see any other issues :D

metmarkosaric:
Just tested it and it looks great @Vigasaurus, thanks! We could try and get this onto staging so we can test with different domain names etc and see if it all could fit on one line somehow.

assets/js/dashboard/filters.js
```diff
@@ -1,93 +1,209 @@
...
+import React, { useState, useEffect, useLayoutEffect } from 'react';
 import { withRouter } from 'react-router-dom'
-import {navigateToQuery, removeQueryParam} from './query'
+import { countFilters, navigateToQuery, removeQueryParam } from './query'
 import Datamap from 'datamaps'
+import Transition from "../transition.js";
+import { useScreenClass } from './screenclass-hook';
 
-function filterText(key, value, query) {
+const filterText = (key, value, query) => {
```
ukutaht, line 8: Why the new style of declaring functions? It's inconsistent with the rest of the project.

ukutaht (commented):
I had a question about the function declaration style. Also, all the other React components are defined as classes in this project (functions only for pure components). I'm not a fan of mixing multiple different styles.

Overall this functions well and I'm eager to merge.

On the code level I think the `screenclass-hook` and `recheck` variable add unnecessary complexity. I'd like to get rid of them if possible. Will take a look on my machine to see if it's possible

Vigasaurus:
Yeah that all makes sense. My biggest reason for using functional here was from how I was able to think up the functionality it needed/the ease in implementation that functional could afford, but I bet it is possible to do this in a class component. I'll take a look at that and removing the complexity you mentioned too and get it refactored to match 👍

ukutaht:
nice @Vigasaurus, thanks for this!

PR #510, opened 2020-12-24
Fix NaN issue in percent_change

mertbulan:
### Changes

Hello! Today, I saw that if there is no change in unique visitors (and probably also for the other ones), instead of showing `0`, the app shows `NaN`. This PR fixes this issue.

Screenshot 2020-12-24 at 16 28 53

### Tests

### Changelog

### Documentation


mertbulan:
Seems like this is an expected behaviour. I thought because of `0/0`, we get an error and because of that `N/A` is shown.

PR #556, opened 2021-01-07
Make Prettier and ESLint a lot happier

maco:
### Changes

I ran `eslint --fix` and `prettier --write`, which changed a lot on their own. Also tweaked the
configs for them as I went through the more manual changes. 

There are still a ton of more manual
changes to do for ESLint, but they're things like "use a stateless function instead of a class" and "add all
the accessibility stuff this component needs" and "stop using bind that way," which really do need
to be manual.

### Tests
no frontend test suite (...yet...I might get there)

### Changelog

### Documentation


TkDodo:
@maco any chance we can get this merged? Solving the conflicts likely means discarding the changes and running prettier / eslint again. As it stands, I find it very hard to contribute without prettier / eslint 😅

If you don't want to continue here, I'm happy to take over the PR (not sure what's technically the best way to do this though)

TkDodo:
also, not sure how the build process works, but having a CI check that make sure that prettier and eslint are not failing would ensure that future changes would adhere to this. Could be part of the "testing" pipeline...

ukutaht:
Yes I totally dropped the ball on this and now it seems the project has moved on a lot. I appreciate your work here @maco and I'm sorry I didn't merge this when the time was right.

If nobody else does it I will at some point fix Prettier and Eslint for the whole project and add it to the CI pipeline.

maco:
I'm gonna close this and start over.

PR #558, opened 2021-01-07
Dialyzer and Credo checks

oliver-kriska:
### Changes

- added Dialyzer
- fixed Dialyzer errors
- added Dialyzer check to GitHub Actions with cache
- added Credo
- fixed Credo Warnings
- added Credo Warnings check to GitHub Actions with cache
- added compile warnings check to GitHub Actions
- reformated GitHub Actions YAML
- upgraded deprecated version
  Upgraded:
    appsignal 2.0.5 => 2.0.7
    certifi 2.5.2 => 2.5.3
    hackney 1.16.0 => 1.17.0
    idna 6.0.1 => 6.1.1
    parse_trans 3.3.0 => 3.3.1
    unicode_util_compat 0.5.0 => 0.7.0 (minor)

First run takes about 12 minutes but it takes less than minute after that

### Tests

### Changelog

### Documentation


lib/plausible/application.ex
```diff
@@ -40,8 +42,7 @@ defmodule Plausible.Application do
   def report_cache_stats() do
     case Cachex.stats(:user_agents) do
       {:ok, stats} ->
-        IO.puts("User agent cache stats:")
```
ukutaht, line 43: I specifically used `IO.puts` here because it should print these stats regardless of the configured log level. Is there a way to keep that behaviour?

ukutaht (approved):
Overall love the changes here and I'm a fan of adding more strict checks in the CI as well. Thanks for this!

lib/plausible/application.ex
oliver-kriska: I will check it but I think IO is not recommended in prod. We can add some skip config but... Also I like `pretty: true` but I didn't find easy way to do it with Logger. https://hexdocs.pm/logger/Logger.html#module-metadata

PR #603, opened 2021-01-14
Self hosted improvements

ukutaht:
Prepping for v1.2


patschi:
If I'm allowed to add my 2 cents here: Would be cool if the Plausible footer could be minimalized too when using self-hosted setup. All sections below could be removed and simplified to basically "About Plausible" with just the left logo and Twitter/GitHub/GitHub Sponsors links. (Personally I think the footer is too large anyway)

ukutaht:
> If I'm allowed to add my 2 cents here: Would be cool if the Plausible footer could be minimalized too when using self-hosted setup. All sections below could be removed and simplified to basically "About Plausible" with just the left logo and Twitter/GitHub/GitHub Sponsors links. (Personally I think the footer is too large anyway)

Makes sense @patschi. I also think the footer could be more minimal for logged-in users on our hosted service. Just like the difference between [github.com](https://github.com) footer when you're logged in vs when you're not. Marketing stuff is less relevant for someone who already decided to sign up.

I'll leave the footer for now because I want to get the release candidate out ASAP but thanks for your suggestion. I'll bring it up with Marko.

marcbria:
Just installed plausible and I like the zen sensation you give compared to other solutions... but this peace of mind is kind of broken with a footer full of links that are useless in a self-hosted installation.

As Patschi suggested, it also encourage replacing this footer by a "About Plausible", "Developed by Plausible", etc.

Anyway, it's a great tool. You help to make internet more diverse and respectful with user's privacy. Thanks!

PR #618, opened 2021-01-19
Change datepicker caret from pink to indigo

shornuk:
### Changes

I couldn't understand why the dropdown caret/chevron was pink. It actually looks more red on my screen and I thought I had an error on the datepicker. Changed to indigo to match the rest of the theme. Appreciate this might be a preference thing, so feel free to ignore.

### Tests

### Changelog

### Documentation


Vigasaurus:
If this is getting changed (and I do agree that it should) - it'll also need to be updated for the dropdown in `assets/js/dashboard/filters.js`

shornuk:
@Vigasaurus - good catch. Took me a while to find where that's shown. Have updated that too.

ukutaht:
Thanks @shornuk!

> I couldn't understand why the dropdown caret/chevron was pink

I think it's leftover from when the UI looked quite different. Indigo is more on brand for sure!

PR #624, opened 2021-01-20
Add Comparison Line and Improved Tooltip to Visitors Graph

Vigasaurus:
### Changes

- Adds Comparison line with previous period
- Adds new and improved tooltip with full freedom of format
- Cleans up various VisitorGraph visual and functional issues
- Fixes a bug with the "month" based plot generation

Resolves #532

### Tests

### Changelog
Still unsure of what you want done for current changelog updates given the intermediate time in updates - lmk if you just want it underr 1.1.2 still. 

### Documentation
I don't *think* this needs a docs update/is mentioned anywhere, but am unsure. 

### Screenshots
image
image
image
image
image
image


ukutaht:
Looks good!

I'll merge and put it on staging for further testing. I'll also fix the changelog.

I really like the UI for when I'm interested in the comparison. However, I'd say most of the time I'm checking the stats, the comparison is not important (this could be very personal). This PR does add some clutter to the UI if the user is not interested in the comparison.

I think it would be interesting to test something like this: All comparisons are off by default to keep the UI clean and fast(er). From the datepicker you can choose to add a comparison with month-over-month/year-over-year/whatever logic. Then we show comparsions for everything: top stats, graph, graph tooltip, sources, pages, countries, goals, etc. What do you think @metmarkosaric @Vigasaurus ? Not for this PR but an idea for a future update.

metmarkosaric:
That could be a good alternative to try when we introduce more elaborate comparisons. 

I've been testing this on staging now and it still feels fast on our live demo and works very well.

Only comment I have is to remove the decimal points from the percentage number to keep it cleaner. i think showing 109.4% is not any more useful than just showing 109%

ukutaht:
I've done some more testing with some specific comments on this:

1. In MTD view, the comparison line and current line are cut off at the same point

Screenshot from 2021-01-20 16-23-33@2x


Now, as a user, what would you expect the percentages at the top to show? I think naturally most people would say that the totals for both lines are compared but that is not the case. Actually at the top we are comparing `2021-01-01,2021-01-20` with `2020-12-11,2020-12-31`. So I'm a bit afraid that this UI could mislead people to think that the comparison line and top percentages are related when they are showing completely different things.

2. The 'Today' view behaves differently from 'Month to date'. Instead of the comparison line cutting off at the same place as the current period, it spans the whole day:

Screenshot from 2021-01-20 16-31-34@2x

 
3. How to handle months that are not the same length? If you view the live demo for March 2020, you'll see that the dates in the comparison do not line up. The comparison line goes from `30 Jan - 29 Feb`

ukutaht:
@Vigasaurus Ignore most of my comments, I thought the graph was mixing previous period with month over month but it is not. All correct.

My 2nd comment on the difference between month to date and today view still stands.

metmarkosaric:
Yeah we should change the "Today" view to fit with the "MTD" view in order to keep our comparisons consistent (always comparison to the previous period). 

Then at a later stage we can introduce a more elaborate comparison section where you can compare year over year, month over month or day over day and for those we can showcase the full historical trend.

metmarkosaric:
Another thing I've noticed is that the tooltip cuts off like this when on the edge of the screen:

Screenshot from 2021-01-20 16-24-42

Vigasaurus:
> difference between month to date

Yeah, this is definitely something that I think we *should* do, it's just a bit annoying given we'd need to convert query and prev_query to use DateTime instead of Date. I can definitely take a look at it though

> cutting off

Let me see what I can do. I think I should be able to have it show the tooltip to the left pretty easily if we're in the right 1/4 of the window. 

> Decimal

Ah yeah, I forgot to remove that from your earlier comment, will do.

Vigasaurus:
> From the datepicker you can choose to add a comparison with month-over-month/year-over-year/whatever logic

100%. My concern with this has always been the UI of the date picker though, as on mobile especially, there's not much space on most viewports to make the drop-down much wider, or more than about 20% taller. I personally think maybe something like a checkbox and a simple drop-down as the last two items could be best - with the drop-down having options of:
- Previous Adjacent Period
- Previous Month
- Previous Year
- Previous Custom (this would open up another Flatpickr)

I definitely think GA's UI is a bit too tall and not compact enough given Plausible's location being 1. In a drop-down, 2. Not spanning the whole screen. 


Thoughts? @metmarkosaric @ukutaht

metmarkosaric:
thanks @Vigasaurus! was looking for some ideas. here's how GA v4 does it:

Screenshot from 2021-01-20 17-58-08

And when you switch on the "compare":

Screenshot from 2021-01-20 17-57-59

Vigasaurus:
Yeah I think that largely should work for us. The biggest thing imo is just figuring out the UI - as GA's version is definitely too tall for most screens in our dropdown style of datepicker. 

Sidenote: What do those arrows for `This Week` and `Last Week` do?

metmarkosaric:
GA definitely has way too many options for everything... we're in a good position that we can decide to curate our options and not necessarily include everything imaginable 

Screenshot from 2021-01-20 18-03-16

Screenshot from 2021-01-20 18-03-11

Vigasaurus:
> The 'Today' view behaves differently from 'Month to date'. Instead of the comparison line cutting off at the same place as the current period, it spans the whole day

I actually have a question/thoughts about how you'd want to do this - I've seen conflicting wants in the discussion, and I wanted to make sure. So there's really 2 big options for the partial-day view/comparison; either we do `previous set of hours` or `this time yesterday` where in the first it'd be very similar to the current MTD implementation (which I do think is flawed, more on that later), in that if we're looking at 00:00-13:00, the comparison would be for 11:00-00:00 (the previous 13 hours) causing some mishap in the graph line up and the hour span of the day (i.e. if you look at 08:00 today, that'd get compared to 19:00 yesterday), whereas in the second it'd be something which in my opinion makes much more sense for days, that we'd compare to the exact same times yesterday (this way the graph would also make sense), such that 00:00-13:00 today, would get compared to 00:00-13:00 yesterday. 

I think this same method actually makes more sense for MTD too, even though Marko did bring up the point that its not necessarily the standard for MTD. I just think that in the graph especially, having them not be largely the same period but in separate chunks would cause issues - as it doesn't really compare well, in the tooltip especially to compare `Jan 21` to `Dec 31` just because they're both the ends of their respective periods (Jan 1-Jan 21) and compared to (Dec 10-Dec 31), instead of intuitively comparing the MTD view of right now to the MTD for the previous month (Jan 1-Jan 21) and (Dec 1-Dec 21). This does bring up the issues of longer months, but I personally think that having the data line up across periods is far more important than potentially missing 1 day/mo in one view. 

@metmarkosaric @ukutaht

metmarkosaric:
Thanks @Vigasaurus! 

There are two main ways to do comparison in analytics and we should try not to mix them in the UI but we should support them both eventually. If you look at GA, you can have "previous period" (which is the default) or "over time" comparisons as the main two. The one we do now is the previous period comparison and we do it throughout the dashboard except in "Today". 

The current live version of Today is broken in terms of comparison. If I select Today right now it compares not even half of the day today with the full day yesterday which is not accurate comparison. We fixed this issue with Month To Date for the month view but we need to fix it for Today too to basically have the same previous period comparison like we do on Month To Date.

So basically the "previous period" comparison that we support currently, we basically take the current period of time on the report you're viewing (say January 1-21) and compare it to the same period of time right before the start of the current period (so December 10-31).

The "over time" comparison for the same report would be January 1-21 compared to (if the user selects month over month) December 1-21 or (if the user selects year over year) January 1-21 previous year. We don't support this right now but it is something that people have asked for and we should support using a new item in the menu such as the "Comparison" menu we discussed.

So perhaps the previous period comparison may not be the best one for the new comparison line (even though other platforms including GA support it) and maybe we should not have it on the main dashboard but only do it if people choose the manual "over time" comparison in the menu?

Vigasaurus:
Okay cool, that all makes sense, I think. 
So let me just make sure I understood you right, because I am working on this part of this soon-ish; for right now you think we should treat the `day-to-hour`  or current-day comparison as previous period, to match everything else as a temporary consistency solution (i.e. 00:00-13:00 is compared to yesterday's 11:00-24:00)?

But should plan for in the better compare menu to not have this behavior be the default, but instead have the default be "over time" as GA calls it (or "over same parent-period" as I understand it) where 1-21 would be compared to 1-21?

metmarkosaric:
What is clear is that our "Today" view comparison is broken and shows inaccurate percentage (comparing part of a day to full day).

For everything else, I don't think there's a clear answer. I just checked few tools including Search Console, GA, GAv4 and Paddle and they all have different ways of doing it. 

So perhaps we need to take a bit longer time to decide what we think is the best way to do it or perhaps present few options and ask the people in social media to say what they prefer.

ukutaht:
> So let me just make sure I understood you right, because I am working on this part of this soon-ish; for right now you think we should treat the `day-to-hour` or current-day comparison as previous period, to match everything else as a temporary consistency solution (i.e. 00:00-13:00 is compared to yesterday's 11:00-24:00)?

I think this is the only realistic option to make everything consistent for now. Then we can think about more clever comparisons on top of this.

Vigasaurus:
Sounds good. I'll poke around at it 👍

PR #625, opened 2021-01-20
Allow for `Esc` to clear all filters

Vigasaurus:
### Changes

Resolves #575

### Tests

### Changelog
Same as my other PRs heh - lmk what you want for changelog format and I'll get it sorted

### Documentation
🤷‍♂️


metmarkosaric:
Nice! Our keyboard shortcuts game is getting stronger! I like it!

Vigasaurus:
I'm actually considering something else too, but I did actually want y'all's feedback/opinions. 
firefox_2021-01-16_01-34-31

I don't know that actually showing them here is necessary, but I thought something like this very similar to how calendar apps work could be neat. (This is just a mock up made in paint)
For the Custom I thought it would be cool for it to just open up the Calendar input automatically, and for the others to do what you'd expect (i.e. exactly what clicking the button would do)

metmarkosaric:
that's very cool and i was thinking to share that idea. i didn't know it's a google calendar thing but last week we started using Hey for email and they do that everywhere and I really like it (just need to get used to the different keybindings)

Screenshot from 2021-01-20 09-47-25

Vigasaurus:
Sounds good (I've always been meaning to try hey but ... lazy)! How do you feel about the following, then? I think some of the more obscure ones don't need a binding, but lmk if you think there should be one for any. 

Dropdown Action | Keybind
--|--
Today | D
Realtime | R
Last 7 days | W
Last 30 days |
Month to Date | M
Last month |
Last 6 months | 
Last 12 months | Y
Custom range | C

metmarkosaric:
looks good to me @Vigasaurus! good that it follows google calendar as i assume that may be the standard that's used in more apps too.

assets/js/dashboard/filters.js
```diff
@@ -52,10 +54,21 @@ class Filters extends React.Component {
...
   componentWillUnmount() {
+    document.removeEventListener("keyup", this.handleKeyup);
     document.removeEventListener('mousedown', this.handleClick, false);
     window.removeEventListener('resize', this.handleResize, false);
   }
 
+  handleKeyup(e) {
+    const {query, history} = this.props
+
+    if (e.ctrlKey || e.ctrlKey || e.altKey) return
```
ukutaht, line 65: Seems like `ctrlKey` is checked for twice

ukutaht (changes requested):
Feel free to add to the `1.2 - unreleased` section on the changelog.

assets/js/dashboard/filters.js
Vigasaurus: Ah haha - I just copied this from the DatePicker arrows, didn't even notice 🙈

ukutaht (approved):

PR #629, opened 2021-01-20
Comparison Tooltip Part 2

Vigasaurus:
### Changes
#624 Part 2

- Rounding to 0 Decimals
- Tooltip Position when close to right side of page

### Tests

### Changelog

### Documentation


Vigasaurus:
Going to also add the fix for this: https://twitter.com/lawik/status/1352194943968825344 here, soon. (re: 7d actually showing 8 days of data and skewing day of the week comparisons)

Vigasaurus:
Also @ukutaht I'm thinking about adding a basic toggle switch in the DatePicker dropdown for now (synced to localStorage) so we can have it toggleable in the UI, and then that can get reused later when the comparison options are added. Thoughts?

It'd just be React-only (hiding the second line, and making the tooltip just the one line) - as I figure having the server not send the data is out of scope for now (as we'd need to add the `comparison` query param later anyways)

metmarkosaric:
There's some more feedback on Mastodon: https://fosstodon.org/@booligoosh/105593366137184279

"I think the thing is that the "visitors" text doesn't change as you move the mouse across the different days.

Before, it was easy to hover on the exact date I wanted, because I could just move my mouse along (scanning the big date text) until I got to the right one. Now, I have to look a lot more closely to get the right date, since it's not the big bold text.

It feels like a waste of scannablity to use the big bold text for something that doesn't change, if that makes sense."

metmarkosaric:
In general, I'm starting to think this should not be on by default and not an option either but that we should only introduce it with the "Comparison" section. So when you're in the Comparison section and comparing time periods you get the comparison line in the graph too. Otherwise it is not there.

Vigasaurus:
Yeah I'm thinking kinda the same thing; I think without more granular selection (which I was planning on doing next anyways) it comes across sort of half-baked, as per a lot of the feedback. 

As for changing the big text, I really do think the current layout is really good, but I can agree with swapping the text sizes/weights - but I think any other re-arranging would likely cause either repetition or an awkward reading flow. 

If @ukutaht agrees, I'll close this PR and we can revert the other PR from master for now; and I'll continue building upon it with the feedback so far + granularity options.

metmarkosaric:
Sounds good to me @Vigasaurus, thank you! As comparison line will be part of the comparison section only, I don't think we need to change the big text. I also like it the way it is. The important part is that the current experience on the live version stays as it is so we don't change that text there while this comparison line is just a very cool extra you get when using comparisons.

ukutaht:
Yeah OK I will revert the previous one and stand by for a more polished PR.

About the tooltip I think making the dates bigger or the label smaller is a good thing to try. Pretty happy with how it looks already though.

I do think it would be good to prevent the backend from querying the comparison if it's not enabled. I'm currently working on the API and I'm designing it to allow comparison modes. Probably not in the first release but if the API is designed right we can release comparisons later without making any breaking changes.

Same goes for intervals. I'm designing the API to allow things like `Last 6 months, reported daily` vs `Last 6 months, reported monthly`. Not to complicate things too much but this switch should also fit somewhere in the UI. (Maybe in the graph above the download button?)

Neither of these features will probably be in the v1 launch of it. So @Vigasaurus if you want to implement comparison modes and different intervals in the backend you can go ahead. Obviously I can help or work on the backend implementation myself.

I really think these two features will take the graph to a whole new level, excited!

Vigasaurus:
Neat, sounds cool! I'll continue on my `532-comparison-line` branch for this (and merge this into that). What I'm mostly planning is that every period will have a set of allowed comparisons, the entire set of them being: `Preceding Period (adjacent)`, `Last Week`, `Last Month`, and `Last Year` - each of which will only show up if the period is shorter than the comparison - i.e. `Day` will see all of these, `7d` will see all except for `Last Week`, `30d` and other months will see only `Preceding`, `Last Month` and `Last Year`, and Custom will see only `Preceding` and `Last Year`(if it's sub 364 days), otherwise just `Preceding`. And of course all of them will have `Custom` available, which will just take a DateRange

ukutaht:
@Vigasaurus sounds great!

PR #640, opened 2021-01-22
fix(register): disallow registration when registration is disabled

TheodoorP:
### Changes
Currently, only visiting the form is disabled when registration is disabled. If a user manually makes a post request to the register endpoint they can bypass the environment variable.

This PR fixes that issue by checking for the environment variable when registering.

### Tests

### Changelog

### Documentation

### DIFF
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -32,36 +32,41 @@ defmodule PlausibleWeb.AuthController do
   end
 
   def register(conn, params) do
-    user = Plausible.Auth.User.new(%Plausible.Auth.User{}, params["user"])
-
-    if PlausibleWeb.Captcha.verify(params["h-captcha-response"]) do
-      case Repo.insert(user) do
-        {:ok, user} ->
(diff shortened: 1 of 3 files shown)


ukutaht:
I think a cleaner option would be to only attach the routes for `GET /register` and `POST /register` when the app is configured to allow registration. This could be done in `lib/plausible_reb/router.ex` like so:

```ex
if !Keyword.fetch!(Application.get_env(:plausible, :selfhost), :disable_registration) do
    get "/register", AuthController, :register_form
    post "/register", AuthController, :register
end
```

I feel like it keeps the controller action nice and clean. It would behave a bit different and return 404 on these pages instead of redirecting to /login though.

Thoughts?

ukutaht:
Welp, that doesn't work as seen in the CI. Looks like routes are attached at compile-time which doesn't make sense in this case. We use the same compiled image for both self-hosted and Cloud so this config option must be applied at runtime.

Sorry, my bad for leading in the wrong direction. If you revert the latest commit I will hit merge.

ukutaht:
Thanks!

PR #673, opened 2021-02-02
Make /sites live update

KristerV:
### Tests

### Changelog

### Documentation

### Changes

- stats update every 30 sec
- user can change period shown

image

I had assumed the stats update all along :)

### Problems

I understand the *real* solution would be to Reactify the whole thing, but it was a little too many things to bite off at once (even though I got half way with React).

### Areas to check by reviewer

- Is using session for storing period okay?
- **It's hard to test with no data. Does the dropdown and periods actually work?**
- At the end of the template is a raw `<script>` - is that ok?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 ## [1.2] - Unreleased
 
 ### Added
+
 - Ability to add event metadata plausible/analytics#381
-- Add tracker module to automatically track outbound links  plausible/analytics#389
+- Add tracker module to automatically track outbound links plausible/analytics#389
 - Display weekday on the visitor graph plausible/analytics#175
 - Collect and display browser & OS versions plausible/analytics#397
 - Simple notifications around traffic spikes plausible/analytics#453
@@ -16,8 +18,10 @@ All notable changes to this project will be documented in this file.
 - Escape keyboard shortcut to clear all filters plausible/analytics#625
 - Tracking exclusions, see our documentation [here](https://docs.plausible.io/excluding) and [here](https://docs.plausible.io/excluding-pages) for details plausible/analytics#489
 - Keybindings for selecting dates/ranges plausible/analytics#630
+- Live update stats in /sites and allow changing period
 
 ### Changed
+
 - Use alpine as base image to decrease Docker image size plausible/analytics#353
 - Ignore automated browsers (Phantom, Selenium, Headless Chrome, etc)
 - Display domain's favicon on the home page
@@ -34,6 +38,7 @@ All notable changes to this project will be documented in this file.
 - Changed caret/chevron color in datepicker and filters dropdown
 
 ### Fixed
+
 - Do not error when activating an already activated account plausible/analytics#370
 - Ignore arrow keys when modifier keys are pressed plausible/analytics#363
 - Show correct stats when goal filter is combined with source plausible/analytics#374
@@ -47,20 +52,24 @@ All notable changes to this project will be documented in this file.
 - Various UI/UX issues plausible/analytics#503
 
 ### Security
+
 - Do not run the plausible Docker container as root plausible/analytics#362
 
 ## [1.1.1] - 2020-10-14
 
 ### Fixed
+
 - Revert Dockerfile change that introduced a regression
 
 ## [1.1.0] - 2020-10-14
 
 ### Added
+
 - Linkify top pages [plausible/analytics#91](https://github.com/plausible/analytics/issues/91)
-- Filter by country, screen size, browser and operating system  [plausible/analytics#303](https://github.com/plausible/analytics/issues/303)
+- Filter by country, screen size, browser and operating system [plausible/analytics#303](https://github.com/plausible/analytics/issues/303)
 
 ### Fixed
+
 - Fix issue with creating a PostgreSQL database when `?ssl=true` [plausible/analytics#347](https://github.com/plausible/analytics/issues/347)
 - Do no disclose current URL to DuckDuckGo's favicon service [plausible/analytics#343](https://github.com/plausible/analytics/issues/343)
 - Updated UAInspector database to detect newer devices [plausible/analytics#309](https://github.com/plausible/analytics/issues/309)
@@ -68,9 +77,11 @@ All notable changes to this project will be documented in this file.
 ## [1.0.0] - 2020-10-06
 
 ### Added
+
 - Collect and present link tags (`utm_medium`, `utm_source`, `utm_campaign`) in the dashboard
 
 ### Changed
+
 - Replace configuration parameters `CLICKHOUSE_DATABASE_{HOST,NAME,USER,PASSWORD}` with a single `CLICKHOUSE_DATABASE_URL` [plausible/analytics#317](https://github.com/plausible/analytics/pull/317)
 - Disable subscriptions by default
 - Remove `CLICKHOUSE_DATABASE_POOLSIZE`, `DATABASE_POOLSIZE` and `DATABASE_TLS_ENABLED` parameters. Use query parameters in `CLICKHOUSE_DATABASE_URL` and `DATABASE_URL` instead.
--- a/lib/plausible_web/controllers/site_controller.ex
+++ b/lib/plausible_web/controllers/site_controller.ex
@@ -5,7 +5,7 @@ defmodule PlausibleWeb.SiteController do
 
   plug PlausibleWeb.RequireAccountPlug
 
-  def index(conn, _params) do
+  def index(conn, params) do
     user = conn.assigns[:current_user]
 
     sites =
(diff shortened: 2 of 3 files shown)


oliver-kriska:
I think, it should be better to use PhoenixLiveView. It's easy to do it for this use case.

KristerV:
LiveView was actually my first thought. Then it turned out that 1) :phoenix_live_view is not installed and 2) the current frontend is a mixture of templates and React. So mixing in a third frontend framework seems excessive. Or what have you guys discussed on this topic before?

oliver-kriska:
It's really depends what plausible team think about it. I think there is plenty of work for Phoenix and Phoenix LV. I'm not sure how important is react in this project. Maybe they had reason to not use LV. We will see when they response. 

Your solution is good, it works. But it's another solution and from my point of view it's same as another framework. So maybe it's enough and maybe not. Reason why I like LV even for this kind of small features/reasons because it open your possibilities. So just imagine you made it in LV and another dev upgrade it with a few lines of code with small stats graph or something else. Also it can open possibilities for another features/issues when people will be inspired by your solution to make another LV.

Maybe I ate too much sugar for lunch....

KristerV:
Yeah I'd love to know why the current solution is templates + React. Seems like a very difficult way to approach frontend. I think Uku started long enough ago that LV just wasn't mature enough, but now perhaps is. And your argument about my solution being also another layer definitely holds water.

How hard is it integrating LV into the already existing codebase? Does a lot need to change or can it just be used wherever? Last I remember checking the configuration seemed a huge headache - that's why I haven't tried.

oliver-kriska:
It's easy. The hardest part or the most problematic part is endpoint configuration. Because doc says you should use @session_config constant and use it for LV config and session config. It has to have same config. It's not good because they are compiled at compile time so when you change session secret on production. This change will be not applied. It's ok for running production in mix but not for release or docker of course. Yes, there are some new elements and maybe new approach. That can be hard but this is easy example. Which can be done for couple of minutes, 1 hour at max. I can show you some example tomorrow. Today, I have to finish refactoring of one old LV with 600 LoC :/

KristerV:
@ukutaht waiting for your comment.

ukutaht:
A few thoughts on this.

As a solo maintainer of this project I view all code as a liability. This is why my approach has been to build/accept contributions only for things that a large number of our users are screaming for.

It has never crossed my mind that the /sites page should update live. I've also never heard a user or customer ask for it. I truly appreciate your contributions but they have to make sense in the larger context and product roadmap. At the moment I feel like merging will not improve the product but it will add more code and endpoints to worry about. Happy to be convinced otherwise :)

For example, I was happy to accept contributions on dark mode and embedded dashboards are being worked on. These are super high-value features that many people want.

The code looks good except for one thing. It introduces an N+1 query which is avoided in the current implementation on purpose. Some customers have a lot of sites, so I'm afraid the N+1 would make their /sites page really slow.

I started this project in Dec 2018 which I can see is when the first blog post introducing LiveView was published. It was in its unstable infancy and so I chose tech I was more comfortable with at the time, namely React.

I've been spending some Fridays hacking on the dashboard with LiveView. I think it's a great dev experience and has a lot of promise for reducing both latency and bandwidth. Changing the whole dashboard is a much bigger job than this /sites page. I'd like to build a prototype of the dashboard with LiveView and compare performance and code structure / maintainability.

KristerV:
Not gonna lie, this is sad. I was looking forward to having an actual overview of my sites. But your position does make sense.

It does confuse me a little how people (by the metric of what's being asked) value dark mode more than having an overview of their websites, I mean that's what the product is for. But I guess they may ask for it later or whatever.

**Proposition:**

I was actually mostly waiting to change the time window from 24h to 30 days since 24h just doesn't tell me much of anything. If I remove the live update stuff and just keep the dropdown - would that change be acceptable?

metmarkosaric:
thanks @KristerV for the contribution! 

just a bit more info on the feature requests:

we keep track of all the requests as we speak to many people every day (subscribers, trial users, people using GA and other tools that want to switch to Plausible...). our development is prioritized by the requests we get combined with our vision for what Plausible is (simple, lightweight, fast, privacy-first). dark theme and API were on the top of the most requested list over the last few months.

most demanded features at this stage (i would say that these 7 get asked for more often than all the other features combined):

city/state
user accounts
import from GA
embedded mode
time on page
entry/exit pages
funnels/userflows

ukutaht:
Well yeah I also didn't understand why so many people were demanding a dark mode to be honest :) Now that it's here I actually really like it.

I think there have been quite a few requests for making the /sites page more useful:
https://github.com/plausible/analytics/discussions/286
https://github.com/plausible/analytics/discussions/674
https://github.com/plausible/analytics/discussions/617

I do like your proposition to remove the live updating and just show more useful information on the /sites page. I would love to have a date picker like you added and small sparkline graphs on that page. I thought about adding them last time I updated the design for that page but I didn't because there's a massive [yak to be shaved](https://americanexpress.io/yak-shaving/) if we want to add this information. Here goes.

Some people have a lot of sites. Even if we add pagination, this page would have to fetch stats for something like 20 sites. Issuing 20 separate DB queries and waiting for all to return is really slow, especially when querying sites with a lot of traffic over a long timeframe. The processing time grows linearly with the amount of events being aggregated on the fly. We could maybe parallelize this (?) but it's still a bad solution IMO. Instead I wanted to fetch the stats in one query.

Now, why can't we get stats for 20 sites in a single query with a `WHERE domain IN ('site1.com', 'site2.com')` etc? Because each site could be configured in a different time zone. This means the time range being queried in the database are different between different sites. It's a side effect of storing timestamps in UTC and it means we cannot query multiple sites in one query.

This is why the /sites page shows last 24h at the moment. Last 24h is the only time period in Plausible that isn't relative to a time zone. Every other time period is relative to the site's timezone.

My long-term plan with this page is to start generating daily rollups for pageviews and unique visitors in the site's configured timezone. This is made easy with materialized views in Clickhouse. This gives us a fast, constant-time way to fetch total pageview and unique visitor numbers. Since the timestamps in this table would be in the site's timezone, multiple sites could be accessed at once. Daily aggregates on the DB level would make it feasible to also show 20 graphs on the /sites page. It would also massively speed up the visitor graph in the main dashboard.

Circling back to your proposition

> I was actually mostly waiting to change the time window from 24h to 30 days since 24h just doesn't tell me much of anything. If I remove the live update stuff and just keep the dropdown - would that change be acceptable?

I would have to test it with an account that has, let's say, 20 sites with 10m monthly pageviews. So we'd be aggregating 200m pageviews on the fly each time the user checks their /sites page. My gut feel says that it will be unacceptably slow, especially considering the non-parallelized N+1 query [here](https://github.com/plausible/analytics/pull/673/files#diff-090303076e22ad4dfdef24a4e7a0df766fc58120336418eae89fd072039c28dfR25).

If those queries are run in parallel, it has a chance of working but still no promises :)

KristerV:
> Last 24h is the only time period in Plausible that isn't relative to a time zone. Every other time period is relative to the site's timezone.

Last 30 days is 24h * 30 so timezone isn't an issue there either.

> I would have to test it with an account that has, let's say, 20 sites with 10m monthly pageviews. 

10m monthly with 20 sites (and 1000 users) will be extremely slow unless pre-aggregated. I know, [I've tried](https://code.krister.ee/postgresql-slow-with-million-row-aggregation-how-to-debug/) with Bashboard. In my case I needed more complex aggregations, but the performance ceiling for above million row queries still comes fast enough that you don't want to risk it, even if it doesn't show it's ugly head immediately. In this case daily materialized views would work wonders (I needed live stats for Bashboard so eventually made cache tables with pre-aggregated data).

Okay I understand the problem better now. Uhmm.. it seems like I'm in a great position to dive in and make the proper change along with those other features, considering my experience. But from the same experience I know it's a rabbit hole and needs thorough tests, performance testing (I made an automatic performace test for Bashboard) and probably multiple rewrites to get it right. DB performance was basically the reason why Bashboard took 3 months and not 3 weeks. I hate to say it but this seems out of my reach time wise even if it does seem exiting. I could be hired for this, but I'm not sure if it makes a lot of sense from your end, because you'll still need to input time also.

I suppose this PR can just be closed. I'll wait a bit for comments and close it later.

ukutaht:
Cheers @KristerV. Didn't realize how much experience you have in analytics and DB stuff. But yeah working on the internal guts of Plausible requires close collaboration. Something like dark mode can easily be built independently but the DB performance stuff is too deep of a rabbit hole for that.

> Last 30 days is 24h * 30 so timezone isn't an issue there either.

Hmm, I think it is. When we talk about a 'day' we mean a calendar day in the timezone the site is configured. Let's say I'm doing a last 30 day query right now for a site configured in `UTC-12` vs a site configured in `UTC+12`. To simplify, let's think about 2 days instead of 30 but the same problem applies.

| Timezone | Range in local time                                          | Range in UTC                                     |
|----------|--------------------------------------------------------------|--------------------------------------------------|
| UTC-12   | `2021-02-11T00:00:00Z-12:00` to `2021-02-12T23:59:59Z-12:00` | `2021-02-10T12:00:00Z` to `2021-02-12T11:59:59Z` |
| UTC+12   | `2021-02-11T00:00:00Z+12:00` to `2021-02-12T23:59:59Z+12:00` | `2021-02-12T12:00:00Z` to `2021-02-13T11:59:59Z` |

What you see here is that when querying for 2 consecutive days and sites are on opposite ends of the world, there is literally zero overlap in the UTC timestamp range. The problem of course is less pronounced when querying over longer timeframes but it's still incorrect to use the same start and end UTC timestamps when sites are configured in different timezones and reporting is relative to calendard days/months etc.

> In this case daily materialized views would work wonders (I needed live stats for Bashboard so eventually made cache tables with pre-aggregated data).

Yeah I think it's a matter of time until we start doing pre-aggregrations as well

KristerV:
> Didn't realize how much experience you have in analytics and DB stuff. 

It was actually quite a fun rabbit hole. I'm glad my company allowed me to pursue this. Went deep on Postgres indexes and how it scans the tables. It is very intelligent, but still bound by the physical world :)

> Hmm, I think it is.

Lets say I want to know the stats for the last minute. No timezone applies. How about last hour? Still no timezone. Last 24h? Last 48h? Last 720h? Last one is effectively a month. I know you differentiate between "last 30 days" and "a month". I was talking about "last 30 days" in this case. Perhaps me explaining will bring this feature a little bit closer, but probably not :)

Ok, was nice talking and looking into the codebase, but I guess that's all for now :)

PR #675, opened 2021-02-03
Expose port 8000 by default

### LINKED ISSUE #666: EXPOSE 8000 missing in Dockerfile
One should not rely on the user to do a port mapping in order to be able to access the service, dockerized applications should use EXPOSE to list the ports they listen for connections at.

This causes issues when running under a reverse proxy such as Traefik. Without the "useless" port mapping to the host it is impossible for it to detect the service, as it has no exposed ports.

Sorry for no PR but I'm not an expert in the container build process, and in order to leverage caching etc I don't know the best place for the EXPOSE would be.

RealOrangeOne:
### Changes

Fixes #666
Re-implements #237 

This doesn't bind it to the host port by default, it's more a hint to tools which integrate with Docker directly.

### Tests

### Changelog

### Documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,4 +65,5 @@ RUN chown -R plausibleuser:plausibleuser /app
 USER plausibleuser
 WORKDIR /app
 ENTRYPOINT ["/entrypoint.sh"]
+EXPOSE 8000
 CMD ["run"]


ukutaht:
Thanks!

Issue #95, opened 2020-06-19
API to show data on dashboard


metmarkosaric:
"As we currently use a dashboard that pulls in data via an API (or other connection methods) it would nice to be able to gave an API that can pull in details. E.g. /api/pageviews/last7days to get the last 7 days page views.

This would also be useful to make rankings of most visited posts on a blog.

Which pages get the most traffic. Ideally the API would be able to return specific page types, ie tell us the most popular product pages, ignoring home page, informational and support pages, for a specified time period."

([imported from the old roadmap](https://plausible.nolt.io/22))

binaryfire:
Awesome to see this is on the roadmap!

fabio-smu:
@metmarkosaric what's the reason to this issue have been removed from roadmap?

fabio-smu:
@metmarkosaric this issue is not on roadmap anymore?

metmarkosaric:
thanks @fabio-smu! we've basically reduced our roadmap to only show the most immediate things we're working on. there's only a couple of items outstanding on the short-term roadmap. you can see it at https://github.com/plausible/analytics/projects/1. 

after that is completed, the API and the export/import of data will be the next two big projects we will focus on. both of these we want to do and they are some of the most upvoted features. we'll update the roadmap again as we complete the current tasks and as we priortize the upcoming work. it is difficult to promise any timelines at this stage on when the API will be completed but we will keep you posted!

alexphelps:
I'm also very interested in this, with an API to pull data for a dashboard, Plausible would be a great fit for us.

lmiller1990:
Going to keep an eye on this issue - pretty keen. The real product, to me, is not plausible the UI, but my analytics tracked in a non-invasive way. Access to that data is the next most important thing, this would let me perform my own analysis more easily.

Loving Plausible so far. 👍

mishushakov:
Agreeing with all the things said above
our scenario: provide analytics for websites we host for our customers

having API would allow us to add the said websites to Plausible and make it possible to fetch and display the data in our customer-facing console

metmarkosaric:
We plan to work on the API as the next big project at the start of next year. In order to deliver the product that satisfies your needs, we would love to hear from people who want to offer Plausible Analytics to their customers, clients etc using this API.

What do you plan do to? What do you need and expect from the API? Do you have any specific requirements that we should consider when building this? 

You can share your feedback here or reach out to us on hello@plausible.io. Thanks!

[...]

PR #679, opened 2021-02-04
Stats API

ukutaht:
### Changes

Relevant discussion: #95

This adds a read-only API to retrieve stats from your Plausible dashboard. I will create a PR with full documentation on our docs repository soon. Will also publish a Postman collection for testing. Some relevant points to discuss:

### Authentication
I decided to go with a simple Bearer token for authentication. The API key is generated with Erlang's `crypto:strong_rand_bytes(64) |> Base.url_encode64()`. This creates 64 strong random bytes. Before storing in the database, this value along with the server secret is hashed using sha256. The first 6 letters are kept as plaintext to make it easier to recognise the API key from the UI.

I didn't want to use Bcrypt here because it adds an artificial delay of about 250ms to each request. OK when you're loggin in, but not OK for API requests. Since the API 'password' is 64 random bytes, we have much more entropy there and a brute force attack is less likely than user-generated passwords. Would be good for someone to confirm whether this is a reasonably secure approach.

### Endpoints

Endpoints are namespaced with `/api/v1/stats`. The API is quite low level and flexible, almost like a restricted database interface. I worry about getting the abstractions right at this stage. I am building it with the aim of being able to rebuild the main dashboard by just using the public API endpoints. The endpoint are:

* `/api/v1/stats/realtime/visitors` for current visitors
* `/api/v1/stats/aggregate` for aggregated stats like the top row of the dashboard (visitors, bounce rate, etc)
* `/api/v1/stats/timeseries` for graph data like the main graph on the dashboard
* `/api/v1/stats/breakdown` for breaking down properties like 'Top sources', 'Top pages' etc. Could also be called `group-by`. Not part of this PR, under construction still 🚧

Again, full documentation will be available soon on the docs repo.

The endpoints are built to accommodate features that we have hard plans to build in the near future:
* Different filter operators (currently only `==` supported)
* Combining filters with `AND` and `OR` (currently only `AND` supported)
* Different comparison modes (previous period, month-over-month, year-over-year, etc)
* Configurable interval for timeseries data (e.g. view last year of data but group daily instead of monthly)

### Tests

### Changelog

### Documentation

### DIFF
--- a/lib/plausible/auth/api_key.ex
+++ b/lib/plausible/auth/api_key.ex
@@ -2,8 +2,9 @@ defmodule Plausible.Auth.ApiKey do
   use Ecto.Schema
   import Ecto.Changeset
 
-  @required [:user_id, :key]
+  @required [:user_id, :key, :name]
   schema "api_keys" do
+    field :name, :string
     field :key, :string, virtual: true
     field :key_hash, :string
     field :key_prefix, :string
@@ -11,5 +12,29 @@ defmodule Plausible.Auth.ApiKey do
     belongs_to :user, Plausible.Auth.User
 
     timestamps()
+  end
+
+  def changeset(schema, attrs \\ %{}) do
+    schema
+    |> cast(attrs, @required)
+    |> validate_required(@required)
+    |> process_key
+  end
+
+  def process_key(%{errors: [], changes: changes} = changeset) do
+    hash =
+      :crypto.hash(:sha256, [secret_key_base(), changes[:key]])
+      |> Base.encode16()
+      |> String.downcase()
+
+    prefix = binary_part(changes[:key], 0, 6)
+    change(changeset, key_hash: hash, key_prefix: prefix)
+  end
+
+  def process_key(changeset), do: changeset
+
+  defp secret_key_base() do
+    Application.get_env(:plausible, PlausibleWeb.Endpoint)
+    |> Keyword.fetch!(:secret_key_base)
   end
 end
--- a/lib/plausible/auth/api_key.ex
+++ b/lib/plausible/auth/api_key.ex
@@ -1,0 +1,15 @@
+defmodule Plausible.Auth.ApiKey do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @required [:user_id, :key]
+  schema "api_keys" do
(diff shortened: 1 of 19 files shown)


oliver-kriska:
What about Phoenix Token for api authentication?  Performance should be fine, even you can use lower key iterations than default 1000. Security should be maybe better. You can store api_key id in this token and do some verification or even skip it, whatever you want. In case of huge need for performance, you can use cachex for cache of actually used keys and verification. 

https://hexdocs.pm/phoenix/Phoenix.Token.html#decrypt/4
There is code how it works: https://github.com/elixir-plug/plug_crypto/blob/master/lib/plug/crypto/message_encryptor.ex#L58-L67

oliver-kriska:
The best practice is just show once the api key. But you can store it with [Cloak](https://hex.pm/packages/cloak_ecto) and allow to show it. This storing type is not good for search but you don't have to search because token can contain regular api key id inside.

ukutaht:
> What about Phoenix Token for api authentication? Performance should be fine, even you can use lower key iterations than default 1000. Security should be maybe better. You can store api_key id in this token and do some verification or even skip it, whatever you want. In case of huge need for performance, you can use cachex for cache of actually used keys and verification.
> 
> https://hexdocs.pm/phoenix/Phoenix.Token.html#decrypt/4
> There is code how it works: https://github.com/elixir-plug/plug_crypto/blob/master/lib/plug/crypto/message_encryptor.ex#L58-L67

Like you mentioned the best practice is to just show the key once. After the key is generated, I would prefer to treat it like a password. This means that even in case of the database and secrets leaking, it should not be possible to retrieve the plaintext api keys.

To achieve this I wanted to use a one-way hash function instead of a two-way encryption function. This is why I didn't use Phoenix Token. Because in case a of the database and secrets leaking, the attacker can just `Phoenix.Token.decrypt/4` them all and get access to stats.

With the current `sha256` approach, the attacker would not have an easy way to retrieve the plaintext value of the api key. Even if they knew the server secret, they would have to brute force 10^77 possible combinations.

oliver-kriska:
@ukutaht I got it. I understand what you want to achieve. We will see how it will work in production. BTW: it will be bigger problem when DB and phx secret will be compromised ;)

PR #709, opened 2021-02-12
Fixes case-sensitivity and residual keys in date-keybinds

### LINKED ISSUE #707: Keyboard shortcuts: cmd does not seem to be ignored correctly
Since keyboard shortcuts for time period navigation were introduced, I've internalized them and navigating Plausible is now much faster. Great contribution by @Vigasaurus, thank you!!

The only issue both myself and Marko have ran into is the following: when you kill another tab with `cmd+w` and land on an open Plausible dashboard, it seems to pick up the `w` and navigate to the `Last 7 days` view. I never intend to go to the week view when closing another tab so this is a strange conflict that we should ignore.

Looking at the code, it should already ignore the keyboard shortcut when meta key or ctrl is pressed at the same time but for some reason it isn't working. I'm on Elementary OS running Firefox.

Any ideas @Vigasaurus? Can you reproduce?

Vigasaurus:
### Changes

- Fixes #707 
- Ensures that keys pressed with a capitalization modifier are properly handled

### Tests

### Changelog
With 1.2 now released, I'm unsure where you want changelog updates/the new version you're targeting

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Stats API [currently in beta] plausible/analytics#679
+- 30 day and 6 month keybindings (`3` and `6`, respectively) plausible/analytics#709
+
+### Fixed
+- Capitalized date/time selection keybinds not working plausible/analytics#709
 
 ## [1.2] - 2021-01-26
 
--- a/assets/js/dashboard/datepicker.js
+++ b/assets/js/dashboard/datepicker.js
@@ -31,19 +31,19 @@ class DatePicker extends React.Component {
   }
 
   componentDidMount() {
-    document.addEventListener("keyup", this.handleKeyup);
+    document.addEventListener("keydown", this.handleKeyup);
     document.addEventListener("mousedown", this.handleClick, false);
   }
 
   componentWillUnmount() {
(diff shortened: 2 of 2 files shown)


ukutaht (commented):
Awesome @Vigasaurus I can confirm my issue with cmd+w is fixed on your testing box.

I've updated the CHANGELOG.md on master. You can add this under 1.3 - unreleased.

I did notice some weirdness. On current master, shift+w doesn't do anything. With this PR hitting shift+w seems to clear the `period` query param. I would expect it to either do nothing or navigate to `?period=7d`. Is this intentional or just happening on my machine?

Vigasaurus:
Hmm that seems backwards in respect to the issue, to me. 

On the current master pressing `shift+<any accepted key>` would reset the period and date params, which was one of the bugs this was fixing. 

In this PR, pressing `shift+<any accepted>` should do the same as the non-shift version of the key (and it behaves correctly like this for me). And in this PR, if the period is already set to the one corresponding to the key pressed, the period is left unchanged, but the date, to and from params are reset to being unset.

If you're having issues with the shift+key resetting the period, I'd have to assume that something's up with your webpack or caching; I just tried on a fresh pull and its seemingly working as described above.

ukutaht:
You're correct. I must have mixed up my testing environments or something. Thanks for the contribution!

Issue #108, opened 2020-06-19
Entry pages for goal conversions


metmarkosaric:
Entry pages for goal conversions

([imported from the old roadmap](https://plausible.nolt.io/49))

Issue #101, opened 2020-06-19
Exit Pages


metmarkosaric:
"It would be nice if you could track exit pages / exit rate."

([imported from the old roadmap](https://plausible.nolt.io/32))

seblz432:
Yes! This is the one major thing missing for me and what's stopping me from just switching to Plausible tbh

Issue #458, opened 2020-12-10
Allow sorting Top Sources by each of the three categories


preslavrachev:
## Prerequisites
- [x] I have searched open and closed issues to make sure that the feature has not yet been requested.

## Feature request
At the moment, the full list of top sources is ordered by the number of visitors. It would be very helpful to make it possible to sort them based on any of the remaining categories (e.g. Bounce Rate and Visit Duration)

![image](https://user-images.githubusercontent.com/86060/101527221-01024f80-398e-11eb-925d-cb7655467d96.png)

**How would you like it to work**
By default, Top Sources will remain sorted by the number of visitors, but the column will have a sort icon, suggesting it is the one that results have been sorted upon. All column headers will be clickable. By clicking on any of the remaining columns, it will become the actively sorted upon and the list of results will be re-rendered to reflect that 

---
**NOTE:** As usually, If there is enough interest in getting this feature implemented, I can implement it myself and send a PR.

ukutaht:
Agreed, I think this would be a great addition

metmarkosaric:
Would be great for Top Pages too!

PR #712, opened 2021-02-13
Adds entry and exit pages to Top Pages module

### LINKED ISSUE #226: Entry pages
"The only thing what I miss that we can't check which page was the first what the user have opened in a session."

request from [Twitter](https://twitter.com/f3r4lh34rt/status/1283400173335584768)

And we already have request #101 for Exit pages.

Vigasaurus:
### Changes

- Adds Entry and Exit pages tabs for all views/filter settings
- Adds entry and exit page viewing for goal filters
- Adds normal page views for referral sources (pulls session data etc to query accurately)
- Adds modal views with incremental loading for Entry and Exit pages
- Adds Exit Rate for exit pages
- Adds session duration based on entry pages
- Resolves #226, #108 and #101

### Tests

### Changelog

### Documentation
Unsure if this needs a docs update - I'd imagine not since it's straightforward but dunno. 

### Screenshots
image
image
image
image
image


metmarkosaric:
This is great @Vigasaurus, thank you! Entry/exit pages is for sure one of those frequently requested features!

Perhaps we should rename them in the dashboard to only say "Top", "Entry" and "Exit" as I assume the current presentation would be too long for many mobile screens?

Average session duration would be a useful metric for sure for the "Entry Pages" report. Perhaps we name it "Visit Duration" so it follows the same naming principles as we already use "Visit Duration" in the top graph? 

I'm happy to keep the "Bounce Rate" in "Entry Pages" too until we do the #591.

"Exit Rate" is definitely something we should try to add to the "Exit Pages" report. Makes it easy to identify pages that perform worse than others.

This all will be even more powerful when we do the #458

Vigasaurus:
100% agree on renaming the tab names.

I do worry that adding visit duration to entry pages will make it overflow a lot on mobile (I intend to tackle this next probably, and the modal UI in general). So maybe we can hold off until I do that to add more columns there. I do honestly kinda think there shouldn't be the duplication between pages and entry pages on this anyways (as long as they do show the same values, which they do as of now), so maybe we add session duration in replacement on entry?

Exit rate I agree for sure. Do you have an agreed upon formula in mind? I've seen a couple of varying ones in my research, would you rather just do the way GA does it, i.e. exits/total, or unique exits/unique visitors - or something else?


I do definitely think this all be awesome with #458, and maybe I can help out with that after I do some of the other stuff I'm planning; but I'll coordinate with the person who started on it earlier when that time comes.

metmarkosaric:
Nice!

Another note: I think it's best to retitle "Unique Entrants" to "Unique Entrances" to keep everything consistent.

I prefer to keep the GA formula in this case (and in general too unless when we can do something much more accurate/useful). New users frequently compare our numbers to what they see in GA so they get a smoother experience when our definitions/numbers match to what they expect to see.

Not sure what you mean by "maybe we add session duration in replacement on entry"? You mean to replace unique or total entrances column with Visit Duration or?

Mobile is a bit of an issue in general. On my own mobile, the last column is already cut (I cannot slide over to see the bounce rate) but perhaps it's not the same experience on other screens where rather than cut from display they overflow:

photo_2021-02-13_09-23-35

Vigasaurus:
Can do!

Okay cool, will do. 

I meant to replace one of the two current instances of bounce rate (currently it's both in top pages modal and entry modal) - but maybe any change there would end up just being temporary ahead of #591 anyways. 

Yeah the modals and a lot of the text overflow is problematic still, it's generally been looking like my fixes to the drop-downs are working alright on mobile; I'll see how I can improve them when I get to them. And no, currently all of them get cut off or compressed, which is of course problematic either way.

metmarkosaric:
Ahh I understand. Yeah, I like that. We could remove Bounce Rate and include Visit Duration in the Entry Pages report. Then it would have Unique Entrances, Total Entrances and Visit Duration.

Vigasaurus:
This should be pretty much ready for review. I might go back through and do some refactoring and code cleanup tomorrow, but it should be in its final state as far as my first pass goes.

metmarkosaric:
Nice, thanks very much @Vigasaurus! Looking forward to checking this out!

Vigasaurus:
Actually now that I think about it, I think visit duration in entry pages is misleading; I think users would think that is the time on the specific page, not the entire visit. 

I'm looking into getting time-on-page working (it's a bit complex a query) - which I think would be a better column there if I can get it working. Otherwise, perhaps we should remove the visit duration from entry pages all together.

metmarkosaric:
"Time on page" is another popular request but it makes most sense to be in the "Top Pages" report itself. It shows average time spent on a specific page.

"Visit duration" fits for the "Entry pages" report as it shows the average session/visit duration for those visits/sessions that started with an entrance on a specific page. 

This is exactly the same way GA displays "Session Duration" in their "Entry Pages" report and something people are used to.

ukutaht:
@Vigasaurus I'm a bit swamped at the moment. Looks great, I'll get to review it later this week :)

Vigasaurus:
@ukutaht All good, and thanks! Take your time for sure. 

I'm still whacking about at getting the query for generating time on page given what data is already stored, but its seeming like clickhouse is missing what SQL has to make something like that straightforward. If I get that working any time soon I'll add that to this PR.

Vigasaurus:
Also I actually just noticed a bug where I compute exit rates for goal-based filters completely wrong (if not crashing all together) - so I'm gonna work on solving that.

ukutaht:
**General thoughts**

Testing it out for correctness and usability first before code review.

I ❤️❤️❤️ seeing entry pages for goals. Now we have a much better idea which blog posts and landing pages are driving conversions. Not in scope for this PR but in the future I'd also love to get a conversion rate for each entry page (as well as each source, country, etc).

After playing around with it for a while I don't see that much value in surfacing so many different numbers. I am not that deep into analytics but I think optimizing the signal to noise ratio is an important part of the product. Personally I think I would prefer two reports:

Top pages: Total pageviews, visitors, exit rate
Entry pages: Total entrances, bounce rate, visit duration

This doesn't really work if we want to add Time on Page and roll bounce rate and visit duration into a single metric. I'm just throwing it out there as a discussion point. How much value is there in seeing total vs uniqe exits, for example? To me it's just noise but maybe there are people who think this is really important.

/General thoughts

**Some specific issues I found**

1. Exit pages does not load when using a filter (I was testing with `screen=Desktop`)
2. When clicking on a row in 'entry pages', let's say `/plausible.io`, then it should add a filter `entry_page=/plausible.io`. Currently it adds a filter `page=/plausible.io`.
3. Same for exit pages. There should be a new filter `exit_page=/plausible.io` instead of adding a `page` filter when clicked.
4. Add a goal filter. Then click on a row in 'exit pages'. The whole dashboard throws an error `TypeError: page is null`
5. Scroll down in exit pages modal. Click on 'load more'. The whole dashboard throws an error `TypeError: page is null` (no filters)

2 and 3 are likely intentional? The `page` filter is weird in general and there are so many strange interactions it can have with other metrics/filters. But I think it's worth introducing new filters for entry/exit pages as well as to redefine what is loaded in the entry/exit page reports when a `page` filter is being used.

For example, what should be shown in 'entry pages' when we have applied a `page=/plausible.io` filter? Currently it shows how many entrances there were from `/plausible.io`. I think that's a reasonable thing to show when using `entry_page=/plausible.io`. But when using `page=/plausible.io`, I would argue that in entry pages we should show the entry pages for sessions that ended up viewing `/plausible.io` at some point in the session. Does that make sense?

Same for exit pages. When using a `page=/plausible.io` then it makes sense to show where people drop off after viewing `/plausible.io`

metmarkosaric:
I personally see no need to do something different with this. All 3 are very common reports (top, entry, exit) and have their use cases. 

Removing bounce rate from top pages makes that report less useful to understand any sense of quality of a page. And when we get to add time on page and engaged visitors to top pages it will make that one report very crowded (especially without the option to search/filter/sort). 

Adding entry and exit like it is in the screenshots makes sense to me. Our signal will increase a lot by adding them and I see no extra noise or confusion added to the dashboard. It pretty much follows the design we already do for UTM and devices. Smart use of space and a valuable addition that makes Plausible much stronger in depth without adding any complexity.

We should just change "more" to "details" or something like that as even now some people don't understand that there's extra insights under "more" rather than just a longer list of the same.

Points 2 and 3 should definitely be part of this as they add a lot of value. Being able to filter the dashboard by a specific entry (or even exit) page I can see being very useful for a lot of different analysis even for us.

Last 2 paragraphs make perfect sense too. We should show session based data there.

Vigasaurus:
> I ❤️❤️❤️ seeing entry pages for goals. Now we have a much better idea which blog posts and landing pages are driving conversions. Not in scope for this PR but in the future I'd also love to get a conversion rate for each entry page (as well as each source, country, etc).

Yeah that sounds really cool, I think leaning into goals will definitely be a big thing. 

> After playing around with it for a while I don't see that much value in surfacing so many different numbers. I am not that deep into analytics but I think optimizing the signal to noise ratio is an important part of the product. Personally I think I would prefer two reports:
> Top pages: Total pageviews, visitors, exit rate
> Entry pages: Total entrances, bounce rate, visit duration

I can agree with that, only issue I could foresee is that without a dedicated Exit Pages report, I dunno how a user could select a `exit_page=/` filter, also as you mentioned, I think there are quite a few other columns which would want to get added (namely Time on Page) - so I worry that there will be too many columns in Top Pages like that. I do kinda think that seeing uniques vs total across the board is beneficial and not extraneous, but that may just be me - maybe @metmarkosaric (or a Twitter poll) could lend some better context? (Marko just put his comment above as I was typing, I definitely agree that swapping more for details should help ensure that it makes sense too. Maybe as more columns get added a FE method for hiding specific columns could help. 

>     1. Exit pages does not load when using a filter (I was testing with `screen=Desktop`)

Hmm, I thought I had fixed that, will take a second look. 

>     2. When clicking on a row in 'entry pages', let's say `/plausible.io`, then it should add a filter `entry_page=/plausible.io`. Currently it adds a filter `page=/plausible.io`.
>     3. Same for exit pages. There should be a new filter `exit_page=/plausible.io` instead of adding a `page` filter when clicked.

I think those make sense, should make everything more verbose and specific. 

>     4. Add a goal filter. Then click on a row in 'exit pages'. The whole dashboard throws an error `TypeError: page is null`
>     5. Scroll down in exit pages modal. Click on 'load more'. The whole dashboard throws an error `TypeError: page is null` (no filters)

I thought I fixed both of these too 🤔 - maybe I missed an edge case somewhere. 

> 2 and 3 are likely intentional? The `page` filter is weird in general and there are so many strange interactions it can have with other metrics/filters. But I think it's worth introducing new filters for entry/exit pages as well as to redefine what is loaded in the entry/exit page reports when a `page` filter is being used.

Yeah, I think doing them based on their specific pageview types per session would be great, I considered that in the original but didn't want to add the complexity without discussing it first.

Vigasaurus:
ngl, this is fantastic. It basically gives you funnels, without all the middle bits, (yet)
image

metmarkosaric:
Beautiful!

Vigasaurus:
Alright, this should be ready to go with the new changes. I got everything swapped over to work with entry and exit page filters; only things I wasn't able to test fully was `filter_converted_sessions` and `conversions_from_referrer` in `lib/plausible/stats/clickhouse.ex` - in theory they should swap to using `base_query_w_sessions` if they actually want to adhere to the exit and entry filters - but I couldn't quite figure out how/where those get called or how to properly test if they need said swap. 

Past that, I've now:
- Swapped More->Details
- Added the Entry and Exit Filters
- Fixed the issues with the exit page modal being broken
- Fixed the issue with the exit page report selection w/ filter

I do still need to do a bit more extensive testing, but I think it's mostly ready for review/comments - currently running on [my dev box](https://dev.vigneshjoglekar.com/secrethitler.io?date=2021-01-01&period=month), as always - but I think something like staging may also help with that, since my data set doesn't use any of the UTM stuff and doesn't get much referring traffic so I can't test that as heavily as I would want.

lib/plausible/stats/clickhouse.ex
```diff
@@ -155,7 +155,7 @@ defmodule Plausible.Stats.Clickhouse do
 
   def total_pageviews(site, query) do
     ClickhouseRepo.one(
-      from e in base_query(site, query),
+      from e in base_query_w_sessions(site, query),
```
ukutaht, line 158: Why the change here?

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
...
         select: %{
           name: s.entry_page,
-          count: fragment("uniq(?) as count", s.user_id)
+          count: fragment("uniq(DISTINCT ?) as count", s.user_id),
+          entries: fragment("count(DISTINCT ?) as entries", s.session_id),
+          visit_duration: fragment("avg(?) as visit_duration", s.duration)
         }
       )
 
+    q = add_goal_filter(site, query, q)
```
ukutaht, line 506: This seems to do a pretty similar job as `filter_converted_sessions/3`. Did you just miss it or is there a good reason to introduce a new function?

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
         offset: ^offset,
         select: %{
           name: s.entry_page,
-          count: fragment("uniq(?) as count", s.user_id)
+          count: fragment("uniq(DISTINCT ?) as count", s.user_id),
```
ukutaht, line 500: This applies to all the changes in this file: the rest of the project uses Clickhouse's `uniq(?)` function for getting uniques. No need for `uniq(DISTINCT ?)` or `COUNT(DISTINCT ?)`.

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
...
-      Enum.map(pages, fn url -> Map.put(url, :bounce_rate, bounce_rates[url[:name]]) end)
+    result = ClickhouseRepo.all(q)
+
+    if Enum.count(result) > 0 do
+      pages = Enum.map(result, fn r -> r[:name] end)
+
+      q2 =
+        from(
+          e in base_query_w_sessions(site, query),
+          select: %{session_id: fragment("DISTINCT ?", e.session_id)}
```
ukutaht, line 538: Each session already has a unique ID. Should be safe to remove `DISTINCT` from here.

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
...
+        )
+
+      event_q = from(e in base_query_bare(site,query),
```
ukutaht, line 541: I feel like this should work fine by just using `base_query_with_sessions/2` without having to join with `q2`. That base query already joins with sessions if any filters are being used.

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
...
+      from(
+        s in base_session_query(site, query),
+        group_by: s.exit_page,
+        order_by: [desc: fragment("count")],
+        limit: ^limit,
+        offset: ^offset,
+        select: %{
+          name: s.exit_page,
+          count: fragment("count(DISTINCT ?) as count", s.user_id),
+          exits: fragment("count(DISTINCT ?) as exits", s.session_id),
```
ukutaht, line 524: Session IDs are already unique, no need to uniq again.

lib/plausible/stats/clickhouse.ex
```diff
@@ -497,25 +497,74 @@ defmodule Plausible.Stats.Clickhouse do
...
+            fragment("count(*)")
+          }
+      )
+
+      total_pageviews = ClickhouseRepo.all(event_q) |> Enum.into(%{})
+
+      Enum.map(result, fn r ->
+        if Map.get(total_pageviews, r[:name]) do
+          exit_rate = r[:exits] / Map.get(total_pageviews, r[:name]) * 100
+          if exit_rate > 100 do
```
ukutaht, line 557: How can exit rate be over 100?

Maybe I'm missing something but it feels like a logical error. There should never be more sessions that exited on a certain page than pageviews in total on that page.

ukutaht:
FYI, this can be tested on https://staging.plausible.io/plausible.io

ukutaht:
@Vigasaurus Not sure which version of Clickhouse you're testing this with. On our staging (which is the same as prod), I get the following error:

```
    ** (Clickhousex.Error) Code: 46, e.displayText() = DB::Exception: Unknown function uniqDistinct (version 20.3.8.53 (official build))

```

ukutaht:
It seems to me like the entry/exit page queries are not quite right when a page filter is being used. When I click on a row in Top Pages to add a `page=/plausible.io` filter for example, the entry/exit page reports don't change.

lib/plausible/stats/clickhouse.ex
Vigasaurus: This needed the swap to join with sessions because otherwise the exit and entry filters wouldn't apply - this way they're accounted for appropriately (they're only available on sessions, obv).

lib/plausible/stats/clickhouse.ex
Vigasaurus: I 100% just missed it - removed and swapped to existing 👍

lib/plausible/stats/clickhouse.ex
Vigasaurus: Makes sense - changes made everywhere.

lib/plausible/stats/clickhouse.ex
Vigasaurus: Yeah that makes sense - I had done it the different way originally because of the whole "needing distinct on sessions", but as you pointed out earlier that was unnecessary. Changed to base_w_sessions 👍

lib/plausible/stats/clickhouse.ex
Vigasaurus: Originally when I had to have that in there with the joins in exit_pages, there would be occasions where with filters like Screen Size, that exist both in sessions and events (and aren't part of the user_id generator) where a user would have resized a window and had a different screen size on the event, bringing its counting down for the filter (also at least partially  was related to the uniqDistinct error) - but with the removal of the double join (and the uniq distinct) that doesn't seem to happen anymore, in my dataset at least. If it reproduces in staging, or for other users, that is the reason, most likely

Vigasaurus:
> 
> 
> @Vigasaurus Not sure which version of Clickhouse you're testing this with. On our staging (which is the same as prod), I get the following error:
> 
> ```
>     ** (Clickhousex.Error) Code: 46, e.displayText() = DB::Exception: Unknown function uniqDistinct (version 20.3.8.53 (official build))
> ```

I'm not exactly sure what the cause of this could be - I have to assume it's the places where I was using `uniq(distinct session_id)` - maybe just changing those to `uniq`s fixed this? I've never seen it, but lmk if it shows up on staging again after my most recent changes and I can poke around. 


> It seems to me like the entry/exit page queries are not quite right when a page filter is being used. When I click on a row in Top Pages to add a `page=/plausible.io` filter for example, the entry/exit page reports don't change.

This should now be fixed, I was missing a pretty glaring part of generating sessions that had visited a page (to do this I changed up `filter_converted_sessions` to also apply when a page filter was set, but I'm not entirely sure that I did it the "Elixir-way" - the current is like:
```elixir
defp filter_converted_sessions(db_query, site, query) do
    goal = query.filters["goal"]
    page = query.filters["page"]

    if is_binary(goal) || is_binary(page) do
      converted_sessions =
        from(e in base_query(site, query),
          select: %{session_id: fragment("DISTINCT ?", e.session_id)}
        )

      from(s in db_query,
        join: cs in subquery(converted_sessions),
        on: s.session_id == cs.session_id
      )
    else
      db_query
    end
  end
  ```
  but it could potentially be more "elixir-style" doing it this way:
  ```elixir
  defp filter_converted_sessions(db_query, site, %Query{filters: %{"goal" => goal}} = query)
       when is_binary(goal) do
    converted_sessions =
      from(e in base_query(site, query),
        select: %{session_id: e.session_id}
      )

    from(s in db_query,
      join: cs in subquery(converted_sessions),
      on: s.session_id == cs.session_id
    )
  end

  defp filter_converted_sessions(db_query, site, %Query{filters: %{"page" => page}} = query)
       when is_binary(page) do
    converted_sessions =
      from(e in base_query(site, query),
        select: %{session_id: fragment("DISTINCT ?", e.session_id)}
      )

    from(s in db_query,
      join: cs in subquery(converted_sessions),
      on: s.session_id == cs.session_id
    )
  end

  defp filter_converted_sessions(db_query, _site, _query), do: db_query
  ```
  lmk. 
  

I did also have to change the tests a bit, as there was an instance where there were 3 exits on `test-site.com``/`, on `2019-01-01`, but only 2 pageviews for the same, which as discussed, doesn't make sense and is impossible as far as data is concerned. I simply added an additional pageview on the same day to accomodate this and make the exit_rate the more accurate 100%.   


All the rest of the comments should be resolved/answered - will love to see the new changes up on staging, I think it's in a great spot now.

File: lib/plausible/stats/clickhouse.ex [hunk 8]
```diff
@@ -291,20 +291,24 @@ defmodule Plausible.Stats.Clickhouse do
...
 
-    from(s in db_query,
-      join: cs in subquery(converted_sessions),
-      on: s.session_id == cs.session_id
-    )
-  end
+    if is_binary(goal) || is_binary(page) do
+      converted_sessions =
+        from(e in base_query(site, query),
+          select: %{session_id: fragment("DISTINCT ?", e.session_id)}
```
Vigasaurus, line 301: This is the one spot that does actually need the `DISTINCT session_id`, as this is only used to join, and when there are multiple successful hits per session, every instance of the session_id gets returned as its own row, and then events get joined to every instance of every session (even on repeated `session_id`s), thus ending up with lots of duplication and super skewed numbers in `base_query_w_sessions_bare`, for example (which propagates through entry and exit pages).

ukutaht (commented):
Looking good!

Three things left on my radar at the moment:
* I cannot integrate this work until I've fixed the historical exit page data. Waiting on the Clickhouse repo to get some help with that
* I noticed the loading states are different between the Top Pages and Top Sources reports now. During loading, Top Pages shows the header and tabs but Top Sources doesn't. I don't really have a preference, probably showing what we can is a good things, but the loading states should be consistent. Up to you which style you go with.
* I think the query for total pageviews for exit rate can be simplified. Not a fan of sending boolean flags to modify the behaviour of a function. I worked out a simpler way, can you confirm this works without regressions? https://github.com/plausible/analytics/compare/pr/712...pr-712-queries

ukutaht:
Found a regression: When adding a filter `page=/blog/remove-google-analytics` for example, the `Bounce Rate` stat at the top does not show bounce rate for the page, it shows the overall bounce rate.

This is caused by removing [this part](https://github.com/plausible/analytics/blob/54b6acf6b0ca24e6a9478d9cd2266b54bdd0baac/lib/plausible/stats/clickhouse.ex#L1150) from the `base_session_query`. I think it should be added back in which would break the behaviour we want for entry and exit page reports with a `page` filter. For those we could modify the Ecto query like so:
```
from(s in base_session_query(site, %Query{query | filters: Map.delete(query.filters, "page")}))
```

BTW @Vigasaurus this code is some of the most tangled stuff I've written... I'm basically rewriting the query interface for the API to see if I can find some good abstractions that make reasoning about this easier. If you have ideas let me know. Everything is pretty basic and easy except for making the `page` and `goal` filters work properly.

Vigasaurus:
>     * I cannot integrate this work until I've fixed the historical exit page data. Waiting on the Clickhouse repo to get some help with that

Yeah, absolutely. Surfacing completely wrong data is definitely 😬

>     * I noticed the loading states are different between the Top Pages and Top Sources reports now. During loading, Top Pages shows the header and tabs but Top Sources doesn't. I don't really have a preference, probably showing what we can is a good thing, but the loading states should be consistent. Up to you which style you go with.

Yep makes sense. I think `Devices` also does the former, so will probably just make them match that. I wonder if we should change all of them (Countries and Goals) to also show their headers during load - just to make them all consistent. @metmarkosaric And preference here? I don't really see loading screens enough to know/have a preference. 

>     * I think the query for total pageviews for exit rate can be simplified. Not a fan of sending boolean flags to modify the behaviour of a function. I worked out a simpler way, can you confirm this works without regressions? [pr/712...pr-712-queries](https://github.com/plausible/analytics/compare/pr/712...pr-712-queries)

Yeah I definitely felt a bit gross doing that, will take a look at your way 👍 . I will say, I don't *fully*understand that format, elixir pattern matching is weird; but I get the gist. 

>Bounce Rate

Don't you think we can now properly use the page-based filters for bounce rate? i.e.
Page: `/` - All sessions that have at least one event that's a pageview on `/`
Entry Page: `/` - All sessions filtered by entry
Exit Page: `/` - All sessions filtered by exit

I'm not necessarily a huge fan of having that filter mean something that it isn't (especially not when that other thing is also an available filter) with bounce rate. This will obviously add some more complexity to that bounce rate calculation, but I think it makes more sense like that. But honestly I can see it both ways in that page for bounce rate may be accurate-ish. Thoughts?

>Tangled

Yeah I think pulling stuff up would definitely help, but it also kinda feels like pulling stuff up and keeping the bases very bare would end up with a decent chunk of repetition in each upper function - this is exemplified with how many of the current functions are able to use the more complex bases so well, and with redone bases they'd need to do all of that themselves. Honestly though, I think the actual query generation bits are alright and very powerful at the current moment, they just get a bit messy with how much each function does alone - e.g. base_w_sessions_bare - and their names aren't necessarily super conducive to show what they do. So I wonder if just splitting up pieces - e.g. reusing a version of `base_session` as `base_session_bare` and then using that in the top half of `base_w_sessions_bare` and similar could make the entire process more DRY and potentially easier to understand. 

As far as page and goals are concerned, I think the biggest thing for pages is going to be coming up with a consistent definition for how you want pages to function now with exit and entry pages - i.e. should it represent all sessions *ever* visited the page, and ensuring that every function uses that consistent definition (I think they currently do, but this could get out of hand fast). As for goals, I think the biggest pain point is just that some queries absolutely shouldn't use goals (the `bare` ones), but aren't necessarily consistent or super clear as to which function uses what (again, just naming). 

To be clear though, I do think the current system is very powerful and honestly pretty straightforward once it becomes clear, but it just sometimes take a while of prodding to get everything clear.

Vigasaurus:
> I think the query for total pageviews for exit rate can be simplified. Not a fan of sending boolean flags to modify the behaviour of a function. I worked out a simpler way, can you confirm this works without regressions? pr/712...pr-712-queries

This doesn't quite work the way you'd want, because while removing the page filter works to satisfy that second condition on applying the `pathname=page` where condition, it doesn't add the page goal as part of the converted session, nor does it take into account the goal. I'll see if I can put together a better way of doing it, maybe by just writing a different version of base_w_sessions_bare just for exits, to prevent clogging up the basic version.

ukutaht:
New query looks good! Thanks for making the change.

**On bounce rate**
> Don't you think we can now properly use the page-based filters for bounce rate? 

As a programmer I agree that this feels nice and would make the backend code cleaner. But it could lead to very questionable user experience.

Let's say someone clicks on the `/vs-matomo` page in their Top Pages report and sees a bounce rate metric. The most natural interpretation of it would be that we're showing a bounce rate for the `/vs-matomo` page. Not for all sessions that at some point visited `/vs-matomo`. And looking at some percentage it's difficult for the user to know which one we're showing.

Same would go for the 'Total pageviews' number, for example. If we accept a common definition that a `page` filter is used to segment sessions, then it would have to show total pageviews from all sessions that at some point visited `/vs-matomo`. This would include pageviews on other paths within the session. But a more natural number to show is the number of pageviews specifically on `/vs-matomo`.

BTW when building the API I've checked how Google Analytics works to make sure we're not too far off industry standard. Everything we do is pretty consistent with them. When you request `ga:bounceRate` to be broken down by `ga:pagePath` then it gives you the equivalent of getting bounce rates for `ga:entryPage`.

> I'm not necessarily a huge fan of having that filter mean something that it isn't (especially not when that other thing is also an available filter) with bounce rate.

I would love to able to have a consistent definition of what the `page` filter means. But user expectations have driven us to this situation where it can mean different things depending on the context. The root of the problem is we're mixing concepts between the `event` and `session` levels. But all this mixing is done in the name of better UX and information architecture for the user. If I had to give a definition for the page filter it would be something like:

_When querying for bounce_rate and visit_duration then page means entry_page. When querying for entry and exit pages or using a goal filter, then page filter segments sessions that visited that page. When querying for total pageviews, then page filter applies at the event level without sessions taken into account_

I'm open to brainstorming some ways we could make the page filter consistent without confusing our users. Maybe we could figure out ways of labelling things differently so they are easier to understand

Vigasaurus:
Yeah that makes sense, I kinda figured it would be the case. The programmatic approach is of course what I'd prefer but that UX would definitely be problematic for anyone not familiar with the actual setup. (And I'm sure a couple months ago, before I looked at any of the backend, I'd have agreed that the page filter alone should indeed mean entry page in those contexts)

And considering that overall visit duration and overall bounce rate are now the only two remaining instances where this is a concern, I think kicking the can down the road for better explanations and preferring the UX is 100% the right call. 

I'll make the relevant change there :+1:

ukutaht (approved):

Vigasaurus:
👏 this should be a good one, think people will like it ~~I totally haven't been running it on my self-hosted box for the past 4 days already 👀~~

metmarkosaric:
looking forward to playing with this one too! thanks @Vigasaurus!

PR #744, opened 2021-02-22
Prevent user id collisions in session tracking

ukutaht:
### Changes

We keep all active sessions in memory. Currently they are keyed with just `user_id` which has caused some collisions to happen. Basically, two different users on two different sites can end up affecting a single session in the database. This PR completely splits them so even if two users on different sites happen to get the same `user_id`, their session stats won't be mixed up.

There is still historical data that is wrong. I will probably rebuild the sessions table on the production server.

### Tests

### Changelog

### Documentation

### DIFF
--- a/lib/plausible/session/store.ex
+++ b/lib/plausible/session/store.ex
@@ -30,7 +30,7 @@ defmodule Plausible.Session.Store do
             on: s.session_id == ls.session_id and s.timestamp == ls.timestamp,
             order_by: s.timestamp
         )
-        |> Enum.map(fn s -> {s.user_id, s} end)
+        |> Enum.map(fn s -> {{s.domain, s.user_id}, s} end)
         |> Enum.into(%{})
       rescue
         _e -> %{}
@@ -48,7 +48,11 @@ defmodule Plausible.Session.Store do
         _from,
         %{sessions: sessions, buffer: buffer} = state
       ) do
-    found_session = sessions[event.user_id] || (prev_user_id && sessions[prev_user_id])
+    session_key = {event.domain, event.user_id}
+
+    found_session =
+      sessions[session_key] || (prev_user_id && sessions[{event.domain, prev_user_id}])
+
     active = is_active?(found_session, event)
 
     updated_sessions =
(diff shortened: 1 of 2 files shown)


ukutaht:
tagging @Vigasaurus. I need to integrate these changes before entry/exit pages can go live. Marko found this issue by scrolling all the way down in exit pages.

Vigasaurus:
Huh that's super interesting - looking at the exit pages report on staging it totally makes sense, but I'd probably have never noticed it myself. 

Is there already a pre-existing way to rebuild the sessions table, or are you going to write something for it?

ukutaht:
> Is there already a pre-existing way to rebuild the sessions table, or are you going to write something for it?

Nope. Luckily the session data can be completely inferred from historical event data.

I'll see today if I can find a way to do it without rebuilding the whole table. Clickhouse does support updates so maybe just rebuilding the `exit_page` field of the `sessions` table is enough.

ukutaht:
Waiting for comment in https://github.com/ClickHouse/ClickHouse/issues/21099

Vigasaurus:
Dunno if you've already seen/tried this, but this seems like a promising way to do this: https://stackoverflow.com/questions/55764093/updating-rows-with-values-from-another-table-in-clickhouse

ukutaht:
Might work. I'll give it a go later

Vigasaurus:
I was able to get a working version - not sure if it's necessarily the most optimal, but it definitely seems to work (and took <1 second on a db with 400k sessions, so it's definitely not *terribly* slow). I think all of these are synchronous in the background though, so doing them one-by-one and checking `mutations` in between/ensuring they finish is probably the *right* way to do it, but I'm not sure. 

```sql
CREATE TABLE exit_pages (session_id UInt64, exit_page String) Engine = Join(ANY, LEFT, session_id);

INSERT INTO exit_pages SELECT session_id, anyLast(pathname) as exit_page from events group by session_id;

ALTER TABLE sessions 
UPDATE exit_page = joinGet('exit_pages', 'exit_page', session_id)
WHERE 1;

DROP TABLE exit_pages; -- optional, just cleanup
```

ukutaht:
@Vigasaurus Thanks for experimenting with this.

I'll see if I can rewrite the same update process to work with a compound key of `(domain, session_id)` instead. I believe that should fix the issue.

ukutaht:
This seems to work well in my testing. Could you confirm with your dataset that the issue you saw previously does not happen when you use this migration code instead?

```sql
CREATE TABLE exit_pages (domain String, session_id UInt64, exit_page String) Engine = Join(ANY, LEFT, domain, session_id);

INSERT INTO exit_pages SELECT domain, session_id, anyLast(pathname) AS exit_page FROM events GROUP BY domain, session_id;

ALTER TABLE sessions UPDATE exit_page = joinGet(exit_pages, 'exit_page', domain, session_id) WHERE 1;

DROP TABLE exit_pages;

```

Vigasaurus:
I'll try running it in a bit, but from first glance, it wouldn't fix it I don't think. That table I showed a screenshot of was events, and theoretically one of those events should have had a completely different session. While yes, this would be the migration to recalculate exit pages after this PR gets merged and the session_id calculation was changed, I don't think this actually solves the issue of changing the session_id of events on affected sessions to delink them from events actually linked, nor does it make additional sessions for the user_jd+domain combinations that didn't have a unique session_id associated with them originally. 

Does that make sense? That might have been a bit convoluted an explanation. 

I think the only real solution here would be to recreate session_id's by getting every unique pair of domain+user_id from events, splitting them by whatever time length dictates an abandoned session (and making more session_ids as needed) (I'm not sure how the abandoned session stuff works, so this part may be wrong), altering in all those newly computed session_ids into the events table accordingly, and then finally going through all those newly calculated sessions and generating rows in the sessions table accordingly.

ukutaht:
I think clashing `session_id` values are not as terrible as you think. While I agree that rebuilding the session_ids would make it proper _correct_, there should be no difference in practice. When querying stats, we always query the `session_id` and `user_id` in combination with a `domain` value. So even though there might be unrelated events/sessions with the same session ID, they will never mix in the UI because they have different `domain` values.

This data migration is supposed to just fix the `exit_page` values for the sessions, not rebuild them from scratch. I believe this is data migration and PR are all we need to get correct exit_page data. 

Our prod server has about 50GB data and I really want to  avoid rebuilding half of that. I can live with some bad session IDs that don't really create any problems for us in real life. As long as the data we present to users is correct I'm happy.

We'll always have hash collisions for `user_id` and randomness collisions in `session_id` since we're dealing with a lot of data. Best we can do is scope these values to a certain `domain` to minimize the effect of these clashes and to make sure we don't mix session data between two sites.

Vigasaurus:
Makes sense - and also, your query did definitely fix my table to that extent, the exit page does point to the correct domain's exit, so that should be all great :D

Issue #241, opened 2020-07-25
Regex in Pageview Goals path


alpha077412:
## Prerequisites
- [x] I have searched open and closed issues to make sure that the feature has not yet been requested.

## Feature request
**Is your feature request related to a problem? Please describe.**
Currently Pageview Goal paths must be identical to those tracked in Plausible, meaning it is not possible to create goals for dynamic URLs without using custom events. For example, a Woocommerce checkout page redirects users to _/checkout/order-received/12345/_ where the final part of the URL is unique for each order. 

**How would you like it to work**
Allow entry of regex or more simply an asterisks * in Pageview Goal paths. For example, _/checkout/order-received/*_

ukutaht:
Makes sense, thanks for the request!

youngbrioche:
Plausible should at least allow for partly URL matching (e.g. „ends with“). Would make it possible to track pageviews like

- /de/signup/thanks
- /en/signup/thanks

Without configuring two goals.

n0th:
Yes, please, this is sooo important for dynamic pages like Woocommerce

PR #750, opened 2021-02-24
Adds same glob patterns as exclusions for pageview goals

Vigasaurus:
### Changes

screenshot

Addresses #241

### Tests

### Changelog

### Documentation

Although I must say, I sometimes feel bad with how many PRs I've forced you to review recently heh, this should be the last one for a bit though


ukutaht:
Code looks good. This should be thoroughly tested for sure.

Vigasaurus:
> Code looks good.

Huh, honestly a bit surprised there. I guess my Elixir has gotten pretty decent after all - this felt a bit hacky to me at first to be honest, but doing it any other way than a separate CH query-per-goal was causing significant issues with counting uniques correctly, so I settled for this setup. (This code is technically v3, heh)

>This should be thoroughly tested for sure.

100%. I'll write significant tests for this, partially because my elixir string replacement/regex is weak anyways, and that I had multiple instances of the clickhouse `match()` function not working as I'd generally expect, so there are decent odds that the same may present itself again. 

As a note here actually, do you know of an easy (or not easy, if needed, I suppose) way for me to have `mix test` set up the test CH db, but not proceed with any more of the tests? In other words, keep the DB initialized as it would be at test-time for me to run CLI queries on it.

ukutaht:
Ah, I didn't actually notice the N+1 query here. Would be good to combine it all back into a single query in case there are a lot of goals. There seem to be loads of functions designed for searching multiple different regex patterns in CH: https://clickhouse.tech/docs/en/sql-reference/functions/string-search-functions/#multimatchanyhaystack-pattern1-pattern2-patternn

> As a note here actually, do you know of an easy (or not easy, if needed, I suppose) way for me to have mix test set up the test CH db, but not proceed with any more of the tests?

You can try setting a `IEx.pry()` somewhere in your test code. It will stop the execution but keep the environment up without tearing it down. More info: https://hexdocs.pm/iex/IEx.html#pry/0

Then you can run arbitrary Elixir or connect to your CH with their client.

Vigasaurus:
Hmm some of those CH functions actually seem promising now that I think about it. The hard part here (and the problem with v1 and v2 of my code) were that I could really only group over the basic pathname effectively, which made unique user_id's be double counted per every basic pathname. I wonder if there's a way to group by something other than a basic column; I'll look into it.

ukutaht:
> Hmm some of those CH functions actually seem promising now that I think about it. The hard part here (and the problem with v1 and v2 of my code) were that I could really only group over the basic pathname effectively, which made unique user_id's be double counted per every basic pathname. I wonder if there's a way to group by something other than a basic column; I'll look into it.

Yes we'd need to find a way to group over the matching regex basically. I wonder if you could use `multiMatchAnyIndex` to get the index of the matching regex. Then group by the index. And after the query in elixir you could map the index of the match back to the regex/goal name that corresponds to it.

Vigasaurus:
Yeah that sounds good, I was thinking about pretty much the same, but with the actual regexes. Doing it just over the index might be better and cleaner though. Will poke around and see what I can get 👍

lib/plausible/stats/clickhouse.ex
```diff
@@ -870,25 +870,29 @@ defmodule Plausible.Stats.Clickhouse do
...
-          select: %{
-            count: fragment("uniq(user_id) as count"),
-            total_count: fragment("count(*) as total_count")
-          }
-        )
-        |> ClickhouseRepo.one()
-        |> Map.put(:name, "Visit #{x[:name]}")
+      from(
+        e in base_query_w_sessions(site, query),
+        where: fragment("multiMatchAny(?, array(?))", e.pathname, ^regex_goals),
```
ukutaht, line 880: Feels like this query is going to run the regex engine twice: once for the `where` clause and once for the index.

What does `multiMatchAllIndices` return if there's no match? Maybe the where clause can use that return value for filtering instead so we can avoid running the regex twice for each entry in the database.

Vigasaurus: Yep I'm still experimenting with that bit - seemingly this works, just need to port it to Ecto and test it out, but it seems promising. 

```sql
SELECT
	uniq(user_id) as count,
	count(*) as total_count,
	arrayJoin(indices) as i
FROM
	plausible_events_db."events" AS e0
WHERE
	(e0."domain" = 'secrethitler.io')
	AND (e0."name" = 'pageview')
	AND notEmpty(multiMatchAllIndices(pathname, ['(^/rules/?$)','(^/tou/?$)','(^/how-to-[^/]*/?$)','(^/game/#/table/.*/?$)', '(^/game/#/table/[^/]*/?$)']) as indices)
group by i
```

lib/plausible/stats/clickhouse.ex
```diff
@@ -863,24 +863,37 @@ defmodule Plausible.Stats.Clickhouse do
...
+        select: %{
+          index: fragment("arrayJoin(indices) as index"),
+          count: fragment("uniq(user_id) as count"),
+          total_count: fragment("count(*) as total_count")
+        },
+        group_by: fragment("index")
+      )
+      |> ClickhouseRepo.all()
+      |> Enum.map(fn x ->
+        Map.put(x, :name, "Visit #{Enum.at(goals, x[:index] - 1)}") |> Map.delete(:index)
```
Vigasaurus, line 895: This probably needs a `|| 0`

lib/plausible/stats/clickhouse.ex
```diff
@@ -863,24 +863,37 @@ defmodule Plausible.Stats.Clickhouse do
...
-          select: %{
-            name: fragment("concat('Visit ', ?) as name", e.pathname),
-            count: fragment("uniq(user_id) as count"),
-            total_count: fragment("count(*) as total_count")
-          }
-        )
+    if Enum.count(goals) > 0 do
+      regex_goals =
+        Enum.map(goals, fn g ->
+          "(^#{g}\/?$)"
```
Vigasaurus, line 873: This can be cleaned up a bit, it's still in my old format from when I was concatenating all the regexes into one regex

lib/plausible/goal/schema.ex
```diff
@@ -18,17 +18,22 @@ defmodule Plausible.Goal do
...
+      |> add_error(:page_path, "this field is required and must start with a /")
     end
   end
 
-  defp present?(changeset, field) do
-    value = get_field(changeset, field)
-    value && value != ""
+  defp validate_page_path(changeset) do
+    value = get_field(changeset, :page_path)
+    value && String.match?(value, ~r/^\/.*/)
```
Vigasaurus, line 32: I'd like to do better validation of pathnames here, and add good validation for preventing duplicate goals; but I didn't want to make a disgusting regex if you knew of an easy Elixir helper for validating URLs (this may not work I guess because of the blobs), and I couldn't get the latter to work.

Vigasaurus:
Alright, I think this should be ready for review now. I do again still need to do more testing across the board, but I figure if you see any issues off the bat I can handle them beforehand. 

I did just add the one main test, adding more tests, or more specific tests would need a lot more detailed events added to test that out effectively. If you think those should be added, I can do that, I just didn't want to immediately add a bunch of events there for this one test alone unless necessary. 

I'll do docs later.

lib/plausible/goal/schema.ex
ukutaht: I think it's OK as is.

lib/plausible/stats/clickhouse.ex
```diff
@@ -924,24 +924,37 @@ defmodule Plausible.Stats.Clickhouse do
...
+          |> String.replace(~r/\*\*/, ".*")
+          |> String.replace(~r/(?<!\.)\*/, "[^/]*")
+        end)
 
-      ClickhouseRepo.all(q)
+      from(
+        e in base_query_w_sessions(site, query),
+        where:
+          fragment(
+            "notEmpty(multiMatchAllIndices(?, array(?)) as indices)",
```
ukutaht, line 943: Nice, this looks good to me.

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -79,4 +79,53 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
              ]
     end
   end
+
+  describe "GET /api/stats/:domain/conversions - with glob goals" do
+    setup [:create_user, :log_in, :create_site]
+
+    test "returns correct and sorted glob goal counts", %{conn: conn, site: site} do
```
ukutaht, line 86: Are double asterisks (`**`) supported? If so, we should also test for them

ukutaht (approved):

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
Vigasaurus: They are; currently none of the existing test events could really benefit from it (they're all single-depth paths), but I can add a couple events to do that if you want 👍

ukutaht: Yeah I think that would be valuable

lib/plausible/goal/schema.ex
herminiotorres: What do you think about it to extract this regex and put in the module attribute? And use a better name to describe the intentions of what this regex do...

ukutaht (approved):

ukutaht:
Thanks @Vigasaurus !

ukutaht:
The goals seem to work OK but it also needs to be implemented as a filter. Currently, when you click on the goal you land on a broken dashboard:

https://plausible.io/plausible.io?goal=Visit+%2Fblog%2F**

metmarkosaric:
Yeah, when you click on the goal, all metrics show 0 and "no data yet", You can see it on our live demo for "Visit /blog/**".

@Vigasaurus: This feature could be very cool to add to our "Top Pages" report to make it easier to use. When you click on "details" in Top Pages you could get a search box at the top which allows you to type whatever you want to filter the pages by that. I could type "blog" which should only list pages that include "blog" in their paths.

And then we could add one header to that table that summarizes the numbers of the pages you searched for:

* you could see percentage of all pageviews that the searched for pages stand for (say "blog" pages have 20% of total pageviews in the time period)
* the total number of pageviews between the searched for pages (say "blog" pages have 2335 pageviews in total in the time period)

Vigasaurus:
Oh yeah huh I never even thought to test for it being used as a goal. Fixing that should be not that terrible though, maybe. I'll look into it.

PR #757, opened 2021-02-25
Add metadata to annotate the JavaScripts according to LibreJS

bhavin192:
### Changes
Reference: https://github.com/plausible/analytics/discussions/719#discussioncomment-374238

- Uses the web labels method for .js files being generated, the table
  is available at the /licenses page.
- References this table with a visible link on the main footer and with
  invisible link on other layouts.
- Annotate the other inline scripts with magnet links to the licenses.
- Set the license of assets/package.json to AGPL-3.0-or-later from
  MIT (old).
- Reference:
  https://www.gnu.org/software/librejs/free-your-javascript.html

This fixes https://github.com/plausible/analytics/discussions/719

### Tests

### Changelog

### Documentation

### Open questions
- I have added the link `JavaScript license` to the  'Why Plausible?' section of the footer, other section were already crowded. Where should I put it?
- Is the `app.js` a compiled form of all the dependencies we use? If that is the case, I think we need to mention all the licenses in the table.


metmarkosaric:
thanks @bhavin192! is there a good reason to add this to the footer across the site? and also to have an "invisible link"? both of these seem like bad SEO moves that may hurt our search traffic

bhavin192:
@metmarkosaric yeah, the link with a `data-jslicense=1` needs to be on all the pages which use one the JavaScripts mentioned in the table. I have added the invisible link because those pages don't have the footer and just have a minimal one line at the end.

> Finally, you must add a link to the new page from all the pages loading these files. You can tell LibreJS about the link by setting either rel="jslicense" or data-jslicense="1" on the <a> tag. 
>  rel="jslicense" doesn't pass HTML5 validation, so if that's an issue you can safely use data-jslicense="1". 

Point number 3.1 JavaScript Web Labels, https://www.gnu.org/software/librejs/free-your-javascript.html

Will having some more info on the licenses page help with maintaining SEO?

metmarkosaric:
i guess we could noindex it but invisible links can cause penalties. i still don't understand why this is required to be put on all the pages just to prove that our little javascript file is open source? we can put it on the about page or on github like we do with all the other licensing stuff. where it's relevant.

bhavin192:
Ah, so there is this tool called LibreJS which looks for such link and finds the table from there. If it does not find a license for a script it just blocks it.

metmarkosaric:
ok thanks. LibreJS doesn't seem to be that popular, only for firefox and also looking at their firefox plugin it doesn't seem to be installed that much even on firefox.

does this also mean that all of our customers would need to include license info on their pages? Or if we include it on our site even those who use our JS on their pages won't have it blocked?

bhavin192:
For the users of the tracker:
The only option is that the users have to add a similar table on their site. Other option is to add comments in the tracker script which we can't do, as it will increase the size of the script. I have tried to explain it here: https://github.com/plausible/analytics/discussions/719#discussioncomment-371127

metmarkosaric:
Ok thanks, I'm personally not sure I see much benefit in doing this. I see more potential negatives from the SEO side of things than any benefits.

bhavin192:
Can you help me understand the SEO related issue? May be you can point me to some links? Is the invisible link the only issue? Or the dedicated license page as well?

metmarkosaric:
Both can cause us trouble. The risk may be low but as we don't do paid ads, we rely on organic search traffic. So I feel any changes that may hurt us in SEO (even low risks) need to have an otherwise strong benefit for us to do. Which I'm not sure I see in this case.

You can search for invisible links SEO and something like this shows up: https://developers.google.com/search/docs/advanced/guidelines/hidden-text-links
Same for thin content SEO or footer site-wide links SEO. There's a lot of different angles you can discuss this one on (it's a thin content page without much value to humans and then we link to it from the footer which dilutes our link equity to important pages without adding anything important back).

bhavin192:
I can understand your concern here. I'm just trying to see if there is any way we can have this without hurting SEO of the site.

Does using something like this will affect SEO?
```html
<head>
  <link rel="jslicense" href="/licenses"> 
</head>
```

Also, what if we add some content related to project license etc to that page, so that it is not just blank page?

metmarkosaric:
> I can understand your concern here. I'm just trying to see if there is any way we can have this without hurting SEO of the site.
> 
> Does using something like this will affect SEO?
> 
> ```
> <head>
>   <link data-jslicense="1" href="/licenses"> 
> </head>
> ```

something like that might work as i don't think it would affect anything with seo. we should probably also no index this page and maybe include link to this: https://plausible.io/blog/open-source-licenses

bhavin192:
I have updated the pages to use `<link>` instead of footer link and blank links. In addition to `<meta name="robots" content="noindex">`, should I add this page to `priv/static/robots.txt`?

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -11,4 +13,12 @@ defmodule PlausibleWeb.PageController do
       render(conn, "index.html")
     end
   end
+
+  def licenses_table(conn, _) do
+    render(conn, "licenses.html",
+      title: "License information | Plausible Analytics",
+      noindex: true,
```
ukutaht, line 20: I like this page being noindex but I think a cleaner solution is to use the `X-Robots-Tag` header instead of the `<meta>` tag. It can be a single line in this controller action and no changes to the template needed.

There's an example in our StatsController: https://github.com/plausible/analytics/blob/master/lib/plausible_web/controllers/stats_controller.ex#L22

lib/plausible_web/templates/error/error.html.eex
```diff
@@ -9,7 +9,7 @@
     <title><%= assigns[:title] || "Plausible · Web analytics" %></title>
     <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
     <script type="text/javascript" data-pref="<%= @conn.assigns[:current_user] && @conn.assigns[:current_user].theme %>" src="<%= Routes.static_path(@conn, "/js/applyTheme.js") %>"></script>
-
+    <link data-jslicense="1" href="/licenses">
```
ukutaht, line 12: Just to be super safe in terms of SEO I'd add `rel="license"` on these invisible links.

ukutaht (changes requested):

lib/plausible_web/controllers/page_controller.ex
bhavin192: Ah, yes. I was not aware of this header, it will be much cleaner with the header.

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -11,4 +13,13 @@ defmodule PlausibleWeb.PageController do
       render(conn, "index.html")
     end
   end
+
+  def licenses_table(conn, _) do
+    conn
+    |> put_resp_header("x-robots-tag", "noindex")
+    |> render("licenses.html",
+      title: "License information | Plausible Analytics",
+      layout: {PlausibleWeb.LayoutView, "focus.html"}
```
herminiotorres, line 22: What do you think to extract these pieces of code and put in a module attribute and called here? And use a name to describe your better intentions about these data. I presume these codes wouldn't change too much and keep cleaner, and the name of the module attribute will give a good context about it...

bhavin192: I'm not sure if I understood correctly, I'm new to Elixir and I have tried to follow the patters as of other modules in the code, for example the auth controller, https://github.com/plausible/analytics/blob/master/lib/plausible_web/controllers/auth_controller.ex
Will it be different pattern if we use module attributes to store the "licenses.html", title and the layout object?

bhavin192:
@ukutaht I have made the changes. Let me know when you folks plan to merge this, then I will rebase the PR for the merge.

Issue #213, opened 2020-07-09
Embeddable mode


cassidyjames:
>## Precheck
>
>Please note that this tracker is only for feature requests. Do not use the issue tracker for help or support. [Our docs](https://docs.plausible.io/) are a great place for most answers, but if you can’t find your answer there, you can [contact us](https://plausible.io/contact). Thanks!

👍

>## Prerequisites

- [x] I have searched open and closed issues to make sure that the feature has not yet been requested.

>## Feature request
>**Is your feature request related to a problem? Please describe.**

I am making a [native/web hybrid Plausible app](https://github.com/cassidyjames/plausible) (think: similar to Electron, but more native) for my own use on elementary OS and probably distribution on [Flathub](https://flathub.org) or [AppCenter](https://developer.elementary.io). I would like to have more of the app native than just having one big web view, so I can follow the native design patterns of elementary OS.

>**How would you like it to work**
>A clear and concise description of what you want to happen.

I have thought of three potential solutions:

1. An “embedded” flag, i.e. via URL query parameter, that tells Plausible to hide the header, footer, and other bits of navigation. The embedding app would be expected to implement that navigation in its own (native) way.

2. Individually embeddable widgets, so a native app could have smaller web frames that just show the specific graphs/widgets as needed. These could also be useful if a site wanted to share some social proof—similar to #119.

3. A full REST/JSON API of some sort. That would be a lot more work for everyone, but ultimately the most flexible. 

For my uses, I prefer solutions 1 or 2 the most.

>**Screenshots (If applicable)**
>If applicable, add screenshots to help explain the proposed solution.

Here's what I'm doing with my app:

![20200709_134939.jpg](https://user-images.githubusercontent.com/611168/87084192-0eeb5300-c1eb-11ea-8f1c-1ffc47b51a65.jpg)

I hide the top and bottom navigation manually by injecting CSS, and then show certain buttons in the native HeaderBar depending on the current page.

ArsalaBangash:
Hi! I'd love this feature as well to display on my org's website org.grey.software/stats

I'm in favor of the first implementation as a minimally viable solution at this point.

ArsalaBangash:
Hi @cassidyjames, I've emailed Plausible Analytics about this issue and CC'ed you. This would be a very useful feature for @grey-software because I'd like to display weekly analytics on https://grey.software/this-week

How did you manage to get the screenshot working? 

I get the following error: 

[...]

PR #772, opened 2021-03-01
Embeddable mode #213

aymanterra:


ukutaht:
I am testing this out now and will be trying to get it across the finish line this week. Will probably pop my questions and comments in here as I go.

First, I made a pretty big change to how shared links work so that it uses a separate cookie from all the other session stuff we do. This allows us to have a `SameSite=None; Secure` cookie for shared links while being more strict for normal authenticated sessions.

This PR requires some rework in the plugs and routes and such to align with the new cookie stuff. I am happy to continue the work on this. The second thing I would probably want to change is moving the `embeddable` field from the `site` record down to the `shared_link` record. I think that's a more natural scope for the field. Let me know if you see any issues with that.

I did have a question about `X-Frame-Origin` - why restrict it at all? Generally speaking I get it, we don't want anyone to embed our settings page for example, because it opens up a clickjacking attack vector. But what if we disabled `X-Frame-Option` on the embedded endpoints and kept it everywhere else? I can't think of any attacks it would open up and it would allow embeds to work cross origin as I understand it. I might be missing something but I'll test this approach out.

bradkane:
@ukutaht So glad you are (generally) pleased with the changes made.  I agree with you that @aymanterra did a great job!  I think your comments and suggestions are all great improvements.  These are areas where deep knowledge of the architecture and code can make our work even better and it is terrific that you are adding your limited time to this effort as well.  Let us know how we can help.

ukutaht:
Closed in favour of #812

PR #781, opened 2021-03-02
Fixes goal filters for glob goals

Vigasaurus:
### Changes

Fixes as per the issue introduced with #750 

### Tests

### Changelog

### Documentation

### DIFF
--- a/lib/plausible/stats/clickhouse.ex
+++ b/lib/plausible/stats/clickhouse.ex
@@ -1162,31 +1162,7 @@ defmodule Plausible.Stats.Clickhouse do
   end
 
   defp base_query_w_sessions(site, query) do
-    q = base_query_w_sessions_bare(site, query)
-
-    {goal_event, path} = event_name_for_goal(query)
-
-    q =
-      if goal_event do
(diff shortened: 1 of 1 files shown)


lib/plausible/stats/clickhouse.ex
```diff
@@ -1418,7 +1427,16 @@ defmodule Plausible.Stats.Clickhouse do
 
     q =
       if path do
-        from(e in q, where: e.pathname == ^path)
+        if String.match?(path, ~r/\*/) do
```
ukutaht, line 1430: Since the code is identical, I think it should be extracted into a separate function

Vigasaurus: Just did that 👍 I do honestly almost think that it consolidated a bit too much (but everything there was verbatim the same), `base_query_w_sessions` and `base_query` basically are just one-liners now - but I think them being named still has value. Thoughts?

ukutaht: Yeah I think they still have value to be named as they are. Maybe there are better names we can think for these base functions but the behaviour is good and I like the extracted function.

PR #786, opened 2021-03-03
Removes static assignment of tracker script variations

Vigasaurus:
### Changes

- Cuts down on the need for manual combinatoric generation for script variant names and aliases
- Changes manual process for adding script types to just appending string to an array (in two places)
- Assigns standard format for variant variable names (variant name, with `-` replaced with `_`)

*Notes:*
- The file name generation grows at `2^n`, so it could potentially be manageable without this
- The alias generation grows at `n!` which is certainly unmanageable manually
- From some quick and dirty tests, up to at least 9 variants are handled somewhat gracefully and in appropriate time with this system (this is 500+ tracker files, so hopefully if there are really more than 9 distinct variants of the script down the line, a better system can be created then) - and this process only happens on image creation

### Tests

### Changelog

### Documentation


Vigasaurus:
I've done pretty significant testing and have verified this to work how it should. (And didn't see much code cleanup opportunity without adding unnecessary complexity). 

I considered adding tests originally, but the only way I could see them being maintainable was if they checked for the primary files, and some random subset of variations - but that felt a bit whacky. Let me know if you have any ideas on tests/similar.

ukutaht (approved):

ukutaht:
Sorry for the long delay, looks good!

PR #850, opened 2021-03-18
Stop analytics tracking if running under Cypress tests

roughike:
I'm using [Cypress](https://www.cypress.io/) to run end-to-end tests on my website locally, and I'm using Plausible Analytics on that website.

At the time, the website in question wasn't advertised anywhere, it was not indexed by Google, and it didn't have any visitors.
After introducing some Cypress tests, my visitor numbers spiked in a way that I just don't see realistically happening.

Putting one and one together, I think the cause was because Plausible tracked events when being run by Cypress. The existing methods of checking if we're running inside an end-to-end test didn't work because Cypress takes a different approach to running end-to-end tests and doesn't use webdrivers.

### Changes
- Detect if we're running under Cypress. If we are, stop tracking. [See here for the official Cypress documentation](https://docs.cypress.io/faq/questions/using-cypress-faq.html#Is-there-any-way-to-detect-if-my-app-is-running-under-Cypress).

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Capitalized date/time selection keybinds not working plausible/analytics#709
 - Invisible text on Google Search Console settings page in dark mode plausible/analytics#759
+- Disable analytics tracking when running Cypress tests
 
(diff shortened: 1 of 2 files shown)


ukutaht:
Makes sense. Thanks!

PR #907, opened 2021-04-02
Disable self-tracking with self hosting

ishitatsuyuki:
### Changes

Disables self-tracking when self hosted.

In my case, I was staring at mysterious entries for my website. It showed up because I hosted Plausible on a subdomain of my primary site (plausible.example.com) and it accidentally matched my site. This PR disables this, as probably no one self-hosting would want it.

### Tests

### Changelog

### Documentation


ukutaht:
Makes sense. Thanks!

PR #951, opened 2021-04-22
Fix weekly report time range

ukutaht:
### Changes

Bug: weekly reports are sent out on Monday. The time range that stats are calculated for are using the `Query.from(%{"period" => "7d"})` which is problematic because:
* The resulting date range is last Tuesday -> this Monday
* Stats for 'this Monday' are incomplete, so sending a report is for that is wrong

The correct date range is 'last monday' -> 'last sunday'. The stats for that range are complete and that's what most people consider a 'week'.

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [1.3] - Unreleased
+## Unreleased
+
+### Fixed
+- Fix weekly report time range plausible/analytics#951
+
+
+## [1.3] - 2021-04-14
 
 ### Added
 - Stats API [currently in beta] plausible/analytics#679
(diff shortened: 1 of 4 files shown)


anmipo:
> The correct date range is 'last monday' -> 'last sunday'. The stats for that range are complete and that's what most people consider a 'week'.

I don't want to throw a wrench, but "most people" is perhaps too optimistic:

> the United States, Canada, Brazil, Japan and other countries consider Sunday as the first day of the week

> the week begins with Saturday in much of the Middle East

[Source: [Wikipedia](https://en.wikipedia.org/wiki/Week)]

ukutaht:
Sure, in the future we can offer customizations so people can choose.

Meanwhile we need to choose something as the default. I think the secular ISO-8601 standard (week starting with Monday) is the most reasonable and inclusive. Other standards come from certain religions and I really don't want to get in the business of choosing whether we should use the Bible or the Quran as the source of truth :)

PR #952, opened 2021-04-23
Additional metrics for timeseries API call and CSV export

ukutaht:
### Changes

Adds a new option `metrics` for the `/api/v1/stats/timeseries` API call. Non-breaking change. The `metrics` parameter defaults to `visitors` but it can be specified as `visitors,pageviews,visit_duration,bounce_rate` or any combination of metrics to get more timeseries information.

CSV export will now include all 4 metrics by default.
CSV export uses non-sampled queries to provide the most accurate data.

### Tests

### Changelog

### Documentation
- [x] https://github.com/plausible/docs/pull/76


gaodeng:
@ukutaht  
Could you please add another metric - events

PR #963, opened 2021-04-25
Group pages by hostname and pathname

baptisteArno:
### Changes
Display pages by taking into account the hostname.

The label is formatted by following these rules:
- If page view is from the exact same configured domain: show only the pathname
- If page view is from a subdomain of the configured domain: show the pathname with subdomain as a tag label next to it (displayed here with parentheses)
- If page view is from a completely different domain: display the entire domain + pathname

CleanShot 2021-04-25 at 11 42 10@2x

Filtering has also been adapted.

This is a result of the discussion in #478

### Tests
- [x] Automated tests have been adapted

### Changelog
- [x] Entry has been added to the changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - New parameter `metrics` for the `/api/v1/stats/timeseries` endpoint plausible/analytics#952
 - CSV export now includes pageviews, bounce rate and visit duration in addition to visitors plausible/analytics#952
+- Page views are now grouped by hostname and pathname on the dashboard.
 
 ### Fixed
 - Fix weekly report time range plausible/analytics#951
--- a/assets/js/dashboard/filters.js
+++ b/assets/js/dashboard/filters.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { withRouter } from 'react-router-dom'
-import { countFilters, navigateToQuery, removeQueryParam } from './query'
+import { countFilters, navigateToQuery, parsePageAttributes, removeQueryParam } from './query'
 import Datamap from 'datamaps'
 import Transition from "../transition.js";
 
@@ -148,7 +148,8 @@ class Filters extends React.Component {
       return <span className="inline-block max-w-2xs md:max-w-xs truncate">Country: <b>{selectedCountry.properties.name}</b></span>
     }
     if (key === "page") {
-      return <span className="inline-block max-w-2xs md:max-w-xs truncate">Page: <b>{value}</b></span>
+      const {pageLabel, subdomain} = parsePageAttributes({domain: this.props.site.domain, pageName: value})
+      return <span className="inline-block max-w-2xs md:max-w-xs truncate">Page: <b>{pageLabel}</b> {subdomain && <span className="bg-gray-600 px-1 rounded-md">{subdomain}</span>}</span>
     }
     if (key === "entry_page") {
       return <span className="inline-block max-w-2xs md:max-w-xs truncate">Entry Page: <b>{value}</b></span>
--- a/assets/js/dashboard/filters.js
+++ b/assets/js/dashboard/filters.js
@@ -148,8 +148,8 @@ class Filters extends React.Component {
       return <span className="inline-block max-w-2xs md:max-w-xs truncate">Country: <b>{selectedCountry.properties.name}</b></span>
     }
     if (key === "page") {
-      const {pageLabel, subdomain} = parsePageAttributes({domain: this.props.site.domain, pageName: value})
-      return <span className="inline-block max-w-2xs md:max-w-xs truncate">Page: <b>{pageLabel}</b> {subdomain && <span className="bg-gray-600 px-1 rounded-md">{subdomain}</span>}</span>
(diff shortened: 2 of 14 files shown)


Vigasaurus:
I think this is a nice implementation as far as the UI is concerned, but I really think that

1. Not having the ability to turn the grouping on and off on the fly is problematic (and doing this sort of grouping only in react will likely be a large slowdown) - see the response to the old PR - https://github.com/plausible/analytics/pull/478#issuecomment-747938667

2. I really don't think that having 2 completely unique view styles for a subdomain and for a different website is that great - it really breaks down the alignment of the report and makes it much harder to parse which website everything is from if you have to keep bouncing back and forth. Maybe if the domain could be put in a similar label as opposed to prepending it to the pathname - that way it's consistent, and out of the way while reading.

baptisteArno:
> 1. Not having the ability to turn the grouping on and off on the fly is problematic (and doing this sort of grouping only in react will likely be a large slowdown) - see the response to the old PR - [#478 (comment)](https://github.com/plausible/analytics/pull/478#issuecomment-747938667)

Who wants to have `blog.app.com/` and `app.com/` and `othersite.com/` count as the same page view event? This shouldn't be an option to turn on or off IMO! Can you provide legitimate examples?

> 2. Maybe if the domain could be put in a similar label as opposed to prepending it to the pathname - that way it's consistent, and out of the way while reading.

I just made the changes, it looks better indeed!
CleanShot 2021-04-25 at 21 41 17@2x

Vigasaurus:
> This shouldn't be an option to turn on or off IMO! Can you provide legitimate examples?

It depends on the version of the site - some people will have different domains for differing languages/the same app (google.ru google.co.uk) whereas some may have a landing page and a differring app with combined stats (discord.com discordapp.com) - in the former example, a user may want to have their entire site stats in one go, then later on decide if they'd like to break down by hostname or not.

Vigasaurus:
Also as an addition here - maybe the same hostname with a different TLD could also be shown in your implementation the same way, where the label can show `.com` and `.app` for example (since we don't need to repeat the basic hostname)

Other than that, I do definitely think the UI looks super clean like this.

ukutaht:
@baptisteArno Thanks, looks very nice!

This is a huge change to the dashboard and many things need to be considered. I really want these problems to be solved and I'm happy to merge a PR that satisfies all the requirements.

**1. Grouping should be optional**
I agree with @Vigasaurus that grouping should be optional and there should be a switch to turn it on and off. The first requests we got for this feature were people who run their website on separate subdomains for different languages:

`mydomain.com`
`es.mydomain.com`
`it.mydomain.com`
`de.mydomain.com`
etc.

These users asked us explicitly to provide an option. Normally they want to see everything combined but sometimes they want to have them broken down by separate hostnames.

**2. Subfolder installations**
I tested this PR with a subfolder installation and it doesn't handle it very well. A pretty standard example would be github pages where one user will configure their domain to be something like this:
`ukutaht.github.io/site1`

Now when they get a pageview on `https://ukutaht.github.io/site1/page1` for example, it should probably recognize that it's a pageview on the root domain. Currently the `ukutaht.github.io` domain will be shown in a label as if the pageview happened on a different domain.

**3. Breakdown by hostname**
There should be a way to see traffic by hostname. Thinking of example nr 1, I want to see a breakdown of traffic by different hostnames:


`mydomain.com        -> 1000 visitors`
`es.mydomain.com   -> 99 visitors`
`it.mydomain.com     -> 98 visitors`
etc

**4. Data representation**
Just something I noticed about the way filters work - now the `page` API response and filter always seems to include the hostname along with the pathname. I think these two should be treated as independent pieces of data, partly because of point 1 from this comment. So if you want to filter by hostname and pathname, it should add two separate filters: `Host: blog.mydomain.com; Page: /page-url` or something like that.

I *really* like the UI for this PR. It looks really nice. Can you think of ways to keep this nice UI while solving all the problems?

I'm leaning towards a solution we've discussed before that doesn't look as nice but addresses the issues I brought up here: https://github.com/plausible/analytics/pull/478

bmenant:
I can’t agree more with @ukutaht and @Vigasaurus 

IMHO, the suggestions [made by @ukutaht](https://github.com/plausible/analytics/pull/478#issuecomment-805816732) remains the best answer in term of usability and functionality.

If it helps, I can detail our own use cases: an _private_ platform where we’ve got hundreds of subdomains and domains plugged onto a single service platform. Internally (for the organization or company say), each domain or subdomain relates to a project, campaign or operation (whatever you call them). The platform is private, but the campaigns are public. Each campaign might have its own theme and features... but, for the most part, mere mechanisms and structure remain the same from a campaign to another. 
If you need an example of such _private_ platform, I think you can have a look at Champaign (campaigning platform application used by SumOfUs).

At a platform scale, there are times we want everything to be grouped by domains (surge of visits on the “I’ve got a problem” page, for example, which would hint at a potential issue on our platform). There’re times we want to compare how different domains (campaigns) perform.

At a campaign scale, we usually want to filter by domain.

Hope it helps and thanks for the PR anyway! :+1:

baptisteArno:
I agree with all you suggested. I will work on it eventually. :)

ukutaht:
Great @baptisteArno. I'll close this PR for now. Looking forward to another one if you get a chance to work on it :)

PR #1051, opened 2021-05-18
Adds time on page as a top stat when a page filter is active

Vigasaurus:
### Changes

As per https://github.com/plausible/analytics/pull/1007#issuecomment-843234020

*Notes*:
- I used the same setup as the bounce+time on page async, but with the current and previous query for the rate of change calculation. 
- In the above setup, I reduced the timeout to 5 seconds in this version, since I figure waiting longer for the top graph just for the one stat is probably not the best UX, so if it takes more than 5 seconds, it'll return the entire graph and all top stats except the time on page (it's just not present). 

### Tests

### Changelog
It's bundled with regular time on page, nbd

### Documentation

### Screenshots
image

### DIFF
--- a/lib/plausible/stats/clickhouse.ex
+++ b/lib/plausible/stats/clickhouse.ex
@@ -691,7 +691,7 @@ defmodule Plausible.Stats.Clickhouse do
     |> Enum.into(%{})
   end
 
-  defp page_times_by_page_url(site, query, page_list) do
+  def page_times_by_page_url(site, query, page_list) do
     q =
       from(
         e in base_query_w_sessions(site, %Query{
@@ -725,7 +725,10 @@ defmodule Plausible.Stats.Clickhouse do
         FROM (#{base_query_raw}))
       WHERE s=s2 AND p IN tuple(?)
       GROUP BY p,p2,s)
-    GROUP BY p" |> ClickhouseRepo.query(base_query_raw_params ++ [page_list ++ ["/"]])
+    GROUP BY p"
+    |> ClickhouseRepo.query(
+      base_query_raw_params ++ [(Enum.count(page_list) > 0 && page_list) || ["/"]]
+    )
   end
 
   defp add_percentages(stat_list) do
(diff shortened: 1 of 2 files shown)


lib/plausible/stats/clickhouse.ex
```diff
@@ -725,7 +725,10 @@ defmodule Plausible.Stats.Clickhouse do
         FROM (#{base_query_raw}))
       WHERE s=s2 AND p IN tuple(?)
       GROUP BY p,p2,s)
-    GROUP BY p" |> ClickhouseRepo.query(base_query_raw_params ++ [page_list ++ ["/"]])
+    GROUP BY p"
+    |> ClickhouseRepo.query(
+      base_query_raw_params ++ [(Enum.count(page_list) > 0 && page_list) || ["/"]]
```
ukutaht, line 730: Why is the `|| ["/"]` needed?

ukutaht (approved):

lib/plausible/stats/clickhouse.ex
Vigasaurus: CH threw an error if you pass an empty array into `tuple()` - so I just chose the page path that is always safely present in all sites just in case there are no pages in the time period. This doesn't apply in this usage of the function since we always pass in the page filter, but in the other use case, there's a chance that the page array is empty if there are no pageviews.

ukutaht: Ah I see.

metmarkosaric:
A question [from Twitter](https://twitter.com/juergenroehm/status/1394965407602204674): "Is that an average value? Does it take into account the visibility of a page? What happens to “Time on page”, if a user has a page open in an inactive browser tab, e.g. for two days?"

ukutaht:
> Is that an average value?
Yes

> Does it take into account the visibility of a page?
No

> What happens to “Time on page”, if a user has a page open in an inactive browser tab, e.g. for two days?
Time on page is calculated as the difference between two pageviews within a single session. If the user is inactive for 30 minutes, their session will expire, the page they were on will be the exit page and therefore time on page will not be counted at all.

metmarkosaric:
thanks!

PR #1052, opened 2021-05-18
Postgres socket

Ma27:
### Changes

This introduces the environment variables 

### Tests

### Changelog

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated. Todo, but I'd like to get some feedback first before investing more time :)

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -107,7 +107,15 @@ config :plausible, PlausibleWeb.Endpoint,
   http: [port: port],
   secret_key_base: secret_key_base
 
-config :plausible, Plausible.Repo, url: db_url
+case System.get_env("DATABASE_SOCKET_DIR", "") do
+  "" ->
+    config :plausible, Plausible.Repo, url: db_url
+
+  x ->
+    config :plausible, Plausible.Repo,
+      socket_dir: x,
+      database: System.get_env("DATABASE_NAME")
+end
 
 config :sentry,
   dsn: sentry_dsn,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 - New parameter `metrics` for the `/api/v1/stats/timeseries` endpoint plausible/analytics#952
 - CSV export now includes pageviews, bounce rate and visit duration in addition to visitors plausible/analytics#952
 - Send stats to multiple dashboards by configuring a comma-separated list of domains plausible/analytics#968
+- To authenticate against a local postgresql via socket authentication, the environment-variables
+  `DATABASE_SOCKET_DIR` & `DATABASE_NAME` were added.
 
 ### Fixed
 - Fix weekly report time range plausible/analytics#951


config/runtime.exs
```diff
@@ -107,7 +107,15 @@ config :plausible, PlausibleWeb.Endpoint,
   http: [port: port],
   secret_key_base: secret_key_base
 
-config :plausible, Plausible.Repo, url: db_url
+case System.get_env("DATABASE_SOCKET_DIR", "") do
```
happysalada, line 110: Just a question here, feel free to ignore if you don't have time.
Do you need the case clause here? I think I remember from a past deployment that postgrex is smart enough to figure out how to connect to postgres. i.e. if there is a socket_dir var it will use it and if not it will check the db url var. I could be wrong here. However testing might simplify the code.

happysalada:
Also not related, but are you working on a nixos implementation? Perhaps we can coordinate our efforts.

config/runtime.exs
ukutaht: If that's true I'd prefer to remove the case clause. However, we need to make sure what the parameter precedence actually is.

ukutaht (commented):
Looks good to me. Needs to be documented and I'm also interested in what @happysalada suggested to remove the case clause from our code if the parameter precedence in Ecto is clear.

Ma27:
> Also not related, but are you working on a nixos implementation? Perhaps we can coordinate our efforts.

Yep, you caught me! I'm around 85% through, I'm mostly missing an actually working module (+test of course), an update script and a few cleanups :) 

> Needs to be documented and I'm also interested in what @happysalada suggested to remove the case clause from our code if the parameter precedence in Ecto is clear

Will try to do that later today! :)

happysalada:
@Ma27 amazing! Haha, the only reason I guessed, was the fact that I was to do the exact same PR :-)
if you are looking for early feedback or want to discuss anything, feel free to share.
I am currently pushing for better elixir release mechanism on nixos. We are right now integrating a mix2nix tool.
I won't spam this PR with more nix stuff, but if you need testing feedback or any help, know you have an ally here!
(on that note I noticed the state of clickhouse on nixos is a little lacking. I thought that could be a small quickly mergeable PR if any adjustments need to be made for a potential integration with plausible).

config/runtime.exs
Ma27: I tried that out by replacing the `case`-clause here with this:

```
config :plausible, Plausible.Repo,
  url: db_url,
  socket_dir: System.get_env("DATABASE_SOCKET_DIR", ""),
  database: System.get_env("DATABASE_NAME", "")
```

When only setting `DATABASE_SOCKET_DIR` and `DATABASE_NAME`, plausible fails to connect to the DB because of a failed peer authentication (it tries to pick up `postgres` which seems to be the default from `DATABASE_URL`). From my understanding this means that the URL param can interfere with the two vars I've added, so the `case` may be a better choice.

happysalada: I looked for the connection logic
https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/protocol.ex#L110
it seems to be prioritizing the socket_dir then checking hostname.
The documentation also mentions that this takes precendence over hostname
https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex.ex#L85

Perhaps this is because of the url. 
After checking my existing running app, I don't have the url param set up but I have it brokent down into it's components. (hostname, database, username, password)

I guess here the choice, is either use a db url or use the separate params. The separate params implies passing several environment variables instead of 1.

I guess the case clause is easier at this point.

ukutaht:
Thanks @Ma27 and @happysalada !

Issue #1017, opened 2021-05-11
Lock down Postgres and Clickhouse Docker image versions in the Makefile for a consistent development environment


MaybeThisIsRu:
## Bug report
**Describe the bug**
When I followed the contributing guidelines, and brought up the Postgres and Clickhouse containers as instructed in step one, I noticed the images are not locked to any version at all. This might result in unexpected issues locally as production might be running Postgres 12.x whereas I might have Postgres 13.x locally (for an example).

**Expected behavior**
Postgres and Clickhouse would be locked to at least the same major version (per SemVer) as the one on production for bringing the local and production environment closer.

**Screenshots (If applicable)**
On running `make postgres`:

![image](https://user-images.githubusercontent.com/3974019/117815703-5d420600-b283-11eb-82c2-14e850f6d708.png)

On running `make clickhouse`:

![image](https://user-images.githubusercontent.com/3974019/117815803-76e34d80-b283-11eb-8690-fb247127341d.png)

ukutaht:
Thanks @hirusi. I've pinned the Postgres and Clickhouse versions in our hosting repo for the same reason. We should probably use the same in development: https://github.com/plausible/hosting/blob/master/docker-compose.yml

hirusi:
> Thanks @hirusi. I've pinned the Postgres and Clickhouse versions in our hosting repo for the same reason. We should probably use the same in development: https://github.com/plausible/hosting/blob/master/docker-compose.yml

Great! Thanks for pointing that out - I've opened a PR to have specific versions locally.

There's one more improvement we could make -- that is to use volumes for the postgres container. At the moment, I have to sign up fresh on Plausible each time I boot the database container. If you're happy to have that change, I'll open a PR later on.

ukutaht:
Yes that sounds good as well :)

PR #1057, opened 2021-05-19
Improvements for local develoment environment

hirusi:
### Changes

Per our discussion on #1017, this PR makes one important change:

* The Postgres container is updated to use a named volume for its data. This lower friction for repeat contributions where one would otherwise sign up and activate their accounts again and again each time.

Additionally, it introduces two new `Makefile` commands and documents them:

* This change adds two new commands to gracefully stop and remove the Postgres and Clickhouse docker containers. To do so, it also gives them a recognizable name i.e. `plausible_db` and `plausible_clickhouse`.

### Tests

### Changelog

### Documentation


ukutaht:
Looks great! Thanks

Issue #265, opened 2020-08-03
search for specific page + stats on this page


devrand:
There is a strong need to check stats for pages that are not in main listing. 

**How would you like it to work**
Search with text field by URL or part of URL, get stats on this page

metmarkosaric:
Thanks @devrand! Is this because you cannot see the specific page in the list when you click on the "More" button on the "Top Pages" report or because you'd like an easier way to find a specific page in that list?

devrand:
It's both, but first user case  is more important - I have a site with thousands of pages

Spone:
Same need here, I added a :+1:

davidsteeb:
Same here.

PR #1067, opened 2021-05-21
Adds manual filters on FE, capability for exclusion filters (and the globbing we're used to) for path-based filters

Vigasaurus:
### Changes

- Adds button in the UI to add filters manually, to help prevent users from having to find the clickable version to apply the filter #265
- Adds availability of the globbing syntax in all 3 of the path-based filters (page, entry_page, exit_page) to help users not need to use goals to do this sort of filtering #774
- Adds ability for users to edit existing filters without needing to delete + re-add them #943
- Adds ability for the same 3 path-based filters to have their filter be an exclusion rather than an inclusion https://github.com/plausible/analytics/discussions/265#discussioncomment-434270

*Notes*:
- ~~This isn't fully done yet, as per the commit descriptions I have a few more things to add (code format, tests) - but the user-facing portion should be done, and I'd love any feedback there~~ Yes it is
- I planned out the system for exclusions broadly enough that it could be extended to more than just the path-based filters relatively easily down the line, but for now I figured the path-based ones are the most desirable
- As discussed below - the UI for explaining that the filter is an exclusion vs an inclusion is up for discussion - I've currently left the `(not)` UI, but we can explore other ideas that make it clear, but not too clunky. I thought a potential would be to use the `=` and `≠`  symbols directly instead of the current setup with colons. 

### Tests

### Changelog

### Documentation

### Screenshots

image
*I'm not necessarily super happy with this UI for the excluded filters, but I couldn't quite come up with anything better without it getting unwieldy*

Add Filter UI
image

Edit Filter UI
image

Add non-excludable filter UI
image

---

Inclusion Filter
image

Exclusion Filter
image


metmarkosaric:
Nice! Some quick ideas on the UI from Search Console in which I use this type of functionality often:

Rather than displaying (not), perhaps use the minus sign?

Screenshot from 2021-05-21 11-04-27

To edit the filter you could perhaps just click on the existing filter that you want to edit, no need to have any visual signs. When you click on the existing filter for this thing pop ups in SC:

Screenshot from 2021-05-21 11-07-00

Vigasaurus:
> Rather than displaying (not), perhaps use the minus sign?

Yeah I thought about this, but with the primary use of this starting with a `/` I wasn't sure that the minus stuck out enough - maybe with a color change or bolded though. Like, `-/` barely looked like a minus sign to me, but we can experiment. 

> To edit the filter you could perhaps just click on the existing filter that you want to edit, no need to have any visual signs.

Yeah actually that may be kinda nice (for the non-dropdown display at least) - to take the cue from SC - where on hover over anything but the `X` the icon changes to a pencil and let's you edit on click there. I think the dropdown-view could potentially stay more verbose with the always-on edit icon since that UI is more mobile-focused anyways, and hovering isn't really a thing.

metmarkosaric:
> Yeah actually that may be kinda nice (for the non-dropdown display at least) - to take the cue from SC - where on hover over anything but the `X` the icon changes to a pencil and let's you edit on click there. I think the dropdown-view could potentially stay more verbose with the always-on edit icon since that UI is more mobile-focused anyways, and hovering isn't really a thing.

yeah, i like this!

assets/css/app.css
```diff
@@ -310,3 +310,11 @@ iframe[hidden] {
 .pagination-link[disabled] {
   @apply cursor-default bg-gray-100 pointer-events-none;
 }
+
+.filterListText:hover ~ .filterListEdit {
```
ukutaht, line 314: Use `kebab-case` instead of `snakeCase` for css classes.

assets/js/dashboard/filters.js
```diff
@@ -133,28 +136,28 @@ class Filters extends React.Component {
...
-      return <span className="inline-block max-w-2xs md:max-w-xs truncate">{browserName}.Version: <b>{value}</b></span>
+      return <span className="inline-block max-w-2xs md:max-w-xs truncate">{browserName} Version: <b>{value}</b></span>
     }
     if (key === "os") {
       return <span className="inline-block max-w-2xs md:max-w-xs truncate">Operating System: <b>{value}</b></span>
     }
     if (key === "os_version") {
       const osName = query.filters["os"] ? query.filters["os"] : 'OS'
-      return <span className="inline-block max-w-2xs md:max-w-xs truncate">{osName}.Version: <b>{value}</b></span>
+      return <span className="inline-block max-w-2xs md:max-w-xs truncate">{osName} Version: <b>{value}</b></span>
```
ukutaht, line 146: I prefer the natural language ones (`is` vs `:`). I think they're especially nice for negations:

Screenshot from 2021-05-25 15-24-12@2x

What do you think about changing everything to say `is` or `is not`? It would also be great for the future if we want to do filters like `starts with` or `includes` which are difficult to explain with symbols instead of words to regular (non-developer) folks.

assets/js/dashboard/filters.js
```diff
@@ -171,18 +174,50 @@ class Filters extends React.Component {
...
+        {['goal', 'props'].includes(key) ? (
+          <span className="flex w-full h-full items-center py-2 pl-3">
+            {this.filterText(key, value, query)}
+          </span>
+        ) : (
+          <>
+            <Link title={`Edit filter: ${formattedFilters[key]}`} className="filterListText flex w-full h-full items-center py-2 pl-3" to={{ pathname: `/${encodeURIComponent(this.props.site.domain)}/filter/${key}`, search: window.location.search }}>
+              {this.filterText(key, value, query)}
+            </Link>
+            <span className="filterListEdit hidden h-full w-full px-2 cursor-pointer text-indigo-700 dark:text-indigo-500 items-center">
```
ukutaht, line 213: The only weird bit I found with this is:
1. Hover on a filter
2. See that the icon changes to the edit icon
3. Now if I actually do want to edit, my natural reaction is to move the mouse to the edit icon and click that
4. But now the edit icon changes to X and I'm removing the filter

I'm just putting it out there that this could be a bit confusing at times. It's better than any alternative I can think of at the moment. If you have ideas, let me know what they are. Otherwise, happy to keep it as is.

assets/js/dashboard/stats/modals/filter.js
```diff
@@ -0,0 +1,137 @@
+import React from "react";
```
ukutaht, line 1: I was skeptical about putting this in a modal (as opposed to a dropdown) but I really really like it. Very nice UX.

ukutaht: Future improvements that I can think of:
* For screen size, provide the only 4 valid options in a select box.
* For country codes, provide incremental (typeahead) search for the list of possible countries.
* For other fields, provide incremental (typeahead) search based on the actual recorded values in the stats database

I don't think these are in scope for this PR but I'm just writing these down for future reference.

lib/plausible/stats/clickhouse.ex
```diff
@@ -1561,4 +1510,91 @@ defmodule Plausible.Stats.Clickhouse do
       q
     end
   end
+
+  defp check_negated_filter(filter) do
+    negated = String.at(filter, 0) == "!"
```
ukutaht, line 1515: The way negating is done here is fine for the moment I think. I have a task in my TODO list to make the filtering in the UI work underneath with the same sort of expressions as the API: https://plausible.io/docs/stats-api#filtering. Basically the UI would stay the same but the serialization would change.

In that system I think a negated filter would get its own symbol `!=` instead of `==`.

lib/plausible/stats/clickhouse.ex
```diff
@@ -1561,4 +1510,91 @@ defmodule Plausible.Stats.Clickhouse do
...
+    {negated, updated_filter}
+  end
+
+  defp convert_path_regex(path) do
+    contains_regex = String.match?(path, ~r/\*/)
+
+    regex =
+      "^#{path}\/?$"
+      |> String.replace(~r/\*\*/, ".*")
+      |> String.replace(~r/(?<!\.)\*/, "[^/]*")
```
ukutaht, line 1527: Again, not in scope for this PR but something I've been thinking about - the glob expressions might be a bit too advanced for regular users. Programmers know glob but most of our users don't. Maybe a simple `*` to mean wildcard (e.g.  `.*` in regex) would suffice. And then we could provide a different completely regex-based system for power use.

assets/js/dashboard/filters.js
```diff
@@ -198,11 +233,15 @@ class Filters extends React.Component {
 
   renderDropDownContent() {
     const { viewport } = this.state;
-    const { history, query } = this.props;
+    const { history, query, site } = this.props;
 
     return (
       <div className="absolute mt-2 rounded shadow-md z-10" style={{ width: viewport <= 768 ? '320px' : '350px', right: '-5px' }} ref={node => this.dropDownNode = node}>
         <div className="rounded bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 font-medium text-gray-800 dark:text-gray-200 flex flex-col">
+          <Link to={`/${encodeURIComponent(site.domain)}/filter${window.location.search}`} className="group border-b flex border-gray-200 dark:border-gray-500 px-4 sm:py-2 py-3 md:text-sm leading-tight hover:text-indigo-700 dark:hover:text-indigo-500 hover:cursor-pointer">
```
ukutaht, line 241: To me, this button is a bit too prominent and attention-grabbing. While filtering is an important feature, it feels like it's visually grabbing attention away from the hero of the dashboard - the stats themselves. Could you try making it a bit more subtle?

I can also give it a go if you run out of ideas :)

ukutaht (changes requested):
Looks great! I think this will be super useful for many users.

A couple things I'd like to discuss before approving for merge:
1. Using `is` and `is not` instead of `:` and `(not):`
2. The button style - I'd prefer something more subtle

I like the choices you made here. The modal feels good and it's a good start to build on top of.

assets/js/dashboard/filters.js
Vigasaurus: Yeah that's a great case of best of both worlds in length and clarity.

assets/js/dashboard/filters.js
Vigasaurus: Yeah the alternative was the original way I had set it up where if you went back and hovered over the edit icon it stayed, which was theoretically fine, but it cause a decent chunk of UX issues when a user would hover over the filter without passing the X, and then would need to explicitly leave the filter, and come back to remove it, since there would be no way to reveal the X. 

I think this method is nice, but definitely we can see what user feedback looks like, I'm honestly not sure.

assets/js/dashboard/stats/modals/filter.js
Vigasaurus: Yeah these will all be great things, even stuff like auto-complete and suggesting recently used things, and otherwise suggested "quick select" filters could be nice options. 

Should make for a nice and powerful feature as it grows down the line.

lib/plausible/stats/clickhouse.ex
Vigasaurus: Yeah that could make sense and simplify things. My only fear is that now people have had goals and exclusions with this format for a while, changing that up is effectively like a major API deprecation. I guess it'd just be a UI and user teaching thing though

assets/js/dashboard/filters.js
Vigasaurus: I'll take a stab at it 👍 - might do something where the bg is the same color as the date picker bg, and the text is the similar indigo+white, much like the datepicker drop-down, will see what it looks like though.

assets/js/dashboard/filters.js
Vigasaurus: One thing here though, I made this setup before I added the delete button in the edit view - maybe it's less necessary now that even if a user accidentally ends up in the edit view, they can still remove the filter easily?

assets/css/app.css
Vigasaurus: I always forget, heh, too used to CSS-in-JS, I guess 🙈

metmarkosaric:
One idea to make it less prominent would be to use a transparent search or filter icon. This then expands into the full top row when clicked. 

Then it would be cool to allow the user to start typing in the custom filter search box which allows the user to search the whole dashboard. No need to choose a metric category. 

If you type Germany we know that it is a country that you're looking for. And if the query you search for has items within several different reports/categories we can include a little tag next to the result so you can select which you want. Say I search for "my-blog-post", I can select between Top Pages, Entry Pages, Exit Pages. 

Kind of similar to how search works in Gnome on Linux or Spotlight on Mac.

Vigasaurus:
Yeah that sort of wizard approach could be nice down the line when the auto-complete setup etc is in place, and guesses can be very accurate, but I'm not sure that it should be the *only* way of doing it, unless it stays nice and consistent for users. I'll probably just try a few UI options for now though.

Vigasaurus:
Updated filter text:
image

Here's a version I was able to come up with that I think is still nice and noticeable, but doesn't jut out as much and fits the rest of the theme of that top bar (I also made all of these top elements now be somewhat consistent in their colors and hover-behaviors). 
image

One other iteration I had potentially, to just make the button a bit smaller. On smaller viewports we already make this change to just `+ Filter` instead of `+ Add Filter` - but I figured it could potentially do well to make the button less aggressive on the full version. 
image

assets/js/dashboard/filters.js
ukutaht: I can also see it being annoying if you have to make 2 clicks to remove a filter. I'd say let's keep it as is and see if we get any feedback about it

ukutaht:
Looks good @Vigasaurus! I'll put this on our staging server for further testing

metmarkosaric:
Thanks @Vigasaurus! I've been playing with this for a bit. I think autocomplete for filter values is necessary before we can release it.

It is a bit complicated to search now. If I type country name such as Denmark it gives me all 0's on the dashboard (it wants me to type a country code DNK instead of the country name). If I type google it gives me all 0's (it wants me to type Google). It should not be possible to filter by things that don't exist. Would be better if it autopredicts/autocompletes Denmark to DNK and google to Google etc.

It also might be better to make every single property value into a select-search box like this: https://react-select-search.com/?path=/story/single-select--search

Also would be nice to add some info icon to explain that you can do "contains" search for pages by using things such as **blog**. Right now if I search for blog it gives me all 0's. I need to type /blog to get the blog home page and **blog** to get all pages that include blog.

Is it possible to have the "Page" filter selected by default so you don't need to go "Select a filter" first? I suspect Page will be the most used one so by choosing it as default would save two clicks every time. People can still click on it to choose a different filter.

Vigasaurus:
Yeah I can work on auto complete for there, definitely. I didn't actually realize the country codes would work like that, I had only tested it with the US, which if you typed in `United States of America` it accepted fine. 

As for your blog, you just need to precede the path with a `/`, all paths must start with that - so you could do `/blog**`

But either way, I'll work on getting some auto complete and some suggestions for filtering.

metmarkosaric:
Is there a way to merge source and referrers into one somehow then? So if I just write the name of the site or the domain name that it gives me options? so i can write google or Google or https://google.com or google.com to see Google traffic

If not then we just have to explain better what the difference is and how to properly use it. There's a lot of white space below the search box now so we can use that to give tips

Also I notice I cannot filter by goal? Is this difficult to add?

Vigasaurus:
Yeah I think the referrer/source thing will definitely just need a bit of documentation+user education - most other changes would end up being over engineered probably. 

As for goals - I could add those back in with the auto complete setup I suppose, I had just excluded that and props originally because they would have a similar issue to referrer where people would need to be very specific/we'd have to make guesses as to if they want a page goal or a custom goal, probably adding confusion.

ukutaht:
> Is there a way to merge source and referrers into one somehow then? So if I just write the name of the site or the domain name that it gives me options? so i can write google or Google or https://google.com or google.com to see Google traffic

The source `Google` includes traffic from `https://google.com` but also `https://google.co.uk` and `https://google.com.au` etc. So showing traffic from `https://google.com` is not the same thing as traffic from `Google` - one is a subset of the other. This is why they're separate properties in the database and filters.

I think this is a big benefit of Plausible over many other tools. Instead of showing `google.com`, `google.co.uk` and `google.com.au` as separate referrer domains, we categorize them into a single _Source_ and show it as one entry named `Google`. Same for `www.reddit.com` and `old.reddit.com` being categorized into `Reddit`, for example.

My initial reaction was 'no way we can merge these fields into one' but we might just be using/thinking in different terms. The Source and Referrer must stay separate fields in the database because `https://google.com` is **not**  equivalent to the source `Google`.

However, the *filters* for these could in theory be merged into one and I think it could be very nice. I'm imagining a single search box for `Traffic source` or something like that. And if I start typing `google`, the suggestions will be:

```
Source: Google
Referrer URL: https://google.com/search
Referrer URL: https://google.co.uk/search
Referrer URL: https://google.com.au/search
... etc
```

So in this model, the database fields and filters stay exactly the same. The only difference would be in how the user creates a filter through the UI.

> If not then we just have to explain better what the difference is and how to properly use it. There's a lot of white space below the search box now so we can use that to give tips

Yeah I would look for opportunities to rename things to make them clearer (I suggested `Referrer -> Referrer URL`). And I totally agree, the filter modal is the ideal place for some guidance and education on what the properties mean and how to use the filters.   I can take care of adding these messages once autocomplete is ready and Marko if you have any suggestions let me know.

metmarkosaric:
> However, the _filters_ for these could in theory be merged into one and I think it could be very nice. I'm imagining a single search box for `Traffic source` or something like that. And if I start typing `google`, the suggestions will be:
> 
> ```
> Source: Google
> Referrer URL: https://google.com/search
> Referrer URL: https://google.co.uk/search
> Referrer URL: https://google.com.au/search
> ... etc
> ```

yeah, this is what I meant. no need to change anything the way it works in the background/database as it makes sense the way it is now. i meant changing the UI of the filter/search so it is easier to understand/use for people. the way it is right now with both Source and Referrer in the filter list would not be too clear to many including me the first time I looked at it. the way you have it above is so much more intuitive

some description on filter modals to inform people on best practices for search will be useful. as long as i understand what different options/wildcards etc we have for when searching, no problem to write it out and it won't take long

Vigasaurus:
@ukutaht @metmarkosaric If y'all are around and want to discuss it a bit, here's my current (basically finished, just needs some polish) iteration of the auto-complete setup - let me know what you think. I personally think it's great and makes adding filters super easy. 
https://dev.vigneshjoglekar.com/secrethitler.io

As for adding further guidance, let me know what you'd want it to say/if you think with this auto-complete it makes it clear enough for a user (I think the immediate suggestions, do, but lmk).

Vigasaurus:
> you're fast. The suggestions look good to me, that's exactly what we're after.
👏 

> feels like it waits a whole second before fetching suggestions (?)
Yeah - I'll bring it down, I had it up mostly for testing, had it at 200ms before, and with with the speed of clickhouse even on my smallish dev server, the volume of them is not a problem at all. It's running CH for like 50-100ms for most of the suggestions. 

> consistent search-select box 
Yeah - I'll bring them all into the setup to match the rest (I've never really liked most of the available components to do this, they are often annoyingly hard to style, especially with tailwind). I did Screen and Countries first and with a static local-list, so they were done different. I'll likely just merge them into the same UI

metmarkosaric:
Looks much better for sure!

Like discussed, would be nice to put Referrers and Sources as one filter and then split results with some tag for Sources and Referrers similar to what Uku described.

Perhaps we can add tags to browser/os version results too? Maybe people know which version numbering is what OS/browser but for someone like me it just shows numbers and I have no idea what is Firefox, Chrome etc

Definitely allowing search for countries would be nice/much faster too in addition to the list as of now.

When I search and select a filter will it show the dashboard immediately or do I still need to press on "Add filter" button? I feel it will be best if it automatically shows the filtered dashboard without needing another click.

Looking forward to testing this on our own live demo where I'm more familiar with the data.

Vigasaurus:
> Like discussed, would be nice to put Referrers and Sources as one filter and then split results with some tag for Sources and Referrers similar to what Uku described.

Yeah maybe, I'll see if I can make that work without major changes, I just fear that with a `Source:` preceding the name, the box may not be wide enough to not look cramped, and not putting it there is just as confusing, probably. (Honestly, this'll probably need to get abstracted to a different PR and out of this one's scope, unless I can find a straightforward way to get it working)

> Perhaps we can add tags to browser/os version results too? Maybe people know which version numbering is what OS/browser but for someone like me it just shows numbers and I have no idea what is Firefox, Chrome etc

Perhaps, but for clarity, the auto-complete suggestions are based on the current filters, so if you filter by Firefox, then try to add a version, it'll only offer Firefox versions, same for OS. 

> When I search and select a filter will it show the dashboard immediately or do I still need to press on "Add filter" button? I feel it will be best if it automatically shows the filtered dashboard without needing another click.

It's all set up with keyboard shortcuts, so you can navigate the suggestions with the arrow keys, press tab or enter to select the suggestion, and then enter at any time to add the filter. So you could just navigate to a suggestion, and press enter twice to add the suggestion. I guess selecting a suggestion could automatically add the filter, I again just worry it could cause frustration or annoyance if a user mis-selects or wants to go back and change the exclusion or similar, if it happens automatically.

ukutaht:
> I've never really liked most of the available components to do this, they are often annoyingly hard to style, especially with tailwind

Agreed. The one I linked seems interesting, I haven't used it but it has a Headless mode which seems ideal for styling with Tailwind.

> Honestly, this'll probably need to get abstracted to a different PR and out of this one's scope, unless I can find a straightforward way to get it working

Yes, as step 1 let's get autocomplete functional in a 1-to-1 mapping to the filters themselves. As the next step we can look at collapsing some filters into a single search box. For browsers I think a similar idea could work as I offered for Sources. Start typing `Chrome` and the suggestions are:

```
Chrome
Chrome 90.0
Chrome 91.0
Chrome 89.0
etc...
```

Seems like a bit of a nightmare to make that work in Clickhouse but I can take a stab at this once basic autocomplete is ready :)

Vigasaurus:
> Seems like a bit of a nightmare to make that work in Clickhouse

Maybe, maybe not - I had to do it pretty much individually for each filter in a big `case`, since Ecto doesn't really like dynamically setting the column name in fragments (which are necessary for the `ilike`) so doing this can be as simple as finishing the grouping to make the suggestions as normal, and then for grouped fields, grouping the versions right on top - i.e. `group by browser, browser_version` and just return each pair as a suggestion. The UI would definitely be a bit more complex to get right without adding 50 suggestions :P

Also one thing I was considering adding and forgot for the first pass, and I thought your feedback could be good on, was having suggestions for the globs directly, i.e. suggest all the normal suggestions, then `/blog/*` and `/blog/**` as the last 2 always (assuming the input was `/blog/` - thoughts? Or you just think a tooltip saying globs are available is enough?

ukutaht:
@Vigasaurus I wouldn't worry about suggesting glob patterns. 

Once this work is done we can focus on adding new filter types `contains`, `starts with` and `ends with`. I'm thinking these are much more accessible for 90%+ of the userbase compared to glob patterns. Even I would probably prefer clicking on 'contains' and writing 'blog' instead of typing '\*\*blog\*\*'.

metmarkosaric:
yep, agree! contains, exact match etc is what i use all the time while i never used (or even heard about) glob patterns before 🤣

PR #1073, opened 2021-05-23
Added parameters for Clickhouse flush interval and max buffer size

ameshkov:
### Changes

Currently, both "flush interval" and "max buffer size" are hard-coded to relatively small values.
This may cause an unnecessarily high load on the Clickhouse instance.
It makes sense to make those configurable.

### Tests

### Changelog

### Documentation

Docs haven't been updated, but if you're okay with this PR, I'll open a PR there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - CSV export now includes pageviews, bounce rate and visit duration in addition to visitors plausible/analytics#952
 - Send stats to multiple dashboards by configuring a comma-separated list of domains plausible/analytics#968
 - Time on Page metric available in detailed Top Pages report plausible/analytics#1007
+- Added `CLICKHOUSE_FLUSH_INTERVAL_MS` and `CLICKHOUSE_MAX_BUFFER_SIZE` configuration parameters
 
 ### Fixed
 - Fix weekly report time range plausible/analytics#951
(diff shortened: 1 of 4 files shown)


ukutaht:
I think this is a good addition. I missed the fact that we flush sessions every second instead of every 5-10 seconds which seems more reasonable to me.

What values are you using? Or is this PR in preparation for testing on your end? The values that are hard-coded at the moment are pretty arbitrary so I'm curious to hear what settings work well in the wild.

ukutaht:
Would you be so kind as to document the changes as well? https://github.com/plausible/docs/blob/master/docs/self-hosting-configuration.md

ameshkov:
Hi @ukutaht, thanks for accepting the PR, I'll prepare the documentation changes asap.

> What values are you using? Or is this PR in preparation for testing on your end? The values that are hard-coded at the moment are pretty arbitrary so I'm curious to hear what settings work well in the wild.

Tbh, we had to configure [clickhouse-bulk](https://github.com/nikepan/clickhouse-bulk) to mitigate this issue without the need to fork Plausible.

Regarding the acceptable values, a lot depends on your Clickhouse instance's overall load and how many instances write to it in parallel. In our case, we have a pretty huge instance under heavy load from several applications so adding writes from Plausible every second was problematic.

Clickhouse devs [suggest](https://clickhouse.tech/docs/en/introduction/performance/#performance-when-inserting-data) this:
> We recommend inserting data in packets of at least 1000 rows, or no more than a single request per second

If Plausible is the only client of Clickhouse and there's just a single Plausible node, I guess 5 secs / 10k buffer would be okay. But once you start scaling it and adding more nodes, you should probably consider using higher values.

ukutaht:
Ah OK that makes sense. Indeed once every second is probably too fast, even for a single node.

I'll see what this update does to our Clickhouse host metrics and I'll probably try a 10 second interval as well.

PR #1096, opened 2021-05-30
make tzdata dir configurable

happysalada:
### Changes

tzdata will regularly update its data and write the result to a directory.
This will default to the priv directory.
This PR adds a way to configure it.
this enables the whole release to be read only.

### Tests

### Changelog

### Documentation

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -15,9 +15,7 @@ end
 base_url = URI.parse(base_url)
 
 if base_url.scheme not in ["http", "https"] do
-  raise "BASE_URL must start with `http` or `https`. Currently configured as `#{
-          System.get_env("BASE_URL")
-        }`"
+  raise "BASE_URL must start with `http` or `https`. Currently configured as `#{System.get_env("BASE_URL")}`"
 end
 
 secret_key_base =
@@ -303,3 +301,5 @@ config :logger, Sentry.LoggerBackend,
   capture_log_messages: true,
   level: :error,
   excluded_domains: []
+
+config :tzdata, :data_dir, System.get_env("TZDATA_DIR", "priv")


ukutaht:
>  this enables the whole release to be read only.

Is this a requirement for NixOS packages? (sorry I don't know anything about that ecosystem :/ )

We have ideas and plans to add more automatically updated databases for things like:
* User-agent database
* IP-to-location database
* Referrer database

These are currently baked into the release and updated during build time. However, the plan is for the Plausible release to keep its databases for Timezones, IP addreses, Referrers, etc. in sync automatically.

I suppose the directory for each of these can be separately configured. But maybe it would be more future-proof to name it something like `STORAGE_DIR` instead and use that as the base directory for any current or future files that are managed by the release. `priv` is still the right value for it. What do you think?

happysalada:
Here is a bit more information on nixos.
- nixos is really just a bunch of shell scripts wrapping a linux kernel.
- For every package that it builds (plausible is an example), it will calculate a unique hash of the resulting file and have everything in that package being read only. So that internally plausible 1.3 is stored at `/nix/store/lkgyfmla311npa9rv00byhw9qn8lkz3h-plausible-1.3.0`. The reason things are done like this is twofold. If you want to install other versions of plausible on your system, now you can. If on the other hand you are running two different services with the same version, then they share the package but will just have different 'state' directory. The state could be different environment variables passed to it, different directories where it stores variable data...
- You can then work with the OS tools directly for process management (essentially systemd) and not need docker (you can still run it if you want to).

There is more to it, but I tried being concise. Feel free to ask any questions you want.
So to answer your question. Any data that changes and that is not part of the original package should be written in a separate directory. It's part of the 'state' of the running service, and the package should stay the same.

So that STORAGE_DIR is exactly what it is. If you want I can make the change to STORAGE_DIR directly in this PR.

On another note, the PR to add plausible to nixos has just been merged. It will be a few more days before it becomes available on the unstable channel. I plan to test this a bit before writing some documentation. If you're open, I can add the doc to this repo.

Let me know if anything.

happysalada:
Did the requested change in https://github.com/plausible/analytics/pull/1104

ukutaht:
>  So to answer your question. Any data that changes and that is not part of the original package should be written in a separate directory. It's part of the 'state' of the running service, and the package should stay the same.

That actually sounds like a reasonable design principle for our Docker setup as well. If the geolocation data is updated, for example, the updated data should survive server restarts and even version upgrades. So it would be a good idea to have a `STORAGE_DIR` as a mounted volume when running through Docker.

I like this change. Will merge the other PR.

Issue #972, opened 2021-04-28
Mobile dashboard design / UX / a11y


metmarkosaric:
We're aware that the Plausible dashboard is not as optimized on smaller screens as on laptops/desktops. We will collect/combine all the main problems in one thread and prioritize the work to get them fixed. I'll start with some which I notice on my own device:

## The traffic numbers on the main chart get squeezed and overlap 

The Y value on the main chart that shows the numbers of unique visitors can be squeezed and overlaps all depending on what traffic you're getting. Here's an example of how it looks like when not optimized:

![photo5785283878305511168](https://user-images.githubusercontent.com/34340819/116375526-a4ee7980-a80f-11eb-8a05-b26bf36b0b72.jpg)

## Details view doesn't allow scroll left or right

For every chart with the "Details" button (Top Sources and Top Pages), the report that opens up when you click on "Details" is not optimized:

* some of the entries are not displayed correctly depending on their length 
* it is not possible to scroll the table left or right so some metrics are missing 

![photo5785283878305511167](https://user-images.githubusercontent.com/34340819/116375645-c51e3880-a80f-11eb-82ef-97020a11205b.jpg)

## Goals area has display issues depending on the name of your custom events

The goals area is not displayed correctly all depends on what type of goals you have set. Some examples here:

![photo5785283878305511166](https://user-images.githubusercontent.com/34340819/116375709-d36c5480-a80f-11eb-8d8f-80632b73fa1a.jpg)

![photo5785283878305511164](https://user-images.githubusercontent.com/34340819/116375793-e717bb00-a80f-11eb-99ec-0d42e7f88ab4.jpg)

## It is not possible to view labels for outbound links, 404 and custom props 

When you open up the goals (404 error pages, outbound link clicks and any goal with custom properties) you cannot see the main label on the portrait orientation. You have to switch to landscape mode to see them. There's a thread with more details here https://github.com/plausible/analytics/issues/534

![photo5785283878305511165](https://user-images.githubusercontent.com/34340819/116375766-df581680-a80f-11eb-88fe-2a7fedee5769.jpg)

![photo5785283878305511163](https://user-images.githubusercontent.com/34340819/116375909-07477a00-a810-11eb-9b40-97c26015d985.jpg)

## Details button issue on Firefox specifically

Firefox specific. The "Details" button on charts goes over the main chart and makes it impossible to click on the main entries:

[...]

PR #1114, opened 2021-06-08
Fix for details button misbehaving on mobile

hirusi:
### Changes

This is a "first pass" for the details button issue being tracked in #972 titled "Details button issue on Firefox specifically". I'd love some testing and feedback on this. I'm certain I may have missed something! 😅 

While it includes a bunch of changes related to development documentation and (optionally) usage (detailed below), I'd love feedback on https://github.com/hirusi/analytics/commit/4cec202a1b3e6445b4aa66814ed0eff96d47f2fb, which is the one and only commit to ultimately build on and tackle the mobile responsiveness issue.

While I have tried to add some comments at some places, they were not a focus so far. If something is unclear, please let me know, and I'll work on adding some. Context is easy to lose. :)

For development related changes...

- It introduces two more convenience commands that I really enjoy using: `make install` and `make server`. They are documented in the CONTRIBUTING.md file. Since I am not familiar with the Phoenix framework, this saves me a ton of time whenever I boot up the project.
- It also adds a note about docker volumes.
- And finally, there are instructions for using `pre-commit`, which was PR'ed by @maco some six months ago. 🧡
- `pre-commit` requires Python so that has also been added to the `asdf` config file.

A further commit is pending for updating the  Changelog.

### Tests

### Changelog

### Documentation

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,4 @@
 elixir 1.11.3-otp-23
 erlang 23.2.1
 nodejs 15.3.0
+python 3.9.4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,4 +36,4 @@ Note: Since we are deleting the containers, be careful when deleting volumes wit
 
 ### Pre-commit hooks
 
-`pre-commit` requires Python to be available locally and covers JavaScript and CSS. Set up with `pip install --user pre-commit` followed by `pre-commit install`.
+`pre-commit` requires Python to be available locally and covers JavaScript and CSS. Set up with `pip install --user pre-commit` followed by `pre-commit install`. Conversely, if the prompts are far too bothersome, remove with `pre-commit uninstall`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,12 +9,13 @@ Make sure Docker, Elixir, Erlang and Node.js are all installed on your developme
 ### Start the environment:
 
 1. Run both `make postgres` and `make clickhouse`.
-2. Run `mix deps.get`. This will download the required Elixir dependencies.
-2. Run `mix ecto.create`. This will create the required databases in both Postgres and Clickhouse.
-3. Run `mix ecto.migrate` to build the database schema.
-4. Run `npm ci --prefix assets` to install the required node dependencies.
-5. Run `mix phx.server` to start the Phoenix server.
-6. The system is now available on `localhost:8000`.
+2. You can then get set up with the following bits in one go with `make install`.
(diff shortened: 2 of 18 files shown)


ukutaht (commented):
Works great! The button no longer overlaps and I really like that you've added it some extra spacing for the button in mobile. Fat-fingered users like myself with appreciate it :)

All the quality-of-life improvements are also much appreciated ❤️

Looking through the code, 2 main things stand out to me. 
1. I know people have differing opinions on TailwindCSS or utility-first CSS in general. On this project we've went the utility-based CSS route. I think we should completely commit to it because mixing multiple different styles is way worse than just picking one. The goal is to do everything we can using default Tailwind classes so ideally the `app.css` file would be empty. I remember adding a `.stats-item` class because I couldn't do a specific height or `calc(50% - 6px)` widths in Tailwind. So some things need to stay in the css file currently but everything that can be added directly to the elements should be added directly to keep a consistent style. I'd like to see how much stuff we can move out of `app.css` and in the context of this PR, how much we can keep from being added to it.

Let me know what you think.

=== OUT OF SCOPE FOR THIS PR ===

I think we could get rid of the whole `.stats-item` class by using Tailwind's `grid` primitives instead of `flex`. It has sligthly worse browser support but I don't think we need to worry about IE in our dashboard. 

2. I also know it _feels_ bad to duplicate the Tailwind classes over and over. This is something I've been very bad at doing but the idea in Tailwind is to extract something like a React component instead of a CSS class. (like `assets/js/dashboard/stats/more-link.js`).

The codebase is really at a point where further duplication is pointless and some refactoring is needed. To explain my approach, I like to let duplication get pretty bad before tackling it. Once you have many duplicates of the same thing, it's much clearer to see how one should extract it into a reusable thing. Inspired by [Sandi Metz](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction) :)

hirusi:
🎉 That's great to hear!

Happy to write them as classes instead. It really helped me to name them so I could understand the structure, but I'm absolutely happy to make those changes now. I think we should be able to move most or all from the classes I introduced to Tailwind style declaration. :)

As for the CSS grid property, I'm generally not too familiar with the vanilla syntax (and therefore Tailwind's), so I'll probably refrain from using that here.

P.s. Solid link-drop there, perhaps I can take some of that into my work today. 😀

maco:
👍🏼  thanks for the precommit docs :)

hirusi:
That should take care of the issue now, @ukutaht. :)

ukutaht:
Great, thanks @hirusi !

metmarkosaric:
this is now live and the details buttons work all fine on my mobile firefox! thanks @hirusi!

PR #1119, opened 2021-06-10
update to latest elixir and erlang

happysalada:
### Changes

This updates the tool-versions
Elixir to 1.12.1
OTP to 24.0.2
node to 16.3.0
I've checked compilation on local.
I've been using those tools with another project in production and didn't observer major problems.

Note that I didn't update the docker image since the tags from the origin image are old
https://hub.docker.com/r/hexpm/elixir/tags?page=1&ordering=last_updated&name=alpine
Elixir 1.12.1 isn't there
alpine is old too, latest is at 3.13.5

### Tests

### Changelog

### Documentation


.tool-versions
```diff
@@ -1,3 +1,3 @@
-elixir 1.11.3-otp-23
-erlang 23.2.1
-nodejs 15.3.0
+v1.12.1-otp-24
```
ukutaht, line 1: this doesn't seem right.. maybe that's why the build is failing? Should be:

```
elixir 1.12.1-otp-24
```

happysalada: Oh, my bad

happysalada: should be good to go

happysalada:
Just checked and the failure is mix format on non related files

ukutaht:
Is there any rush with this? I'd prefer to upgrade when the new Docker image is available as well

happysalada:
No rush at all.

ukutaht:
Closing in favour of #1169

PR #1125, opened 2021-06-13
feat: add plausible_config and allowLocalhost

riderx:
The goal of this PR to allow users to send from localhost, In case needed, like in Cordova apps or Capacitor
Do to so, I had to add a plausible_config system is to save config across calls. 

I choose to implement it as `allowLocalhost` instead of @metmarkosaric suggestion `debug`, whitch is less opinionated, and define better what it does.

### Changes

I added plausible_config in window local storage, you can set it with `plausible('plausible_config', {allowLocalhost: true})`
Then each call to plausible will get data from locastorage and merge it with given options.
given options will override plausible_config.

Check the doc here: https://github.com/plausible/docs/pull/103

### Tests

### Changelog
- [X ] Entry has been added to changelog

### Documentation
- [ X] [Docs](https://github.com/plausible/docs) have been updated 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
   `DATABASE_SOCKET_DIR` & `DATABASE_NAME` were added.
 - Time on Page metric available in detailed Top Pages report plausible/analytics#1007
 - Added `CLICKHOUSE_FLUSH_INTERVAL_MS` and `CLICKHOUSE_MAX_BUFFER_SIZE` configuration parameters
+- Added `plausible_config` event to save global options
+- Added `allowLocalhost` option to let plausible send localhost event in certain conditions
 
 ### Fixed
 - Fix weekly report time range plausible/analytics#951
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -17,7 +17,10 @@
   }
 
   function trigger(eventName, options) {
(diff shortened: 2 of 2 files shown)


Vigasaurus:
I think a better implementation may be to use a script variant, so someone could just load `plausible.localhost.js` - and could this could be more easily used, and would properly work for the initial pageload.

riderx:
@Vigasaurus seems ok to me, it will make the script lighter, that will duplicate a lot of code on other hands, @metmarkosaric what do think ?

Vigasaurus:
What do you mean code duplication? There shouldn't be any - if you look in the plausible.js file, you'll see some sections wrapped with `{{#if condition}} ... {{/if}}`, similarly `allow_localhost` could be added as an option (and the condition might read `{{#if !allow_localhost}}`, and then in `/tracker/compile.js` and `/lib/plausible_web/plugs/tracker.ex` you'd add the script variant `allow-localhost` to the arrays of script types (with exclusions, hash, etc).

ukutaht:
Thanks for the PR @riderx!

I agree with @Vigasaurus, adding it as a script variant/extension makes a lot more sense to me. Happy to merge a PR using that approach!

riderx:
@Vigasaurus oh yes I didn't get where this was coming from so I just ignored it !
I will do it thanks @Vigasaurus @ukutaht

riderx:
I created a new pr following your recomendation @Vigasaurus  : https://github.com/plausible/analytics/pull/1218

PR #1160, opened 2021-06-28
Feature/mobile graph

hirusi:
### Changes

This solves the "The traffic numbers on the main chart get squeezed and overlap" issue in #972. Unsure if this requires any tests?

### Tests

### Changelog

### Documentation


ukutaht:
I'm just testing this and while I can reproduce the issue on an [older version](https://testing.plausible.io/plausible.io) of the dashboard, I cannot seem to reproduce the issue on the [current version](https://plausible.io/plausible.io) of the dashboard. I upgraded from Chart.js v2 to v3 between the two versions so it's possible that it just fixed it.

@hirusi Can you reproduce on the current live demo?

metmarkosaric:
i just tested it and it looks fine both on firefox and brave for me. seems like this one is all sorted now

hirusi:
Hmm, @ukutaht is right at first glance @metmarkosaric. I tested staging and production, and staging seems fine. I'm assuming that's a different deploy to this PR.

I'll check without my fix but latest `master` and report back.

hirusi:
Right - I just tested this locally on the latest master without my fix, and there are no issues on the Y axis

image

If there is any interest in reducing the clutter on the X axis, this PR will be helpful.

X axis with _this_ fix and `Last 30 days` selected:

image


X axis without _this_ fix and `Last 30 days` selected:

image

---

If X axis is not a concern, I think we can close this PR.

ukutaht:
@hirusi I prefer it with your fix, it's cleaner and more readable. Thanks!

PR #1165, opened 2021-07-06
Feature/goals bar text

hirusi:
### Changes

This fixes two issues from #972:

- Goals area has display issues depending on the name of your custom events
- It is not possible to view labels for outbound links, 404 and custom props

Some screenshots:

image

Totals are hidden on mobile:

image

When text spills over, Bar updates its height naturally:

image

I would also recommend testing every single place that uses a bar, since this PR touches every single place. :)

### Tests

### Changelog

### Documentation


hirusi:
PR is up-to-date with the latest `master`.

assets/js/dashboard/stats/pages/entry-pages.js
```diff
@@ -40,15 +40,20 @@ export default class EntryPages extends React.Component {
...
+          bg="bg-orange-50 dark:bg-gray-500 dark:bg-opacity-15"
+          content={(
+            <span className="flex px-2 py-1.5 group dark:text-gray-300 relative break-words z-9">
+              <Link to={{pathname: window.location.pathname, search: query.toString()}} className="block hover:underline">{page.name}</Link>
+              <a target="_blank" href={`http://${  this.props.site.domain  }${page.name}`} className="hidden group-hover:block">
+                <svg className="inline w-4 h-4 ml-1 -mt-1 text-gray-600 dark:text-gray-400" fill="currentColor" viewBox="0 0 20 20"><path d="M11 3a1 1 0 100 2h2.586l-6.293 6.293a1 1 0 101.414 1.414L15 6.414V9a1 1 0 102 0V4a1 1 0 00-1-1h-5z"></path><path d="M5 5a2 2 0 00-2 2v8a2 2 0 002 2h8a2 2 0 002-2v-3a1 1 0 10-2 0v3H5V7h3a1 1 0 000-2H5z"></path></svg>
+              </a>
+            </span>
+          )}
+          maxWidthDeduction="4rem"
```
ukutaht, line 55: Looks like you're giving more space for the Bar components in the Devices tab. I'm assuming that's because these reports also show a percentage so there's less space for the bar.

I would prefer to be completely consistent with these so there are only two 'sizes' for the Bar component. Bigger one for the Conversions report and a smaller one for all others. The Bar component could default to the smaller size so the `maxWidthDeduction` would only have to be passed from the bigger components. What do you think?

assets/js/dashboard/stats/conversions/index.js
```diff
@@ -26,39 +48,51 @@ export default class Conversions extends React.Component {
...
         <div className="flex items-center justify-between my-2">
-          <div className="relative w-full h-8 dark:text-gray-300" style={{maxWidth: 'calc(100% - 16rem)'}}>
-            <Bar count={goal.count} all={this.state.goals} bg="bg-red-50 dark:bg-gray-500 dark:bg-opacity-15" />
-            {this.renderGoalText(goal.name)}
-          </div>
+          <Bar
+            count={goal.count}
+            all={this.state.goals}
+            bg="bg-red-50 dark:bg-gray-500 dark:bg-opacity-15"
+            text="dark:text-gray-300"
```
ukutaht, line 89: Why does only this one Bar need a different text color?

ukutaht (changes requested):

assets/js/dashboard/stats/conversions/index.js
```diff
@@ -26,39 +48,51 @@ export default class Conversions extends React.Component {
...
     if (this.props.query.period === 'realtime') {
-      return <span className="block px-2" style={{marginTop: '-26px'}}>{goalName}</span>
+      return <span className="block px-2 py-1.5 relative z-9 break-words">{goalName}</span>
     } else {
       const query = new URLSearchParams(window.location.search)
       query.set('goal', goalName)
 
       return (
-        <Link to={{pathname: window.location.pathname, search: query.toString()}} style={{marginTop: '-26px'}} className="block px-2 hover:underline">
+        <Link to={{pathname: window.location.pathname, search: query.toString()}} className="block px-2 py-1.5 hover:underline relative z-9 break-words">
```
ukutaht, line 69: Missing a text color override for dark mode: 
Screenshot from 2021-07-06 16-12-15@2x

assets/js/dashboard/stats/conversions/index.js
```diff
@@ -8,11 +8,33 @@ import numberFormatter from '../../number-formatter'
 import * as api from '../../api'
 import LazyLoader from '../../lazy-loader'
 
+const FALLBACK_WIDTH = 639
```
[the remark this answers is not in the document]
hirusi: Interesting. I reviewed this again too after you brought it up and looks like we may not need it at all. At least, not in these two components for conversions.

assets/js/dashboard/stats/conversions/index.js
hirusi: Hmm, good question. Might need to default the dark color, also taking into account https://github.com/plausible/analytics/pull/1165/files#r664542615.

assets/js/dashboard/stats/conversions/index.js
```diff
@@ -26,39 +48,51 @@ export default class Conversions extends React.Component {
...
-          <div className="relative w-full h-8 dark:text-gray-300" style={{maxWidth: 'calc(100% - 16rem)'}}>
-            <Bar count={goal.count} all={this.state.goals} bg="bg-red-50 dark:bg-gray-500 dark:bg-opacity-15" />
-            {this.renderGoalText(goal.name)}
-          </div>
+          <Bar
+            count={goal.count}
+            all={this.state.goals}
+            bg="bg-red-50 dark:bg-gray-500 dark:bg-opacity-15"
+            text="dark:text-gray-300"
+            content={this.renderGoalText(goal.name)}
```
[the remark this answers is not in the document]
hirusi: Correct. You're right here too! My React is a bit rusty. This is an obvious improvement. :)

assets/js/dashboard/stats/pages/entry-pages.js
hirusi: For the most part, I have simply copied down whatever the max width deduction was already being made for any given component. I haven't taken a look at them holistically.

The only new value I have introduced for max width deduction is a 10rem, down from 16rem, in conversions on mobile view, where one more column is removed and the bar is free to take up more space on account of this column removal.

From what I can from the commit, there appear to be three values now in total: 4rem, 16rem, and 10rem. We can probably standardize 4rem as the default since that one repeats the most.

What do you think?

ukutaht: Ah OK. For some reason I thought this was something you introduced in this PR. No need to worry about it then, I can also take a look later :)

assets/js/dashboard/stats/conversions/index.js
ukutaht: I checked again and it seems like this doesn't do anything. It sets the `color` property on the bar itself (not the text in it). Since the bar is colored using `bg-color` anyways, this should not affect anything.

I'm pretty sure the `text` property can be safely deleted.

ukutaht (approved):

ukutaht:
Amazing, thank you @hirusi!

PR #1179, opened 2021-07-13
Fix shell scripts

nh2:
### Changes

There were various mistakes in the shell scripts, such as lack of quoting; they would break them immediately e.g. if `BIN_DIR` contained a space.

Pointed out by [`shellcheck`](https://www.shellcheck.net/).

### Tests

### Changelog

### Documentation


ukutaht:
Thanks! Looks good to me

PR #1190, opened 2021-07-15
Allow user to specify listen interface via LISTEN_IP

nh2:
This enables safer deployments to allow localhost-only or VPN-interface-only listening.

### Tests

### Changelog

### Documentation

- [x] [Docs](https://github.com/plausible/docs) have been updated: https://github.com/plausible/docs/pull/112

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -7,6 +7,18 @@ end
 
 config_dir = System.get_env("CONFIG_DIR", "/run/secrets")
 
+# Listen IP supports IPv4 and IPv6 addresses.
+listen_ip = (
+  str = get_var_from_path_or_env(config_dir, "LISTEN_IP") || "0.0.0.0"
+  case :inet.parse_address(String.to_charlist(str)) do
+    {:ok, ip_addr} ->
+      ip_addr
+
+    {:error, reason} ->
+      raise "Invalid LISTEN_IP '#{str}' error: #{inspect(reason)}"
+  end
+)
+
 # System.get_env does not accept a non string default
 port = get_var_from_path_or_env(config_dir, "PORT") || 8000
 
@@ -190,7 +202,7 @@ config :plausible, :selfhost,
 
 config :plausible, PlausibleWeb.Endpoint,
   url: [scheme: base_url.scheme, host: base_url.host, path: base_url.path, port: base_url.port],
-  http: [port: port, transport_options: [max_connections: :infinity]],
+  http: [port: port, ip: listen_ip, transport_options: [max_connections: :infinity]],
   secret_key_base: secret_key_base
 
 if is_nil(db_socket_dir) do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 
 ## Unreleased
+- Added `LISTEN_IP` configuration parameter plausible/analytics#1189
 
 ### Added
 - API route `PUT /api/v1/sites/goals` with form params `site_id`, `event_name` and/or `page_path`, and `goal_type` with supported types `event` and `page`


nh2:
For some extra motivation on why restricting listen addresses to localhost is a good default practice, see MongoDB's history, or e.g. https://github.com/NixOS/nixpkgs/issues/130244 for Erlang.

This PR only adds the option, it does not change the default away from `0.0.0.0`.

nh2:
CC @Ma27 @happysalada

config/runtime.exs
```diff
@@ -7,6 +7,18 @@ end
 
 config_dir = System.get_env("CONFIG_DIR", "/run/secrets")
 
+# Listen IP supports IPv4 and IPv6 addresses.
+listen_ip = (
+  str = get_var_from_path_or_env(config_dir, "LISTEN_IP") || "0.0.0.0"
```
happysalada, line 12: Just a thought, shouldn't the default be 127.0.0.1 ?
https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html#module-loopback-vs-public-ip-addresses

happysalada:
An alternative solution to this would be to set RELEASE_DISTRIBUTION to `none` on our side.

I'm not commited to any particular solution, just mentioning an alternative.

https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-environment-variables

config/runtime.exs
nh2: I personally would be happy if the default was `127.0.0.1`, like most software does it, as a safe default.

However, I didn't want to change it here without asking the project first if they want to change the default.

nh2:
> An alternative solution to this would be to set `RELEASE_DISTRIBUTION` to `none`

@happysalada `RELEASE_DISTRIBUTION` does not affect the port that the Plausible HTTP web server is listening on. (This PR is about the HTTP web server, not the Erlang inter-node communication and `epmd` ports.)

However, I've added your recommendation to the docs PR at https://github.com/plausible/docs/pull/112, which talks about both topics.

happysalada:
I was trying to put myself in their shoes and answer the following question
`Why do we need this?`
My understanding was that it was to prevent a security issue where if the cookie is empty and the node is connecting on a public IP then there is a possibility for remote code execution through a connection to EPMD.
(Trying to re-read the link you gave, the issue didn't pop out clearly).
If that is the problem we are trying to fix, then I was just going through alternative solutions so that they could evaluate.

I could be wrong on the problem, and so on the proposed solution.

(Thank you for taking the time to do this BTW!)

happysalada:
Reading more about the docs, I see this PR is just to configure the IP the server is listening on and has nothing to do with a security issue, understood, thanks!

nh2:
@ukutaht I have rebased the PR to fix the conflict, could you have a look at it?

It's a trivial change, the main question is if you also want to switch to the safer default that is `127.0.0.1`, like many other web projects do in their default configuration.

ukutaht:
Thanks @nh2 for this.

I like having more safe and secure defaults. I believe in order to run Plausible in a docker container, it needs to bind to `0.0.0.0`. But we can achieve that by adding `ENV LISTEN_IP=0.0.0.0` to the Dockerfile, thus allowing the elixir release to default to `127.0.0.1`. Nix is the only use-case I know of anyone running Plausible in production but without Docker. 

While erlang is really eager to connect nodes and form a cluster, we don't use any of those features within the Plausible codebase. In fact, Plausible is currently limited to running in a single node because the service itself is stateful with no synchronization between nodes.

We will move to a multi-node setup at some point but I think the synchronization will happen at the level of the database, not the app server. App servers will be stateless and independent. Therefore we can add extra safety and set `RELEASE_DISTRIBUTION=none` by default in all Plausible instances. If we end up needing it in the future, we'll enable it then. Better to be safe in the meanwhile.

I will merge this for now and test it out with some docker containers.

ukutaht:
This seems like a good default to me: https://github.com/plausible/analytics/commit/69576aa253e5dff5f6eb512be8dc2e2c13e21b43

PR #1202, opened 2021-07-23
Hotfix/bar overflow

hirusi:
### Changes

There are two changes made here:

1. An issue with larger screens overflowing when too many list items wrap and expand height-wise.

2. An issue with words breaking very conservatively such that spacing is violated.

Before:

image

After:

image

### Tests

### Changelog

### Documentation


hirusi:
One concern I have is that Tailwind doesn't support max-width responsive styling by default. Would it be worth adding one to limit where `break-words`/`break-all` apply?

ukutaht:
> One concern I have is that Tailwind doesn't support max-width responsive styling by default. Would it be worth adding one to limit where break-words/break-all apply?

Hmm I don't think I fully understand. Could you give a bit more context?

ukutaht:
Looks really good, I think this is a great solution.

The only thing is I think the breakpoint we should target is `md:truncate` instead of `lg:truncate`. If you make the window gradually smaller, there's a point where the reports are still side by side with a fixed height but the text starts wrapping.

Screenshot from 2021-07-23 13-45-05@2x

hirusi:
> Hmm I don't think I fully understand. Could you give a bit more context?

Sure, so instead of being able to limit break-worrds/break-all to lg and below, we have it working all across. And then truncate for the *child* text takes over and truncates it such that it never breaks into the next line.

So this works, but we’d have to [use Tailwind’s configuration](https://tailwindcss.com/docs/breakpoints#max-width-breakpoints) to add a new `uptolg:` breakpoint or similar.

We can’t use the lowest one -- `sm` -- since that will block it from applying to <`sm` ones.

hirusi:
That makes sense. I'll get that committed in, now. :)

July 23, 2021 4:18 PM, "Uku Taht" ***@***.*** ***@***.******@***.***>)> wrote:
	Looks really good, I think this is a great solution. 

	The only thing is I think the breakpoint we should target is md:truncate instead of lg:truncate. If you make the window gradually smaller, there's a point where the reports are still side by side with a fixed height but the text starts wrapping. 

	 (https://user-images.githubusercontent.com/3731516/126771732-e2d9a980-3be1-495a-af29-c20cd2458d3d.png)

ukutaht:
> Sure, so instead of being able to limit break-worrds/break-all to lg and below, we have it working all across. And then truncate for the _child_ text takes over and truncates it such that it never breaks into the next line.
> 
> So this works, but we’d have to [use Tailwind’s configuration](https://tailwindcss.com/docs/breakpoints#max-width-breakpoints) to add a new `uptolg:` breakpoint or similar.
> 
> We can’t use the lowest one -- `sm` -- since that will block it from applying to <`sm` ones.

Ah OK I see. Usually in Tailwind there's a utility class to reset a style if you need it. Instead of `uptolg:break-all` maybe something like this would work? `break-all lg:break-normal`

At least that's how I've been getting around not having max-width breakpoints so far.

But if the current solution works I don't see much reason to add a max-width breakpoint.

hirusi:
That's fine then.

I didn't think the undo variant worked. But I might have tested it in jest.

July 23, 2021 5:04 PM, "Uku Taht" ***@***.*** ***@***.******@***.***>)> wrote:
	Sure, so instead of being able to limit break-worrds/break-all to lg and below, we have it working all across. And then truncate for the child text takes over and truncates it such that it never breaks into the next line. 

	So this works, but we’d have to use Tailwind’s configuration (https://tailwindcss.com/docs/breakpoints#max-width-breakpoints) to add a new uptolg: breakpoint or similar. 

	We can’t use the lowest one -- sm -- since that will block it from applying to <sm ones.  

	Ah OK I see. Usually in Tailwind there's a utility class to reset a style if you need it. Instead of uptolg:break-all maybe something like this would work? break-all lg:break-normal 

	At least that's how I've been getting around not having max-width breakpoints so far. 

	But if the current solution works I don't see much reason to add a max-width breakpoint.

ukutaht:
Perfect, I'm ready to merge as soon as the breakpoint is corrected (`md:truncate` as opposed to `lg:truncate`)

hirusi:
That's ready now. :)

ukutaht:
Thanks!

hirusi:
#972 (linking these together)

PR #1209, opened 2021-07-26
Add City level details

aymanterra:
#### Add City level details to the country modal in the stats page.

- Currently we are showing the geoname_ids, next phase we will query another DB to translate this geoname_ids to readable names according to the language specified.
- The hierarchy works as follow:
   - Country:  name as currently provided in Plausible
   - Subdivision1:  geoname_id comes from...
    (1)  First element in the subdivisions array returned by Geolix (if there is more than 1 subdivision value in the array returned by Geolix) (can contain 1, 2 or [potentially] no values) OR IF subdivisions array length < 2 THEN
    (2)  Country geoname_id (can contain 1 or [potentially] no values) OR IF country array length = 0 THEN
    (3)  NULL
   - Subdivision2:  geoname_id comes from...
    (1)  Second element in the subdivisions array (if there is more than 1 subdivision value in the array returned by Geolix) OR IF subdivisions array length < 2 THEN
    (2)  First element in the subdivisions array (if there is only 1 element in the subdivisions array) OR IF subdivisions array length = 0 THEN
    (3)  NULL
   - City:  geoname_id comes from...
    (1)  City geoname_id from city dictionary returned by Geolix OR IF city dictionary exists THEN
    (2)  Second element in the subdivisions array (if array contains two elements) OR IF subdivisions array length < 2 THEN
    (3)  NULL

Attached a screen recording.

https://user-images.githubusercontent.com/26826457/126918462-a23bc715-2b07-498c-805b-8a106dad004a.mp4


priv/clickhouse_repo/migrations/20210712214034_add_more_location_details.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Plausible.ClickhouseRepo.Migrations.AddMoreLocationDetails do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events) do
+      add :continent_geoname_id, :"LowCardinality(String)"
```
ukutaht, line 6: I think it would make more sense to store these as `UInt32`. The geoname ID should be an integer ([this](https://www.ip2location.com/free/geoname-id) resource says it's `INT(10)` which is equivalent to `UInt32` in clickhouse). Storing them as such will save space and feel more natural I think.

lib/plausible_web/controllers/api/external_controller.ex
```diff
@@ -171,15 +176,61 @@ defmodule PlausibleWeb.Api.ExternalController do
...
+        do: Integer.to_string(result.city.geoname_id),
+        else:
+          if(result && result.subdivisions && length(result.subdivisions) > 1,
+            do: Integer.to_string(Enum.at(result.subdivisions, 1).geoname_id),
+            else: nil
+          )
+
+    _output = %{
+      country_code: country_code,
+      continent_geoname_id: continent_geoname_id,
```
ukutaht, line 228: Is the continent ID used anywhere?

assets/js/dashboard/stats/modals/countries.js
```diff
@@ -21,19 +24,108 @@ class CountriesModal extends React.Component {
...
+
   renderCountry(country) {
     const query = new URLSearchParams(window.location.search)
     query.set('country', country.name)
 
     return (
       <tr className="text-sm dark:text-gray-200" key={country.name}>
         <td className="p-2">
+        <b onClick={() => this.getSubdivisions1(country.name)} style={{ display: this.state.expanded_country != country.name ? "inline-block" : "none" }} className="mr-2 cursor-pointer hover:text-indigo-700 dark:hover:text-indigo-500"> + </b>
+        <b onClick={() => this.setState({expanded_country: ''})} style={{ display: this.state.expanded_country == country.name ? "inline-block" : "none" }} className="mr-2 cursor-pointer hover:text-indigo-700 dark:hover:text-indigo-500"> - </b>
```
ukutaht, line 116: Hiding this component can be achieved via Tailwind classes as follows:

```js
// at the top
import classNames from 'classnames';

<b ... 
  className={classNames("mr-2 cursor-pointer hover:text-indigo-700 dark:hover:text-indigo-500", 
    {"inline-block": this.state.expanded_country === country.name, 
     "hidden": this.state.expanded_country !== country.name})
> - </b>
```

We're trying to avoid `style=` wherever possible and do it with utility classes instead.

ukutaht (commented):
Looks good to me overall. Glad you spent the time to figure out how to properly look up the data from Geolix.

A couple comments but I can hopefully continue work on this later this week.

- [ ] Translate geoname IDs to names when presenting to the user
- [ ] Add filters for subdivision 1, subdivision 2 and city

ukutaht:
City-level geolocation data has been merged to master and is being tested in our production environment at the moment. There's still some work left to translate the location names properly but it's going well

PR #1211, opened 2021-07-27
Feature/details modal mobile

hirusi:
### Changes

This updates the modals to use a specified width on smaller screens for the first table column. They can be scrolled horizontally as well as vertically. Spacing is reduced so that this is easier to do. Longer text wraps into a new line from previous fixes, so thihs is not a big deal.

On desktop, the previous behaviour is retained: set the width to 800px. Additionally, we now allocate either 1/2 or 1/3 of the 800px width to the first column to ensure it gets at least a give space on a bigger screen.

### Tests

### Changelog

### Documentation


ukutaht:
**On mobile**

Looks good to me. The only thing I would change is reduce the padding on the left and right so more content is visible before scrolling. 

Screenshot from 2021-07-30 10-00-36@2x

**On desktop**

The detail modals seem pretty broken to me. They're wider than the screen so the stats are not visible.

Screenshot from 2021-07-30 10-01-43@2x

hirusi:
FYI @ukutaht - this is ready for review. :)

ukutaht:
Works great! Any reason to change the default modal width from 860px to 800px?

I personally prefer the wider (860px) modals on desktop.

hirusi:
Good catch! 😇 Oversight from me, sorry.

ukutaht:
Looks good, thanks!

hirusi:
#972 (linking these together)

PR #1256, opened 2021-08-20
Condense and show all site nav/options on mobile

hirusi:
### Changes

Per #972, fix the following:

> "Also, on mobile, the add filter button doesn't fit well, and there are no more the arrows to switch to previous day"

> "noticed that custom filter breaks the layout on mobile screens a bit"

### Tests

### Changelog

### Documentation


ukutaht:
Looks good to me!

PR #1267, opened 2021-08-24
Hotfix/outbound links

hirusi:
### Changes

This -- attempts to -- add a `is_url` key on the breakdown stats. I think it'll need to be run only on url breakdown type. I've also been unable to test this locally since I don't have the relevant test data.

### Tests

### Changelog

### Documentation


ukutaht:
Fixed in https://github.com/plausible/analytics/commit/021cffaf112f802af193fe94560781ebac27a05d

PR #1299, opened 2021-09-04
Add conversion_rate to sources api and source table

ro-savage:
### Changes

As per #1296 and #1115 this adds the conversation rate to the Top Sources table when filtering by a goal.

The API will only return the conversion_rate when a goal filter has been applied.

### Screenshots
image

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated (Not required)


ro-savage:
@ukutaht - Any feedback on if this is the correct way to go about getting the conversion right, and if there is anything I can do to make the elixir code nicer.

Once you're happy with that, then I can update tests, changelog, docs, etc.

assets/js/dashboard/number-formatter.js
```diff
@@ -49,3 +49,7 @@ export function durationFormatter(duration) {
     return `${seconds}s`
   }
 }
+
+export function percentageFormatter(num) {
+  return +(Math.round(num + "e+2")  + "e-2");
```
ukutaht, line 54: Why is this necessary?

assets/js/dashboard/stats/sources/source-list.js
```diff
@@ -66,7 +73,8 @@ class AllSources extends React.Component {
             </Link>
           </span>
         </Bar>
-        <span className="font-medium dark:text-gray-200">{numberFormatter(referrer.count)}</span>
+        <span className="font-medium dark:text-gray-200 w-20 text-right">{numberFormatter(referrer.count)}</span>
+        {showCR && <span className="font-medium dark:text-gray-200 w-20 text-right">{percentageFormatter(referrer.conversion_rate)}%</span>}
```
ukutaht, line 77: Does this look OK on mobile? If it gets too busy we can also only show it on desktop or something like that.

ukutaht (commented):
Don't know what you were talking about in the other thread, the Elixir looks good here!

Tests are missing, I'm happy to take it from here or I can also leave it up to you. Let me know

ro-savage:
Thanks @ukutaht - I'll add give adding the tests a go over the weekend and if I get stuck hand it over to you.

assets/js/dashboard/number-formatter.js
ro-savage: Good spotting. This can be removed now.

It rounds to 2 decimals places, and was left over from when I was calculating the % in the react app.

assets/js/dashboard/stats/sources/source-list.js
ro-savage: Sounds like a good idea. I'll check on mobile and see what it looks like.

ukutaht:
> Thanks @ukutaht - I'll add give adding the tests a go over the weekend and if I get stuck hand it over to you.

Sounds good :)

assets/js/dashboard/stats/sources/source-list.js
ro-savage: Looks good in mobile - I'll leave this on for mobile unless you'd prefer different.

image

ro-savage:
@ukutaht - This should be ready to go if you are happy with the formatting.

I also added the CR to the modals, and I changed the naming convention and style for the 'should show' function to match what you already had in the app.

metmarkosaric:
nice work, thanks @ro-savage! 

would it be possible to add the CR to the top pages as well to make it all consistent? the original feature request mentions both sources and pages.

i assume when you click on "details" you also see the CR for all the other sources and not only the top sources? 

curious to play around with this and test it. wondering if it looks too busy on the main screen or not. normally we hide the other metrics behind the "details" screen and don't show them immediately but perhaps for this one it might work

no need to change the docs at this stage

ro-savage:
@metmarkosaric - Thanks for the feedback.

1. Happy to add it to pages - originally I had planned to but then felt it probably won't that important for pages and might just add noise. But happy to implement it if you think that is valuable to your customers.

2. When you click details for anything in the "Top Sources" list it will show. Referrers, campaigns, etc

3. The CR only shows when you are filtering by a goal - so this would be when the information is already important to you. So I don't think it gets to busy, but happy to hear your thoughts.

The reason I think CR is highly valuable, at least for a SASS company like mine, is the most important metric we have for our marketing site is "Where did the visitors come from and did they sign up" this lets us focus our marketing spend in the right place. We choose to do our marketing without using any pixel tracking, or in fact any form of tracking beyond UTM. Because of our privacy-focused marketing, UTM becomes the most important and only thing, we can use to ensure we are targeting the right audience. Be that Google Adwords, LinkedIn campaigns, blog content, or sponsors posts. Therefore CR is the metric I (personally) want to see the most when looking at the analytics.

metmarkosaric:
Sounds good, thanks! I will definitely find this addition useful myself as well so no worry about the usefulness of this. I value it! I meant it more in terms of design layout - we'll need to test how it looks on mobile devices, on sites with a lot of traffic, on referral sources/pages that are longer... but for sure the addition of CR will be very useful.

>     1. Happy to add it to pages - originally I had planned to but then felt it probably won't that important for pages and might just add noise. But happy to implement it if you think that is valuable to your customers.

Yeah, CR for pages will definitely be useful for some of the users. It all depends on how you've set goals up. Some might mostly focus on external referrers but some might also look at learning more and optimizing the conversion rate from internal referrers/pages where they have those goals setup. You can see for instance the public dashboard of Elementary OS and there having access to which internal pages drive best CR would be very interesting https://plausible.io/elementary.io?goal=Payment

ukutaht:
Sorry @ro-savage I'm on holiday so this will go slow... don't have time to test/put it on staging at the moment.

re: CR for top pages.

Hopefully it doesn't get confusing between top pages and entry pages. The natural thing to show in Top pages is how well the page itself is converting. For example, this would be useful to see which blog posts are convincing people to sign up to the blog newsletter on our live demo: https://plausible.io/plausible.io?period=30d&goal=Newsletter+signup. 

If you look at our Signup event, the Top Pages report is pretty pointless: https://plausible.io/plausible.io?period=30d&goal=Signup. It will probably just show a very high conversion rate for `/register`. But really the more interesting thing would be to see CR broken down by entry page instead of the actual page of the Signup event.

So that is to say that CR is needed for both Top Pages and Entry Pages.

@ro-savage if you're willing to work on this go ahead but there's no pressure. I'm happy to merge this as is and take it from here as well. It will be much slower that way, though :)

metmarkosaric:
> So that is to say that CR is needed for both Top Pages and Entry Pages.

exactly! depending on the site/goals tracked/situation, either of these three could be the most valuable metric that people care about the most (CR for referrals, top pages or entry pages)

ukutaht:
@ro-savage As a developer I like the idea of doing it piecemeal but we discussed it with Marko and he made a good point. There are a lot of eyes on the project and we want to start announcing features in a relatively complete state.

So, sorry but I'll wait until I can also add CR to the pages card before merging it

ro-savage:
@ukutaht @metmarkosaric - I have now added the CR to all the panels, including the pages and details.

Video: https://www.loom.com/share/50c81e77bdea474aaa01f43873334abd

I ended up adding them to every panel as it seemed weird that it would just be available in some of the panels, and it may turn out to be useful to know the CRs for countries or devices. (E.g. maybe you shouldn't target people on mobile devices or people from Australia)

I haven't been able to manually test for countries (wasn't sure how to fake this with curl) but it should work. I've also updated the tests to include goal filtering tests that include checks of the cr.

Let me know if any other changes are required.

image

metmarkosaric:
this looks great @ro-savage, thank you! looking forward to trying it out on our staging

ukutaht (approved):
Everything looks good to me!

This PR really highlights the amount of duplication we have in the dashboard code :) Thanks for making through it all.

PR #1364, opened 2021-10-01
Adds Main Graph Metric Selection

### LINKED ISSUE #117: Toggle between different metrics in the main graph
"The main graph should be able to show both unique visitors and pageviews, depending on what the user selects."

([imported from the old roadmap](https://plausible.nolt.io/64))

Vigasaurus:
### Changes

Resolves #117 and #532 

This does also include (as discussed) the concept that if a user chooses to deselect the currently selected top stat from the graph, only the top stats show, and the graph gets hidden. 

I did, however - for now - not include choosing time on page or conversion (rate/users) as an available graph selection, purely because they're not well supported (at all) in the new timeseries interface, and implementing them would not be trivial & probably shouldn't be in the scope of this PR alone. Right now, the top stat selection for showing as the graph data supports the 4 primary and default top stats (unique visitors, pageviews, bounce rate, and visit duration). 

I'm also not necessarily super happy with the UI for the comparison enable/disable button in the datepicker, so I'm definitely open to ideas there.

https://github.com/plausible/analytics/pull/1364/commits/2575756740d55e11988bea175e8f392a5ec00350 is referring to this comical tooltip:
image

I'll set up some tests for this (and fix the existing tests to include the new main-graph route) soonish, but I figured it could get pushed for feedback anyways.

### Tests

### Changelog

### Documentation
I imagine this will need both normal docs updates, and API doc updates

### Screenshots

image
image
image


ukutaht:
Thanks @Vigasaurus. Just letting you know I don't have time today to look at it and I'm off until Thu next week. So the review will take some time, sorry. Will that block your other stuff?

Vigasaurus:
@ukutaht All good - and no, not at all; I'm working on the interval selection for the graph technically on top of this in a separate branch, but without much reliance on it (it's really only to make sure I don't cause UI-based merge conflicts, since the file change overlap is almost 100%), and if nothing else, I'll have all the tests etc all sorted by the time you're back.

Vigasaurus:
I took inspiration from the tabs in the other reports for the top stat selection styling - lmk what you think. I do kinda like it personally, but I almost feel like all the headers being bolded with the potential for the new color & underline is maybe a bit much, but that may just be me. 

image

Vigasaurus:
Yeah there's not really any spots in the design where the background changes for text on an interaction, unless it's the entire parent container changing background (like a button or a dropdown item). It's kinda hard to do anything but text color/decoration changes for these, since they don't really have a clean parent we could styles apply to. There is always the way that GA does it where that entire rectangle container changes background color, but I think that'd be really jarring in our design language.

ukutaht:
I like the design as well. Good choice!

assets/js/dashboard/stats/visitor-graph.js
```diff
@@ -107,68 +171,144 @@ class LineGraph extends React.Component {
...
+                tooltipEl.innerHTML = innerHtml;
               }
+              tooltipEl.style.opacity = 1;
             }
           },
         },
         responsive: true,
         onResize: this.updateWindowDimensions,
-        elements: {line: {tension: 0}, point: {radius: 0}},
+        elements: { line: { tension: 0.1 }, point: { radius: 0 } },
```
ukutaht, line 305: Is this just a stylistic choice? Personally I prefer the sharp edges.

assets/js/dashboard/stats/visitor-graph.js
```diff
@@ -107,68 +171,144 @@ class LineGraph extends React.Component {
...
-                const dataset = item.dataset
-                if (!this.drawnLabels[dataset.label]) {
-                  this.drawnLabels[dataset.label] = true
-                  const pluralizedLabel = item.formattedValue === "1" ? dataset.label.slice(0, -1) : dataset.label
-                  return ` ${item.formattedValue} ${pluralizedLabel}`
-                }
-              },
-              footer: function(_dataPoints) {
+            position: 'average',
+            external: function (context) {
```
ukutaht, line 195: Why use external tooltip instead of the default one?

Side note: the whole file is over 600 lines so we have to start organising this a bit better. I think the tooltip code is a prime candidate for extracting in a separate file.

ukutaht: In my testing the tooltip looks nice but doesn't animate as smoothly as the native canvas tooltip

assets/js/dashboard/comparison-consumer-hoc.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ComparisonContext } from './comparison-context'
```
ukutaht, line 2: Just wanted to ask you if you're still working on comparisons? I guess we can leave this stuff in if it's coming soon, otherwise I would get rid of it.

assets/js/dashboard/stats/visitor-graph.js
```diff
@@ -360,53 +554,87 @@ class LineGraph extends React.Component {
...
 
   onVisible() {
     this.fetchGraphData()
     if (this.props.timer) this.props.timer.onTick(this.fetchGraphData.bind(this))
   }
 
-  componentDidUpdate(prevProps) {
-    if (this.props.query !== prevProps.query) {
-      this.setState({loading: true, graphData: null})
+  componentDidUpdate(prevProps, prevState) {
```
ukutaht, line 587: When changing the current metric, there's no reason to unload the top stats and reload them. This bothered me about the UI. When changing between metrics, I think the top stats should stay visible the whole time and just the graph at the bottom should reload.

I totally understand why you went this way, it fits more naturally with the code we already have. But I think it's worth separating the top stats from the graph (maybe the API endpoints as well) so that they can be reloaded independently.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -16,11 +17,21 @@ defmodule PlausibleWeb.Api.StatsController do
...
-    plot = Enum.map(timeseries_result, fn row -> row["visitors"] end)
+    if selected_metric == "none" do
+      json(conn, %{
+        top_stats: top_stats,
+        sample_percent: sample_percent
+      })
+    end
+
+    timeseries_task =
+      Task.async(fn -> Stats.timeseries(site, timeseries_query, [selected_metric]) end)
```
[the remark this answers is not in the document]
ukutaht: Extra note: if you split the API routes for top stats and the plot then they will naturally be parallelized over separate HTTP connections so this point becomes moot.

ukutaht (commented):
Finally have some time to look at this.

Looks great overall! I think the active style is really good. I left comments about a few points I had in the code. Also I noticed that when filtering for a goal, the display is a bit off:

Screenshot from 2021-11-26 14-53-47@2x

There's probably more work under the hood but I think for the user it should simply highlight the `Unique conversions` label at the top.

lib/plausible_web/controllers/api/stats_controller.ex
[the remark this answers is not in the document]
Vigasaurus: Yeah I think splitting those fully could make that a bit nicer overall, both for usability and speed. Will do.

assets/js/dashboard/stats/visitor-graph.js
Vigasaurus: Ah yeah this was just a leftover merge from when I did this super long ago - it came from a change in the Chartjs param spec. I'll get it swapped back.

assets/js/dashboard/stats/visitor-graph.js
Vigasaurus: The main reason for the custom tooltip here was to make it look significantly better - the basic tooltip didn't allow for a lot of styling within (changing the color of the delta percentage, changing the size/font weight of the delta percentage, some other things with formatting the date labels). 

From what I had noticed the smoothness was near negligible, given the benefits of the custom tooltip. 

I might be able to smooth the behavior a bit more, but I do think the custom tooltip is the way to go. 

(Also I will do the file splitting there)

assets/js/dashboard/comparison-consumer-hoc.js
Vigasaurus: I am indeed - I just wanted to make sure both this and the interval selection PRs were nearing completion/finalized overall structure before doing significant work on comparisons to avoid merge conflict hell.

assets/js/dashboard/stats/visitor-graph.js
Vigasaurus: Yeah I did notice that too, but originally had thought it was out of scope of this PR/shouldn't be all bundled together - but if we're splitting it by requests it should be good. Will do.

assets/js/dashboard/stats/visitor-graph.js
ukutaht: Cool makes sense. Fully custom makes sense especially if we start showing comparisons

assets/js/dashboard/comparison-consumer-hoc.js
ukutaht: Fair enough, sorry about the delay on my end!

Vigasaurus:
Just a heads up - @ukutaht both this and #1574 are good to go + ready for review whenever you get a chance. Just wanted to give them a bump to help avoid future conflicts + so I can push further on the comparisons stuff as discussed once they're more finalized.

ukutaht:
Nice, I'll give it another review soon

Vigasaurus:
👀 I fear this'll fall into merge conflict hell soon

ukutaht:
Sorry @Vigasaurus I'm stretched really thin.

This is the next thing on my list and I think I can take care of the conflicts

Vigasaurus:
All good no worries at all, lmk if there's anything I can do to help

Vigasaurus:
By the way - I did try resolving the merge conflicts here but was getting errors in importing my own GA data to validate that the import indicator etc was working correctly post-merge, so I didn't quite yet. 

@ukutaht Let me know if you want any of the errors I had in the import or if you know of it already/if you do need it tested further (it was a NULL column input error from very late in my GA visitors data)

ukutaht:
@Vigasaurus We're still uncovering edge cases with the import. I finalized everything from my own testing yesterday so actually testing the import on master right now would be quite valuable.

Vigasaurus:
@ukutaht Gotcha - I figured. I'll run another test on both my GA data sets and report back.

Vigasaurus:
Update: Merge conflicts here are all resolved - there are a couple of edge cases in the split top-stat and main-graph loading UI/timing that is improved in #1574 (I'll get the conflicts resolved there + some minor improvements I've seen necessary in my self-host usage of it after this is merged).

Vigasaurus:
I'm seemingly unable to reproduce either of those - do you see them [here](https://dev.vigneshjoglekar.com/secrethitler.io) too?

The intended behavior is that with a goal filter you can only graph `Unique Conversions`, and all three of the rest remain unclickable to indicate that.

ukutaht:
OK, I've reviewed everything and I'm ready to merge this. Could you please:
1. Cherry pick this commit: https://github.com/plausible/analytics/commit/fa603452662e35e01a7a680113560c90c53f6f36
2. Resolve conflicts on this branch

And then I will merge?

I can try to resolve conflicts myself but I'm sure it's quicker if you do it since you have a better context

Vigasaurus:
Alright, should be all good to go! I also was able to add the slide up and down transition for the graph getting hidden fully as you'd mentioned + fixed up some UX on the top stat selectors.

ukutaht:
Amazing @Vigasaurus, thank you for sticking with it and keeping the PR up to date. Heroic effort!

ukutaht:
Some feedback:

> For us the most important info on the tooltip would be the actual number - but now it's quite condensed. Maybe if you could take the number to a separate line and make it bold?

I agree with this as well. The actual metric number is not as easy to see anymore. I see why you did it - to fit comparisons in the tooltip as well. But maybe just play around with the text styles a bit to make the metric more prominent?

I can take a look as well, just posting here in case you have time to play around with it :)

Vigasaurus:
Yeah for sure - I could see it being not bad by just scaling up + bolding the number, but I do definitely see how the hierarchy is kinda skewed in the current version where the data is basically the least prominent item. 

0. Before:
image

1. Maybe something like this?
image

2. Or maybe by reducing the prominence of the title, the data can shine a bit more:
image

3. Or maybe combining the previous + reducing the weight of the "Click to view month", and changing the color of the actual data text:
image

4. Similarly with the last one, making the data color be the line color, it might make sense to hide the circle color swatch:
image

I personally don't love removing the swatch, especially when the comparisons are visible (also it makes the tooltip look a bit bare in my opinion). 

I personally most like either 2 or 3, potentially with the reduction of weight in the "Click to view" of 3, and the still white color of 2; but lmk what you think.

metmarkosaric:
thanks @Vigasaurus! I also like your favorite one (also don't mind if you make the metric in color too). And I see no problem with making "click to view" less prominent or completely removing it. I've shared this with the person who write this in the first place to see what they say.

Vrq:
The suggested version looks good!
Does anyone have a Screenshot of how the tooltip looked before?

metmarkosaric:
Thanks, let's go with that one then!

It looked something like this:

Screenshot from 2022-04-22 10-25-18

Vigasaurus:
Cool - so just to validate, this version looks good for everyone?

https://dev.vigneshjoglekar.com/plausible.io

metmarkosaric:
I get 404 on your dashboard

Vigasaurus:
I forgot to make it public - should be good now

metmarkosaric:
thanks! looks fine to me

jpomykala:
I found a bug. Last 7 days + visit duration. Now, I click on Unique Visitors, and change time to Last 30 days. Now, I got highlighted two labels. image

Vigasaurus:
@jpomykala Are you sure both are actually showing as selected and it's not just showing the hover color change for one of them? (This can happen on mobile if you press and hold on the button/top stat - it gets treated as a hover). 

You should be able to check by just tapping anywhere else in the whitespace of the page to see if it goes back to having just the actual one highlighted. 

I don't particularly know of anything that could cause such an issue outside of something like that, as all the properties are fully mutually exclusive otherwise.

jpomykala:
Hey @Vigasaurus you are right! This probably because the hover color is the same as active color. So when I change time range, it looks like that both options are active. 😄

Vrq:
I am not sure if this will be fixed with this PR but it would be nice to go back to detailed number, so not 1.2k but for example 1256.
Or maybe add rounding for bigger numbers, like 100s of thousands.

ukutaht:
@Vigasaurus this PR introduced a regression. To reproduce:

1. Scroll down on the dashboard
2. Click on 'details' in Top Sources
3. Close the popup modal
4. Expected behaviour: your previous scroll position is restored. Sticky header works.
5. Actual behaviour: browser jumps to the top of the page. Sticky header is broken until refresh.


Could you please take a look and fix?

Vigasaurus:
@ukutaht Ah yeah super weird - looks like it was caused by the `overflow: hidden` on `app.css:35` interfering with the modal - that can just be removed all together (it was only relevant to an old version of the tooltip). I can PR if you want, but it'd probably just be easy if you want to commit it.

ukutaht:
Ah sweet, thanks for letting me know! I'll commit it on master

Vrq:
> I am not sure if this will be fixed with this PR but it would be nice to go back to detailed number, so not 1.2k but for example 1256. Or maybe add rounding for bigger numbers, like 100s of thousands.

@metmarkosaric any chance to bring this back as well? :)

metmarkosaric:
I miss that one myself too. Any chance we can explore couple of designs that allow full numbers in the tooltip again @Vigasaurus?

metmarkosaric:
Another thing I noticed is that our mobile dashboard is no longer responsive. Not sure if it's related to this PR but it did start happening around the same time. Basically I can now scroll all the way on the right on our dashboard to an empty screen on my mobile. And when I click on any "details" views, they no longer show centrally on my screen.

Vigasaurus:
> Full numbers

Yeah I can look into some of those, or at least just tweaking the shorting criteria for them. 

> Broken modals & width

That is quite interesting, I notice it too (I don't use mobile enough to have seen it earlier), and I don't quite know anything in this PR that could have caused that, but it's possible. I'll take a look.

metmarkosaric:
fyi: the mobile issue seems to have been fixed. i don't see the problems on my own phone anymore

PR #1366, opened 2021-10-03
Adds support for selecting a new time interval (graph detail) in main graph

### LINKED ISSUE #282: Choose a different interval on the top graph
Right now the top graph shows the interval according to the time period selected in the date picker. Here's an example:

Screenshot from 2020-08-12 16-49-58

* If "Today" selected, the graph shows hourly intervals 
* If "Last 30 days" selected, it shows daily intervals
* If "Last 6 months" selected, it shows monthly intervals

Would be great with the option to manually select a different interval. For instance, I'd love to be able to select "weekly" interval for the graph when I'm looking on a longer term time period such as "last 3 months" or "last 6 months".

We could make this completely flexible and allow users to select whichever interval they want:

* If "Today" selected, the graph shows hourly intervals by default but I can switch it to "Daily" if I want
* If "Last 30 days" selected, it shows daily intervals by default but I can switch it to "Weekly" or "Monthly" if I want
* If "Last 6 months" selected, it shows monthly intervals by default but I can switch it to "Daily" or "Weekly" if I want

Vigasaurus:
This is built on top of #1364 - should be merged only after it to best prevent conflicts (I'll go back through here after 1364 is merged and resolve conflicts all at once, cause I'm sure some will come up from 1364's review) - **this will definitely cause conflicts now that the comparison stuff is getting descoped and moved out**

Resolves #282 ~~partially (I can definitely add week-intervals too, but **will need some input** on what exactly the split there should be, i.e. should it just be treated like a 7-day split starting from the first day given, or always split on Sundays/Mondays etc)~~

### Changes

Adds user choice for graph detail on main graph, by selecting variable intervals for plot. 

Currently includes limits in the UI to prevent excessive detail/issues on larger data sets with the following limits of interval choice per chosen time period:
see comment

Includes UI fallback for if /main-graph returns with 500 for the DB timeout (usually if the detail level is set too high with a long custom date range, and the DB query times out)

### Tests

### Changelog

### Documentation

### Screenshots
image
image
image
image
image


metmarkosaric:
thanks @Vigasaurus! i don't think we need hourly, we keep it only on daily charts. having hourly on a monthly chart is not useful even when not considering the resource usage. weekly is definitely a key one to have. something like we had in #282:

- If "Today" selected, the graph shows hourly intervals by default but I can switch it to "Daily" if I want
- If "Last 7 days" selected, the graph shows daily intervals by default but I can switch it to "Weekly" if I want
- If "Last 30 days" selected, it shows daily intervals by default but I can switch it to "Weekly" or "Monthly" if I want
- If "This month" selected, it shows daily intervals by default but I can switch it to "Weekly" or "Monthly" if I want
- If "Last month" selected, it shows daily intervals by default but I can switch it to "Weekly" or "Monthly" if I want
- If "Last 6 months" selected, it shows monthly intervals by default but I can switch it to "Daily" or "Weekly" if I want
- If "Last 12 months" selected, it shows monthly intervals by default but I can switch it to "Daily" or "Weekly" if I want
- Custom range should depending on the period a person selects should allow "Daily" or "Weekly" or "Monthly"

Vigasaurus:
@metmarkosaric Sounds good - those are pretty much everything that is available right now, I'll pull hourly back to just daily, the only consideration to discuss will be what I mentioned in the PR as far as the weekly split is concerned then - 
> but I'll need some input on what exactly the split there should be, i.e. should it just be treated like a 7-day split starting from the first day given, or always split on Sundays/Mondays etc

metmarkosaric:
thanks! i think weeks are best kept from monday to sunday

Vigasaurus:
So, let's say someone selects a month view, and is on October 2021 - would we show data outside of the selected period on weekly mode? 

Where they'd see the following as data points:
- Week of 9/26 (9/26-10/2)
- Week of 10/3
- Week of 10/10
- Week of 10/17
- Week of 10/24
- Week of 10/31 (10/31-11/6)

or the alternative, where it also slices based on the period selected - so in the above example the data points would be:
- Week of 10/1 (10/1 -10/2)
- Week of 10/3
- Week of 10/10
- Week of 10/17
- Week of 10/24
- Week of 10/31 (10/31 only)

I'd imagine the latter - but lmk what you think/what makes most intuitive sense

metmarkosaric:
> One thing though, in that example it brings up setting a Today view to `daily` or a Last month view to `monthly` - both of which make little sense, no? They'd just be one data point

true. we can probably simplify it and remove unnecessary options. i just went through all the time periods and this sounds like best for me:

Today is fine as it is
Last 7 days is fine as it is
Last 30 days needs "weekly" only
Month to date needs "weekly" only
Last month needs "weekly" only
Last 6 months needs "weekly" and "daily" 
Last 12 months needs "weekly" and "daily" 
Custom range needs "weekly" and "monthly"

what do you think?

metmarkosaric:
>     * Week of 10/1 (10/1 -10/2)
> 
>     * Week of 10/3
> 
>     * Week of 10/10
> 
>     * Week of 10/17
> 
>     * Week of 10/24
> 
>     * Week of 10/31 (10/31 only)
> 
> 
> I'd imagine the latter - but lmk what you think/what makes most intuitive sense

i say the latter too

Vigasaurus:
> true. we can probably simplify it and remove unnecessary options. i just went through all the time periods and this sounds like best for me:

Eh I think those are good, but at the same time, giving daily as an option for the longer ones definitely doesn't hurt imo (and it can make selecting specific days for drill-down easier than the two clicks & extra load it normally takes) - especially since the UI for selection is pretty minimal already. I can definitely agree that having hourly on the longer periods makes little sense (the graph becomes really hard to read), and that everything longer than 7d should have weekly as well. 

Either way - I'll work on the weekly split for now and we can reconvene once Uku's back too.

metmarkosaric:
> Eh I think those are good, but at the same time, giving daily as an option for the longer ones definitely doesn't hurt imo (and it can make selecting specific days for drill-down easier than the two clicks & extra load it normally takes) - especially since the UI for selection is pretty minimal already. I can definitely agree that having hourly on the longer periods makes little sense (the graph becomes really hard to read), and that everything longer than 7d should have weekly as well.
> 
> Either way - I'll work on the weekly split for now and we can reconvene once Uku's back too.

daily is already there now. we should not remove it. we should add the new interval as an addition to what's there now without removing anything we already do as the default

metmarkosaric:
sounds good! 🤣

Vigasaurus:
So apparently I did some silly force push 3 months ago and broke most of this PR - but I'd have to rebase it on top of the newly updated #1364 anyways (given the heavy merge conflicts they'd have) - so I'll close this for now + remake it following the merge of #1364 (then focus on getting comparisons as we'd discussed ready to go).

ukutaht:
Makes sense, thanks!

PR #1374, opened 2021-10-05
Fix ESLint errors

maco:
### Changes

1. Remove the "airbnb" eslint plugin since it conflicts with prettier and so was just annoying
2. Get rid of all existing ESLint errors.
2a. Turned off `react/display-name` because I couldn't figure out how to make that wrapped component have a display name. If anyone can figure it out, that'd be great, because that makes things nicer when using the React debugger.
2b. The part where it says `plausible()` is undefined in `app.js` is bothering me. I disabled the check because I can't figure out where that actually comes from to put in the proper import.
2c. Told ESLint we're using Babel.
3. Added `npm run format` and `npm run lint` commands.

This is a good stopping point for a PR because right now ESLint is happy. I intend future PRs to enable more ESLint rules and make things comply with them, but for now at least there won't be pre-commit stuff getting in people's way. Sorry for taking so long on this.

I should also note that Prettier is currently not enabled for any files. The plan there is to enable it for a file, fix up that file, and put the Prettier pragma at the top of it so that that file would then have style enforced on it. One file at a time, so it's not a ton to do at once.

### Tests

### Changelog

### Documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,6 @@ repos:
         files: "assets/js"
         additional_dependencies:
           - eslint@7.2.0
-          - eslint-config-airbnb@18.2.0
           - eslint-plugin-import@2.22.1
           - eslint-plugin-jsx-a11y@6.4.1
           - eslint-plugin-react@7.21.5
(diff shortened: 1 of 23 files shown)


ukutaht:
I've added linting and a format check to our CI runs as well to make sure we don't break it going forward: https://github.com/plausible/analytics/runs/3868554597

PR #1387, opened 2021-10-12
Start putting together means to export all aggregate data

m-col:
This commit adds two things that are first steps in adding a feature to
download all aggregate data:

1. A button to the dashboard to request the download, and
2. internal code to handle the request package CSV data into a zip file.

--

There are a few things to discuss:

- UI change: should we keep the current button on the top graph as is, and introduce a new button elsewhere to download the zip file? If so, where? The PR as is adds a new button at the top beside the site selector.
- Currently the zip file only contents the CSV of the top graph. What other CSV files could be added? Essentially the values plotted in the other 4 graphs?
- How should we package data with respect to dates? Currenty `30d` is hard-coded for demonstration. Am I correct in thinking that 'days' is the highest resolution, and therefore perhaps we can set the request to any number of days and package those?

Related discussion: https://github.com/plausible/analytics/discussions/1385

(Tests, docs and changelog entry TODO following discussion.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Add Conversion Rate to Top Sources, Top Pages Devices, Countries when filtered by a goal plausible/analytics#1299
 - Add list view for countries report in dashboard plausible/analytics#1381
 - Add ability to view more than 100 custom goal properties plausible/analytics#1353
+- Data exported via the download button will contain CSV data for all visible graps in a zip file.
 
 ### Fixed
 - Fix weekly report time range plausible/analytics#951


metmarkosaric:
thanks @m-col! (note that i cannot read any of the code so this is just me writing without any code review...)

couldn't we just keep the old button but make it download the new zip file instead? we don't need the current CSV download feature after this new full download feature is introduced. it could completely replace it. we plan to introduce some other icons in there for new features next to the download one in the future (such as the option to change the interval on the chart) as that location is nice 

ideally every report would be included in the export. so we would have the list of sources and their stats (including UTM medium, source, campaign), the list of pages (including entry and exit pages), countries (we're working on introducing cities and regions so these should be there too), devices, browsers (including browser versions), OS (including OS versions) and goals ([including custom properties](https://plausible.io/docs/custom-event-goals#using-custom-props)). that's pretty much every report you see in the dashboard. if this all is too much/heavy/resource intensive we could split them up and then after clicking the download icon it gives you a list as a menu and you have to pick what report to download individually (sources, pages etc)

i think it makes sense for it all to be based on days and people that want can then manually add up days together when they do their analysis in excel or whatever. another note is that the export should be based on the date range selected in the top right. so if i select the period of last month and click on download then it gives me the export of data from last month rather than all the data since day 1

m-col:
OK, makes sense. It wouldn't be much extra work for a user to merge them too, if that is what they wanted.

>  does any of this make a difference in terms of the development work?

Nope so all good

metmarkosaric:
@m-col @ukutaht what's worth considering at this early stage is the CSV download of the filtered view. if i filter by page or whatever other metric, when i click on the download icon should i be able to get the export of that filtered view only? the way it works now is that any filters are ignored and i get the download of the normal unfiltered dashboard. would the download of filtered view add much extra complexity to this project? more info in #1050

metmarkosaric:
> Yeah my expectation would be that the filters are applied when running the export.
> 
> From the development perspective I don't think it adds too much complexity. We have everything in place for that in the backend. You can `Query.parse` and get all the filters and stuff for free without any extra work.

that's great news! supporting downloads of filtered views would be perfect. basically you download/get what you see in the dashboard

m-col:
I've restored the way the top graph data is saved so it is unchanged from the current behaviour. This will need updating when #1366 is merged.

> The only thing I would explore a bit is using Accept header together with get_format/1 instead of the csv=true parameter to control whether to render JSON or CSV. Content type negotiation is built into HTTP so there shouldn't be any need to create a non-standard query parameter for that. This is all for the functions in PlausibleWeb.Api.StatsController.

For posterity, we discussed this privately and decided that the parameter is better due it its ease; we'd have to hack into the js to specify headers to be passed along to the stats controller.

> Conversions should be included in the csv export if the site has goals configured.

I've added this into the zip, and this seems to work fine when there are goals. If there are no goals configured, the fetched data will be empty, and so we will see a CSV with just headers, I think? Interestingly I am unable to delete a goal locally to test. Perhaps a bug?

m-col:
And now fixed the test failures. There are 2 failures that I am getting locally but these appear to be also present on master: `Plausible.Workers.CleanInvitationsTest` and `PlausibleWeb.Api.StatsController.MainGraphTest`.

m-col:
Nice, that totally makes sense! I've updated the tests accordingly. There are now three tests for the exports: test export with default settings, test with shared link, and test asking for the past 6 months of data. All dates are hard-coded (both in the data and in the request) so I don't think there should be any more problems with tests just because time as passed.

Issue #136, opened 2020-06-22
Filter Page URLs


metmarkosaric:
"I have some pages that contain UUIDs that are specific to users. I don't care which date or UUID is called and want to aggregate them.

Example:
Before: /account/550e8400-e29b-11d4-a716-446655440000
After: /account/{uuid}

A simple search and replace functionality with regex would be great!"

([imported from the old roadmap](https://plausible.nolt.io/95))

pelme:
This would be helpful from a privacy *and* analytics perspective for us.

marcobusemann:
Same here. Would be nice to specify multiple patterns to aggregate page URLs. Currently having dynamic URL parts creates a mess in the statistic.

tlghnb:
It is indeed creating a mess and completely unnecessary to show stats for those URLs. Would love to see this feature added anytime soon!

PR #1398, opened 2021-10-20
[WIP] Adds 'custom-location' script extension (in response to #136)

EmilePerron:
### Changes

This PR gives users the possibility of specifying a custom tracking location for their pages via a new `custom-location` script extension.

The main advantage of this addition is that it allows users to aggregate and/or redact URLs to improve the privacy and usefulness of their analytics. This helps to resolve the issues mentioned in #136.

As discussed with @ukutaht in #136, the implementation of this feature works as follows:
  
> If you want to provide a plain value (for ex. using the templating language from a server-generated page) you can use the standard configuration syntax:
> ```html
> <script data-domain="my-domain.com" data-location="https://my-domain.com/my-custom-location" src="https://plausible.io/js/plausible.custom-location.js"></script>
> ```
> 
> And if you want to generate it client-side:
> ```html
><script>window.myLocationGetter = function() { return location.href + 'foo';  }</script>
><script data-domain="my-domain.com" data-get-location="myLocationGetter" src="https://plausible.io/js/plausible.custom-location.js"></script>
> ```
> The contract would be that data-get-location will provide the name of the function and if it exists of window, it will get executed and the return value used as the location value.

### Tests

### Changelog

### Documentation

The documentation PR is here: https://github.com/plausible/docs/pull/136


EmilePerron:
I implemented the basic feature and tested it locally, and it works fine. 

I'll be adding the documentation and changelog for this soon, 

However, I do have a few things I'd like to hear your thoughts about:

**How would you prefer to have this interact with the `exclusions` extension?**
My thoughts is that the exclusions should still be checking the page's regular location (as it currently does), not its custom location (if one is provided). My thought process behind this is that checking for exclusion on the custom location would require usage of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL), which:
- would add weight to the script;
- would prevent users from using characters that are not necessarily URL friendly, but that are likely to be used, such as curly braces (ex.: `https://dummy.site/project/{id}/settings`);
- would require an alternative when combined with `compat`, which would make the script much heavier (and likely more complex).
Plus, it could require an additional change to make for users who already have exclusions set up, and who would like to add custom locations (if the new custom locations don't match their existing URL exclusion patterns).

**Are there tests for the tracker?**
I looked around and couldn't find any tests for the tracking scripts, so I'm wondering if I should write tests for this or not.

I'll wait for your thoughts on these two things before I complete this.

Cheers!

EmilePerron:
I added the changelog entry as well as the documentation for this new feature (in the following draft PR https://github.com/plausible/docs/pull/136).

The only thing left would be to define how this should interact with `exclusions`, as mentioned in my previous comment.

@ukutaht any thoughts?

ukutaht:
@EmilePerron I did just have an idea how to override the URL on the client side. I'm not sure how I didn't think of this before. I recently added a `manual` script extension that doesn't trigger pageviews automatically. So all we need is a way for the `plausible` function to accept a parameter to override the URL like so:

```html
<script defer data-domain="yourdomain.com" src="https://plausible.io/js/plausible.manual.js"></script>
<script>
// define the `plausible` function to manually trigger events
window.plausible = window.plausible || function() { (window.plausible.q = window.plausible.q || []).push(arguments) }

// Trigger pageview with a custom location
plausible('pageview'. {u: '/my/custom/location'})

// Alternatively, use a function
plausible('pageview'. {u: getMyLocation()})
</script>
```

After writing this, it seems like the obvious way to accomplish a custom url. What do you think?

We could start with this and see what people think. It does require development skill but adding something more streamlined on top of this would not invalidate being able to override the url of an event. It's just a very basic and natural feature so it doesn't become an ultimatum between the two approaches we've discussed so far.

ukutaht:
> Should I still get to work on the data-url-patterns? Or would you prefer simply documenting this URL overriding option with the manual extension and waiting to see if there's still a demand for the simpler option afterwards?

Overriding the url variable is not currently possible so you if you want to help you can implement and document that :) Otherwise I'm not sure where I can fit it into on my todo list.

I don't think we need to start with the patterns yet. Let's see if there's demand once we explain this option,

PR #1407, opened 2021-10-22
Add integration with Google Analytics to display historic data in the dashboard main graph

m-col:
### Changes

This starts the work to allow users to fetch Google Analytics data from their google accounts for displaying in their dashboard. The feature is not currently complete so I am posting for feedback on what has been implemented.

What has been implemented so far (corresponding to the 5 commits):

1. Update the settings UI so the 'Search Console' page is now a more general 'Google Integration' page, with the top panel for connecting to the Google account, and two following panels for search console integration and analytics integration respectively.
2. Add boolean flags to the `GoogleAuth` configuration of each `Site` to represent which of these two features are enabled for that site. A migration is needed to update existing sites which has set all search consoles to enabled and GA integration to disabled.
3. Update the UI to wrap the two feature panels in slider toggle buttons which enable/disable these features.
4. Update the `main-graph` API so that it returns additional data for sites with GA integration enabled. This is then plotted on the main graph in red.
5. Make the request to Google's API so that the data forward to the main graph corresponds to the same data range etc as the plotted Plausible data. This last commit is what I am currently working on and so is unfinished. Still figuring out Google's API for this!

Here are some pics:

The graph (placeholder values are just the Plausible data halved):
image

Settings panel - both features can be enabled individually:
image

Search console settings required it to be enabled first:
image

### Tests

### Changelog

### Documentation


m-col:
As discussed, I've updated the settings page to display a dropdown showing the user's google analyics profiles:
image

It's now more flexible (they can select a profile with a different host domain if required) and less prone to cause problem based on the assumption that that was the case, and that users have only one profile. Now all they need to do is select a profile that looks the most obvious i.e. their host domain of interest, and 'All Web Site Data' for access permissions.

m-col:
The test failure is the same test fixed in my other PR.

metmarkosaric:
> As discussed, I've updated the settings page to display a dropdown showing the user's google analyics profiles:

Looks nice! Much easier to use this way. Thanks!

m-col:
There are some updates I need to make in line with some of the new features, but in its current form it is working 100%!

I'm first going to rebase and make those updates but will then be focussing on getting tests written for all of the changes. Any review/comments on its current form would be appreciated!

m-col:
I posted that on the wrong PR!

ukutaht:
@m-col guessing this one can be closed?

m-col:
Yep!

PR #1441, opened 2021-11-03
Don't enforce longer SECRET_KEY_BASE than Phoenix

lorenz:
Phoenix only requires SECRET_KEY_BASE to be longer than 32 characters. 32 characters even with hex encoding is 128 bits of entropy which is more than enough. Some secret generation tools only generate 32 characters/128 bits by default which makes this relatively arbitrary limit annoying. Let's change it to what Phoenix requires.

### Tests

### Changelog

### Documentation


ukutaht:
Seems good to me

Issue #190, opened 2020-06-28
use beacon api?


scottjg:
i love how small the plausible script is, so i'm a little hesistant to suggest more stuff in it, but it would be nice to use the beacon api instead of xhr. that way, you can shoot off events "in the background", as opposed to having them block the page nav. browser support for this is ok but not supported on IE11 which maybe is a dealbreaker for some of your users. in our case, it isn't though.

i guess in theory you could have different scripts served based on the user agent, though i can see why that would be a hassle to maintain.

oliverpool:
> i'm a little hesistant to suggest more stuff in it

I think that the addition can be quite small, and restricted to the `registerOutboundLinkEvents` mode.


> not supported on IE11

One can make a simple `if(window.navigator.sendBeacon)` to detect its support and fallback to the current "delay method".

---

I would be willing to make a PR for the tracking script if the maintainers are ok with it.

_One thing to be aware, is that `sendBeacon` sends a POST request and I don't know if the backend supports it yet._

PR #1446, opened 2021-11-04
More improvements for CSV export

m-col:
This rounds the `time_on_page` metric returned as part of the `pages`
API to the nearest second. While this shows no apparent change in the
web UI, it removes the decimal from the exported data in `pages.csv`.

### Tests

### Changelog

### Documentation

### To do list

To do:

- [x] the total column for goals is empty (fixed in previous PR).
- [x] add test for goals in conversions.csv
- [x] comment above - round in database?
- [x] see if it's possible to enable spinner upon request + disable upon download ready
    - https://stackoverflow.com/a/4168965


m-col:
Actually it would be nice to add a time_on_page entry to the test data and check that in a test - I'll take a look to see if it's possible

m-col:
@ukutaht perhaps you could give me a pointer on this, if you agree it's worth including in the test. The `time_on_page` is calculated on the fly rather than stored: https://github.com/plausible/analytics/blob/master/lib/plausible/stats/breakdown.ex#L228-L249. Consequently the `populate_stats` test helper can't simply be passed data with `time_on_page` specified. My SQL is pretty weak - any ideas on how I could make the test data output a `time_on_page`? Does it need to be given a page exit time?

lib/plausible/stats/breakdown.ex
```diff
@@ -248,7 +248,10 @@ defmodule Plausible.Stats.Breakdown do
       GROUP BY p"
 
     {:ok, res} = ClickhouseRepo.query(time_query, base_query_raw_params ++ [pages])
-    res.rows |> Enum.map(fn [page, time] -> {page, time} end) |> Enum.into(%{})
+
+    res.rows
```
ukutaht, line 252: I think it would make more sense to do the rounding in the database instead of the app server

ukutaht:
> My SQL is pretty weak - any ideas on how I could make the test data output a time_on_page? 

Time on page is calculated based on the difference in timestamps of the `pageview` events. You can get it to be returned from the API by providing timestamps for the test data. See https://github.com/plausible/analytics/blob/master/test/plausible_web/controllers/api/stats_controller/pages_test.exs#L28

lib/plausible/stats/breakdown.ex
m-col: That sounds nice, but could it have unintended side effects? The main potential problem I'm thinking is rounding causing some pageviews to be 0 s, when it would make sense for pagesviews to only ever be 1 second or more, right?

m-col:
Great thanks. In the case where a visitor views a single page, is there then no 'time on page' value for it?

m-col:
Got it from the test - it's `nil` when it is 0, and the exit page isn't included in that report.

metmarkosaric:
> Great thanks. In the case where a visitor views a single page, is there then no 'time on page' value for it?

Time on page only shows people who visit more than one page. This is similar to how GA does it. We plan to move to a different method to make it more accurate in the future. See #190

lib/plausible/stats/breakdown.ex
ukutaht: > The main potential problem I'm thinking is rounding causing some pageviews to be 0 s, when it would make sense > for pagesviews to only ever be 1 second or more


Hmm I don't understand the concern here. If there's a result with <0.5seconds then it will be shown as zero, but the same will happen if the rounding is done in the app server. Pushing this average calculation down to the database should not change the behavior at all.

m-col:
I'll have to rebase as the changes I made to the tests here probably conflict with #1447. I've made them much nicer - the test data is stored as CSVs instead of hard-coded, making it much easier to read and modify the tests.

lib/plausible/stats/breakdown.ex
m-col: If I've understood the calculation correctly, as 

> Time on page is calculated based on the difference in timestamps of the pageview events

then timestamps of 3.4 and 3.6 would round to 3 and 4 in the database, and would become a visit duration of 1 instead of 0. The inverse can also happen e,g, with 3.6 and 4.4 where rounding in the database changes the 1s to 0s.

Perhaps this is splitting hairs though... :smile:

ukutaht: Sorry I might have not been completely clear - by 'in the database' I didn't mean that the timestamps on pageviews would be rounded as they're stored on disk. I meant that the rounding can be applied in the database query for time on page. It would probably mean applying the `round` function to the result of this line: https://github.com/plausible/analytics/blob/master/lib/plausible/stats/breakdown.ex#L233

By rounding just the result of the `avg` calculation, it should not skew any data.

m-col:
I'm working on getting the spinner to be visible only for the duration of the download request, and then remove it when the download is ready.

The approach I'm trying to use, inspired by the stackoverflow link in the top post above, involves the JS simply polling every 1s and then undoing the spinner when a cookie appears. JS diff here:

```diff
diff --git a/assets/js/dashboard/stats/visitor-graph.js b/assets/js/dashboard/stats/visitor-graph.js
index f83a4a4..1b5c17c 100644
--- a/assets/js/dashboard/stats/visitor-graph.js
+++ b/assets/js/dashboard/stats/visitor-graph.js
@@ -304,12 +304,17 @@ class LineGraph extends React.Component {
     return stats
   }
 
+  pollExportReady() {
+    if (document.cookie.indexOf('export_ready') > -1 ) {
+      this.setState({exported: false})
+    } else {
+      setTimeout(this.pollExportReady.bind(this), 1000);
+    }
+  }
+
   downloadSpinner() {
     this.setState({exported: true});
-    setTimeout(
-      () => this.setState({exported: false}),
-      4000
-    );
+    setTimeout(this.pollExportReady.bind(this), 1000);
   }
 
   downloadLink() {
```

This is simple enough but the elixir code doesn't quite work. The current change is only:
```diff
diff --git a/lib/plausible_web/controllers/stats_controller.ex b/lib/plausible_web/controllers/stats_controller.ex
index 42beaf8..96c9a29 100644
--- a/lib/plausible_web/controllers/stats_controller.ex
+++ b/lib/plausible_web/controllers/stats_controller.ex
@@ -93,6 +93,7 @@ defmodule PlausibleWeb.StatsController do
     conn
     |> put_resp_content_type("application/zip")
     |> put_resp_header("content-disposition", "attachment; filename=\"#{filename}\"")
+    |> put_resp_cookie("export_ready", "true")
     |> send_resp(200, zip_content)
   end
 ```
However the cookie is never created on the page. I'm thinking this might be because the page is not redirected, and so the cookie needs to be applied to current page which already exists.

Anything here jump out as not quite right?

m-col:
I've rebased and completed the other fixes + tests so this is at a good state.

In the meantime I'll be continuing on the spinner thing.

m-col:
Got the spinner timeout working! The spinner is now only displayed until the download is ready, so this PR is ready for review.

ukutaht:
Looks good to me!

metmarkosaric:
Everything looks/works great now from what we discussed earlier! 

There's one last thing we need to fix before this can go live to the public: Download gets stuck and nothing happens if you filter by a goal or a custom property and click on the download button.

We should make it work so that if you filter by a goal/custom property and click on the download, the export includes exactly what you see on the dashboard. For instance if I filter by 404 error pages or external link clicks, we export a list of those etc

PR #1456, opened 2021-11-09
Fix custom property total conversions value not displayed & Export custom properties

m-col:
The custom property conversion metrics are not consistent with the other
metrics resulting in the total conversions not being displayed in the
dashboard. This fixes that.

### Tests

### Changelog

### Documentation


m-col:
This is looking good now

test/plausible_web/controllers/CSVs/30d-filtered/prop_breakdown.csv
```diff
@@ -0,0 +1 @@
+prop,name,unique_conversions,total_conversions
```
ukutaht, line 1: I think we should add some data to test that the breakdown is working as expected here as well

m-col: I've added a commit attempting to do this but the test exported data doesn't seem to match the populated data. It's possible the test is fine but the lib code is wrong, but I suspect I've not setup the test correctly. Could you possibly take a look at the test to confirm the populated data and the filter make sense for this?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -611,13 +611,49 @@ defmodule PlausibleWeb.Api.StatsController do
...
+    site = conn.assigns[:site]
+    query = Query.from(site.timezone, params) |> Filters.add_prefix()
+
+    headers = ["prop", "name", "unique_conversions", "total_conversions"]
+
+    prop_names =
+      if query.filters["event:goal"] do
+        {_, _, goal} = query.filters["event:goal"]
+
+        Stats.props(site, query)
```
m-col, line 642: It seems like this is returning `%{}` during the test

m-col: Looks like this is due to the time period being queried - if `period=day` then we get the props, otherwise we do not.

m-col: OK, think I got it, the timestamps were not setup correctly.

m-col:
Improved test with goal + prop in CSV added.

ukutaht:
Sorry I force pushed to master to clean up some sensitive information that should not have been there. Looks like it messed up this PR a bit as well. Just formatting needs to be fixed and then I think it's ready to go

Issue #1272, opened 2021-08-27
Mobile issues (realtime dashboard, favicon etc)


metmarkosaric:
* A way for anonymous/logged-out viewers to see domain name on the embed/public mobile version. right now they only see the favicon and would be nice by tapping the favicon they can see the full domain name too

* if i'm logged in and tap on the favicon to display the full domain / other sites i have, i cannot tap the favicon again to close that little menu. i must tap on another site or somewhere else on the display to close the menu. would be nice if i can tap the favicon again

* The text headings for unique visitors and pageviews in "realtime" view overflow

![index4](https://user-images.githubusercontent.com/34340819/131127126-1a6cb65a-d636-4747-b79f-3ad2af365ef6.jpg)

* footer menus need to be aligned properly

![index3](https://user-images.githubusercontent.com/34340819/131127172-4e3a3b56-73f9-4f96-b532-c1dd57486081.jpg)

* top pages:

![top-pages](https://user-images.githubusercontent.com/34340819/131127061-bb3be0c7-67c6-48aa-ba58-ad48054cf7e3.png)

Vigasaurus:
The issues with the top stat headers I have fixed in my graph overhaul, should be coming soonish.

metmarkosaric:
Another little issue is on the home page for non-logged in users. the "2 months free" overflows so should be placed either above or below "yearly billing"

![Screenshot_20210901-141848_Firefox.jpg](https://user-images.githubusercontent.com/34340819/131670154-6b14e7ab-1f4e-49ce-9ce5-3a2c7fe9763a.jpg)

hirusi:
@metmarkosaric - got it. think I'll start tackling this new stuff this weekend/early next week.

hirusi:
Also:

https://user-images.githubusercontent.com/3974019/133571822-6e448428-fd28-4cba-9624-bc962dd66f40.mp4

Not a "mobile issue" though.

hirusi:
> Another little issue is on the home page for non-logged in users. the "2 months free" overflows so should be placed either above or below "yearly billing"

Does this look good?

![Screen Shot 2021-09-28 at 11 20 39](https://user-images.githubusercontent.com/3974019/135030730-e591df85-8825-4c0b-8fdc-5d18f5df403c.png)

If so, I'll finish this tomorrow (tablet+desktop adjustments) and open a PR.

hirusi:
> The text headings for unique visitors and pageviews in "realtime" view overflow

[...]

PR #1498, opened 2021-11-28
UI fix: footer alignment on mobile

hirusi:
### Changes

This fixes the issue raised in https://github.com/plausible/analytics/issues/1272, specifically:

> footer menus need to be aligned properly

I have opted for a bit of a hacky approach as far as the alignment for the "Comparisons" column goes (`mt-32`). If that is unwarranted, let me know please.

#### Tested with

- Firefox 94
- Edge 96
- Chromium 96

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 - Region and city-level geolocation plausible/analytics#1449
 - The `u` option can now be used in the `manual` extension to specify a URL when triggering events.
 
+### Fixed
+
+- UI fix to align footer columns
+
 ## v1.4.1
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ukutaht:
This footer is duplicated in the website repo so it should be fixed there as well

hirusi:
Ahh, you're right. On it.

hirusi:
Sorry @ukutaht, I don't see this on the landing website?

image

I did refer to that implementation to get this correct here, and it seemed OK.

hirusi:
The problem happened primarily because of a difference in the number of list items for the first block ("Why Plausible?").

metmarkosaric:
Ahh I see. I updated the one on the home page as that's the one I knew how to update. I didn't get to update the one in app. The footer on the home page is the correct one and the one we should use in app too. Perhaps you could simply replicate what we have on the home page in app too?

hirusi:
That's exactly what I've done. :)

With the exception of a small change, that is, different top margin, since one column is shorter here compared to the home page.

I'll share some screenshots later to explain more clearly.

ukutaht:
Ah OK then it's all good

PR #1500, opened 2021-11-28
Site switcher toggle functionality

hirusi:
### Changes

This tackles an issue raised in https://github.com/plausible/analytics/issues/1272#issue-981200594, specifically:

> if i'm logged in and tap on the favicon to display the full domain / other sites i have, i cannot tap the favicon again to close that little menu. i must tap on another site or somewhere else on the display to close the menu. would be nice if i can tap the favicon again.

More details in https://github.com/plausible/analytics/issues/1272#issuecomment-981074054

### Tests

### Changelog

### Documentation


ukutaht:
The `handleClick` handler's job is to close the dropdown when it's open and you click somewhere outside of it on the page. This change seems to break that behaviour. I think we should try to figure out a way to preserve it

hirusi:
Ahh, that makes sense now. I'll take a look.

ukutaht:
Thanks!

Issue #149, opened 2020-06-22
User flows


metmarkosaric:
"One of the main reasons for me to want Analytics is to see the users flow inside the Website itself, so that I can tailor it to the way the users are actually using it. The way I did it previously (when I had Google Analytics) was to have a Sankey Diagram which showed me how many users were flowing from one page to another. Is this something supported by Plausible or planned to be supported at some point?

Details:
For this to be useful for larger websites, I will describe how I would imagine it for a Blog (The web’s standard Hello World of websites :P )

The user would define custom page types. In this case, it would be:
HomePage: Has it’s own type
About: Has it’s own type
Posts: All posts share this type.

Then the Analytics allows me to select a page type (Say HomePage) and see how users flow from there (maybe 90% go to a Posts type page, and 10% to the About page).

Does this make sense?"

![ldud4xldqpwft8kfaovh](https://user-images.githubusercontent.com/34340819/85281041-9f0a6800-b489-11ea-960e-6fd25c00b748.png)

([imported from the old roadmap](https://plausible.nolt.io/109))

bradonomics:
This type of user flow data would be a prerequisite for moving people off Google Analytics who wanted to use predictive prefetching libraries like [Guess.js](https://guess-js.github.io/).

PR #1506, opened 2021-12-03
feature/manage-goals-api-routes

abbaseya:
### Changes

- API route `PUT /api/v1/sites/goals` with form params `site_id`, `goal_type` and `goal_value` with supported types `event` and `page`
- API route `PUT /api/v1/sites/goals/:goal_id` with form params `site_id`, `goal_type` and `goal_value` with supported types `event` and `page`
- API route `DELETE /api/v1/sites/goals/:goal_id` with form param `site_id`

Documentation changes at PR [#149](https://github.com/plausible/docs/pull/149)

### Tests

### Changelog

### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- API route `PUT /api/v1/sites/goals` with form params `site_id`, `goal_type` and `goal_value` with supported types `event` and `page`
+- API route `DELETE /api/v1/sites/goals/:goal_id` with form params `site_id`
+
+### Added
 - Data exported via the download button will contain CSV data for all visible graps in a zip file.
 - Region and city-level geolocation plausible/analytics#1449
 - The `u` option can now be used in the `manual` extension to specify a URL when triggering events.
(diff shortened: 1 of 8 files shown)


ukutaht:
Thanks @abbaseya!

Definitely a worthwhile addition to the API. I'm happy to merge but the following points need to be addressed first:
* Automated tests
* Documentation for the endpoint
* Changelog entry

abbaseya:
@ukutaht Thank you for your response. Changes pushed.

ukutaht:
Thanks for the updates


What's the purpose of the `update` endpoint?

Personally I think it would be better to just use `delete` and `find_or_create` a new one if you want to change how a goal behaves. The update endpoint can also generate confusion. What happens if I take an existing custom even goal and pass update it with `{goal_type: 'page', goal_value: '...'}`?

The second point I wanted to bring up is that I think the `goal_value` parameter should be renamed to what it actually represents (given the different goal types). So to create a goal I would prefer this:

```json
{"goal_type": "event", "event_name": "Signup"}
{"goal_type": "page", "page_url": "/blabla"}
```

instead of

```json
{"goal_type": "event", "goal_value": "Signup"}
{"goal_type": "page", "goal_value": "/blabla"}
```

abbaseya:
@ukutaht Well, your note on omitting the update endpoint actually makes more sense. Thanks for pointing that out.

As for the shape of the response. Just pushed an update that expects `event_name` or `page_path` instead of `goal_value` (_based on the `goal_type`_)

ukutaht:
Thanks @abbaseya, looking good!

It just needed to be formatted which I've done in https://github.com/plausible/analytics/pull/1527. The code is now merged. I will deploy it onto our staging server for some testing and then it will be available on our hosted service next week.

PR #1393, opened 2021-10-17
Add `utm_content` and `utm_term`.

BlenderDefender:
Support `utm_content` and `utm_term` as requested in #515.

### Changes
I have added support for `utm_content` and `utm_term` and made a separate tab for these parameters. It works as far as I've tested it, but I honestly don't know if I did something wrong, because it's my first time scripting with Elixir. One improvement that could/should be made for this is splitting the `utm_term` across some separator like a comma or a semicolon:
How it looks right now:
image
How it should look:
image

### Tests
- [x] This PR does not require tests

### Changelog
- [x] Entry has been added to the changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated

PR #1542, opened 2021-12-16
PR 1393 continued

ukutaht:
### Changes

Follows up on the work done in https://github.com/plausible/analytics/pull/1393. This PR adds a dropdown for UTM reports.


FarisZR:
In the self hosted version, there is still no way to select utm_content and utm_term
image

PR #1574, opened 2021-12-31
Adds ability to select graph detail (interval)

Vigasaurus:
### Changes

Recreation of #1366 rebased on top of the newly updated #1364 with several improvements & fixes (again, this should only be merged after #1364)

- Performs the same as #1366 did at the end of discussions
- Now only reloads the graph view on changing interval as opposed to entire page
- More cleanly loads and unloads graph views (keeps top stats in view unless they are changing)
- Stores & maintains interval at the graph-level instead of polluting the query object with value only used by one report

I could potentially see some benefit here in keeping the interval prop in the query schema to allow for sharing a link with the interval pre-selected, but I figured the better UX of not reloading the entire page on updating the interval was a better tradeoff. 

### Tests

### Changelog

### Documentation

### Dark mode

### Screenshots

image
image
image
image


Vigasaurus:
This is mostly up to date and functional, I just need to update the interactions with the new periods available (all time & YTD), which I'll add in soon.

ukutaht:
Nice. No rush, I still have some work to do on the main graph selections with imported data.

Vigasaurus:
> main graph selections with imported data

I can't say I'm super surprised there were some issues with that heh - I haven't been able to get the import fully working on my self hosted box nor my dev environment yet to test that out fully. The issue on my self hosted box is a bit weird honestly - the oban job for the import has just been stuck on available for a while 🤷‍♂️

Vigasaurus:
Alright @ukutaht - this should be all good to go now. It had some issues with the imported data handling of the new week interval (naturally), and needed updating for the new periods on the FE which should both be sorted. I did also end up finding and fixing various UI/UX edge cases + got some code cleanup done. 

I must say, I'm really not happy with the default formatting that gets forced on the longer fragments in `imported.ex` and `timeseries.ex` but I guess it is what it is (maybe those strings can be pulled into a separate variable earlier in the function to not have that deep of nesting? - I'm unsure of how that works with Ecto).

lambaek:
Could it be considered to add the "graph detail" as a dropdown in the upper right corner, which is similar to what the user is already familiar with from the "Top Sources" section?

I made a quick prototype for it which can be [seen here in Figma](https://www.figma.com/proto/KZz7G1SEDUuHdVCaD08C6q/Plausible?node-id=3%3A38&scaling=min-zoom&page-id=3%3A8&starting-point-node-id=3%3A38)

With this approach, we don't confuse the user by adding a new icon that
a) They don't know what does.
b) Will most likely be hidden amongst the other icons in the existing space

https://user-images.githubusercontent.com/10897924/176861875-8131ec29-f7ae-4aff-bbea-47aedf516317.mp4

.

segments-tobias:
@Vigasaurus Any update on this? Would be really helpful since we have a strong weekly pattern in our data 🙏

vinibrsl:
Thanks for this extensive work, @Vigasaurus! We'll be reviewing it this week and start some testing in our staging environment this week.

Vigasaurus:
Sounds great! I haven't looked at this much since it was first completed, so do let me know if there are any new incompatibilities or issues with anything new you discover and I'd be happy to help resolve them (though I have no doubt y'all's expanded team can handle them now too 😀). 

I'd also be happy to pick up, continue, or hand off the partially completed comparisons work down the line too - but @ukutaht we can discuss that later if you'd like.

vinibrsl:
> Sounds great! I haven't looked at this much since it was first completed, so do let me know if there are any new incompatibilities or issues with anything new you discover and I'd be happy to help resolve them (though I have no doubt y'all's expanded team can handle them now too 😀).
> 
> I'd also be happy to pick up, continue, or hand off the partially completed comparisons work down the line too - but @ukutaht we can discuss that later if you'd like.

Thanks for your cooperation. There is one bug we couldn't figure out. When hovering the `Tooltip` component inside `TopStats` row this exception is thrown:

image

```
Uncaught TypeError: current is undefined

The above error occurred in the <Tooltip> component:
    in Tooltip (created by TopStats)
    in TopStats (created by LineGraph)
    in div (created by LineGraph)
    in div (created by FadeIn)
    in FadeIn (created by LineGraph)
    in div (created by LineGraph)
    in LineGraph (created by Context.Consumer)
    in withRouter(LineGraph) (created by VisitorGraph)
    in div (created by FadeIn)
    in FadeIn (created by VisitorGraph)
    in div (created by VisitorGraph)
    in div (created by _default)
    in _default (created by VisitorGraph)
    in VisitorGraph (created by Historical)
    in div (created by Historical)
    in Historical (created by _class)
    in _class (created by Dashboard)
    in Dashboard (created by _class)
    in _class (created by Context.Consumer)
    in withRouter(_class) (created by Router)
    in Route (created by Router)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by Router)
    in Router
    in ErrorBoundary

React will try to recreate this component tree from scratch using the error boundary you provided, ErrorBoundary.
```

We tried debugging but no luck. Could you please help us figure it out?

Vigasaurus:
Absolutely - I'll take a look today.

Vigasaurus:
@vinibrsl I was able to find it!

That was a surprisingly easy fix for something that was relatively difficult to find haha - such a small amount of a diff. 

Do let me know if anything else comes up! 

Also, I do realize in using this a bit that it opens up the door for a better UX if a week-long date picker option is also available, to allow for the same behavior as months and days in the graph (click to view) for weeks as well.

vinibrsl:
@Vigasaurus thanks for the prompt response!

I checked out your Elixir changes only, and it's compatible with the current/master version of the dashboard. This means we can ship API support for intervals without any breaking changes. I'm [co-authoring](https://mokacoding.com/blog/how-to-add-coauthors-to-a-git-commit/) any commits I make so you are credited accordingly.

Here's my plan for releasing this feature:
- [ ] Deploy the back-end with a separate PR (#2417)
- [ ] Make a separate PR with JS changes unrelated to intervals
- [ ] Re-review the original PR (that will have fewer changes due 1 and 2)
- [ ] Deploy dashboard changes under a feature flag

If it's not too much to ask, could you please help reviewing #2417?

vinibrsl:
This pull request is now ready to be reviewed, @plausible/developers. Back-end support for intervals is already in production, and it's playing perfectly with the master version of the dashboard ✨

The last step is to merge and deploy this pull request, which includes the dashboard changes to support intervals. I added a feature flag called `intervals` that shows the "Graph detail" button conditionally, so we can release gradually.

image

metmarkosaric:
wow nice work @vinibrsl! i need this enabled for my site so i can play around! :)

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ All notable changes to this project will be documented in this file.
 - New and improved Session tracking module for higher throughput and lower latency. [PR#1934](https://github.com/plausible/analytics#1934)
 - Do not display ZZ country code in countries report [PR#1934](https://github.com/plausible/analytics#2223)
 - Add fallback icon for when DDG favicon cannot be fetched [PR#2279](https://github.com/plausible/analytics#2279)
+- Add ability to select time intervals on main graph plausible/analytics#1574 (thanks to @Vigasaurus for this feature)
```
aerosol, line 72: Would it be good to mention that it's kept behind a feature flag? Or is the plan to get rid of it when the release lands?

aerosol (approved):

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -0,0 +1,134 @@
...
+
+export const INTERVAL_LABELS = {
+  'minute': 'Minutes',
+  'hour': 'Hours',
+  'date': 'Days',
+  'week': 'Weeks',
+  'month': 'Months'
+}
+
+export default class IntervalPicker extends Component {
```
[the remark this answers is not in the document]
ukutaht: Function components can be stateful or not. One pattern we've been using a lot with class components is extracting functions from the `render` function. But I think the more readable way would be to extract those into their own little stateless function components. Dirty pseudocode:

```javascript
function DropdownContent({query, graphData}) {
  // renders dropdown content
}

function IntervalPicker(props) {
  // a bunch of hooks and logic

  return (
    // jsx jsx...
      // instead of {this.renderDropdownContent()}
     <DropdownContent query={props.query} graphData={props.graphData} />
   // jsx jsx...
  )
}
```

For convenience, all props can also be passed down as well but I prefer the explicit style of only passing what's needed to render the child. Makes debugging easier.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -67,7 +69,17 @@ class LineGraph extends React.Component {
             grid: { display: false },
             ticks: {
               maxTicksLimit: 8,
-              callback: function(val, _index, _ticks) { return dateFormatter(graphData.interval)(this.getLabelForValue(val)) },
+              callback: function (val, _index, _ticks) {
+                if (graphData.interval === 'hour' && query.period !== 'day') {
+                  return `${dateFormatter("date", false, query.period)(this.getLabelForValue(val))}, ${dateFormatter(graphData.interval, false, query.period)(this.getLabelForValue(val))}`
```
RobertJoonas, line 74: I'm a bit confused why we return two formatted dates separated with a comma here. What do you think about leaving a comment about why/what this is doing?

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -0,0 +1,134 @@
...
+    super(props);
+    this.state = {
+      open: false
+    };
+
+    this.handleClick = this.handleClick.bind(this);
+    this.renderDropDownContent = this.renderDropDownContent.bind(this);
+  }
+
+  componentDidMount() {
```
ukutaht, line 37: All of this and more are implemented in Headless UI, developed by the Tailwind team: https://headlessui.com/react/menu. It's already used as a dependency in this project as well.

I think we should move as much of our hand-rolled components to use Headless UI instead because it considers keyboard navigation, focus states, tabindex, accessibility, and a bunch of other stuff that make interactive elements better.

I've started converting things that I touched in the last year or so but it's not been done for everything yet.  ` git grep '<Menu'` for examples. One can be found here: https://github.com/plausible/analytics/blob/e522a2d7c182f5a0c7b2d95bf3ea248567a529e0/assets/js/dashboard/filters.js#L274

Get familiar with everything that Headless UI has to offer: https://headlessui.com. When we work on something that looks and feels like one of the component's they've already built, we should consider using Headless UI to simplify our own code.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -67,7 +69,17 @@ class LineGraph extends React.Component {
...
+                }
+
+                if (graphData.interval === 'minute' && query.period !== 'realtime') {
```
RobertJoonas, line 77: From `INTERVAL_MAPPING` it looks like the `day` period also supports `minute` interval:

```javascript
export const INTERVAL_MAPPING = {
  'realtime': ['minute'],
  'day': ['minute', 'hour'],
  ...
}
```

Why do we still override the interval to `hour` here?

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -278,36 +297,70 @@ export default class VisitorGraph extends React.Component {
...
+      } else {
+        this.setState({ loadingTopStats: true, topStatData: null })
+      }
       this.fetchTopStatData()
     }
 
     if (metric !== prevState.metric) {
-      this.setState({loading: 1, graphData: null})
-      this.fetchGraphData()
+      this.setState({loadingMainGraph: 2}, this.validateInterval)
```
ukutaht, line 359: Could this magic number be extracted to a constant enum with names? And maybe the values should be strings that are actually readable. When using React DevTools for example, you see the component state. Seeing `"loading"` would be way better than: `2`

At some point I'd like to start introducing typescript so we can use proper Enums (https://www.typescriptlang.org/docs/handbook/enums.html). For now, a simple map would do:
```javascript
const LOADING_STATE = {
  loading: 'loading',
  success: 'success',
  error: 'error'
}
```

Or something like this. Not sure what the appropriate names are in this case.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -335,45 +388,54 @@ export default class VisitorGraph extends React.Component {
...
+    }
+
+    const url = `/api/stats/${encodeURIComponent(this.props.site.domain)}/main-graph`
+    let params = {metric: this.state.metric || 'none'}
+    if (this.state.interval) { params.interval = this.state.interval }
+
+    api.get(url, this.props.query, params)
       .then((res) => {
-        this.setState((state) => ({ loading: state.loading-2, graphData: res }))
+        this.setState({ loadingMainGraph: false, graphData: res })
```
ukutaht, line 402: Also looks like the `loadingMainGraph` key is mixing booleans with numbers. After learning some statically typed languages I find this very confusing to work with.

RobertJoonas (approved):
Looks good to me in general. Left a couple of suggestions and questions inline.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -335,45 +388,54 @@ export default class VisitorGraph extends React.Component {
...
 
     const theme = document.querySelector('html').classList.contains('dark') || false
 
-    if ((loading <= 1 && topStatData) || (topStatData && graphData)) {
-      return (
-          <LineGraphWithRouter graphData={graphData} topStatData={topStatData} site={site} query={query} darkTheme={theme} metric={metric} updateMetric={this.updateMetric} />
-      )
-    }
+    return (
+      <FadeIn show={(!loadingTopStats && (!loadingMainGraph || loadingMainGraph === 2) && (topStatData && (!metric || loadingMainGraph === 2) || !!(topStatData && graphData)))}>
```
ukutaht, line 426: This is pretty brain-breaking. I think some naming could be useful here:

```javascript
const showCondition = (!loadingTopStats && (!loadingMainGraph || loadingMainGraph === 2)
const anotherShowCondition = (topStatData && (!metric || loadingMainGraph === 2)
// etc..

const showGraph = showCondition 
  || anotherShowCondition
   || //etc...

return (
  <FadeIn show{showGraph}>
  </FadeIn>
)
```

The names are just placeholders, of course. And if it's too long, a function like `shouldShowGraph()` can be extracted.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -335,45 +388,54 @@ export default class VisitorGraph extends React.Component {
...
 
   render() {
-    const {metric, topStatData, graphData} = this.state
+    const {metric, loadingMainGraph, loadingTopStats} = this.state
 
     return (
       <LazyLoader onVisible={this.onVisible}>
         <div className={`relative w-full mt-2 bg-white rounded shadow-xl dark:bg-gray-825 transition-padding ease-in-out duration-150 ${metric ? 'main-graph' : 'top-stats-only'}`}>
-          {this.state.loading > 0 && <div className="graph-inner"><div className={`${topStatData && !graphData ? 'pt-52 sm:pt-56 md:pt-60' : metric ? 'pt-32 sm:pt-36 md:pt-48' : 'pt-16 sm:pt-14 md:pt-18 lg:pt-5'} mx-auto loading`}><div></div></div></div>}
+          {(loadingMainGraph || loadingTopStats) && <div className="graph-inner"><div className={`${loadingMainGraph === 2 ? 'pt-52 sm:pt-56 md:pt-60' : (!metric) ? 'pt-16 sm:pt-14 md:pt-18 lg:pt-5' : 'pt-32 sm:pt-36 md:pt-48'} mx-auto loading`}><div></div></div></div>}
```
ukutaht, line 438: This is crying out for using the `classNames` function: https://github.com/JedWatson/classnames. It's already imported in this file.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -278,36 +297,70 @@ export default class VisitorGraph extends React.Component {
...
     }
     this.onVisible = this.onVisible.bind(this)
     this.updateMetric = this.updateMetric.bind(this)
     this.fetchTopStatData = this.fetchTopStatData.bind(this)
     this.fetchGraphData = this.fetchGraphData.bind(this)
+    this.validateInterval = this.validateInterval.bind(this)
+    this.updateInterval = this.updateInterval.bind(this)
+  }
+
+  validateInterval() {
```
ukutaht, line 313: Can we think of a better name for this? I wouldn't expect a `validateX` function to modify the state of a component. Validations, in my mind, should not have side effects.

Maybe it's better to keep a `validateInterval` without side-effects but use it in a new `updateIfValid` function?

I'm not sure why we would need to validate a response from our own API anyways. I've tried to only return stuff that's actually valid from the API and make the frontend code optimistic. In what case does this validation fail?

assets/js/dashboard/stats/graph/graph-util.js
```diff
@@ -27,39 +29,65 @@ export const METRIC_FORMATTER = {
   'conversions': numberFormatter,
 }
 
-export const dateFormatter = (interval, longForm) => {
+export const dateFormatter = (interval, longForm, period, full) => {
```
ukutaht, line 32: I think this function should be teased apart, it's literally starting to look like the [gilded rose kata](https://github.com/emilybache/GildedRose-Refactoring-Kata/blob/main/elixir/lib/gilded_rose.ex) which is an exercise is pulling apart too many nested conditionals :)

I don't think it's in scope for this PR or for the intervals work in general. Many code improvements are already being done and I don't think we need to go down all rabbitholes we encounter along the way. 

Just a mental reminder to stop adding stuff to this in the future. If we need to change it again in the future, refactor it first.

CHANGELOG.md
ukutaht: Good point. As long as it's being tested behind a flag, it's probably not a good idea to put it in the changelog.

We could do a follow-up PR which removes the flag once it's tested, and adds it to the changelog.

ukutaht (commented):
Huge props for teasing this PR apart and making it easy to review @vinibrsl! And big thanks to @Vigasaurus for the initial implementation ❤️ This is very exciting and highly requested feature at the moment.

I usually try to keep comments to a reasonable amount but I saw so many potential improvements here. I think a big part of it is the overall messiness of the frontend code which doesn't set a good example. Many comments are about things you didn't add @vinibrsl, but I think this PR is a good opportunity to start setting better standards in the React code.

Sorry for tearing it apart, but I really want to start raising the bar in the React codebase. I think the [broken window theory](https://medium.com/@learnstuff.io/broken-window-theory-in-software-development-bef627a1ce99) applies here. I feel bad about leaving a legacy like this that needs to be cleaned up 😕

About getting to approve this PR. As I said in the last demo, I don't trust anything other than manual tests for UI stuff. So I'd need to be able to play around with this PR on staging in order to approve it.

assets/js/dashboard/stats/graph/graph-util.js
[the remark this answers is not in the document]
ukutaht: Yes, good point: https://martinfowler.com/bliki/FlagArgument.html

I think the whole function needs a bigger refactoring, but might be a good place to start from. The bigger refactoring is probably too much for this PR.

ukutaht:
> This pull request is now ready to be reviewed, @plausible/developers. Back-end support for intervals is already in production, and it's playing perfectly with the master version of the dashboard ✨
> 
> The last step is to merge and deploy this pull request, which includes the dashboard changes to support intervals. I added a feature flag called `intervals` that shows the "Graph detail" button conditionally, so we can release gradually.
> 
> image
> 
> Shout-out to @Vigasaurus who developed this highly-requested feature :) thank you again, and sorry this is taking so long! We want to be as safe as we can when releasing big changes.

Waiting anxiously to test this on staging :) From the screenshot I can think of two improvements to test:
1. Switch the positions of the CSV export button and the graph detail selector
2. Maybe always showing the actual interval is better than 'Graph detail'. If it shows 'Daily ↓', it's more informative immediately and the arrow would be a good indicator that there are more options.

Both of these are just ideas that we could experiment with. I don't want to give any further UI feedback without playing around with it interactively.

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Great suggestion! Changed in 98964c4a.

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Good call! Changed in 0a67e8f6321f20c89ad00c90682804e497c8256f.

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -12,109 +13,45 @@ export const INTERVAL_LABELS = {
...
-          <Transition
-            show={this.state.open}
-            as={Fragment}
-            enter="transition ease-out duration-100"
-            enterFrom="transform opacity-0 scale-95"
-            enterTo="transform opacity-100 scale-100"
-            leave="transition ease-in duration-75"
-            leaveFrom="transform opacity-100 scale-100"
-            leaveTo="transform opacity-0 scale-95"
+      <Menu.Item onClick={() => { this.props.updateInterval(interval) }} key={interval}>
```
ukutaht, line 52: The `Transition` import is no longer used. I assume it's built into the Headless UI component itself? Anyways, we should remove the import.

Feels like something a linter would catch as well.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -335,45 +401,70 @@ export default class VisitorGraph extends React.Component {
...
+      <FadeIn show={showGraph}>
+        <LineGraphWithRouter graphData={graphData} topStatData={topStatData} site={site} query={query} darkTheme={theme} metric={metric} updateMetric={this.updateMetric} updateInterval={this.updateInterval}/>
+      </FadeIn>
+    )
   }
 
   render() {
-    const {metric, topStatData, graphData} = this.state
+    const {metric, loadingMainGraph, loadingTopStats} = this.state
+    const graphClassName = classNames({
```
ukutaht, line 456: The convention that makes classes easier to read/scan is:
1. If class is not conditional, list it as a string in the first argument
2. If class is conditional, list it as object key in second argument

```suggestion
    const graphClassName = classNames('mx-auto loading', {
      'pt-52 sm:pt-56 md:pt-60': loadingMainGraph == 2,
      'pt-32 sm:pt-36 md:pt-48': loadingMainGraph !== 2 && metric,
      'pt-16 sm:pt-14 md:pt-18 lg:pt-5': loadingMainGraph !== 2 && !metric
    })
```

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -0,0 +1,66 @@
...
+      </Menu>
+    )
+  }
+
+  renderDropdownOption(interval) {
+    return (
+      <Menu.Item onClick={() => { this.props.updateInterval(interval) }} key={interval}>
+        {({ active }) => (
+          <span
+            className={classNames(
```
ukutaht, line 55:
```suggestion
            className={classNames({
            'block px-4 py-2 text-sm',
            'bg-gray-100 dark:bg-gray-900 text-gray-900 dark:text-gray-200 cursor-pointer': active,
            'text-gray-700 dark:text-gray-200': !active
            
            })
```

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -0,0 +1,134 @@
+import { Transition } from '@headlessui/react';
+import { ChevronDownIcon } from '@heroicons/react/20/solid'
+import React, { Component, Fragment } from 'react';
+
+export const INTERVAL_MAPPING = {
```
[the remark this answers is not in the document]
ukutaht: ❤️❤️❤️

Love it. Had one comment/question left on the commit diff itself.

> fetching yearly data in a hourly interval 

Don't see why anyone would need it. Let's start with more restricted options and keep our eyes/ears open for any feedback about the granularity.

⚠️ One important thing that I realized we need to test/implement: The CSV export should always follow what's shown the dashboard. When a custom interval is configured, it should export the CSV with the requested interval.

assets/js/dashboard/stats/graph/interval-picker.js
ukutaht: Beautiful, had a comment about unused import on the diff itself.

I really appreciate the habit of posting the commit diff for feedback comments. It's so helpful for reviews 🙂

assets/js/dashboard/stats/graph/visitor-graph.js
ukutaht: 👍

Left some suggestions about more idiomatic use of classNames:
https://github.com/plausible/analytics/pull/1574/files#r1026171716
https://github.com/plausible/analytics/pull/1574/files#r1026168066

assets/js/dashboard/stats/graph/visitor-graph.js
ukutaht: Yes. I would consider it a bug if the backend is sending bad responses.

I wonder if it has to do with a race condition when the user changes intervals faster than the API can respond. I've asked @Vigasaurus if he remembers and can clarify: https://github.com/plausible/analytics/commit/1ae8c07d2f33f7f5d32695510b81c75fac48f920#r90321392

If the problem is a race condition with backend API calls, then the better option would be for the frontend to cancel the  old API request before requesting a new batch of data. This is already [implemented](https://github.com/plausible/analytics/blob/3bedf9281c69b125445b96b87f643dacdfb6403a/assets/js/dashboard/index.js#L40).

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Thanks! Fixed in [9dbc310a503d957c9eb84ea498415144ecbc47fd](https://github.com/plausible/analytics/pull/1574/commits/9dbc310a503d957c9eb84ea498415144ecbc47fd) and [5cc5e609cec18a1bbfcc9f7be93417c3135d2a2e](https://github.com/plausible/analytics/pull/1574/commits/5cc5e609cec18a1bbfcc9f7be93417c3135d2a2e)

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -387,72 +387,6 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
   describe "GET /api/stats/main-graph - varying intervals" do
     setup [:create_user, :log_in, :create_new_site]
 
-    test "displays visitors for a month on an hourly scale", %{conn: conn, site: site} do
```
vinibrsl, line 390: I removed these tests because as of [6076650384c1240f958784616686e91b5938f47c](https://github.com/plausible/analytics/pull/1574/commits/6076650384c1240f958784616686e91b5938f47c), these period + interval combinations are not supported anymore, and we already have tests for invalid combinations.

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Good catch! Fixed in [76e54d64fbdd388bcce5f86ba0c602c11cbb43c7](https://github.com/plausible/analytics/pull/1574/commits/76e54d64fbdd388bcce5f86ba0c602c11cbb43c7).

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Fixed in [76e54d64fbdd388bcce5f86ba0c602c11cbb43c7](https://github.com/plausible/analytics/pull/1574/commits/76e54d64fbdd388bcce5f86ba0c602c11cbb43c7).

assets/js/dashboard/stats/graph/interval-picker.js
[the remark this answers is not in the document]
vinibrsl: Good call! Thanks for clarifying. I transformed `IntervalPicker` into a function component as suggested.

React Context looks promising and fits well our use-case where `site` and `query` is required almost everywhere.

vinibrsl:
I think I got every code review comment covered. Thanks all for your inputs 🙌

I pushed [one extra commit](https://github.com/plausible/analytics/pull/1574/commits/d1a92f7689e67e5933fcff4b3b89238eb77377d4) reviewing the overall user experience, including smooth transitions, dark mode support, fixed a bug where the interval input would blink when changing intervals, etc. I also implemented @ukutaht's suggestion of displaying the selected interval replacing the "Graph detail" label.

Screen Shot 2022-11-18 at 23 34 18

I tested the CSV export, and it is not taking intervals into account. We should definitely fix that. Can we do that in a separate pull request, as I'm guessing that would be mainly back-end code? I feel like this pull request is starting to grow again, and I want to avoid that.

If you have some time, please give this branch a try. [Seeding the database](https://github.com/plausible/analytics/blob/master/CONTRIBUTING.md#seeds) may help. Let me know if there are any other UX improvements I can make, or bugs you find.

assets/js/dashboard/stats/graph/visitor-graph.js
ukutaht: Good catch! Client-side state makes everything more complicated 😕

For this specific edge case, I think there's an approach available that would make the logic more straightforward. Instead of waiting for the round-trip to the server, we could validate the interval directly when fetching it from `localStorage`. We have the set of valid intervals available from the backend.

This could even be extracted to a function like `storage.getValidInterval(site, period, validIntervalsByPeriod)`. We can start utilizing the `storage` module more in general, and start encapsulating the keys in there as well so they don't leak into UI code.

I feel like it would make the purpose of the validation more obvious to the reader. Unless, of course, there are other reasons why the response validation is needed. What do you think?

assets/js/dashboard/stats/graph/visitor-graph.js
ukutaht: :chefskiss:

assets/js/dashboard/stats/graph/interval-picker.js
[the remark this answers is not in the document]
ukutaht: beautiful 👍

metmarkosaric:
> If you have some time, please give this branch a try. [Seeding the database](https://github.com/plausible/analytics/blob/master/CONTRIBUTING.md#seeds) may help. Let me know if there are any other UX improvements I can make, or bugs you find.

i don't know how to give this branch a try but i'd love to test this feature and send any feedback in order not to delay things too much. for this purpose, it would be nice if we can get these types of features that make important visual changes to the UI in staging itself so everyone can test there with for instance our live demo data.

vinibrsl:
> > it would be nice if we can get these types of features that make important visual changes to the UI in staging itself so everyone can test there with for instance our live demo data.
> 
> We don't have live demo data mirroring from prod->staging yet. Staging only has some fake data. Not great for UI testing because the dashboard doesn't 'feel real'. We can implement the live mirroring at some point next year for some public dashboards that are good for testing.
> 
> Correct me if I'm wrong @vinibrsl but the best way to test this right now is in production. It will be deployed behind a feature flag which means it can be enabled for your personal site @metmarkosaric for testing before we enable it for the public. We can use the same approach for comparisons, funnels and other new features that don't modify existing UI too much.

You're right! The best path for assessing this feature is in production, as we don't have _real_ data patterns in staging yet. It is under a feature flag, so we should be good to go. Once it's deployed I'll add your user IDs to the feature flag.

metmarkosaric:
nice one, thank you!

ukutaht (approved):

vinibrsl:
Amazing work, @Vigasaurus! We'll release this feature incrementally during the week 💙

PR #1576, opened 2022-01-01
change sites to stats

metmarkosaric:
changing sites to stats as couple of people mentioned it sounds like we will lock their websites so nobody can visit them


ukutaht:
Thanks!

PR #1609, opened 2022-01-14
Change `Country` to `Location` in the Add filter dropdown

RobertJoonas:


ukutaht:
Thanks for this. The rename is appropriate but incomplete. The URL should also be changed from `/filter/country`  to `/filter/location` and the header of the filter modal should be changed from 'Country' to 'Location'

ukutaht:
Closing in favour of https://github.com/plausible/analytics/pull/1634

PR #1610, opened 2022-01-14
Update session screen_size if unknown

ssendev:
### Changes

If a session is created without a screen size and later requests specify a screen size update the session
fixes #1605 

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
I like this solution. I think it would make sense for other dimensions as well (like browser, OS, country etc.)

PR #1661, opened 2022-02-09
optional ipv6 for repo

ericsullivan:
### Changes

I tried to self-host on fly.io, but it uses IPv6 for DB access:
https://hexdocs.pm/phoenix/fly.html#runtime-configuration

This changes follows the current Phoenix generator setup:
https://github.com/phoenixframework/phoenix/blob/master/installer/lib/phx_new/generator.ex#L355

It had been defaulted to inet6:
https://github.com/phoenixframework/phoenix/pull/4289

but that caused an issue:
https://github.com/phoenixframework/phoenix/issues/4389

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - If a session was started without a screen_size it is updated if an event with screen_size occurs
 - Added `LISTEN_IP` configuration parameter plausible/analytics#1189
 - The breakdown endpoint with the property query `property=event:goal` returns custom goal properties (within `props`)
+- Added IPv6 Ecto support (via the environment-variable `ECTO_IPV6`)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ukutaht:
Thanks! Definitely interested in getting this in.

Did you consider parsing the `DATABASE_URL` config parameter? So if the hostname is an IPv6 then the `[:inet6]`config option will automatically be applied. I don't understand why Ecto doesn't work that way to begin with

ericsullivan:
@ukutaht The DATABASE_URL uses a DNS entry like:
```
DATABASE_URL=postgres://<USERNAME>:<PW>@top2.nearest.of.plausible-db.internal:5432/plausible
```
So you can't tell it's an IPv6 without doing the lookup. Since it's fly.io they're all IPv6, which is why I used the ENV approach

ukutaht:
Oh, gotcha. That makes sense.

ericsullivan:
Thank you

PR #1730, opened 2022-03-02
Reduce tracker size by using `navigator.sendBeacon`

willfarrell:
### Changes

Replace `XMLHttpRequest` for `navigator.sendBeacon`. Also adds in the ability to allow events to be triggered on page exit.

This will likely need an update to the back end (sorry, don't know elixir) to change what headers are used (ie use `ping-from` over `Referer` maybe?). Not sure what others header are send along with the request that maybe be used by plausible. Some testing will be needed.

> It's intended to be used for sending analytics data to a web server, and avoids some of the problems with legacy techniques for sending analytics, such as the use of [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest). ~ MDN

- HTTP on page exit: https://css-tricks.com/send-an-http-request-on-page-exit/
- MDN: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon
- Browser Support: https://caniuse.com/beacon

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 - Cache the tracking script for 24 hours
+- Replace `XMLHttpRequest` for `navigator.sendBeacon`
 
 ## v1.4.1
 
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -69,17 +69,7 @@
     payload.h = 1
     {{/if}}
 
-    var request = new XMLHttpRequest();
-    request.open('POST', endpoint, true);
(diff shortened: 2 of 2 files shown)


ukutaht:
Hi @willfarrell, thanks for the PR!

I think you already read the discussions we've had about this. I think currently Beacon is not [well-supported enough](https://caniuse.com/?search=sendbeacon) to make it the default way of sending pageviews. We have talked about and might incorporate beacon as an optional extra in the future. But we definitely won't use it as the default anytime soon.

fellmirr:
@ukutaht has browser support matured enough for this to become a viable option?

PR #1746, opened 2022-03-09
Extra note for the domain taken message

metmarkosaric:
I wanted to add an extra note to the "has already been taken" message. We now regularly get questions about this message and in most cases it happens because a team member already registered the site but didn't share the access to the rest of the team. I hope I've added it correctly.


ukutaht:
Needed some more changes. I've made them here: https://github.com/plausible/analytics/commit/e026148fb09b4eeffcd6e97f8dd234c6cf287585

metmarkosaric:
thanks!

PR #1759, opened 2022-03-12
Improve stats api

RobertJoonas:
### Changes

Added the ability to query the `events` metric from Stats API breakdown endpoint when breaking down on a `visit:` property. Modified one test to also use this metric in such a query.

Stats API docs don't say anything about this not working, so no update required there.

### Tests

### Changelog

### Documentation

### Dark mode


RobertJoonas:
I removed the changelog update that I made earlier, as it's basically already written there and these changes will just fix that behaviour.

PR #1775, opened 2022-03-17
added file-downloads script extension

RobertJoonas:
### Changes

Adds the `file-downloads` script extension that can be used to automatically track file downloads.

### Tests

### Changelog

### Documentation

### Dark mode


tracker/src/plausible.js
```diff
@@ -117,6 +117,48 @@
   registerOutboundLinkEvents()
   {{/if}}
 
+  {{#if file_downloads}}
+  var defaultFileTypes = ['.pdf', '.xlsx', '.docx', '.txt', '.rtf', '.csv', '.exe', '.key', '.pps', '.ppt', '.pptx', '.7z', '.pkg', '.rar', '.gz', '.zip', '.avi', '.mov', '.mp4', '.mpeg', '.wmv', '.midi', '.mp3', '.wav', '.wma']
+  var fileTypesToTrack = scriptEl.getAttribute('file-types') || defaultFileTypes;
```
ukutaht, line 122: I'm confused how this line can work, does the `file-types` attribute get converted to an Array automatically somehow?

tracker/src/plausible.js
```diff
@@ -117,6 +117,48 @@
...
+    var middle = event.type == "auxclick" && event.which == 2;
+    var click = event.type == "click";
+
+    while(link && (typeof link.tagName == 'undefined' || link.tagName.toLowerCase() != 'a' || !link.href)) {
+      link = link.parentNode
+    }
+
+    if (link && link.href && isDownloadToTrack(link.href)) {
+
+      if (middle || click)
```
ukutaht, line 136: I don't think we should rely on whitespace in Javascript `if` statements. Let's use curly braces to be sure the right block gets executed.

tracker/src/plausible.js
```diff
@@ -117,6 +117,48 @@
...
+            location.href = link.href;
+          }, 150);
+          event.preventDefault();
+        }
+      }
+    }
+  }
+
+  function isDownloadToTrack(url) {
+    var pathArray = url.split('/');
```
ukutaht, line 152: How about 
```js
return fileTypesToTrack.some(function(fileType) {
  return url.endsWith(filetype)
})
```

ukutaht (changes requested):
Looks good overall. Very similar to outbound links

ukutaht (approved):

ukutaht:
Well done, thanks!

In the future you can hit 'resolve' on the specific pieces of feedback in the review as you address them.

Also, for future PRs let's try to use the workflow built into Github. If the PR is not ready for review, it should be marked as a draft PR. Once it's ready, you can request a review from me through Github so you don't need to ping me a message somewhere else. And if I leave feedback and you address it, request another review through Github.

PR #1799, opened 2022-03-25
1421 contains filter

andreamazza89:
### Changes

- Moves `page_entry` and `page_exit` filters to be part of the same filter group as `page`
- Adds a "contains" option as a filter type for page-based filters, which has the same effect as using double asterisks before and after the keyword (i.e. `**<keyword>**`).
- Fixes two small UI bugs (see the screenshots below)

I picked tilde (`~`) as the character to signify a 'contains' query, I hope that's ok?

This is what the new filter type looks like:

image
image

The small UI bugs are:

- the 'is not' filter type would display as `is_not` once selected:
image
- if I select any 'is not' filter, save and then edit the filter, `!` is added in front of the filter keyword:
image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,6 @@ All notable changes to this project will be documented in this file.
 - The breakdown endpoint with the property query `property=event:goal` returns custom goal properties (within `props`)
 - Added IPv6 Ecto support (via the environment-variable `ECTO_IPV6`)
 - New filter type: `contains`, available for `page`, `entry_page`, `exit_page`
-- Move `entry_page` and `exit_page` to be part of the `Page` filter group
 
 ### Fixed
 - UI fix where multi-line text in pills would not be underlined properly on small screens.
@@ -31,6 +30,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 - Cache the tracking script for 24 hours
+- Move `entry_page` and `exit_page` to be part of the `Page` filter group
 
 ## v1.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ All notable changes to this project will be documented in this file.
 - Added `LISTEN_IP` configuration parameter plausible/analytics#1189
 - The breakdown endpoint with the property query `property=event:goal` returns custom goal properties (within `props`)
 - Added IPv6 Ecto support (via the environment-variable `ECTO_IPV6`)
+- New filter type: `contains`, available for `page`, `entry_page`, `exit_page`
+- Move `entry_page` and `exit_page` to be part of the `Page` filter group
 
 ### Fixed
 - UI fix where multi-line text in pills would not be underlined properly on small screens.
--- a/assets/js/dashboard/stats/modals/filter.js
+++ b/assets/js/dashboard/stats/modals/filter.js
@@ -42,48 +42,31 @@ function getFormState(filterGroup, query) {
   }, {})
 }
 
-const filterTypes = {
+const FILTER_TYPES = {
   isNot: 'is not',
   contains: 'contains',
   is: 'is'
 };
 
-const filterPrefixes = {
-  isNot: '!',
-  contains: '~',
-  is: ''
+const FILTER_PREFIXES = {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ All notable changes to this project will be documented in this file.
 - Added `LISTEN_IP` configuration parameter plausible/analytics#1189
 - The breakdown endpoint with the property query `property=event:goal` returns custom goal properties (within `props`)
 - Added IPv6 Ecto support (via the environment-variable `ECTO_IPV6`)
+- New filter type: `contains`, available for `page`, `entry_page`, `exit_page`
+- Move `entry_page` and `exit_page` to be part of the `Page` filter group
```
ukutaht, line 24: I would say the new filter type fits in the  `Added` categry but moving filters around should go under `Changed`.

See all change types: https://keepachangelog.com/en/1.0.0/#types

assets/js/dashboard/stats/modals/filter.js
```diff
@@ -43,6 +42,55 @@ function getFormState(filterGroup, query) {
...
+
+export function valueWithoutPrefix(value) {
+  return [filterTypes.isNot, filterTypes.contains].includes(toFilterType(value))
+    ? value.substring(1)
+    : value;
+}
+
+function toFilterQuery(value, type) {
+  let prefix;
+  switch (type) {
```
ukutaht, line 76: If you squint hard enough then both of the switch statements in this file could be map lookups. It's up to you if it feels cleaner/more readable but I tend to use a pattern like this often:

```js
const filterPrefixes = {
  [filterTypes.isNot]: '!',
  [filterTypes.contains]: '~',
  [filterTypes.is]: ''
};

function toFilterQuery(value, type) {
  const prefix = filterPrefixes[type]
  return prefix + value.trim()
}
```

ukutaht: Also, we've been naming global / magic variables in all uppercase. See examples below. Probably best to keep one style.

https://github.com/plausible/analytics/blob/c83e038af0a8044ff7d57ece899d7bb9de22109f/assets/js/dashboard/stats/devices/index.js#L93

https://github.com/plausible/analytics/blob/c83e038af0a8044ff7d57ece899d7bb9de22109f/assets/js/dashboard/index.js#L9

https://github.com/plausible/analytics/blob/c83e038af0a8044ff7d57ece899d7bb9de22109f/assets/js/dashboard/stats/conversions/prop-breakdown.js#L9

ukutaht (commented):
Very nice! Everything looks good overall, I just had a couple small comments about style.

assets/js/dashboard/stats/modals/filter.js
andreamazza89: Awesome, I like it 👍 

In `toFilterType` I've gone for a 'reverse lookup' (find a key for a given value) on the prefix map as that feels cleaner than having another map like

```js
const FILTER_PREFIX_TO_TYPE = {
  [FILTER_PREFIXES.isNot[FILTER_TYPES.isNot]]: FILTER_TYPES.isNot,
  [FILTER_PREFIXES.contains[FILTER_TYPES.contains]]: FILTER_TYPES.contains,
  [FILTER_PREFIXES.is[FILTER_TYPES.is]]: FILTER_TYPES.is
};
```

Hope this makes sense and I haven't misunderstood your point?

andreamazza89:
Thank you for your feedback!

I've pushed changes, please let me know if I've misunderstood or missed anything 😄

ukutaht:
Sweet, thanks @andreamazza89!

ukutaht:
This can now be tested at https://staging.plausible.io/plausible.io

@metmarkosaric let me know if you have any feedback about the UI. From my testing it seems to work exactly as expected.

assets/js/dashboard/stats/modals/filter.js
ukutaht: Hehe, yeah . Looks good :) Wish the reverse lookup operation was built-in to the Object/Map/Hash data type in standard libraries.

metmarkosaric:
looks good to go!

PR #1816, opened 2022-04-06
Pageview custom dimensions

RobertJoonas:
### Changes

- [ ] Add the ability to send pageview dimensions based on meta tags ([discussion comment](https://github.com/plausible/analytics/discussions/320#discussioncomment-179904))
- [x] Add the `is <value>` filtering functionality for dimensions
- [x] Add the `is not <value>` filtering functionality for dimensions
- [x] Complete the dimension filter functionality in the UI
- [ ] Globally rename `props` to `dimensions`

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/base.ex
```diff
@@ -158,6 +174,48 @@ defmodule Plausible.Stats.Base do
           )
       end
 
+    sessions_q =
```
ukutaht, line 177: Is this ever used? I don't see how one could add a filter for `visit:entry_props`.

ukutaht (approved):

lib/plausible/stats/base.ex
RobertJoonas: It currently gets added to the Top Sources query via the `treat_prop_as_entry_prop` function ([code_ref](https://github.com/plausible/analytics/pull/1816/files#diff-d90d0982415340f17b8c7949591770cbc1979d14a0faae8666c5b7b9bf6aa113R275)). We agreed to change this behaviour and add entry props as a separate filter instead, but this functionality should remain the same I think

ukutaht: Yeah, makes sense

PR #1824, opened 2022-04-08
Update the internal /sites api to paginate results

andreamazza89:
image

### Changes

This PR updates the internal api to paginate `/sites` results and updates the site switcher in the dashboard to add a `View all` link when there are more than one page worth of sites.

This is in preparation of adding a search functionality to the `/sites` page (see [this discussion](https://github.com/plausible/analytics/discussions/1049)).
The idea for having these changes separate is to limit the size of PRs. Do let me know if you prefer larger PRs to incremental ones.

I am not in love with the way the `View all` link looks next to `+ Add Site`, but could not think of anything nicer. Do let me know if you have something else in mind.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/api/internal_controller.ex
```diff
@@ -11,14 +13,37 @@ defmodule PlausibleWeb.Api.InternalController do
...
-  def sites(conn, _) do
-    if conn.assigns[:current_user] do
-      user = Repo.preload(conn.assigns[:current_user], :sites)
-      json(conn, Enum.map(user.sites, & &1.domain))
+  def sites(conn, params) do
+    current_user = conn.assigns[:current_user]
+
+    if current_user do
+      sites =
+        sitesFor(current_user, params)
```
ukutaht, line 21: Use `snake_case` in Elixir instead of `camelCase`

lib/plausible_web/controllers/api/internal_controller.ex
```diff
@@ -11,14 +13,37 @@ defmodule PlausibleWeb.Api.InternalController do
...
+        order_by: s.domain
+      ),
+      params
+    )
+  end
+
+  defp buildResponse({sites, pagination}, conn) do
+    %{
+      data: Enum.map(sites, &%{domain: &1.domain}),
+      pagination: paginate(conn, pagination)
```
ukutaht, line 46: I use `import` very sparingly because it can be confusing where certain functions are defined. This is a good example - I wanted to see what `paginate` does and it took me a few seconds to figure out it's coming from a dependency.

I think importing `Ecto.Query` for example makes a lot of sense because you're basically importing a whole syntax / DSL. Or when you have a lot of shared base functionality like `Plausible.Stats.Base`.

But I think in this case the best for readability would be to just use the module right where the function is called. If that's too verbose then use `alias` instead.

ukutaht (commented):
> I am not in love with the way the View all link looks next to + Add Site, but could not think of anything nicer. Do let me know if you have something else in mind.

Agreed. How about we just remove the `+ Add site` from there completely? And replace that with 'view all' which is shown always, regardless of how many sites you have?

I doubt many people use the dropdown to actually enter the new site flow. And even if they do, it's not that difficult to hit `View all -> Add site`. Adding a new site is not an everyday activity for most users. Going to your site listing is probably much more common, so always displaying `View all` feels like a sound choice to improve navigation.

What do you think @metmarkosaric? The context here is that Andrea is working on adding a search box to the /sites page but also looking at some other cases where we don't handle a large number of sites well. One of them is the site switcher dropdown. What should it do when you have 1500 sites?

We decided that adding a search box to the dropdown would expand the scope a bit too much but as a temporary measure we thought it would be nice to limit the dropdown to 20 results and add a `View all` link so you could go to your listing page and start searching.

ukutaht:
Oh, love the incremental PRs by the way! Keep em coming

metmarkosaric:
removing Add Site and adding View All sounds like a good solution to me!

andreamazza89:
Thank you for your comments 👍 

This should now be ready for merging - let me know if there's anything else

ukutaht:
Thanks @andreamazza89!

PR #1831, opened 2022-04-12
added add-file-types attribute + now ignores query params

RobertJoonas:
### Changes

- Can now specify the `add-file-types` attribute tag to add custom file types without overriding defaults
- Can now also track file downloads for a link href that contains query parameters

### Tests

### Changelog

### Documentation


tracker/src/plausible.js
```diff
@@ -138,7 +139,7 @@
     if (link && link.href && isDownloadToTrack(link.href)) {
 
       if (middle || click) {
-        plausible('File Download', {props: {url: link.href}})
+        plausible('File Download', {props: {url: link.href.split("?")[0]}})
```
ukutaht, line 142: Instead of duplicating the `.split("?")[0]` here and below, I think it would be better to declare a variable like so:

```js
var linkTarget = link.href.split("?")[0]

isDownloadToTrack(linkTarget)
...
plausible('File Download', {props: {url: linkTarget}})
```

RobertJoonas: Good point! Extracted it into a `linkTarget` variable with the new commit

PR #1849, opened 2022-04-22
Re-adds underline to active top stat header

Vigasaurus:
### Changes

Re-adds the underline (now thinner) for the currently selected top stat, & metric, to differentiate the hover state from the actual active state. I noticed this got removed in the cherry picked commit. I figure it wasn't fully intentional, as the hover state and the active state being identical is probably a bit problematic (as can be seen from the recent discussion in #1364), but if it was fully intentional feel free to close this PR.

image

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
It was intentional because I felt the underline was too heavy. You're right that it seems to cause a problem so let's try with an underline

Vigasaurus:
Makes sense - yeah, looking back at it that 2px thick underline was indeed too heavy - if the new 1px is still too distracting we could also try different shades of indigo for the hover vs active state as needed (but that's not really a defined style in the current system).

PR #1892, opened 2022-05-06
Have entrypoint script run required steps directly

RealOrangeOne:
### Changes

This removes the need for a custom `command` in `docker-compose.yml`. https://github.com/plausible/hosting/blob/f7682057104671b42e3fff3a106f93a32ea14775/docker-compose.yml#L30

It intentionally removes `init-admin` from the setup, as this should really only be done once.

This should make the installation much simpler, as it works more expected, and means people can unset the environment variables required for the admin setup script. It should also make it start up _marginally_ faster.

Interested to hear what people think.

### Tests

### Changelog

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated (I'll do this once the approach is approved, but before merge)

### Dark mode

### DIFF
--- a/.gitlab/build-scripts/docker-entrypoint.sh
+++ b/.gitlab/build-scripts/docker-entrypoint.sh
@@ -1,14 +1,10 @@
 #!/bin/sh
+
 set -e
 
-if [ "$1" = 'run' ]; then
-      exec /app/bin/plausible start
+# Confirm database exists in the required state
+/app/createdb.sh
+/app/migrate.sh
 
-elif [ "$1" = 'db' ]; then
-      exec /app/"$2".sh
- else
-      exec "$@"
-
-fi
-
-exec "$@"
+# Start Plausible
+/app/bin/plausible start
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
 - Paginate /api/sites results and add a `View all` link to the site-switcher dropdown in the dashboard.
 - Remove the `+ Add Site` link to the site-switcher dropdown in the dashboard.
 - `DISABLE_REGISTRATIONS` configuration parameter can now accept `invite_only` to allow invited users to register an account while keeping regular registrations disabled plausible/analytics#1841
+- Simplify container entrypoint to remove need for custom `command`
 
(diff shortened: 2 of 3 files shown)


ukutaht:
I'm generally in favour of this.

How would people run `init-admin` with this setup? Eventually I would like to build that behaviour into the UI so people can just create the admin account through the UI. But for now we would need to be able to run `init-admin` manually during setup.

RealOrangeOne:
> How would people run `init-admin` with this setup

Once they start the plausible container (and its dependencies), it'd be something like `docker-compose exec plausible /app/init-admin.sh`.

I'll get a docs PR together which updates the install docs, which should help the picture a little too

schnerring:
> It should also make it start up marginally faster.

Why? Because `init-admin` isn't executed every time on startup? It's one user lookup by email address per startup. The performance gain is negligible IMO.

`create-db` is also something that should be run only once. Also, running migrations implicitly is not something everybody would want. We should at least add a `DISABLE_MIGRATIONS` env var switch.

Having said all that, I don't really see the point of these changes then.

---

As example, [I'm currently implementing a Helm chart for the TrueCharts community repo](https://github.com/truecharts/apps/pull/3055). The resulting Kubernetes deployment does this:

1. Run an init container awaiting both Plausible and ClickHouse to accept connections (instead of naively running `sleep 10`)
2. Run one of the following init containers:
    - when **installing** the Helm chart, run: `createdb`, `migrate`, `init-admin`
    - when **upgrading** the Helm chart, run: `migrate`
3. Run the main container with default `run`

RealOrangeOne:
When running in a k8s environment, yes using init containers is a better solution. But outside of that (eg docker-compose), that doesn't exist.

I agree though it's definitely best to allow users to opt for the functionality they want. So in the k8s case, being able to disable migrations etc on container start is probably ideal (even if most of the time they'd noop)

ukutaht:
Having thought about it, this change will not work for us since we run this docker image in a non-self-hosted environment. We want to be in manual control over migrations etc. and don't want the image itself to run these scripts. I think the `plausible/hosting` repo is an appropriate place for that.

The long term view is to build this admin stuff into the UI so people don't need to run the init-admin script and stuff like that themselves.

PR #1903, opened 2022-05-13
CRM improvements

RobertJoonas:
### Changes

- [x] Can now unlock and remove the grace period from a user from "Actions -> Unlock"
- [x] Now displays user email address and role in the "other members" column
- [x] The "subscription plan" column for users now contains links to manage the subscription in Paddle
- [x] Created a new `Mix.Task` to speed up the subscription creating process in dev mode

### DIFF
--- a/lib/mix/tasks/create_sandbox_subscription.ex
+++ b/lib/mix/tasks/create_sandbox_subscription.ex
@@ -1,0 +1,68 @@
+defmodule Mix.Tasks.CreateSandboxSubscription do
+  use Mix.Task
+  use Plausible.Repo
+  alias Plausible.{Repo, Auth.User, Billing.Subscription}
+  require Logger
+
+  # Steps to create a subscription in dev environment
+  #
+  # 1) Subscribe to a sandbox plan in the UI > User Settings. Instructions:
+  #    https://developer.paddle.com/getting-started/ZG9jOjIxODY4NjYx-sandbox
+  #
+  # 2) find the created subscription_ID here:
+  #    https://sandbox-vendors.paddle.com/subscriptions/customers
+  #
+  # 3) run from command line:
+  #    mix create_sandbox_subscription <subscription_ID>
+
+  @headers [
+    {"Content-type", "application/json"},
+    {"Accept", "application/json"}
+  ]
+
+  def run([paddle_subscription_id]) do
+    Mix.Task.run("app.start")
+
+    config = Application.get_env(:plausible, :paddle)
+
+    endpoint = Plausible.Billing.PaddleApi.vendors_domain() <> "/api/2.0/subscription/users"
+
+    params = %{
+      vendor_id: config[:vendor_id],
+      vendor_auth_code: config[:vendor_auth_code],
+      subscription_id: paddle_subscription_id
+    }
+
+    case HTTPoison.post(endpoint, Jason.encode!(params), @headers) do
+      {:ok, response} ->
+        body = Jason.decode!(response.body)
+
+        if body["success"] do
+          res = body["response"] |> List.first()
+          user = Repo.get_by!(User, email: res["user_email"])
+
+          subscription = %{
+            paddle_subscription_id: res["subscription_id"] |> to_string(),
+            paddle_plan_id: res["plan_id"] |> to_string(),
+            cancel_url: res["cancel_url"],
+            update_url: res["update_url"],
+            user_id: user.id,
+            status: res["state"],
+            next_bill_date: res["next_payment"]["date"],
+            next_bill_amount: res["next_payment"]["amount"] |> to_string(),
+            currency_code: res["next_payment"]["currency"]
+          }
+
+          Subscription.changeset(%Subscription{}, subscription)
+          |> Repo.insert!()
+
+          Logger.info("Subscription created for user #{user.id} (#{user.email})")
+        else
+          Logger.error(body["error"])
+        end
+
+      {:error, reason} ->
+        Logger.error(reason)
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/create_sandbox_subscription.ex
```diff
@@ -0,0 +1,68 @@
+defmodule Mix.Tasks.CreateSandboxSubscription do
```
ukutaht, line 1: I would prefer to call this something like `SyncSandboxSubscription` or `PullSandboxSubscription` instead. The subscription has already been created in the master database (which is managed by Paddle). We are just pulling down the subscription in order to synchronize with Paddle. Using the word `Sync` or `Pull` implies correctly that the source of truth is external to our application/database.

lib/plausible/auth/user_admin.ex
```diff
@@ -20,7 +20,7 @@ defmodule Plausible.Auth.UserAdmin do
       email: nil,
       inserted_at: %{name: "Created at", value: &format_date(&1.inserted_at)},
       trial_expiry_date: %{name: "Trial expiry", value: &format_date(&1.trial_expiry_date)},
-      subscription_tier: %{value: &subscription_tier/1},
+      subscription_plan: %{value: &subscription_tier/1},
```
ukutaht, line 23: Make the same change to the function name: `&subscription_tier/1 -> &subscription_plan/1`. More consistent that way

lib/plausible/auth/user_admin.ex
```diff
@@ -31,6 +31,14 @@ defmodule Plausible.Auth.UserAdmin do
       remove_grace_period: %{
         name: "Remove grace period",
         action: fn _, user -> remove_grace_period(user) end
+      },
+      lock_sites: %{
```
ukutaht, line 35: I'm a bit worried about this manual switch conflicting with the nightly automatic system and creating confusion. For example, if we unlock all sites for a user manually but forget to remove the grace period, then the sites will be locked again at midnight.

Is there any other use-case for this other than removing a grace period and unlocking all sites immediately? I'm thinking about people who went over limit for just 2 months for example and don't want to upgrade.

If that's the only use-case for manually controlling the lock status then all we need to do here is add `set_lock_status_for(user, false)` to the `remove_grace_period` action instead.

I'm just trying to think if there's a way to design it so that user error is impossible.

ukutaht: >  For example, if we unlock all sites for a user manually but forget to remove the grace period, then the sites will be locked again at midnight.

This will work the other way as well: if you manually lock sites for someone, they will all be unlocked at midnight unless something about their subscription or trial has changed. The automatic system will always be dominant. 

So in the CRM I think we should add tools to affect the automatic system (like removing the grace period) in combination with manual locking to make those changes take effect immediately. But providing a way to just 'lock all sites' manually will only work until midnight so it's kind of pointless. If we need a manual way to lock a whole account we need to make some bigger changes to the automatic system as well.

lib/plausible/auth/user_admin.ex
```diff
@@ -58,7 +71,13 @@ defmodule Plausible.Auth.UserAdmin do
     if user.subscription && user.subscription.status == "active" do
       quota = PlausibleWeb.AuthView.subscription_quota(user.subscription)
       interval = PlausibleWeb.AuthView.subscription_interval(user.subscription)
-      "#{quota} (#{interval})"
+
+      manage_url =
```
ukutaht, line 75: Nice!

lib/plausible/site/admin.ex
```diff
@@ -52,9 +52,17 @@ defmodule Plausible.SiteAdmin do
     Enum.find(site.memberships, fn m -> m.role == :owner end).user.email
   end
 
-  defp get_other_members_emails(site) do
-    memberships = Enum.reject(site.memberships, fn m -> m.role == :owner end)
-    Enum.map(memberships, fn m -> m.user.email end) |> Enum.join(", ")
+  defp get_other_members(site) do
+    Enum.reduce(site.memberships, "", fn m, acc ->
```
ukutaht, line 56: This works OK but at least for my brain here's a much more natural way to build this string:

```elixir
Enum.map(site.memberships, fn m -> 
  case m.role do
    :owner -> m.user.email
    role -> m.user.email <> "(#{role})"
  end
end)
|> Enum.join(",")
```

RobertJoonas: Yeah, point taken. The owner should not show up in the "other members" column though, but I made it look much nicer now. Thanks!

lib/plausible/auth/user_admin.ex
RobertJoonas: I now merged the "unlock sites" and "remove grace period" actions into one single "Unlock" action.

But yeah, for locking users we'd have to make some bigger changes. And considering that this would be very rarely used, let's postpone building this for now.

metmarkosaric: nice one! and yeah, we just need easy way to unlock sites for now

lib/plausible/site/admin.ex
ukutaht: Oh, right, missed a `filter` here as well :)

lib/plausible/site/admin.ex
```diff
@@ -52,9 +52,10 @@ defmodule Plausible.SiteAdmin do
     Enum.find(site.memberships, fn m -> m.role == :owner end).user.email
   end
 
-  defp get_other_members_emails(site) do
-    memberships = Enum.reject(site.memberships, fn m -> m.role == :owner end)
-    Enum.map(memberships, fn m -> m.user.email end) |> Enum.join(", ")
+  defp get_other_members(site) do
+    Enum.filter(site.memberships, &(&1.role != :owner))
```
ukutaht, line 56: Beautiful :)

ukutaht (approved):

PR #1936, opened 2022-05-30
Keep the public and in-app footers consistent

metmarkosaric:
i've tried to make the two footers consistent to include the different links we were missing in the in-app footer (terms, dpa...) but also to display the new job opening for those that are logged in. i don't have a way to test it so hopefully i didn't make any mistakes


ukutaht:
Cheers. I will test and make adjustments if needed

PR #1942, opened 2022-06-01
Add API to retrieve site by domain

kurtmc:
### Changes

Added an API endpoint to get a site based on it's domain. This is required to check if the site exists or not from my terraform provider: https://github.com/mcalpinefree/terraform-provider-plausible

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated: https://github.com/plausible/docs/pull/234

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/external_sites_controller.ex
+++ b/lib/plausible_web/controllers/api/external_sites_controller.ex
@@ -28,6 +28,16 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
     end
   end
 
+  def get_site(conn, %{"site_id" => site_id}) do
+    site = Sites.get_for_user(conn.assigns[:current_user].id, site_id, [:owner, :admin])
+
+    if site do
+      json(conn, site)
+    else
+      H.not_found(conn, "Site could not be found")
+    end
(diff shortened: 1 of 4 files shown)


lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -28,6 +28,16 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
     end
   end
 
+  def get_site(conn, %{"site_id" => site_id}) do
+    site = Sites.get_for_user(conn.assigns[:current_user].id, site_id, [:owner])
```
ukutaht, line 32: Let's expand this to `[:owner, :admin]`. Most other endpoints in this controller allow both owners and admins.

ukutaht (changes requested):
Thanks for this, looks good to me!

The only change I would request is to open this endpoint up for `admin` level users as well.  Otherwise it's ready to merge.

kurtmc:
Thanks @ukutaht ! I've applied that change now.

ukutaht (approved):

PR #1961, opened 2022-06-17
Add basic test harness for browserstack/playwright

ukutaht:
### Changes

Adds a testing harness for the tracker script so that we can write automated tests for the tracker.

Running these tests requires a BrowserStack account. Ask me if you need access.


tracker/test/fixtures/snippet.js
```diff
@@ -0,0 +1,9 @@
+const scriptVariant = document.currentScript.getAttribute('variant')
+
+// how to import the server address from '../support/server.js' ?
```
RobertJoonas, line 3: The idea of this `snippet.js` file is to keep the `http://localhost:3000` server address only in one place (otherwise every HTML fixture would have it written out). I didn't find any good solution for importing it from the Node module. Any ideas? Perhaps keep the configuration for it elsewhere and not in the `server.js` file?

RobertJoonas (dismissed):

vinibrsl (approved):
Looks good! This is a great starting point 💪

Out of curiosity: is this going to run for every deploy?

tracker/test/support/playwright.config.js
```diff
@@ -0,0 +1,49 @@
...
+  timeout: 60000,
+  use: {
+    viewport: null,
+    baseURL: LOCAL_SERVER_ADDR
+  },
+  projects: [
+    // -- BrowserStack Projects --
+    // name should be of the format browser@browser_version:os os_version@browserstack
+    // supported options: https://www.browserstack.com/docs/automate/playwright/browsers-and-os
+    {
```
ukutaht, line 22: It's a bit weird to me that Safari on OSX is missing from this list. I'd like to test with a lot more devices if it doesn't make things too slow.

I think best would be to try to test the whole matrix of `Chrome` ,`Safari` and `Firefox` on both `OSX` and `Windows`.  That should be 6 tests with the latest platforms.

Then I would also find the oldest versions that we want to support, and create another matrix of reach browser on each OS, but with older versions of these browsers to make sure we don't break support.

ukutaht:
@RobertJoonas I know it took you longer than you hoped but the result is great 👍 These tests look really good and readable to me, nicely done.

> Out of curiosity: is this going to run for every deploy?

Good question. I think the ideal logic would be that if a file in the `priv/tracker` folder has changed, then they should be run. If such precision is not possible then we can also just run them unconditionally. I've considered creating a separate repository for the tracker JS but they also have to be served from the phoenix app and I didn't want to mess around with git submodules.

@RobertJoonas the next step would be hooking these tests up to Github Actions so that these tests are executed automatically when we merge a pull request, for example. You'll need to create a file in the `.github/workflows` folder that installs the required dependencies and runs the tests. The current NPM one is the closest to what you need. [Here's](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) some documentation and I'm sure there are blog posts that describe using Github Actions with Playwright. If you run into trouble, feel free to ask @vinibrsl to see if he can help :)

RobertJoonas:
Thanks for the review @vinibrsl and @ukutaht ! 

> It's a bit weird to me that Safari on `OSX` is missing from this list.

True. I changed it to Windows because `playwright-webkit` browser on `OSX` couldn't open links in a new tab. There's a Github issue on this https://github.com/microsoft/playwright/issues/11595 which is marked completed, but it still seems broken to me. On Windows it worked nicely though. I'll poke around in that same issue to see if they can give us any update on that.

Good points about the browser version coverage in general. Testing on more browsers definitely makes sense, and it shouldn't affect the speed that much as I think they will run concurrently in separate workers on BrowserStack.

RobertJoonas:
Wow! I can't believe I didn't think of it before to check the playwright version. For some reason it was a quite old release that we initially installed. Opening links in new tabs works all nicely now with a later version! So removing the Playwright issue mention from here.

@ukutaht I'll go ahead and start testing more combinations of OS-s and browser versions like you mentioned. Once that's ready, I'll try and hook it up with Github actions.

ukutaht:
Perfect, thanks @RobertJoonas!

RobertJoonas:
Hey guys! Some updates on my progress with this. I had no issues with testing on latest browser versions. Everything is running smooth there. But ...

### Problems with testing older browser versions

According to BrowserStack docs, the oldest versions possible to test on BrowserStack with Playwright are:

	14.2 for Safari (from November 2020)
	87.0 for Firefox (from March 2021)
	83.0 for Chrome (from May 2020)
	
A browser version is actually selected through a Playwright version, as I've understood. [Here](https://www.browserstack.com/docs/automate/playwright/playwright-browser-compatibility) is the full list of Playwright versions supported by BrowserStack, and the browser versions that they are using.

Now, when you look at the options for `browser` in this table [here](https://www.browserstack.com/docs/automate/playwright/browsers-and-os), it says that you can use both "Playwright's bundled browsers" (such as playwright-firefox) and branded browsers (such as Chrome). But still, if you give a browser version, say 85 for Chrome, then according to [this](https://www.browserstack.com/docs/automate/playwright/select-playwright-version#Bundled_browsers), a "Playwright version that supports this browser is selected". Sounds like a conflict to me?

Chrome 83 and above supported:
image

And the earliest supported Playwright version supports Chrome 90.0:
image

Still, I've managed to use Chrome 86. Not sure why, and what Playwright version it uses, but it works and the tests pass.

Anyways, I've had a hard time understanding how this browser versioning works. I've tried running the tests with the earliest versions possible, but they fail. And I think it's not because our code is not working on the specific older browser version but because Playwright simply fails to execute some part of the code. A couple of errors I've seen with older Playwright versions:

- event request is sent but it has got no payload (cannot read property n of null) - tested on BS Live and it works there with the same browser/os
- tracing.stop: Must start tracing before stopping
- Cannot open link in new tab on Mac

So the way it seems to me is that the older versions of playwright (needed to run older browser versions) are buggy and do not work on BrowserStack as I'd expect. I've also reached out to BrowserStack support, asking for some clarification on this. Haven't heard back yet.

What I've done now is to use the earliest browser/playwright versions possible for all tests to pass. These are:

	Chrome 86 (from October 2020)
	Firefox 96 (from January 2022)
	Webkit 15.4 (latest)

Looking forward to hearing thoughts from either one of you guys, @ukutaht @vinibrsl, thanks!

vinibrsl:
I paired with @RobertJoonas and we pushed the changes adding test harness conditionally to CI, and fixing the request mock. Tests are passing :)

image

We just need to change the Browserstack secret keys to something more general. It's currently using my work account keys. I'm guessing as you're admin on Browserstack you have access to a CI key or something like this, @ukutaht? CI keys are on plausible/analytics repo > Settings > Secrets > Actions.

Other than that, this is ready to be merged.

ukutaht:
I've added my own keys for browserstack

ukutaht:
Thanks @vinibrsl @RobertJoonas ❤️

PR #1992, opened 2022-07-06
Add option to configure sentry pool size, and geolix worker size

manusajith:
PR add option to configure:

* max hackney pool size for Sentry
* worker size for Geolix


ukutaht:
Thanks!

PR #1997, opened 2022-07-08
Upgrade to Geolix 2.0

aerosol:
### Changes

This PR upgrades Geolix dependency to [release 2.0](https://github.com/elixir-geolix/geolix/releases/tag/v2.0.0).

The main benefit for plausible is the pooling layer removal. That's not really a breaking change in terms of the API the library exposes. 

Previous Geolix version used to spawn a group of processes managed by [poolboy](https://github.com/devinus/poolboy) to query the underlying storage (an ets table holding a binary dump of the whole IP database) with checkout/checkin semantics. This is completely unnecessary, as it only creates a bottleneck. Plausible queries the IP DB on virtually every request leading to congestion (the number of request processes in flight naturally exceeds the number of available pool workers, making the checkouts costly as the traffic peaks). 

In terms of isolated lookup performance there is no significant difference between the two Geolix releases. The gain should be apparent only under load.

```elixir
iex(1)> :ets.info :geolix_mmdb2_storage_geolocation
[
  id: #Reference<0.63578542.2420244481.168329>,
  decentralized_counters: false,
  read_concurrency: true,
  write_concurrency: false,
  compressed: false,
  memory: 438,
  owner: #PID<0.723.0>,
  heir: :none,
  name: :geolix_mmdb2_storage_geolocation,
  size: 1,
  node: :nonode@nohost,
  named_table: true,
  type: :set,
  keypos: 1,
  protection: :protected
]
```

The ets table used is `:protected` meaning that any process can read it directly. 

In Geolix, both IP and metadata lookups are routed through the `adapter` interface (e.g. https://github.com/elixir-geolix/geolix/blob/0b9d99ac421e40c1818f1d933fd8b76c5f31099f/lib/geolix.ex#L332) that ends up [fetching the whole database binary directly from ets](https://github.com/elixir-geolix/adapter_mmdb2/blob/563ac2451d423eba444d6aadb06c149508f4bcd6/lib/mmdb2/storage.ex#L40-L53) for further processing in the [codec](https://github.com/elixir-geolix/mmdb2_decoder) where the actual processing time is spent.

Optionally, a [caching layer](https://github.com/elixir-geolix/adapter_lookup_cache) can be put on top of that, to respond faster on subsequent individual IP addresses lookups. Initially I wanted to give it a go, since Plausible already has the [Cachex dependency required](https://github.com/elixir-geolix/adapter_lookup_cache/blob/935d2db3fa3199321b8eb74cad49a17ff13292bc/lib/lookup_cache/cache_adapter/cachex.ex), but it's unclear whether it's the right thing to do just yet (pending some metrics instrumentation) after release this PR goes live.

Additionally, the PR fixes [improper Geolix invocations](https://github.com/plausible/analytics/commit/387666e3f59026c8b40d9cc57827f87253be322a) leading to unnecessary parallel map over a single element. 

### Tests

### Changelog

### Documentation

### Dark mode


aerosol:
Rebased.

aerosol:
Looking into CI failures...

ukutaht:
Thanks @aerosol!

PR #1999, opened 2022-07-08
Setup promex

manusajith:
### Changes

Setup promex for gathering application insights

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/prom_ex.ex
```diff
@@ -0,0 +1,102 @@
+defmodule Plausible.PromEx do
+  @moduledoc """
```
ukutaht, line 2: This reads like generic Promex documentation which mostly is implemented in this PR. Is there any value in leaving this @moduledoc in?

lib/plausible/prom_ex.ex
```diff
@@ -0,0 +1,102 @@
...
+  @impl true
+  def dashboards do
+    [
+      # PromEx built in Grafana dashboards
+      {:prom_ex, "application.json"},
+      {:prom_ex, "beam.json"},
+      {:prom_ex, "phoenix.json"},
+      {:prom_ex, "ecto.json"},
+      {:prom_ex, "oban.json"}
+      # {:prom_ex, "phoenix_live_view.json"},
```
ukutaht, line 94: Not a fan of these comments. If we want to add stuff later we can just look up the docs for Promex, no?

ukutaht (commented):
Looks good. I would just cut down on some of the generic comments in the Promex module. Seems like leftover from copying an example file?

lib/plausible/prom_ex.ex
manusajith: The file was automatically generated using the generator `mix prom_ex.gen.config ... `

Just cleaned it up.

PR #2000, opened 2022-07-08
feat(tracker): navigator.sendBeacon support

kolodziejczak-sz:
### Changes

This PR introduces `navigator.sendBeacon` support. The API is not supported by Internet Explorer, I think the browser is already dead but since you did provide sort of polyfill for `URL API` I added this piece of code in all variants except `compat`.

This lets you to omit these timeouts which slows down navigation to the next page:

https://github.com/plausible/analytics/blob/2b8e3ea62af5972d5e2a8e5b301eef438fdae922/tracker/src/plausible.js#L122-L130

You can read more details about that api here: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

Unfortunately I'm not an Elixir expert, so I would be grateful if you could test this in some way.

Thanks

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- Add `navigator.sendBeacon` support in all variants except `compat`
 - The `exclusions` script extension now also takes a `data-include` attribute tag
 - A `file-downloads` script extension for automatically tracking file downloads as custom events
 - Integration with [Matomo's referrer spam list](https://github.com/matomo-org/referrer-spam-list/blob/master/spammers.txt) to block known spammers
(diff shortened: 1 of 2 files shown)


RobertJoonas:
Thanks for the PR @kolodziejczak-sz!

We've had some internal discussion on `sendBeacon` and for now have decided not to switch to it in the current version of the Plausible script. It's because we're trying to keep it as simple as possible, and adding Beacon into the mix would complicate things.

I think it would be better to introduce Beacon in a new, more modern tracking script version, where it could be the only technology used for sending all events. Then, people (like you) who believe that IE is dead, could simply run that script on their site.

Good point about the timeouts slowing down navigation. In fact, to tackle this, I'm currently working on the exact same stuff, rebuilding most of it, using callbacks to minimize the delay.

willfarrell:
Related: https://github.com/plausible/analytics/discussions/190 & https://github.com/plausible/analytics/pull/1730

PR #2011, opened 2022-07-12
Reset goal stats

vinibrsl:
### Changes

This pull request adds new button in Site > Settings > Goals to reset goals by deleting all related events. This was suggested in https://github.com/plausible/analytics/discussions/186.

Screen Shot 2022-07-12 at 7 54 04 PM

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated (link: https://github.com/plausible/docs/pull/247)

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ All notable changes to this project will be documented in this file.
 - Add ability to import historical data from GA: plausible/analytics#1753
 - API route `GET /api/v1/sites/:site_id`
 - Hovering on top of list items will now show a [tooltip with the exact number instead of a shortened version](https://github.com/plausible/analytics/discussions/1968)
+- Add ability to [reset goal stats](https://github.com/plausible/analytics/discussions/186)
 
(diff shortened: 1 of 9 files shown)


ukutaht:
This was a fun exercise, but I would like to discuss first whether hard deletes is the right way here. The use-case for deleting data is much broader than this PR:

People often want to remove tracking from a page whose URL was supposed to stay secret, for example. Or maybe a bot generated a bunch of bad data that needs to be cleared. Or maybe they sent data from a dev environment and want to clear that.

Currently we do hard deletes but giving people the power to flexibly remove data is dangerous with hard deletes. I can imagine our support being swarmed by people wanting to recover data they accidentally deleted. It should be possible to UNDO a delete for a while after it was run.

Beyond resetting goals, the more common use-case would be:
1. Filter your dashboard by some filters
2. Remove only the data that matches those filters

Basically, I think we need to implement soft deletes before shipping this feature.

PR #2030, opened 2022-07-18
Keep user filter when listing cities, countries, and regions stats 

### LINKED ISSUE #1982: Cities list features other cities too even when filtered by a city
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

When filtering by a city, everything works fine and all stats are filtered correctly. The only thing is that we still display the other top cities in the cities list itself. On countries and regions, we isolate only the country and region that the dashboard is filtered by and we do not include any other names so we should follow that for cities too. 

### Expected behavior

Cities list should only show the city that the dashboard is filtered by kind of like it works for countries and regions

### Screenshots

_No response_

### Environment

```markdown
- OS:
- Browser:
- Browser Version:
```

vinibrsl:
### Changes

This commit fixes a bug where location filters were filtering stats but not the locations list. This was caused by a `Map.put/3` call that overrides the user filter. I rolled back 5b571432737fb121e1b2a6c9e977500cb70f8880 changes and removed the `Map.put/3` call.

### Tests

### Changelog

### Documentation

### Dark mode

Closes #1982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ All notable changes to this project will be documented in this file.
 - Fix SecurityError in tracking script when user has blocked all local storage
 - Prevent dashboard graph from being selected when long pressing on the graph in a mobile browser
 - The exported `pages.csv` file now includes pageviews again [plausible/analytics#1878](https://github.com/plausible/analytics/issues/1878)
+- Fix a bug where city, region and country filters were filtering stats but not the location list
 
 ### Changed
 - Cache the tracking script for 24 hours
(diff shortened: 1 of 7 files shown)


vinibrsl:
> Thanks @vinibrsl. I actually looked at this one already. Some comments:
> 
> 1. The same issue appears for regions and countries. They should be fixed together.
> 2. I considered fixing with `put_new` but looking at the original commit made me lean a different direction: [5b57143](https://github.com/plausible/analytics/commit/5b571432737fb121e1b2a6c9e977500cb70f8880)
> 
> I think the original commit is applying the DRY principle too much, causing the bug, and should be reverted in full.
> 
> As you saw, there are tests missing for cities as well :) Thanks for adding them

Thanks for the clarification! I agree with rolling back those changes to keep it simple. I added test cases for regions and countries as well.

ukutaht:
Fantastic, thank you!

PR #2034, opened 2022-07-19
International Domain Names (IDN) Support

vinibrsl:
### Changes

This PR adds support for IDNs as discussed in #993. I replaced the latin alphanumeric validation with a more broad one, supporting other alphabets. I also made some changes to URI encoding to make sure our URLs are still good looking.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/schema.ex
+++ b/lib/plausible/site/schema.ex
@@ -43,14 +43,15 @@ defmodule Plausible.Site do
     site
     |> cast(attrs, [:domain, :timezone])
     |> validate_required([:domain, :timezone])
-    |> validate_format(:domain, ~r/^[a-zA-Z0-9\-\.\/\:]*$/,
+    |> clean_domain()
+    |> validate_format(:domain, ~r/^[-\.\\\/:\p{L}\d]*$/u,
       message: "only letters, numbers, slashes and period allowed"
     )
+    |> validate_domain_reserved_characters()
     |> unique_constraint(:domain,
       message:
         "This domain has already been taken. Perhaps one of your team members registered it? If that's not the case, please contact support@plausible.io"
     )
-    |> clean_domain
   end
 
   def make_public(site) do
@@ -113,5 +114,21 @@ defmodule Plausible.Site do
     change(changeset, %{
       domain: clean_domain
     })
+  end
+
+  # https://tools.ietf.org/html/rfc3986#section-2.2
+  @uri_reserved_chars ~w(: ? # [ ] @ ! $ & ' \( \) * + , ; =)
+  defp validate_domain_reserved_characters(changeset) do
+    domain = get_field(changeset, :domain) || ""
+
+    if String.contains?(domain, @uri_reserved_chars) do
+      add_error(
+        changeset,
+        :domain,
+        "must not contain URI reserved characters #{@uri_reserved_chars}"
+      )
+    else
+      changeset
+    end
   end
 end
--- a/lib/plausible_web/templates/site/settings_visibility.html.eex
+++ b/lib/plausible_web/templates/site/settings_visibility.html.eex
@@ -7,21 +7,21 @@
     <% end %>
   </header>
 
-	<%= if @site.public do %>
-		<div class="flex items-center mt-4 space-x-3">
-			<%= button(to: "/sites/#{URI.encode_www_form(@site.domain)}/make-private", method: "POST", class: "bg-indigo-600 relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring") do %>
-				<span class="inline-block w-5 h-5 bg-white rounded-full shadow translate-x-5 dark:bg-gray-800 transform transition ease-in-out duration-200"></span>
-			<% end %>
-			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%></a></span>
-		</div>
-	<% else %>
-		<div class="flex items-center mt-4 space-x-3">
-			<%= button(to: "/sites/#{URI.encode_www_form(@site.domain)}/make-public", method: "POST", class: "bg-gray-200 dark:bg-gray-700 relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring") do %>
-				<span class="inline-block w-5 h-5 bg-white rounded-full shadow translate-x-0 dark:bg-gray-800 transform transition ease-in-out duration-200"></span>
-			<% end %>
-			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%></a></span>
-		</div>
-	<% end %>
+  <%= if @site.public do %>
+    <div class="flex items-center mt-4 space-x-3">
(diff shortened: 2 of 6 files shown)


lib/plausible/site/schema.ex
```diff
@@ -41,14 +41,15 @@ defmodule Plausible.Site do
     site
     |> cast(attrs, [:domain, :timezone])
     |> validate_required([:domain, :timezone])
-    |> validate_format(:domain, ~r/^[a-zA-Z0-9\-\.\/\:]*$/,
+    |> validate_format(:domain, ~r/^[-.\\:\p{L}0-9]*$/u,
```
vinibrsl, line 44: I replaced `a-zA-Z` with `p{L}` which stands for Unicode letter (in any alphabet).

lib/plausible/site/schema.ex
```diff
@@ -112,4 +113,13 @@ defmodule Plausible.Site do
       domain: clean_domain
     })
   end
+
+  defp validate_domain_reserved_characters(changeset) do
+    domain = get_field(changeset, :domain) || ""
+    has_reserved_chars? = domain |> String.to_charlist() |> Enum.any?(&URI.char_reserved?/1)
```
vinibrsl, line 119: As we're being more permissive, I added this check to ensure we don't have any URI chars in domains (e.g. `/`, `&`, `?`)

lib/plausible/sites.ex
```diff
@@ -72,7 +72,7 @@ defmodule Plausible.Sites do
 
   def shared_link_url(site, link) do
     base = PlausibleWeb.Endpoint.url()
-    domain = "/share/#{URI.encode_www_form(site.domain)}"
+    domain = "/share/#{site.domain}"
```
vinibrsl, line 75: We don't need to encode the domain, as it does not contain any URI chars. Also, this prevents ugly URLs, e.g.:

```
/share/maçã.com -> /share/ma%C3%A7%C3%A3.com
```

Browsers handle diacritic in domains and URLs without having to encode, that is why I did not have to change this in the client-side.

lib/plausible_web/templates/site/settings_visibility.html.eex
```diff
@@ -12,14 +12,14 @@
 			<%= button(to: "/sites/#{URI.encode_www_form(@site.domain)}/make-private", method: "POST", class: "bg-indigo-600 relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring") do %>
 				<span class="inline-block w-5 h-5 bg-white rounded-full shadow translate-x-5 dark:bg-gray-800 transform transition ease-in-out duration-200"></span>
 			<% end %>
-			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%></a></span>
+			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> @site.domain %></a></span>
```
vinibrsl, line 15: Same URI encoding/diacritic comment.

lib/plausible_web/templates/site/settings_visibility.html.eex
```diff
@@ -12,14 +12,14 @@
...
 		</div>
 	<% else %>
 		<div class="flex items-center mt-4 space-x-3">
 			<%= button(to: "/sites/#{URI.encode_www_form(@site.domain)}/make-public", method: "POST", class: "bg-gray-200 dark:bg-gray-700 relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer transition-colors ease-in-out duration-200 focus:outline-none focus:ring") do %>
 				<span class="inline-block w-5 h-5 bg-white rounded-full shadow translate-x-0 dark:bg-gray-800 transform transition ease-in-out duration-200"></span>
 			<% end %>
-			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%></a></span>
+			<span class="text-sm font-medium text-gray-900 leading-5 dark:text-gray-100">Make stats publicly available on <a href="<%= plausible_url() <> "/" <> URI.encode_www_form(@site.domain)%>" class="text-indigo-500"><%= plausible_url() <> "/" <> @site.domain%></a></span>
```
vinibrsl, line 22: Same URI encoding/diacritic comment.

lib/plausible_web/templates/site/settings_visibility.html.eex
ukutaht: The domain needs to be encoded because we support 'domain names' like this: `ukutaht.github.io/website` to support Github Pages for example. Changing this would break a lot of dashboards.

lib/plausible/site/schema.ex
ukutaht: At least `/` should be allowed as explained in the other comment

ukutaht (changes requested):

lib/plausible_web/templates/site/settings_visibility.html.eex
vinibrsl: You're right! I created a function that encodes only slashes, which is the only URI-reserved character we support.

metmarkosaric:
@vinibrsl nice! for marketing/communication purposes: when this is merged, we will support any character for the domain name? or this adds support for some specific international characters only? basically need to know what to communicate to people

vinibrsl:
> @vinibrsl nice! for marketing/communication purposes: when this is merged, we will support any character for the domain name? or this adds support for some specific international characters only? basically need to know what to communicate to people

@metmarkosaric this pull request adds support for any Unicode character in the "letter" category, not any character. In other words, a letter in any language. Plus numbers, dots, and slashes.

RobertJoonas (approved):
Looks good to me!

ukutaht (changes requested):
The PR looks good overall!

The new encoding function is used in a couple places but there are many many more places where we encode domains. Try running `git grep 'URI.encode_www_form'` in the project to see all of them.

I need some convincing in the first place that the old function isn't going to work:
```elixir
iex(7)> URI.encode_www_form("wèbsite.com/wat")
"w%C3%A8bsite.com%2Fwat"
```

When I paste this into the browser, I get:
Screenshot 2022-09-26 at 16 43 04

Which looks correct to me. So maybe we don't need to re-do the encoding?

ukutaht:
> 

I've always thought the techincal debt we have is not using the Phoenix Route helpers which automatically url-encode your params. So instead of:

```elixir
    |> redirect(to: "/#{URI.encode_www_form(site.domain)}/settings/danger-zone")
```

We could do
```elixir
    |> redirect(to: Routes.site_path(conn, :settings_danger_zone, site.domain))
```

But I think the behaviour between them is exactly the same

vinibrsl:
> The PR looks good overall!
> 
> The new encoding function is used in a couple places but there are many many more places where we encode domains. Try running `git grep 'URI.encode_www_form'` in the project to see all of them.
> 
> I need some convincing in the first place that the old function isn't going to work:
> 
> ```elixir
> iex(7)> URI.encode_www_form("wèbsite.com/wat")
> "w%C3%A8bsite.com%2Fwat"
> ```
> 
> When I paste this into the browser, I get: Screenshot 2022-09-26 at 16 43 04
> 
> Which looks correct to me. So maybe we don't need to re-do the encoding?

You're right, we don't need the new encoding function for the URLs to work. However, there are two places where the escaped URL is displayed: (1) dashboard visibility toggle and (2) shared links. Here's how they look without the new function:

Screen Shot 2022-09-27 at 12 29 35

At least for the visibility toggle we should have some kind of escaping. I moved the new function to be used only in that view (d49fb6a29797f571453515967b6d9c17dba23400). Based on your suggestion I also changed Phoenix links and buttons to the generated `Routes` functions in that same view (31775fc54330b2128467972b227a9e68dedea06b). Here's how it's looking:

Screen Shot 2022-09-27 at 12 27 03

ukutaht (approved):

PR #2040, opened 2022-07-20
Provide a color coded donut chart for the Screen Size widget

eoghanmurray:
Instead of the inline bars

 - the mixed percentages / visitor totals at the right hand site of the bars were confusing for me; instead provide the percent info in a more visually obvious way (exact figures are in the donut tooltips)
 - there are only 4 possible screen sizes which suggests a good fit for a donut/pie chart - any more categories would be too much

Downsides/Todos:

 - the 4-color coding may conflict with other UI elements, e.g. yellow may be reserved to mean 'Pages'. Happy to take feedback on this
 - color is the only thing that maps each of the categories to the legend; this may be problematic to users with visual color impairment

### Changes

### Tests
TODO: Automated tests may need to be added

### Changelog

### Documentation
TODO: https://plausible.io/docs/guided-tour (and possibly others) will need to be updated when the visuals are finalized

### Dark mode
TODO: needs a tweak for dark mode for unused `faded = true`


metmarkosaric:
this looks great @eoghanmurray!

metmarkosaric:
For reference: this is how it looks like in GA4:

Screenshot from 2022-07-21 10-58-43
Screenshot from 2022-07-21 10-58-31

PR #2042, opened 2022-07-20
Add browser language stats

vinibrsl:
### Changes

This pull request adds a new field to store browser language: `preferred_language`. Browsers already send a list of user languages in a header, so nothing was changed in Plausible's script.

The header is [Accept-Language](https://httpwg.org/specs/rfc7231.html#header.accept-language). The client can have multiple languages, but only the topmost is saved, which is the preferred language. Language variation is stripped, e.g. `pt-BR` is transformed into `pt` to ensure it's a valid ISO 639 language code.

I still need to test this a bit more as I might have missed some edge cases. I'll write full documentation when this is reviewed and approved.

output

Also discussed in #865.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
This is a bigger chunk of work, we should first discuss where it should even be displayed becayse it could just as well be grouped with locations instead of devices.

Since it's not a huge priority compared to GA imports and such, I'll close the PR for now to keep things cleaner. We can resurrect this branch in the future if/when we decide to go for it.

PR #2046, opened 2022-07-22
Google Analytics Import Refactor

vinibrsl:
### Changes

This pull request refactors the Google Analytics Import feature. The main changes are:
- [Make parallel requests to Google Analytics Reporting API
](https://github.com/plausible/analytics/pull/2046/files#diff-bb3663885e04a0f5b0ce4e52c6aaa5bcc00626e08d95288864299b8d9f637b69R102)
- [Add buffering GenServer to reduce number of INSERT calls to Clickhouse](https://github.com/plausible/analytics/pull/2046/files#diff-ba12ab8ccbbc43e57db468da33042a79f12877eb3fd084dd6127d0f49090387bR1)
- [Change test strategy to use real Google Analytics data with `exvcr`, asserting number of imported pageviews and inserts per second](https://github.com/plausible/analytics/pull/2046/files#diff-26d4f7b722e115d3c0c53c178d3f4761fde2780680a48000d93ecdb90079ecdeR33)
- Use maps and structs to organize code instead of lists or positional arguments
- Separate HTTP calls from importing logic

### Tests

### Changelog
The changelog already had an unreleased entry for GA

### Documentation
- [X] [Docs](https://github.com/plausible/docs) have been updated (https://github.com/plausible/docs/pull/250)

### Dark mode


ukutaht (commented):
Looks really good @vinibrsl. I think the overall solution is a good one. I don't think the monthly requests were a good idea so the change you made is welcome.

We should play around with the buffer size a bit to find the optimal number there. Theoretically, if the buffer is big enough, it should do 1 insert per `imported_` table for the whole imported dataset. I'm not saying that's optimal, but it should be possible.

The bigger the batch, the longer it takes to serialize and insert it. If inserting takes too long, we would need backpressure and to introduce something like gen_stage as you mentioned. Before going there, I think we should be able to find a `max_buffer_size` that sends a reasonable amount of INSERTs and that can serialize stuff fast enough and avoid having to implement backpressure.

I would start with 10k like I said to see what impact it has on the number of INSERTs. Also, would be good to measure how long it takes to flush a buffer with 10k imported rows. If it takes <1s then there's no problem. If it takes 5 seconds, we might run into trouble where the importer is generating more data than the buffer can insert.

^^ Maybe things like this we should be monitoring in prod. What size are the buffers when they're flushed and how long does it take. We currently have both Promex and OpenTelemetry installed. Promex is probably the better option for now.

lib/plausible/google/buffer.ex
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  defp flush_all(%{buffers: buffers, site: site}) do
+    Enum.each(buffers, fn {dataset, records} -> do_insert_many(dataset, records, site) end)
+  end
+
+  defp max_buffer_size, do: 1000
+
+  defp do_insert_many(dataset, records, site) do
+    records = Plausible.Imported.from_google_analytics(records, site.id, dataset)
```
ukutaht, line 50: I would do this work in the API/importer layer before the buffer. Flushing the buffer should be as fast as possible in order to avoid the Genserver being overloaded. It's better to minimize data processing and serialization (i.e. looping over the whole buffer) in here

lib/plausible/google/api.ex
```diff
@@ -271,85 +132,70 @@ defmodule Plausible.Google.Api do
...
+           request.access_token,
+           request.view_id,
+           request.date_range,
+           request.dimensions,
+           request.metrics,
+           10_000,
+           request.page_token
+         ) do
+      {:ok, {rows, nil}} ->
+        Plausible.Google.Buffer.insert_many(buffer_pid, request.dataset, rows)
```
ukutaht, line 192: << see other comment about `Imported.from_google_analytics`

I would do it here

lib/plausible/google/api.ex
```diff
@@ -271,85 +132,70 @@ defmodule Plausible.Google.Api do
...
-      http_client.post(
-        "https://analyticsreporting.googleapis.com/v4/reports:batchGet",
-        Jason.encode!(%{reportRequests: [report_request]}),
-        [Authorization: "Bearer #{request.access_token}"],
-        timeout: 30_000,
-        recv_timeout: 30_000
-      )
+    case HTTP.get_report(
+           http_client,
+           request.access_token,
```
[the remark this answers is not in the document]
vinibrsl: Good call! I created the `Plausible.Google.ReportRequest` struct.

lib/plausible/google/buffer.ex
```diff
@@ -0,0 +1,53 @@
...
+    flush_all(state)
+    {:reply, nil, put_in(state, [:buffers], %{})}
+  end
+
+  def terminate(_reason, state) do
+    flush_all(state)
+  end
+
+  defp flush_all(%{buffers: buffers, site: site}) do
+    Enum.each(buffers, fn {dataset, records} -> do_insert_many(dataset, records, site) end)
```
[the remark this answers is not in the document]
vinibrsl: I added a `Process.sleep(1000)` when flushing.

More than 1 INSERT per second may happen as `Plausible.Event.WriteBuffer`, `Plausible.Session.WriteBuffer` and `Plausible.Google.Buffer` can run simultaneously. Should we be worried?

lib/plausible/google/buffer.ex
vinibrsl: Good call! I changed this :)

vinibrsl:
I pushed the changes from the code review and fixed tests. I wrote tests for the new buffer, but it is still missing GA tests. I'm planning tests to actually assert GA API is being requested properly, and records are being inserted. I'll play with VCR-like tests so we can be as close as possible to the actual thing.

lib/plausible/google/buffer.ex
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  defp max_buffer_size, do: 1000
```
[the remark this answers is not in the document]
vinibrsl: I increased the buffer size to 10,000 records. I think this is a good number as we are already using it in production buffers.

With a buffer size of 10,000, I could import 1.4 pageviews with 24 Clickhouse INSERT operations, opposed to 492 INSERT operations from the current strategy. I'm using `select count(*) from system.query_log where query_kind = 'Insert'` to assert this.

lib/plausible/google/buffer.ex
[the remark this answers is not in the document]
ukutaht: We also have 3 total nodes running. So even though the import buffer will only run on one of them, the `Event.WriteBufffer` and `Session.WriteBuffer` are running on all 3 nodes concurrently.

I don't think we should be worried. The 1 insert per second number is extremely safe, and the important thing is the average over a longer time period, it doesn't matter how they're spread out.

ukutaht:
Thanks Vini, great work so far!

config/runtime.exs
```diff
@@ -94,6 +94,11 @@ ch_db_url =
   |> get_var_from_path_or_env("CLICKHOUSE_MAX_BUFFER_SIZE", "10000")
   |> Integer.parse()
 
+{google_max_buffer_size, ""} =
```
ukutaht, line 97: There's a function called `get_int_from_path_or_env` which parses the integer for you

test/plausible/google/vcr_test.exs
```diff
@@ -0,0 +1,39 @@
...
+      access_token = "***"
+      view_id = "54297898"
+      date_range = Date.range(~D[2011-01-01], ~D[2022-07-19])
+
+      assert :ok == Plausible.Google.Api.import_analytics(site, date_range, view_id, access_token)
+
+      total_seconds = DateTime.diff(DateTime.utc_now(), before_importing_timestamp, :second)
+      total_inserts = get_insert_count() - inserts_before_importing
+
+      assert total_inserts / total_seconds > 1.0,
```
ukutaht, line 33: Wow, such a cool and creative test :)

ukutaht:
One question about the VCR cassettes: which GA property is this data from? I want to make sure we're not checking anything potentially sensitive into a public repository...

test/plausible/google/vcr_test.exs
vinibrsl: Thank you :)

ukutaht:
I've tested it with some large sites and it works as advertised. Thanks @vinibrsl, let's check how this works in production!

PR #1989, opened 2022-07-05
Support 'AM/PM' and '24h' timeformat in stats graph

sashaafm:
### Changes

This PR aims to add the feature requested [here](https://github.com/plausible/analytics/discussions/217). A `:timeformat` enumerated field was added to the `sites` table. This field defaults to `:'am/pm'` but it can be altered in the site's general settings.

### Tests
- [ ] Automated tests have been added
- [ ] This PR does not require tests

Not yet

### Changelog
- [ ] Entry has been added to changelog
- [ ] This PR does not make a user-facing change

Not yet

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [ ] This change does not need a documentation update

Not yet

### Dark mode
- [X] This PR does not change the UI

Below is a screen record showcasing this feature:
https://user-images.githubusercontent.com/3422400/177301701-da07ce35-1782-42d6-866f-2737535db635.mov

PR #2048, opened 2022-07-23
Show time format of graphs based on browser's language

sashaafm:
### Changes

As discussed in the previous [PR](https://github.com/plausible/analytics/pull/1989) a simpler and more pragmatic way of implementing this is to just present the time format based on the system language. From the research I've done it looks like reading the operation system's language isn't straightforward or implemented in all browser. Thus, I've implemented this by leveraging `navigator.language` which per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language#browser_compatibility) is compatible across all major browsers.

### Tests

### Changelog
- [X] Entry has been added to `CHANGELOG`

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/graph-util.js
+++ b/assets/js/dashboard/stats/graph/graph-util.js
@@ -10,13 +10,26 @@ export const dateFormatter = (interval, longForm) => {
     } else if (interval === 'date') {
       return formatDay(date);
     } else if (interval === 'hour') {
+      var uses12hTime = /h12/.test(
+        Intl.DateTimeFormat(navigator.language, { hour: 'numeric' })
+        .resolvedOptions()
+        .hourCycle
+      );
       const parts = isoDate.split(/[^0-9]/);
       date = new Date(parts[0], parts[1] - 1, parts[2], parts[3], parts[4], parts[5])
       var hours = date.getHours(); // Not sure why getUTCHours doesn't work here
-      var ampm = hours >= 12 ? 'pm' : 'am';
-      hours = hours % 12;
-      hours = hours ? hours : 12; // the hour '0' should be '12'
-      return hours + ampm;
+      if (uses12hTime) {
+        var ampm = hours >= 12 ? 'pm' : 'am';
+        hours = hours % 12;
+        hours = hours ? hours : 12; // the hour '0' should be '12'
+        return hours + ampm;
+      } else {
+        var suffix = "h";
+        if (hours >= 12) {
+          hours = hours % 12 + 12;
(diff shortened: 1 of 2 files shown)


sashaafm:
A possible case to consider here is if an older browser without `Intl` is used. Possibly fallback to a default format or attempt to use another API.

assets/js/dashboard/stats/graph/graph-util.js
```diff
@@ -10,13 +10,26 @@ export const dateFormatter = (interval, longForm) => {
...
+        hours = hours % 12;
+        hours = hours ? hours : 12; // the hour '0' should be '12'
+        return hours + ampm;
+      } else {
+        var suffix = "h";
+        if (hours >= 12) {
+          hours = hours % 12 + 12;
+        }
+        return hours + suffix;
+      }
```
vinibrsl, line 32: As `Intl.DateTime` will be used here, we could use it to format the hour as well. What do you think?

```
Intl.DateTimeFormat("pt-BR", { hour: 'numeric' }).format(new Date);
// 18

Intl.DateTimeFormat("en-US", { hour: 'numeric' }).format(new Date);
// 6 PM
```

sashaafm: @vinibrsl I agree that's a better way to achieve this! 

```javascript
      var uses12hTime = /h12/.test(
        Intl.DateTimeFormat(navigator.language, { hour: 'numeric' })
        .resolvedOptions()
        .hourCycle
      );
      const parts = isoDate.split(/[^0-9]/);
      date = new Date(parts[0], parts[1] - 1, parts[2], parts[3], parts[4], parts[5])
      var hours = Intl.DateTimeFormat(navigator.language, { hour: 'numeric' }).format(date);
      if (uses12hTime) {
        hours = hours.toLowerCase().replace(/\s/g, "");
      } else {
        hours = hours.concat("h");
      }
      return hours;

```

I appended the `h` to the end of the 24h time and lowercased the 12h time since I think the aesthetics fit better into the overall look of Plausible that way. Tell me what you think 😄

ukutaht:
Thank you @sashaafm

❤️ 🧡 💛 💚 💙 💜 🖤

ukutaht:
One thing: Both @vinibrsl and I prefer `15:00` over `15h` as the format, so I've just changed in on master to get it live quickly.

@metmarkosaric Any opinions? `15:00` vs `15h` for Europeans?

metmarkosaric:
15:00 for me too

PR #2054, opened 2022-07-25
Migrate HTTPoison to Finch

manusajith:


lib/plausible/google/api.ex
```diff
@@ -1,5 +1,5 @@
 defmodule Plausible.Google.Api do
```
ukutaht, line 1: @vinibrsl Worth for you checking out these changes as well. I think we'll merge this PR first and then the GA import one so there will be some conflicts.

lib/plausible/http_client.ex
```diff
@@ -0,0 +1,33 @@
...
+    call(:get, url, header, params)
+  end
+
+  defp call(method, url, header, params) do
+    method
+    |> build_request(url, header, params)
+    |> do_request
+  end
+
+  defp build_request(:post, url, header, params) do
```
ukutaht, line 26: Personally I don't see the value of wrapping these functions. It introduces a layer of indirection without making any of the code simpler or easier to read.

ukutaht (approved):

lib/plausible/google/api.ex
vinibrsl: Got it! Thanks for the heads-up.

vinibrsl (approved):

vinibrsl:
I was talking to @ukutaht about this change, and I think that https://github.com/elixir-tesla/tesla would be a better replacement for HTTPoison. Tesla supports [multiple adapters](https://hexdocs.pm/tesla/Tesla.Adapter.Finch.html), meaning that we could use Tesla + Finch. It ships with built-in support for test mocks without having to do dependency injection, and most importantly a [request retrying](https://hexdocs.pm/tesla/Tesla.Middleware.Retry.html) feature that can be really useful for us. The con is having another layer of abstraction, but I see more pros than cons.

What are your thoughts?

manusajith:
> I was talking to @ukutaht about this change, and I think that https://github.com/elixir-tesla/tesla would be a better replacement for HTTPoison. Tesla supports [multiple adapters](https://hexdocs.pm/tesla/Tesla.Adapter.Finch.html), meaning that we could use Tesla + Finch. It ships with built-in support for [test mocks](https://hexdocs.pm/tesla/Tesla.Mock.html) without having to do dependency injection, and most importantly a [request retrying](https://hexdocs.pm/tesla/Tesla.Middleware.Retry.html) feature that can be really useful for us. The con is having another layer of abstraction, but I see more pros than cons.
> 
> What are your thoughts?

I had considered `Tesla` and also `Req`

`Req` is still in early phases of development, so it could introduce breaking changes.

Tesla has too many abstractions imho, and the community is also somewhat shifting towards using Finch(for at least the newer code/libraries etc).

re: request retying, I am not sure some of the places where we use http libraries would benefit from retries.

re: tests/mocks - I was considering to use `bypass`, but did not add it to this PR, as I wanted to keep the changes specific to the adapter.

ukutaht:
> Tesla has too many abstractions imho

In general I'm quite averse to introducing new libraries and layers of indirection as well. I liked the Tesla suggestion because it would simplify testing and retries (which I think would be useful in many cases).

@manusajith Bypass seems to send stuff over an actual TCP connection? I really like that because it also exercises the client library itself as opposed to mocking it out. I have way more confidence in tests like that. The only thing we'd need to inject is the URL of the server to talk to which is a good idea to parameterize in any case. 

It's a cool suggestion, even if we use Tesla I would prefer to use something like Bypass to get more of the system under test. Then Tesla starts losing some of its value.

I suggested retries as a way to reduce timeout errors:

https://sentry.plausible.io/organizations/sentry/issues/640
https://sentry.plausible.io/organizations/sentry/issues/295
https://sentry.plausible.io/organizations/sentry/issues/1799

etc.

Network requests can never be expected to be 100% reliable. If we're updating someone's subscription for example, I think it's a better experience for the customer if we retry the request automatically instead of them seeing an error and having to retry manually. It would also fall under the umbrella of 'reducing sentry errors'. I don't think every single HTTP request should be retried, we need to decide case-by-case.

I've found that for the GA imports we already do [5 retries](https://github.com/plausible/analytics/blob/972ba2482b3538cc6976544eaaf4165eec474e43/lib/plausible/google/api.ex#L326) before giving up. Even if we don't change any behaviour, it would be nice to encapsulate that logic in the `Plausible.HTTPClient` so we can reuse it elsewhere.

So in the end, if we use Bypass, then the only thing we would get by using Tesla is retry logic which we already have implemented in ~5 lines of code. I don't think it's worth it.

lib/plausible/http_client.ex
lovebes: adding to this comment - how about using `defdelegate`?
e.g.)

```
defdelegate build_request(:post, url, header, params), to: Finch, as: :build
```

manusajith:
@ukutaht this is ready for another review now.

ukutaht:
Thanks @manusajith

PR #2056, opened 2022-07-26
Fix error handling for postmark.

manusajith:
Do not log errors to Sentry if the error code is 406(Recipient marked as inactive)

### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/mailer.ex
```diff
@@ -1,29 +1,24 @@
...
-          stacktrace: __STACKTRACE__,
-          extra: %{extra: "Error while sending email"}
-        )
+  defp handle_error(%Bamboo.ApiError{message: message} = error) do
+    case Jason.decode!(message.response) do
+      %{"ErrorCode" => 406} ->
+        {:error, error}
+
+      response ->
+        Sentry.capture_exception(response, extra: %{extra: "Error while sending email"})
```
vinibrsl, line 20: What do you think of using [`Sentry.capture_message/2`](https://hexdocs.pm/sentry/Sentry.html?#capture_message/2) with `response` as an extra?

It seems from the [`Sentry.capture_exception/2` typespec](https://hexdocs.pm/sentry/Sentry.html?#capture_exception/2) that it expects an [`Exception.t()`](https://hexdocs.pm/elixir/Exception.html#t:t/0).

ukutaht:
@vinibrsl's comment seems on point.

The two functions in that module used to do different things - one would reraise the error and the other would swallow it. Both are reported to Sentry though.

This PR keeps the two different function definitions, but removes the distinction between them (neither uses `reraise` anymore). In order to integrate this, I see one of two options:
1. Bring back the distinction and keep the old `reraise` behaviour
2. Verify that the change does not impact user experience negatively wherever the `send_email` without `safe` was used. If `reraise` doesn't add any value, then I'm OK with removing the distinction between these two functions. But that needs to be a bigger refactor because in the current state we have two functions that do exactly the same thing and that's very confusing to someone who is not familiar with the history. So if we remove the `reraise`, this module should have only one function, not two.

PR #2082, opened 2022-08-05
added a test and changed hashmode condition

RobertJoonas:
### Changes

The NPM package sends events with the `h: 0` option. Currently, the if statement only checks if the hash option is present, not its value. With this change, `h: 0` in the event payload will not track the hash part of the URL.

There's an issue and a PR in the tracker repo, to fix this on the tracker side (https://github.com/plausible/plausible-tracker/issues/42) but I think it makes sense to change this one in our app instead.

### Tests

### DIFF
--- a/lib/plausible_web/controllers/api/external_controller.ex
+++ b/lib/plausible_web/controllers/api/external_controller.ex
@@ -273,7 +273,7 @@ defmodule PlausibleWeb.Api.ExternalController do
       |> URI.decode()
       |> String.trim_trailing()
 
-    if hash_mode && uri.fragment do
+    if hash_mode == 1 && uri.fragment do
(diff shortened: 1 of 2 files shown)


ukutaht:
How does the JS tracker in this repo send the `h` parameter, and how does this backend change affect that?

RobertJoonas:
JS tracker in this repo does the following:

```
{{#if hash}}
    payload.h = 1
{{/if}}
```
So it either sends the `h` option with the value `1`, or doesn't send it at all.

This is how we evaluate the hash_mode from event options:

```
"hash_mode" => params["h"] || params["hashMode"]
```
If it's not present, the value is `nil` and `nil == 1` will give false as expected.

ukutaht:
Makes sense, thanks!

PR #2085, opened 2022-08-09
Process per IP ingestion strategy

vinibrsl:
### Changes

This commit creates a GenServer spawned for each incoming event IP. It caches event information to be used in subsequent events from the same IP. This makes the processing asynchronous, therefore, tests were changed accordingly. This is a breaking change as validation errors are no longer returned, returning HTTP 202 instead.

This is work in progress and feedback is more than welcome. After feedback I'll jump into writing tests for the new GenServer.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/ingestion/ingestion.ex
```diff
@@ -0,0 +1,513 @@
...
+
+            {:cont, :ok}
+          else
+            errors = Ecto.Changeset.traverse_errors(changeset, &encode_error/1)
+            {:halt, {:error, errors}}
+          end
+        end)
+
+      case response do
+        :ok -> {:ok, {ua, location_details}}
```
vinibrsl, line 100: The current implementation returns `:ok`, and this change returns the user agent and location details to be cached. This is not the best solution, but it works for now.

My proposal is to split this big function into many functions that transform the `Plausible.ClickhouseEvent` struct. Then we can cache fields from from the struct, and skip some functions when cached data is available. Check this comment for more https://github.com/plausible/analytics/pull/2084#issuecomment-1209421570.

This would be a separate PR :)

lib/plausible/ingestion/user.ex
```diff
@@ -0,0 +1,81 @@
...
+    state = %State{remote_ip: remote_ip} |> renew_grace_period()
+    {:ok, state}
+  end
+
+  @impl true
+  def handle_cast({:send_event, %Plausible.Ingestion.Request{} = request}, %State{} = state) do
+    Logger.debug("Ingestion: Processing new event for IP #{state.remote_ip}")
+
+    # NOTE: Should we queue requests to send mini-batches to the buffer? That would reduce the
+    # main buffers mailboxes. Something to think about.
```
vinibrsl, line 47: I'd like to hear your opinion on this, @ukutaht and @manusajith :)

lib/plausible/ingestion/user.ex
```diff
@@ -0,0 +1,81 @@
...
+      {:noreply, state}
+    end
+  end
+
+  # NOTE: A second timeout could be added to block annoying callers. This timeout would not reset
+  # when receiving events. For example a same IP sending events for hours straight could be added
+  # to a temporary blocklist.
+  #
+  # TODO: Change @grace_period to reasonable timeout
+  @grace_period :timer.minutes(1)
```
vinibrsl, line 75: The grace period, which is the time the process self destructs after not receiving events, is set to 1 minute. Correct me if I'm wrong but sessions are currently set to 30 minutes.

I'm not sure if this process should behave like a Plausible session. Having each process living for at least 30 minutes may result in too many alive processes under the DynamicSupervisor, and memory issues. I decided to leave as-is for now, and we can tweak on that later.

/cc @ukutaht

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -59,6 +74,7 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       conn =
         conn
         |> put_req_header("content-type", "text/plain")
+        |> put_ip_address()
```
vinibrsl, line 77: I added an IP address to each test request here, as now using the same IP address uses a cached user agent from the first request.

lib/plausible/ingestion/ingestion.ex
```diff
@@ -1,13 +1,23 @@
 defmodule Plausible.Ingestion do
   require OpenTelemetry.Tracer, as: Tracer
 
+  @type location_details :: %{
+          country_code: integer() | nil,
```
ukutaht, line 5: `country_code`, `subdivision1_code`, and `subdivision2_code` should be strings. `city_geoname_id` is an integer indeed

vinibrsl:
Checkpoint: I updated this branch with the ExternalController refactoring that is in `master`, and implemented a process per session strategy, removing `Plausible.Session.CacheStore`. I still have plenty of tests to fix as I deleted the CacheStore module. However, `ExternalController` tests are passing and that is a good thing :)

I'll work on making tests green and writing tests for the new Session actor.

ukutaht:
Just an overall observation: I really like all the naming here and the way you've factored the code, plus docs. Love seeing everything get cleaner and cleaner 🙏 Much appreciated.

I think the changes are quite significant and I would like to review this on a video call when we get a chance.

ukutaht:
Can we close this to reduce the amount of open PRs? Can always refer back to the code or reopen in but it has many conflicts with the ongoing ingestion pipeline work.

vinibrsl:
> Can we close this to reduce the amount of open PRs? Can always refer back to the code or reopen in but it has many conflicts with the ongoing ingestion pipeline work.

Fine by me! I'll leave the branch in the remote /cc @aerosol

aerosol:
Ah thanks for the heads up, I had completely forgotten about this, sorry. Let's revisit later.

Issue #1941, opened 2022-06-01
outbound-links script causes media lightboxes to open in new tab


ianmcilwraith:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

I have two sites (WordPress) where lightboxes are in use, and of course on both sites Plausible is enabled with the Outbound Links script. On one site, the lightbox is used for images, and on the second site for videos. In both cases, the behavior witnessed is that the lightbox appears for a quick second before opening the image/video in a new tab. I narrowed down the issue to the Plausible `outbound-links.js` script causing this issue.

This is the relevant section of the `outbound-links.js` script that was highlighted by the browser debugger as the last step before the redirection happened.
`(setTimeout(function(){a.href=e.href},150),t.preventDefault()))`

This behavior does not happen with just the `plausible.js` script.
[This is the second site](https://cslewismovie.com) mentioned — clicking on the "Play Trailer" button will open a popup window for a second before redirecting to the linked video. The outbound links script is enabled on this site.

Happy to provide further debugging info as needed.

### Expected behavior

The expected behavior is simply that the media opens in a lightbox and doesn't redirect to the media URL.

### Environment

```markdown
- OS: macOS Monterey
- Browser: Firefox
- Browser Version: 100.0.2

Same behavior happens in Safari on iOS, Safari on macOS, Chrome on macOS, likely all browsers.
```

ianmcilwraith:
@RobertJoonas Thank you for the fix for this issue in #1938. Any idea on when it will be pushed to master?

RobertJoonas:
@ianmcilwraith, I'm now finalizing this bug fix in https://github.com/plausible/analytics/pull/2109. I wanted to ask if the issue was really in opening the lightbox link in a **new tab**? I tested on Ubuntu/Firefox + macOS Monterey/Firefox and from what I witnessed, the link opened in the **same tab**. I cannot think of any way that the current outbound-links script would cause the link to open in a new tab, as Plausible will not intercept when the link is set to open in a new tab. Can you please confirm?

ianmcilwraith:
@RobertJoonas I just tested again and it the Lightbox link was opening in the same tab, not a new tab. Apologies if I misspoke in the initial comment!

[...]

PR #2109, opened 2022-08-16
refactored outbound links and file downloads

RobertJoonas:
I opened a new PR instead of continuing https://github.com/plausible/analytics/pull/1938, as I think it makes sense to start from scratch with the refactoring before moving on to the `tagged events` extension.

### Changes

- added a Handlebars helper function `any` to represent an `or` statement for multiple script extensions
- broke functionality down to several functions to avoid code duplication
- created an `inFlightRequests` queue to avoid conflict when multiple Plausible events are sent at the same time (for example, a link can be both outbound and a file download)
- to continue default behavior, switched to using callbacks instead of waiting a fixed amount of time for requests to go through
- Stopped continuing the default link click behavior when the default has already been prevented by an external script. That is, if `event.defaultPrevented` is already `true` (this is the fix for https://github.com/plausible/analytics/issues/1941)

### Tests

### Changelog

### Documentation
- [x] this PR does not need a documentation update

### Dark mode


tracker/src/plausible.js
```diff
@@ -231,4 +149,96 @@
...
+          sendCustomEvent(eventName, eventProps, callbackFn)
+      } else {
+          sendCustomEvent(eventName, eventProps, eventCallback(eventName, null))
+      }
+  }
+
+  function shouldFollowlink(link, event) {
+      var targetsCurrentWindow = !link.target || link.target.match(/^_(self|parent|top)$/i)
+      var isRegularClick = event.type === 'click' && !(event.ctrlKey || event.metaKey || event.shiftKey)
+      return targetsCurrentWindow && isRegularClick && (!event.defaultPrevented || plausiblePreventedDefault)
```
ukutaht, line 181: This line breaks by brain. Can you give a name to the `(!event.defaultPrevented || plausiblePreventedDefault)` expression?

ukutaht: It may eve worth adding comments/documentation here to explain:
1. Why the event.defaultPrevented check is necessary
2. Why the `plausiblePreventedDefault` check is necessary

Seeing as our current `master` branch does not check for either, it means the code changes here also change behaviour. If at all possible, the changes in behaviour should be tested. Can you come up with tests for `event.defaultPrevented` and `plausiblePreventedDefault`?

tracker/src/plausible.js
```diff
@@ -231,4 +149,96 @@
...
+          delete inFlightRequests[eventName]
+
+          if (continueFn && Object.keys(inFlightRequests).length == 0) {
+              continueFn()
+          }
+      }
+  }
+
+  function sendCustomEvent(eventName, eventProps, callbackFn) {
+      inFlightRequests[eventName] = true
```
ukutaht, line 195: Consider the case where a second request is fired for the same `eventName` before the first request can finish. In other words - what happens if `inFlightRequests[eventName]` is already true at this point? It means there's an existing in-flight request for `eventName`. Starting a new one would create 2 requests, 2 pending callbacks but just 1 entry in `inFlightRequests`. I *think* that would lead to an error when the second callback finishes.

Concurrent, async programming is complicated....

Maybe a simple fix would be to just prevent having more than 1 in-flight request for each `eventName`.

tracker/src/plausible.js
```diff
@@ -231,4 +149,96 @@
...
+
+  function sendLinkClickEvent(link, event, eventName, eventProps) {
+      if (shouldFollowlink(link, event)) {
+          event.preventDefault()
+          plausiblePreventedDefault = true
+          var callbackFn = eventCallback(eventName, function() { window.location = link.href })
+          setTimeout(callbackFn, 5000)
+          sendCustomEvent(eventName, eventProps, callbackFn)
+      } else {
+          sendCustomEvent(eventName, eventProps, eventCallback(eventName, null))
```
ukutaht, line 174: Is there any possibility that the `eventCallback` here will never be called? What if the request times out? What if it never finishes? What is has a DNS or some other weird error?

I think it makes sense to add a timeout here as well, even if we don't follow the link, to make sure we don't leave any dangling zombie entries in `inFlightRequests`

ukutaht (commented):
Looking good!

From my experience, a queue like `inFlightRequests` is more complex to manage than you think, especially in an async environment. I'd like the different scenarios to be verified and tested before we roll this out to all the customers. We should think of every way to mess up `inFlightRequests` and `event.defaultPrevented` behaviour and write a test for it.

tracker/src/plausible.js
RobertJoonas: > I'd like the different scenarios to be verified and tested before we roll this out to all the customers. We should think of every way to mess up inFlightRequests and event.defaultPrevented behaviour and write a test for it.

I've now added two tests to test the `inFlightRequests` queue by sending two events with the same click action. One where navigation is expected, and another where default is externally prevented and no navigation should happen.

I also added some comments to explain why `event.defaultPrevented` and `plausiblePreventedDefault` are necessary.

tracker/src/plausible.js
RobertJoonas: Yeah, that's a good point.

> Maybe a simple fix would be to just prevent having more than 1 in-flight request for each eventName.

From the customers perspective, I think you wouldn't want to send more than one same event at the same time. So that would make sense to do. I've now added a check that ignores the event if there's already an event in flight with the same name.

tracker/src/plausible.js
RobertJoonas: Good point. Fixed it.

tracker/test/support/test-utils.js
```diff
@@ -14,6 +14,24 @@ exports.mockRequest = function (page, path) {
   })
 }
 
+// Mocks a specified number of Plausible event requests. Returns a promise that resolves to a list
+// of event names as soon as the specified number of requests is made, or 10 seconds has passed.
+exports.mockManyPlausibleRequests = function(page, path, numberOfRequests) {
```
vinibrsl, line 19: I like that you decided to make this a separate function. I think it makes more sense :)

vinibrsl, line 19: As this is mocking Plausible requests only, we can probably omit the `path` argument, and replace it with `"/api/event"`.

vinibrsl (approved):

ukutaht:
Happy with the code.

It's a big change to a very sensitive and depended on piece of code. What can we do to verify that these changes:
1. Fix the bug that we expect it to fix
2. Don't break existing functionality?

Before and after pushing this live?

tracker/test/support/test-utils.js
RobertJoonas: Makes sense, removed the argument.

RobertJoonas:
> It's a big change to a very sensitive and depended on piece of code. What can we do to verify that these changes:

Agreed, @ukutaht

> 1. Fix the bug that we expect it to fix

I tested the new code via the page console on the [site](https://cslewismovie.com/) shared in [the issue](https://github.com/plausible/analytics/pull/2109). My testing workflow was the following:

1. Specifically disable loading the outbound-links script with an adblocker
2. Refresh the page and confirm that `script.outbound-links.js` fails to load
3. Install the Plausible snippet and the `plausible()` function via the dev tools console
4. Copy all the new code in the console to add the eventListeners
5. Left-click the same link which opens in the same tab with `script.outbound-links.js`

Tested environments:

- Ubuntu/Firefox
- macOS Monterey/Firefox (on BrowserStack Live)

The result was as expected - the outbound link event was sent, and unlike before, the lightbox opened nicely in a popup.

I just read through the issue once again. The customer reported that the link opens in a **new tab** rather than the **same tab** which was actually the case in my testing. I've now asked for their confirmation about that here: https://github.com/plausible/analytics/issues/1941#issuecomment-1232773044

> 2. Don't break existing functionality?

These changes currently only change outbound link and file download tracking and cannot break any other tracker functionality. The BrowserStack tests cover the core part of sending these custom events, but I guess it would also make sense to do some more manual testing on different environments with BrowserStack Live.

> ... after pushing this live?

Not sure if there's any other way than waiting for user feedback?

RobertJoonas:
Closing this PR, as started from a clean slate in https://github.com/plausible/analytics/pull/2208 with smaller commits.

PR #2116, opened 2022-08-17
Remove static tracker files

RobertJoonas:
Co-authored-by: Vinicius Brasil <vini@hey.com>

I overwrote this branch with a cleaner one. Reviewing commit by commit suggested :)

### Changes

Deletes all static `priv/tracker/js/*.js` files from Git. Instead auto-generates them in different environments when needed.

runs `npm run deploy` in the tracker folder before:

- running npm test locally
- running npm test on CI
- ~~running elixir tests locally~~ (we'll do it manually instead, as it helps save 4-5 seconds on every `mix test`)
- running elixir tests on CI
- running mix phx.server

Running on production is [already implemented](https://github.com/plausible/analytics/blob/master/Dockerfile#L39).

### Tests

### Changelog

### Documentation

### Dark mode


mix.exs
```diff
@@ -124,7 +124,15 @@ defmodule Plausible.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate", "test", "clean_clickhouse"],
-      sentry_recompile: ["compile", "deps.compile sentry --force"]
+      sentry_recompile: ["compile", "deps.compile sentry --force"],
+      npm_deploy: &npm_install_and_deploy/1
```
vinibrsl, line 128: What do you think of moving this command alias to `package.json`? CI can use npm to run install and deploy, and we keep `mix.exs` for Elixir-specific commands.

.github/workflows/elixir.yml
```diff
@@ -60,6 +60,8 @@ jobs:
         run: mix format --check-formatted
       - name: Check Compile Warnings
         run: mix compile --warnings-as-errors --all-warnings
+      - name: Generate tracker files
+        run: mix npm_deploy
```
ukutaht, line 64: `npm_deploy` could refer to both the `assets` folder or the `tracker` folder. Maybe a more specific name would be better like `mix compile_tracker`

ukutaht: BTW Is this needed in the `elixir` CI job? I don't think we actually use the tracker files in test.

I think this needs to be triggered in the Dockerfile to make the tracker files available in the deployed image.

RobertJoonas: > BTW Is this needed in the elixir CI job? I don't think we actually use the tracker files in test.

Tracker files are needed in these 3 tests: https://github.com/plausible/analytics/blob/master/test/plausible_web/plugs/tracker_test.exs

> I think this needs to be triggered in the Dockerfile to make the tracker files available in the deployed image.

I believe we're already doing it here: https://github.com/plausible/analytics/blob/master/Dockerfile#L39

vinibrsl (approved):

mix.exs
```diff
@@ -123,7 +123,13 @@ defmodule Plausible.MixProject do
     [
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
-      test: ["ecto.create --quiet", "ecto.migrate", "test", "clean_clickhouse"],
+      test: [
+        "cmd npm run deploy --prefix ./tracker",
```
ukutaht, line 127: Does this slow down the test suite?

ukutaht: Maybe a better question: how long does this take and does it run every time we run the test suite?

RobertJoonas: The more script extensions we have, the longer it takes. I think it's currently around 4-5s. So I agree, that spending this time with every `mix test` is not feasible.

To not generate the tracker files every time, we'll have to do it manually instead. I've done this change in https://github.com/plausible/analytics/pull/2116/commits/0a6d91e738b52555d85d0168627b8be0b7c600dc where I also documented this in `tracker_test.exs`

ukutaht:
Have you tested whether the final Docker image works correctly?

One way to verify would be:
1. Deploy this PR to the staging.src.plausible.io server
2. Hit it with `curl` to make sure all the tracker files are still available

I'm asking because the build steps are different for local development vs prod Docker image. Need to make sure both work

ukutaht:
> 

I know you didn't change how the Dockerfile is build but still worth testing I think

RobertJoonas:
> Have you tested whether the final Docker image works correctly?
> 
> One way to verify would be:
> 
>     1. Deploy this PR to the staging.srv.plausible.io server
> 
>     2. Hit it with `curl` to make sure all the tracker files are still available
> 

We tested it with @vinibrsl. Deployed to staging and the files were generated nicely. We're planning to deploy it tomorrow.

ukutaht:
Nice, let's do it!

PR #2126, opened 2022-08-22
Update site footer to include tools

metmarkosaric:
Updating the site footer to include the "Tools" heading underneath "Comparisons". This is untested so best to double check how it looks. Ideally there should be a blank space underneath "vs Cloudflare" before the "Tools" heading starts so it all is aligned with other columns. Thanks!

### DIFF
--- a/lib/plausible_web/templates/layout/_footer.html.eex
+++ b/lib/plausible_web/templates/layout/_footer.html.eex
@@ -90,6 +90,24 @@
             </ul>
           </div>
         </div>
+        <div class="mt-32 md:mt-0">
+            <h4 class="text-sm font-semibold tracking-wider text-gray-400 uppercase leading-5">
+              Tools
+            </h4>
+            <ul class="mt-4">
+              <li>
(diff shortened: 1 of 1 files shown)


ukutaht:
I haven't tested how it affects the footer yet. Another option would be adding it to the bottom of this dropdown to make it contextual. Thoughts?

Screenshot 2022-08-22 at 15 48 44

metmarkosaric:
Could be cool to integrate it better with the product in the future. Would be best with some type of standalone, popup design of the UTM tag builder in that case so we don't take people away from the dashboard. Could even integrate with the top pages report so people can just pick one of the pages from their Top Pages and add tags.

ukutaht:
Screenshot 2022-09-06 at 12 49 45

How about this for now, and integrating more deeply later?

metmarkosaric:
looks good @ukutaht!

ukutaht:
Closing in favour of https://github.com/plausible/analytics/pull/2195

PR #2152, opened 2022-08-30
Build Docker images for each PR

vinibrsl:
### Changes

This pull request creates a GitHub action triggered for each pull request labeled `deploy-to-staging` that builds and pushes a Docker image.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
What's the rationale for creating a new workflow instead of adding another trigger for the current one? 

https://github.com/plausible/analytics/blob/master/.github/workflows/docker.yml

vinibrsl:
> What's the rationale for creating a new workflow instead of adding another trigger for the current one?
> 
> https://github.com/plausible/analytics/blob/master/.github/workflows/docker.yml

I initially thought of updating the current one, but the workflow file ended up a bit complicated. IMO they should be separate files because:

- Staging workflow pushes to [another Docker Hub repository](https://hub.docker.com/r/plausible/analytics-staging/tags), to avoid pushing staging images to the main repository, which serves production and self-hosted images. As this for internal testing, this could be a private repository as well.
- Staging workflow reacts to pull requests, and not pushes. That is because we want to start the workflow when the labeled pull request code is changed, and also when labels change.
- Staging workflow image tags are labeled using PR reference (e.g. `plausible/analytics-staging:pr-1234`), as GH Actions push events and PR events have different available data. I didn't spend much time debugging this, but `docker/metadata-action@v4` failed with the existing workflow tag generation, so I used `type=ref,event=pr` works for this case.

ukutaht:
Thanks,

> Staging workflow pushes to another Docker Hub repository, to avoid pushing staging images to the main repository, which serves production and self-hosted images.

I like the separation here. If we're going to separate it should be:
* Public repository for `v1.5.0` and tags like this. Only public self-hosted releases.
* Private repository for `master`, `stable`, `pr-213`, etc.

As a migration strategy, we can make sure `master` and `stable` are built for both public and private repos for some time being. Then we can move our `ansible` repo over to using the private repository for both preview envs and also `master` and `stable`. Let's ping Cenk once this is merged but can you add `master` and `stable` as triggers for this one as well?

> Naming

Is it possible to change the naming like so? 
```
plausible/analytics:v1.4.0 // stays the same
plausible/analytics:master // -> plausible-private/analytics:master
plausible/analytics:stable // -> plausible-private/analytics:stable
plausible/analytics:pr-123 // -> plausible-private/analytics:pr-123
```

Or does that not make sense with the way Docker repos/tags work? The way I think about it, we're still getting the same image (i.e. `analytics`) but we're getting it from a different private repo (i.e. `plausible-private`). 

> Staging workflow image tags are labeled using PR reference 

I feel like using the branch name here would be more generic. But if the only way to trigger it is by creating a PR it doesn't matter I guess :)

ukutaht:
I played around with it myself. Registering a new organization in DockerHub is not ideal so let's do it as follows:

```javascript
plausible/analytics:v1.4.0 // stays the same
plausible/analytics:master // -> plausible/analytics-private:master
plausible/analytics:stable // -> plausible/analytics-private:stable
plausible/analytics:pr-123 // -> plausible/analytics-private:pr-123
```

For context, some people use the public `master` and `stable` images with no official support for self-hosting and get mad when it doesn't work. So I'd prefer to make these private as well.

I think then the private repo is more appropriately named `plausible-private` not just `plausible-staging` because it will also include production images. The difference is that they're private.

Happy to keep the two actions workflows separate.

vinibrsl:
> I played around with it myself. Registering a new organization in DockerHub is not ideal so let's do it as follows:
> 
> ```js
> plausible/analytics:v1.4.0 // stays the same
> plausible/analytics:master // -> plausible/analytics-private:master
> plausible/analytics:stable // -> plausible/analytics-private:stable
> plausible/analytics:pr-123 // -> plausible/analytics-private:pr-123
> ```
> 
> For context, some people use the public `master` and `stable` images with no official support for self-hosting and get mad when it doesn't work. So I'd prefer to make these private as well.
> 
> I think then the private repo is more appropriately named `plausible-private` not just `plausible-staging` because it will also include production images. The difference is that they're private. I've registered the `plausible/analytics-private` repo on Dockerhub.
> 
> Happy to keep the two actions workflows separate.

I like your suggestion! I updated the pull request to do the following:

- `build-public-images.yml` listens to pushed Git tag. Publishes images to `plausible/analytics` using semantic versioning
- `build-private-images.yml` subscribes to labeled pull request events + pushes to master/stable. Publishes images to `plausible/analytics-private` using PR reference when available or branch name.

That should do exactly what you suggested. Before merging we'll need to switch plausible.io deployment files to read from `plausible-private`.

ukutaht:
Awesome, that's ideal ✌️

I think having them in a separate repo is enough for now. I'd need to create a paid account to make them completely private but it's not like we have anything sensitive in these so not going to bother for now. Everything is built from public code on the public repo anyways. I just wanted to make sure the `stable` and `master` images are hidden away from the main repo that self-hosters pull from.

@vinibrsl How's your Ansible skills? Mind opening a PR on that repo to switch over the image to `plausible/analytics-private`?

vinibrsl:
> Awesome, that's ideal ✌️
> 
> I think having them in a separate repo is enough for now. I'd need to create a paid account to make them completely private but it's not like we have anything sensitive in these so not going to bother for now. Everything is built from public code on the public repo anyways. I just wanted to make sure the `stable` and `master` images are hidden away from the main repo that self-hosters pull from.
> 
> @vinibrsl How's your Ansible skills? Mind opening a PR on that repo to switch over the image to `plausible/analytics-private`?

Makes sense! There it is: https://github.com/plausible/ansible/pull/9 :)

PR #2156, opened 2022-08-31
Perform calculations in SQL instead of app when creating sites

vinibrsl:
### Changes

This PR removes some Enum calls to rely on the database for aggregating data. This improves performance when creating new sites, especially if the user has multiple sites.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -1,12 +1,12 @@
 defmodule Plausible.Sites do
   use Plausible.Repo
   alias Plausible.Site.SharedLink
+  import Ecto.Query
 
   def create(user, params) do
-    count = Enum.count(owned_by(user))
     limit = Plausible.Billing.sites_limit(user)
 
-    if count >= limit do
+    if owned_sites_count(user) >= limit do
       {:error, :limit, limit}
     else
       site_changeset = Plausible.Site.changeset(%Plausible.Site{}, params)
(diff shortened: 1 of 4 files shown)


lib/plausible/sites.ex
```diff
@@ -116,24 +116,25 @@ defmodule Plausible.Sites do
...
-        where: sm.user_id == ^user.id,
-        select: count(sm)
+  def owned_sites_domains(user) do
+    user
+    |> owned_sites_query()
+    |> select([site], site.domain)
+    |> Repo.all()
+  end
+
+  defp owned_sites_query(user) do
```
ukutaht, line 132: I like extracting the query here. Can the existing `owned_by` function also use it?

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -52,12 +52,10 @@ defmodule PlausibleWeb.SiteController do
   def new(conn, _params) do
     current_user = conn.assigns[:current_user] |> Repo.preload(site_memberships: :site)
 
-    owned_site_count =
-      current_user.site_memberships |> Enum.filter(fn m -> m.role == :owner end) |> Enum.count()
-
+    owned_site_count = Plausible.Sites.owned_sites_count(current_user)
```
ukutaht, line 55: I think this is only worth changing if you can also remove the `Repo.preload(site_memberships: :site)` from above.

lib/plausible/sites.ex
vinibrsl: I actually removed the `owned_by/1` function in favor of `owned_sites_domains/1` and `owned_sites_count/1`. The two new functions are using the extracted query :)

lib/plausible_web/controllers/site_controller.ex
vinibrsl: Oops! You're right.

PR #2150, opened 2022-08-30
Fix Prettier Config

PascalPixel:
### Changes

Please describe the changes made in the pull request here.

Prettier config is currently not working, the config I added seems to leads to the least code changes.

Below you'll find a checklist. For each item on the list, check one option and delete the other.

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #2176, opened 2022-09-05
Fix Prettier Config

PascalPixel:
Only the config fix from https://github.com/plausible/analytics/pull/2150

If we're going to fix files step by step it might be a good time to change the config now as well; perhaps just to the default prettier settings?

### DIFF
--- a/assets/.prettierrc.json
+++ b/assets/.prettierrc.json
@@ -1,5 +1,5 @@
 {
   "singleQuote": true,
-  "requirePragma": true,
-  "insertPragma": true
+  "trailingComma": "none",
+  "semi": false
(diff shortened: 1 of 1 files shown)


ukutaht:
Looks like we do have checks in the CI for this... sorry.

Over the next few weeks we'll we reducing the amount of open PRs. Once we get into a more steady state with few open PRs, I think it's a good idea to do a full `fix` run like you did earlier. But we need to coordinate this with our team so that everyone is on a clean slate and can pull immediately after this is done, to avoid big conflicts.

Sorry, I'll close this for now and plan to do it in the next month or so.

PR #2181, opened 2022-09-06
Fix improper pattern matching on Finch responses

aerosol:
### Changes

This PR fixes improper pattern matching on Finch responses
for 3rd party integration as seen in sentry.

By the way, abstracting away some request building details,
encapsulating redundant request encoding.

Type specs were added which immediately made some mismatches
visible via lsp integration.

Potential follow-ups to this:
  - a more comprehensive mocking setup for 3rd party
    integrations (preferably via https://github.com/dashbitco/mox)
    so that erroneous code paths can be tested
  - CI linting (such as credo to enforce typespecs/moduledocs in any
    files touched in a given PR) + perhaps a dialyzer run

### Tests
- [x] This PR requires extra tests but they can't be added yet (e.g. paddle API docs excluded the subscription preview docs). Should be done in a follow up PR. Otherwise this is only refactoring, no public interfaces are changed.

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/http.ex
+++ b/lib/plausible/google/http.ex
@@ -97,20 +97,30 @@ defmodule Plausible.Google.HTTP do
     url = "https://www.googleapis.com/webmasters/v3/sites"
     headers = [{"Content-Type", "application/json"}, {"Authorization", "Bearer #{access_token}"}]
 
-    {:ok, response} = HTTPClient.get(url, headers)
+    case HTTPClient.get(url, headers) do
+      {:ok, response} ->
+        response
+        |> Map.get(:body)
+        |> Jason.decode!()
+        |> then(&{:ok, &1})
 
-    response
-    |> Map.get(:body)
-    |> Jason.decode!()
-    |> then(&{:ok, &1})
+      {:error, reason} = e ->
+        Logger.error("Google Analytics: failed to list sites: #{reason}")
+        e
+    end
   end
 
   def fetch_access_token(code) do
     url = "https://www.googleapis.com/oauth2/v4/token"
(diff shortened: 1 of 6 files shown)


lib/plausible/google/http.ex
```diff
@@ -201,31 +219,30 @@ defmodule Plausible.Google.HTTP do
         |> Map.get("error")
         |> then(&{:error, &1})
 
-      {:error, %Finch.Response{body: body}} ->
-        Sentry.capture_message("Error fetching Google queries", extra: Jason.decode!(body))
+      {:error, %{reason: reason}} ->
```
aerosol, line 222: No such thing as `{:error, %Finch.Response{...`

lib/plausible/google/http.ex
```diff
@@ -245,9 +262,13 @@ defmodule Plausible.Google.HTTP do
 
         {:ok, date}
 
-      {:error, %Finch.Response{body: body}} ->
+      {:ok, %Finch.Response{body: body}} ->
```
aerosol, line 265: that'd never match

lib/plausible/google/http.ex
```diff
@@ -245,9 +262,13 @@ defmodule Plausible.Google.HTTP do
...
         Sentry.capture_message("Error fetching Google view ID", extra: Jason.decode!(body))
         {:error, body}
+
+      {:error, %{reason: reason}} ->
```
aerosol, line 269: `Mint` exposes two error types both of which contain the `reason` field.

RobertJoonas (approved):

vinibrsl (approved):
Looks good! We recently switched HTTP clients and some matches were odd. This reads so much better now ✨

PR #2184, opened 2022-09-07
List all Google Analytics views during import

vinibrsl:
### Changes

This commit fixes a bug where different Google Analytics views with the same name and URI were not shown. This was caused because GA views were stored as a map, that naturally doesn't support duplicate keys.

This change updates the GA views list to display view IDs, making it clearer to know what is being imported. The dropdown is now grouped by website URL.

New:
Screen Shot 2022-09-06 at 22 55 23

Current:
image

### Tests

### Changelog
- [X] Entry has been added to changelog (Google Analytics entry is still Unreleased)

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/api.ex
+++ b/lib/plausible/google/api.ex
@@ -3,6 +3,8 @@ defmodule Plausible.Google.Api do
   use Timex
   require Logger
 
+  @type google_analytics_view() :: {view_name :: String.t(), view_id :: String.t()}
+
   @scope URI.encode_www_form(
            "https://www.googleapis.com/auth/webmasters.readonly email https://www.googleapis.com/auth/analytics.readonly"
          )
@@ -53,27 +55,52 @@ defmodule Plausible.Google.Api do
     end
   end
 
-  def get_analytics_view_ids(access_token) do
+  @spec list_views(access_token :: String.t()) ::
+          {:ok, %{(hostname :: String.t()) => [google_analytics_view()]}} | {:error, term()}
+  @doc """
+  Lists Google Analytics views grouped by hostname.
+  """
+  def list_views(access_token) do
     case HTTP.list_views_for_user(access_token) do
       {:ok, %{"items" => views}} ->
-        view_ids = for view <- views, do: build_view_ids(view), into: %{}
-        {:ok, view_ids}
+        views = Enum.group_by(views, &view_hostname/1, &view_names/1)
+        {:ok, views}
 
       error ->
         error
     end
   end
 
-  defp build_view_ids(view) do
-    uri = URI.parse(Map.get(view, "websiteUrl", ""))
-
-    if !uri.host do
-      Sentry.capture_message("No URI for view ID", extra: view)
+  defp view_hostname(view) do
+    case view do
+      %{"websiteUrl" => url} when is_binary(url) -> url |> URI.parse() |> Map.get(:host)
+      _any -> "Others"
     end
+  end
 
-    host = uri.host || Map.get(view, "id", "")
-    name = Map.get(view, "name")
-    {"#{host} - #{name}", Map.get(view, "id")}
+  defp view_names(%{"name" => name, "id" => id}) do
+    {"#{id} - #{name}", id}
+  end
+
+  @spec get_view(access_token :: String.t(), lookup_id :: String.t()) ::
+          {:ok, google_analytics_view()} | {:ok, nil} | {:error, term()}
+  @doc """
+  Returns a single Google Analytics view if the user has access to it.
+  """
+  def get_view(access_token, lookup_id) do
+    case list_views(access_token) do
+      {:ok, views} ->
+        view =
+          views
+          |> Map.values()
+          |> List.flatten()
+          |> Enum.find(fn {_name, id} -> id == lookup_id end)
+
+        {:ok, view}
+
+      {:error, cause} ->
+        {:error, cause}
+    end
   end
 
   @per_page 10_000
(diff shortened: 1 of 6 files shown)


lib/plausible/google/api.ex
```diff
@@ -56,24 +56,23 @@ defmodule Plausible.Google.Api do
...
+  defp view_hostname(view) do
+    case view do
+      %{"websiteUrl" => url} when is_binary(url) -> url |> URI.parse() |> Map.get(:host)
+      _any -> "Others"
     end
+  end
 
-    host = uri.host || Map.get(view, "id", "")
-    name = Map.get(view, "name")
-    {"#{host} - #{name}", Map.get(view, "id")}
```
vinibrsl, line 76: This is the culprit. The combination of host and name is not unique, as you can have multiple views with that.

lib/plausible/google/api.ex
```diff
@@ -56,24 +56,23 @@ defmodule Plausible.Google.Api do
   def get_analytics_view_ids(access_token) do
     case HTTP.list_views_for_user(access_token) do
       {:ok, %{"items" => views}} ->
-        view_ids = for view <- views, do: build_view_ids(view), into: %{}
-        {:ok, view_ids}
+        views = Enum.group_by(views, &view_hostname/1, &view_names/1)
```
aerosol, line 59: Do you think this could be made client's interface concern? i.e. instead of having `HTTP.list_views_for_user/1` return a raw response, we could parse and test the details there, even if that extra behavior is made optional?

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -714,8 +714,10 @@ defmodule PlausibleWeb.SiteController do
     end_date =
       Plausible.Stats.Clickhouse.pageview_start_date_local(site) || Timex.today(site.timezone)
 
-    {:ok, view_ids} = Plausible.Google.Api.get_analytics_view_ids(access_token)
-    {view_id_name, _} = Enum.find(view_ids, fn {_, v} -> v == view_id end)
+    {:ok, views} = Plausible.Google.Api.get_analytics_view_ids(access_token)
```
aerosol, line 717: Again, perhaps we could cover this with  something like `Google.Api.get_view_by_id(view_id)`?

Is there a controller test necessary here?

ukutaht: +1 I like that suggestion

ukutaht (commented):
Good stuff, thanks @vinibrsl 

@aerosol had some comments about potential refactorings. The separation between `Google.Api` and `Google.HTTP` is not super clear to me, maybe @vinibrsl you have stronger opinions between where to put stuff.

But I think moving logic from controllers to service modules like `Google.Api` is almost always a good idea. Definitely in favour of that

ukutaht:
The dropdown itself looks so much nicer on the screenshots 💯

lib/plausible_web/controllers/site_controller.ex
vinibrsl: Thanks for the suggestion, @aerosol! I extracted this in a function inside `Plausible.Google.Api`.

lib/plausible/google/api.ex
vinibrsl: To put this into context, we recently refactored the entire Google Analytics import feature (#2046). This is a big feature and there is still a lot to improve codewise :)

I created two separate modules, `Google.HTTP` and `Google.Api`. `Google.HTTP` should make requests and return raw responses, while `Google.Api` translates those raw responses to Plausible's domain.

vinibrsl:
Thanks for the reviews! I updated the PR with your suggestions, added tests (with Bypass), and extracted some functions.

test/plausible_web/controllers/site_controller_test.exs
```diff
@@ -722,6 +722,37 @@ defmodule PlausibleWeb.SiteControllerTest do
...
+        fn conn ->
+          response_body = File.read!("fixture/ga_list_views.json")
+          Plug.Conn.resp(conn, 200, response_body)
+        end
+      )
+
+      :plausible
+      |> Application.get_env(:google)
+      |> Keyword.put(:api_url, "http://0.0.0.0:#{bypass.port}")
+      |> then(&Application.put_env(:plausible, :google, &1))
```
aerosol, line 744: Unfortunately this won't scale well -- tests running in parallel will keep racing for that shared state resulting in random failures. Moreover, it feels like Bypass isn't ideal here - it spawns TCP listeners that we shouldn't really require for 3rd party integration tests. Perhaps we should park this till [Mox](https://github.com/dashbitco/mox) is incorporated - it has enough flexibility to let us improve the contract test scenarios against various responses.

aerosol: > Can you elaborate on this? Where is Bypass appropriate and where isn't it?

I _think_ it's appropriate where the actual TCP/HTTP needs to be involved in testing; the only valid case for it IMHO is testing the HTTP Client itself. So a real HTTP server becomes the system under test, with all its failure modes and semantics, e.g. socket timeouts.

For mocking responses from a 3rd party service and keeping the integration contract somewhat healthy, a combination of pre-recorded HTTP responses (with e.g. exvcr that we seem to already use) and _concurrent_ mocks that a) implement an interface that is shared with the live client, b) are not concerned with the underlying network behavior, should strike a good balance between regression tracking and testing agility.

ukutaht:
Agreed. I'll deploy this tomorrow :)

PR #2197, opened 2022-09-09
Manually lock outgrown enterprise accounts

vinibrsl:
### Changes

This pull request adapts the account locking and grace period flow to enterprise customers. A quick summary of the changes:

- Enables the grace period mechanism for enterprise accounts on `Workers.CheckUsage`, therefore showing existing dashboard alerts
- Creates a `manual_lock` field in the `grace_period` JSON, that when `true`, does not expire the grace period automatically, therefore preventing `Workers.LockSites` to lock enterprise customers
- Adds a lock/unlock button to the internal CRM users page
- Tests new existing and new functionality

I [recorded this screencast](https://zipmessage.com/rvgm10bf/032d02b1242cda0fa7c0300aa) previewing the feature. There's been some refactoring, so small commits can help reviewing.

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl:
I still want to test the email templates I changed manually. That is why I opened this PR as a draft.

lib/plausible/auth/user_admin.ex
```diff
@@ -31,28 +31,47 @@ defmodule Plausible.Auth.UserAdmin do
...
+        "--"
 
-  defp grace_period_status(user) do
-    if user.grace_period.is_over do
-      "ended"
-    else
-      days_left = Timex.diff(user.grace_period.end_date, Timex.now(), :days)
-      "#{days_left} days left"
+      %{manual_lock: true} ->
+        "Manual lock"
```
ukutaht, line 67: If I understand the logic correctly, the grace period is started automatically by the `check_usage` background job. I'm slightly confused by what _'manual lock'_ here refers to.

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -14,10 +14,11 @@ defmodule PlausibleWeb.AuthPlug do
       id ->
         user =
           Repo.get_by(Plausible.Auth.User, id: id)
-          |> Repo.preload(
+          |> Repo.preload([
+            :enterprise_plan,
```
ukutaht, line 18: Can we avoid this database query on the hot path?

This plug is executed for every logged-in request so it should keep data fetching to a minimum. It should only fetch data that is absolutely necessary for every request after this plug.

lib/plausible_web/templates/email/over_limit.html.eex
```diff
@@ -7,14 +7,14 @@ This is a notice that your traffic has exceeded your subscription tier two month
 In order to keep your stats running, we require you to upgrade your account to accommodate your new traffic levels. If you do not upgrade your account within the next 7 days, we will lock your stats and they won't be accessible.
 <br /><br />
 In the last billing cycle (<%= date_format(@last_cycle.first) %> to <%= date_format(@last_cycle.last) %>), your account has used <%= PlausibleWeb.StatsView.large_number_format(@usage) %> billable pageviews.
-<%= if @usage <= 20_000_000 do %>
+<%= if @suggested_plan == :enterprise do %>
```
ukutaht, line 10: Nice, much better :)

lib/plausible_web/views/billing_view.ex
```diff
@@ -37,4 +37,9 @@ defmodule PlausibleWeb.BillingView do
         "custom"
     end
   end
+
+  def enterprise?(%Plausible.Auth.User{enterprise_plan: %Plausible.Billing.EnterprisePlan{}}),
```
ukutaht, line 41: In order to avoid fetching the enterprise_plan in the plug, you can add:
```elixir
user = Repo.preload(user, :enterprise_plan)
```

here. It will only make the extra DB roundtrip if the `enterprise_plan` is not already loaded.

mix.exs
```diff
@@ -123,4 +127,19 @@ defmodule Plausible.MixProject do
       sentry_recompile: ["compile", "deps.compile sentry --force"]
     ]
   end
+
+  defp docs do
+    [
+      main: "readme",
+      logo: "assets/static/images/icon/plausible_favicon.png",
+      extras:
+        Path.wildcard("guides/**/*.md") ++
```
ukutaht, line 136: Would be good to add `CONTRIBUTING.md` to the docs as well

ukutaht (commented):
Looks good overall 👍 Some comments inline

lib/plausible/auth/user_admin.ex
vinibrsl: Grace period is started by `CheckUsage` indeed. This new `manual_lock` boolean is `true` for enterprise customers, and prevents `SiteLocker` (the other background job) to lock enterprise sites.

mix.exs
vinibrsl: Good call!

lib/plausible_web/plugs/auth_plug.ex
vinibrsl: Oops! You're right, we don't need that there :)

lib/plausible_web/views/billing_view.ex
vinibrsl: Good call! Done :)

lib/plausible/auth/user_admin.ex
ukutaht: Yeah I'm thinking more from the users' perspective. If Marko or Amanda see 'Manual lock' in the CRM, it's easy to assume that the account is already locked.

Basically, there should be a clear indicator for CS whether the users' sites are locked or not. Currently, the label does not seem to change before or after the CS agent clicks 'lock' in the CRM. How about something like:
1. First the label is `Grace period -> Indefinite` 
2. CS clicks `Lock` to manually lock the site
3. The label changes to `Grace period -> Manual lock`

I'm not sure if `Indefinite` is the right word but it's basically trying to say that the account is on a grace period, but it will not automatically end, rather it needs to be ended manually by CS.

vinibrsl: Ohhh I see! I agree that the current label is confusing. What do you think of this? (b6215b9e2ccc7147c0924163283750838fc143af)

1. Label: `Waiting for manual lock`
2. Lock action
3. Label: `Manually locked
4. Unlock action
5. Label: `--`

Preview

ukutaht: Looks good, thanks!

ukutaht (approved):

PR #2208, opened 2022-09-12
Rework outbound links and file downloads

RobertJoonas:
### Changes

- Refactoring
  - Handlebars helper function to inject code for multiple script extensions at the same time
  - Extracted several functions and improved readability
- Made the script not prevent default if it has been prevented by an external script (fixes https://github.com/plausible/analytics/issues/1941)
- Used callbacks instead of waiting 150ms for the event request to be sent
- Disallowed sending two events at the same time by only having one click handler
  - If using `script.file-downloads.outbound-links.js` and clicking an outbound download link, only outbound link click event is sent  

### Tests

### Changelog

### Documentation

### Dark mode


tracker/src/plausible.js
```diff
@@ -177,16 +177,14 @@
   }
 
   function handleOutbound(event) {
+    if (event.type === 'auxclick' && event.which !== 2) { return }
```
RobertJoonas, line 177: That's the only case where the click-handler function is called but we don't want to send the event

tracker/src/plausible.js
```diff
@@ -169,19 +169,14 @@
...
-
-  function handleOutbound(event) {
+  function handleLinkClickEvent(event) {
     if (event.type === 'auxclick' && event.which !== 2) { return }
 
     var link = getLinkEl(event.target)
+    var hrefWithoutQuery = link && link.href && link.href.split('?')[0]
 
-    if (isOutboundLink(link)) {
+    if (typeof isOutboundLink === 'function' && isOutboundLink(link)) {
```
RobertJoonas, line 182: `typeof isOutboundLink === 'function'` will return `false` if the function does not exist. This function will only exist in outbound-links extension

tracker/compile.js
```diff
@@ -17,7 +17,11 @@ function compilefile(input, output, templateVars = {}) {
   const template = Handlebars.compile(code)
   const rendered = template(templateVars)
   const result = uglify.minify(rendered)
-  fs.writeFileSync(output, result.code)
+  if (result.code) {
+    fs.writeFileSync(output, result.code)
+  } else {
+    throw new Error(`Failed to compile ${output.split('/').pop()}.\n${result.error}\n`)
+  }
```
RobertJoonas, line 24: Currently, when script compilation fails on `npm run deploy`, we do not show the actual error message on why it failed

vinibrsl: Good catch! This will give more visibility when generating the files.

IMO that string interpolation is quite complicated to read. Maybe it's a good idea to separate in variables?

tracker/src/plausible.js
vinibrsl: I was curious about `event.which` in the if clause, and I found that there is a deprecation warning on [`UIEvent.which`](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which). Is that the same `event.which` we are using?

In any case, I think it's a good idea to extract the `event.which` value in a constant. For example:

```
if (event.type === 'auxclick' && event.which !== mySpecialMouseClick) { return }
```

tracker/src/plausible.js
ukutaht: Could this be decided at compile-time instead?

What I mean is wrapping it in handlebars like so:
```javascript
{{#if outbound_links}}
  if (isOutboundLink(link)) {
    // send event
  }
{{/if}}
```

tracker/compile.js
ukutaht: Nice addition!

ukutaht (commented):
Reviewing this commit-by-commit was a pleasure, good job @RobertJoonas!

The only change I would request is removing the `typeof` check and doing it with handlebars instead.

vinibrsl (approved):
Nice work!

tracker/src/plausible.js
```diff
@@ -174,21 +174,30 @@
...
   function handleLinkClickEvent(event) {
-    if (event.type === 'auxclick' && event.which !== 2) { return }
+    var middleMouseButton = 1
+    if (event.type === 'auxclick' && event.button !== middleMouseButton) { return }
 
     var link = getLinkEl(event.target)
     var hrefWithoutQuery = link && link.href && link.href.split('?')[0]
 
-    if (typeof isOutboundLink === 'function' && isOutboundLink(link)) {
+    // if clause for syntax - allows dynamically injecting 0 or more `else if` clauses
```
ukutaht, line 183: Maybe a simpler way to factor this:

```
 {{#if outbound_links}}
     if (isOutboundLink(link)) {
       var eventName = 'Outbound Link: Click'
       var eventProps = {url: link.href}
       return sendLinkClickEvent(event, link, eventName, eventProps)
     }
     {{/if}}

   {{#if file_downloads}}
     if (isDownloadToTrack(hrefWithoutQuery)) {
       var eventName = 'File Download'
       var eventProps = {url: hrefWithoutQuery}
       return sendLinkClickEvent(event, link, eventName, eventProps)
     }
     {{/if}}
```

Adding the `return` allows you to exit the function early so you don't need the chain of `if-else` necessarily, it can be composed of plain `if` statements.

tracker/src/plausible.js
```diff
@@ -174,21 +174,30 @@
   }
 
   function handleLinkClickEvent(event) {
-    if (event.type === 'auxclick' && event.which !== 2) { return }
+    var middleMouseButton = 1
```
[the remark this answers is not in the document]
RobertJoonas: As Vini mentioned, the [`event.which`](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which) has a deprecation warning. So I've now changed it to use [event.button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) instead of `event.which` as suggested in the docs.

> Did this change from 2 to 1?

Yes, the middle button has a value of 1 in the `event.button` property.

Noted about the general convention. Changed it now.

tracker/src/plausible.js
RobertJoonas: Missed answering some of your comments :)

Changed `event.which` to `event.button` and extracted the constant in https://github.com/plausible/analytics/pull/2208/commits/e20777f96cf2cef9eb03ffe16ab99f4dab8422da

tracker/src/plausible.js
RobertJoonas: Changed in https://github.com/plausible/analytics/pull/2208/commits/1dff10a5b568466c00ee40e31b201b121e8f17a7

PR #2212, opened 2022-09-12
Build docker images with new tags

aerosol:
### Changes
 
This adds a new GitHub Action that runs on new tag creation. The workflow should result with a tagged docker image.

### Tests

### Changelog

### Documentation

### Dark mode


.github/workflows/build-tagged-images.yml
```diff
@@ -0,0 +1,43 @@
+name: Build Tagged Images
+
+on:
+  push:
+      tags:
+          - '**'
```
vinibrsl, line 6: I'm not sure, but I think adding this to `.github/workflows/build-private-images.yml` and putting `type=ref,event=tag` up top would do it all in a single file.

Happy to go with this file, but just checking if you considered that.

vinibrsl (approved):

.github/workflows/build-tagged-images.yml
ukutaht: Yes, I think extending `.github/workflows/build-private-images.yml` is a good idea unless there's some reason to keep this separate.

aerosol: > unless there's some reason to keep this separate.

Not at the moment, no. On it.

aerosol: OK I went with `r*` tagging scheme for now as discussed with @ukutaht elsewhere.

https://github.com/plausible/analytics/pull/2212/commits/0d2ae743d3016b2cf2a40be9a094858fdb207aa9 - if you think that'll work, I'll be happy to give it a go

ukutaht:
I was going to do it myself but maybe better to flag: unless there's a reason to keep the branch around, let's delete it after the PR is merged to keep the nr of branches in the whole git tree reasonable.

aerosol:
I just did, was mine after all! Apologies

PR #2216, opened 2022-09-13
Allow admins to grant admin rights

ukutaht:
### Changes

Sentry error: https://sentry.plausible.io/organizations/sentry/issues/1945/events/2f6ea4f040284deebea16469384e0f60/?project=1

This error is caused by people taking a reasonable action: someone with `admin` rights wants to grant someone else `admin` rights as well. I think this action should be supported but it currently isn't.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/site/membership_controller.ex
+++ b/lib/plausible_web/controllers/site/membership_controller.ex
@@ -119,9 +119,14 @@ defmodule PlausibleWeb.Site.MembershipController do
     Owner - Can update anyone's role except for themselves. If they want to change their own role, they have to use the 'transfer ownership' feature.
     Admin - Can update anyone's role except for owners. Can downgrade their own access to 'viewer'. Can promote a viewer to admin.
   """
+  @role_mappings Membership
+                 |> Ecto.Enum.mappings(:role)
+                 |> Enum.map(fn {k, v} -> {v, k} end)
+                 |> Enum.into(%{})
+
   def update_role(conn, %{"id" => id, "new_role" => new_role_str}) do
     %{site: site, current_user_role: current_user_role} = conn.assigns
-    new_role = intern_role(new_role_str)
+    new_role = Map.fetch!(@role_mappings, new_role_str)
 
     if can_grant_role?(current_user_role, new_role) do
       membership =
@@ -149,10 +154,6 @@ defmodule PlausibleWeb.Site.MembershipController do
       |> redirect(to: Routes.site_path(conn, :settings_people, site.domain))
     end
   end
-
-  defp intern_role("owner"), do: :owner
-  defp intern_role("admin"), do: :admin
-  defp intern_role("viewer"), do: :viewer
 
   defp can_grant_role?(:owner, :admin), do: true
   defp can_grant_role?(:owner, :viewer), do: true
--- a/lib/plausible_web/controllers/site/membership_controller.ex
+++ b/lib/plausible_web/controllers/site/membership_controller.ex
@@ -13,6 +13,17 @@ defmodule PlausibleWeb.Site.MembershipController do
   plug PlausibleWeb.AuthorizeSiteAccess,
        [:owner, :admin] when action not in @only_owner_is_allowed_to
 
+  @moduledoc """
+    This controller deals with user management via the UI in Site Settings -> People. It's important to enforce permissions in this controller.
+
+    Owner - Can manage users, can trigger a 'transfer ownership' request
+    Admin - Can manage users
+    Viewer - Can not access user management settings
+    Anyone - Can accept invitations
+
+    Everything else should be explicitly disallowed.
+  """
+
   def invite_member_form(conn, %{"website" => site_domain}) do
     site = Sites.get_for_user!(conn.assigns[:current_user].id, site_domain)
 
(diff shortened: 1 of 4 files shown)


lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -13,6 +13,17 @@ defmodule PlausibleWeb.Site.MembershipController do
   plug PlausibleWeb.AuthorizeSiteAccess,
        [:owner, :admin] when action not in @only_owner_is_allowed_to
 
+  @moduledoc """
```
aerosol, line 16: Amazing, thank you for including `@moduledoc` :star_struck: 
Just by convention I think it should always start at the second line, right after module definition.

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -101,28 +112,54 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+      )
+      |> redirect(to: redirect_target)
+    else
+      conn
+      |> put_flash(:error, "You are not allowed to grant the #{new_role} role")
+      |> redirect(to: Routes.site_path(conn, :settings_people, site.domain))
+    end
   end
 
+  defp intern_role("owner"), do: :owner
```
aerosol, line 153: I _think_ you meant "internal" here (?), but perhaps we could use a map like 

```elixir
@role_mappings Membership
               |> Ecto.Enum.mappings(:role) 
               |> Enum.map(fn {k,v} -> {v,k} end)  
               |> Enum.into(%{})
```

so we can

```elixir
new_role = Map.fetch!(@role_mappings, new_role)
```

This way, we won't have to get back here anymore once a new role is introduced for instance.

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -101,28 +112,54 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+  defp intern_role("admin"), do: :admin
+  defp intern_role("viewer"), do: :viewer
+
+  defp can_grant_role?(:owner, :admin), do: true
+  defp can_grant_role?(:owner, :viewer), do: true
```
aerosol, line 158: Wondering if we could move that logic to a service-ish module, even if very modest.
I'm not sure how likely it's for us to call any role assignments outside of that controller in the future. Feel free to disregard the comment.

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -101,28 +112,54 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+  defp can_grant_role?(:admin, :admin), do: true
+  defp can_grant_role?(:admin, :viewer), do: true
```
aerosol, line 160: Sanity check - this means one can also downgrade themselves from admin/owner to viewer, is that intentional? Don't we risk a lock-out here?

aerosol (commented):
Comments inline

lib/plausible_web/controllers/site/membership_controller.ex
ukutaht: I actually did mean `intern`. I learned that word when I tried to write a programming language: https://en.wikipedia.org/wiki/String_interning

Seems like that word isn't widely known so I'm happy to get rid of it :) Let's just call them Atoms like Elixir does.

I like your suggestion, I'll go with the `@role_mappings` solution

lib/plausible_web/controllers/site/membership_controller.ex
ukutaht: Done

lib/plausible_web/controllers/site/membership_controller.ex
aerosol: > https://en.wikipedia.org/wiki/String_interning

:exploding_head: learn something new every day

lib/plausible_web/controllers/site/membership_controller.ex
ukutaht: Currently they are prevented from doing it because the option isn't shown in the UI.

I've added another commit that ensures that an owner lock-out cannot happen. More strictness doesn't hurt

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -101,28 +112,66 @@ defmodule PlausibleWeb.Site.MembershipController do
...
-    )
-    |> redirect(to: redirect_target)
+    if can_grant_role do
+      membership =
+        membership
+        |> Membership.changeset(%{role: new_role})
+        |> Repo.update!()
+
+      redirect_target =
+        if membership.user.id == current_user.id && new_role == :viewer do
```
aerosol, line 147:
```suggestion
        if membership.user.id == current_user.id and new_role == :viewer do
```

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -101,28 +112,66 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+                 |> Enum.map(fn {k, v} -> {v, k} end)
+                 |> Enum.into(%{})
+
+  def update_role(conn, %{"id" => id, "new_role" => new_role_str}) do
+    %{site: site, current_user: current_user, current_user_role: current_user_role} = conn.assigns
+
+    membership = Repo.get!(Membership, id) |> Repo.preload(:user)
+    new_role = Map.fetch!(@role_mappings, new_role_str)
+
+    can_grant_role =
```
aerosol, line 133:
```suggestion
    can_grant_role? =
```

aerosol (approved):

PR #2219, opened 2022-09-13
Remove parallel requests to Google Analytics

vinibrsl:
### Changes

This makes Google Analytics HTTP requests synchronous. Parallel requests were added in the recent refactor (#2046) as a strategy to make importing faster, but turns out the change that made it faster was fetching GA records paginated instead of monthly batches, not parallelism. We decided to make it synchronous because:

- Makes request mocking easier
- Removes unnecessary complexity
- Mitigates HTTP connectivity issues (e.g. timeouts)

Me and @aerosol tested this against a 1M pageviews GA import, and importing was finished in 45 seconds without parallel requests, which is good. Parallel requests are in fact faster in most cases, but given the reasons above, not worth the complexity in this instance.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/api.ex
+++ b/lib/plausible/google/api.ex
@@ -104,7 +104,6 @@ defmodule Plausible.Google.Api do
   end
 
   @per_page 10_000
-  @one_day_in_ms 86_400_000
   @doc """
   API reference:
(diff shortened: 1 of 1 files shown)


aerosol (approved):
Looks good. :nail_care: points for single pipe removal maybe, but frankly we should have credo rule for stuff like that.

ukutaht (approved):

PR #2224, opened 2022-09-14
add finch preset

ruslandoga:
👋

### Changes

This makes finch less memory hungry and allows using smaller instances with less variance in CPU usage.

Image with this change: [ghcr.io/ruslandoga/analytics:pr-3](https://github.com/ruslandoga/analytics/pkgs/container/analytics/41633489?tag=pr-3)

### Tests
- [x] Tests (or rather env files) have been updated

### Changelog

### Documentation

### Dark mode


ruslandoga:
Screenshot 2022-09-14 at 11 26 07

From Sept 12 and up to yesterday I was running with finch configured with large pool sizes. It resulted in instance becoming unavailable a few times.

ruslandoga:
Should anything be done regarding changelog / docs?

config/runtime.exs
```diff
@@ -373,6 +373,7 @@ config :plausible, :hcaptcha,
   secret: hcaptcha_secret
 
 config :plausible, Plausible.Finch,
+  preset: get_var_from_path_or_env(config_dir, "FINCH_PRESET", "SELF_HOSTED"),
```
vinibrsl, line 376: I think the default value should map to the current behavior, which is the complete pool.

config/.env.dev
```diff
@@ -11,6 +11,7 @@ ADMIN_USER_IDS=1
 SHOW_CITIES=true
 PADDLE_VENDOR_AUTH_CODE=895e20d4efaec0575bb857f44b183217b332d9592e76e69b8a
 PADDLE_VENDOR_ID=3942
+FINCH_PRESET=CLOUD
```
vinibrsl, line 14: What do you think of naming this FINCH_POOL_CONFIG or FINCH_POOL_CONFIG_PRESET?

lib/plausible/application.ex
```diff
@@ -43,7 +43,45 @@ defmodule Plausible.Application do
...
+    sentry_dsn = Application.get_env(:sentry, :dsn)
+
+    if is_binary(sentry_dsn) do
+      Map.put(pool_config, sentry_dsn,
+        size: config[:sentry_pool_size],
+        count: config[:sentry_pool_count]
+      )
+    else
+      pool_config
+    end
```
vinibrsl, line 80: What do you think of putting this in a separate function, as we're changing this?

lib/plausible/application.ex
```diff
@@ -43,7 +43,45 @@ defmodule Plausible.Application do
...
+    pool_config =
+      case config[:preset] do
+        "CLOUD" ->
+          finch_cloud_preset(config)
+
+        "SELF_HOSTED" ->
+          finch_self_hosted_preset()
+
+        other ->
+          raise """
```
vinibrsl, line 55: Good stuff! This error message is really descriptive ✨

lib/plausible/application.ex
```diff
@@ -43,7 +43,45 @@ defmodule Plausible.Application do
   defp finch_pool_config() do
     config = Application.fetch_env!(:plausible, Plausible.Finch)
 
-    pool_config = %{
+    pool_config =
+      case config[:preset] do
```
vinibrsl, line 47: I have a naming suggestion that might denote better the meaning of this new setting:

- `FINCH_PRESET` -> `FINCH_POOL_CONFIG_PRESET` or `FINCH_POOL_CONFIG`
- `CLOUD` -> `COMPLETE` or `FULL` (default to maintain compatibility)
- `SELF_HOSTED` -> `MINIMAL`

Let me know what you think!

vinibrsl (commented):
Thanks for your pull request! I left some suggestions and considerations.

lib/plausible/application.ex
ruslandoga: I'll leave the naming up to you :)

vinibrsl:
> Should anything be done regarding changelog / docs?

Yes, please! You can add your change to the "Unreleased" section under "Added".

config/runtime.exs
ukutaht: Yeah @vinibrsl it's the same with a bunch of other config options as well. The defaults are appropriate for self-hosting so that self-hosters don't have to manage a large complex configuration file. We override these options in prod for Plausible.Cloud to keep it simple for self-hosters.

So when people upgrade, they just get this config change automatically. If we didn't make it the default, self-hosters would have to update their configs when they upgrade.

vinibrsl: Got it! I'll make sure Plausible Cloud has that environment variable set then :)

ruslandoga:
> Now, the question is whether these high `:count` values are beneficial in any way for Plausible.Cloud? Maybe it's fine to run with a single pool for each.

For HTTP2 connections `:count` is practically the number of TCP connections to the domain. You can try `iperf3` or similar tool to measure how much of a benefit you get from additional TCP connections.

aerosol:
> This PR creates two different pooling presets, one for beefier machines and the other one that saves memory.

I dislike that this is done as a custom conditional layer not a specific `Mix.env` config. We should have `config/selfhosted.exs` for any overrides like this going forward.

> Now, the question is whether these high :count values are beneficial in any way for Plausible.Cloud? Maybe it's fine to run with a single pool for each.

Finch allows us to configure two values per each domain:
  - `size` that describes the number of connections
  - `count` that describes the number of pools

If you are not aware of any reasons to increase the number of pools I'd leave them at 1. Should be fine, was fine for much more busy systems I've seen. We should only tweak size if necessary, unless proven otherwise.

ruslandoga:
`:size` is no-op for http connections. `config/selfhosted.exs` wouldn't work at runtime?

aerosol:
@ruslandoga ah you're right about http2. It's multiplexing though, isn't it enough to be a sane default for all? Have we measured anything in that regard? I don't think I've ever used http2 explicitly with Finch so I can't advise here.

> config/selfhosted.exs wouldn't work at runtime?

Yeah I only mean we can have MIX_ENV=selfhosted provided when building the app, right? Doesn't have to be that specifically, just maybe using very verbose and redundant env vars such as FINCH_*_POOL_SIZE, FINCH_*_POOL_COUNT etc. could get us better coverage in satisfying different setups. Two presets don't necessarily fit all, that was my point.

ruslandoga:
@aerosol I see. I'll try replacing presets with more granular pool configs + selfhosted compile time mix env to pick domains.

ukutaht:
Based on this conversation, I think we should just try removing the excessive `count` configuration and see if it causes any problems for us in prod. According to reports `count: 1` works just fine for much bigger systems . So maybe maintaining this config split is not needed in the first place

aerosol:
@ukutaht I can only vouch for http1 setup, never tried http2

ruslandoga:
@ukutaht @aerosol I've been using finch with http2 for some time and it mostly works fine. In my head, I equate http2's max_concurrent_streams on a single connection to the number of http1 connections. Here're the `max_concurrent_streams` values for the domains:

- analyticsreporting.googleapis.com --  100
- icons.duckduckgo.com -- 128
- www.googleapis.com -- 100

ukutaht:
> If at most 100 reqs at a time is fine, then a single pool is probably fine too. Once 101th comes in, mint rejects it without sending, so it's cheap to retry.

OK, so we can lower the `:count` to 1 and then monitor for `:too_many_concurrent_requests`. If we run into that, we have to increase the count. Does that sound right?

ruslandoga:
@ukutaht I would react to `:too_many_concurrent_requests` after a few attempts at retrying to mellow out the spikes.

```elixir
case retry(fn -> finch_request() end, attempts: 5) do
  {:ok, ...} -> ...
  {:error, %Mint.HTTPError{module: Mint.HTTP2, reason: :too_many_concurrent_requests}} ->
    # sentry, logger, etc.
end
```

ukutaht:
Yes I think HTTP timeouts should also be retried a few times before giving up, depending on the use-case.

@ruslandoga I'll try out out `count=1` in production tomorrow. If it works fine then the excessive pooling is not needed and we can keep it at `count=1` for everyone. It's a very simple change and I'll close this PR for now.

PR #2228, opened 2022-09-15
Add API to list sites

francois2metz:
### Changes

This implement the following user request: https://github.com/plausible/analytics/discussions/1280

The API is paginated, there is no real paginated format right now in the API, so I created something. I expect some feedbacks to improve it.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -93,6 +93,19 @@ defmodule Plausible.Sites do
     )
   end
 
+  def get_sites_for_user(user_id, roles \\ [:owner, :admin, :viewer]),
+    do: Repo.all(get_sites_for_user_q(user_id, roles))
+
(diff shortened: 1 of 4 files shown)


aerosol:
Thank you for you contribution @francois2metz! :heart: We are discussing this internally and will come back to you once we agree on what's the best way to move it forward. Thanks for your patience!

aerosol:
@francois2metz Having discussed this internally, we think the best solution for the product in the long run would be to implement cursor based pagination, rather than one utilizing offsets (we're thinking of either https://github.com/duffelhq/paginator or https://github.com/sketch-hq/paginator -- to be evaluated). If you would like to contribute to building this feature still, we are happy to go into details. If not, we can start working on this ourselves soon. Again, many thanks for taking the time to contribute and apologies for communication delays :pray:

PR #2243, opened 2022-09-19
Treat non-200 HTTP responses as errors

aerosol:
### Changes

We're universally changing the way HTTPClient abstraction works - we'll now tag every non-200 response with an `:error`. Additionally we'll send a Sentry message every time get_invoices/1 fails.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/paddle_api.ex
+++ b/lib/plausible/billing/paddle_api.ex
@@ -89,15 +89,40 @@ defmodule Plausible.Billing.PaddleApi do
     end
   end
 
+  @spec get_invoices(Plausible.Billing.Subscription.t()) ::
+          {:ok, list()}
+          | {:error, :request_failed}
+          | {:error, :no_subscription}
   def get_invoices(nil), do: {:error, :no_subscription}
 
   def get_invoices(subscription) do
-    case do_get_invoices(subscription) do
-      {:ok, response} ->
-        response
+    config = get_config()
 
-      {:error, _} = e ->
-        e
+    params = %{
+      vendor_id: config[:vendor_id],
+      vendor_auth_code: config[:vendor_auth_code],
+      subscription_id: subscription.paddle_subscription_id,
+      is_paid: 1,
+      from: Timex.shift(Timex.today(), years: -5) |> Timex.format!("{YYYY}-{0M}-{0D}"),
+      to: Timex.shift(Timex.today(), days: 1) |> Timex.format!("{YYYY}-{0M}-{0D}")
+    }
+
+    with {:ok, response} <- HTTPClient.post(invoices_url(), @headers, params),
+         {:ok, body} <- Jason.decode(response.body),
+         true <- Map.get(body, "success"),
+         [_ | _] = response <- Map.get(body, "response") do
+      Enum.sort(response, fn %{"payout_date" => d1}, %{"payout_date" => d2} ->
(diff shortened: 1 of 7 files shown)


lib/plausible/billing/paddle_api.ex
```diff
@@ -92,39 +92,15 @@ defmodule Plausible.Billing.PaddleApi do
...
-      vendor_id: config[:vendor_id],
-      vendor_auth_code: config[:vendor_auth_code],
-      subscription_id: subscription.paddle_subscription_id,
-      is_paid: 1,
-      from: Timex.shift(Timex.today(), years: -5) |> Timex.format!("{YYYY}-{0M}-{0D}"),
-      to: Timex.shift(Timex.today(), days: 1) |> Timex.format!("{YYYY}-{0M}-{0D}")
-    }
-
-    case HTTPClient.post(invoices_url(), @headers, params) do
+    case do_get_invoices(subscription) do
```
ukutaht, line 95: This refactoring makes apparent that the return value of this function is inconsistent. Errors are wrapped in `{:error, e}` but success response is not tagged with `{:ok, res}`. Would be nice to add the `:ok` wrapping for consistency. The return value is pattern matched in a template which is a quick change.

lib/plausible/google/http.ex
```diff
@@ -138,7 +138,7 @@ defmodule Plausible.Google.HTTP do
       {:ok, %Finch.Response{body: body, status: 200}} ->
```
ukutaht, line 138: I suppose the `status: 200` match has become insignificant, and could even be misleading if the reader thinks it is significant.

lib/plausible/billing/paddle_api.ex
```diff
@@ -92,39 +92,15 @@ defmodule Plausible.Billing.PaddleApi do
   def get_invoices(nil), do: {:error, :no_subscription}
 
   def get_invoices(subscription) do
```
ukutaht, line 94: Good opportunity to add a test for this function as well

ukutaht (approved):

lib/plausible/billing/paddle_api.ex
aerosol: @ukutaht we can't yet, it's mocked at test time with this https://github.com/plausible/analytics/blob/treat-non200-http-responses-as-errors/test/support/paddle_api_mock.ex#L29-L52 :(

lib/plausible/billing/paddle_api.ex
aerosol: 9a44605b

lib/plausible/google/http.ex
aerosol: That depends. It looks like in some cases we're trying to decode the JSON body in case of non-200 responses (as opposed to other errors) and those might end up displayed in the UI templates. Maybe let's not touch that in this PR?

lib/plausible/billing/paddle_api.ex
ukutaht: Gotcha

RobertJoonas (approved):
Nice and quick fix, thanks @aerosol for your help! People will now be able to view their user settings page even when `get_invoices/1` fails, which I think is the most critical thing in this PR. And we'll also have some diagnostics of Paddle responses going forward. :slightly_smiling_face:

PR #2244, opened 2022-09-19
Improve GA Imports error reporting

vinibrsl:
### Changes

This pull request improves error reporting and documentation for Google Analytics imports. Currently unexpected errors from GA (unauthorized, timeouts, connection closed, etc) are raised as an exception. I changed this to a regular error tuple, reporting the site domain, error and GA response body to Sentry.

### Tests

I also tested this in our staging Sentry app.

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/api.ex
+++ b/lib/plausible/google/api.ex
@@ -151,12 +151,8 @@ defmodule Plausible.Google.Api do
     result
   end
 
-  @spec fetch_and_persist(Plausible.Site.t(), ReportRequest.t(),
-          buffer: pid(),
-          attempt: non_neg_integer(),
-          sleep_time: non_neg_integer(),
-          http_client: module()
-        ) :: :ok | {:error, term()}
+  @spec fetch_and_persist(Plausible.Site.t(), ReportRequest.t(), Keyword.t()) ::
+          :ok | {:error, term()}
   def fetch_and_persist(site, %ReportRequest{} = report_request, opts \\ []) do
     buffer_pid = Keyword.get(opts, :buffer)
     attempt = Keyword.get(opts, :attempt, 1)
--- a/lib/plausible/google/api.ex
+++ b/lib/plausible/google/api.ex
@@ -181,7 +181,7 @@ defmodule Plausible.Google.Api do
       {:error, cause} ->
         if attempt >= @max_attempts do
           Sentry.capture_message("Failed to import from Google Analytics",
-            extra: %{site: site.domain, error: cause}
+            extra: %{site: site.domain, error: inspect(cause)}
           )
 
           {:error, cause}
--- a/lib/plausible/google/api.ex
+++ b/lib/plausible/google/api.ex
@@ -104,11 +104,26 @@ defmodule Plausible.Google.Api do
   end
 
   @per_page 10_000
+  @max_attempts 5
+  @spec import_analytics(Plausible.Site.t(), Date.Range.t(), String.t(), String.t()) ::
+          :ok | {:error, term()}
   @doc """
-  API reference:
-  https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#ReportRequest
+  Imports stats from a Google Analytics UA view to a Plausible site.
 
-  Dimensions reference: https://ga-dev-tools.web.app/dimensions-metrics-explorer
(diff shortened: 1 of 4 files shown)


lib/plausible/google/api.ex
```diff
@@ -104,36 +104,59 @@ defmodule Plausible.Google.Api do
...
   @per_page 10_000
+  @max_attempts 5
+  @spec import_analytics(Plausible.Site.t(), Date.Range.t(), String.t(), String.t()) ::
+          :ok | {:error, term()}
   @doc """
-  API reference:
-  https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#ReportRequest
+  Imports stats from a Google Analytics UA view to a Plausible site.
+
+  This function fetches Google Analytics reports in batches of #{@per_page} per
```
aerosol, line 113: that's some clever use of interpolation :star:

lib/plausible/google/api.ex
```diff
@@ -104,36 +104,59 @@ defmodule Plausible.Google.Api do
...
+            page_size: @per_page
+        }
+
+        case fetch_and_persist(site, report_request, buffer: buffer) do
+          :ok -> {:cont, :ok}
+          {:error, _} = error -> {:halt, error}
+        end
+      end)
 
     Plausible.Google.Buffer.flush(buffer)
```
aerosol, line 148: Sanity check: do we still want to flush the buffer when an import has failed half way through?

lib/plausible/google/api.ex
```diff
@@ -104,36 +104,59 @@ defmodule Plausible.Google.Api do
...
     Plausible.Google.Buffer.stop(buffer)
 
-    :ok
+    result
   end
 
-  @max_attempts 5
+  @spec fetch_and_persist(Plausible.Site.t(), ReportRequest.t(),
+          buffer: pid(),
```
aerosol, line 155: I think you need to either use `Keyword.t()` or, if you like to be specific define a union of options such as 

```elixir
..., list(
  {:attempt, pos_integer()*} 
  | {:sleep_time, pos_integer()}
  | {:http_client | module()}}
)
```

Otherwise this enforces order, just like pattern matching on Keywords.

*`non_neg_integer` allows 0

lib/plausible/google/http.ex
```diff
@@ -41,6 +41,13 @@ defmodule Plausible.Google.HTTP do
          token <- Map.get(report, "nextPageToken"),
          {:ok, report} <- convert_to_maps(report) do
       {:ok, {report, token}}
+    else
+      {:ok, %{status: _non_http_200, body: body}} ->
```
aerosol, line 45: :roll_eyes: Heads up: this will break when https://github.com/plausible/analytics/pull/2243 is merged -- whoever goes second needs to rebase

aerosol (changes requested):
Typespecs need change, otherwise LGTM :+1:

ukutaht:
@aerosol's comments are on point. LGTM otherwise.

lib/plausible/google/api.ex
vinibrsl: Good thinking!

The worker that calls this function deletes all imported records when it fails for whatever reason. The buffer is actually flushed every 10,000 records, this additional call here is to kill the process safely without losing remaining records. So we should be fine :)

lib/plausible/google/http.ex
vinibrsl: Thank you! I'm second 😅

lib/plausible/google/api.ex
vinibrsl: I went with `Keyword.t()`. This function should be private anyways, but we have tests using it. If we migrate to Mox we'll have a chance to test this properly :)

lib/plausible/google/http.ex
aerosol: :hand_over_mouth: thanks

aerosol: I can still see the `{:ok, ...}` clause here though - it should be tagged with `:error` now as per #2243

aerosol: apologies, I didn't realise this is calling finch directly :see_no_evil:

PR #2250, opened 2022-09-20
Make Finch pools lighter for self-hosting

ukutaht:
### Changes

Previous discussion: https://github.com/plausible/analytics/pull/2224

The current pool config eats a lot of memory for self-hosted installations. The pool count seems excessive so this is an attempt to make the app more lean by default.

If it turns out we need beefier pool counts for Plausible.Cloud now or in the future, we can add a config option but keep the lighter pools in the default configuration for easy self-hosting.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -372,12 +372,6 @@ config :plausible, :hcaptcha,
   sitekey: hcaptcha_sitekey,
   secret: hcaptcha_secret
 
-config :plausible, Plausible.Finch,
-  default_pool_size: get_int_from_path_or_env(config_dir, "DEFAULT_FINCH_POOL_SIZE", 50),
-  default_pool_count: get_int_from_path_or_env(config_dir, "DEFAULT_FINCH_POOL_COUNT", 1),
-  sentry_pool_size: get_int_from_path_or_env(config_dir, "SENTRY_FINCH_POOL_SIZE", 50),
-  sentry_pool_count: get_int_from_path_or_env(config_dir, "SENTRY_FINCH_POOL_COUNT", 1)
-
 config :plausible, Plausible.Sentry.Client,
   finch_request_opts: [
     pool_timeout: get_int_from_path_or_env(config_dir, "SENTRY_FINCH_POOL_TIMEOUT", 5000),
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -43,7 +43,6 @@ defmodule Plausible.Application do
   defp finch_pool_config() do
     base_config = %{
       "https://icons.duckduckgo.com" => [
-        protocol: :http2,
         conn_opts: [transport_opts: [timeout: 15_000]]
       ]
     }
@@ -70,7 +69,6 @@ defmodule Plausible.Application do
     cond do
       paddle_conf[:vendor_id] && paddle_conf[:vendor_auth_code] ->
         Map.put(pool_config, Plausible.Billing.PaddleApi.vendors_domain(),
-          protocol: :http2,
           conn_opts: [transport_opts: [timeout: 15_000]]
         )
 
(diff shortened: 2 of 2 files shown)


lib/plausible/application.ex
```diff
@@ -41,41 +41,61 @@ defmodule Plausible.Application do
...
+    base_config
+    |> maybe_add_sentry_pool()
+    |> maybe_add_paddle_pool()
+    |> maybe_add_google_pools()
+  end
+
+  defp maybe_add_sentry_pool(pool_config) do
+    case Application.get_env(:sentry, :dsn) do
+      dsn when is_binary(dsn) ->
+        Map.put(pool_config, dsn, size: 50)
```
ukutaht, line 60: This is the only pool where we don't currently specify `:http2`

I'm not sure what the benefits of using it elsewhere is, and why we don't use http2 for Sentry. Didn't want to dive into this rabbit hole for now but something I noticed

lib/plausible/application.ex
```diff
@@ -41,41 +41,61 @@ defmodule Plausible.Application do
...
       "https://icons.duckduckgo.com" => [
         protocol: :http2,
-        count: 100,
         conn_opts: [transport_opts: [timeout: 15_000]]
       ]
     }
 
-    sentry_dsn = Application.get_env(:sentry, :dsn)
+    base_config
+    |> maybe_add_sentry_pool()
```
ukutaht, line 52: Not all these pools are needed for self-hosting so I'm leaving them out here if the relevant integrations have not been configured

lib/plausible/application.ex
aerosol: Maybe it's https://forum.sentry.io/t/sentry-requests-using-http-1-1-instead-of-h2/13371

I am looking to learn what is the rationale for using http2 everywhere else? Is there an obvious benefit for our use cases?

config/runtime.exs
```diff
@@ -372,12 +372,6 @@ config :plausible, :hcaptcha,
   sitekey: hcaptcha_sitekey,
   secret: hcaptcha_secret
 
-config :plausible, Plausible.Finch,
```
aerosol, line 375: Do we prefer hardcoded values to system env vars?

aerosol (approved):

lib/plausible/application.ex
ukutaht: > I am looking to learn what is the rationale for using http2 everywhere else? Is there an obvious benefit for our use cases?

No idea

aerosol: Perhaps we shouldn't use it until we find a good reason then? AFAIU http2 characteristics is the reason for this PR :grimacing:

vinibrsl (approved):
Looks good!

config/runtime.exs
aerosol: Yeah I mean we could advise some users to tweak those values in case they're low on resources. In the future we might want to deploy various configs (without rebuilding images) to take measurements. But I suppose this works too :+1:

lib/plausible/application.ex
ruslandoga: @ukutaht @aerosol sentry.io is a docs/marketing server, ingests servers support h2, I've been using it for some time with no issues.

Try https://sentry.io and then https://o1012425.ingest.sentry.io/ on https://tools.keycdn.com/http2-test

config/runtime.exs
ruslandoga: > self-hosters don't use Sentry

I actually do, and I think some other users might as well.

lib/plausible/application.ex
ukutaht: Thanks for the insights. I think it's best to aim for stability so I'll try removing `http2` config and using normal http1 pools with size 50.

SaphireLattice:
Bumped my self-hosted version to the 1.5 RC with this, absolutely decimated the memory usage by almost a whole gig. Which is pretty huge when the entire VPS is only 2GB of RAM. Before that, with some minor services running (nginx, weechat), plausible with clickhouse ended up getting total system memory usage to around 1.7GB

ruslandoga:
@SaphireLattice 👋

Did you enable any new features like city-level geolocation?

SaphireLattice:
Oh, I meant that the 1.5 is using 1GB _less_ memory than previous version! And huh, is that enabled by default?

PR #2257, opened 2022-09-22
Fix favicons

### LINKED ISSUE #2256: Favicon missing
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

It looks like some favicons are broken for the past 2 days. The URL it's calling is: `https://plausible.io/favicon/sources/Google` which just results in a not found image with http status code 200. There's a bunch of them that don't work in my dashboard.

Screen Shot 2022-09-22 at 11 01 40

### Expected behavior

The favicon to show like before instead of an empty white square.

### Screenshots

See above

### Environment

```markdown
- OS: macOS
- Browser: Safari
- Browser Version: Version 16.0 (17614.1.25.9.10, 17614)
```

ukutaht:
### Changes

Some HTTP changes broke header handling for the favicon proxy plug.

We remove the `transfer-encoding` header because it's only meant for one hop and breaks things if it's forwarded to the client. This filter stopped working, probably when we changed HTTP clients, because response headers used to not be processed, but now they are downcased by the HTTP lib.

There were no tests for `PlausibleWeb.Favicon` so I took the opportunity to try out Mox for this use-case. @aerosol curious to see what you think. Recommended to review commit-by-commit.

Closes #2256

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -50,6 +50,13 @@ defmodule Plausible.HTTPClient do
     call(:get, url, headers, nil)
   end
 
+  # TODO: Is it possible to tell the type checker that we're returning a module that conforms to the
+  # Plausible.HTTPClient.Interface behaviour?
+  @spec impl() :: module()
+  def impl() do
+    Application.get_env(:plausible, :http_impl, __MODULE__)
+  end
+
   defp call(method, url, headers, params) do
     {params, headers} = maybe_encode_params(params, headers)
 
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -16,7 +16,7 @@ defmodule Plausible.HTTPClient.Interface do
   @type params() :: Finch.Request.body() | map()
   @type response() ::
           {:ok, Finch.Response.t()}
-          | {:error, Mint.Types.error() | Finch.Error.t() | Non200Error.t()}
+          | {:error, Mint.Types.error() | Finch.Error.t() | Plausible.HTTPClient.Non200Error.t()}
 
   @callback get(url(), headers()) :: response()
   @callback get(url()) :: response()
@@ -54,7 +54,7 @@ defmodule Plausible.HTTPClient do
 
   # TODO: Is it possible to tell the type checker that we're returning a module that conforms to the
   # Plausible.HTTPClient.Interface behaviour?
-  @spec impl() :: module()
+  @spec impl() :: Plausible.HTTPClient
   def impl() do
     Application.get_env(:plausible, :http_impl, __MODULE__)
   end
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -19,6 +19,7 @@ defmodule Plausible.HTTPClient.Interface do
           | {:error, Mint.Types.error() | Finch.Error.t() | Non200Error.t()}
 
   @callback get(url(), headers()) :: response()
+  @callback get(url()) :: response()
   @callback post(url(), headers(), params()) :: response()
 end
 
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -24,7 +24,6 @@ defmodule Plausible.HTTPClient.Interface do
 end
 
 defmodule Plausible.HTTPClient do
-  @behaviour Plausible.HTTPClient.Interface
   @moduledoc """
   HTTP Client built on top of Finch.
 
@@ -38,6 +37,8 @@ defmodule Plausible.HTTPClient do
   @doc """
   Make a POST request
   """
+  @behaviour Plausible.HTTPClient.Interface
+
   @impl Plausible.HTTPClient.Interface
   def post(url, headers \\ [], params \\ nil) do
     call(:post, url, headers, params)
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -1,26 +1,16 @@
-defmodule Plausible.HTTPClient do
-  @moduledoc """
-  HTTP Client built on top of Finch.
+defmodule Plausible.HTTPClient.Non200Error do
+  defstruct reason: nil
 
-  By default, request parameters are json-encoded.
+  @type t :: %__MODULE__{reason: Finch.Response.t()}
 
-  If a raw binary value is supplied, no encoding is performed.
-  If x-www-form-urlencoded content-type is set in headers, 
-  URL encoding is invoked.
-  """
-
-  defmodule Non200Error do
-    defstruct reason: nil
-
-    @type t :: %__MODULE__{reason: Finch.Response.t()}
-
-    @spec new(Finch.Response.t()) :: t()
-    def new(%Finch.Response{status: status} = response)
-        when is_integer(status) and (status < 200 or status >= 300) do
-      %__MODULE__{reason: response}
-    end
+  @spec new(Finch.Response.t()) :: t()
+  def new(%Finch.Response{status: status} = response)
+      when is_integer(status) and (status < 200 or status >= 300) do
+    %__MODULE__{reason: response}
   end
+end
 
+defmodule Plausible.HTTPClient.Interface do
   @type url() :: Finch.Request.url()
   @type headers() :: Finch.Request.headers()
   @type params() :: Finch.Request.body() | map()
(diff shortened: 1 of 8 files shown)


lib/plausible/http_client.ex
```diff
@@ -1,49 +1,63 @@
...
   @doc """
   Make a GET request
   """
-  @spec get(url(), headers()) :: response()
+  @impl Plausible.HTTPClient.Interface
   def get(url, headers \\ []) do
     call(:get, url, headers, nil)
   end
 
+  # TODO: Is it possible to tell the type checker that we're returning a module that conforms to the
```
aerosol, line 54: we can say `EitherThisModule | OrEitherThatModule` I think?

lib/plausible_web/plugs/favicon.ex
```diff
@@ -1,10 +1,37 @@
 defmodule PlausibleWeb.Favicon do
+  @referer_domains_file "priv/referer_favicon_domains.json"
+  @moduledoc """
+    A Plug that fetches favicon images from DuckDuckGo and returns them
+    to the Plausible frontend.
+
+    The proxying is there so we can reduce the number of third-party domains that
+    the browser clients need to connect to. Our goal is to have 0 third-party domain
+    connections on the website for privacy reasons.
```
aerosol, line 9: :raised_hands: thank you for documenting that aspect as well!

lib/plausible_web/plugs/favicon.ex
```diff
@@ -1,10 +1,37 @@
...
+    The mapping from source category -> source hostname is stored in "#{@referer_domains_file}" and
+    managed by `Mix.Tasks.GenerateReferrerFavicons.run/1`
+  """
   import Plug.Conn
   alias Plausible.HTTPClient
 
   def init(_) do
     domains =
-      case File.read(Application.app_dir(:plausible, "priv/referer_favicon_domains.json")) do
+      case File.read(Application.app_dir(:plausible, @referer_domains_file)) do
```
aerosol, line 34: can't we `File.read!` here? This file is part of version control and init/1 is called on compile time so we can assume it _has_ to be there, no?

lib/plausible_web/plugs/favicon.ex
```diff
@@ -34,13 +68,23 @@ defmodule PlausibleWeb.Favicon do
...
+    "keep-alive",
+    "proxy-authenticate",
+    "proxy-authorization",
+    "te",
+    "trailers",
+    "transfer-encoding",
+    "upgrade"
+  ]
+  defp remove_hop_by_hop_headers(headers) do
+    Enum.filter(headers, fn {key, _} -> key not in @hop_by_hop_headers end)
```
aerosol, line 88:
```suggestion
    Enum.reject(headers, fn {key, _} -> key in @hop_by_hop_headers end)
```

aerosol: I'm wondering though, instead of forwarding the response and stripping out the headers, can't we only copy DDG's body without cloning anything else? Are there headers we explicitly want to keep and forward? If so, can we document that?

test/plausible_web/plugs/favicon_test.exs
```diff
@@ -0,0 +1,90 @@
...
+
+    %{plug_opts: opts}
+  end
+
+  test "ignores request on a URL it does not need to handle", %{plug_opts: plug_opts} do
+    conn =
+      conn(:get, "/irrelevant")
+      |> Favicon.call(plug_opts)
+
+    refute conn.halted
```
aerosol, line 20: perhaps we can assert here that `new_conn == old_conn` if we're testing it's intact?

test/plausible_web/plugs/favicon_test.exs
```diff
@@ -0,0 +1,90 @@
...
+
+    conn =
+      conn(:get, "/favicon/sources/plausible.io")
+      |> Favicon.call(plug_opts)
+
+    assert conn.halted
+    assert conn.status == 200
+
+    assert conn.resp_headers == [
+             {"content-type", "should-pass-through"}
```
aerosol, line 87: ditto - maybe we can have a "forward headers list" instead of a block list?

lib/plausible_web/plugs/favicon.ex
```diff
@@ -1,10 +1,37 @@
 defmodule PlausibleWeb.Favicon do
+  @referer_domains_file "priv/referer_favicon_domains.json"
```
vinibrsl, line 2: Nice job adding docs! ✨

I think the convention is having `@moduledoc` first, and every other attribute after.

lib/plausible/http_client.ex
```diff
@@ -1,49 +1,63 @@
...
           {:ok, Finch.Response.t()}
           | {:error, Mint.Types.error() | Finch.Error.t() | Non200Error.t()}
 
+  @callback get(url(), headers()) :: response()
+  @callback get(url()) :: response()
+  @callback post(url(), headers(), params()) :: response()
+end
+
+defmodule Plausible.HTTPClient do
+  @behaviour Plausible.HTTPClient.Interface
```
vinibrsl, line 27: Some `@moduledoc` being the first attribute suggestion.

lib/plausible/http_client.ex
vinibrsl: Not dynamically possible unfortunately. Adam's suggestion is our best bet to typespec this.

test/test_helper.exs
```diff
@@ -1,6 +1,8 @@
 {:ok, _} = Application.ensure_all_started(:ex_machina)
 Plausible.Test.ClickhouseSetup.run()
 Mimic.copy(FunWithFlags)
+Mox.defmock(Plausible.HTTPClient.Mock, for: Plausible.HTTPClient.Interface)
+Application.put_env(:plausible, :http_impl, Plausible.HTTPClient.Mock)
```
vinibrsl, line 5: Any concerns moving this to `config/test.exs`?

lib/plausible_web/plugs/favicon.ex
ukutaht: The compiler disagrees :) This is what I get when I move the @moduledoc above the @referrer_domains_file attribute

```bash
➜  analytics git:(fix-favicons) mix compile
dependency :hut is using Rebar 2, which is no longer maintained and no longer works in recent Erlang/OTP versions. Remove the :manager option or set it to :rebar3 instead
Compiling 5 files (.ex)
warning: undefined module attribute @referer_domains_file, please remove access to @referer_domains_file or explicitly set it before access
  lib/plausible_web/plugs/favicon.ex:25: PlausibleWeb.Favicon (module)

Generated plausible app
```

lib/plausible/http_client.ex
ukutaht: This one I can do :)

lib/plausible/http_client.ex
ukutaht: Managed by just returning `Plausible.HTTPClient`

lib/plausible_web/plugs/favicon.ex
ukutaht: > init/1 is called on compile time

TIL. Thanks :)

test/plausible_web/plugs/favicon_test.exs
ukutaht: 👍

test/plausible_web/plugs/favicon_test.exs
```diff
@@ -0,0 +1,90 @@
+defmodule PlausibleWeb.FaviconTest do
+  use Plausible.DataCase
+  use Plug.Test
+  alias PlausibleWeb.Favicon
+
+  import Mox
+  setup :verify_on_exit!
```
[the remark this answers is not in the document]
ukutaht: Not sure why Mox does export `use Mox` macro that would expand to:

```elixir
import Mox
setup :verify_on_exit!
```

Basically I forgot the `verify_on_exit` when I used Mox the first time and am salty about it now :)

test/plausible_web/plugs/favicon_test.exs
ukutaht: Yeah makes sense - I've changed it to just copy `content-type,cache-control,expires` headers

test/test_helper.exs
ukutaht: Did it. Couldn't move the `Mox.defmock(Plausible.HTTPClient.Mock, for: Plausible.HTTPClient.Interface)` because it needs to be evaluated at runtime

vinibrsl (approved):

PR #2266, opened 2022-09-23
Stats API: allow escaping `|` literal character with `\|` when filtering

RobertJoonas:
### Changes

- Allow escaping the `|` character with `\|` in Stats API filter values, to prevent the filter type being parsed to `:member`

### Tests

### Changelog

### Documentation
- [x] Documentation update PR is pending: https://github.com/plausible/docs/pull/279

### Dark mode

### DIFF
--- a/lib/plausible/stats/filter_parser.ex
+++ b/lib/plausible/stats/filter_parser.ex
@@ -16,26 +16,39 @@ defmodule Plausible.Stats.FilterParser do
     |> Enum.into(%{})
   end
 
+  @non_escaped_pipe_regex ~r/(?<!\\)\|/
   defp parse_single_filter(str) do
-    [key, val] =
+    [key, raw_value] =
       String.trim(str)
       |> String.split(["==", "!="], trim: true)
       |> Enum.map(&String.trim/1)
 
     is_negated = String.contains?(str, "!=")
-    is_list = String.contains?(val, "|")
-    is_wildcard = String.contains?(val, "*")
+    is_list = Regex.match?(@non_escaped_pipe_regex, raw_value)
+    is_wildcard = String.contains?(raw_value, "*")
+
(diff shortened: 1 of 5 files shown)


aerosol (approved):
Looks fine, it's great that we have parser-level tests now so we'll keep track of regressions moving forward :heart:

lib/plausible/stats/filter_parser.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  ### Examples:
+
+      iex> FilterParser.parse_filters(%{"filters" => "{\\"page\\":\\"/blog/**\\"}"})
+      %{"page" => "/blog/**"}
+
+      iex> FilterParser.parse_filters(%{"filters" => "visit:browser!=Chrome"})
+      %{"visit:browser" => {:is_not, "Chrome"}}
+  """
+  def parse_filters(%{"filters" => filters}) when is_binary(filters) do
```
ukutaht, line 25: Only thing that bothers me a bit is this module taking a Map as input. Feels like it's concerned with the internals of the `Query` struct and doing something that's unrelated to filter parsing. I think the separation would be cleaner if this module just took the value of the `filters` key:

```elixir
# in query.ex
FilterParser.parse_filters(query["filters"])
```

```elixir
# in filter_parser.ex
def parse_filters(filters) when is_binary(filters) do
   ...
end
```

ukutaht (commented):
Looks good! One comment inline

lib/plausible/stats/filter_parser.ex
RobertJoonas: Good point! Did it now: https://github.com/plausible/analytics/pull/2266/commits/6ae6a8f3d9e6c0235ca69507e662ca56813e3e1f

ukutaht (approved):

PR #2272, opened 2022-09-26
Add script to send pageview

ukutaht:
### Changes

For development we need a convenient way to generate fake pageview events.

Currently this is done via `make dummy_event` but it's not very convenient because you cannot
override parameters from the command line like so:

```bash
$ make dummy_event --ip 127.0.0.2
$ make dummy_event --domain some-test-domain.com
```

In order to support command-line options like this, I thought I would write a small Mix task for it:
```bash
$ mix send_pageview --ip 127.0.0.2
$ mix send_pageview --domain some-test-domain.com
```

I'm not sure of the value of `Make` in this project in general. Mix is a great build tool and task runner, do we need another one?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ Make sure Docker, Elixir, Erlang and Node.js are all installed on your developme
 1. Navigate to `http://localhost:8000/register` and fill in the form.
 2. Fill in the rest of the forms and for the domain use `dummy.site`
 3. Skip the JS snippet and click start collecting data.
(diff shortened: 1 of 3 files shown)


vinibrsl (approved):
This is really helpful! I was often changing my Makefile to do some tests.

> I'm not sure of the value of Make in this project in general. Mix is a great build tool and task runner, do we need another one?

I don't think we need Make. I also think that our Docker commands inside the Makefile could be a docker-compose file, which suits that kind of scripting.

PR #2274, opened 2022-09-26
Increase GA import request receive_timeout to 30s

vinibrsl:
### Changes

This commit increases the receive_timeout from 15s to 30s for Google Analytics import requests. Timeouts have been reported, and I could replay some requests that took 15-20 seconds to complete when fetching 10,000 records in a single request.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/http.ex
+++ b/lib/plausible/google/http.ex
@@ -34,7 +34,7 @@ defmodule Plausible.Google.HTTP do
         [{"Authorization", "Bearer #{report_request.access_token}"}],
         params
       )
-      |> http_client.request(Plausible.Finch)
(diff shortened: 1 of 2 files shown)


vinibrsl:
> Could the same be achieved by changing the value on this line?
> 
> https://github.com/plausible/analytics/blob/519a1fa669f622f9fa8ddcd098e190e1ba946062/lib/plausible/application.ex#L87

I thought it would, but those are sent to [`Mint.HTTP.connect/4`](https://hexdocs.pm/mint/1.4.0/Mint.HTTP.html#connect/4-transport-options) which only takes `connect_timeout`, the initial phase of the request. The timeout we want is `receive_timeout`, which is the total time of the request.

ukutaht (approved):

PR #2285, opened 2022-09-29
Redirect user after upgrading only when subscription is created

vinibrsl:
### Changes

We've had people upgrading twice, resulting in duplicate subscriptions. This happens because subscriptions are created from Paddle asynchronous webhooks, and this process may take some time.

This pull request redirect the user after upgrading only when the subscription is created, using a polling strategy.

image

### Tests

### Changelog
Not sure if this needs a changelog update?

### Documentation

### Dark mode


metmarkosaric:
nice 👏

lib/plausible_web/templates/billing/upgrade_success.html.eex
```diff
@@ -1,18 +1,40 @@
...
     <div class="w-full pt-2 text-xl font-bold dark:text-gray-100">
-      Subscription created successfully
+      Your account is being upgraded...
     </div>
 
-    <div class="w-full text-gray-500 dark:text-gray-200 text-sm py-4">
-      Thank you for upgrading your subscription. You will be redirected to your
-      account within 5 seconds.
-    </div>
+    <p class="w-full text-gray-500 dark:text-gray-200 text-sm py-4">
```
ukutaht, line 7: Can we make this page show a big spinner or some other type of animating thing that's known to increase patience in UX? I think it would be a good idea to get people to linger on this page a bit longer.

Additionally, I think we should add a timeout after 15 seconds or so that adds another message like "Your subscription is taking longer than usual to upgrade. If you're not redirected soon, please contact hello@plausible.io".

ukutaht (commented):
Looks good. My only concern is the UX on that page, I think it can be improved.

lib/plausible_web/templates/billing/upgrade_success.html.eex
vinibrsl: There is currently an animated spinner.

I added another message that shows up after 15 seconds (c798e8263336406318e42cd480856c877dfa4f32). Let me know what you think!

image

lib/plausible_web/templates/billing/upgrade_success.html.eex
```diff
@@ -4,34 +4,45 @@
...
     </div>
 
-    <p class="w-full text-gray-500 dark:text-gray-200 text-sm py-4">
+    <p class="text-gray-500 dark:text-gray-200 text-sm py-4">
       Thank you for upgrading your subscription! We're still working on 
       upgrading your account, and you'll be automatically redirected in a few
       seconds.
     </p>
 
+    <p hidden id="timeout-notice" class="text-gray-500 dark:text-gray-200 text-sm">
```
ukutaht, line 13: TIL about the `hidden` HTML attribute

lib/plausible_web/templates/billing/upgrade_success.html.eex
ukutaht: Fantastic, thank you!

ukutaht (approved):

PR #2291, opened 2022-10-01
Improve user input validation

aerosol:
### Changes

This improves our input data validation at various places to address the following Sentry errors:

  - APP-1JA
  - APP-22Y
  - APP-21T
  - APP-A6

We now don't crash but rather silently discard the garbage sent our way.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -317,18 +317,8 @@ defmodule Plausible.Stats.Query do
   defp parse_single_date(tz, params) do
     case params["date"] do
       "today" -> Timex.now(tz) |> Timex.to_date()
-      date when is_binary(date) -> sanitize_date(tz, date)
+      date when is_binary(date) -> Date.from_iso8601!(date)
       _ -> Timex.now(tz) |> Timex.to_date()
-    end
-  end
-
-  defp sanitize_date(tz, input_date) do
-    case Date.from_iso8601(input_date) do
-      {:ok, parsed_date} ->
-        parsed_date
-
-      _ ->
-        Timex.now(tz) |> Timex.to_date()
     end
   end
 
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -317,8 +317,18 @@ defmodule Plausible.Stats.Query do
   defp parse_single_date(tz, params) do
     case params["date"] do
       "today" -> Timex.now(tz) |> Timex.to_date()
-      date when is_binary(date) -> Date.from_iso8601!(date)
+      date when is_binary(date) -> sanitize_date(tz, date)
       _ -> Timex.now(tz) |> Timex.to_date()
+    end
+  end
+
+  defp sanitize_date(tz, input_date) do
+    case Date.from_iso8601(input_date) do
+      {:ok, parsed_date} ->
+        parsed_date
+
+      _ ->
+        Timex.now(tz) |> Timex.to_date()
     end
   end
 
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -924,10 +924,22 @@ defmodule PlausibleWeb.Api.StatsController do
   end
 
   defp parse_pagination(params) do
-    limit = if params["limit"], do: String.to_integer(params["limit"]), else: 9
-    page = if params["page"], do: String.to_integer(params["page"]), else: 1
+    limit = to_int(params["limit"], 9)
+    page = to_int(params["page"], 1)
     {limit, page}
   end
+
+  defp to_int(string, default) when is_binary(string) do
+    case Integer.parse(string) do
+      {i, ""} when is_integer(i) ->
+        i
+
+      _ ->
+        default
+    end
+  end
+
+  defp to_int(_, default), do: default
 
   defp maybe_add_percentages(stat_list, query) do
     if Map.has_key?(query.filters, "event:goal") do
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -5,8 +5,6 @@ defmodule PlausibleWeb.Api.StatsController do
   alias Plausible.Stats
   alias Plausible.Stats.{Query, Filters}
 
-  require Logger
-
   def main_graph(conn, params) do
     site = conn.assigns[:site]
 
@@ -1053,8 +1051,6 @@ defmodule PlausibleWeb.Api.StatsController do
   end
 
   defp bad_request(conn) do
-    Logger.error("Input validation error in #{__MODULE__}, returning Bad Request to the client.")
-
     conn
     |> put_status(400)
     |> json(%{error: "input validation error"})
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -5,56 +5,72 @@ defmodule PlausibleWeb.Api.StatsController do
   alias Plausible.Stats
   alias Plausible.Stats.{Query, Filters}
 
+  require Logger
+
   def main_graph(conn, params) do
     site = conn.assigns[:site]
-    query = Query.from(site, params) |> Filters.add_prefix()
 
-    selected_metric =
-      if !params["metric"] || params["metric"] == "conversions" do
-        "visitors"
-      else
-        params["metric"]
-      end
+    with :ok <- validate_params(params) do
+      query = Query.from(site, params) |> Filters.add_prefix()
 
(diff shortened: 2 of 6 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -924,11 +924,23 @@ defmodule PlausibleWeb.Api.StatsController do
...
 
   defp parse_pagination(params) do
-    limit = if params["limit"], do: String.to_integer(params["limit"]), else: 9
-    page = if params["page"], do: String.to_integer(params["page"]), else: 1
+    limit = to_int(params["limit"], 9)
+    page = to_int(params["page"], 1)
     {limit, page}
   end
 
+  defp to_int(string, default) when is_binary(string) do
```
ukutaht, line 932: I feel like this function should be part of the standard lib. Always miss it.

ukutaht (approved):

lib/plausible/stats/query.ex
```diff
@@ -317,11 +317,21 @@ defmodule Plausible.Stats.Query do
   defp parse_single_date(tz, params) do
     case params["date"] do
       "today" -> Timex.now(tz) |> Timex.to_date()
-      date when is_binary(date) -> Date.from_iso8601!(date)
+      date when is_binary(date) -> sanitize_date(tz, date)
```
vinibrsl, line 320: It's nice to have fallbacks, but in this specific case, wouldn't it be better to respond with a date format error message and HTTP 400?

As an API user wanting to use a date filter, I rather have an error message than a fallback + HTTP 200 that may seem like my date format is correct.

What do you think?

aerosol: Right, I see your point. 

This module (`Query`) seems to be shared in different contexts:
  - `SendEmailReport` and `SpikeNotifier` (`Oban` workers)
  - `StatsController` (private react API)
  - `ExternalStatsController` (public API)
 
The issue I was trying is to fix was `StatsController` abuse (someone trying to send malicious payload to an endpoint that expects somewhat "trusted" input assumed to be crafted by the react front-end). 

At the `Query` module level where this change was made there is no HTTP context, and no validations done, so I'm wondering if I should revert that change and maybe implement a controller-level validation similar to https://github.com/plausible/analytics/blob/12a513f3cdcfdd5e046538de7d8b0076e72d8c62/lib/plausible_web/controllers/api/external_stats_controller.ex#L190-L218 in `StatsController`? Do you think that'd be better?

lib/plausible_web/controllers/api/stats_controller.ex
aerosol: Yeah it feels like `String.to_integer` should be `String.to_integer!` :)

lib/plausible/stats/query.ex
ukutaht: Hmm yeah good point @vinibrsl.

Ideally, I suppose, the `Query` module would all the validations and return appropriate error tuples when the date, filters, or anything like that cannot be parsed. And then `ExternalStatsController` and `StatsController` can decide what to do on the HTTP level if parsing fails.

What do you think @aerosol ?

aerosol: I am a little conflicted on that one @ukutaht. 

On one hand, with `Query` being central to all those contexts, it makes sense to push the responsibility downstream. On the other, it's unclear which parts of the `Query` interface are meant for which context specifically. 

Correct me if I'm wrong please -- there seem to be differences in what can be built in public vs private Stats APIs _and_ sometimes the input is purely internal, so trusted by definition (in case of Oban workers). Perhaps a little bit of redundancy doesn't hurt if that leads to a clearer picture. 

What else do we need to validate apart from date and filters alone? I'm seeing `sample_threshold` and `interval` coming from `params` too. Moving validations to `Query` for the sake of DRY alone is not great IMHO.

I'm thinking that ideally we'd split the generic `Query.from` into specific use cases to draw a clearer scope. Let me know your thoughts.

lib/plausible_web/controllers/api/stats_controller.ex
aerosol: I was pointed at

```elixir
Ecto.Type.cast(:integer, "42")
{:ok, 42}
```

lib/plausible/stats/query.ex
ukutaht: Yeah the use-cases are slightly different

**Background job / internal context:**
1. Validations aren't strictly needed
2. Could use a dedicated `from` function which doesn't do any parsing. Currently we do `Date.to_iso8601` in the Oban worker and then parse it in Query, which is a weird way to create the data structure

**ExternalStatsController & StatsController**
Since these are both exposed to the world, and coming over the network, parsing & validations are clearly needed. The same rules apply, we need to parse and validate `period,date,filters,interval,sample_threshold,include_imported`

There are some nuances though:
1. `include_imported` is only used in the internal stats controller. Imported stats are not part of the public API at the moment. Although they accidentally might be at the moment? Hyrum's law and what not
2. `sample_threshold` is not documented, built for a specific customer in the public API for exporting unsampled data. It is not used by anyone else that I know of. In the future, configuring sample threshold should probably be some sort of premium API access plan.

aerosol: @vinibrsl thoughts on this?

vinibrsl: I understand that there are different scopes and rules for each caller. That being said I think `Query` should have a strong contract with types and permitted parameters, and the contexts (HTTP for example) should be responsible for parsing that into `Query`'s contract.

If that is too much work, we can have the date format error as I suggested first, and add this improvement to the backlog.

aerosol: > That being said I think Query should have a strong contract with types and permitted parameters, and the contexts (HTTP for example) should be responsible for parsing that into Query's contract.

Could you rephrase that please? Where exactly do you see validations taking place? I'm a little confused what do you mean by "parse into" here.

vinibrsl: I thought this through again, and rephrasing my last comment, my suggestion's end goal is similar to yours.

> @aerosol: I'm thinking that ideally we'd split the generic Query.from into specific use cases to draw a clearer scope.

Having different use cases for each caller (internal calls, public HTTP API and "private" HTTP API). I believe `Query` should have a strong contract, that is, for example, timestamps are `%DateTime{}` structs and not strings. That moves the responsibility of parsing external input and returning errors to the caller. I see ISO-8601 strings being parsed by the controller before passing to `Query`.

aerosol: I like that @vinibrsl, thanks, let me see what I can do about it

ukutaht: Maybe separate out `QueryParser` from `Query`. Then the background jobs could use `Query` directly whereas the controllers would use `QueryParser` which parses and validates untrusted input.

ukutaht: Just a thought

aerosol: This is more involved than it looked like initially (I see there's also `Plausible.Stats.Filters` involved that patches the `Query` conditionally). If you don't mind I'm going to move the DateTime parsing bit to the other controller for now and plant a backlog item. Don't want to spend too much time on this PR, as shipping it as is would be already an improvement.

vinibrsl: Agreed!

ukutaht: +1

aerosol: Yeah, I didn't give up just yet but there's so many quirks -- the internal API happily accepts "realtime" period (that is in turn changed to 30m), whereas the external API doesn't like those. I think I'm going to validate date alone for now in the controller.

aerosol: addc5e50 and 6c664956

ukutaht: > the internal API happily accepts "realtime" period (that is in turn changed to 30m)

Oh boy, that's a bit messy indeed. The realtime dashboard is very quirky, it mixes 5m time period with 30m time period. We've had some complaints about this from customers as well: https://github.com/plausible/analytics/discussions/1532

The realtime dashboard also has some bugs and we're intending to look at it soonish in the backlog. Let's not worry about it here and now, but I wanted to provide this context because we'll need to improve the realtime dashboard soonish.

ukutaht: `period=realtime` has been requested for the public API as well: https://github.com/plausible/analytics/discussions/1744

aerosol: I am so bookmarking this thread :sweat_smile:

aerosol:
Made it green with enough retries :) @ukutaht shall we merge it?

vinibrsl (approved):

PR #2310, opened 2022-10-05
Override stats_start_date in CRM

vinibrsl:
### Changes

This pull request adds the `stats_start_date` field to the CRM site form. This allows this field to be manually overridden, so Google Analytics imports date ranges can be overridden also.

Reviewing commit-by-commit can help understand some decisions and minor refactoring we made. 

### Tests
We don't have a test suite for the CRM.

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/admin.ex
+++ b/lib/plausible/site/admin.ex
@@ -21,7 +21,8 @@ defmodule Plausible.SiteAdmin do
     [
       domain: %{update: :readonly},
       timezone: nil,
-      public: nil
+      public: nil,
+      stats_start_date: nil
     ]
   end
 
@@ -130,4 +131,7 @@ defmodule Plausible.SiteAdmin do
     end)
     |> stringify_fields()
   end
+
+  def create_changeset(schema, attrs), do: Plausible.Site.crm_changeset(schema, attrs)
+  def update_changeset(schema, attrs), do: Plausible.Site.crm_changeset(schema, attrs)
 end
--- a/lib/plausible/site/admin.ex
+++ b/lib/plausible/site/admin.ex
@@ -19,7 +19,7 @@ defmodule Plausible.SiteAdmin do
 
   def form_fields(_) do
     [
-      domain: nil,
+      domain: %{update: :readonly},
       timezone: nil,
       public: nil
     ]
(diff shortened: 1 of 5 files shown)


lib/plausible/sites.ex
```diff
@@ -39,18 +39,28 @@ defmodule Plausible.Sites do
...
-      start_date = Plausible.Stats.Clickhouse.pageview_start_date_local(site)
+  @spec stats_start_date(Plausible.Site.t()) :: Date.t() | nil
+  @doc """
+  Returns the date of the first event of the given site, or `nil` if the site
+  does not have stats yet.
+
+  If this is the first time the function is called for the site, it queries
+  Clickhouse and saves the date in the sites table.
+  """
+  def stats_start_date(site)
```
aerosol, line 50: minor nitpick: stats and start look so similar it reads a little weird, maybe consider `events_start_at`?

aerosol (approved):

lib/plausible/site/admin.ex
```diff
@@ -19,9 +19,10 @@ defmodule Plausible.SiteAdmin do
 
   def form_fields(_) do
     [
-      domain: nil,
+      domain: %{update: :readonly},
       timezone: nil,
```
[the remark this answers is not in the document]
vinibrsl: Good call! https://github.com/plausible/analytics/pull/2310/commits/e51304751f6db73f77b984befefef4055c7d2106

lib/plausible/sites.ex
vinibrsl: I agree with you. However I think we should stick with `stats_start_date`, as that is the name of the field in the database and other places as well. Maybe it is better to have one not-to-good name than two different ones over the codebase?

lib/plausible/site/admin.ex
```diff
@@ -20,7 +20,7 @@ defmodule Plausible.SiteAdmin do
   def form_fields(_) do
     [
       domain: %{update: :readonly},
-      timezone: nil,
+      timezone: %{choices: Plausible.Timezones.options()},
```
aerosol, line 23: That's nice but won't let you quickly select one by typing in first letters

lib/plausible/sites.ex
aerosol: fine by me :+1:

lib/plausible/site/admin.ex
aerosol: right but I meant as opposed to https://github.com/plausible/analytics/pull/2310#discussion_r988018168 -- no strong feelings whatsoever

PR #2315, opened 2022-10-07
Fix and enable dialyzer

ukutaht:
### Changes

Fixed dialyzer errors and enabled it in CI.

I had to mark one function as `:nowarn` because we use an ecto feature that does not have a proper typespec.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Generate tracker files
         run: npm run deploy --prefix ./tracker
       - name: Check Credo Warnings
-        run: mix credo suggest --files-included $(git diff --name-only origin/master HEAD | grep -E 'lib/|test/')
+        run: git diff --name-only master HEAD | grep -E 'lib/|test/' | xargs -I % mix credo suggest --files-included %
       - name: Run tests
         run: mix test
       - name: Check Dialyzer
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -66,11 +66,11 @@ jobs:
         run: mix compile --warnings-as-errors --all-warnings
       - name: Generate tracker files
         run: npm run deploy --prefix ./tracker
+      - name: Check Credo Warnings
+        run: mix credo diff --from-git-merge-base master
       - name: Run tests
         run: mix test
-      #- name: Check Dialyzer
-      #  run: mix dialyzer <-- false negatives
-      - name: Check Credo Warnings
-        run: mix credo --strict --only warning --ignore Consistency
+      - name: Check Dialyzer
+        run: mix dialyzer
 env:
   MIX_ENV: test
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -41,6 +41,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Read .tool-versions
         uses: marocchino/tool-versions-action@v1
         id: versions
@@ -67,7 +69,7 @@ jobs:
       - name: Generate tracker files
         run: npm run deploy --prefix ./tracker
       - name: Check Credo Warnings
-        run: mix credo diff --from-git-merge-base master
+        run: mix credo suggest --files-included $(git diff --name-only master HEAD | grep -E 'lib/|test/')
       - name: Run tests
         run: mix test
       - name: Check Dialyzer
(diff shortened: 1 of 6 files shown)


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -217,12 +218,13 @@ defmodule PlausibleWeb.StatsController do
 
   defp find_shared_link(domain, auth) do
     link_query =
-      from link in Plausible.Site.SharedLink,
```
ukutaht, line 220: Sorry I think my editor added some formatting changes as well

ukutaht:
What do we think about Credo? 

> 1396 mods/funs, found 2 consistency issues, 26 refactoring opportunities, 114 code readability issues, 2 software design suggestions.

ukutaht:
> What do we think about Credo?
> 
> > 1396 mods/funs, found 2 consistency issues, 26 refactoring opportunities, 114 code readability issues, 2 software design suggestions.

The issues are quite simple, we could go down the list, fix them and then enable Credo in CI as well

vinibrsl:
> What do we think about Credo?
> 
> > 1396 mods/funs, found 2 consistency issues, 26 refactoring opportunities, 114 code readability issues, 2 software design suggestions.

Credo would be a nice addition to our CI pipeline. It would require fixing current issues I think, but helps us maintaining good Elixir conventions and style consistency.

ukutaht:
> helps us maintaining good Elixir conventions and style consistency.

For me the biggest value is taking some discussions / arguments / disagreements off the table. I find dev teams spend a lot of time bikeshedding over code style which is down to personal preference. Having an automated rule system means we don't need to spend time arguing about style and can spend more energy on the substance.

ukutaht:
@ruslandoga has opened a PR on Ecto that will allow us to remove the `:nowarn` hint https://github.com/elixir-ecto/ecto/pull/4026 ❤️

lib/plausible_web/controllers/stats_controller.ex
aerosol: `mix format` should use our local `.formatter.exs` where we import `:ecto` and where all the macros as listed (https://github.com/elixir-ecto/ecto/blob/master/.formatter.exs)

ukutaht:
Credo has a diff mode so we can at least fail the build when new issues are introduced: https://github.com/rrrene/credo/blob/master/CHANGELOG.md#new-diff-options

This just ensures no new issues are introduced in a PR but it won't force us to fix any of the existing ones. Better than nothing but it won't force us to fix old problems along the way.

I'm not sure how to get it to do a full run on only the changed files. I tried the following without success:
```bash
$ mix credo suggest --files-included $(git diff --name-only master HEAD)
```

ukutaht:
```
➜  analytics git:(fix-dialyzer) mix credo suggest --files-included $(git diff --name-only master HEAD)
dependency :hut is using Rebar 2, which is no longer maintained and no longer works in recent Erlang/OTP versions. Remove the :manager option or set it to :rebar3 instead
info: Some source files could not be parsed correctly and are excluded:

   1) .github/workflows/elixir.yml

No files found!

Please report incorrect results: https://github.com/rrrene/credo/issues

Analysis took 0.05 seconds (0.04s to load, 0.01s running 52 checks on 0 files)
0 mods/funs, found no issues.

Showing priority issues: ↑ ↗ →  (use `mix credo explain` to explain issues, `mix credo --help` for options).
```

vinibrsl (approved):

vinibrsl:
I added two commits that runs Credo against Git changes only. CI is green and happy :)

.github/workflows/elixir.yml
```diff
@@ -66,11 +68,11 @@ jobs:
         run: mix compile --warnings-as-errors --all-warnings
       - name: Generate tracker files
         run: npm run deploy --prefix ./tracker
+      - name: Check Credo Warnings
+        run: mix credo diff --from-git-merge-base origin/master
```
aerosol, line 72: is this expected to halt the CI with error?

.github/workflows/elixir.yml
```diff
@@ -40,7 +40,9 @@ jobs:
             --health-retries 5
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1
```
aerosol, line 43: Why has this version had to change?

.github/workflows/elixir.yml
vinibrsl: It halts CI with error with the exit code.

.github/workflows/elixir.yml
vinibrsl: Previous commits were using Credo piping through some Git magic to list changed files. I changed the Credo command to `mix credo diff --from-git-merge-base origin/master` which does a different Git magic that does not need `checkout@v1`. I also updated actions/checkout to the latest version.

PR #2317, opened 2022-10-07
OpenTelemetry (OTEL) Implementation

vinibrsl:
### Changes

This pull request improves the current OpenTelemetry implementation. Currently only 1% of the spans are sent, due to the high volume of ingestion requests to `/api/event`. I enabled the 1% sampling to `/api/event` only, recording 100% of the other traces.


lib/plausible_web/endpoint.ex
```diff
@@ -31,7 +31,6 @@ defmodule PlausibleWeb.Endpoint do
 
   plug Plug.RequestId
   plug PromEx.Plug, prom_ex_module: Plausible.PromEx
-  plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
```
ukutaht, line 34: Doesn't this also remove Phoenix stats from Promex?

We still want /api/event to be monitored with Prometheus/Grafana.

vinibrsl: Nice catch! I tested locally and it removes /api/event stats from PromEx indeed. [OTEL supports](https://hexdocs.pm/opentelemetry/readme.html#samplers) passing a custom sampling module that dynamically decides whether to count a span or not. I'll test if that works.

ukutaht: Seems like a good approach!

vinibrsl: It worked! [76e00af](https://github.com/plausible/analytics/pull/2317/commits/76e00af3a81f11a8ed729d669d5ddbe100cc2ac8)

lib/plausible/open_telemetry.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Plausible.OpenTelemetry.Sampler do
+  @moduledoc """
+  [Custom OpenTelemetry sampler](https://hexdocs.pm/opentelemetry/readme.html#samplers)
+  implementation that ignores `/api/event` requests.
```
ukutaht, line 4: Hmm.... i wonder now... could this be used to set sampling at 1% for /api/event and 100% for the rest of the endpoints? Maybe worth giving it a shot, I thought it was more complicated and needed a change in the OTEL library :)

ukutaht: The issue with sampling was that Honeycomb was not aware of it and didn't inflate the stats by 100x on their end which I would expect. Honeycomb does say it's sampling-aware so I wonder if there's something missing in the implementation for that.

No need to spend a ton of time on this but maybe just something to look at. Having sampled metrics for /api/event would be nice to have, especially because Honeycomb is good at high-cardinality data. That means we could graph the top 10 domains sending data to /api/event for example, which would be really useful during incidents :)

ukutaht (commented):
Nice! Excited for this :)

Can we add some useful tags to the spans? For better o11y I would add:
1. The hostname of the app (app01, app02, etc)
2. Current release tag (also should add this to Sentry)
3. Current user (if logged-in)
4. Domain/site id (if relevant to the endpoint)

I think all these would also be really useful to add to Sentry tracking.

For 3 and 4, we have Plugs that authorize users / sites so that's the natural place to add this information.

lib/plausible/open_telemetry.ex
vinibrsl: Good call! I updated the custom sampler to record 100% of traces but sample `/api/event` at 1%. I did that by forwarding `/api/event` traces to the native sampler `:otel_sampler_trace_id_ratio_based`, which should take care of informing Honeycomb of the sampling rate therefore inflating the numbers.

vinibrsl:
> Nice! Excited for this :)
> 
> Can we add some useful tags to the spans? For better o11y I would add:
> 
> 1. The hostname of the app (app01, app02, etc)
> 2. Current release tag
> 3. Current user (if logged-in)
> 4. Domain/site id (if relevant to the endpoint)
> 
> I think all these would also be really useful to add to Sentry tracking. It should already have the current user but not the other tags.
> 
> For 3 and 4, we have Plugs that authorize users / sites so that's the natural place to add this information.

I added the four to both Sentry and OTEL. https://github.com/plausible/analytics/pull/2317/commits/5ff467d2463d2a3f235ba70cbc7552175f2b386f, https://github.com/plausible/analytics/pull/2317/commits/8c93c253ff4701506ee5f4a25f8d01d42c14f41c, https://github.com/plausible/analytics/pull/2317/commits/ae874938498c3a50bfdadc1e8f4a26127b7802cc

ukutaht (commented):
Looking very good.

I think we do want to add `user_id` in every case possible, even if AuthPlug isn't running. We don't run AuthPlug in some cases because the current user account is not necessary for the purpose of the controller action.

There's also a plug `PlausibleWeb.AuthorizeSitesApi` which works similarly to `PlausibleWeb.AuthorizeStatsApi` and it would make sense to add tracing metadata to that plug as well.

Let's make these small improvements and test it out in staging!

config/runtime.exs
```diff
@@ -218,8 +233,8 @@ config :sentry,
   dsn: sentry_dsn,
   environment_name: env,
   included_environments: included_environments,
-  release: app_version,
```
aerosol, line 221: was the intention to remove `APP_VERSION` reference as well? It's still here unused but also in `Plausible.MixProject.project().version`

lib/plausible/open_telemetry.ex
```diff
@@ -0,0 +1,69 @@
...
+      {"plausible.user.email", user.email}
+    ])
+  end
+
+  # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service
+  def resource_attributes(build_metadata) do
+    [
+      {"service.name", "analytics"},
+      {"service.namespace", "plausible"},
+      {"service.instance.id", "app00"},
```
aerosol, line 26: `build_metadata[:host]` ?

lib/plausible/open_telemetry.ex
```diff
@@ -0,0 +1,69 @@
...
+  @behaviour :otel_sampler
+  require OpenTelemetry.Tracer, as: Tracer
+
+  @ratio_sampler :otel_sampler_trace_id_ratio_based.setup(0.01)
+
+  @impl true
+  def setup(_sampler_opts), do: []
+
+  @impl true
+  def description(_sampler_config), do: "#{__MODULE__}"
```
aerosol, line 48:
```suggestion
  def description(_sampler_config), do: inspect(__MODULE__)
```

unless we want to keep the `Elixir.` prefix

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -7,24 +7,22 @@ defmodule PlausibleWeb.AuthPlug do
...
-        if user do
-          Sentry.Context.set_user_context(%{id: user.id, name: user.name, email: user.email})
-          assign(conn, :current_user, user)
-        else
-          conn
-        end
-    end
+    Plausible.Auth.User
+    |> Repo.get(id)
+    |> Repo.preload(subscription: last_subscription_q)
```
aerosol, line 26: unrelated to this PR, but we should change this to an inner join so only one query is executed

lib/plausible/open_telemetry.ex
```diff
@@ -0,0 +1,69 @@
...
+      {"service.instance.id", "app00"},
+      {"service.version", build_metadata[:version]}
+    ]
+  end
+end
+
+defmodule Plausible.OpenTelemetry.Sampler do
+  @moduledoc """
+  [Custom OpenTelemetry sampler](https://hexdocs.pm/opentelemetry/readme.html#samplers)
+  implementation that samples 1% of the `/api/event` traces, but records 100% of
```
aerosol, line 35: Why is 1% used, and not e.g. 10%? Why is 1% any good?

aerosol (approved):
Looks good, minor questions/suggestions inline

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -7,24 +7,22 @@ defmodule PlausibleWeb.AuthPlug do
...
+      from(s in Plausible.Billing.Subscription, order_by: [desc: s.inserted_at])
 
-        if user do
-          Sentry.Context.set_user_context(%{id: user.id, name: user.name, email: user.email})
-          assign(conn, :current_user, user)
-        else
-          conn
-        end
-    end
+    Plausible.Auth.User
```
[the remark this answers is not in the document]
aerosol: ^ that's still two, unless you join subscriptions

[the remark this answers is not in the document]
aerosol: image

lib/plausible/open_telemetry.ex
ukutaht: Build metadata is compiled into the docker image so it's identical in all the running hosts. It's created at build-time. I'm not sure how to get the host information into the container but it definitely needs to be at runtime.

@cnkk we want to add the hostname of the app (e.g. `app01.srv.plausible.io`, `app02.srv.plausible.io` etc) into our opentelemetry traces. How can this be accessed in the running docker container?

lib/plausible/open_telemetry.ex
ukutaht: Pricing.

We get about 2b metrics submissions per month, each generating multiple spans in Honeycomb. Sending everything would cost about the same as the rest of our hosting costs combined. Our current plan allows us to send 100m events per month for $150/m. We can increase it to 500m and pay more for it, but I would need some evidence that our usage of Honeycomb is actually providing some value to the team. At the moment it's not.

1% is not a hard decision. We can start using Honeycomb properly and if there's headroom on our plan, or we think it's worth upgrading the plan to send more samples, then I'd be happy to go for it. The value of tracing on `/api/event` is questionable anyways and not even planned for this chunk of work but we decided to give it a go to see if we get something useful out of it.

The main thing that would be useful during incidents is seeing is top 10 domains sending us traffic to `/api/event` at a given time. We're giving it a go to have something rather than nothing. But in the long term this can also be implemented using metrics (rather than traces) which would use much less volume and $$$.

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -41,6 +41,8 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
         end
 
       if role in allowed_roles do
+        Sentry.Context.set_extra_context(%{site_id: site.id, domain: site.domain})
+        Plausible.OpenTelemetry.add_site_attributes(site)
```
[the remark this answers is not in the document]
vinibrsl: Nice catch! I added user_id to `AuthorizeSiteAccess` too.

lib/plausible/open_telemetry.ex
vinibrsl: Cenk added the `APP_HOST` environment variable. I updated

lib/plausible/open_telemetry.ex
```diff
@@ -0,0 +1,87 @@
+defmodule Plausible.OpenTelemetry do
+  @moduledoc false
+
+  require OpenTelemetry.Tracer, as: Tracer
+
+  def add_site_attributes(site) do
+    case site do
+      %Plausible.Site{} = user ->
```
aerosol, line 8:
```suggestion
      %Plausible.Site{} ->
```

PS: we should make sure those warnings make PRs red

ukutaht (approved):

ukutaht:
I created a little dashboard in HoneyComb for staging: https://ui.honeycomb.io/plausible/board/y9peu6J7Msq/Staging-Stats-dashboard. Pretty cool. Won't deploy to prod on a Friday but excited to get this deployed next week :)

PR #2333, opened 2022-10-13
Add tagged-events script extension

RobertJoonas:
Reviewing commit by commit suggested.

### Changes

This PR adds a new script extension that allows tracking interactions with specific HTML elements on a website. For example - to track link clicks on one specific `<a>` element, you can tag it like this:

```html
<a href=... class="plausible-event-name=<your_event_name>">
```

And you can also tag the link with custom property names and values:

```html
<a href=... class="plausible-event-name=<your_event_name> plausible-event-<your_custom_prop>=<your_value>">
```

Tagging a link as above will send a custom event with the given name and props, if a `click` or `auxclick` browser event happens, and targets the link element.

The tracking behavior is somewhat different based on the HTML element type:  

- `<a>` 
  - triggers on `click` and `auxclick` events
  - intercepts navigation based on the same rules as `outbound-links` and `file-downloads`
- `<form>`
  - triggers on `submit` event
  - always intercepts navigation (calls `form.submit()` after preventing default and sending the Plausible event)
- other (``, `<button>`, `<span>`, `<div>`, `<h2>`, etc ...)
  - triggers on `click` and `auxclick` events
  - does not prevent default to intercept possible navigation. Simply calls Plausible with the event name and props read from the element class list.

### Tests

### Changelog

### Documentation
- [x] Documentation update is pending: https://github.com/plausible/docs/pull/295

### Dark mode


tracker/src/plausible.js
```diff
@@ -242,4 +249,72 @@
...
+
+    classList.forEach(function (className) {
+      // Only for link elements: url is a special prop with the dynamic link.href value
+      if (htmlElement.tagName.toLowerCase() === 'a' && className === 'plausible-event-url') {
+        eventAttrs.props.url = htmlElement.href
+        return
+      }
+
+      var matchList = className.match(/plausible-event-(.+)=(.+)/)
+      if (!matchList) { return }
```
vinibrsl, line 270: Maybe it's worth checking that `matchList` has the two required elements, key and value.

tracker/src/plausible.js
```diff
@@ -242,4 +249,72 @@
...
+    var formSubmitted = false
+
+    function submitForm() {
+      if (!formSubmitted) {
+        formSubmitted = true
+        form.submit()
+      }
+    }
+
+    setTimeout(submitForm, 5000)
```
vinibrsl, line 300: Are we waiting for 5 seconds before submitting tagged forms? Why is that?

tracker/src/plausible.js
```diff
@@ -242,4 +249,72 @@
...
+    plausible(eventAttrs.name, { props: eventAttrs.props, callback: submitForm })
+  }
+
+  function handleOtherElementClickEvent(event) {
+    if (event.type === 'auxclick' && event.button !== MIDDLE_MOUSE_BUTTON) { return }
+
+    var tagName = event.target.tagName.toLowerCase()
+    if (tagName === 'form' || tagName === 'a') { return }
```
vinibrsl, line 308: I'd suggest having a constant up top defining those special tags, e.g.

```
var SPECIAL_TAGS = ['form', 'a']

// ...

if (SPECIAL_TAGS.include(tagName) { return }
```

This makes clear what `form` and `a` are and why they are skipped from this function.

tracker/test/tagged-events.spec.js
```diff
@@ -0,0 +1,36 @@
+const { test } = require('./support/harness')
+const { mockRequest, mockManyRequests, isMac, expectCustomEvent } = require('./support/test-utils')
+const { expect } = require('@playwright/test')
+
+test.describe('tagged-events extension', () => {
```
vinibrsl, line 5: Question: with `tagged-events` should `a` elements without tags be still tracked?

vinibrsl (commented):
Looks good overall. Thanks for the small commits, @RobertJoonas!

tracker/src/plausible.js
RobertJoonas: MDN Web Docs says [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value) about `String.prototype.match()` return value:

> An Array whose contents depend on the presence or absence of the global (g) flag, or null if no matches are found.
[ ... ]
If the g flag is not used, only the first complete match and its related capturing groups are returned.

My thinking was that the `!matchList` check should be enough because if the match function returns an array (not null), it has to be of length 3 (first complete match + two captured groups). Played around with this in a JS console and it worked consistently.

tracker/src/plausible.js
RobertJoonas: `submitForm` is also called right after plausible finishes sending the event request (`callback` argument).

`plausible(eventAttrs.name, { props: eventAttrs.props, callback: submitForm })`

If for some reason Plausible fails to send the event, it falls back to submitting the form after 5 seconds.

tracker/test/tagged-events.spec.js
RobertJoonas: Nope. They will only be tracked if they have a `plausible-event-name=<name>` class. Also, if there are classes for custom props `plausible-event-<prop>=<value>` but no event name, they're not tracked, because a custom event needs to have a name.

vinibrsl: Thanks for clarifying! In that case I suggest adding a test asserting events are not sent and the user is redirected to the correct URL when clicking an `a` element without our special class.

tracker/src/plausible.js
RobertJoonas: Good point! I've now done this in https://github.com/plausible/analytics/pull/2333/commits/871ac10471bfe7b8d00f10a51af9a01eadc221ac

However, I used `Array.some` instead of `Array.includes` because it's supported in IE. Doesn't look very nice, but I tried my best, wrapping the long line of code in a helper function.

vinibrsl (approved):
Looks good! We should add some public documentation and guides, and we're good to go!

RobertJoonas:
@vinibrsl could you please review the new commits I made (starting from https://github.com/plausible/analytics/pull/2333/commits/470e3fcdff4c86a2d7aae67756fa6ae11a76b425)?

Here are the two main problems solved with the recent commits:

- Ability to track HTML elements when their children are clicked (links already had that ability)
  - add eventListeners directly on the tagged elements (instead of `document`), this gives us access to the tagged element itself via the `event.curretTarget` property even when a child of this element is actually clicked (i.e. the child is the `event.target`)
- make tagged-events IE 11 compatible
  - I tested it on IE 11 with BrowserStack Live and as it turned out, `classList.forEach` was not compatible. After changing it to a simple `for i` loop, IE was able to send the event request nicely with the correct name and props.

With these changes, it should be ready for testing on staging.

RobertJoonas:
UPDATE: Just pushed a [new commit](https://github.com/plausible/analytics/pull/2333/commits/93843c13ebee78872e553f57f13cda0c77256aab) that adds a test in `custom-event-edge-cases.spec.js`. This is necessary to fix the priority between outbound / download / tagged events and make sure only one event is fired every time.

vinibrsl (approved):
Looks good!

RobertJoonas:
**UDPATE:** The [latest commit](https://github.com/plausible/analytics/pull/2333/commits/a42f87df101d6f1dcfb7a975ee9567babff6382f) adds the ability to track forms that have a direct tagged parent (the same case as with links). Added a test for this as well, and also covered the "form submit with enter key" case that was not tested before.

I have been manually testing all the edge cases I could think of, in these environments:

- Firefox (latest) on Ubuntu
- IE 11 on Windows 10
- Opera 88.0 on Windows 10
- Safari 12.1 on macOS Mojave
- Chrome 100 on macOS Mojave

Everything worked perfectly, except for tracking clicks on `<svg>` elements on IE 11 ([doesn't support `HTMLElement.classList` on svg-s](https://caniuse.com/?search=classList)) which I think is not a concern.

So I think this is now ready to go live, but I'd like for @ukutaht to review the decisions made regarding this feature. For the best overview of what tagged-events will support, I'd suggest reading through the docs I wrote for this: https://github.com/plausible/docs/pull/295. Thanks!

tracker/src/plausible.js
```diff
@@ -242,4 +250,104 @@
...
+    var classList = element && element.classList
+    if (classList) {
+      for (var i = 0; i < classList.length; i++) {
+        if (classList.item(i).match(/plausible-event-name=(.+)/)) { return true }
+      }
+    }
+    return false
+  }
+
+  document.addEventListener('submit', handleFormSubmitEvent)
```
ukutaht, line 341: There seems to be an inconsistency here.

The event listener for forms is attached to the top-level document. But the event listener for clicks are attached to the actual elements themselves. 

Why does this matter?

Imagine the customer has some javascript on their page that modifies the HTML dynamically after the content is loaded. If they dynamically add a form, it will get tracked because all events bubble up to the document root. However, if they dynamically add a new link or button, it will not be tracked because we only attach listeners once when the content is loaded.

Maybe try testing this on our React dashboard? SPA frameworks like React constantly modify the DOM. The tagged-events script should work everywhere, including modal dialogs which are rendered dynamically after content is loaded.

Is there any problem with registering the 'click' and 'auxclick' events to the document root as well? It would dynamically catch every possible click, even if the DOM has changed after the initial load.

tracker/src/plausible.js
```diff
@@ -242,4 +250,104 @@
     })
   }
   {{/if}}
+
+  {{#if tagged_events}}
```
ukutaht, line 254: This file is starting to grow into unmanageable territory.

@vinibrsl how much do you know about JS compilation? I wonder if we could extract some of these supporting functions into a separate file and `import()` them conditionally to be only compiled for the specific script extensions where they're needed.

ukutaht (commented):
LGTM overall. I love that we're testing everything here now.

Had some comments, let me know what you think

tracker/src/plausible.js
vinibrsl: I agree, however, including a JS file in a JS file was introduced in ES6. I don't think we can require multiple files while maintaining compatibility with other browsers.

One solution would be to use Handlebars partials, and split the code in multiple files.

```
{{#if tagged_events}}
{{> taggedEvents }}
{{/if}}
```

tracker/src/plausible.js
RobertJoonas: Good point about the dynamic DOM changes. I just tested it on the React dashboard, and for clicks, it doesn't work like that indeed. I can see two options to tackle this issue:

1) switch back to listening for click events on the document level

This would mean that for every click on any element, we'd have to bubble up and search for a tagged parent element. This would be quite expensive to do, and I think in this case we should define a limit to how far up we bubble. For example, with code like this:

```html
<div class="plausible-event-name=CustomEvent">
    <div>
        <div>
            <button> I'm clicked </button>
        </div>
    </div>
</div>
```

we'll be able to capture the event, if we bubble up at least 3 levels.

2) re-add the event listeners every time DOM is updated (do something like this: https://stackoverflow.com/questions/3219758/detect-changes-in-the-dom). That would avoid the bubbling phase on every click, but for links with tagged parents we'll still have to bubble.

vinibrsl: The first one seems a good solution, as we already bubble up in other cases.

tracker/src/plausible.js
ukutaht: Ah yes, we don't actually compile ES6 here. Handlebars partial is a great solution!

tracker/src/plausible.js
ukutaht: We had a quick discussion about this with Robert.

The first option is 'quick and dirty' and should work for most cases. I think the second option is much more robust but harder to implement.

Let's try the first option with some live customers before announcing this feature publicly. If we don't see any issues, we can move forward with it. If the 3-node limit turns out to be a problem, we can improve this by using a `MutationObserver` and attaching the listeners directly to the relevant nodes.

tracker/src/plausible.js
```diff
@@ -186,10 +186,10 @@
     var hrefWithoutQuery = link && link.href && link.href.split('?')[0]
 
     {{#if tagged_events}}
-    var eventAttrs = getTaggedEventAttributes(link)
-    if (eventAttrs.name) {
-      eventAttrs.props.url = link.href
-      return sendLinkClickEvent(event, link, eventAttrs)
+    if (isElementOrParentTagged(link, 0)) {
```
RobertJoonas: This doesn't send any events here anymore, but we still need it to know when to ignore outbound links or file downloads.

tracker/src/plausible.js
```diff
@@ -300,32 +300,40 @@
...
 
   function isForm(element) {
-    return element.tagName.toLowerCase() === 'form'
+    return element && element.tagName && element.tagName.toLowerCase() === 'form'
   }
 
-  function handleOtherElementClickEvent(event) {
+  var PARENTS_TO_SEARCH_LIMIT = 3
+
+  function handleTaggedElementClickEvent(event) {
```
RobertJoonas: aA opposed to before, this now handles link clicks too.

tracker/compile.js
```diff
@@ -8,6 +8,8 @@ Handlebars.registerHelper('any', function (...args) {
   return args.slice(0, -1).some(Boolean)
 })
 
+Handlebars.registerPartial('customEvents', Handlebars.compile(fs.readFileSync(relPath('src/customEvents.js')).toString()))
```
RobertJoonas, line 11: I agree that it would look cleaner to split into multiple files, but the Handlebars `if` clauses are nested inside and outside of functions, and in multiple places. So having every extension code in another file is not feasible. Moreover, I imagine it would make it pretty hard to find function references/declarations (you'd have to jump around in multiple files).

I only moved custom event tracking into a separate file atm, because all those functions make up a very different chunk of code than the rest of the plausible script. I **didn't** break it down into separate custom event extension files, because there are many shared functions.

However, we could also separate the `exclusions` and `dimensions` extensions. Those are only injected into one place. The `local` extension is also injected in one place only, but it's two lines of code. Let me know if you think these should be in separate files too.

tracker/src/plausible.js
```diff
@@ -118,41 +117,41 @@
   }
 
   {{#unless manual}}
```
RobertJoonas, line 119: Everything in this 'unless' block just moved back one indent. No other changes here.

RobertJoonas: Outdated - just deleted that commit. Let's do the linting in a separate PR.

ukutaht (approved):

ukutaht:
The partial looks good to me

PR #2334, opened 2022-10-13
Remove inserts per second Google Analytics test assertion 

vinibrsl:
This commit removes a flaky test assertion from the GA suite. The initial purpose of this test was to ensure we are not inserting to Clickhouse more than 1 time per second during an import.

Although it is nice to have an assertion like this, the test is flaky and fails sometimes. To get the total inserts during a time range, it queries Clickhouse internal tables for insert commands, and this number is not deterministic.

In a real-world scenario with Clickhouse, the app runs in multiple nodes not aware of their neighbors' CH calls. In that case, Grafana provides us visibility over CH, and these tests do not suffice.

### DIFF
--- a/test/plausible/google/vcr_test.exs
+++ b/test/plausible/google/vcr_test.exs
@@ -6,21 +6,8 @@ defmodule Plausible.Google.Api.VCRTest do
 
   setup [:create_user, :create_site]
 
-  defp get_insert_count do
-    Plausible.ClickhouseRepo.aggregate(
-      from(ql in "query_log",
-        prefix: "system",
-        where: ql.query_kind == "Insert" and ql.is_initial_query == true
-      ),
-      :count
-    )
-  end
-
   test "imports page views from Google Analytics", %{site: site} do
     use_cassette "google_analytics_import#1", match_requests_on: [:request_body] do
-      inserts_before_importing = get_insert_count()
-      before_importing_timestamp = DateTime.utc_now()
-
       view_id = "54297898"
       date_range = Date.range(~D[2011-01-01], ~D[2022-07-19])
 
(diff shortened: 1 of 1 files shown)


aerosol (approved):

ukutaht:
> In a real-world scenario with Clickhouse, the app runs in multiple nodes not aware of their neighbors' CH calls. In that case, Grafana provides us visibility over CH, and these tests do not suffice.

The initial reason for this test was to ensure that we don't send too many insert statements to Clickhouse which results in `TOO_MANY_PARTS` error in Clickhouse.

I agree that this should be solved with infrastructure monitoring rather than a unit test.

Currently we have alerting in place for when we run into the `TOO_MANY_PARTS` error.  But at the moment this is fired too late, when we actually see the error. At that point we're already losing inserts which happened during this summer.

So the question is how do we monitor this specific aspect of Clickhouse. How do we know if we're within the acceptable range of inserts sent / parts created? How do we know we're close to hitting the `TOO_MANY_PARTS` error? Some help from Altinity would be needed here.

vinibrsl:
> > In a real-world scenario with Clickhouse, the app runs in multiple nodes not aware of their neighbors' CH calls. In that case, Grafana provides us visibility over CH, and these tests do not suffice.
> 
> The initial reason for this test was to ensure that we don't send too many insert statements to Clickhouse which results in `TOO_MANY_PARTS` error in Clickhouse.
> 
> I agree that this should be solved with infrastructure monitoring rather than a unit test.
> 
> Currently we have alerting in place for when we run into the `TOO_MANY_PARTS` error. But at the moment this is fired too late, when we actually see the error. At that point we're already losing inserts which happened during this summer.
> 
> So the question is how do we monitor this specific aspect of Clickhouse. How do we know if we're within the acceptable range of inserts sent / parts created? How do we know we're close to hitting the `TOO_MANY_PARTS` error? Some help from Altinity would be needed here.

Thanks to @cnkk, our Clickhouse Performance Monitor dashboard on Grafana now has an inserts per second query. In a healthy state, it averages below 1,5 inserts per second. The last time we saw `TOO_MANY_PARTS` error, this number was above 7 inserts per second:

image

I added an alert on Grafana when greater than 2 inserts/sec so we can prepare before a too many parts error.

PR #2353, opened 2022-10-18
Double maximum header length

vinibrsl:
### Changes

This commit makes the permitted header length more permissive, 8,192 bytes, doubling the Phoenix default.

Related to https://github.com/4lejandrito/next-plausible/issues/67

### Tests
I tried testing with ExUnit, but I could not reproduce the HTTP 431 Request Header Fields Too Large response. I guess this is because the HTTP server does not actually run in test environment. However, I tested this locally with curl and the HTTP 431 is fixed.

### Changelog

### Documentation

### Dark mode


aerosol (approved):
Let's revisit this once we start dropping unwanted headers earlier in the pipeline and/or nextjs gets this fixed.

PR #2354, opened 2022-10-18
Parse Google Analytics failed response before reporting to Sentry

vinibrsl:
This commit parses failed GA responses to JSON before reporting to Sentry. This makes Sentry's PII filtering smarter, redacting only specific keys from the response, instead of the whole string.


aerosol (approved):
note to myself: I'll have to incorporate this in https://github.com/plausible/analytics/pull/2358

PR #2356, opened 2022-10-19
Update Makefile with tracker generation

aerosol:
### Changes

This PR updates Makefile with tracker generation and also makes the CI use it.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -59,13 +59,11 @@ jobs:
           key: ${{ runner.os }}-mix-v2-${{ hashFiles('**/mix.lock') }}
           restore-keys: ${{ runner.os }}-mix-v2-
       - name: Install dependencies
-        run: mix deps.get && npm install --prefix ./tracker
+        run: make install
       - name: Check Formatting
         run: mix format --check-formatted
       - name: Check Compile Warnings
         run: mix compile --warnings-as-errors --all-warnings
-      - name: Generate tracker files
-        run: npm run deploy --prefix ./tracker
       - name: Run tests
         run: mix test
       #- name: Check Dialyzer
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -10,6 +10,7 @@ install: ## Run the initial setup
 	mix download_country_database
 	npm install --prefix assets
 	npm install --prefix tracker
+	npm run deploy --prefix tracker
```
RobertJoonas, line 13: We initially agreed with Vini that before running tests locally, the tracker files will have to be generated manually once. But this change sounds good, as it will just automate it.

.github/workflows/elixir.yml
```diff
@@ -59,13 +59,11 @@ jobs:
           key: ${{ runner.os }}-mix-v2-${{ hashFiles('**/mix.lock') }}
           restore-keys: ${{ runner.os }}-mix-v2-
       - name: Install dependencies
-        run: mix deps.get && npm install --prefix ./tracker
+        run: make install
```
RobertJoonas, line 62: Looks clean, but I think in Elixir CI it's fine without these two commands:

- mix download_country_database
- npm install --prefix assets

Also, creating and migrating databases is done in the `mix test` alias. Is there a reason for this change other than making it prettier?

aerosol: Good point about redundancy of DB provisioning, I wanted to couple `make install` so the CI ensures the environment is ready to use after invoking it. I can either revert this change, or make a new [alias](https://github.com/plausible/analytics/blob/make-the-ci-use-makefile/mix.exs#L123) (such as `test.ci`) that will skip DB creation/migration steps. Any preference?

aerosol: I'm not sure what's better - running the two steps you mentioned somewhat makes it part of the pipeline, which is good - we'll know quickly if they ever break, but then it's done at the expense of CI time...

aerosol: On the other hand it's unclear to me which step used to generate tracker files on the CI. It passed after all, so `priv/` had to be populated already by something?

aerosol: reverting the CI change with f4001b7

vinibrsl (approved):

PR #2357, opened 2022-10-19
improve first launch experience for self-hosters

ruslandoga:
### Changes

This PR removes admin creds set in env vars and replaces them with web ui setup on first launch.

- [ ] Maybe log invitation link instead of redirecting all requests to `/register`? That would provide a bit more protection if the admin goes to sleep between launching the app and opening it in the browser.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/application.ex
```diff
@@ -30,7 +30,12 @@ defmodule Plausible.Application do
     setup_sentry()
     setup_opentelemetry()
     Location.load_all()
-    Supervisor.start_link(children, opts)
+
+    with {:ok, _pid} = started <- Supervisor.start_link(children, opts) do
+      # TODO maybe do this after Plausible.Repo is started and before endpoint
+      Plausible.Release.set_first_launch()
```
ruslandoga, line 36: - [ ] maybe do this after Plausible.Repo is started and before endpoint, this would solve a race-condition when `first_launch` in `nil`

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -24,52 +24,61 @@ defmodule PlausibleWeb.AuthController do
...
-        layout: {PlausibleWeb.LayoutView, "focus.html"}
-      )
+  defp maybe_disable_registration(conn, _opts) do
+    selfhost_config = Application.get_env(:plausible, :selfhost)
+    disable_registration = Keyword.fetch!(selfhost_config, :disable_registration)
+    first_launch? = Release.first_launch?()
+
+    cond do
+      first_launch? ->
+        conn
```
ruslandoga, line 36: when first_launch?, it still shows the registration forms, even with registration disabled, otherwise the user won't be able to get into the system

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -24,52 +24,61 @@ defmodule PlausibleWeb.AuthController do
...
-      else
-        render(conn, "register_form.html",
-          changeset: user,
-          captcha_error: "Please complete the captcha to register",
-          layout: {PlausibleWeb.LayoutView, "focus.html"}
-        )
+    if PlausibleWeb.Captcha.verify(params["h-captcha-response"]) do
+      case Repo.insert(user) do
+        {:ok, user} ->
+          if Release.first_launch?(), do: Release.set_first_launch(false)
```
ruslandoga, line 62: once registered, first_launch? is set to false

- [ ] forgot to test that

test/plausible_web/controllers/admin_auth_controller_test.exs
```diff
@@ -8,6 +18,7 @@ defmodule PlausibleWeb.AdminAuthControllerTest do
       assert redirected_to(conn) == "/login"
     end
 
+    @tag skip: true
     test "logs admin user in automatically when authentication is disabled", %{conn: conn} do
```
ruslandoga, line 22: not sure what to do about this test and `DISABLE_AUTH` in general

rel/overlays/init-admin.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/sh
-# Create an admin user
-
-BIN_DIR=$(dirname "$0")
-
-"${BIN_DIR}"/bin/plausible eval Plausible.Release.init_admin
```
ruslandoga, line 6: or should these be kept but made no-op?

test/plausible_web/controllers/admin_auth_controller_test.exs
ruslandoga: I'm keeping it as before to avoid breaking anyones setup

rel/overlays/init-admin.sh
ruslandoga: I restored this script and everything else related to `init_admin`

lib/plausible/application.ex
```diff
@@ -9,6 +9,7 @@ defmodule Plausible.Application do
     children = [
       Plausible.Repo,
       Plausible.ClickhouseRepo,
+      {Plausible.Release.BlockingChild, fn -> Plausible.Release.set_first_launch() end},
```
aerosol, line 12: At this point you can already access application env so there's no point in running this code in a non-self-hosted setup -- you can dynamically generate the list of children started. Also the GenServer for this seems to be an overkill, a regular function should block it just fine, no? 
My concern is this is prone to race conditions, you have no guarantees the Repo is up by this point, I think it starts asynchronously?

ruslandoga: Where would the function call be placed?

ruslandoga: I've added one more commit, but I think generating the children dynamically it's a worse approach since it splits the logic and executes the setup code less often in tests.

lib/plausible/application.ex
```diff
@@ -9,6 +9,9 @@ defmodule Plausible.Application do
     children = [
       Plausible.Repo,
       Plausible.ClickhouseRepo,
+      {Plausible.Release.BlockingChild,
+       fn -> Plausible.Release.wait_for_repo(Plausible.Repo) end},
```
ruslandoga, line 13: waiting for repo unconditionally as it is required for `Plausible.Session.Salts` as well

lib/plausible/application.ex
aerosol: Let me have a proper look on Monday and get back to you

lib/plausible_release.ex
```diff
@@ -7,6 +7,52 @@ defmodule Plausible.Release do
     :ecto
   ]
 
+  def self_hosted? do
+    Application.fetch_env!(@app, :is_selfhost)
```
aerosol, line 11: Do you know what is setting the env var for me?

```
$ SELFHOST=true iex -S mix
...
iex(1)> System.get_env "SELFHOST"
"false"
```

ruslandoga: Probably `config/.env.dev`

aerosol: sneaky, thanks :pray:

test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -111,6 +111,34 @@ defmodule PlausibleWeb.AuthControllerTest do
 
       assert html_response(conn, 200) =~ "Please complete the captcha"
     end
+
+    test "sets first_launch? to false", %{conn: conn} do
+      alias Plausible.Release
+
+      prev_env = Application.get_all_env(:plausible)
+      on_exit(fn -> Application.put_all_env(plausible: prev_env) end)
```
aerosol, line 119: we have now helpers to do that: https://github.com/ruslandoga/analytics/blob/first-launch/test/support/test_utils.ex#L12-L31

test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -111,6 +111,34 @@ defmodule PlausibleWeb.AuthControllerTest do
...
+
+      :ok = Application.put_env(:plausible, :is_selfhost, true)
+      :ok = Release.set_first_launch()
+      true = Release.first_launch?()
```
aerosol, line 123:
```suggestion
      assert Release.first_launch?()
```

test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -111,6 +111,34 @@ defmodule PlausibleWeb.AuthControllerTest do
...
+            email: "user@example.com",
+            password: "very-secret",
+            password_confirmation: "very-secret"
+          }
+        )
+
+      assert conn.status == 302
+
+      assert Release.should_be_first_launch?() == false
+      assert Release.first_launch?() == false
```
aerosol, line 140:
```suggestion
      refute Release.should_be_first_launch?()
      refute Release.first_launch?()
```

lib/plausible/application.ex
```diff
@@ -9,6 +9,9 @@ defmodule Plausible.Application do
...
+      maybe_set_first_launch(),
```
aerosol, line 14: Those two blocking children are in conflict now, you need IDs to run them. However as mentioned elsewhere, you don't need a GenServer that shuts down immediately for this, child specs can be built like this:

```elixir 
%{id: Ignored.WaitForRepo, start: {__MODULE__, :wait_for_repo, []}} # returns :ignore,
%{id: Ignored.MaybeSetFirstLaunch, start: {Release, :maybe_set_first_launch, []} # returns :ignore
```
Everything happens in the application startup process and children are started sequentially. Since you wait for the Repo anyway, there's no need for any wrappers. At the end `Enum.reject` all that returned `:ignore`.

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -24,52 +24,61 @@ defmodule PlausibleWeb.AuthController do
...
+          conn = set_user_session(conn, user)
+
+          case user.email_verified do
```
aerosol, line 65: `if`?

lib/plausible/application.ex
```diff
@@ -116,4 +120,15 @@ defmodule Plausible.Application do
     OpentelemetryEcto.setup([:plausible, :clickhouse_repo])
     OpentelemetryOban.setup()
   end
+
+  defp maybe_set_first_launch do
+    alias Plausible.Release
```
aerosol, line 125: let's keep aliases at the top

aerosol (changes requested):
Some comments inline. 

In general I'd consider simplifying it further, without introducing that much application env fiddling -- so that the plug kicks in only in self-hosted environment and the first launch is determined by checking if any User exists, on each self-hosted browser request. This extra query in the `:browser` pipeline hurts no one I believe in terms of load/performance and gets us much better clarity.

test/plausible_web/controllers/auth_controller_test.exs
ruslandoga: I don't think those helpers would work here since I'm reseting all plausible env to prev values after the test has run. There are several keys that are changed.

test/plausible_web/controllers/auth_controller_test.exs
ruslandoga: That's not really the assertion of this test case.

test/plausible_web/controllers/auth_controller_test.exs
ruslandoga: `refute` passes on `nil`. I want to ensure the value returned is `false`. I can assert on `is_boolean` though.

test/plausible_web/controllers/auth_controller_test.exs
aerosol: Sure ok, see my final comment about not using that many application env conditionals, perhaps it'd help. That comes with an extra bonus of not having to wait for the repo ;)

test/plausible_web/controllers/auth_controller_test.exs
aerosol: still crashes on match error though :shrug:

lib/plausible_web/controllers/auth_controller.ex
ruslandoga: I'd like to keep the changes only relevant to the PR :)

ruslandoga:
> In general I'd consider simplifying it further, without introducing that much application env fiddling -- so that the plug kicks in only in self-hosted environment and the first launch is determined by checking if any User exists, on each self-hosted browser request.

Right, I'll update the PR.

test/plausible_web/controllers/auth_controller_test.exs
ruslandoga: I don't think it's a problem that it can crash since code in `setup` blocks can also crash. But I'll update the PR.

test/plausible_web/controllers/auth_controller_test.exs
ruslandoga: We'd still need to wait for the repo for Salts genserver.

test/plausible_web/plugs/first_launch_plug_test.exs
```diff
@@ -0,0 +1,94 @@
+defmodule PlausibleWeb.FirstLaunchPlugTest do
+  use PlausibleWeb.ConnCase
+  import Plug.Test
+
+  alias PlausibleWeb.FirstLaunchPlug
+  alias Plausible.Release
+
+  setup do
```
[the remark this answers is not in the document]
ruslandoga: does `patch_env` handle multiple invocations for the same key?

test/plausible_web/controllers/admin_auth_controller_test.exs
```diff
@@ -34,6 +44,17 @@ defmodule PlausibleWeb.AdminAuthControllerTest do
       conn = get(conn, "/register")
       assert redirected_to(conn) == "/login"
     end
+
+    test "disable registration + first launch", %{conn: conn} do
+      set_config(disable_registration: true)
```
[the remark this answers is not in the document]
ruslandoga: I updated that function to use `patch_env` underneath. It has more logic that is handled by `patch_env` right now.

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -24,52 +24,63 @@ defmodule PlausibleWeb.AuthController do
...
     end
   end
 
-  def register(conn, params) do
-    if Keyword.fetch!(Application.get_env(:plausible, :selfhost), :disable_registration) != false do
-      redirect(conn, to: Routes.auth_path(conn, :login_form))
-    else
-      user = Plausible.Auth.User.new(params["user"])
+  defp is_selfhost(conn, _opts) do
+    assign(conn, :is_selfhost, Plausible.Release.is_selfhost())
```
aerosol, line 48: What is the purpose of the assign in contrast to fetching from env each time?

aerosol: Ah I see it now, templates

ruslandoga: pure view functions and all that

ruslandoga: I can remove it and use app env in templates. That would be in line with the conventions of this project, I think.

lib/plausible_web/templates/auth/register_form.html.eex
```diff
@@ -1,5 +1,11 @@
 <div class="mx-auto mt-6 text-center dark:text-gray-300">
-  <h1 class="text-3xl font-black">Register your 30-day unlimited-use free trial</h1>
+  <h1 class="text-3xl font-black">
+    <%= if assigns[:is_selfhost] do %>
+    Register your Plausible Analytics account
+    <% else %>
+    Register your 30-day unlimited-use free trial
+    <% end %>
```
ruslandoga, line 7: I'm using invitation template wording here ...

lib/plausible_web/templates/auth/register_form.html.eex
```diff
@@ -53,7 +59,15 @@
...
     <% end %>
 
-    <%= submit "Start my free trial →", class: "button mt-4 w-full" %>
+    <%
+      submit_text =
+        if assigns[:is_selfhost] do
+          "Create my account →"
+        else
+          "Start my free trial →"
+        end
```
ruslandoga, line 68: ... and here

lib/plausible_web/templates/auth/register_form.html.eex
```diff
@@ -1,5 +1,11 @@
 <div class="mx-auto mt-6 text-center dark:text-gray-300">
-  <h1 class="text-3xl font-black">Register your 30-day unlimited-use free trial</h1>
+  <h1 class="text-3xl font-black">
+    <%= if assigns[:is_selfhost] do %>
```
ruslandoga, line 3: I'm using `assigns[:is_selfhost]` to avoid possible problems when this template is rendered outside of auth controller. I checked the project and it doesn't seem there are any places that do this, but still, better be safe :)

lib/plausible_release.ex
```diff
@@ -7,9 +7,13 @@ defmodule Plausible.Release do
...
 
+  @spec is_selfhost :: boolean
+  def is_selfhost do
+    Application.fetch_env!(@app, :is_selfhost)
+  end
+
   def should_be_first_launch? do
-    self_hosted? = Application.fetch_env!(@app, :is_selfhost)
-    self_hosted? and not (_has_users? = Repo.exists?(Plausible.Auth.User))
+    is_selfhost() and not (_has_users? = Repo.exists?(Plausible.Auth.User))
```
ruslandoga, line 16: I moved repo call after `and` to avoid calling it in the cloud edition.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 - Manually lock and unlock enterprise users plausible/analytics#2197
 - ARM64 support for docker images plausible/analytics#2103
 - Add support for international domain names (IDNs) plausible/analytics#2034
+- Allow self-hosted users to register via Web UI
```
ruslandoga, line 42: not sure if it's descriptive enough. Maybe "Allow first time users to register via Web UI in self-hosted version"

lib/plausible_web/templates/auth/register_form.html.eex
aerosol: It's safe to use @is_selfhost and then crash loud if it's missing. IMHO tests should be responsible for ensuring it's there.

CHANGELOG.md
aerosol: "Allow self-hosters to register an account on first launch"?

aerosol (approved):

ukutaht (approved):

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,14 +1,16 @@
...
   def index(conn, _params) do
     if conn.assigns[:current_user] do
       user = conn.assigns[:current_user] |> Repo.preload(:sites)
       render(conn, "sites.html", sites: user.sites)
     else
-      render(conn, "index.html")
+      conn
+      |> put_session(:login_dest, conn.request_path)
+      |> redirect(to: "/login")
+      |> halt()
```
ruslandoga, line 13:
```elixir
conn
|> put_session(:login_dest, conn.request_path)
|> redirect(to: "/login")
|> halt()
```

is copied from https://github.com/plausible/analytics/blob/101e5a68b5ff907978f1f7b220b17d2cf53da4b6/lib/plausible_web/plugs/auto_auth_plug.ex#L19-L22

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,14 +1,16 @@
 defmodule PlausibleWeb.PageController do
   use PlausibleWeb, :controller
   use Plausible.Repo
-  plug PlausibleWeb.AutoAuthPlug
```
ruslandoga, line 4: which used to be "invoked" here

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,14 +1,16 @@
...
 
   def index(conn, _params) do
     if conn.assigns[:current_user] do
       user = conn.assigns[:current_user] |> Repo.preload(:sites)
       render(conn, "sites.html", sites: user.sites)
     else
-      render(conn, "index.html")
```
ruslandoga, line 11: I'm removing `render` since it wasn't called due to `halt` in `AutoAuthPlug` anyway.

ruslandoga: actually, to keep the previous auto auth plug behaviour, we need to redirect all requests to `/login` from this controller action, regardless of whether the user is signed in.

lib/plausible_release.ex
```diff
@@ -9,13 +9,13 @@ defmodule Plausible.Release do
     :ecto
   ]
 
-  @spec is_selfhost :: boolean
-  def is_selfhost do
+  @spec selfhost? :: boolean
+  def selfhost? do
```
ruslandoga, line 13: credo made me do it

ruslandoga:
@ukutaht that should be fixed in https://github.com/plausible/analytics/pull/2357/commits/923682080d5cd6efbaf054c0f2f9799db3cea759

lib/plausible_web/controllers/page_controller.ex
```diff
@@ -1,14 +1,10 @@
 defmodule PlausibleWeb.PageController do
   use PlausibleWeb, :controller
   use Plausible.Repo
-  plug PlausibleWeb.AutoAuthPlug
```
ukutaht, line 4:
```suggestion
@doc """
The root path is never accessible in Plausible.Cloud because it is handled by the upstream reverse proxy.

This controller action is only ever triggered in self-hosted Plausible. It redirects the user to /login where `PlausibleWeb.RequireLoggedOutPlug` plug kicks in. If they are already logged in, they are redirected to /sites, otherwise they'll see the /login page.
"""
```

ukutaht: I think a nice TODO for v1.6.0 would be to create a nice splash page that welcomes the user to their self-hosted Plausible installation and links to the /register and /login pages.

aerosol (approved):

StreetStrider:
I'm a self-hoster and I'm looking for a way to automatically setup admin user and some seeded data during plausible setup in our infra. The setup is (ofc) automatic.
I don't get the ratio behind this change, at some angle it looked like it's removing the ability to create admin user via scripting.
I'm struggling to understand, can someone help me with how do I setup admin user (and possible seed some data) during plain shell script or while docker image baking in Dockerfile?

ruslandoga:
👋 @StreetStrider 

https://github.com/plausible/analytics/discussions/2460#discussioncomment-4318576 might be helpful.

StreetStrider:
@ruslandoga I was able to populate `/app/lib/plausible-0.0.1/priv/repo/seeds.exs` with some debug code (like IO.puts) and run it via `/app/bin/plausible eval "Plausible.Release.seed()"`. I was able to actually run this code.

However, when I put an actual migration code ([like this one](https://github.com/ruslandoga/plausible-seeds-ci/blob/d650b951eed6ae3ec6d056ee0d588d7bf781e6a5/ci/seeds.exs)) in it, I got an error related to environment:
```
#0 0.518 ERROR! Config provider Config.Reader failed with:                                                                                                                                                         
#0 0.519 ** (RuntimeError) BASE_URL configuration option is required. See https://plausible.io/docs/self-hosting-configuration#server 
#0 0.519     (stdlib 3.17.1) erl_eval.erl:685: :erl_eval.do_apply/6                                                                                                                                                
#0 0.519     (elixir 1.13.4) src/elixir.erl:296: :elixir.recur_eval/3                                                                                                                                              
#0 0.519     (elixir 1.13.4) src/elixir.erl:274: :elixir.eval_forms/3
…
```
I've double-checked that process environment is correctly propagated, however, it seemes that runner requires some additional step to pass env to it. I don't know how to fix this, this looks like a plausible/elixir specific issue.

ruslandoga:
Can you please share `plausible-conf.env`, `seeds.exs`, and `docker-compose.yml`?

StreetStrider:
@ruslandoga 
```yaml
  plausible:
    # image: plausible/analytics:latest
    build:
      context: .
    # restart: always
    command: sh -c "sleep 10 && /entrypoint.sh db createdb && /entrypoint.sh db migrate && /entrypoint.sh run"
    depends_on:
      - plausible_db
      - plausible_events_db
    #  - mail
    # ports:
    #   - 8083:8000
    environment:
      SECRET_KEY_BASE: ${SECRET_KEY_BASE:?err}
      BASE_URL: ${BASE_URL:?err}
      # ADMIN_USER_EMAIL: ${ADMIN_USER_EMAIL:?err}
      # ADMIN_USER_NAME: ${ADMIN_USER_NAME:?err}
      # ADMIN_USER_PWD: ${ADMIN_USER_PWD:?err}
```
We retrieve env from dotenv provided by compose. I've commented out `image` to be able to replace it with Dockerfile instead, to run seed. Dockerfile is simple at the moment:
```Dockerfile
FROM plausible/analytics:latest

COPY seed.exs lib/plausible-0.0.1/priv/repo/seeds.exs
```

```
alias Plausible.Repo
alias Plausible.Auth.{User, Password}

_user =
  Repo.insert!(%User{
    name: "A B",
    email: "a@b.com",
    password_hash: Password.hash("abc"),
    email_verified: true
  })
```
I don't get on what stage should I run migrations. The seed is not running automatically.

StreetStrider:
It seemes that most convenient way is to put `/app/bin/plausible eval "Plausible.Release.seed()"` into compose's `command` before run, but it would be executed every time. I'm not sure is it idempotent as migrate command or not. If not, it would create seed's dups on every start.

ruslandoga:
I think the easiest way is to mount the seeds script as a volume and run it in `command`. I can help you make it idempotent.

StreetStrider:
@ruslandoga okay, I will continue investigating. Thank you for help.

StreetStrider:
For the record, I ended up using `COPY seed` + `command: eval Plausible.Release.seed` as above. The seed script got guard condition to determine db is empty, something like:
```exs
unless Repo.one(from(User)) do
  # run seed code
end
```

ErikHumphrey:
@StreetStrider Are you using anything to automatically sign in as the admin user? This seems to remove that as well for people that used an OAuth2 proxy, Keycloak, etc. to sign into a single user. Since sites aren't shared across users, you can't really use this in an organization anymore, right?

StreetStrider:
@ErikHumphrey we do not sign in automatically, that's still too much tech for us. We use the same secret storage to deploy and to login, but login is manual. The password is plain in our case.

StreetStrider:
@ErikHumphrey if you was referring to `sites` in plausible, then you can actually add any memberships. My seed script adds single member to single site, but it is flexible as any plain script, so you can create multiple sites and memberships.
```exs
	Repo.insert!(%Membership {
		site: site, # ref to previously created site
		user: owner, # ref to previously created admin user
		role: :owner,
	})
```

tintin10q:
This is so inconvenient. How does this make the first launch experience better for self hosters at all? You could have just kept the `ADMIN_USER_PWD` environments vars working. Also it is not stated anywhere really that these env vars got removed so I spend quite some time trying to plausible work, automatically registering the user but it was never going to work. 

Very bad change in my opinion. It is so much more complicated while it was working fine. You could have just also kept the env vars.

nh2:
Yeah, this bit me too. Easy automated CI deployment now no longer works. I'll try to use a workaround as described above, but it is not nice that one now needs to deal with internals instead of having an official automatic way to auto-deploy.

In NixOS this allowed an automatic fully-declarative deploy approach; unfortunately due to this change, that had to get removed in https://github.com/NixOS/nixpkgs/pull/356221

PR #2358, opened 2022-10-19
Google APIs integration improvements

aerosol:
### Changes

This PR aims to fix several bugs and regressions. Unfortunately due to tight coupling it's not really possible to split this into smaller pieces. Reviewing commits one by one should help a little.

Summary of changes:

1. Referrer drilldown should no longer crash in case of Google API errors, be it invalid grant, insufficient scope or random HTTP errors (sentry: APP-1ZY)

2. Any search console integration errors should be now caught correctly, both in the search console setting and in the react component. React component displays an info visible only to the site owners:

image

This is a small improvement. At first I made an attempt at translating all the possible error codes but that gets messy very quickly - arbitrary lengths and messages don't play well with this small UI window. We don't have a standardised way of translating error codes into human readable error messages in react either, so there's definitely room for future improvement here. However, from there the user can be taken to the search console settings where we do our best to reveal the details (a regression, now tested, was found for the "invalid grant" error too):

image

It is currently my working assumption that the "invalid grant" error can happen any any stage, including Google Token refresh and is always carried with HTTP 400. In that case the following error message is presented to the user:

> Invalid Grant error returned from Google. See here on how to fix it. 

(linking to https://plausible.io/docs/google-search-console-integration#i-get-the-invalid-grant-error)

In case of 401 and 403 errors, we assume this is permanent integration error - either invalid API scope (the user forgot to tick all the boxes) or an invalid token. In this case, the user is presented with a re-link suggestion:

> Your Search Console account hasn't been connected successfully. Please unlink your Google account and try linking it again.

If the above fails, and we're yet experiencing another type of error, a generic message is shown:

> Something went wrong, but looks temporary. If the problem persists, try re-linking your Google account.

3. Non-user facing changes: tests were added and cleaned up a little, as we're slowly moving away from Double mocks in favor of Mox. This helps us finally test any integration failure modes going forward.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/google/http.ex
+++ b/lib/plausible/google/http.ex
@@ -178,25 +178,20 @@ defmodule Plausible.Google.HTTP do
     url = "#{api_url()}/webmasters/v3/sites/#{property}/searchAnalytics/query"
     headers = [{"Authorization", "Bearer #{access_token}"}]
 
-    case HTTPClient.post(url, headers, params) do
+    case HTTPClient.impl().post(url, headers, params) do
       {:ok, %Finch.Response{body: body, status: 200}} ->
         {:ok, body}
 
-      {:error, %{reason: %Finch.Response{body: body, status: 401}}} ->
-        Sentry.capture_message("Error fetching Google queries", extra: %{body: inspect(body)})
-        {:error, :invalid_credentials}
+      {:error, %{reason: %Finch.Response{body: _body, status: status}}}
+      when status in [401, 403] ->
+        {:error, "google_auth_error"}
 
-      {:error, %{reason: %Finch.Response{body: body, status: 403}}} ->
-        Sentry.capture_message("Error fetching Google queries", extra: %{body: inspect(body)})
-        {:error, get_in(body, ["error", "message"])}
+      {:error, %{reason: %{body: %{"error" => error}}}} ->
+        {:error, error}
 
-      {:error, %{reason: %Finch.Response{body: body}}} ->
-        Sentry.capture_message("Error fetching Google queries", extra: %{body: inspect(body)})
-        {:error, :unknown}
-
-      {:error, %{reason: _} = e} ->
-        Sentry.capture_message("Error fetching Google queries", extra: %{error: inspect(e)})
-        {:error, :unknown}
+      {:error, reason} ->
+        Logger.error("Google Analytics: failed to list stats: #{inspect(reason)}")
+        {:error, "failed_to_list_stats"}
     end
   end
 
--- a/lib/plausible/google/http.ex
+++ b/lib/plausible/google/http.ex
@@ -104,13 +104,19 @@ defmodule Plausible.Google.HTTP do
     url = "#{api_url()}/webmasters/v3/sites"
     headers = [{"Content-Type", "application/json"}, {"Authorization", "Bearer #{access_token}"}]
 
-    case HTTPClient.get(url, headers) do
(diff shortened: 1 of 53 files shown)


lib/plausible/google/http.ex
```diff
@@ -40,12 +40,12 @@ defmodule Plausible.Google.HTTP do
          {:ok, report} <- convert_to_maps(report) do
       {:ok, {report, token}}
     else
-      {:error, %{reason: %{status: _non_http_200, body: body}}} ->
-        Sentry.Context.set_extra_context(%{google_analytics_response: body})
+
+      {:error, %{reason: %{status: status, body: body} = response}} ->
+        Sentry.Context.set_extra_context(%{ga_response: %{body: body, status: status}})
```
aerosol, line 45: cc @vinibrsl - this is https://github.com/plausible/analytics/commit/95aafb477d46857af2a330653bc6f76e027a5329 incorporated

lib/plausible_web/templates/site/settings_search_console.html.eex
```diff
@@ -36,12 +36,24 @@
...
         <% {:error, error} -> %>
-          <p class="text-gray-700 dark:text-gray-300 mt-6">The following error happened when fetching your Google Search Console domains.</p>
+          <p class="text-gray-700 dark:text-gray-300 mt-6">The following error happened when fetching your Google Search Console domains:</p>
 
-          <%= if error == "invalid_grant" do %>
-            <p class="text-red-700 font-medium mt-3"><a href="https://plausible.io/docs/google-search-console-integration#i-get-the-invalid-grant-error">Invalid Grant error returned from Google. <span class="text-indigo-500">See here on how to fix it.</a></p>
-          <% else %>
-            <p class="text-red-700 font-medium mt-3">Something went wrong</p>
+          <%= case error do %>
+            <% "invalid_grant" ->  %>
```
vinibrsl, line 42: What do you think of returning atoms instead of strings?

vinibrsl (approved):
This is a huge improvement! Thanks a lot :)

lib/plausible_web/templates/site/settings_search_console.html.eex
aerosol: We can mix them, but the errors dynamically fetched from Google responses are strings after all

aerosol:
> Were you able to test it locally with GA OAuth?

To an extent, I tested some of the failures locally. We should probably put that on staging to make sure there's no regressions.

lib/plausible_web/templates/site/settings_search_console.html.eex
vinibrsl: Let's keep it as strings then

aerosol: I think going forward as we clean this up, we could introduce error structs containing error codes and messages for example. But that's for another day.

aerosol:
@vinibrsl @RobertJoonas let's not merge this just yet, I have some more testing to do

test/support/test_utils.ex
```diff
@@ -2,6 +2,34 @@ defmodule Plausible.TestUtils do
...
   alias Plausible.Factory
 
+  defmacro __using__(_) do
+    quote do
+      require Plausible.TestUtils
+      import Plausible.TestUtils
+    end
+  end
+
+  defmacro patch_env(env_key, value) do
```
ruslandoga, line 12: can it be a function?

Issue #2166, opened 2022-09-02
IP Geo Location showing ISP HQ location


Why8046:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

Geo locations shows not current location but the location of the ISP.

Visited chisca.com from my office in Chamblee GA from laptop and mobile phone.

Stats show:
Marietta GA for laptop (comcast)
New York City for mobile phone (Verizon)

This will become a problem when we do local marketing and cannot see real location of visitors.
![Screenshot_3](https://user-images.githubusercontent.com/112729984/188232325-f3554910-a778-49b5-9559-cb0dc6942bbc.jpg)

On another site, it shows me logged in 3 hours from where I actually am.

### Expected behavior

Stats should show exact location where I am when visiting the site.

### Screenshots

![Screenshot_3](https://user-images.githubusercontent.com/112729984/188232442-af3004b3-ee42-4b30-9ecb-9c348d681b0a.jpg)

### Environment

```markdown
- OS: Windows 10
- Browser: Chrome
- Browser Version: Version 104.0.5112.102 (Official Build) (64-bit)


- OS: Android
- Browser: DuckDuckGo 
- Browser Version:5.1300
```

metmarkosaric:
Thanks for reporting that @Why8046!

We use the [DB-IP](https://db-ip.com/) database to look up the location of an IP address. Could you please test your IP addresses on their website to see what location they show?

We tested several geolocation database providers over several months and DB-IP was able to locate the most IP addresses in the most accurate way. Here's what they say on their website:

"We aggregate data from a large number of sources and through agreements with popular ISPs. We constantly evolve our machine learning models for more accurate IP address geolocation."

What we've noticed is that in situations where they don't have the exact match on the location, rather than not showing any location for that visitor they tend to look at the wider area. Other database providers we tested would simply not show any location in those cases so for us this was a better solution.

Unfortunately, when DB-IP gives us the wrong city for some IP address, there's nothing we can do on our end other than report it and hope that their database improves. We constantly get any improvements that they make so when they make updates, our geolocation accuracy improves too.

Sorry about that and thanks for understanding!

metmarkosaric:
[...]

PR #2362, opened 2022-10-22
replace geolix with locus

ruslandoga:
### Changes

This PR replaces geolix with [locus](https://github.com/g-andrade/locus) to simplify self-hosted setup. `locus` can auto-update maxmind dbs which are [recommended](https://github.com/plausible/analytics/discussions/2324#discussioncomment-3885027) for self-hosters if they want city-level geolocation. [`locus` is also a bit faster.](https://github.com/ruslandoga/locus-geolix-bench)

This PR also uses a test mmdb file from https://github.com/maxmind/MaxMind-DB for e2e geolocation tests without stubs.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/geo/locus.ex
```diff
@@ -0,0 +1,48 @@
...
+  def lookup(ip_address) do
+    case :locus.lookup(@db, ip_address) do
+      {:ok, entry} ->
+        entry
+
+      :not_found ->
+        nil
+
+      {:error, reason} ->
+        raise "failed to lookup ip address #{inspect(ip_address)}: " <> inspect(reason)
```
ruslandoga, line 45: raising on invalid ip address

lib/plausible/geo/locus.ex
```diff
@@ -0,0 +1,48 @@
...
+      path = opts[:path] ->
+        :ok = :locus.start_loader(@db, path)
+
+      true ->
+        raise "failed to load geolocation db: need :path or :license_key to be provided"
+    end
+
+    unless opts[:async] do
+      {:ok, _version} = :locus.await_loader(@db)
+    end
```
ruslandoga, line 22: waiting for db to be loaded by default, blocks root supervisor startup

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -967,7 +967,7 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
 
       conn
       |> put_req_header("user-agent", @user_agent)
-      |> put_req_header("x-forwarded-for", "982.32.12.1")
+      |> put_req_header("x-forwarded-for", "82.32.12.1")
```
ruslandoga, line 970: not sure if it was a typo or intentional but in the case of an invalid ip address, `locus:lookup` would return `{:error, {:invalid_address, "982.32.12.1"}}` which would result in a `raise`d exception in `Plausible.Geo`

```elixir
** (RuntimeError) failed to lookup ip address "982.32.12.1": {:invalid_address, "982.32.12.1"}
  ...__STACKTRACE__...
```

lib/plausible/geo/locus.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @impl true
+  def load_db(opts) do
+    cond do
+      license_key = opts[:license_key] ->
+        edition = opts[:edition] || "GeoLite2-City"
+        :ok = :locus.start_loader(@db, {:maxmind, edition}, license_key: license_key)
+
+      path = opts[:path] ->
+        :ok = :locus.start_loader(@db, path)
```
ruslandoga, line 14: if `locus:start_loader` returns an error, this function crashes with a match error, I think it's good enough as the error reason would still be easily visible in logs

lib/plausible/geo/locus.ex
```diff
@@ -0,0 +1,48 @@
...
+
+      true ->
+        raise "failed to load geolocation db: need :path or :license_key to be provided"
```
ruslandoga, line 17: raising here too in case not enough info was provided

test/support/geo_stub.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Plausible.Geo.Stub do
```
ruslandoga, line 1: using a stub as before, a mox mock can be used instead in tests now that the geolocation lib is wrapped in a behaviour module

lib/plausible/geo/adapter.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Plausible.Geo.Adapter do
+  @moduledoc "Behaviour to be implemented by geolocation adapters"
+
+  @type entry :: map
+  @type opts :: Keyword.t()
+  @type ip_address :: :inet.ip_address() | String.t()
+
+  @callback load_db(opts) :: :ok
+  @callback database_type :: String.t() | nil
+  @callback lookup(ip_address) :: entry | nil
```
ruslandoga, line 10: - [ ] I think I'll make `entry` more strictly typed and make it contain only the info used in events

ruslandoga: I'll try out test db tomorrow: https://github.com/maxmind/MaxMind-DB

ruslandoga: https://github.com/ruslandoga/analytics/pull/19

test/plausible_web/controllers/api/external_controller_test.exs
vinibrsl: What are other cases for exceptions?

I suggest returning an error tuple and leaving geolocation-dependent fields empty in case of failure, instead of throwing an exception and rejecting the event.

vinibrsl (commented):
Good job on this!

test/plausible_web/controllers/api/external_controller_test.exs
ruslandoga: @vinibrsl https://github.com/g-andrade/locus/blob/a0a580e09cca7d1bd3ffefe5739649e06ac49db5/src/locus.erl#L480-L482

Other causes are configuration mistakes.

ruslandoga: I'll make it return an error tuple, but I wonder what should be done with logs. I wouldn't want lookup failures to go unnoticed.

lib/plausible/geo/adapter.ex
ruslandoga: ok to merge it into this branch? `ruslandoga:real-locus -> ruslandoga:locus`

config/runtime.exs
```diff
@@ -430,17 +432,24 @@ config :kaffy,
...
+        edition: maxmind_edition
+      ]
+
+    ip_geolocation_db ->
+      [path: ip_geolocation_db]
+
+    true ->
+      # adapter would fail on incorrect / missing opts anyway
+      # so leaving unhandled cases as empty opts
+      []
```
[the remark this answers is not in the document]
ruslandoga: I'm not sure either if geolocation is required but country mmdb is included by default and I guess most people would want it to be enabled.

test/plausible_web/controllers/api/external_controller_test.exs
ruslandoga: done in https://github.com/plausible/analytics/pull/2362/commits/4e94cebe643ccc65fc75800978a83f4b631c4828

test/support/geo_stub.ex
ruslandoga: using a test mmdb instead now: https://github.com/ruslandoga/analytics/pull/19

test/plausible_web/controllers/api/external_controller_test.exs
vinibrsl: Looks good! Thanks

vinibrsl:
I pushed your changes to staging, and it's looking good! Could you please help me doing some manual testing? https://staging.srv.plausible.io/

I believe we can merge and deploy it to production on Monday.

ruslandoga:
I'll send some events to dummy.site :)

lib/mix/tasks/download_country_database.ex
```diff
@@ -34,7 +34,7 @@ defmodule Mix.Tasks.DownloadCountryDatabase do
 
     if res.status_code == 200 do
       File.mkdir("priv/geodb")
-      File.write!("priv/geodb/dbip-country.mmdb", res.body)
+      File.write!("priv/geodb/dbip-country.mmdb.gz", res.body)
```
ruslandoga, line 37: `locus` trusts the file extension. `locus` didn't try to gunzip the file because it didn't have a `.gz` suffix.

ruslandoga: For reference, the error that `locus` logged was

```
[error] [locus] [geolocation] database failed to load (:filesystem): {:unpack_database_from, :mmdb_blob,
 {:bad_metadata,
  {:marker_not_found,
   <<171, 205, 239, 77, 97, 120, 77, 105, 110, 100, 46, 99, 111, 109>>}}}
```

metmarkosaric:
> I sent some events to dummy.site with "forwarded" headers to simulate different ip addresses.
> 
> One thing I noticed now and also before these changes is that for some locations regions seem to be showing cities and cities seem to be showing city districts. Like [`Region=Tokyo, City=Shinjuku`](https://staging.srv.plausible.io/dummy.site?city=11790353&city_name=Shinjuku) I think it's a dbip quirk.
> 
> As a user, for the Shinjuku example above I'd expect to see `Region=Kanto, City=Tokyo[, District=Shinjuku]`

yeah, that's not ideal. we've been manually changing some of these districts into cities ourselves to make it better when people report it but it's difficult to do the whole world like that. see https://github.com/plausible/analytics/blob/master/lib/plausible/ingestion/city_overrides.ex. is there a better way to do it?

ruslandoga:
Before I go into speculation on how to work around this, I'll post an example response from dbip-city-lite for a Shinjuku ip address to better understand what we are dealing with:

```elixir
iex(6)> :locus.lookup(:dbip, "124.150.157.214")
{:ok,
 %{
   "city" => %{"names" => %{"en" => "Shinjuku"}},
   "continent" => %{
     "code" => "AS",
     "geoname_id" => 6255147,
     "names" => %{
       "de" => "Asien",
       "en" => "Asia",
       "es" => "Asia",
       "fa" => " آسیا",
       "fr" => "Asie",
       "ja" => "アジア大陸",
       "ko" => "아시아",
       "pt-BR" => "Ásia",
       "ru" => "Азия",
       "zh-CN" => "亚洲"
     }
   },
   "country" => %{
     "geoname_id" => 1861060,
     "is_in_european_union" => false,
     "iso_code" => "JP",
     "names" => %{
       "de" => "Japan",
       "en" => "Japan",
       "es" => "Japón",
       "fa" => "ژاپن",
       "fr" => "Japon",
       "ja" => "日本",
       "ko" => "일본",
       "pt-BR" => "Japão",
       "ru" => "Япония",
       "zh-CN" => "日本"
     }
   },
   "location" => %{"latitude" => 35.6967, "longitude" => 139.709},
   "subdivisions" => [%{"names" => %{"en" => "Tokyo"}}]
 }}
```

The full dbip response from the paid version would also contain geonames ids.

ruslandoga:
So what we have is basically wrong data, in my opinion. DB-IP thinks Shinjuku is a city, there is no way around it just by using the data we have from this response. But since the paid version includes geoname id, it can be used to lookup the geoname info by the id and check what kind of place it is: https://www.geonames.org/1852083/shinjuku-ku.html

Shinjuku is listed as a second-order administrative division, that information can be used to automatically rename the "city" key into something like "district" before inserting into the DB.

ruslandoga:
> So what we have is basically wrong data, in my opinion. DB-IP thinks Shinjuku is a city, there is no way around it just by using the data we have from this response. But since the paid version includes geoname id, it can be used to lookup the geoname info by the id and check what kind of place it is: https://www.geonames.org/1852083/shinjuku-ku.html
> 
> Shinjuku is listed as a second-order administrative division, under Tokyo, maybe that information can be used to automatically rename the "city_geoname_id" key into something like "district_geoname_id" before inserting into the DB. Something similar is already being done in https://github.com/plausible/location, but on the "presenting" side, when turning geoname ids into city names before rendering the dashboard.

I tried to research a bit what a second-order administrative division is, and it usually means something like a city in most countries. But from https://en.wikipedia.org/wiki/Municipalities_of_Japan

> Each of the [23 special wards](https://en.wikipedia.org/wiki/23_special_wards) of [Tokyo](https://en.wikipedia.org/wiki/Tokyo) is legally equivalent to a city, though sometimes the 23 special wards as a whole are regarded as one city.

So maybe in this particular case, Shinjuku is indeed a city-like administrative division, even though it sounds a bit strange.

metmarkosaric:
would be great if we can improve this somehow. we get three types of feedback for cities (everyone seems to be happy with countries and regions). here they are ranked by how often we hear about them:

1. you show my city as my ISP location or some big city a few hundreds kilometers away from me (similar to this https://github.com/plausible/analytics/issues/2166). not sure if there's something to do here other than update the database 

2. this is a district and not a city (we manually override these now if/when people report them)

3. you don't show any city data for a big part of my visitors (some people expect to see cities for all their visitors). in our testing before we released this, db-ip actually showed city data for more visitors than other databases did

vinibrsl:
Hey 👋! I pushed some commits to this branch and merged with master. We're also ready to give MaxMind in Plausible Cloud a try after some benchmarking.

lib/plausible/geo.ex
```diff
@@ -0,0 +1,164 @@
...
+  require Logger
+
+  @db :geolocation
+
+  @doc """
+  Starts the geodatabase loading process. Two options are supported, local file and maxmind key.
+
+  Loading a local file:
+
+  iex> load_db(path: "/etc/plausible/dbip-city.mmdb")
```
ruslandoga, line 15: I think these code snippets need to be indented to show up as code blocks in docs.

lib/plausible/ingestion/event.ex
```diff
@@ -168,7 +168,7 @@ defmodule Plausible.Ingestion.Event do
   end
 
   defp put_geolocation(%__MODULE__{} = event) do
-    result = Plausible.Ingestion.Geolocation.lookup(event.request.remote_ip)
+    result = Plausible.Ingestion.Geolocation.lookup(event.request.remote_ip) || %{}
```
ruslandoga, line 171: Since `get_in` is used, is there a need for `%{}` fallback?

lib/plausible/application.ex
```diff
@@ -119,4 +120,9 @@ defmodule Plausible.Application do
     OpentelemetryEcto.setup([:plausible, :clickhouse_repo])
     OpentelemetryOban.setup()
   end
+
+  defp setup_geolocation do
+    opts = Application.fetch_env!(:plausible, Plausible.Geo)
+    :ok = Plausible.Geo.load_db(opts)
```
ukutaht, line 126: Since this seems to be async by default, I think it would also be a good idea to add this loading step to the `/api/health` endpoint like we do with the sites cache. This way the load balancer will not route traffic to a newly restarted node until the geo database is loaded.

Side question: how long does it take to download and be ready to run geolocation lookups with the Maxmind license key?

Side note: not a concern for this PR but I think we're getting to a point where we should really separate [liveness vs readiness](https://developers.redhat.com/blog/2020/11/10/you-probably-need-liveness-and-readiness-probes) probes as suggested by @cnkk. Loading the geolocation DB is required for the app to be 'ready' but it's independent from it being 'live'.

ukutaht (commented):
Looks good to me, just one comment about adding geolocation readiness status to our healthcheck endpoint.

Once this is tested and we've settled on using Locus+Maxmind, we also need to document this for self-hosted and include it in the release notes for the next release (cc @ruslandoga)

lib/plausible/application.ex
ruslandoga: I think it's sync by default in this implementation.

lib/plausible/geo.ex
```diff
@@ -0,0 +1,164 @@
...
+      path = opts[:path] ->
+        :ok = :locus.start_loader(@db, path)
+
+      true ->
+        raise "failed to load geolocation db: need :path or :license_key to be provided"
+    end
+
+    unless opts[:async] do
+      {:ok, _version} = :locus.await_loader(@db)
+    end
```
ruslandoga, line 40: We are waiting for loader to complete unless `async: true` is provided.

ukutaht: Oh right. I sometimes have trouble reading `unless` clauses...

lib/plausible/application.ex
ruslandoga: > Side question: how long does it take to download and be ready to run geolocation lookups with the Maxmind license key?

On my laptop on meh co-working wifi:

```elixir
Mix.install [:locus]

:timer.tc fn ->
  Application.put_env(:locus, :license_key, "XXdDMc5OhchOTazu") # I'll delete it in a few days
  :ok = :locus.start_loader(:city, {:maxmind, "GeoLite2-City"}, [:no_cache])
  {:ok, _} = :locus.await_loader(:city)
end

#=> {6204296, {:ok, {{2023, 1, 10}, {15, 33, 48}}}}
```

ukutaht: Hmm. Seeing as this is sync and called before `Location.load`, I assume this will slow down the whole startup process. If this loading was async, it could run in parallel and probably finish before `Location.load` is done if I understand right. That seems desirable so the whole startup process is only limited by the slowest piece which would be `Location.load`. What do you think @ruslandoga @vinibrsl ?

ruslandoga: Right, maybe we can fire it off before `Location.load`, and await for it with something like `Plausible.Geo.await_setup()` after `Location.load` (just in case).

ukutaht: Yes, that's a good idea

lib/plausible/geo.ex
vinibrsl: Good catch. Fixed and added some more documentation in 18a904c6db186859b892dbf7d68d35c7074d00b5

Screenshot 2023-01-11 at 11 16 14

lib/plausible/application.ex
vinibrsl: I tried using `:locus.await_load` but I couldn't know when the task was completed. The alternative I considered is to use [`Task.await_many/2`](https://hexdocs.pm/elixir/1.12/Task.html#await_many/2) to start Location and Geolix in parallel. This ensures the app will never be in an invalid state where Geolix or Location haven't yet finished start. Changed in 506c2f1ccac483c2082325830c15b9ab7a4015aa. Let me know what you think.

ruslandoga: > I tried using :locus.await_load but I couldn't know when the task was completed.

What was the problem?

lib/plausible/application.ex
```diff
@@ -30,10 +30,18 @@ defmodule Plausible.Application do
...
     setup_opentelemetry()
-    setup_geolocation()
-    Location.load_all()
+
+    Task.await_many(
+      [
+        Task.async(&setup_geolocation/0),
+        Task.async(&Location.load_all/0)
+      ],
+      15_000
```
ruslandoga, line 42: This is probably too low. Some users are having problems with even the 30s default in `Location.load_all`

lib/plausible/application.ex
vinibrsl: The problem was that I was using it wrong 😅. Fixed in c775070e678c3a825550c0f6ec1a4ac611274e5b

vinibrsl (approved):

PR #2371, opened 2022-10-25
ignore .DS_Store

ruslandoga:
### Changes

I can make it globally ignored but I prefer having it in `.gitignore` instead.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ plausible-report.xml
 
 # Auto-generated tracker files
 /priv/tracker/js/*.js
+
+# For apple fanboys
+.DS_Store


.gitignore
```diff
@@ -66,3 +66,6 @@ plausible-report.xml
 
 # Auto-generated tracker files
 /priv/tracker/js/*.js
+
+# For apple fanboys
```
aerosol, line 70: The comment sounds hostile, but also this is specific ignore can be kept [globally](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c#global-gitignore-contents) outside of the project - nothing to do with it after all.

PR #2382, opened 2022-10-25
Trim goals user input

aerosol:
### Changes

People are likely to enter (copy/paste) goals from external sources
which can lead to whitespace characters appended by accident.
That renders the goal unusable and hard to distinct visually.

Normally to fix up existing goals we would use a data migration,
but this should be good enough to check if the problem
with never appearing goals resurfaces.

Going forward it's a good idea to trim all the input coming from forms,
maybe we could put that in the backlog cc @vinibrsl @RobertJoonas 

### Tests

### Changelog

### Documentation

### Dark mode


RobertJoonas (approved):
Looks good!

PR #2406, opened 2022-10-31
Changing the settings link in the renewal email

metmarkosaric:
Someone mentioned there's no way to cancel from the link we include in this email (https://plausible.io/billing/upgrade) so I'm changing it to the settings page instead https://plausible.io/settings. My own https://plausible.io/billing/upgrade is a bit different than normal subscribers see it so I don't actually know if there's a way to cancel on it. If there is a way there too, feel free to ignore this


ukutaht (approved):

ukutaht:
Thanks @metmarkosaric !

PR #2409, opened 2022-11-01
Remove ingestion from OTEL tracing

vinibrsl:
This commit removes the current 1% sampling on ingestion to completely ignore ingestion traces. We are not getting enough value from those, and we'll keep dashboard, API and background jobs traces only.


aerosol (approved):
:fire:

PR #2417, opened 2022-11-02
Add API support for intervals

vinibrsl:
This pull request adds API support for different intervals in queries (weekly, daily, hourly, etc). I cherry-picked @Vigasaurus' changes from the Intervals PR (#1574) so we can release this feature incrementally and review in chunks.

### DIFF
--- a/lib/plausible/stats/fragments.ex
+++ b/lib/plausible/stats/fragments.ex
@@ -47,30 +47,52 @@ defmodule Plausible.Stats.Fragments do
   end
 
   @doc """
-  Returns the nearest Monday not past a given date. If the nearest Monday is
-  past the given date, returns the latter.
+  Returns the weekstart for `date`. If the weekstart is before the `not_before`
+  boundary, `not_before` is returned.
+
+  ## Examples
+
+  In this pseudo-code example, the fragment returns the weekstart. The
+  `not_before` boundary is set to the past Saturday, which is before the
+  weekstart, therefore the cap does not apply.
+
+    iex> this_wednesday = ~D[2022-11-09]
+    ...> past_saturday = ~D[2022-11-05]
+    ...> weekstart_not_before(this_wednesday, past_saturday)
+    ~D[2022-11-07]
+
+
+  In this other example, the fragment returns Tuesday and not the weekstart.
+  The `not_before` boundary is set to Tuesday, which is past the weekstart,
+  therefore the cap applies.
+
+    iex> this_wednesday = ~D[2022-11-09]
+    ...> this_tuesday = ~D[2022-11-08]
+    ...> weekstart_not_before(this_wednesday, this_tuesday)
+    ~D[2022-11-08]
+
   """
-  defmacro nearest_monday_not_past(date_to_round, not_past_date) do
+  defmacro weekstart_not_before(date, not_before) do
     quote do
       fragment(
         "if(toMonday(?) < toDate(?), toDate(?), toMonday(?))",
-        unquote(date_to_round),
-        unquote(not_past_date),
-        unquote(not_past_date),
-        unquote(date_to_round)
+        unquote(date),
+        unquote(not_before),
+        unquote(not_before),
+        unquote(date)
       )
     end
   end
 
   @doc """
-  Same as Plausible.Stats.Fragments.nearest_monday_not_past/2 but converts dates
-  to the specified timezone.
+  Same as Plausible.Stats.Fragments.weekstart_not_before/2 but converts dates to
+  the specified timezone.
   """
-  defmacro nearest_monday_not_past(date_to_round, not_past_date, timezone) do
+  defmacro weekstart_not_before(date, not_before, timezone) do
     quote do
-      nearest_monday_not_past(
-        to_timezone(unquote(date_to_round), unquote(timezone)),
-        unquote(not_past_date)
+      weekstart_not_before(
+        to_timezone(unquote(date), unquote(timezone)),
+        unquote(not_before)
       )
     end
   end
--- a/lib/plausible/stats/fragments.ex
+++ b/lib/plausible/stats/fragments.ex
@@ -35,6 +35,46 @@ defmodule Plausible.Stats.Fragments do
     end
   end
 
+  @doc """
+  Converts time or date and time to the specified timezone.
+
+  Reference: https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions/#totimezone
+  """
+  defmacro to_timezone(date, timezone) do
+    quote do
+      fragment("toTimeZone(?, ?)", unquote(date), unquote(timezone))
+    end
+  end
+
+  @doc """
+  Returns the nearest Monday not past a given date. If the nearest Monday is
+  past the given date, returns the latter.
+  """
+  defmacro nearest_monday_not_past(date_to_round, not_past_date) do
+    quote do
+      fragment(
+        "if(toMonday(?) < toDate(?), toDate(?), toMonday(?))",
+        unquote(date_to_round),
+        unquote(not_past_date),
+        unquote(not_past_date),
+        unquote(date_to_round)
+      )
+    end
+  end
+
+  @doc """
+  Same as Plausible.Stats.Fragments.nearest_monday_not_past/2 but converts dates
+  to the specified timezone.
+  """
+  defmacro nearest_monday_not_past(date_to_round, not_past_date, timezone) do
+    quote do
+      nearest_monday_not_past(
+        to_timezone(unquote(date_to_round), unquote(timezone)),
+        unquote(not_past_date)
+      )
+    end
+  end
+
   defmacro __using__(_) do
     quote do
       import Plausible.Stats.Fragments
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -3,6 +3,7 @@ defmodule Plausible.Stats.Imported do
   alias Plausible.Stats.Query
   import Ecto.Query
   import Plausible.Stats.Base
+  import Plausible.Stats.Fragments
 
   @no_ref "Direct / None"
 
@@ -42,26 +43,8 @@ defmodule Plausible.Stats.Imported do
     {first_datetime, _} = utc_boundaries(query, site.timezone)
 
     imported_q
-    |> group_by(
-      [i],
-      fragment(
-        "if(toMonday(?) < toDate(?), toDate(?), toMonday(?))",
-        i.date,
-        ^first_datetime,
-        ^first_datetime,
-        i.date
-      )
-    )
-    |> select_merge([i], %{
-      date:
-        fragment(
-          "if(toMonday(?) < toDate(?), toDate(?), toMonday(?))",
-          i.date,
-          ^first_datetime,
-          ^first_datetime,
-          i.date
-        )
-    })
+    |> group_by([i], nearest_monday_not_past(i.date, ^first_datetime))
+    |> select_merge([i], %{date: nearest_monday_not_past(i.date, ^first_datetime)})
   end
 
   defp apply_interval(imported_q, _query, _site) do
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -43,8 +43,8 @@ defmodule Plausible.Stats.Imported do
     {first_datetime, _} = utc_boundaries(query, timezone)
 
     imported_q
-    |> group_by([i], nearest_monday_not_past(i.date, ^first_datetime))
-    |> select_merge([i], %{date: nearest_monday_not_past(i.date, ^first_datetime)})
+    |> group_by([i], weekstart_not_before(i.date, ^first_datetime))
+    |> select_merge([i], %{date: weekstart_not_before(i.date, ^first_datetime)})
   end
 
   defp apply_interval(imported_q, _query, _timezone) do
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -23,7 +23,7 @@ defmodule Plausible.Stats.Imported do
         select: %{}
       )
       |> select_imported_metrics(metrics)
-      |> apply_interval(query, site)
+      |> apply_interval(query, site.timezone)
 
     from(s in Ecto.Query.subquery(native_q),
       full_join: i in subquery(imported_q),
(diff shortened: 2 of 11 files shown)


lib/plausible/stats/timeseries.ex
```diff
@@ -60,23 +65,48 @@ defmodule Plausible.Stats.Timeseries do
...
+
+  defp buckets(%Query{interval: "date"} = query) do
     Enum.into(query.date_range, [])
   end
 
-  def buckets(%Query{interval: "hour"} = query) do
-    Enum.map(0..23, fn step ->
+  defp buckets(%Query{interval: "hour"} = query) do
+    n_hours = Timex.diff(query.date_range.last, query.date_range.first, :hours)
+    n_buckets = if n_hours == 0, do: 23, else: n_hours
```
vinibrsl, line 84: @Vigasaurus Would you mind elaborating on why we need 23 hours instead of 24 here? :)

Vigasaurus: Hey, yeah this was kept as 23 to keep the convention as the rest of the buckets functions, which used an inclusive range selector (like on line 92, for example) - so we're going from [0 to 23] inclsuive, which is 24 buckets, instead of [0 to 24), excluding the 24th index which would be the 25th bucket. This could be swapped out if you feel showing the 24 as the number there is easier to understand long-term though.

lib/plausible/stats/timeseries.ex
```diff
@@ -60,23 +65,48 @@ defmodule Plausible.Stats.Timeseries do
...
-  def buckets(%Query{period: "30m", interval: "minute"}) do
+  defp buckets(%Query{period: "30m", interval: "minute"}) do
     Enum.into(-30..-1, [])
   end
 
-  def select_bucket(q, site, %Query{interval: "month"}) do
+  @day_in_minutes 24 * 60
+  defp buckets(%Query{interval: "minute"} = query) do
+    n_minutes = Timex.diff(query.date_range.last, query.date_range.first, :minutes)
+    n_buckets = if n_minutes == 0, do: @day_in_minutes, else: n_minutes
```
aerosol, line 100: Can we have a test for when `n_buckets` is `!@day_in_minutes`?

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -383,4 +383,251 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+      assert %{"plot" => plot} = json_response(conn, 200)
+
+      assert Enum.count(plot) == 181
+      assert List.first(plot) == 1
+      assert Enum.at(plot, 14) == 2
+      assert Enum.at(plot, 45) == 1
+      assert List.last(plot) == 1
+    end
+
+    test "displays visitors for custom on an month scale", %{conn: conn, site: site} do
```
aerosol, line 453:
```suggestion
    test "displays visitors for a custom period on a monthly scale", %{conn: conn, site: site} do
```

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -383,4 +383,251 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+        )
+
+      assert %{"plot" => plot} = json_response(conn, 200)
+
+      assert Enum.count(plot) == 1441
+      assert List.first(plot) == 1
+      assert Enum.at(plot, 15) == 2
+    end
+
+    test "displays visitors for 6mo on an day scale", %{conn: conn, site: site} do
```
aerosol, line 429:
```suggestion
    test "displays visitors for 6mo on a day scale", %{conn: conn, site: site} do
```

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -383,4 +383,251 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+        )
+
+      assert %{"plot" => plot} = json_response(conn, 200)
+
+      assert Enum.count(plot) == 721
+      assert List.first(plot) == 1
+      assert Enum.at(plot, 1) == 1
+    end
+
+    test "displays visitors for a day on an minute scale", %{conn: conn, site: site} do
```
aerosol, line 409:
```suggestion
    test "displays visitors for a day on a minute scale", %{conn: conn, site: site} do
```

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -383,4 +383,251 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+          "/api/stats/#{site.domain}/main-graph?period=month&date=2021-01-01&metric=visitors&interval=month"
+        )
+
+      assert %{"plot" => plot} = json_response(conn, 200)
+
+      assert Enum.count(plot) == 1
+      assert List.first(plot) == 2
+    end
+
+    test "displays visitors for a day on an month scale", %{conn: conn, site: site} do
```
aerosol, line 493:
```suggestion
    test "displays visitors for a day on an monthly scale", %{conn: conn, site: site} do
```

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -383,4 +383,251 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+          "/api/stats/#{site.domain}/main-graph?period=day&date=2021-01-01&metric=visitors&interval=month"
+        )
+
+      assert %{"plot" => plot} = json_response(conn, 200)
+
+      assert Enum.count(plot) == 1
+      assert List.first(plot) == 3
+    end
+
+    test "displays visitors for a month on an week scale", %{conn: conn, site: site} do
```
aerosol, line 512:
```suggestion
    test "displays visitors for a month on a weekly scale", %{conn: conn, site: site} do
```

aerosol (commented):
Comments inline

lib/plausible/stats/timeseries.ex
ukutaht: Welp, I thought coverage would show that this is untested but:

```
100.0% lib/plausible/stats/timeseries.ex             225       74        0
```

Don't have time to dig into how this branch is tested, or whether Elixir coverage reporting is broken.

ukutaht: In general I think it's a good tool when refactoring a module to find opportunities to test legacy code. Consider my code as legacy going forward, it's all basically a 'move-fast-and-break-things' MVP that grew out of proportion 😀

ukutaht (commented):
Thanks for teasing this apart, very helpful for more rigorous reviewing and improvements along the way ❤️ Comments below

lib/plausible/stats/timeseries.ex
aerosol: I'm currently doing it like a caveman either by planting IO.inspect calls or raising an exception in the code to see if it's tested 🤡

lib/plausible/stats/query.ex
```diff
@@ -67,7 +67,7 @@ defmodule Plausible.Stats.Query do
 
     %__MODULE__{
       period: "realtime",
-      interval: "minute",
+      interval: Map.get(params, "interval", "minute"),
```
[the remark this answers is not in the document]
vinibrsl: Good point! 7623153cf72df94eafd9537ccf1b0c94fa295dac adds a fallback for invalid values.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -35,21 +35,49 @@ defmodule PlausibleWeb.Api.StatsController do
 
       labels = Enum.map(timeseries_result, fn row -> row[:date] end)
       present_index = present_index_for(site, query, labels)
+      full_intervals = build_full_intervals(query, labels)
```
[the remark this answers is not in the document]
vinibrsl: @Vigasaurus could you please elaborate on this? I know it's to be used by the intervals client-side but I'm not sure why it's needed.

lib/plausible/stats/timeseries.ex
```diff
@@ -87,7 +117,47 @@ defmodule Plausible.Stats.Timeseries do
...
 
-  def select_bucket(q, site, %Query{interval: "date"}) do
+  defp select_bucket(q, site, %Query{interval: "week"} = query) do
+    {first_datetime, _} = utc_boundaries(query, site.timezone)
+
+    from(
+      e in q,
+      select_merge: %{
+        date:
+          fragment(
```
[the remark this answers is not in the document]
vinibrsl: Good call! I extracted both Timeseries and Imported functions to Fragments (1926e4de)

lib/plausible/stats/timeseries.ex
```diff
@@ -60,23 +65,48 @@ defmodule Plausible.Stats.Timeseries do
...
+  end
+
+  defp buckets(%Query{interval: "date"} = query) do
     Enum.into(query.date_range, [])
   end
 
-  def buckets(%Query{interval: "hour"} = query) do
-    Enum.map(0..23, fn step ->
+  defp buckets(%Query{interval: "hour"} = query) do
+    n_hours = Timex.diff(query.date_range.last, query.date_range.first, :hours)
```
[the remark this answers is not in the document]
vinibrsl: I tried your suggestion and other variations of it, but it also returns `0` when using same day filtering. I changed the conditional to make clearer what the edge case actually is:

```elixir
@full_day 23
defp buckets(%Query{interval: "hour"} = query) do
  n_buckets =
    if query.date_range.first == query.date_range.last do
      @full_day
    else
      Timex.diff(query.date_range.last, query.date_range.first, :hours)
    end

  Enum.map(0..n_buckets, fn step ->
    Timex.to_datetime(query.date_range.first)
    |> Timex.shift(hours: step)
    |> Timex.format!("{YYYY}-{0M}-{0D} {h24}:{m}:{s}")
  end)
end
```
Full commit: 801310b4. Let me know what you think :)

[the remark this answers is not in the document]
ukutaht: Yeah, looks better

lib/plausible/stats/fragments.ex
```diff
@@ -35,6 +35,46 @@ defmodule Plausible.Stats.Fragments do
...
+    quote do
+      fragment("toTimeZone(?, ?)", unquote(date), unquote(timezone))
+    end
+  end
+
+  @doc """
+  Returns the nearest Monday not past a given date. If the nearest Monday is
+  past the given date, returns the latter.
+  """
+  defmacro nearest_monday_not_past(date_to_round, not_past_date) do
```
ukutaht, line 53: It took me along time to understand what this does, and I'm still not sure if I 100% get it. I think part of it is the naming. What does `not past` imply here? It immediately points me to think that it scans the week going forward after the `date_to_round` but in fact it seems to scan backwards if you know what I mean.

Also the name seems to imply that it always returns a Monday which is not the case. For example, in the tests for the month of 2021-09, it returns 2021-09-01 which is a Wednesday.

I'm searching for a name in my head that better describes what this does. If I understand it correctly, isn't it something like `to_monday_but_not_before_a_certain_date_boundary`?

or more succintly, `weekstart_not_before` or something like that?

lib/plausible_web/controllers/api/stats_controller.ex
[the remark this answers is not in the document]
ukutaht: Looking at [the tests](https://github.com/plausible/analytics/blob/801310b445bbadfc59ebc0d8a00deef380bc3e87/test/plausible_web/controllers/api/stats_controller/main_graph_test.exs#L546-L552) for weekly intervals, my understanding is that it tells the client whether we're looking at an interval that's been cut off by the requested date range or not.

So for example, if looking at a month week-by-week, some weeks may be cut off by the month boundaries. It's useful to adjust the graph display slightly in case the interval is not 'full' so that the user understands why the numbers might be lower for those partial periods. I think this should be documented.

Depending on your philosophy, this could also be a frontend concern. But fiddling with dates in JS is a nightmare so I prefer doing it here in Elixir.

lib/plausible/stats/fragments.ex
ukutaht: Looks good!

lib/plausible/stats/timeseries.ex
vinibrsl: Thanks for clarifying! I decremented the minutes bucket as you suggested in e05052aa1272c0f8b064e57235cae502dd13cbf8 :)

lib/plausible/stats/timeseries.ex
vinibrsl: Good call! I added the missing test case in 1bb555f4cd688c6f234223f542b5795b0cb957c6

lib/plausible_web/controllers/api/stats_controller.ex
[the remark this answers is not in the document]
vinibrsl: Thanks for looking that up! I agree that we should have docs for this, and that we should keep this calculation on the Elixir side. I invested some time on reading and documenting the `main_graph` controller action, including the new `full_intervals` map. Let me know if I missed anything: ef6458161414a30be71ff31fbd78bdf3ba111663

lib/plausible/stats/timeseries.ex
```diff
@@ -50,8 +50,13 @@ defmodule Plausible.Stats.Timeseries do
     |> ClickhouseRepo.all()
   end
 
-  def buckets(%Query{interval: "month"} = query) do
-    n_buckets = Timex.diff(query.date_range.last, query.date_range.first, :months)
+  defp buckets(%Query{interval: "month"} = query) do
```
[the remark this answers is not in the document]
vinibrsl: After some testing I don't think this change is necessary. `Timex.diff/3` doesn't seem to care about the day when doing month comparisons. 

```
Timex.diff(~D[2022-01-30], ~D[2020-01-01], :month)
# => 24

Timex.diff(~D[2022-01-30], ~D[2020-01-01], :month)
# => 24
```

I reverted this change in ec3140a7df0a2df05e0deda8473017b7c446e156

lib/plausible/stats/imported.ex
```diff
@@ -31,13 +32,39 @@ defmodule Plausible.Stats.Imported do
...
 
-  defp apply_interval(imported_q, %Plausible.Stats.Query{interval: "month"}) do
+  defp apply_interval(imported_q, %Plausible.Stats.Query{interval: "month"}, _site) do
     imported_q
     |> group_by([i], fragment("toStartOfMonth(?)", i.date))
     |> select_merge([i], %{date: fragment("toStartOfMonth(?)", i.date)})
   end
 
-  defp apply_interval(imported_q, _query) do
+  defp apply_interval(imported_q, %Plausible.Stats.Query{interval: "week"} = query, site) do
```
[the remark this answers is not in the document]
vinibrsl: Good catch! I added the missing test in f588bd86

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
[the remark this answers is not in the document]
vinibrsl: Good call! I don't think we should support weird patterns like `period=day,interval=month` or even `period=month,interval=month`.

I tried different approaches to validate this input, and I decided to go simple with a static list, which is easier to read, change if needed and check valid intervals for each period. Changed in 241d9352d90a769b14e27ed5d41d40bde0b60bf2.

The con of using a static list is `custom` and `all` periods are not strictly validated, meaning you could request intervals bigger than the period. For example: `period=custom,from=2020-01-01,to=2020-01-02,interval=month`.

In any case, `custom` and `all` have to be handled dynamically and are different from static periods (e.g. `month`). Especially if we want to limit heavy queries, as you suggested. I'd like to tackle interval limits in a separate PR after the client-side changes are ready, if you agree. I have some ideas to approach this by limiting the number of buckets.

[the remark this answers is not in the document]
ukutaht: > I have some ideas to approach this by limiting the number of buckets.

I like this direction, sounds cool.

lib/plausible/stats/timeseries.ex
ukutaht: boo 👎 for line strategy

ukutaht: > I'm currently doing it like a caveman either by planting IO.inspect calls or raising an exception in the code to see if it's tested 🤡

I like replacing return values (from functions, if branches, case clauses etc) with random static values that conform to the correct type. In this case just replacing one if branch to return `40` or something. Works nicely in functional languages where everything is an expression. Tests not only that a line is run, but also that the assertion actually checks that it does something meaningful. Still caveman style but I haven't found a better way to poke holes in tests.

ukutaht (approved):

PR #2419, opened 2022-11-03
Trace external HTTP requests

vinibrsl:
This commit adds OTEL tracing to external HTTP requests made with the `Plausible.HTTPClient` module. It records the elapsed time, host, path, and response status code.

### DIFF
--- a/lib/plausible/http_client.ex
+++ b/lib/plausible/http_client.ex
@@ -36,6 +36,9 @@ defmodule Plausible.HTTPClient do
   URL encoding is invoked.
   """
 
+  require OpenTelemetry.Tracer
+  alias Plausible.HTTPClient.Non200Error
+
   @doc """
   Make a POST request
   """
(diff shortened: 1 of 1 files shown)


lib/plausible/http_client.ex
```diff
@@ -62,13 +64,17 @@ defmodule Plausible.HTTPClient do
...
-    |> maybe_decode_body()
-    |> tag_error()
+    OpenTelemetry.Tracer.with_span "http_client.request" do
+      {params, headers} = maybe_encode_params(params, headers)
+
+      method
+      |> build_request(url, headers, params)
+      |> trace_request()
+      |> do_request(finch_req_opts)
+      |> trace_response()
```
aerosol, line 74: Shouldn't we move this after the error is tagged? Otherwise we're not marking e.g. 500s as failed, not sure if intentional.

aerosol (approved):
Just one comment inline

lib/plausible/http_client.ex
vinibrsl: I'm marking `http_client.request.failed` as `true` when the request fails and there is no response, e.g. connection errors, timeouts, etc. What do you think?

aerosol: Sure I mean it that's only failure we're looking to track. Just wondering if a sudden spike in client/server errors would be useful to highlight.

vinibrsl: Good call!

PR #2427, opened 2022-11-09
ensure correct scheme is set for clickhouse repo

ruslandoga:
### Changes

Possibly fixes https://github.com/plausible/analytics/discussions/1391 by adding `:scheme` option.

More info: https://github.com/plausible/analytics/discussions/1391#discussioncomment-4093013

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -84,6 +84,12 @@ ch_db_url =
     "http://plausible_events_db:8123/plausible_events_db"
   )
 
+ch_db_url_scheme =
+  case URI.parse(ch_db_url) do
+    %URI{scheme: "https"} -> :https
+    _other -> :http
+  end
+
 {ch_flush_interval_ms, ""} =
   config_dir
   |> get_var_from_path_or_env("CLICKHOUSE_FLUSH_INTERVAL_MS", "5000")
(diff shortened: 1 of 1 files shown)


config/runtime.exs
```diff
@@ -84,6 +84,12 @@ ch_db_url =
     "http://plausible_events_db:8123/plausible_events_db"
   )
 
+ch_db_url_scheme =
+  case URI.parse(ch_db_url) do
+    %URI{scheme: "https"} -> :https
+    _other -> :http
```
ruslandoga, line 90: I'm keeping the prev behaviour this way, by defaulting to `:http`.

ukutaht (commented):
Is this how other Ecto libraries work as well?


It feels slightly weird to me that the library itself doesn't use the scheme we provide in the URL. Could this be an issue with clickhouse-ecto or clickhousex instead? If so, I'd prefer to fix it in the library rather than here.

ruslandoga:
Ecto handles the url option itself and passes down only the info extracted from the url. Ecto ignores the scheme. Other adapters are generally not dependant on the scheme and use `:ssl` options instead.

ruslandoga:
Closing as https://github.com/elixir-ecto/ecto/pull/4052 would allow to get the scheme info in the adapter.

PR #2428, opened 2022-11-09
Add sites metadata rate limit migration

aerosol:
### Changes

This PR adds postgres migration extending the `sites` tagle with metadata fields required for ingestion rate limiting.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20221109082503_add_rate_limiting_to_sites.exs
+++ b/priv/repo/migrations/20221109082503_add_rate_limiting_to_sites.exs
@@ -2,6 +2,9 @@ defmodule Plausible.Repo.Migrations.AddRateLimitingToSites do
   use Ecto.Migration
 
   def change do
-
+    alter table(:sites) do
+      add :ingest_rate_limit_scale_seconds, :integer, null: true
+      add :ingest_rate_limit_threshold, :integer, null: true
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20221109082503_add_rate_limiting_to_sites.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Plausible.Repo.Migrations.AddRateLimitingToSites do
+  use Ecto.Migration
+
+  def change do
+    alter table(:sites) do
+      add :ingest_rate_limit_scale_seconds, :integer, null: false, default: 60
+      add :ingest_rate_limit_threshold, :integer, null: true
```
ukutaht, line 7: According to the design:

>  If the value is non-0 the limit applies, negative/0 value disables rate-limiting. 

Would it make more sense to make this field non-nullable and 0 by default?

ukutaht: I suppose `null` would also be a good way to disable it :) probably clearer

aerosol: Yup, I think `null` is better to disable it. `0` would mean effectively banned, and negative numbers make no sense :)

ukutaht: yeah that's more intuitive to me as well

ukutaht (approved):

PR #2430, opened 2022-11-10
Fix an issue with hourly stats with non-UTC Clickhouse

ameshkov:
This commit resolves an issue with Clickhouse configured to use a timezone different from UTC.

With the old code we had the following situation:
1. The hourly buckets are built without taking site's timezone into account. In our case these were just 00:00, 01:00, etc.
2. At the same time Clickhouse query was returning dates that did take both the timezone and UTC into account, i.e. hourly stats were starting from 21:00 previous day (as the Clickhouse server lives in GMT+3 timezone in our case).

### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/timeseries.ex
+++ b/lib/plausible/stats/timeseries.ex
@@ -7,7 +7,7 @@ defmodule Plausible.Stats.Timeseries do
   @event_metrics [:visitors, :pageviews]
   @session_metrics [:visits, :bounce_rate, :visit_duration]
   def timeseries(site, query, metrics) do
-    steps = buckets(site, query)
+    steps = buckets(query)
 
     event_metrics = Enum.filter(metrics, &(&1 in @event_metrics))
     session_metrics = Enum.filter(metrics, &(&1 in @session_metrics))
@@ -50,7 +50,7 @@ defmodule Plausible.Stats.Timeseries do
     |> ClickhouseRepo.all()
   end
 
-  def buckets(_site, %Query{interval: "month"} = query) do
+  def buckets(%Query{interval: "month"} = query) do
     n_buckets = Timex.diff(query.date_range.last, query.date_range.first, :months)
 
     Enum.map(n_buckets..0, fn shift ->
@@ -60,12 +60,17 @@ defmodule Plausible.Stats.Timeseries do
     end)
   end
 
-  def buckets(_site, %Query{interval: "date"} = query) do
+  def buckets(%Query{interval: "date"} = query) do
     Enum.into(query.date_range, [])
   end
 
-  def buckets(site, %Query{interval: "hour"} = query) do
-    {first_datetime, _last_datetime} = utc_boundaries(query, site.timezone)
+  def buckets(%Query{interval: "hour"} = query) do
+    # Adjusts hourly buckets in accordance with the Clickhouse server timezone.
+    # See: https://github.com/plausible/analytics/issues/2432.
+    # TODO: replace with ClickHouse's timezone when it's supported by HTTP API.
+    {:ok, res} = ClickhouseRepo.query("SELECT timezone()")
+    [[timezone]] = res.rows
+    {first_datetime, _last_datetime} = utc_boundaries(query, timezone)
 
     Enum.map(0..23, fn step ->
       first_datetime
@@ -74,7 +79,7 @@ defmodule Plausible.Stats.Timeseries do
     end)
   end
 
-  def buckets(_site, %Query{period: "30m", interval: "minute"}) do
+  def buckets(%Query{period: "30m", interval: "minute"}) do
     Enum.into(-30..-1, [])
   end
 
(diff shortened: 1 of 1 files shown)


ameshkov:
Well, not that easy it appears:(

lib/plausible/stats/timeseries.ex
```diff
@@ -65,8 +65,16 @@ defmodule Plausible.Stats.Timeseries do
   end
 
   def buckets(%Query{interval: "hour"} = query) do
+    # Adjusts hourly buckets in accordance with the Clickhouse server timezone.
+    # See: https://github.com/plausible/analytics/issues/2432.
+    # It would be better to replace it with ClickHouse's client timezone when
+    # it is supported by HTTP API.
+    {:ok, res} = ClickhouseRepo.query("SELECT timezone()")
```
aerosol, line 72: Hey @ameshkov, wondering if this could be made a static configuration entry defaulting to UTC rather than something we query in runtime. What do you think? I assume this is for self-hosted instances when the server's tz is known upfront.

cc @vinibrsl as this possibly affects the intervals work

ukutaht: So far I've been comfortable telling people that Plausible only works with Clickhouse configured in the UTC timezone.

Seeing 3 upvotes here, I'd like to understand why you can't run Clickhouse in the UTC timezone? Personally I've never run a database in any other timezone than UTC and think it should be banned 😀Sorry for my ignorance or non-understanding at the moment, but I'd like this to be justified @D13410N3 @sc0rp10 @z1m1n @ameshkov 

If we're going to support this, I do think it should be a static config rather than anything at runtime.

ameshkov: > Personally I've never run a database in any other timezone than UTC and think it should be banned

It makes sense to keep all servers in the same time zone, but it shouldn't necessarily be UTC. Historically, we have all our servers running in GMT+3 and this includes a large Clickhouse instance which we use to host Plausible's database.

The reason for using GMT+3 everywhere is that most of the employees live in GMT+2/GMT+3 and it's not too comfortable for them to do mental gymnastics with time zone conversion all the time. Take data analysts that work with Clickhouse directly and do selects, they'd better see the results in the time zone they're used to.

> If we're going to support this, I do think it should be a static config rather than anything at runtime.

The whole issue is rather specific to Clickhouse and its inability of using the client's timezone via HTTP interface. When they allow that (as I understand, they're going to), there will be no good reason to have that static config. My thinking is that the current PR makes a small change that solves the immediate issue, but in the long run it can and should be removed.

ukutaht: OK I think I understand your use-case better here.

> My thinking is that the current PR makes a small change that solves the immediate issue, but in the long run it can and should be removed.

Currently it puts a database round-trip in a fairly performance-sensitive part of the code. In its current form I won't accept this PR, sorry.

Doing anything like this at runtime is too error-prone IMO. We can accept a static config option that will be hard-deprecated as soon as timezone support is added to Clickhouse's HTTP interface.

ameshkov: What's better, a static config or initializing a variable on startup?

If static config is better, what would be a better name for it? `CLICKHOUSE_TZ`?

aerosol: @ameshkov both. We normally allow overwriting application env config with env vars, see https://github.com/plausible/analytics/blob/master/config/runtime.exs 

As far as names go, there's no harm in being extra verbose about it. Maybe something like:

```elixir
clickhouse_server_timezone = get_var_from_path_or_env(config_dir, "CLICKHOUSE_SERVER_TIMEZONE", "UTC")

config :plausible, Plausible.ClickhouseRepo,
  ...
  server_timezone: clickhouse_server_timezone
```

WYT? /cc @ukutaht

ameshkov: By initializing a variable I meant running a Clickhouse query on startup instead of doing it every time.

If you prefer config then I'll update the PR accordingly.

ukutaht: @ameshkov I've reviewed some of our code to make a better judgement on this.

We have a number of places in the code that assume both databases (Postgres and Clickhouse) are running in the UTC timezone. For the Clickhouse stuff, it's not just about the way we calculate hourly time buckets. It also affects time boundaries for all queries, for example.

In order to support flexible timezone configurations, it would be a lot of work to implement in both the code and tests. Not only would it be hard to implement, it would hurt the long-term maintenance of our codebase.

We can easily support flexible timezones if Clickhouse implements it within their HTTP API. This would not come with any complicated development work or maintenance issues.

The decision at this point is to not accept this PR or any attempt to support non-UTC timezones within the Elixir code itself. The Elixir codebase will always assume that the database is running in UTC. If you want your use-case to work, please direct your communication towards Clickhouse team itself so they implement the HTTP interface.

ameshkov: No problem, your argument makes sense to me.

Issue #2120, opened 2022-08-18
Links in the traffic spike alert emails


metmarkosaric:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

The link to our website is not clickable in the actual email in some email clients (Fastmail for instance) and the link to the dashboard is not sent in the email even though there's a mention of it [in the code](https://github.com/plausible/analytics/blob/e1bb4368460ebb3a0bb86151b143176797b686cc/lib/plausible_web/templates/email/spike_notification.html.eex).

Comment from Twitter: "looks like @link isn’t defined. Also, it’s probably missing the `link` method call to make it clickable for most email clients"

(reported on twitter)

### Expected behavior

Add link to the dashboard and make Plausible link clickable 

### Environment

```markdown
- OS:
- Browser:
- Browser Version:
```

metmarkosaric:
fixed and deployed earlier today by @RobertJoonas!

PR #2431, opened 2022-11-10
Fix email links

RobertJoonas:
### Changes

- [x] Fix a bug where dashboard links were not rendered in traffic spike emails (+ test)
  - bug: https://github.com/plausible/analytics/issues/2120
- [x] Wrap the `plausible_url()` in the email footer into a link element to make it clickable in every email client
- [x] Create a standard email layout template called `base_email` that does the following:
  - greets the user by first name if a user is given in the template assigns
  - renders the https://plausible.io link in the footer
  - renders the Postmark unsubscribe placeholder, if `unsubscribe: true` is in the template assigns.
  - adds
    ```
    Thanks,
    Uku and Marko
    ```
    to the email footer, or
    ```
    Thanks,
    Plausible
    ```
    when application is self-hosted

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/layout/base_email.html.eex
+++ b/lib/plausible_web/templates/layout/base_email.html.eex
@@ -1,0 +1,24 @@
+<%= case Map.get(assigns, :user) do
+  %{name: name} when is_binary(name) -> "Hey #{String.split(name) |> List.first()},"
+  _any -> "Hey,"
+end %>
+<br /><br />
+
+<%= @inner_content %>
+<br /><br />
+
+Thanks,<br />
+<%= if Application.get_env(:plausible, :is_selfhost) do %>
+Plausible
+<% else %>
+Uku and Marko
+<% end %>
+<br /><br />
+
+--
+<br /><br />
+<%= link(plausible_url(), to: plausible_url()) %><br />
+
+<%= if Map.get(assigns, :unsubscribe, false) do %>
+{{{ pm:unsubscribe }}}
+<% end %>
--- a/lib/plausible_web/email.ex
+++ b/lib/plausible_web/email.ex
@@ -305,11 +305,21 @@ defmodule PlausibleWeb.Email do
     })
   end
 
-  defp base_email() do
+  defp base_email(), do: base_email(%{layout: "base_email.html"})
+
+  defp base_email(%{layout: layout}) do
     mailer_from = Application.get_env(:plausible, :mailer_email)
 
     new_email()
     |> put_param("TrackOpens", false)
     |> from(mailer_from)
+    |> maybe_put_layout(layout)
+  end
+
+  defp maybe_put_layout(email, nil), do: email
+
+  defp maybe_put_layout(email, layout) do
+    email
+    |> put_html_layout({PlausibleWeb.LayoutView, layout})
   end
 end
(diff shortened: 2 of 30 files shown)


metmarkosaric:
thanks @RobertJoonas! perhaps you can just use something like "The Plausible Team 💌" as the signature name for all emails. that's what we started using recently on newer created emails such as the GA import email. eventually we can replace all signatures with that rather than specific names now that we're a bigger team

RobertJoonas:
> thanks @RobertJoonas! perhaps you can just use something like "The Plausible Team love_letter" as the signature name for all emails. that's what we started using recently on newer created emails such as the GA import email. eventually we can replace all signatures with that rather than specific names now that we're a bigger team

Makes sense, let's do it! What do you think about the "Thanks," part in the footer? I don't think it should be there for all emails. In some emails we thank the customer at the end of the actual email (not in the footer). For example:

- "Thanks for understanding and for being a Plausible subscriber!"
- "Thanks again for using our product and for your support!"

So those emails would look like this in the end:

> ...
>  
> Thanks again for using our product and for your support!
>
> Thanks,
The Plausible Team :love_letter:

Perhaps let's do the thanking part in the actual email and keep the footer universal?

metmarkosaric:
> Perhaps let's do the thanking part in the actual email and keep the footer universal?

sounds good to me!

for this one, perhaps something like:

Congrats on the spike in traffic!

ukutaht:
We could replace 'Thanks' with 'Regards' as a universal signature or is that too formal? I can't think of anything with more personality at the moment :)

ukutaht:
Beautiful work @RobertJoonas! This will make our email templates much more maintainable and consistent ❤️

lib/plausible_web/templates/layout/base_email.html.eex
```diff
@@ -0,0 +1,24 @@
+<%= case Map.get(assigns, :user) do
```
ukutaht, line 1: We have a function `user_salutation` in the email_view which I think could be used here?

As a general principle, if you're doing anything more complicated than a simple conditional in Phoenix templates, it's better to extract it into a view function which is pure elixir and much more readable thank the interpolated Elixir in HTML templates.

ukutaht (approved):
Code looks good to me! Approved pending some copy changes from the discussion

metmarkosaric:
> We could replace 'Thanks' with 'Regards' as a universal signature or is that too formal? I can't think of anything with more personality at the moment

regards is fine

lib/plausible_web/templates/layout/base_email.html.eex
RobertJoonas: Agreed! I created something similar to `user_salutation` - `greet_recipient` (commit: https://github.com/plausible/analytics/pull/2431/commits/f9fc0a741e38beeefc90ae9dae7a4fdeedf729ce). This functionality is already tested in `email_test.exs`. 

Removed the `user_salutation` function and its test as those are not needed anymore.

RobertJoonas:
@metmarkosaric

> eventually we can replace all signatures with that rather than specific names now that we're a bigger team

Just following up on this to clarify. This PR changes the signatures of every email, **except**:
- enterprise over limit (internal email)
- password reset email
- weekly report email

The new signature applied to every other email will be this:

"""
Regards,
The Plausible team :love_letter: 
"""

metmarkosaric:
> This PR changes the signatures of every email

ahh ok i didn't understand that. all fine for me!

PR #2444, opened 2022-11-16
Update sites cache implementation

aerosol:
### Changes

This is a follow-up PR to https://github.com/plausible/analytics/pull/2434 and a ground work for rate limiting ingestion, cherry-picked from a work in progress PR to hopefully keep the changes isolated and easier to review.
 
Reviewing commits one by one should give a better picture of how the ideas evolved.

Summary:
  - the tests have been reorganized
  - the `Cache` interface makes it now possible to either refresh the whole cache or re-query a single item once to account for deleted/missing domains (this will be again addressed later in greater detail)
  - warmer responsibility was reduced, tests simplified
  - more consistent/comprehensive telemetry instrumentation

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/cache.ex
+++ b/lib/plausible/site/cache.ex
@@ -54,8 +54,8 @@ defmodule Plausible.Site.Cache do
     )
   end
 
-  @spec prefill(Keyword.t()) :: :ok
-  def prefill(opts) do
+  @spec refresh_all(Keyword.t()) :: :ok
+  def refresh_all(opts) do
     cache_name = Keyword.fetch!(opts, :cache_name)
 
     sites_by_domain_query =
--- a/lib/plausible/site/cache.ex
+++ b/lib/plausible/site/cache.ex
@@ -39,6 +39,8 @@ defmodule Plausible.Site.Cache do
      ingest_rate_limit_threshold
    )a
 
+  @type t() :: Site.t() | Ecto.NoResultsError.t()
+
   def name(), do: @cache_name
 
   @spec child_spec(Keyword.t()) :: Supervisor.child_spec()
@@ -65,6 +67,7 @@ defmodule Plausible.Site.Cache do
 
     sites_by_domain = Plausible.Repo.all(sites_by_domain_query)
 
+    Cachex.clear!(cache_name)
     true = Cachex.put_many!(cache_name, sites_by_domain)
     :ok
   end
@@ -81,7 +84,7 @@ defmodule Plausible.Site.Cache do
     Map.get(stats, :hit_rate, 0)
   end
 
-  @spec get(String.t(), Keyword.t()) :: nil | Site.t()
+  @spec get(String.t(), Keyword.t()) :: t() | nil
   def get(domain, opts \\ []) do
     cache_name = Keyword.get(opts, :cache_name, @cache_name)
     force? = Keyword.get(opts, :force?, false)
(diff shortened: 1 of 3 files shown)


lib/plausible/site/cache.ex
```diff
@@ -19,8 +19,26 @@ defmodule Plausible.Site.Cache do
   database counterpart -- to spare bandwidth and query execution time,
   only selected database columns are retrieved and cached.
 
+  There are two modes of refreshing the cache: `:all` and `:single`.
+
+    * `:all` means querying the database for all Site entries and should be done
+      periodically (via `Cache.Warmer`). All existing Cache entries all cleared 
+      prior to writing the new batch.
+
+    * `:single` attempts to re-query a specific site by domain and should be done
```
ukutaht, line 28: Do I understand the idea correctly?

When an event hits `/api/event`, we attempt to `Cache.get(domain)` which may return `nil`. Since this `nil` could be a newly created site, we attempt to refresh that single entry from the database. If it gets filled with `%Ecto.NoResultsError{}`, then it doesn't get looked up again from the DB until the next cache refresh.

I see how this would work for new sites but not deleted ones... For deleted sites, `Cache.get(domain)` would return the site struct until the next cache refresh period, no? Can't think of any indication that the single entry should be refreshed, other than a PubSub message at the moment.

BTW 15 minutes is probably fine for deleted sites. Especially if we make best effort to notify other nodes via PubSub later on.

lib/plausible/site/cache.ex
```diff
@@ -52,20 +73,27 @@ defmodule Plausible.Site.Cache do
...
+        from s in Site,
+          select: {
+            s.domain,
+            %{struct(s, ^@cached_schema_fields) | from_cache?: true}
+          }
 
-    sites_by_domain = Plausible.Repo.all(sites_by_domain_query)
+      sites_by_domain = Plausible.Repo.all(sites_by_domain_query)
+
+      Cachex.clear!(cache_name)
```
aerosol, line 90: In another iteration I'm planning to look into the transactional capabilities of Cachex. There's a very small window here in which the cache can be completely empty otherwise. Out of scope for this PR.

lib/plausible/site/cache/warmer.ex
```diff
@@ -35,9 +30,11 @@ defmodule Plausible.Site.Cache.Warmer do
 
   @spec child_spec(Keyword.t()) :: Supervisor.child_spec() | :ignore
   def child_spec(opts) do
+    child_name = Keyword.get(opts, :child_name, {:local, __MODULE__})
```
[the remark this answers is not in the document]
aerosol: Out of scope for this PR, trying to keep it minimal :sweat: - I'll have to check if we index timestamps too, so there will be possibly another migration involved.

ukutaht (approved):

lib/plausible/site/cache.ex
ukutaht: Good catch!

lib/plausible/site/cache/warmer.ex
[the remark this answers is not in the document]
ukutaht: > able to re-use it for a more frequent refresh cycle where we'll keep looking for sites most recently added only (DB timestamp within the last 15 minutes)

Oh, I think I've misunderstood below in my comment how new sites will work. This seems like a great idea. As you said as long as there's an index on the timestamp (probably not), then this could be run very frequently. From a UX perspective, I don't thin anyone will get an implementation up and running within 30 seconds of adding a site, for example.

I guess I don't really get what the use-case for the single cache refresh is then.

lib/plausible/site/cache.ex
aerosol: @ukutaht Your understanding is correct here. 

The idea here was to handle events for newly created sites with a minimal delay (see my other comment about refreshing the cache more frequently only for just-in sites) and at the same time act reasonably towards bursts of events when a site has been deleted from our DB but the script.js is still working out there.

PubSub is on my list, but I believe we can make it right for 99% cases without the extra complexity that comes with it (setting up clustering dynamically, handling netsplits and still some latency/throughput side effects of broadcasting those messages). 

I believe more work around deletions will have to be done regardless.

Once Rate Limiter lands with full instrumentation, I plan to enable it gradually with a feature flag so we'll hopefully catch the edge cases without causing too much distress.

ukutaht: Fair. I'll wait for more code to see how it all fits together :)

aerosol: The problem right now is we don't know the difference between a deleted site and a site that's never been created in the first place and that's an important distinction to make, MAYBE. As I said, deletions is going to be another chunk of work I'm not very focused on yet.

vinibrsl (approved):

lib/plausible/site/cache.ex
ukutaht: If we accept a 15-minute window to stop accepting stats from deleted sites, the distinction doesn't matter does it?

PR #2449, opened 2022-11-16
Seed database with pageviews

vinibrsl:
This commit adds basic support for database seeding useful for local testing, especially dashboard changes, like intervals.

It creates two years of pageviews with random timestamps. There is lot of room for improvement, such as adding sources, entry pages, geolocation, devices, custom events, but this already helps us with testing.

Screen Shot 2022-11-16 at 12 56 42

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -47,7 +47,7 @@ defmodule Plausible.MixProject do
   end
 
   # Specifies which paths to compile per environment.
-  defp elixirc_paths(:test), do: ["lib", "test/support"]
+  defp elixirc_paths(env) when env in [:test, :dev], do: ["lib", "test/support"]
   defp elixirc_paths(_), do: ["lib"]
 
   # Specifies your project dependencies.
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -71,9 +71,8 @@ defmodule Plausible.MixProject do
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
       {:double, "~> 0.8.0", only: :test},
       {:ecto_sql, "~> 3.0"},
-      {:elixir_uuid, "~> 1.2", only: :test},
```
vinibrsl, line 74: This was used in tests, and I replaced with `Ecto.UUID.generate/0` which should be the same.

ukutaht:
Looks very useful for local testing. I think we should also add a note about this in the `CONTRIBUTING.md` document.

mix.exs
aerosol: :fire:

aerosol (approved):
Awesome!

vinibrsl:
> Looks very useful for local testing. I think we should also add a note about this in the `CONTRIBUTING.md` document.

Good call!

ukutaht (approved):

PR #1335, opened 2021-09-15
Allow URLs without hostname setting it to (none)

RobertJoonas:
### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

PR #2462, opened 2022-11-22
Rework ingestion pipeline processing

aerosol:
### Changes

This PR contains changes against (on top of) #2351 to address concerns expressed in the originating [thread](https://github.com/plausible/analytics/pull/2351#discussion_r1021580109). With this change applied it should be fairly straight-forward to integrate and test Rate Limiting (ref #2452)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -62,6 +62,7 @@ defmodule Plausible.Ingestion.Event do
       &put_user_agent/1,
       &put_basic_info/1,
       &put_referrer/1,
+      &put_utm_tags/1,
       &put_geolocation/1,
       &put_screen_size/1,
       &put_props/1,
@@ -134,6 +135,14 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_referrer(%__MODULE__{} = event) do
     ref = parse_referrer(event.request.uri, event.request.referrer)
+
+    update_attrs(event, %{
+      referrer_source: get_referrer_source(event.request, ref),
+      referrer: clean_referrer(ref)
+    })
+  end
+
+  defp put_utm_tags(%__MODULE__{} = event) do
     query_params = event.request.query_params
 
     update_attrs(event, %{
@@ -141,9 +150,7 @@ defmodule Plausible.Ingestion.Event do
       utm_source: query_params["utm_source"],
       utm_campaign: query_params["utm_campaign"],
       utm_content: query_params["utm_content"],
-      utm_term: query_params["utm_term"],
-      referrer_source: get_referrer_source(event.request, ref),
-      referrer: clean_referrer(ref)
+      utm_term: query_params["utm_term"]
     })
   end
 
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -117,10 +117,10 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_basic_info(%__MODULE__{} = event) do
     host =
-      if event.request.uri && event.request.uri.host == "" do
-        "(none)"
-      else
-        event.request.uri && event.request.uri.host
+      case event.request.uri do
+        %{host: ""} -> "(none)"
+        %{host: host} when is_binary(host) -> host
+        _ -> nil
       end
 
     update_attrs(event, %{
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -1,57 +1,259 @@
 defmodule Plausible.Ingestion.Event do
-  alias Plausible.Ingestion.{Request, CityOverrides}
-
-  @spec build_and_buffer(Request.t() | [Request.t()]) ::
-          :ok | :skip | {:error, Ecto.Changeset.t()}
-  @doc """
-  Builds events from %Plausible.Ingestion.Request{} and adds them to Plausible.Event.WriteBuffer.
-  This function reads geolocation data and parses the user agent string. Returns :skip if the
-  request is identified as spam, or blocked.
+  @moduledoc """
+  This module exposes the `build_and_buffer/1` function capable of
+  turning %Plausible.Ingestion.Request{} into a series of events that in turn
+  are uniformly either buffered in batches (to Clickhouse) or dropped
+  (e.g. due to spam blocklist) from the processing pipeline.
   """
-  def build_and_buffer(request_or_requests)
+  alias Plausible.Ingestion.{Request, CityOverrides}
+  alias Plausible.ClickhouseEvent
 
-  def build_and_buffer(requests) when is_list(requests) do
-    Enum.reduce_while(requests, :ok, fn %Request{} = request, _acc ->
-      case build_and_buffer(request) do
-        :ok -> {:cont, :ok}
-        error -> {:halt, error}
+  defstruct domain: nil,
+            clickhouse_event_attrs: %{},
+            clickhouse_event: nil,
+            dropped?: false,
+            drop_reason: nil,
+            request: nil,
+            salts: nil
+
+  @type drop_reason() ::
+          :bot
+          | :domain_blocked
+          | :spam_referrer
+          | {:error, Ecto.Changeset.t()}
+
+  @type t() :: %__MODULE__{
+          domain: String.t() | nil,
+          clickhouse_event_attrs: map(),
+          clickhouse_event: %ClickhouseEvent{} | nil,
+          dropped?: boolean(),
+          drop_reason: drop_reason(),
+          request: Request.t(),
+          salts: map()
+        }
+
+  @spec build_and_buffer(Request.t()) ::
+          {:ok, %{dropped: [t()], buffered: [t()]}}
+  def build_and_buffer(%Request{domains: domains} = request) do
+    processed_events =
+      if spam_referrer?(request) do
+        for domain <- domains, do: drop(new(domain, request), :spam_referrer)
+      else
+        Enum.reduce(domains, [], fn domain, acc ->
+          if domain_blocked?(domain) do
+            [drop(new(domain, request), :domain_blocked) | acc]
+          else
+            processed =
+              domain
+              |> new(request)
+              |> process_unless_dropped(pipeline())
+
+            [processed | acc]
+          end
+        end)
+      end
+
+    {dropped, buffered} = Enum.split_with(processed_events, & &1.dropped?)
+    {:ok, %{dropped: dropped, buffered: buffered}}
+  end
+
+  defp pipeline() do
+    [
+      &put_user_agent/1,
+      &put_basic_info/1,
+      &put_referrer/1,
+      &put_geolocation/1,
+      &put_screen_size/1,
+      &put_props/1,
+      &put_salts/1,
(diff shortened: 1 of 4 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -1,57 +1,259 @@
...
+    {dropped, buffered} = Enum.split_with(processed_events, & &1.dropped?)
+    {:ok, %{dropped: dropped, buffered: buffered}}
+  end
 
-  def build_and_buffer(requests) when is_list(requests) do
-    Enum.reduce_while(requests, :ok, fn %Request{} = request, _acc ->
-      case build_and_buffer(request) do
-        :ok -> {:cont, :ok}
-        error -> {:halt, error}
+  defp pipeline() do
```
aerosol, line 60: I'm considering making this definition customizable for test purposes, so that we can test reaching each stage in isolation (and in parallel). 

Something like:

```elixir
# in test
build_and_buffer(
  request, 
  pipeline_overrides: %{
    write_to_buffer: &acknowledge_sent_to_buffer/1
  })
```

I didn't want to make this PR grow out of proportion. For now I'm assuming no existing test regressions is good enough.

vinibrsl: Good call on changing the `with` clause stack to this function pipeline. I think it makes sense because every step can either drop or change the event.

The only concern I have, and it is the same concern I had with the `with` clause, is that some steps in the pipeline are dependent on each other. `register_session/1`, for example, depends on the user agent parsing. Not that this is a problem, but it's not 100% clear who depends on who, and with pipeline skipping this can get a bit more complex. I'm not suggesting a change, just making a note to know what you think about this :)

lib/plausible_web/controllers/api/external_controller.ex
```diff
@@ -11,28 +11,51 @@ defmodule PlausibleWeb.Api.ExternalController do
...
-         :ok <- Ingestion.Event.build_and_buffer(requests) do
-      conn |> put_status(202) |> text("ok")
-    else
-      :skip ->
-        conn |> put_status(202) |> text("ok")
+    case Ingestion.Request.build(conn) do
+      {:ok, request} ->
+        Sentry.Context.set_extra_context(%{request: request})
+
+        case Ingestion.Event.build_and_buffer(request) do
```
vinibrsl, line 18: Have you considered extracting the second `case` in separate function? Or using a `with`/`else` clause?

vinibrsl (approved):

lib/plausible/ingestion/event.ex
aerosol: > register_session/1, for example, depends on the user agent parsing. 

How exactly? I have trouble finding it.

lib/plausible_web/controllers/api/external_controller.ex
aerosol: How about d45f19d0?

lib/plausible/ingestion/event.ex
aerosol: Anyway, depending on how declarative we'd like it to be, we could have more verbose pipeline definition. Or be very strict about pattern matching in each step's function head. Not sure if worth it at this stage though, but happy to make changes if you feel it helps.

lib/plausible_web/controllers/api/external_controller.ex
vinibrsl: Looks good. Thanks

lib/plausible/ingestion/event.ex
vinibrsl: > > register_session/1, for example, depends on the user agent parsing.
> 
> How exactly? I have trouble finding it.

I actually re-read the code, and it uses the raw user agent string, so it's not dependent on the parsing function.

https://github.com/plausible/analytics/blob/8516c859bcb9f2ad026d35b456726b0bd1cf8cd9/lib/plausible/ingestion/event.ex#L384

Sorry, you can ignore my comment.

lib/plausible/ingestion/event.ex
```diff
@@ -1,57 +1,259 @@
...
+          else
+            processed =
+              domain
+              |> new(request)
+              |> process_unless_dropped(pipeline())
+
+            [processed | acc]
+          end
+        end)
+      end
```
aerosol, line 54: @ukutaht I think this should be where we collect grafana metrics for permitted/rejected events tagged with simplified reason label.

lib/plausible/ingestion/event.ex
```diff
@@ -1,57 +1,259 @@
...
+
+  @spec build_and_buffer(Request.t()) ::
+          {:ok, %{dropped: [t()], buffered: [t()]}}
+  def build_and_buffer(%Request{domains: domains} = request) do
+    processed_events =
+      if spam_referrer?(request) do
+        for domain <- domains, do: drop(new(domain, request), :spam_referrer)
+      else
+        Enum.reduce(domains, [], fn domain, acc ->
+          if domain_blocked?(domain) do
```
aerosol, line 43: @ukutaht Should we get rid of static domain blocking when Rate Limiting is integrated? Blocking a domain will require setting the threshold to explicit 0 via CRM and will be a matter of 30 seconds with this upcoming "updated recently" cache refresher

ukutaht: Yes, agreed. Having a single method is better.

lib/plausible/ingestion/event.ex
ukutaht: Sounds good.

lib/plausible/ingestion/event.ex
```diff
@@ -1,57 +1,259 @@
...
+      _any ->
+        event
     end
   end
 
-  defp put_basic_info(%{} = event, %Request{} = request) do
+  defp put_basic_info(%__MODULE__{} = event) do
     host =
-      if request.uri && request.uri.host == "" do
+      if event.request.uri && event.request.uri.host == "" do
```
ukutaht, line 120: I wish we had a conditional chaining operator in Elixir. Something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Some time during this project I learned about [`Kernel.get_in/2`](https://hexdocs.pm/elixir/1.14/Kernel.html#get_in/2) which can be used to access nested fields safely.

```elixir
event.request.uri && event.request.uri.host
```

can be rewritten as

```elixir
get_in(event.request, [:uri, :host])
```

or if `event.request` can also be nil:

```elixir
event.request && event.request.uri && event.request.uri.host
```

can be rewritten as

```elixir
get_in(event, [:request, :uri, :host])
```

I find it slightly easier to scan when reading code than chaining `&&`.

lib/plausible/ingestion/event.ex
```diff
@@ -1,57 +1,259 @@
...
+      hostname: Request.sanitize_hostname(host),
+      pathname: get_pathname(event.request.uri, event.request.hash_mode)
+    })
+  end
+
+  defp put_referrer(%__MODULE__{} = event) do
+    ref = parse_referrer(event.request.uri, event.request.referrer)
+    query_params = event.request.query_params
+
+    update_attrs(event, %{
```
ukutaht, line 139: Not a blocker at all but maybe could be nice to separate `put_referrer` and `put_utm_tags` into separate pipeline stages. Then the utm one could be done with 

```elixir
Map.take(event.request.query_params, [
  "utm_medium", 
  "utm_source", 
  "utm_campaign", 
  "utm_content", 
  "utm_term"
]
```

ukutaht (commented):
Had some optional stylistic comments but overall looks great to me!

lib/plausible/ingestion/event.ex
ukutaht: Or even better `Map.take(event.request.query_params, @utm_tags)`

lib/plausible/ingestion/event.ex
aerosol: Which reminds me, in the OG PR I asked when is host equal to `""` exactly. Isn't the event invalid then?

aerosol: sauce: https://github.com/plausible/analytics/pull/1335

aerosol: So at this point the uri is guaranteed to be `URI.parse/1` result so the check is excessive and can be rewritten as e.g.:

```elixir
    host = event.request.uri.host
    if host in [nil, ""], do: "(none)", else: host
```

aerosol: Wait, no, the validation is run too late :facepalm:

aerosol: and event.request is a struct so has to implement Access behavior to make get_in work :clown_face:

aerosol: Best I can do right now

```elixir
    host =
      case event.request.uri do
        %{host: ""} -> "(none)"
        %{host: host} when is_binary(host) -> host
        _ -> nil
      end
```

lib/plausible/ingestion/event.ex
aerosol: `Map.take` won't work, we're fetching stringified keys from `query_params` but need atoms for the `attrs`

aerosol: I've separated them: e7d0d714

lib/plausible/ingestion/event.ex
ukutaht: Ah, the Access protocol... right. Probably not worth it then.

> Which reminds me, in the OG PR I asked when is host equal to "" exactly. Isn't the event invalid then?

No, Plausible is sometimes used outside of the WWW context. The incoming `url` parameter does need to conform to the URI spec but people can put anything in there, including leaving the hostname blank. 
[https://github.com/RobertJoonas/analytics/blob/ee68e659df92195cee5f3512cbae10f4da3c3dcb/test/plausible_web/controllers/api/external_controller_test.exs#L696-L712](Test case) that shows an example from a mobile app.

lib/plausible/ingestion/event.ex
ukutaht: > Map.take won't work, we're fetching stringified keys from query_params but need atoms for the attrs

Ah you're right, of course.

lib/plausible/ingestion/event.ex
ukutaht: > Best I can do right now
> 
> ```elixir
>     host =
>       case event.request.uri do
>         %{host: ""} -> "(none)"
>         %{host: host} when is_binary(host) -> host
>         _ -> nil
>       end
> ```

Looks good. One thing this reminds me: I've wanted to replace the current Ecto validations and instead run validations as early in the pipeline as possible. Currently we process the whole event and if the end result is invalid, drop it. But we don't need to event start processing events that are obviously invalid from the beginning, or that turn out to be invalid some time in the processing pipeline. Like in this case, if the `uri` is `nil`, it means the event is invalid and does not need to be processed any further.

Not in scope for this PR but something I wanted to flag. The shiny new pipeline seems flexible enough to allow each step to return `(drop, :validation_error_key)` or something like this.

aerosol: Agreed :+1:

PR #2463, opened 2022-11-22
Fix target container is not a DOM element

vinibrsl:
This commit fixes a bug where opening modals resulted in a `target container is not a DOM element` error. This was caused by a change in #1574, that was not closing the div that mounts React as expected.


RobertJoonas (approved):
Nice quick fix @vinibrsl, thanks!

PR #2470, opened 2022-11-24
Add keyboard navigation to intervals

vinibrsl:
### Changes

This pull request adds keyboard navigation to intervals. Pressing `<i>` will open the interval menu, and pressing `<i>` again will close it. You can navigate with arrows and press `<Enter>` to select an interval.

out

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/datepicker.js
+++ b/assets/js/dashboard/datepicker.js
@@ -22,6 +22,7 @@ import {
   isAfter,
 } from "./util/date";
 import { navigateToQuery, QueryLink, QueryButton } from "./query";
+import { shouldIgnoreKeypress } from "./keybinding.js"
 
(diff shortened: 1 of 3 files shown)


metmarkosaric:
nice!

ukutaht:
Looks good to me!

Was thinking of a way to make it more discoverable but couldn't come up with anything nice.

ukutaht (approved):

vinibrsl:
> Looks good to me!
> 
> Was thinking of a way to make it more discoverable but couldn't come up with anything nice.

I thought of adding a tooltip, but I don't think many people would see it anyways. I like the idea of having a `<?>` keybinding that lists every shortcut. Mastodon has something like this:

image

PR #2472, opened 2022-11-24
Gatekeep ingestion pipeline

aerosol:
### Changes

This PR introduces event ingestion pipeline protection and instrumentation. 

Apologies for the sizeable chunk of code in one go; reviewing commit by commit should help with illustrating the thought process as I proceeded.

The event processing logic has been reworked, so that we return any client request errors early if possible. When the initial `Request` validation passes, it's turned into a series of `Ingestion.Event`s (to handle multiple domains uniformly) that are either buffered to ClickHouse or dropped with a reason. We keep telemetry for both cases (counters).

The new behavior from backend perspective is we'll now any drop traffic for sites that are not found in the database/cache.

A new `Site.Cache` warmer is now spawned, refreshing any recently updated/created sites every 30 seconds (ref #2467 - the migration has been applied on prod). Both cache refresh modes are also instrumented with prometheus metrics (histograms).

The ingestion pipeline also checks for any rate-limits, optionally imposed via CRM, in which any limits applied are also made visible in the site index page.

image

This PR also hard-deprecates the [DOMAIN_BLACKLIST](https://github.com/plausible/analytics/pull/2472/commits/4a9b8a29d34111aa20cbcb2d9b247fd40dacab49) setting, leveraging CRM threshold/scale settings. Similarly, the `block_event_ingest` feature flag is no longer used. On prod, appropriate sites have been updated with ingest threshold=0 to keep the blocking effective. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -158,11 +158,6 @@ log_level =
   |> get_var_from_path_or_env("LOG_LEVEL", "warn")
   |> String.to_existing_atom()
 
-domain_blacklist =
-  config_dir
-  |> get_var_from_path_or_env("DOMAIN_BLACKLIST", "")
-  |> String.split(",")
-
 is_selfhost =
   config_dir
   |> get_var_from_path_or_env("SELFHOST", "true")
@@ -195,8 +190,7 @@ config :plausible,
   site_limit: site_limit,
   site_limit_exempt: site_limit_exempt,
   is_selfhost: is_selfhost,
-  custom_script_name: custom_script_name,
-  domain_blacklist: domain_blacklist
+  custom_script_name: custom_script_name
 
 config :plausible, :selfhost,
   enable_email_verification: enable_email_verification,
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -23,6 +23,7 @@ defmodule Plausible.Application do
       ),
       {Plausible.Site.Cache, []},
       {Plausible.Site.Cache.Warmer, []},
+      {Plausible.Site.Cache.Warmer.RecentlyUpdated, []},
       PlausibleWeb.Endpoint,
       {Oban, Application.get_env(:plausible, Oban)},
       Plausible.PromEx
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -21,12 +21,12 @@ defmodule Plausible.Application do
       Supervisor.child_spec({Cachex, name: :sessions, limit: nil, stats: true},
         id: :cachex_sessions
       ),
+      Plausible.PromEx,
       {Plausible.Site.Cache, []},
       {Plausible.Site.Cache.Warmer.All, []},
       {Plausible.Site.Cache.Warmer.RecentlyUpdated, []},
       PlausibleWeb.Endpoint,
-      {Oban, Application.get_env(:plausible, Oban)},
-      Plausible.PromEx
+      {Oban, Application.get_env(:plausible, Oban)}
     ]
 
     opts = [strategy: :one_for_one, name: Plausible.Supervisor]
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -22,7 +22,7 @@ defmodule Plausible.Application do
         id: :cachex_sessions
       ),
       {Plausible.Site.Cache, []},
-      {Plausible.Site.Cache.Warmer, []},
+      {Plausible.Site.Cache.Warmer.All, []},
       {Plausible.Site.Cache.Warmer.RecentlyUpdated, []},
       PlausibleWeb.Endpoint,
       {Oban, Application.get_env(:plausible, Oban)},
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -8,7 +8,7 @@ defmodule Plausible.Ingestion.Request do
   use Ecto.Schema
   alias Ecto.Changeset
 
-  schema "" do
+  embedded_schema do
     field :remote_ip, :string
     field :user_agent, :string
     field :event_name, :string
@@ -37,28 +37,22 @@ defmodule Plausible.Ingestion.Request do
 
     case parse_body(conn) do
       {:ok, request_body} ->
-        changeset =
-          changeset
-          |> put_remote_ip(conn)
-          |> put_uri(request_body)
-          |> put_hostname()
-          |> put_user_agent(conn)
-          |> put_request_params(request_body)
-          |> put_pathname()
-          |> put_query_params()
-          |> map_domains(request_body)
-          |> Changeset.validate_required([
-            :event_name,
-            :hostname,
-            :pathname,
-            :timestamp
-          ])
-
-        if changeset.valid? do
-          {:ok, Changeset.apply_changes(changeset)}
-        else
-          {:error, changeset}
-        end
+        changeset
+        |> put_remote_ip(conn)
+        |> put_uri(request_body)
+        |> put_hostname()
+        |> put_user_agent(conn)
+        |> put_request_params(request_body)
+        |> put_pathname()
+        |> put_query_params()
+        |> map_domains(request_body)
+        |> Changeset.validate_required([
+          :event_name,
+          :hostname,
+          :pathname,
+          :timestamp
+        ])
+        |> Changeset.apply_action(nil)
 
       {:error, :invalid_json} ->
         {:error, Changeset.add_error(changeset, :request, "Unable to parse request body as json")}
@@ -96,7 +90,9 @@ defmodule Plausible.Ingestion.Request do
   end
 
   defp put_pathname(changeset) do
-    pathname = get_pathname(changeset.changes[:uri], changeset.changes[:hash_mode])
+    uri = Changeset.get_field(changeset, :uri)
+    hash_mode = Changeset.get_field(changeset, :hash_mode)
+    pathname = get_pathname(uri, hash_mode)
     Changeset.put_change(changeset, :pathname, pathname)
   end
 
@@ -117,7 +113,7 @@ defmodule Plausible.Ingestion.Request do
         Changeset.put_change(changeset, :domains, domains)
 
       nil ->
-        from_uri = sanitize_hostname(changeset.changes[:uri])
+        from_uri = sanitize_hostname(Changeset.get_field(changeset, :uri))
 
         if from_uri do
           Changeset.put_change(changeset, :domains, [from_uri])
@@ -144,7 +140,7 @@ defmodule Plausible.Ingestion.Request do
 
   defp put_hostname(changeset) do
     host =
-      case changeset.changes[:uri] do
+      case Changeset.get_field(changeset, :uri) do
         %{host: host} when is_binary(host) and host != "" -> host
         _ -> "(none)"
       end
@@ -192,7 +188,7 @@ defmodule Plausible.Ingestion.Request do
   end
 
   defp put_query_params(changeset) do
-    case changeset.changes[:uri] do
+    case Changeset.get_field(changeset, :uri) do
       %{query: query} when is_binary(query) ->
         Changeset.put_change(changeset, :query_params, URI.decode_query(query))
 
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -128,19 +128,25 @@ defmodule Plausible.Ingestion.Request do
   end
 
   defp put_uri(changeset, %{} = request_body) do
-    if url = request_body["u"] || request_body["url"] do
-      Changeset.put_change(changeset, :uri, URI.parse(url))
-    else
-      changeset
+    url = request_body["u"] || request_body["url"]
+
+    case url do
+      nil ->
+        Changeset.add_error(changeset, :url, "is required")
+
+      url when is_binary(url) ->
+        Changeset.put_change(changeset, :uri, URI.parse(url))
+
+      _ ->
+        Changeset.add_error(changeset, :url, "must be a string")
     end
   end
 
   defp put_hostname(changeset) do
     host =
       case changeset.changes[:uri] do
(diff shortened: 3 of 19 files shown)


lib/plausible/ingestion/request.ex
```diff
@@ -1,51 +1,74 @@
...
-          referrer: term(),
-          domains: list(String.t()),
-          screen_width: term(),
-          hash_mode: term(),
-          props: map(),
-          query_params: map()
-        }
-
-  @spec build(Plug.Conn.t()) :: {:ok, t()} | {:error, :invalid_json}
+  use Ecto.Schema
```
ukutaht, line 8: Nice, good call making it `Ecto.Schema`

lib/plausible/ingestion/request.ex
```diff
@@ -61,38 +84,68 @@ defmodule Plausible.Ingestion.Request do
...
+      changeset
     end
   end
 
+  defp put_hostname(changeset) do
+    host =
+      case changeset.changes[:uri] do
+        %{host: empty} when empty in ["", nil] -> "(none)"
+        %{host: host} when is_binary(host) -> host
+        _ -> nil
```
ukutaht, line 143: This old implementation by me seems work in a quite indirect and confusing way:
1. `URI.parse` will always return a non-nil value if input is a binary (according to both typespec and [implementation](https://github.com/elixir-lang/elixir/blob/a64d42f5d3cb6c32752af9d3312897e8cd5bb7ec/lib/elixir/lib/uri.ex#L531))
2. The only reason this case branch would run is if `params["url"] || params["u"]` is missing or a non-binary value
3. In that case, hostname gets `nil` value
4. Validation fails on the hostname field because it's required.

I think a better approach would be to validate the URL early and add more specific errors:
1. If url parameter is missing, add `Changeset.add_error(changeset, :url, "is required")`
2. If url is parameter is non-binary, add `Changeset.add_error(changeset, :url, "must be a string")`
3. Always default hostname to `"(none)"` on the line I'm commenting on

lib/plausible_web/controllers/api/external_controller.ex
```diff
@@ -35,10 +35,10 @@ defmodule PlausibleWeb.Api.ExternalController do
           end
       end
     else
-      {:error, :invalid_json} ->
+      {:error, %Ecto.Changeset{} = changeset} ->
         conn
         |> put_status(400)
-        |> json(%{errors: %{request: "Unable to parse request body as json"}})
+        |> json(%{errors: traverse_errors(changeset)})
```
ukutaht, line 41: Does this change the public API by returning the specific validation errors as JSON? If so, I think that's a good thing and we should wrap in in a test so we don't lose it.

lib/plausible/site/gate_keeper.ex
```diff
@@ -1,25 +1,39 @@
...
+  @block :block
+  @deny :deny
+  @throttle :throttle
+
+  @type policy() :: :allow | :deny | :block | :throttle
+
+  @policy_for_non_existing_sites @deny
+  @policy_on_rate_limiting_backend_error @allow
+
+  defstruct passed?: false,
```
ukutaht, line 12: I feel that an API like

```elixir
> GateKeeper.check(domain)
:allow
> GateKeeper.check(domain)
{:deny, :rate_limit}
> GateKeeper.check(domain)
{:deny, :site_does_not_exist}
```

Would be clearer, at least for me. I understand it's not as extensible but at least for now much quicker to scan, get what it's doing and why. What do you think?

We're in control of the whole codebase with a small team and this is not part of any public API. So I tend to prefer for ease of readability > extensibility.

lib/plausible/site/admin.ex
```diff
@@ -40,7 +40,16 @@ defmodule Plausible.SiteAdmin do
       timezone: nil,
       public: nil,
       owner: %{value: &get_owner_email/1},
-      other_members: %{value: &get_other_members/1}
+      other_members: %{value: &get_other_members/1},
+      limits: %{
```
ukutaht, line 44: Nice, that looks good.

Kaffy is not great. It's also stopping us from upgrading phoenix_html at the moment unless they've recently updated deps.

I think doing it all with custom code would be better tbh. I brought Kaffy in mostly to get a default UI and not have to worry about it. But we have also paid for [TailwindUI](https://tailwindui.com/components) which now has components for most admin-related things we would ever need. At some point writing a custom admin panel using Tailwind UI would be a cool project.

config/runtime.exs
```diff
@@ -158,11 +158,6 @@ log_level =
   |> get_var_from_path_or_env("LOG_LEVEL", "warn")
   |> String.to_existing_atom()
 
-domain_blacklist =
```
ukutaht, line 161: Yay!

lib/plausible_web/controllers/api/external_controller.ex
aerosol: it's tested and backwards-compatible. The only public facing change is we don't return confusing /from API consumer's perspective/ error messages anymore such as "user_id can't be blank".

ukutaht: Sweet

lib/plausible/site/cache.ex
```diff
@@ -99,18 +108,20 @@ defmodule Plausible.Site.Cache do
...
+      new = MapSet.new(Enum.into(new_items, [], fn {k, _} -> k end))
+      old = MapSet.new(old_keys)
 
-    old
-    |> MapSet.difference(new)
-    |> Enum.each(fn k ->
-      Cachex.del(cache_name, k)
-    end)
+      old
+      |> MapSet.difference(new)
```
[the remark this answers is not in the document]
aerosol: > I don't like that we have to manually clean this

Why not?

lib/plausible/application.ex
```diff
@@ -23,6 +23,7 @@ defmodule Plausible.Application do
       ),
       {Plausible.Site.Cache, []},
       {Plausible.Site.Cache.Warmer, []},
```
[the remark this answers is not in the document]
aerosol: fb1636e6 I did this, because we share child specs function with `Warmer.RecentlyUpdated`, so it'd look more confusing if we were calling `Warmer.All` there.

lib/plausible/site/gate_keeper.ex
aerosol: 4e8d7195

lib/plausible/site/admin.ex
aerosol: Do you feel we can go live with this before the CRM is fully customized?

lib/plausible/site/cache.ex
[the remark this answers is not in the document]
ukutaht: To use a technical term it seems.. fiddly? 😀

Replacing old cache contents with new sounds really simple in theory but I get why the code cannot be as direct and straight-forward as it seems at first. Transactional guarantees are not easy. We don't need to focus on this at all. I'm happy you caught the edge case and addressed it.

lib/plausible/site/admin.ex
ukutaht: yes, totally. 'At some point' in my comment above is far in the future.

lib/plausible/site/cache.ex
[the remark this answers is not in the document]
aerosol: OK cool, tbh it's tested and it's reasonably fast for what we need, so I'm not too worried.

lib/plausible/application.ex
```diff
@@ -21,12 +21,12 @@ defmodule Plausible.Application do
       Supervisor.child_spec({Cachex, name: :sessions, limit: nil, stats: true},
         id: :cachex_sessions
       ),
+      Plausible.PromEx,
```
aerosol, line 24: Making PromEx earlier in the tree so that periodic measurements are gathered at first probe

lib/plausible/site/gate_keeper.ex
```diff
@@ -46,11 +43,6 @@ defmodule Plausible.Site.GateKeeper do
     "ingest:site:#{domain}"
   end
 
-  @spec policy_telemetry_event(policy()) :: list(atom())
```
aerosol: All the instrumentation happens uniformly at the event processing pipeline now (so we can also gather drop reasons currently unrelated to GateKeeper, such as "spam")

lib/plausible/ingestion/request.ex
aerosol: OK, thanks for this breakdown, I had indeed trouble reasoning about this. Your proposed change seems to work fine with a minor caveat - we no longer return "hostname can't be blank" error, instead "url is required" will be used. Correct me if I'm wrong but that's better - since the API never accepted `hostname` from user input anyway. See: d267f937

lib/plausible/ingestion/request.ex
```diff
@@ -1,51 +1,74 @@
...
-          hash_mode: term(),
-          props: map(),
-          query_params: map()
-        }
-
-  @spec build(Plug.Conn.t()) :: {:ok, t()} | {:error, :invalid_json}
+  use Ecto.Schema
+  alias Ecto.Changeset
+
+  schema "" do
```
vinibrsl, line 11: Good call on making this an `Ecto.Schema`! The first time I saw Ecto being used for validating in-memory data was in Hubert Łępikci's [(Ab)using Ecto as Elixir data casting and validation library](https://www.amberbit.com/blog/2017/12/27/ecto-as-elixir-data-casting-and-validation-library/) amazing blog post.

I think `embedded_schema/1` is a better candidate for this case. I use `schema/2` for persisted data, and `embedded_schema/1` for embedded JSON and in-memory schemas. Good thing we can remove that empty table name string too. Have you considered using `embedded_schema/1`?

> On the other hand, [embedded_schema/1](https://hexdocs.pm/ecto/Ecto.Schema.html#embedded_schema/1) is used for defining schemas that are embedded in other schemas or only exist in-memory. For example, you can use such schemas to receive data from a command line interface and validate it, without ever persisting it elsewhere. Such structs do not contain a __meta__ field, as they are never persisted. https://hexdocs.pm/ecto/Ecto.Schema.html

lib/plausible/ingestion/request.ex
```diff
@@ -1,51 +1,74 @@
...
+            :hostname,
+            :pathname,
+            :timestamp
+          ])
+
+        if changeset.valid? do
+          {:ok, Changeset.apply_changes(changeset)}
+        else
+          {:error, changeset}
+        end
```
vinibrsl, line 61: Nitpick: [`Ecto.Changeset.apply_action/2`](https://hexdocs.pm/ecto/Ecto.Changeset.html#apply_action/2) can be your friend. It returns a tuple with the applied data or the changeset in case of validation errors.

```elixir
Ecto.Changeeset.apply_action(valid_changeset,  nil) 
# => {:ok, %{name: "Adam"}}

Ecto.Changeeset.apply_action(invalid_changeset,  nil) 
# => {:error, Ecto.Changeset.t()}
```

lib/plausible/ingestion/request.ex
```diff
@@ -61,38 +84,74 @@ defmodule Plausible.Ingestion.Request do
...
+      raw when is_binary(raw) ->
+        domains =
+          raw
+          |> String.split(",")
+          |> Enum.map(&sanitize_hostname/1)
+
+        Changeset.put_change(changeset, :domains, domains)
+
+      nil ->
+        from_uri = sanitize_hostname(changeset.changes[:uri])
```
vinibrsl, line 120: We don't load any data into this changeset, we start from a clean slate and use `put_change/3` later on. For this reason `changeset.changes` always has all the fields.

Still this clean slate has default values, e.g. timestamp. I don't think default values are included in the changes map. I suggest using [`Ecto.Changeset.get_field/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#get_field/3) for almost every case. This function looks at the changes and then falls back on the data, which might be useful as we have schema defaults.

vinibrsl, line 120: We don't load any data into this changeset, we start from a clean slate and use `put_change/3` later on. For this reason `changeset.changes` always has all the fields.

Still this clean slate has default values, e.g. timestamp. I don't think default values are included in the changes map. I suggest using [`Ecto.Changeset.get_field/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#get_field/3) for almost every case. This function looks at the changes and then falls back on the data, which might be useful as we have schema defaults.

```suggestion
        from_uri =
          changeset
          |> Ecto.Changeset.get_field(:uri)
          |> sanitize_hostname()
```

I'd suggest the same for every `changeset.changes` call.

lib/plausible/ingestion/request.ex
ukutaht: Yes, the "url is required" error is a much more helpful error message than "hostname can't be blank"

ukutaht (approved):

lib/plausible/ingestion/request.ex
aerosol: good call, will do, cheers

lib/plausible/ingestion/request.ex
aerosol: Right, it's just the `nil` action makes me pause a little. I don't mind, will apply your suggestion :+1:

lib/plausible/ingestion/request.ex
aerosol: Cheers, will do!

aerosol: 3163b68b

lib/plausible/ingestion/request.ex
aerosol: 3163b68b

vinibrsl (approved):

lib/plausible/site/gate_keeper.ex
ukutaht: Yeah good point :)

PR #2479, opened 2022-11-29
Fix: application startup, event timestamps

aerosol:
### Changes

This PR:
  - pushes PromEx to the bottom of supervision stack to avoid Endpoint instrumentation failure
  - ensures the site cache is ready by exposing it through the health check endpoint
  - fixes event timestamps being calculated at compile time, with regression unit and integration tests

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/cache.ex
+++ b/lib/plausible/site/cache.ex
@@ -55,6 +55,7 @@ defmodule Plausible.Site.Cache do
 
   @type t() :: Site.t()
 
+  @spec name() :: atom()
   def name(), do: @cache_name
 
   @spec child_spec(Keyword.t()) :: Supervisor.child_spec()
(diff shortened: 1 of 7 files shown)


lib/plausible/site/cache.ex
```diff
@@ -68,6 +69,22 @@ defmodule Plausible.Site.Cache do
...
   end
 
+  @doc """
+  Ensures the cache has non-zero size unless no sites exist.
+  Useful for orchestrating app startup to prevent the service
+  going up asynchronously with an empty cache.
+  """
+  @spec ready?(atom()) :: boolean
+  def ready?(cache_name \\ @cache_name) do
+    case size(cache_name) do
```
ukutaht, line 79: `Sites.Cache.size/1` can fail because it [asserts a match](https://github.com/plausible/analytics/blob/0ceadbeffa2626c9c85cc8dece61a162044b053e/lib/plausible/site/cache.ex#L148) on `{:ok, size}`

I guess that's fine...? Let it crash and all that. If Cachex itself isn't ready yet then I suppose `/api/health`  will return 500 due to a `MatchError` in the `size/1` function but recover fine once Cachex itself is ready. Just confirming that this is how you expect it to behave as well

ukutaht (approved):

lib/plausible/site/cache.ex
aerosol: Yep. I thought of that scenario. There's almost no chance for this to happen, the Cachex process alone is started very early in relation to the Endpoint. And if it somehow does that, let it crash indeed.

PR #2480, opened 2022-11-30
Trim the value in case it contains newlines

spapas:
### Changes

Hello friends, I'm trying to self-host plausible on my server and have observed that if I add a file named (for example) `secrets/LISTEN_IP` to configure the bind IP I'll get the following error:

```
Compiling 1 file (.ex)
** (RuntimeError) Invalid LISTEN_IP '172.19.130.83
' error: :einval
    /home/serafeim/plausible/config/runtime.exs:20: (file)
```

(notice that I get the same error when configuring the DATABASE_URL or whatever comes first)

It seems that for whatever reason the File.read! will insert a newline at the end of the file. Notice that I have double checked the file with vim and *it does not contain a newline at the end* or at least I don't know *how* to make the file not have a newline at the end.

Thus I'm adding a `String.trim()` there to fix that thing; even if I am doing something wrong and my files *do* contain a newline at the end, adding a `String.trim()` there definite won't hurt :)

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (approved):
Thanks @spapas! Good addition ❤️ Files do often contain newlines at the end as well so it makes sense to trim these.

Could you also please add a note to the changelog under `FIXED`?

spapas:
Done, thank you!

PR #2486, opened 2022-12-01
Use fetch with keepalive when available

sandstrom:
### Changes

For link tracking (sending an event when clicking a link), it's useful if immediate navigation is allowed, while still having the event sent.

```js
// example code; real usage would be slightly different
window.addEventListner('click', (event) => {
  if (event.target.matches('a[href]')) {
    plausibleTracker.trackEvent('Outbound Link: Click', { props: { url: event.target.href });
  }
});
```

The `keepalive` flag in fetch [does just that](https://css-tricks.com/send-an-http-request-on-page-exit/).

I don't know if you consider fetch [available enough](https://caniuse.com/fetch) to ditch XHR, but for now I kept XHR as a fallback choice.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -98,16 +98,32 @@
     payload.h = 1
     {{/if}}
 
-    var request = new XMLHttpRequest();
-    request.open('POST', endpoint, true);
-    request.setRequestHeader('Content-Type', 'text/plain');
+    if (window.fetch) {
+      // if fetch is available, use it because the 'keepalive' flag
+      // will ensure delivery even if the user navigates away
+      window.fetch(endpoint, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(payload),
+        keepalive: true,
+      }).then(() => {
+        options && options.callback && options.callback();
+      });
+    } else {
+      // fallback to XMLHttpRequest
+      var request = new XMLHttpRequest();
+      request.open('POST', endpoint, true);
+      request.setRequestHeader('Content-Type', 'text/plain');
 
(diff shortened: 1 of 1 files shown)


ukutaht:
Hey 👋 thanks for the suggestion @sandstrom.

We are aware of the problems with the XMLHTTPRequest API. We are planning to work on evaluating `sendBeacon` and fetch with keepalive in order to remove the need to wait for requests to complete before unloading the page (like navigation).

Unofortunately there are many complexities with these APIst that need to be considered holistically: https://xgwang.me/posts/you-may-not-know-beacon/

1. Fetch with keepalive is [not supported](https://caniuse.com/mdn-api_request_keepalive) on all modern browsers (e.g. Firefox)
2. SendBeacon is [better supported](https://caniuse.com/beacon) but blocked by uBlock origin and friends.

Changing how the underlying transport mechanism works also implies many changes to the script extensions we've build (outbound links and tagged-events) so they don't wait for the response before continuing the default behaviour.

We also have to consider the script size: adding a bunch of fallbacks is a good idea for backwards compatibility but will balloon up the size of the script which we are targeting to stay under 1KB.

To roll this out, we have to test it extensively since we are handling billions of pageviews per month from tens of thousands of websites. We don't want to create big regressions where suddenly we have a nicer user experience on some browsers but events stop coming altogether from other browsers.

For these reasons, given that the scope of work is huge and will need heavy involvement from the core development team, I'm going to reject this PR for now until we have time to address this properly.

Every PR requires some work for testing and integrations from the core team. I would recommend discussing and getting some alignment from the team before opening a PR, otherwise it's likely that it will get rejected.

PR #2505, opened 2022-12-06
Ga better errors

vinibrsl:
### Changes

This pull request adds error handling to GA authentication errors. These errors were previously resulting in unhandled exceptions. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -561,6 +561,39 @@ defmodule PlausibleWeb.AuthController do
     |> redirect(to: redirect_to)
   end
 
+  def google_auth_callback(conn, %{"error" => error, "state" => state} = params) do
+    [site_id, _redirect_to] = Jason.decode!(state)
+    site = Repo.get(Plausible.Site, site_id)
+
+    case error do
+      "access_denied" ->
+        conn
+        |> put_flash(
+          :error,
+          "We were unable to authenticate your Google Analytics account. Please check that you have granted us permission to 'See and download your Google Analytics data' and try again."
+        )
+        |> redirect(to: Routes.site_path(conn, :settings_general, site.domain))
+
+      message when message in ["server_error", "temporarily_unavailable"] ->
+        conn
+        |> put_flash(
+          :error,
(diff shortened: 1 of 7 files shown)


lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -561,6 +561,39 @@ defmodule PlausibleWeb.AuthController do
...
+          :error,
+          "We were unable to authenticate your Google Analytics account. Please check that you have granted us permission to 'See and download your Google Analytics data' and try again."
+        )
+        |> redirect(to: Routes.site_path(conn, :settings_general, site.domain))
+
+      message when message in ["server_error", "temporarily_unavailable"] ->
+        conn
+        |> put_flash(
+          :error,
+          "We are unable to authenticate your Google Analytics account at the moment. The authentication service is currently unavailable. Please try again in a few moments."
```
ukutaht, line 581:
```suggestion
          "We are unable to authenticate your Google Analytics account because Google's authentication service is temporarily unavailable. Please try again in a few moments."
```

Trying to clarify that:
1. The issue is on Google's side, not us
2. Should be a temporary thing, trying again should work

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -715,7 +731,7 @@ defmodule PlausibleWeb.SiteController do
     case start_date do
       {:ok, nil} ->
         site = conn.assigns[:site]
-        view_ids = Plausible.Google.Api.list_views(access_token)
+        {:ok, view_ids} = Plausible.Google.Api.list_views(access_token)
```
ukutaht, line 734: Is the idea here that authentication errors are caught earlier in the flow so at this point it's not expected to get any sort of auth error?

vinibrsl: Exactly. We could handle them here too, but I haven't seem Sentry errors at this point.

lib/plausible_web/controllers/auth_controller.ex
vinibrsl: Good one!

ukutaht (approved):

PR #2445, opened 2022-11-16
Realtime dashboard improvements

RobertJoonas:
### Changes

This PR aims to start automatically updating all visible dashboard reports in the realtime view, fixing https://github.com/plausible/analytics/issues/1589. It also fixes the faulty behavior that we subscribe for updates, but do not unsubscribe from those updates when report components unmount.

This PR introduces a new timer object that dispatches a 'tick' browser event that reports can follow individually. That makes updates possible without passing the timer object around. Every time a dashboard report is mounted, it starts listening for the 'tick' events to fetch new data.  As soon as the component unmounts, we will unsubscribe from updates to not make unnecessary requests.

### Tests
- [x] Writing unit tests for the React dashboard is currently out of scope

### Changelog
- [x] Entry has been added to changelog

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #2532, opened 2022-12-19
Visual improvements to automatic realtime updates

RobertJoonas:
### Changes

1. Use the `react-flip-move` library to make updates to `ListReports` look nicer.
2. Unify the color of 'No data yet' in darkmode.

Screencasts to showcase 1 and 2: [before](https://user-images.githubusercontent.com/56999674/208523855-ad1cecd8-3947-4228-b0db-f34b3ce514c6.mp4) -> [after](https://user-images.githubusercontent.com/56999674/208524152-9cbe41e6-ee39-40f4-8880-88961e1c7f8b.mp4) (the "Before" screencast already includes the auto-updating work done in https://github.com/plausible/analytics/pull/2445)

3. Lose the flicker in the graph when updated in the realtime mode. A console log demonstrates the moment when new data is fetched. Screencasts: [before](https://user-images.githubusercontent.com/56999674/208525401-9c7e8b99-d099-441e-bd48-3b56d4f3d49f.mp4) -> [after](https://user-images.githubusercontent.com/56999674/208525521-9205f897-a411-4434-9d8b-ff5df109729d.mp4) (before = master)

4. Fix a bug where graph data was requested twice upon changing the interval. Screencasts: [before](https://user-images.githubusercontent.com/56999674/208525808-699de926-3c8b-453c-9f4a-2d4645fea958.mp4) -> [after](
https://user-images.githubusercontent.com/56999674/208525873-733a4674-239e-4e94-994c-5499767c5bc4.mp4) (before = master)

5. Some refactoring commits to `visitor-graph.js`

### Tests
- [x] Writing unit tests for the React dashboard is currently out of scope

### Changelog
- [x] https://github.com/plausible/analytics/pull/2445 updates changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -395,7 +395,7 @@ export default class VisitorGraph extends React.Component {
   }
 
   updateMetric(clickedMetric) {
-    const newMetric = clickedMetric !== this.state.metric ? clickedMetric : ""
+    const newMetric = clickedMetric === this.state.metric ? "" : clickedMetric
 
     storage.setItem(`metric__${this.props.site.domain}`, newMetric)
     this.setState({ metric: newMetric })
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -362,10 +362,10 @@ export default class VisitorGraph extends React.Component {
     const { metric, topStatData } = this.state;
 
     if (this.props.query !== prevProps.query) {
-      if (metric) {
-        this.setState({ mainGraphLoadingState: LOADING_STATE.loading, topStatsLoadingState: LOADING_STATE.loading, graphData: null, topStatData: null }, this.maybeRollbackInterval)
-      } else {
+      if (this.isGraphCollapsed()) {
         this.setState({ topStatsLoadingState: LOADING_STATE.loading, topStatData: null })
+      } else {
+        this.setState({ mainGraphLoadingState: LOADING_STATE.loading, topStatsLoadingState: LOADING_STATE.loading, graphData: null, topStatData: null }, this.maybeRollbackInterval)
       }
       this.fetchTopStatData()
     }
@@ -388,6 +388,10 @@ export default class VisitorGraph extends React.Component {
     }
   }
 
+  isGraphCollapsed() {
+    return !this.state.metric
+  }
+
   componentWillUnmount() {
     document.removeEventListener('tick', this.maybeRollbackInterval)
     document.removeEventListener('tick', this.fetchTopStatData)
@@ -401,13 +405,13 @@ export default class VisitorGraph extends React.Component {
   }
 
   fetchGraphData() {
-    if (!this.state.metric) {
+    if (this.isGraphCollapsed()) {
       this.setState({ mainGraphLoadingState: LOADING_STATE.loaded, graphData: null })
       return
     }
 
     const url = `/api/stats/${encodeURIComponent(this.props.site.domain)}/main-graph`
-    let params = {metric: this.state.metric || 'none'}
+    let params = {metric: this.state.metric}
     if (this.state.interval) { params.interval = this.state.interval }
 
     api.get(url, this.props.query, params)
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -360,8 +360,9 @@ export default class VisitorGraph extends React.Component {
 
   componentDidUpdate(prevProps, prevState) {
     const { metric, topStatData } = this.state;
+    const { query, site } = this.props
 
-    if (this.props.query !== prevProps.query) {
+    if (query !== prevProps.query) {
       if (this.isGraphCollapsed()) {
         this.setState({ topStatsLoadingState: LOADING_STATE.loading, topStatData: null })
       } else {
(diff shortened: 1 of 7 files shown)


assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -388,6 +388,10 @@ export default class VisitorGraph extends React.Component {
     }
   }
 
+  isGraphCollapsed() {
+    return !this.state.metric
```
vinibrsl, line 389: What do you think of making this state clearer by comparing to an empty string, which is the state when you collapse the graph?

```js
return this.state.metric === ""
```

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -401,13 +405,13 @@ export default class VisitorGraph extends React.Component {
...
   fetchGraphData() {
-    if (!this.state.metric) {
+    if (this.isGraphCollapsed()) {
       this.setState({ mainGraphLoadingState: LOADING_STATE.loaded, graphData: null })
       return
     }
 
     const url = `/api/stats/${encodeURIComponent(this.props.site.domain)}/main-graph`
-    let params = {metric: this.state.metric || 'none'}
+    let params = {metric: this.state.metric}
```
vinibrsl, line 411: Good catch!

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -330,16 +329,9 @@ export default class VisitorGraph extends React.Component {
     return validIntervals.includes(interval)
   }
 
-  maybeRollbackInterval() {
```
vinibrsl, line 336: This function checked for invalid states at the very first paint of the component. For example if Plausible ever changes an interval name or removes an interval for a period, users' localStorage would still keep the old invalid value. We need to check that and reset the interval in this case.

Additionally we could delete the localStorage entry whenever an invalid interval value is found. Currently we only reset the state, but not the stored value.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -448,8 +448,8 @@ export default class VisitorGraph extends React.Component {
 
     const theme = document.querySelector('html').classList.contains('dark') || false
 
-    const topStatsReadyOrRefreshing = (topStatsLoadingState === LOADING_STATE.ready || topStatsLoadingState === LOADING_STATE.refreshing)
-    const mainGraphReadyOrRefreshing = (mainGraphLoadingState === LOADING_STATE.ready || mainGraphLoadingState === LOADING_STATE.refreshing)
+    const topStatsReadyOrRefreshing = (topStatsLoadingState === LOADING_STATE.loaded || topStatsLoadingState === LOADING_STATE.refreshing)
+    const mainGraphReadyOrRefreshing = (mainGraphLoadingState === LOADING_STATE.loaded || mainGraphLoadingState === LOADING_STATE.refreshing)
```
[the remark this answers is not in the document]
RobertJoonas: Agreed. Done in https://github.com/plausible/analytics/pull/2532/commits/216d50bb9b0a79346a7254d61d97db552ccc1bd4.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -393,14 +393,11 @@ export default class VisitorGraph extends React.Component {
...
-  updateMetric(newMetric) {
-    if (newMetric === this.state.metric) {
-      storage.setItem(`metric__${this.props.site.domain}`, "")
-      this.setState({ metric: "" })
-    } else {
-      storage.setItem(`metric__${this.props.site.domain}`, newMetric)
-      this.setState({ metric: newMetric })
-    }
+  updateMetric(clickedMetric) {
+    const newMetric = clickedMetric !== this.state.metric ? clickedMetric : ""
```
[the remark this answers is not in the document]
RobertJoonas: Good point, just makes it harder to read. Changed in https://github.com/plausible/analytics/pull/2532/commits/39f2186f5dc4dd2d6fe394139af21f106b260650

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Agreed: https://github.com/plausible/analytics/pull/2532/commits/ef7a5d9274a130cd372ddfe8e0f90c00be96e194

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Thanks for pointing that out, @vinibrsl! I totally missed thinking about this. It was indeed hard to understand that this was the reason `maybeRollbackInterval` was called instead of `fetchGraphData`. Applied your suggested changes in https://github.com/plausible/analytics/pull/2532/commits/b9c9da3591ac2545e0d0d347a31674eef4f556bc and also left a comment.

ukutaht: I'm still not a fan of the way we have to sync localStorage to the component state along with worrying about whether it's invalid or not.

This got me thinking that maybe the `interval` doesn't need to be stored in the component state at all. So I made a little PoC that you can see here: https://github.com/plausible/analytics/compare/flipmove...interval-without-state?expand=1

The main difference is that the `interval` is gone from the component state. It's always fetched from localStorage when sending a request to the backend. And updating the `interval` value just means updating localStorage. As far as I can tell this works fine, and doesn't allow invalid intervals to cause an error state. What do you think @RobertJoonas @vinibrsl ?

vinibrsl: Good call, @ukutaht! I agree that reading from a single source is better than managing two different sources (localStorage and component state).

RobertJoonas:
@ukutaht 

> I pulled down the code and I saw that reports using ListReport don't show a spinner when changing period from the date picker. Is that a regression?

You're right. The spinner doesn't show anymore. Should I fix it with a new commit in this PR?

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: I like that idea, and I agree that we wouldn't need to keep the interval state in the component itself. As long as we have default interval values, there shouldn't be any concerns with this. @ukutaht, @vinibrsl, is it OK to open that PoC PR that Uku linked, test it and merge into this branch?

vinibrsl (approved):

assets/js/dashboard/stats/graph/visitor-graph.js
vinibrsl: Looks like a good plan. I approved this PR in case you want to do that in a separate PR.

PR #2535, opened 2022-12-21
Realtime prop breakdown

RobertJoonas:
### Changes

This PR makes the custom event property breakdown automatically update in realtime mode. It also disables the "Load more" button in realtime mode (screenshot of the button I mean), because supporting paginated results in realtime is too complex.

image

This is how the automatic updates look like (realtime ticks are sped up from 30s to 5s, and there's an artificial delay in fetching the data to make the loading spinner visible):

https://user-images.githubusercontent.com/56999674/208881570-6554f7e0-89ad-49b3-8dc9-ca7073f309ed.mp4

I also considered using `react-flip-move` as in `ListReport`, but as the container size changes, it causes the animation to overflow the container borders:

https://user-images.githubusercontent.com/56999674/208881785-3d8eac28-b265-4ee7-9e56-972f19c64ce5.mp4

That's why I didn't use FlipMove here.

### Tests
- [x] Writing unit tests for the React dashboard is currently out of scope

### Changelog
- [x] https://github.com/plausible/analytics/pull/2445 updates changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/conversions/prop-breakdown.js
+++ b/assets/js/dashboard/stats/conversions/prop-breakdown.js
@@ -149,7 +149,7 @@ export default class PropertyBreakdown extends React.Component {
   renderLoading() {
     if (this.state.loading) {
       return <div className="px-4 py-2"><div className="loading sm mx-auto"><div></div></div></div>
-    } else if (this.state.moreResultsAvailable) {
+    } else if (this.state.moreResultsAvailable && this.props.query.period !== 'realtime') {
       return (
         <div className="w-full text-center my-4">
           <button onClick={this.loadMore.bind(this)} type="button" className="button">
--- a/assets/js/dashboard/stats/conversions/prop-breakdown.js
+++ b/assets/js/dashboard/stats/conversions/prop-breakdown.js
@@ -46,6 +46,7 @@ export default class PropertyBreakdown extends React.Component {
     }
 
     this.handleResize = this.handleResize.bind(this);
+    this.fetchPropBreakdown = this.fetchPropBreakdown.bind(this)
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/conversions/prop-breakdown.js
```diff
@@ -69,17 +76,34 @@ export default class PropertyBreakdown extends React.Component {
...
-            moreResultsAvailable: res.length === 100
+            moreResultsAvailable: res.length === BREAKDOWN_LIMIT
           })))
     }
   }
 
+  doFetch(nextPage=false) {
+    const page = nextPage ? this.state.page : 1
+    return api.get(`/api/stats/${encodeURIComponent(this.props.site.domain)}/property/${encodeURIComponent(this.state.propKey)}`, this.props.query, {limit: BREAKDOWN_LIMIT, page: page})
+  }
```
vinibrsl, line 103: Could you please clarify why a new function is needed? From my understanding we could change `fetchPropBreakdown` to always concat instead of replacing the entire `state.breakdown` list, with a initial state of `[]`.

RobertJoonas: Good point @vinibrsl! That sounds like a cleaner implementation, indeed. Can I force-push and overwrite the last 2 commits?

vinibrsl: I'd push a new commit as it's easier to review :) force pushing overwrites history, and we squash commits when "merging" PRs anyways.

RobertJoonas: @vinibrsl, I just found one small reason why the current implementation would be better. If we were to reset `state.breakdown` to `[]` before fetching the data, and always concatenating the result, we should display the loading spinner for as long as it loads. To my understanding, we cannot make it look very dynamic that way. All other reports in realtime are updating without the loading spinner.

vinibrsl (approved):

PR #2537, opened 2022-12-21
Added certs to runtime container to enable SSL for SMTP

DavyLandman:
### Changes

If you are sending email via an SMTP server that requires SSL/TLS, you get this warning printed to the output log:

```
'Authenticity is not established by certificate path validation' 
Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
```

And no emails are send.

Adding ca-certificats to the runtime image solves this.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,7 @@ ENV LANG=C.UTF-8
 
 RUN apk upgrade --no-cache
 
-RUN apk add --no-cache openssl ncurses libstdc++ libgcc
+RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates
 
 COPY ./rel/docker-entrypoint.sh /entrypoint.sh
 


ukutaht:
@ruslandoga could you test this and give your opinion? Whenever you have time :)

DavyLandman:
Note, I am running this on a server (actually, I had to build a dedicated docker file)

```dockerfile
FROM plausible/analytics:v1.5.1
USER root
RUN apk --no-cache add ca-certificates
USER plausibleuser
```

and this seems to fix the problem. So I'm not trying to just snipe in a random dependency here.

ruslandoga:
Testing it today :)

DavyLandman:
Interesting that it worked "on my machine". I'll try and login to the server tomorrow.

Note, in my setup I was directly connecting to an external smtp server (that enforced SSL), I didn't use the internal forwarder that is the default docker-compose setup.

DavyLandman:
I just checked, I also still get these warnings in the log:

```
plausible_1               | 13:00:01.039 [warning] Description: 'Authenticity is not established by certificate path validation'
plausible_1               |      Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
plausible_1               |
plausible_1               | 12:00:00.458 [warning] Description: 'Authenticity is not established by certificate path validation'
plausible_1               |      Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
plausible_1               |
plausible_1               | 12:00:00.171 [warning] Description: 'Authenticity is not established by certificate path validation'
plausible_1               |      Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
plausible_1               |
plausible_1               | 12:00:00.720 [warning] Description: 'Authenticity is not established by certificate path validation'
plausible_1               |      Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
plausible_1               |
```

But, the difference is, it actually works :| so now it's just a warning. Before adding `ca-certificates` it would print this warning and also fail to send something.

> Another approach is to use [`:tls_cacertfile` config option](https://hexdocs.pm/bamboo_smtp/Bamboo.SMTPAdapter.html#module-example-config) from `Bamboo.SMTPAdapter` and set it to [`CAStore.file_path`](https://github.com/elixir-mint/castore#usage) which is already used by Mint.

Sharing the cacert looks like a good approach.

ukutaht:
@ruslandoga what's your verdict on this? Worth merging?

ruslandoga:
I think so, yes :)

ukutaht (approved):

ukutaht:
Thanks @ruslandoga and @DavyLandman !

zenminimalist:
I'm currently seeing this as well in my Plausible container logs. I have set `SMTP_HOST_SSL_ENABLED` to `true`

```
17:19:45.880 [warning] Description: 'Server authenticity is not verified since certificate path validation is not enabled'
     Reason: 'The option {verify, verify_peer} and one of the options \'cacertfile\' or \'cacerts\' are required to enable this.'
```
I'm using the STMP sever credentials I created via [Fastmail](https://www.fastmail.help/hc/en-us/articles/1500000279921-IMAP-POP-and-SMTP), not a sending service e.g. Sendgrid etc.

I'm unsure what to add where. Is there EVN variable I need to add? 

I didn't see a variable for this in the [docs](https://plausible.io/docs/self-hosting-configuration#mailersmtp-setup).

PR #2544, opened 2022-12-26
Trace breakdown queries with OTEL

vinibrsl:
This pull request traces breakdown queries with OTEL. This will be useful to know if some properties are being used like `entry_page`, as we want to migrate that for session metrics, for example.

Honeycomb screenshot:

Screenshot 2022-12-26 at 11 59 43

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -7,6 +7,7 @@ defmodule Plausible.Stats.Query do
             include_imported: false
 
   @default_sample_threshold 20_000_000
+  require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{FilterParser, Interval}
 
   def shift_back(%__MODULE__{period: "year"} = query, site) do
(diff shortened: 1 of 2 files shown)


lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -64,6 +65,11 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
          query <- Query.from(site, params),
          {:ok, metrics} <- parse_metrics(params, property, query),
          {:ok, limit} <- validate_or_default_limit(params) do
+      Tracer.set_attributes([
```
RobertJoonas, line 68: I think doing this in `ExternalStatsController` would only trace Stats API breakdown requests. But we probably want to track all of them (on the dashboard too) right? How about doing this in `Plausible.Stats.Breakdown` instead? I think Stats API requests are also going through the same modules in the end, so they would still be covered.

lib/plausible/stats/query.ex
```diff
@@ -215,10 +225,12 @@ defmodule Plausible.Stats.Query do
           })
         )
         |> Map.put(:period, "all")
+        |> trace_query()
```
RobertJoonas, line 228: Maybe instead of tracing the queries where they're created, we can trace these in `Plausible.Stats.Aggregate`, `Plausible.Stats.Breakdown`, etc ... too? Just thinking we could make do with less `trace_query` calls. What do you think?

RobertJoonas (commented):
Thanks for this @vinibrsl! Having these traces will be useful for observability on my realtime improvements PR, too! :)

Left a couple of suggestions in-line, let me know what you think.

RobertJoonas (approved):

PR #2558, opened 2023-01-03
Wrap paddle actions with DB transactions

aerosol:
### Changes

This change makes all the paddle webhook updates transactional. 
I considered rewriting it to Ecto.Multi-based services but that can be made on top of this in another iteration, since shipping the change itself feels quite critical.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -12,93 +12,78 @@ defmodule Plausible.Billing do
     user_id |> active_subscription_query() |> Repo.exists?()
   end
 
-  defp active_subscription_query(user_id) do
-    from s in Subscription,
-      where: s.user_id == ^user_id and s.status == "active",
-      order_by: [desc: s.inserted_at],
-      limit: 1
-  end
-
   def subscription_created(params) do
-    params =
-      if present?(params["passthrough"]) do
-        params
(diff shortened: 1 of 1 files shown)


vinibrsl (approved):
Good one! And I agree that moving to Ecto.Multi is a good idea and we should do in a separate release.

aerosol:
@vinibrsl made the errors bubble up so it rolls back this time

ukutaht (approved):
Looks good! 

Is there any benefit between `Ecto.Multi` vs `Repo.transaction`? If I understand right it's just cosmetic / style difference or is there more to it?

aerosol:
> Is there any benefit between Ecto.Multi vs Repo.transaction? If I understand right it's just cosmetic / style difference or is there more to it?

Repo.transaction accepts Ecto.Multi structs. The way I see it there's testability benefit - you can test some aspects without having to provision the DB; sometimes all you care about is whether the transaction structure/context has been built valid, not whether the database works. You also get better insight into failures, the failing step is explicitly named & returned. It makes it simpler to mock (or skip) certain transaction steps if need be. You can compose multiple Multis :smiling_imp: together with little to no effort - this supports building "service oriented" code bases.

PR #2562, opened 2023-01-03
Fix breakdown API pagination when using event metrics

### LINKED ISSUE #2561: API pagination's duplicates
### Discussed in https://github.com/plausible/analytics/discussions/2415

<div type='discussions-op-text'>

<sup>Originally posted by **NoeGRANGE** November  2, 2022</sup>
Hello,

I'm using your API to make my own dashboard but I think I got a problem when I try to request breakdown with `limit` and `page` parameters.
Every result from the second page already exist on the first one.

- `https://plausible.io/api/v1/stats/breakdown?site_id=esf-meribel.com&period=custom&date=2022-09-13,2022-11-02&metrics=visitors&property=event:props:orderId&filters=event:name==MseM: Payment success&limit=1000&page=1`

- `https://plausible.io/api/v1/stats/breakdown?site_id=esf-meribel.com&period=custom&date=2022-09-13,2022-11-02&metrics=visitors&property=event:props:orderId&filters=event:name==MseM: Payment success&limit=1000&page=2`

Am I using these parameters wrong?

Thanks in advance 👍</div>

vinibrsl:
This commit fixes a bug where the subsequent breakdown API pages had the same items as the first page. The fix sorts the underlying ClickHouse query by timestamp, keeping the same order between requests, as we use OFFSET/LIMIT pagination.

Closes #2561

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -223,7 +223,7 @@ defmodule Plausible.Stats.Breakdown do
 
   defp breakdown_events(site, query, property, metrics, pagination) do
     from(e in base_event_query(site, query),
-      order_by: [desc: fragment("uniq(?)", e.user_id), asc: fragment("min(?)", e.timestamp)],
+      order_by: [desc: fragment("uniq(?)", e.user_id)],
       select: %{}
     )
     |> do_group_by(property)
@@ -327,7 +327,8 @@ defmodule Plausible.Stats.Breakdown do
       group_by: e.name,
       where: meta.key == ^prop,
       group_by: meta.value,
-      select_merge: %{^prop => meta.value}
+      select_merge: %{^prop => meta.value},
+      order_by: {:asc, meta.value}
     )
   end
 
@@ -340,7 +341,8 @@ defmodule Plausible.Stats.Breakdown do
       inner_lateral_join: meta in fragment("meta"),
       where: meta.key == ^prop,
       group_by: meta.value,
-      select_merge: %{^prop => meta.value}
+      select_merge: %{^prop => meta.value},
+      order_by: {:asc, meta.value}
     )
   end
 
@@ -351,7 +353,8 @@ defmodule Plausible.Stats.Breakdown do
     from(
       e in q,
       group_by: e.name,
-      select_merge: %{name: e.name}
+      select_merge: %{name: e.name},
+      order_by: {:asc, e.name}
     )
   end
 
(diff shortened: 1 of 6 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -223,7 +223,7 @@ defmodule Plausible.Stats.Breakdown do
 
   defp breakdown_events(site, query, property, metrics, pagination) do
     from(e in base_event_query(site, query),
-      order_by: [desc: fragment("uniq(?)", e.user_id)],
+      order_by: [desc: fragment("uniq(?)", e.user_id), asc: fragment("min(?)", e.timestamp)],
```
aerosol, line 226: Given that the timestamp is DateTime with second-precision, is there still a way for results to get scrambled?

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
```diff
@@ -1588,6 +1588,47 @@ defmodule PlausibleWeb.Api.ExternalStatsController.BreakdownTest do
       assert Enum.count(res["results"]) == 2
     end
 
+    test "does not repeat results", %{conn: conn, site: site} do
```
aerosol, line 1591: I've tried running this test without the ordering patch and it passes. Is this an undefined DB behavior I must be lucky to hit?

lib/plausible/stats/breakdown.ex
ukutaht: Another thought: should `breakdown_sessions` get the same treatment? It might have the same problem with undefined ordering and pagination.

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
vinibrsl: Oops! Good catch. I must've changed this somewhere in time and the assertion was wrong. Fixed in f5e026e3413dd275310489e7b470c8aa689552ef

lib/plausible/stats/breakdown.ex
vinibrsl: Yes, you're right. We need a more "unique" key to order to avoid this bug in different scenarios.

I changed the patch to use alphabetical ordering in each breakdown property. The functions are shared between breakdown_sessions and breakdown_events, so it should fix both cases. I tried making it dynamic and "DRY-er" but it grew more complex than I wanted, so I went the simple way. Changed in c4cb566f41efc119f4848344f92c78c589cc4f04.

vinibrsl: Another concern: I'd consider this a breaking change to the API, as the first page (and other pages too) are not the same after releasing this. Do you see this as a problem?

ukutaht: > Do you see this as a problem?

The way I understand is that ordering for props with the same amount of visitors used to be random, but now it's stable. I don't see this as a problem. Previously the users could not rely on any ordering so I don't see it as a problem

ukutaht (approved):

PR #2564, opened 2023-01-04
Refine parallel application-side joins

ukutaht:
### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/clickhouse_repo.ex
+++ b/lib/plausible/clickhouse_repo.ex
@@ -13,7 +13,14 @@ defmodule Plausible.ClickhouseRepo do
 
   @task_timeout 60_000
   def parallel_tasks(queries) do
-    Task.async_stream(queries, fn fun -> fun.() end, max_concurrency: 3, timeout: @task_timeout)
+    ctx = OpenTelemetry.Ctx.get_current()
+
+    execute_with_tracing = fn fun ->
+      OpenTelemetry.Ctx.attach(ctx)
+      fun.()
+    end
+
+    Task.async_stream(queries, execute_with_tracing, max_concurrency: 3, timeout: @task_timeout)
     |> Enum.to_list()
     |> Keyword.values()
   end
--- a/lib/plausible/clickhouse_repo.ex
+++ b/lib/plausible/clickhouse_repo.ex
@@ -10,4 +10,10 @@ defmodule Plausible.ClickhouseRepo do
       import Ecto.Query, only: [from: 1, from: 2]
     end
   end
+
+  @task_timeout 15_000
+  def parallel_tasks(queries) do
+    Enum.map(queries, &Task.async/1)
+    |> Task.await_many(@task_timeout)
+  end
 end
--- a/lib/plausible/clickhouse_repo.ex
+++ b/lib/plausible/clickhouse_repo.ex
@@ -11,9 +11,10 @@ defmodule Plausible.ClickhouseRepo do
     end
   end
 
-  @task_timeout 15_000
+  @task_timeout 60_000
   def parallel_tasks(queries) do
-    Enum.map(queries, &Task.async/1)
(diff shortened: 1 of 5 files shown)


lib/plausible/clickhouse_repo.ex
```diff
@@ -10,4 +10,10 @@ defmodule Plausible.ClickhouseRepo do
       import Ecto.Query, only: [from: 1, from: 2]
     end
   end
+
+  @task_timeout 15_000
+  def parallel_tasks(queries) do
```
ukutaht, line 15: The idea with this is to consolidate the timeouts for parallel clickhouse requests. That way there's a single number to tune instead of having magic numbers all over the place.

aerosol: I'm wondering if we should consider `Task.async_stream` here instead and spawn queries in chunks with `max_concurrency` capped to something low (2? 3? 4?). My worry here is spawning too many queries at the same time is prone to races in DB pool utilization - a single request will potentially acquire too many connections. Not sure if we have telemetry data for that part, would be nice to get a breakdown of how time consuming the queries are and to what extent the parallel joins are efficient/worth it. WYT?

ukutaht: Yes, good point. The CSV export will trigger something like >10 queries in parallel, all of which could potentially be fairly long-running. I think the `max_concurrency` option is good and then I would up the timeout to something significantly longer like 60s to give it an opportunity to finish all the concurrent chunks. I'll make that change.

Ideally it should be possible to push a lot of this stuff down to the database which cuts down on the nr of queries and back-and-forth between the app and DB. But this would require some more advanced query composition in Ecto.

Your question about telemetry made me check Honeycomb and I found an issue with OTEL tracking: when using the `Task` module it spawns a new process and the OTEL package we use is unable to recognise the Task as a child span of the original request. See [this trace](https://ui.honeycomb.io/plausible/datasets/plausible-prod/result/xoyRXDT5GqA/trace/JFMSP5DSAXj?fields[]=c_name&fields[]=c_service.name&span=fade3bf4f4518d42) for example. It's missing the Clickhouse queries and I'm 90%+ sure it's because these are running in `Task` processes. Flagging with @vinibrsl as well.

If we can fix that we should be able to see a beautiful trace with 2 Clickhouse queries in parallel with timings on them. For now, we don't have that yet.

ukutaht: Changed in https://github.com/plausible/analytics/pull/2564/commits/cfd0adf4c01ca41cff1a450a1627eca9a909a677

ukutaht:
https://github.com/plausible/analytics/pull/2564/commits/facc49501d3d8dd823d7940e5724a36e63a48942 is an attempt to propagate the OpenTelemetry context to child Tasks so that the DB queries in Task processes are linked to the original HTTP request.

Unfortunately I cannot test this properly because our Honeycomb account is heavily throttled at the moment and I can't see the traces I create locally (at least not reliably). I think they're dropped by Honeycomb due to the throttling. Do you want to leave this in @vinibrsl and evaluate later? Or should I get rid of it for now?

aerosol (approved):

vinibrsl:
> 



> [facc495](https://github.com/plausible/analytics/commit/facc49501d3d8dd823d7940e5724a36e63a48942) is an attempt to propagate the OpenTelemetry context to child Tasks so that the DB queries in Task processes are linked to the original HTTP request.
> 
> Unfortunately I cannot test this properly because our Honeycomb account is heavily throttled at the moment and I can't see the traces I create locally (at least not reliably). I think they're dropped by Honeycomb due to the throttling. Do you want to leave this in @vinibrsl and evaluate later? Or should I get rid of it for now?

Let's leave it in and evaluate later. I'll create a TODO internally so we can come back and check.

vinibrsl (approved):

PR #2589, opened 2023-01-12
Fix metric undefined

RobertJoonas:
### Changes

This PR fixes a bug introduced in https://github.com/plausible/analytics/pull/2560. For a reproduction of the bug, go to [this dashboard URL](https://plausible.io/plausible.io?goal=Signup&props=%7B%22logged_in%22%3A%22false%22%7D&period=realtime) and remove the `Signup.logged_in is false` filter. After doing so:

- The graph disappears (main-graph API request returns 500)
- An error is shown in the console: `ApiError: ** (CaseClauseError) no case clause matching: :undefined`

This is because `metric` becomes undefined in the `VisitorGraph` state when setting it to `topStatLabels[0]` [on this line](https://github.com/plausible/analytics/blob/b6349df475a17c8fdc3a30cdeace18e88b29c52d/assets/js/dashboard/stats/graph/visitor-graph.js#L386) (if `topStatLabels` is an empty array).

This PR makes sure that we're not relying on the top stat labels when setting the metric. It also does some refactoring to make the code more readable. Reviewing commit-by-commit suggested.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -373,24 +373,22 @@ export default class VisitorGraph extends React.Component {
     if (metric !== prevState.metric) {
       this.setState({mainGraphLoadingState: LOADING_STATE.refreshing}, this.fetchGraphData)
     }
-
-    this.maybeUpdateMetric(prevState)
   }
 
-  maybeUpdateMetric(prevState) {
+  resetMetric() {
     const { metric, topStatData } = this.state;
     const { query, site } = this.props
 
     const savedMetric = storage.getItem(`metric__${site.domain}`)
-    const topStatLabels = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
-    const prevTopStatLabels = prevState.topStatData && prevState.topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
-    if (topStatLabels && `${topStatLabels}` !== `${prevTopStatLabels}`) {
+    const selectableMetrics = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
+
+    if (selectableMetrics) {
       if (query.filters.goal && metric !== 'conversions') {
         this.setState({ metric: 'conversions' })
-      } else if (topStatLabels.includes(savedMetric) && savedMetric !== "") {
+      } else if (selectableMetrics.includes(savedMetric) && savedMetric !== "") {
         this.setState({ metric: savedMetric })
       } else {
-        this.setState({ metric: topStatLabels[0] })
+        this.setState({ metric: selectableMetrics[0] })
       }
     }
   }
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -381,15 +381,14 @@ export default class VisitorGraph extends React.Component {
...
-      if (query.filters.goal) {
-        this.setState({ metric: 'conversions' })
-      } else if (selectableMetrics.includes(savedMetric) && savedMetric !== "") {
-        this.setState({ metric: savedMetric })
-      } else {
-        this.setState({ metric: 'visitors' })
-      }
+    if (query.filters.goal) {
+      this.setState({ metric: 'conversions' })
+    } else if (canSelectSavedMetric || savedMetric === "") {
```
RobertJoonas, line 388: About the `&& savedMetric !== ""` -> `|| savedMetric === ""` change:

As we're not checking whether the top stat labels have changed anymore (and are just resetting the metric every time new top-stat data is fetched), the realtime view will automatically expand a collapsed graph every tick. This change here prevents it, and keeps the graph hidden if the metric saved to localStorage is `""`.

My thinking was that if a user has collapsed the graph, they should expect it to be collapsed after they refresh the dashboard too.

Let me know what you think about this change. Should we be concerned user accidentally hiding the graph and not getting it to display again with a refresh?

ukutaht:
I can't seem to reproduce this locally on `master`, do I need to be on some branch?

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -437,6 +435,7 @@ export default class VisitorGraph extends React.Component {
     api.get(`/api/stats/${encodeURIComponent(this.props.site.domain)}/top-stats`, this.props.query)
       .then((res) => {
         this.setState({ topStatsLoadingState: LOADING_STATE.loaded, topStatData: res })
+        this.resetMetric()
```
ukutaht, line 437: Why move this from `componentDidUpdate` to the API response callback?

RobertJoonas: I did this because the top stat labels can only change when we fetch new data (the labels are returned from the API). And the only condition where we want to "reset the metric" is when the labels have changed.

This will avoid having to write this (or something similar) in `componentDidUpdate`

```js
const topStatLabels = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
const prevTopStatLabels = prevState.topStatData && prevState.topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)

const topStatLabelsChanged = topStatLabels && `${topStatLabels}` !== `${prevTopStatLabels}`
```

(There's another function called `updateMetric` which is used when a new metric is selected)

ukutaht: Makes sense. In this case I think it would be best to run it as a callback to `setState` like so:

```javascript
this.setState({ topStatsLoadingState: LOADING_STATE.loaded, topStatData: res }, this.resetMetrics)
```

assets/js/dashboard/stats/graph/visitor-graph.js
ukutaht: I think it makes sense, no concerns.

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Done in https://github.com/plausible/analytics/pull/2589/commits/f1ef1f41b9761f671d47b5da94e0e6439f40959a

ukutaht (approved):

PR #2596, opened 2023-01-16
feat: allow to use sendgrid mailer

shiipou:
Or other mailer from the list https://hexdocs.pm/bamboo/readme.html#available-adapters

### Changes

I want to use SendGrid API to send mails.
Bamboo, the lib you use to send mails, was shipped with official sendgrid support, we just need to allow us to define the api_key.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -275,12 +275,14 @@ config :plausible, Plausible.ClickhouseRepo,
   max_buffer_size: ch_max_buffer_size
 
 case mailer_adapter do
+
+  # Keep this just in case of backward compatibility
   "Bamboo.PostmarkAdapter" ->
     config :plausible, Plausible.Mailer,
       adapter: :"Elixir.#{mailer_adapter}",
-      request_options: [recv_timeout: 10_000],
+      request_options: [recv_timeout: 10_000],      # POSTMARK_API_KEY must be deprecated, use MAILER_API_KEY instead.
       api_key: get_var_from_path_or_env(config_dir, "POSTMARK_API_KEY")
-
+      
   "Bamboo.SMTPAdapter" ->
     config :plausible, Plausible.Mailer,
       adapter: :"Elixir.#{mailer_adapter}",
@@ -302,7 +304,15 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, adapter: Bamboo.TestAdapter
 
   _ ->
-    raise "Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter"
+    case String.length(get_var_from_path_or_env(config_dir, "MAILER_API_KEY")) ->
+      0 ->
+        raise "Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter"
+      _ ->
+        config :plausible, Plausible.Mailer,
+          adapter: :"Elixir.#{mailer_adapter}",
+          request_options: [recv_timeout: 10_000],
+          api_key: get_var_from_path_or_env(config_dir, "MAILER_API_KEY") # POSTMARK_API_KEY must be deprecated, use MAILER_API_KEY instead.
+    end
 end
 
 config :plausible, PlausibleWeb.Firewall,


config/runtime.exs
```diff
@@ -302,7 +304,15 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, adapter: Bamboo.TestAdapter
 
   _ ->
-    raise "Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter"
+    case String.length(get_var_from_path_or_env(config_dir, "MAILER_API_KEY")) ->
+      0 ->
+        raise "Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter"
+      _ ->
+        config :plausible, Plausible.Mailer,
+          adapter: :"Elixir.#{mailer_adapter}",
```
ruslandoga, line 312: I think that would mean you need to provide the adapter prefixed by `Bamboo.` as `Bamboo.SendgridAdapter`. Is this intentional?

config/runtime.exs
```diff
@@ -302,7 +304,15 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, adapter: Bamboo.TestAdapter
 
   _ ->
-    raise "Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter"
+    case String.length(get_var_from_path_or_env(config_dir, "MAILER_API_KEY")) ->
```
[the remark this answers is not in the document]
ruslandoga: Another way is to actually check for module being defined.

config/runtime.exs
shiipou: With the current config in master, as I understand we  already have to set the var like this : 
```
MAILER_ADAPTER=Bamboo.SMTPAdapter
```

Correct me if I'm wrong.

ruslandoga: Right, then the error message seems incorrect.

> Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter

config/runtime.exs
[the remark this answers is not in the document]
shiipou: I admit I'm not so familiar with Elixir, so idk how I can check if the module exist.
Did you think `code.module_status(:"Elixir.#{mailer_adapter}")` is a good way ?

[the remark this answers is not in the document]
ruslandoga: I'd probably use `module.__info__(:functions)` where module is either `:"Elixir.#{mailer_adapter}"` or `Module.safe_concat(Elixir, String.to_atom(mailer_adapter))` or some combination of the two. In `module.__info__(:functions)` we can additionally look for the `deliver/2` function, but that is probably unnecessary.

config/runtime.exs
shiipou: > > Right, then the error message seems incorrect.
> 
> > Unknown mailer_adapter; expected SMTPAdapter or PostmarkAdapter

Ok I'll correct it then.

config/runtime.exs
```diff
@@ -302,7 +304,18 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, adapter: Bamboo.TestAdapter
 
   _ ->
```
aerosol, line 306: Can't we simplify this by matching "Bamboo.SMTPAdapter" and "PostmarkAdapter" explicitly in their own clauses? What's the purpose of this fallback logic looking up modules and validating entries? 

Context: the simpler this gets, the easier it is for us to implement secondary mailer fallback logic we've been planning (e.g. to use sendgrid when postmark is unavailable).

ruslandoga: I guess the idea is to delegate adapter discovery to bamboo and support whatever it supports. I agree, however, that hardcoding all known adapters is easier and potentially more configurable.

ukutaht: Yes I agree with the hardcoding. I think it's also best to keep the API_KEY variables specific to each backend i.e. `POSTMARK_API_KEY`, `SENDGRID_API_KEY`, etc.

Happy to accept this PR if the current stuff stays backwards compatible, and the new adapter is added with explicit matching like the previous ones.

ruslandoga: I tried the hardcoding approach in https://github.com/plausible/analytics/pull/2649.

I didn't fork from this branch as there wasn't much code in common.

aerosol:
Closing in favor of https://github.com/plausible/analytics/pull/2649

PR #2601, opened 2023-01-17
parse date labels by local timezone instead of UTC

RobertJoonas:
### Changes

This PR fixes the date parsing in the main graph time buckets. The timestamps returned by the API are in `site.timezone`. We should not parse them as UTC.

Here's a GIF showing how it currently behaves. I am in UTC +02, which is why from 2AM (the UTC midnight) the date is changed from 16 -> 17 Jan:

before

After this change:

after

### Tests

### Changelog

### Documentation

### Dark mode


aerosol (approved):
I'm not sure if that wouldn't appear too convoluted but maybe we could include timezone information in that tooltip to leave no room for confusion?

ukutaht:
> I'm not sure if that wouldn't appear too convoluted but maybe we could include timezone information in that tooltip to leave no room for confusion?

Good idea but I tried a couple approaches and it does make it quite convoluted and hard to read..

PR #2604, opened 2023-01-17
Stop logging invalid_address errors from Locus

vinibrsl:
This commit stops logging `{:invalid_address, "ip"}` errors. This reduces noise on Sentry, that is capturing error logs. Other errors are still logged.

### DIFF
--- a/lib/plausible/geo.ex
+++ b/lib/plausible/geo.ex
@@ -181,8 +181,11 @@ defmodule Plausible.Geo do
       :not_found ->
         nil
 
+      {:error, {:invalid_address, _address}} ->
+        nil
+
       {:error, reason} ->
-        Logger.error("failed to lookup ip address: " <> inspect(reason))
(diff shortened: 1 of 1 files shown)


lib/plausible/geo.ex
```diff
@@ -181,8 +181,11 @@ defmodule Plausible.Geo do
       :not_found ->
         nil
 
+      {:error, {:invalid_address, _address}} ->
+        nil
```
ruslandoga, line 185: What were the invalid addresses that were logged?

ruslandoga: Ah, it was `{:invalid_address, "ip"}`.

vinibrsl: @ruslandoga from our internal discussion do you still think we should keep that log? :)

ruslandoga: I think your changes are the way to go :)

ruslandoga (approved):

PR #2605, opened 2023-01-18
Communicate site creation edge cases to the users

aerosol:
### Changes

This PR introduces the following changes:
  - Domains with existing events cannot be created; either asynchronous deletion is in progress or support intervention is necessary.
image

  - I found user-facing code presenting a warning when site limit is reached, but it was never displayed (raw 4xx error page was returned). It is now shown when applicable:
image

  - The whole site creation process is wrapped with a single Postgres transaction
  - Some tests were naively misusing `Repo.exists?` so that it always evaluated to true - this has been fixed

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -191,11 +191,8 @@ defmodule Plausible.Sites do
       )
 
       [
-        domain: """
-        We cannot create '#{domain}' at the moment; if the site was recently deleted, 
-        we are still processing the deletion, in which case please try again later. 
-        Otherwise, please contact support.
-        """
+        domain:
+          "This domain has already been taken. Perhaps one of your team members registered it? If that's not the case, please contact support@plausible.io"
       ]
     else
       []
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -1,5 +1,6 @@
 defmodule Plausible.Sites do
   use Plausible.Repo
+  alias Plausible.ClickhouseRepo
   alias Plausible.Site
   alias Plausible.Site.SharedLink
   import Ecto.Query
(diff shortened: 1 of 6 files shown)


lib/plausible/sites.ex
```diff
@@ -166,4 +183,22 @@ defmodule Plausible.Sites do
...
   end
+
+  defp check_for_existing_events(domain, params) do
+    if has_events?(domain) do
+      Sentry.capture_message("Refused to create a site with existing events",
+        extra: %{params: params}
+      )
+
+      [
+        domain: """
```
ukutaht, line 194: I'm not sure if we need to expose the inner workings of async deletes to users. How about showing the same error message as when the site already exists in the postgres database here?

Screenshot 2023-01-18 at 12 08 32

We would still get the error message sent to devs through Sentry so we can figure it out on our end.

cc @metmarkosaric what do you think?

lib/plausible_web/templates/site/new.html.eex
```diff
@@ -16,7 +16,7 @@
             </h3>
             <div class="mt-2 text-sm text-yellow-700 dark:text-yellow-300">
               <p>
-              Your account is limited to <%= @site_limit %> sites. Please contact hello@plausible.io to add more sites.
+              Your account is limited to <%= @site_limit %> sites. Please contact support to add more sites.
```
ukutaht, line 19: I think 'contact support' with no link or email address is not the best experience for the customer. They'll have to go hunting for how to contact support in the footer or something like this.

Self-hosted installations don't have a site limit so it's no problem to show our contact email address here. I think it's better UX to revert this or link the words 'contact support' to https://plausible.io/contact

aerosol: Cheers, good catch, I initially meant to include https://plausible.io/contact link right under the "create site" modal. I suppose clickable error message text is preferred? 

I didn't like that there's a single screen technically advising using two different emails support@plausible.io and hello@plausible.io depending on the error.

lib/plausible/sites.ex
aerosol: Agreed that's a good call; no need to be very specific here as we'll still keep track of it internally.

metmarkosaric: yeah, i agree. i mentioned the same thing yesterday on the basecamp thread

aerosol: https://github.com/plausible/analytics/pull/2605/commits/eaefca7d5c22ddd90ced43c84e3f403419594e22

lib/plausible_web/templates/site/new.html.eex
aerosol: Included the e-mail back again in https://github.com/plausible/analytics/pull/2605/commits/54b4975038dbc6028c88d2b31205d303a1fd868f

Is that good enough?

metmarkosaric: looks fine to me @aerosol. we get all emails sent to both of these addresses in Help Scout but also happy if you want to pick one and standardize all the messages in the app. 

we typically link to https://plausible.io/contact as there people can find links to the docs/faq. this helps reduce the number of emails we got and works for more generic error messages. when the error is very specific such as this case where people actually need to contact us, it's best experience to just give them the email address directly

vinibrsl (approved):

PR #2608, opened 2023-01-18
Save city name when importing from GA

vinibrsl:
## Summary

This commit adds city data to imported records from Google Analytics. The current implementation sets city to 0 because GA does not use the GeoNames database.

Google Analytics Reporting API uses [Geographical IDs](https://developers.google.com/analytics/devguides/collection/protocol/v1/geoid) to identify cities and countries. Plausible uses [GeoNames](https://geonames.org/) and I couldn't find databases corelating the two.

Fortunately, GA also returns the city name and this commit uses the city name and the country ISO code to find the Geoname ID. To avoid making expensive ETS searches, I created another ETS table in the Location library that uses {country, city} as a key.

Related PR: https://github.com/plausible/location/pull/3

## Preview

I imported from a test site Marko shared with me, and >60% of the events were identified with its GeoNames IDs. This means this strategy does not work in 100% of the cases as it matches by a string, but results are promising still.

Screenshot 2023-02-10 at 11 26 03


lib/plausible/imported/site.ex
```diff
@@ -98,12 +98,24 @@ defmodule Plausible.Imported do
...
+
+    city = Location.City.get_city(city_name, country_code) || %{}
+
     %{
       site_id: site_id,
       date: get_date(row),
-      country: row.dimensions |> Map.fetch!("ga:countryIsoCode") |> default_if_missing(""),
+      country: country_code,
       region: row.dimensions |> Map.fetch!("ga:regionIsoCode") |> default_if_missing(""),
-      city: 0,
```
vinibrsl, line 106: I haven't added tests yet because I wanted to validate this experiment with you first.

lib/plausible/google/report_request.ex
```diff
@@ -57,7 +57,7 @@ defmodule Plausible.Google.ReportRequest do
       },
       %__MODULE__{
         dataset: "imported_locations",
-        dimensions: ["ga:date", "ga:countryIsoCode", "ga:regionIsoCode"],
+        dimensions: ["ga:date", "ga:countryIsoCode", "ga:regionIsoCode", "ga:city"],
```
vinibrsl, line 60: `ga:city` is the Google Analytics dimension that returns the city name in English, e.g. `Warsaw`, `Rio de Janeiro`, `Tallinn`.

lib/plausible/imported/site.ex
```diff
@@ -98,12 +98,24 @@ defmodule Plausible.Imported do
...
+      row.dimensions
+      |> Map.fetch!("ga:city")
+      |> default_if_missing("")
+
+    country_code =
+      row.dimensions
+      |> Map.fetch!("ga:countryIsoCode")
+      |> default_if_missing("")
+
+    city = Location.City.get_city(city_name, country_code) || %{}
```
vinibrsl, line 111: This attempts to get the GeoName record by city name and country code. I had to create another ETS table for this to avoid "querying" using `:ets.match`. My thinking is that this would be called a lot for each row and would probably slow down imports. We have imports with millions of rows.

ukutaht:
> I imported from a test site Marko shared with me, and >60% of the events were identified with its GeoNames IDs

Instead of trying to find a Geoname ID for the imported city name, did you consider leaving the city name as is during import? With this approach we would use Geoname lookups only when querying cities from native Plausible data, but for GA data we would trust the city name that it provides us and use it directly.

vinibrsl:
> > I imported from a test site Marko shared with me, and >60% of the events were identified with its GeoNames IDs
> 
> Instead of trying to find a Geoname ID for the imported city name, did you consider leaving the city name as is during import? With this approach we would use Geoname lookups only when querying cities from native Plausible data, but for GA data we would trust the city name that it provides us and use it directly.

That's an interesting idea! I tried playing with that approach, but I figured the issue becomes querying, instead of importing the data. Filtering the dashboard works with GeoNames, and we would have to adapt that too. It ended up being more complex than I was expecting, and I don't think it's worth given that we're excepting less GA imports in the future.

RobertJoonas:
@vinibrsl

>  That's an interesting idea! I tried playing with that approach, but I figured the issue becomes querying, instead of importing the data. Filtering the dashboard works with GeoNames, and we would have to adapt that too. It ended up being more complex than I was expecting, and I don't think it's worth given that we're excepting less GA imports in the future, at the cost of being less assertive due the string matching.

But GA data will only be shown if no filters are applied, right? Here's [the code](https://github.com/plausible/analytics/blob/master/lib/plausible/stats/query.ex#L336) that does this I believe. And I think filtering imported data is something that we cannot allow in any case. So maybe showing the cities as they come from GA would still be the easier approach, when we can totally ignore any filter scenarios?

vinibrsl:
> And I think filtering imported data is something that we cannot allow in any case.

I didn't know that! Thanks for sharing, @RobertJoonas. In that case I think it makes sense to import city names directly instead of trying to find their GeoName ID.

As we're trying a different approach, I force-pushed this pull request. I have three comments:

- At first I thought I could just save city names to the `imported_locations.city` field, but this field is an integer on ClickHouse. I couldn't change the type because it's a primary key, and I couldn't remove the field either for the same reason. The solution for that was creating a new field: `imported_locations.city_name` and keeping the old field.
- The way we're building the imported query in Elixir expects imported fields to have the same name as native tables. I had to make some small changes to the query building to make this work.
-  I also couldn't make country flags work in the UI, because the country information comes from the GeoName ID.

Let me know what you think!

/cc @ukutaht @RobertJoonas

lib/oban_error_reporter.ex
```diff
@@ -7,18 +11,21 @@ defmodule ObanErrorReporter do
...
     extra = Map.take(meta, ~w(channel payload)a)
 
+    Logger.warn("Failed to complete Oban job. Reason: #{inspect(meta.reason)}")
     Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
   end
 
   def handle_event([:oban, :plugin, :exception], _timing, meta, _) do
     extra = Map.take(meta, ~w(plugin)a)
 
+    Logger.warn("Failed to complete Oban job. Reason: #{inspect(meta.reason)}")
```
aerosol, line 28: Are those warnings necessary and appropriate given that execution errors are already sent to Sentry? Or perhaps we should use error-level logging to do both via a single call?

lib/plausible/imported/site.ex
```diff
@@ -103,7 +103,7 @@ defmodule Plausible.Imported do
       date: get_date(row),
       country: row.dimensions |> Map.fetch!("ga:countryIsoCode") |> default_if_missing(""),
       region: row.dimensions |> Map.fetch!("ga:regionIsoCode") |> default_if_missing(""),
-      city: 0,
+      city_name: row.dimensions |> Map.fetch!("ga:city") |> default_if_missing(),
```
aerosol, line 106: Isn't `default_if_missing()` redundant if the same can be done with `Map.get("ga:city")` and `Map.get("ga:regionIsoCode", "")` ?

lib/plausible/stats/imported.ex
```diff
@@ -119,6 +120,13 @@ defmodule Plausible.Stats.Imported do
       )
       |> select_imported_metrics(metrics)
 
+    imported_q =
```
aerosol, line 123: maybe extract this to `|> maybe_group_by_city()` and get rid of that credo suppression? I assume this is what's causing it? :thinking:

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -832,22 +832,23 @@ defmodule PlausibleWeb.Api.StatsController do
...
-            country_flag: country_info.flag
-          })
-        else
-          Logger.warning("Could not find city info - code: #{inspect(city[:code])}")
-
-          Map.merge(city, %{name: "N/A"})
-        end
+      |> transform_keys(%{city: :code, city_name: :name})
+      |> Enum.map(fn
+        %{name: name} = city when name != nil and name != "" ->
```
aerosol, line 837: is the intention to crash here when name is `nil` or `""`?

aerosol (commented):
Comments/questions inline

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -832,22 +832,23 @@ defmodule PlausibleWeb.Api.StatsController do
...
+          if city_info = Location.get_city(code) do
+            country_info = get_country(city_info.country_code)
+
+            Map.merge(city, %{
+              name: city_info.name,
+              country_flag: country_info.flag
+            })
+          else
+            Logger.warning("Could not find city info - code: #{inspect(city[:code])}")
+            Map.merge(city, %{name: "N/A"})
```
aerosol, line 850:
```suggestion
            Map.put(city, %{name: "N/A"})
```

priv/clickhouse_repo/migrations/20230124140348_add_city_name_to_imported_locations.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.ClickhouseRepo.Migrations.AddCityNameToImportedLocations do
```
aerosol, line 1: Do we need any extra steps in order to run this migration? Shouldn't it be shipped in a separate PR?

lib/plausible_web/controllers/api/stats_controller.ex
vinibrsl: Thanks, applied in 469c937080f275ecb0ec083886dcb62ed0685788

lib/oban_error_reporter.ex
vinibrsl: Oops! Those warnings are not necessary. I was using for debugging locally. Removed in ee93b8207465f24b6dcff552e351b3fa9efafcfc.

lib/plausible/stats/imported.ex
vinibrsl: I tried but the problem is that this function was already too complex, and Credo started failing because we only check changed files on CI. I don't think it's worth refactoring it now, as we don't intend to have Google Analytics imports for much longer.

lib/plausible_web/controllers/api/stats_controller.ex
vinibrsl: This guard should match when `name` is binary and not empty. I rewrote this statement in 3e0a8fa7130378bf91ae909577f8430472ac1ef8. It should be more clear now.

priv/clickhouse_repo/migrations/20230124140348_add_city_name_to_imported_locations.exs
vinibrsl: Good catch! I totally missed this. This PR should go first: https://github.com/plausible/analytics/pull/2633/files

lib/plausible_web/controllers/api/stats_controller.ex
aerosol: ah cool, got it now.

non-empty binary could be matched like this:

```elixir
<<_, _::binary>>
```

vinibrsl: Oh cool! I didn't know that. I'm not sure why does it work though. Do you have any links I can read on this? I'm always confused when pattern matching binaries.

aerosol: > I'm not sure why does it work though.

match a byte and "maybe rest" (just like list's tail in `[h | t]`)

https://elixir-lang.org/getting-started/binaries-strings-and-char-lists.html#binaries

aerosol: https://hexdocs.pm/elixir/main/Kernel.SpecialForms.html#%3C%3C%3E%3E/1

vinibrsl: Changed in d71326b730607cc84e2374616a14f4960a967fa7. Thanks :)

vinibrsl:
I changed the strategy back to finding and saving GeoNames for imported data. It turns out that this will have the best user experience allowing aggregations and country flags. Additionally it will make data more consistent as we'll keep using GeoNames for both imported and native stats. Code is cleaner too.

Tests are failing because https://github.com/plausible/location/pull/3 is not merged yet. As soon it's merged I'll bump the library version and tests should be fixed.

ukutaht (approved):

Issue #2514, opened 2022-12-08
Manual width mode is always enabled


martinspenkevics:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

After upgrading to version 1.5 the embedded dashboard takes full page width even if not using the manual width mode as described in https://plausible.io/docs/embed-dashboard#manual-width-mode

It seems the class "max-width-screen-lg" has no styles associated with it.

### Expected behavior

The manual width mode should work only when enabled.

### Environment

```markdown
- OS: Windows
- Browser: Microsoft Edge
- Browser Version: 108
```

PR #2610, opened 2023-01-18
fix class name typo and include it in safelist

ruslandoga:
### Changes

Fixes https://github.com/plausible/analytics/issues/2514

This PR fixes the class name used for embedded dashboards and includes it in a safelist to avoid purging.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/views/stats_view.ex
+++ b/lib/plausible_web/views/stats_view.ex
@@ -52,7 +52,7 @@ defmodule PlausibleWeb.StatsView do
   def stats_container_class(conn) do
     cond do
       conn.assigns[:embedded] && conn.assigns[:width] == "manual" -> ""
-      conn.assigns[:embedded] -> "max-width-screen-lg mx-auto"
+      conn.assigns[:embedded] -> "max-w-screen-lg mx-auto"
       !conn.assigns[:embedded] -> "container"
     end
   end
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -5,6 +5,11 @@ module.exports = {
     './js/**/*.js',
     '../lib/plausible_web/templates/**/*.html.eex',
   ],
+  safelist: [
+    // PlausibleWeb.StatsView.stats_container_class/1 uses this class
+    // it's not used anywhere else in the templates or scripts
+    "max-w-screen-lg"
+  ],
   darkMode: 'class',
   theme: {
     container: {


lib/plausible_web/views/stats_view.ex
```diff
@@ -52,7 +52,7 @@ defmodule PlausibleWeb.StatsView do
   def stats_container_class(conn) do
     cond do
       conn.assigns[:embedded] && conn.assigns[:width] == "manual" -> ""
-      conn.assigns[:embedded] -> "max-width-screen-lg mx-auto"
+      conn.assigns[:embedded] -> "max-w-screen-lg mx-auto"
```
ruslandoga, line 55: not sure if it was in the tailwind 1.0 -> 2.0 migration or just a typo, but there doesn't seem to be a `max-width-screen-lg` class anymore: https://tailwindcss.com/docs/max-width

vinibrsl (commented):
One concern: the embed dashboard is quite difficult to test because people use it in very different setups. If this class wasn't being applied, what do you think of actually removing it to keep to same behavior?

ruslandoga:
I know @metmarkosaric uses the embedded dashboard: https://markosaric.com/stats/ so we can ask him what to do :)

metmarkosaric:
looks the same to me? any difference that i don't see?

ruslandoga:
Right now it looks like this (full width):

Screenshot 2023-01-19 at 22 07 29

metmarkosaric:
ah ok so with this change we don't allow people to choose the width or if it's not specified we show narrow width rather than full width by default?

ruslandoga:
> if it's not specified we show narrow width rather than full width by default

This.

metmarkosaric:
hmm i'm unsure if this breaks the UI and design of some people that are happy with the full width. is there a reason to change this at all? seems fine to me the way it is. full width by default but you get to pick a manual width if you prefer that. it's already documented too

ruslandoga:
> is there a reason to change this at all?

Full width by default is a bug: https://github.com/plausible/analytics/issues/2514. I only fixed the bug :)

metmarkosaric:
ok cool. i guess it's fine if this change wasn't intentional but a mistake. it's not something we discussed either. i'll let Vini decide

metmarkosaric:
narrow by default does look nicer!

vinibrsl:
Thanks, @ruslandoga!

Issue #1666, opened 2022-02-11
More pageviews in the dashboard with a region filter than no filters. 


whscienceteacher:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Describe the bug

In my current online dashboard, I see more pageviews when I choose a country or region filter than with no filters. I cannot think of a reason why this should be the case. 

[Example with filter](http://whscience.org/bugreports/plausible2022-02-11093710.png) 357 pageviews

[Without filter](https://whscience.org/bugreports/plausible2022-02-11093720.png) 139 pageviews

### Expected behavior

Adding filters should only decrease the number of pageviews. 

### Screenshots

![plausible2022-02-11093710](https://user-images.githubusercontent.com/69084880/153624153-cd56f243-b76b-41b6-b6da-f81d5fcef67a.png)
![plausible2022-02-11093720](https://user-images.githubusercontent.com/69084880/153624165-4da61698-1b25-4a3c-876e-8c37716c8945.png)

### Environment

```markdown
- OS: Windows
- Browser: Edge
- Browser Version: 98.0.1108.43
```

ukutaht:
Thanks for the report. Clearly a bug.

metmarkosaric:
thanks @whscienceteacher! do you still experience this issue?

metmarkosaric:
closing this as cannot replicate the issue but please do share more details if this happens again. thanks!

sptv2reg:
I'm also seeing this bug on our trial:

![image](https://user-images.githubusercontent.com/103186488/213391494-e1dd92fb-247e-47ce-9362-2420fe6916fb.png)

With country filter:

![image](https://user-images.githubusercontent.com/103186488/213391554-ea6f64e6-3486-4e73-a020-a85a743b640f.png)

metmarkosaric:
thanks for reporting this! our developers were able to identify the issue. it seems to be happening on the reports for "today" and "realtime". we're working on a fix and will update here when it's live

metmarkosaric:
Just a quick update that this fix is now online. You might notice a faster loading dashboard after this change too. Thanks again for reporting it and for helping us make Plausible better for everyone!

PR #2612, opened 2023-01-19
Fix returning more pageviews with a visit property filter

RobertJoonas:
### Changes

This PR fixes https://github.com/plausible/analytics/issues/1666. The bug was that we were not querying the CollapsingMergeTree table engine correctly, hence the number of pageviews (also custom events) was wrongly inflated when looking at recent periods (`today`, `realtime`).

As the 'sessions' table is using the CollapsingMergeTree engine, we have to select session_id's distinctively in the join subquery. Otherwise we will get multiple rows (with sign -1 and 1) as long as the background merge hasn't happened.

### Tests

### Changelog
- [x] Not sure whether to log this change

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -150,7 +150,7 @@ defmodule Plausible.Stats.Base do
       from(
         s in "sessions",
         where: s.domain == ^site.domain,
-        where: s.start >= ^first_datetime and s.start < ^last_datetime,
+        where: s.start >= ^first_datetime and s.start < ^last_datetime
       )
       |> add_sample_hint(query)
 
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -14,7 +14,9 @@ defmodule Plausible.Stats.Base do
       sessions_q =
         from(
           s in query_sessions(site, query),
-          select: %{session_id: fragment("DISTINCT ?", s.session_id)}
+          select: %{session_id: s.session_id},
+          where: s.sign == 1,
+          group_by: s.session_id
         )
 
       from(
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/base.ex
```diff
@@ -14,7 +14,7 @@ defmodule Plausible.Stats.Base do
       sessions_q =
         from(
           s in query_sessions(site, query),
-          select: %{session_id: s.session_id}
+          select: %{session_id: fragment("DISTINCT ?", s.session_id)}
```
aerosol, line 17: Is there a reason for not using `where: s.sign == 1` in `query_sessions/2` instead? 🤔 I would expect this to be more efficient.

aerosol: Unless `where: s.sign == 1` does not guarantee unique results? 🤔

aerosol (approved):

aerosol (commented):
We'll run some perf comparisons before deploying this

lib/plausible/stats/base.ex
```diff
@@ -148,7 +150,7 @@ defmodule Plausible.Stats.Base do
       from(
         s in "sessions",
         where: s.domain == ^site.domain,
-        where: s.start >= ^first_datetime and s.start < ^last_datetime
+        where: s.start >= ^first_datetime and s.start < ^last_datetime,
```
aerosol, line 153: :thinking: unnecessary change?

RobertJoonas: Yup, removed in https://github.com/plausible/analytics/pull/2612/commits/e94fe8a9d9d9ac63dd4927f5aff2c21490a18c11

PR #2627, opened 2023-01-26
Remove the ability to collapse the main graph + transition bug fix

vinibrsl:
This PR removes the ability to collapse the main graph. The graph collapsed whenever `metric` was a falsy value (`undefined`, `null`, `""`). I removed all code related to that. `metric` should default to `visitors` always.

We want to add new items to top stats, and this PR will make it easier to change it. Also, there's currently a bug where top stats is randomly collapsing, which should be fixed by this PR.

This pull request also fixes a bug where the graph wasn't fading out when changing metrics.

## Preview

[preview.webm](https://user-images.githubusercontent.com/5093045/215574524-83b307ce-8dd8-4130-952b-c7ff7f748acf.webm)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ All notable changes to this project will be documented in this file.
 - Always show direct traffic in sources reports plausible/analytics#2531
 - Stop recording XX and T1 country codes plausible/analytics#2556
 
+### Removed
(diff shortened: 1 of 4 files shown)


aerosol (approved):
Very hard to track all those state containers, but I've run it locally and it looked good with no errors

PR #2628, opened 2023-01-28
Add case to shift_back function to handle day periods

### LINKED ISSUE #1650: Statistics relative to previous period are misleading
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

On the dashboard, under "Unique visitors", is a percentage showing how today's numbers compare to yesterday's.

Unfortunately, it seems these are based on the entire stats of the previous day, rather than just those at the current time. At least, I think that's the case.

### Expected behavior

These "previous period" statistics take into account how far through the current period we are.

For example, when looking at the "Day" statistics, and it's 4pm, it should only consider stats for yesterday up to 4pm, so the data is actually comparable.

Otherwise, the values aren't exceptionally useful unless going through extreme growth / decline in traffic.

### Screenshots

image

### Environment

```markdown
N/A
```

djanda97:
### Changes

Fixes #1650
This PR adds a case to the `shift_back` function in `lib/plausible/stats/query.ex` to handle day periods.

### Tests

### Changelog

### Documentation

### Dark mode


djanda97:
Since the `shift_back` function handles month and year periods similarly, I tried following this format and adding another case for the function with the relevant parts swapped out for the day period. I haven't added an automated test case for these changes yet, but I wanted to open this PR to check if I'm on the right track. I would appreciate any feedback :smile:

djanda97:
Hi @vinibrsl! It has been a while since I've worked on this issue plus there wasn't much code that was included in this PR, so I agree that it would be best to start over. I'm still interested in working on #1650, so I'll take a look through `analytics/lib/plausible/stats/comparisons.ex` and its associated tests. After taking a closer look, I'll post here with any questions that I have! :smile:

djanda97:
Hi @vinibrsl, I haven't found the time to work on updating this PR.

PR #2629, opened 2023-01-28
Changes to the site settings danger zone

metmarkosaric:
The idea is to:

1. Show a transfer site ownership message to owners of the site only. This will make ownership transfer feature more discoverable. Will also help people transfer ownership of a site rather than deleting the whole thing when that makes more sense
2. Add a note that site deletion is a delayed operation
3. Make some other tiny copy changes

What did I break? :)

### DIFF
--- a/lib/plausible_web/templates/site/settings_danger_zone.html.eex
+++ b/lib/plausible_web/templates/site/settings_danger_zone.html.eex
@@ -33,7 +33,7 @@
     <div class="border-b border-gray-200 dark:border-gray-500"></div>
 
     <li class="py-4 flex items-center justify-between space-x-4">
-      <div class="flex flex-col">
+      <div class="flex max-w-md flex-col">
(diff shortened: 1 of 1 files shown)


lib/plausible_web/templates/site/settings_danger_zone.html.eex
```diff
@@ -23,7 +38,7 @@
         Delete site
         </p>
         <p class="text-sm leading-5 text-gray-500 dark:text-gray-200">
-        Removes all stats along with the site configuration
+        Permanently remove all stats and the site configuration too. This process takes a few hours so you won't be able to immediately register the same site again
```
aerosol, line 41: This doesn't look to good, let me see if I can fix it.

image

aerosol: image

after https://github.com/plausible/analytics/pull/2629/commits/b61b202ede196d5aac4356673f220380dabdd8c3

aerosol (approved):

PR #2637, opened 2023-02-01
Interval label flicking when changing metric

vinibrsl:
This commit fixes a bug where the interval label flickers when changing the dashboard metric. This was caused because changing the metric clears the `graphData` state property until the HTTP request is finished.

This change fixes it by falling back `interval` to the value stored in localStorage when `graphData` is falsy.

## Bug preview (see "Days")

[preview.webm](https://user-images.githubusercontent.com/5093045/216039004-e3803078-1a73-4364-a050-ba748f179a07.webm)

### DIFF
--- a/assets/js/dashboard/stats/graph/interval-picker.js
+++ b/assets/js/dashboard/stats/graph/interval-picker.js
@@ -54,7 +54,7 @@ export function IntervalPicker({ graphData, query, site, updateInterval }) {
   const menuElement = React.useRef(null)
   subscribeKeybinding(menuElement)
 
-  const currentInterval = graphData?.interval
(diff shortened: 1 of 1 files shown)


aerosol:
hmm it still flickers for me with this patch

https://user-images.githubusercontent.com/173738/216056960-c9799202-7045-4e11-86a0-73f7d212f130.mp4

vinibrsl:
> hmm it still flickers for me with this patch
> 
>  flicker-2023-02-01_14.30.50.mp4

Ohhh... thanks for testing! The interval is saved in localStorage when you actually select an interval, and in your case it looks like it's `undefined`.

I guess we'll need to figure out a better way of doing this. Maybe saving the interval in the `query` state property so we'll have it available while the HTTP request is not ready yet.

PR #2642, opened 2023-02-03
Update sentry server_name handling

cnkk:
### Changes

Update sentry server_name handling.

Our infrastructure now sets the correct hostname instead of the mutable container name.
The changes introduced in #2317 to set the APP_HOST env var as server_name also did not have any effect, because we server_name needs to be set as an option (not as a tag).

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl (approved):
Good catch!

PR #2649, opened 2023-02-06
Support alternative mailing services (Mailgun, Mandrill, Sendgrid)

ruslandoga:
### Changes

This PR is similar to https://github.com/plausible/analytics/pull/2596 but hardcodes all adapters that bamboo ships with and adds some config tests.

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated https://github.com/plausible/docs/pull/351

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -278,13 +278,32 @@ config :plausible, Plausible.ClickhouseRepo,
 case mailer_adapter do
   "Bamboo.PostmarkAdapter" ->
     config :plausible, Plausible.Mailer,
-      adapter: :"Elixir.#{mailer_adapter}",
+      adapter: Bamboo.PostmarkAdapter,
       request_options: [recv_timeout: 10_000],
       api_key: get_var_from_path_or_env(config_dir, "POSTMARK_API_KEY")
 
+  "Bamboo.MailgunAdapter" ->
+    config :plausible, Plausible.Mailer,
+      adapter: Bamboo.MailgunAdapter,
+      hackney_opts: [recv_timeout: :timer.seconds(10)],
+      api_key: get_var_from_path_or_env(config_dir, "MAILGUN_API_KEY"),
+      domain: get_var_from_path_or_env(config_dir, "MAILGUN_DOMAIN")
+
+  "Bamboo.MandrillAdapter" ->
+    config :plausible, Plausible.Mailer,
+      adapter: Bamboo.MandrillAdapter,
+      hackney_opts: [recv_timeout: :timer.seconds(10)],
+      api_key: get_var_from_path_or_env(config_dir, "MANDRILL_API_KEY")
+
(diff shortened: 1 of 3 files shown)


config/runtime.exs
```diff
@@ -275,16 +275,37 @@ config :plausible, Plausible.ClickhouseRepo,
   flush_interval_ms: ch_flush_interval_ms,
   max_buffer_size: ch_max_buffer_size
 
+mailer_hackney_opts = [recv_timeout: :timer.seconds(10)]
+
 case mailer_adapter do
   "Bamboo.PostmarkAdapter" ->
     config :plausible, Plausible.Mailer,
-      adapter: :"Elixir.#{mailer_adapter}",
-      request_options: [recv_timeout: 10_000],
```
ruslandoga, line 282: I couldn't find `request_options` in bamboo:

- https://github.com/thoughtbot/bamboo/search?q=request_options
- https://github.com/thoughtbot/bamboo/search?q=hackney_opts

ruslandoga: 🙈 https://github.com/pablo-co/bamboo_postmark/blob/c8b8cbf6103fc38a8ad467a3ab0eb6ebe44a84e6/lib/bamboo/postmark_adapter.ex#L194

config/runtime.exs
```diff
@@ -278,13 +278,32 @@ config :plausible, Plausible.ClickhouseRepo,
...
     config :plausible, Plausible.Mailer,
-      adapter: :"Elixir.#{mailer_adapter}",
+      adapter: Bamboo.PostmarkAdapter,
       request_options: [recv_timeout: 10_000],
       api_key: get_var_from_path_or_env(config_dir, "POSTMARK_API_KEY")
 
+  "Bamboo.MailgunAdapter" ->
+    config :plausible, Plausible.Mailer,
+      adapter: Bamboo.MailgunAdapter,
+      hackney_opts: [recv_timeout: :timer.seconds(10)],
```
ruslandoga, line 288: So I used `hackney_opts` in the built-in adapters.

ukutaht (approved):

tenfourty:
Hi folks,
Thank you for this work, it really adds some flexibility for folks wanting to use other services to send email.

In my case I was looking at using this for Mailgun, unfortunately I don't think this will work for folks who use Mailgun's EU servers.

I had a look at the source code for the implementation, and you might need to add the `base_uri` for Mailgun as well. Mailgun makes a difference in the API base URL between sender domains from within the EU and outside.

You can read more about this in the Bamboo Adapter docs [here](https://hexdocs.pm/bamboo/Bamboo.MailgunAdapter.html#module-api-base-uri-configuration).

I'll try to put a comment inline to the PR as well.

Do let me know if you want me to create a new issue for this. I felt it was important to include this as this is included in the v2 RC and it would be great to have this fix included.

PR #2650, opened 2023-02-06
First pass at Comparisons

vinibrsl:
### Changes

This pull request adds support for comparing the actual showed period on the main graph with the previous one. This is a first pass and it's hidden under a feature flag because it's not feature complete yet as we want to support other comparison modes.

image

Next PRs will include some refactoring, especially in the graph-utils.js file which is calling for a split. And after that I'll push comparison modes.

Reviewing commit-by-commit may help understanding some of my decisions.

### Tests
Not sure if it's worth testing the current behavior as next PRs will include comparison modes (previous period, year over year, custom date).

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/graph-util.js
+++ b/assets/js/dashboard/stats/graph/graph-util.js
@@ -149,11 +149,7 @@ export const GraphTooltip = (graphData, metric, query) => {
 }
 
 const truncateToPresentIndex = function(array, presentIndex) {
-  if (presentIndex) {
-    for (var i = presentIndex; i < array.length; i++) {
-      array[i] = undefined
-    }
-  }
+  return array.slice(0, presentIndex)
 }
 
 export const buildDataSet = (plot, comparisonPlot, present_index, ctx, label) => {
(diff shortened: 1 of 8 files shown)


assets/js/dashboard/stats/graph/graph-util.js
```diff
@@ -124,24 +148,52 @@ export const GraphTooltip = (graphData, metric, query) => {
   }
 }
 
-export const buildDataSet = (plot, present_index, ctx, label, isPrevious) => {
+const truncateToPresentIndex = function(array, presentIndex) {
```
ukutaht, line 151: Currently this seems to mutate a JS data structure that was returned from the API. I think making operations like this immutable is the way to go so we don't get unexpected errors from mutable data.

ukutaht (approved):

assets/js/dashboard/stats/graph/graph-util.js
vinibrsl: Good call! Changed in 2b47ac7f243fa0258617d60492851c54c090e477 :)

Issue #2640, opened 2023-02-02
Transfer ownership to invited user results in HTTP 500


Popescu-PfeifferMarc:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

Transferring ownership to another user, who hasn't accepted the join request results in HTTP 500. 

### Expected behavior

An error message is shown, e.g. "this user hasn't accepted the join request".

Popescu-PfeifferMarc:
The issue was resolved by the invited user accepting the invite. I just wanted to suggest adding a more descriptive error message to help future users running into the same problem

Popescu-PfeifferMarc:
> 👋 @Popescu-PfeifferMarc
> 
> What Plausible version are you running?
> 
> ~A similar issue #2455 has been fixed in [v1.5.](https://github.com/plausible/analytics/discussions/2507)~

It should be the latest release version of plausible (v1.5.1)

PR #2651, opened 2023-02-07
improve transfer ownership error message

ruslandoga:
### Changes

Fixes https://github.com/plausible/analytics/issues/2640 by adding a more descriptive error message in case a transfer to an invited (but not joined) user is requested.

Screenshot 2023-02-13 at 11 40 58

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -116,22 +116,39 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+          |> Plausible.Mailer.send()
+
+          put_flash(conn, :success, "Site transfer request has been sent to #{email}")
+
+        {:error, changeset} ->
+          errors = Ecto.Changeset.traverse_errors(changeset, fn {msg, _} -> msg end)
+
+          message =
+            if errors[:invitation] do
+              "#{email} has already been invited but hasn't yet accepted the join request to #{site_domain}"
```
aerosol, line 143: Can we make this part of the constraint error message and traverse errors in an uniform way so that any extra messages we add in the future don't fall into the mysterious "transfer has failed" bucket?

ruslandoga: Sure. Should it be part of this PR? Re: implementation.

aerosol: Your call 🤗

aerosol (approved):

lib/plausible_web/controllers/site/membership_controller.ex
ruslandoga: Otherwise it'd be hard to distinguish between a double invite and invite -> transwer ownership. I can also use `invitation.role == :owner` but it seems a bit more error prone.

ruslandoga: Or maybe just returning `already invited` for all errors is fine? I also noticed that the flash message disappears very quickly, not enough time to read it if it's long.

aerosol: Yes, I think that's fine. It's clear enough to let the user know it's an existing invitation preventing proceeding. Re: flash message - I think we can double the timeout there: https://github.com/ruslandoga/analytics/blob/master/lib/plausible_web/templates/layout/_flash.html.eex#L37

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -116,22 +116,38 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+
+          put_flash(conn, :success, "Site transfer request has been sent to #{email}")
+
+        {:error, changeset} ->
+          errors = Plausible.ChangesetHelpers.traverse_errors(changeset)
+
+          message =
+            case errors do
+              %{invitation: [error | _]} -> String.capitalize(error)
+              _other -> "Site transfer request to #{email} has failed"
```
ruslandoga, line 144: I'm avoiding returning any other errors since I'm not sure if it's safe...

test/plausible_web/controllers/site/membership_controller_test.exs
```diff
@@ -207,6 +207,31 @@ defmodule PlausibleWeb.Site.MembershipControllerTest do
...
+      conn = get(recycle(conn), redirected_to(conn, 302))
+
+      assert html_response(conn, 200) =~
+               "#{invited} has been invited to #{site.domain} as an admin"
+
+      # transferring ownership to that domain now fails
+
+      conn = post(conn, "/sites/#{site.domain}/transfer-ownership", %{email: invited})
+      conn = get(recycle(conn), redirected_to(conn, 302))
+      assert html_response(conn, 200) =~ "Already sent"
```
ruslandoga, line 233: Not sure if that error is descriptive enough, though.

lib/plausible_web/controllers/site/membership_controller.ex
ruslandoga: Re: flash https://github.com/plausible/analytics/pull/2652

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -116,22 +116,40 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+          |> Plausible.Mailer.send()
+
+          put_flash(conn, :success, "Site transfer request has been sent to #{email}")
+
+        {:error, changeset} ->
+          errors = Plausible.ChangesetHelpers.traverse_errors(changeset)
+
+          message =
+            case errors do
+              %{invitation: [error | _]} -> String.capitalize(error)
```
ruslandoga, line 143: `:invitation` is not a schema field, but a custom error key.

```elixir
  |> unique_constraint([:email, :site_id],
      name: :invitations_site_id_email_index,
      error_key: :invitation,
      message: "invitation already exists"
    )
```

Should we still concat the other possible errors?

aerosol: Just the first one is fine. If anything confusing comes up we can address on case by case basis?

aerosol (approved):

aerosol (changes requested):
Tests fail. Please don't force push with ongoing review, it makes it more difficult to track changes. Thx.

ruslandoga:
Sorry, my bad.

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -116,22 +116,41 @@ defmodule PlausibleWeb.Site.MembershipController do
...
+          |> Plausible.Mailer.send()
+
+          put_flash(conn, :success, "Site transfer request has been sent to #{email}")
+
+        {:error, changeset} ->
+          errors = Plausible.ChangesetHelpers.traverse_errors(changeset)
+
+          message =
+            case errors do
+              %{invitation: ["already sent" | _]} -> "Invitation has already been sent"
```
ruslandoga, line 143: I decided to reuse the approach used for duplicate invitations: https://github.com/plausible/analytics/blob/1d1ec9a673ecfa2857b8d43de1d09c7bef440a1a/lib/plausible_web/controllers/site/membership_controller.ex#L84-L85

aerosol (approved):

PR #2652, opened 2023-02-07
add flash ttl

ruslandoga:
### Changes

This PR adds flash `:ttl` to set the duration in ms that flash is shown. Useful for longer messages like in https://github.com/plausible/analytics/pull/2651.

Usage

```elixir
conn
|> put_flash(:ttl, 8000)
|> put_flash(:error, "some long error message")
```

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
Why make it configurable instead of upping the limit globally?

ruslandoga:
Personal preference :) I tried setting the limit higher, to 8000. It looked fine for the flash with the long message, but the flashes with short messages then appeared to be "stuck". I expect the flash to shown about as long as it takes me to read it.

ukutaht:
Fair enough, that makes sense to me :)

PR #2653, opened 2023-02-07
Fix bug with multiple ARRAY JOIN query

### LINKED ISSUE #1789: [Bug on plausible website] Error showing custom event goal
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

Yesterday I opened this issue, but today I was unhappily surprised that somebody without even asking just closed this issue and converted this to a discussion while it's clearly a bug. It seems like the issue is not even read, because it's also suddenly placed in the 'selfhosting' category while it's obviously not about that, but a bug on the plausible website where we view the analytics and was working before.

Please fix this issue. We pay for Plausible every month and expect things to keep working, not suddenly break what was working for at least a year.

--

When clicking on a custom event an endless wait cursor appears and the data never shows. This was working fine in previous versions of Plausible. The code on our website didn't change and is not sending other data so something on the Plausible website did change.

### Expected behavior

Plausible to keep working as before, not suddenly crash on a feature that worked in all previous versions for at least a year.

### Screenshots

screenshot-issue

### Environment

```markdown
- OS: windows
- Browser: ms edge chromium for instance
- Browser Version: latest (99.0.1150.46 (Official build) (64-bit))
```

ukutaht:
### Changes

Fixes #1789

Small change to query building to make sure we don't `JOIN` with `meta` more than once.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -315,20 +315,13 @@ defmodule Plausible.Stats.Breakdown do
     end
   end
 
-  defp do_group_by(
-         %Ecto.Query{
-           from: %Ecto.Query.FromExpr{source: {"events", _}},
-           joins: [%Ecto.Query.JoinExpr{source: {"meta", _}}]
-         } = q,
-         "event:props:" <> prop
-       ) do
-    from(
-      [e, meta] in q,
-      group_by: e.name,
-      where: meta.key == ^prop,
-      group_by: meta.value,
-      select_merge: %{^prop => meta.value},
-      order_by: {:asc, meta.value}
+  defp has_join_with_table(ecto_q, table) do
+    Enum.any?(
+      ecto_q.joins,
+      fn
+        %Ecto.Query.JoinExpr{source: {t, _}} when t == table -> true
+        _ -> false
+      end
     )
   end
 
@@ -336,9 +329,18 @@ defmodule Plausible.Stats.Breakdown do
          %Ecto.Query{from: %Ecto.Query.FromExpr{source: {"events", _}}} = q,
          "event:props:" <> prop
        ) do
+    q =
+      if has_join_with_table(q, "meta") do
+        q
+      else
+        from(
+          e in q,
+          inner_lateral_join: meta in fragment("meta")
+        )
+      end
+
     from(
-      e in q,
-      inner_lateral_join: meta in fragment("meta"),
+      [e, meta] in q,
       where: meta.key == ^prop,
       group_by: meta.value,
       select_merge: %{^prop => meta.value},
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -315,7 +315,7 @@ defmodule Plausible.Stats.Breakdown do
     end
   end
 
-  defp has_join_with_table(ecto_q, table) do
+  defp joins_table?(ecto_q, table) do
     Enum.any?(
       ecto_q.joins,
       fn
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -315,30 +315,32 @@ defmodule Plausible.Stats.Breakdown do
...
-      group_by: e.name,
-      where: meta.key == ^prop,
-      group_by: meta.value,
-      select_merge: %{^prop => meta.value},
-      order_by: {:asc, meta.value}
+  defp has_join_with_table(ecto_q, table) do
+    Enum.any?(
+      ecto_q.joins,
+      fn
+        %Ecto.Query.JoinExpr{source: {t, _}} when t == table -> true
```
aerosol, line 322:
```suggestion
        %Ecto.Query.JoinExpr{source: {^table, _}} -> true
```

aerosol: But also, is the table always a literal source here or are Schema names allowed here too? cc @ruslandoga

lib/plausible/stats/breakdown.ex
```diff
@@ -315,30 +315,32 @@ defmodule Plausible.Stats.Breakdown do
...
-         "event:props:" <> prop
-       ) do
-    from(
-      [e, meta] in q,
-      group_by: e.name,
-      where: meta.key == ^prop,
-      group_by: meta.value,
-      select_merge: %{^prop => meta.value},
-      order_by: {:asc, meta.value}
+  defp has_join_with_table(ecto_q, table) do
```
aerosol, line 318:
```suggestion
  defp joins_table?(ecto_q, table) do
```

lib/plausible/stats/breakdown.ex
```diff
@@ -315,30 +315,32 @@ defmodule Plausible.Stats.Breakdown do
...
+      end
     )
   end
 
   defp do_group_by(
          %Ecto.Query{from: %Ecto.Query.FromExpr{source: {"events", _}}} = q,
          "event:props:" <> prop
        ) do
+    q =
+      if has_join_with_table(q, "meta") do
```
aerosol, line 333:
```suggestion
      if joins_table?(q, "meta") do
```

aerosol: Nitpick: another option would be to make it `queryable |> join_once("meta")`

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -352,6 +352,51 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
                }
              ]
     end
+
+    test "Property breakdown with prop and goal filter", %{conn: conn, site: site} do
```
aerosol, line 356: Nice, I've checked this test without your patch and it fails as expected 👍

aerosol (approved):
minor nitpicks inline, completely optional

lib/plausible/stats/breakdown.ex
ukutaht: There's no schema for the `meta` field so it should not be a concern. And this change didn't break any tests so I assume this match works in every case we've tested for.

lib/plausible/stats/breakdown.ex
ukutaht: I've committed your suggestions.

The problem with extracting a `join_once` function is named bind for `meta`. Currently we do this:

```elixir
q = from(
  e in q,
  inner_lateral_join: meta in fragment("meta")
)

from(
  [e, meta] in q
  # ...
)
```

The name `meta` is significant here and I don't think it's possible to make it dynamic without using a macro for `join_once`.

PR #2654, opened 2023-02-07
Fix unknown bind meta

ukutaht:
### Changes

At some point we changed which data structures are used for props and filters on the `Plausible.Stats.Query` struct. This PR updates some DB query building code to be aligned with the new format.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/props.ex
+++ b/lib/plausible/stats/props.ex
@@ -7,11 +7,11 @@ defmodule Plausible.Stats.Props do
       Enum.find(query.filters, fn {key, _} -> String.starts_with?(key, "event:props:") end)
 
     if prop_filter do
-      {key, val} = prop_filter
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/props.ex
```diff
@@ -7,11 +7,11 @@ defmodule Plausible.Stats.Props do
       Enum.find(query.filters, fn {key, _} -> String.starts_with?(key, "event:props:") end)
 
     if prop_filter do
-      {key, val} = prop_filter
+      {"event:props:" <> key, {_, val}} = prop_filter
```
aerosol, line 10: minor nitpick: https://github.com/plausible/analytics/commit/3aada7c407898a6d95ec4f2c7bc2bac9eec16737

ukutaht: Thanks! cherry-picked

PR #2661, opened 2023-02-09
Split Clickhouse pools into Read-Only and Read/Write (dedicated to writes)

aerosol:
### Changes

This PR separates clickhouse pools into read-only and read-write (introduces `IngestRepo`).

The former becomes read-only so no insert calls will be allowed via its module (compilation-time check is done in Ecto.Repo to prevent insert function exports).
The latter is dedicated to the following writes:
  - session buffer 
  - events buffer 
  - GA import writes

Therefore the working assumption is pool size of 3 for the `IngestRepo` is enough. There should be no more than 3 parallel inserts executed at a single node. 

We normally configure the regular pool size via DB URL `?pool_size` query string parameter. The `IngestRepo` runtime configuration relies on the fact that static config takes precedence, so the same DB URL can be used. In any case the ingestion pool size can be overridden alone through `CLICKHOUSE_INGEST_POOL_SIZE` environment variable.

I don't think this affects self-hosters -- paging @ruslandoga for sanity check :pleading_face:.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/ingest_repo.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.IngestRepo do
+  use Ecto.Repo,
+    otp_app: :plausible,
+    adapter: ClickhouseEcto
+
+  defmacro __using__(_) do
+    quote do
+      alias Plausible.ClickhouseRepo
```
ruslandoga, line 8: should this still be alaised here?

ruslandoga (approved):

lib/plausible/ingest_repo.ex
aerosol: nice, thx https://github.com/plausible/analytics/pull/2661/commits/a568621d9f340d9ab052f69eef218842c2ef2d95

ruslandoga (commented):
I think maybe IngestRepo should also be used to transfer / delete sessions / events. Or is it ok to do via the read-only repo? I guess `query` is defined anyway via [`Ecto.Adapters.SQL`](https://github.com/elixir-ecto/ecto_sql/blob/389f855dbd0a16022b681531b85748488e1635e7/lib/ecto/adapters/sql.ex#L644-L651) so it should be ok.

aerosol:
> I think maybe IngestRepo should also be used to transfer / delete sessions / events. Or is it ok to do via the read-only repo? I guess query is defined anyway via Ecto.Adapters.SQL so it should be ok.

Great catch, let me check that. If possible, I'd rather not let transfers interfere with ingestion. They are currently somewhat broken anyway and most of the time require manual fiddling (takes too long to complete over HTTP in most cases, proper implementation should use chunks, but there's also some technical debt around site identifiers that we'll try to address first to hopefully improve that area).

aerosol:
@ruslandoga the transfer inserts went fine through `Ecto.Adapters.SQL`, thanks for that! (remember our "why test admin code" conversation? tsk tsk 😅)

config/config.exs
```diff
@@ -1,7 +1,7 @@
 import Config
 
 config :plausible,
-  ecto_repos: [Plausible.Repo, Plausible.ClickhouseRepo]
+  ecto_repos: [Plausible.Repo, Plausible.IngestRepo]
```
ukutaht, line 4: Should this also include `Plausible.ClickhouseRepo`?

ukutaht (approved):

config/config.exs
ruslandoga: I think this config is mostly for `mix ecto.*` commands. If we include both repos, I think it would make the mgirations run twice, etc.

ukutaht: Ah yes, that makes sense

aerosol: yep, 100% - insert is necessary to run the migrations, and with `read_only: true` it can't be done

vinibrsl (approved):

PR #2662, opened 2023-02-09
Refactor top stats and graph positioning

RobertJoonas:
### Changes

This PR is the foundation for displaying additional Top Stat metrics on the dashboard. It replaces the manual (absolute) CSS positioning with relative positioning, which significantly simplifies writing code for a varying number of Top Stat metrics.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -463,16 +463,15 @@ export default class VisitorGraph extends React.Component {
 
   render() {
     const {mainGraphLoadingState, topStatsLoadingState} = this.state
-    const mainGraphRefreshing = mainGraphLoadingState === LoadingState.refreshing
 
     const showLoader =
-      LoadingState.isLoadingOrRefreshing(mainGraphLoadingState) ||
-      topStatsLoadingState === LoadingState.loading
+      [mainGraphLoadingState, topStatsLoadingState].includes(LoadingState.loading) &&
+      mainGraphLoadingState !== LoadingState.refreshing
 
     return (
       <LazyLoader onVisible={this.onVisible}>
         <div className={"relative w-full mt-2 bg-white rounded shadow-xl dark:bg-gray-825"}>
-          {showLoader && !mainGraphRefreshing && renderLoader()}
+          {showLoader && renderLoader()}
           {this.renderInner()}
         </div>
       </LazyLoader>
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -11,6 +11,7 @@ import TopStats from './top-stats';
 import { IntervalPicker, getStoredInterval, storeInterval } from './interval-picker';
 import FadeIn from '../../fade-in';
 import * as url from '../../util/url'
+import classNames from 'classnames';
 
 class LineGraph extends React.Component {
   constructor(props) {
@@ -287,7 +288,7 @@ class LineGraph extends React.Component {
 
   render() {
     const { onlyGraphLoading, updateMetric, metric, topStatData, query, site, graphData } = this.props
-    const extraClass = this.props.graphData && this.props.graphData.interval === 'hour' ? '' : 'cursor-pointer'
+    const canvasClass = classNames('mt-4 select-none', {'cursor-pointer': !['minute', 'hour'].includes(graphData?.interval)})
 
     return (
       <div>
@@ -304,7 +305,7 @@ class LineGraph extends React.Component {
           </div>
           <FadeIn show={graphData}>
             <div className="relative h-96 w-full">
-              <canvas id="main-graph-canvas" className={'mt-4 select-none ' + extraClass}></canvas>
+              <canvas id="main-graph-canvas" className={canvasClass}></canvas>
             </div>
           </FadeIn>
         </div>
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -11,7 +11,6 @@ import TopStats from './top-stats';
 import { IntervalPicker, getStoredInterval, storeInterval } from './interval-picker';
 import FadeIn from '../../fade-in';
 import * as url from '../../util/url'
-import classNames from "classnames";
 
 class LineGraph extends React.Component {
   constructor(props) {
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -287,7 +287,7 @@ class LineGraph extends React.Component {
   }
 
   render() {
-    const { onlyGraphLoading, updateMetric, metric, topStatData, query, site, graphData } = this.props
+    const { mainGraphRefreshing, updateMetric, metric, topStatData, query, site, graphData } = this.props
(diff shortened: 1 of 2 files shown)


assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -423,39 +445,46 @@ export default class VisitorGraph extends React.Component {
...
+        <div className={"relative w-full mt-2 bg-white rounded shadow-xl dark:bg-gray-825"}>
+          {showLoader && !mainGraphRefreshing && renderLoader()}
           {this.renderInner()}
         </div>
       </LazyLoader>
     )
   }
 }
+
+function renderLoader() {
```
RobertJoonas, line 482: This will render a loader icon exactly in the middle of a `position: relative` parent element. The function is called in two places because we want it to have a different parent in order to be placed correctly.

ukutaht (approved):
Looks good to me, I like getting rid of that crazy padding stuff that was used before.

Not relevant for this PR but this review made me think that maybe it's appropriate for the frontend to have some more knowledge. Currently it's the backend that decides which values to show in the top stats container. But what if the frontend made that decision and requested the appropriate top stats from the backend itself?

It may simplify parts of the code. The frontend would always know which top stats it will be showing. This would allow us to always have the correct container height during loading. I think it would also allow us to remove the `resetMetric` callback. Thoughts @RobertJoonas @vinibrsl?

RobertJoonas:
@ukutaht good idea! It would totally make sense if the `TopStats` component itself would be responsible for fetching the metrics allowed. It could also define which metrics can and cannot be selected. Basically, we could move the whole metric selection logic to `TopStats` (it won't have to share the state with the graph anymore, since it's only in localStorage now).

But probably still a good idea to merge this one as it is now, and do these changes in another PR, right?

ukutaht:
> But probably still a good idea to merge this one as it is now, and do these changes in another PR, right?

Yes, I think it's completely independent from any other work. Moving the logic is quite difficult and there may be some roadblocks along the way that we haven't thought about. Go ahead and continue with the current batch of work. Once the new metrics are added, we can have an experimental pairing session on refactoring the top stats logic :)

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -273,24 +273,39 @@ class LineGraph extends React.Component {
...
           <div className="absolute right-4 -top-10 py-2 md:py-0 flex items-center">
             { this.downloadLink() }
             { this.samplingNotice() }
             { this.importedNotice() }
             <IntervalPicker site={site} query={query} graphData={graphData} metric={metric} updateInterval={this.props.updateInterval}/>
           </div>
           <FadeIn show={graphData}>
-            <canvas id="main-graph-canvas" className={'mt-4 select-none ' + extraClass} width="1054" height="342"></canvas>
+            <div className="relative h-96 w-full">
+              <canvas id="main-graph-canvas" className={'mt-4 select-none ' + extraClass}></canvas>
```
vinibrsl, line 307: Not strictly related to the graph positioning, but what do you think of renaming `extraClass`? As I understand this variable styles the cursor when hovering the graph to make clear it's clickable. The only exception is when interval is hourly as the graph is not clickable when viewing this interval.

I suggest using `classNames` [(example here)](https://github.com/plausible/analytics/blob/3b73836136a1816f2c9847969e966e055cf142f7/assets/js/dashboard/stats/graph/top-stats.js#L80) or renaming this variable to be more descriptive, for example `clickableCursorUnlessHourly`.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -423,39 +445,46 @@ export default class VisitorGraph extends React.Component {
 
     const theme = document.querySelector('html').classList.contains('dark') || false
 
-    const mainGraphRefreshing = (mainGraphLoadingState === LoadingState.refreshing)
+    const onlyGraphLoading = (mainGraphLoadingState === LoadingState.refreshing)
```
vinibrsl, line 448: I did not understand this change. Could you please elaborate on it?

This variable confused me a bit. `onlyGraphLoading` would suggest the state is `loading` but it's `refreshing`.

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -423,39 +445,46 @@ export default class VisitorGraph extends React.Component {
...
     const showLoader =
       LoadingState.isLoadingOrRefreshing(mainGraphLoadingState) ||
       LoadingState.isLoadingOrRefreshing(topStatsLoadingState)
 
     return (
       <LazyLoader onVisible={this.onVisible}>
-        <div className={"relative w-full mt-2 bg-white rounded shadow-xl dark:bg-gray-825 main-graph"}>
-          {showLoader && <div className="graph-inner"><div className={loaderClassName}><div></div></div></div>}
+        <div className={"relative w-full mt-2 bg-white rounded shadow-xl dark:bg-gray-825"}>
+          {showLoader && !mainGraphRefreshing && renderLoader()}
```
vinibrsl, line 474: I suggest adding this new condition to the `showLoader` variable instead of comparing it inside JSX. Also, it seems to me this new condition (`!mainGraphRefreshing`) negates the `showLoader` condition. It needs testing, but I think this would suffice:

```
const showLoader = LoadingState.isLoading(mainGraphLoadingState) || LoadingState.isLoadingOrRefreshing(topStatsLoadingState)
```

vinibrsl (changes requested):

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Nice catch! Done in https://github.com/plausible/analytics/pull/2662/commits/fc9031b6fbcc2b9db72c8721fead277fc838761a. I also added the condition for a `minute` intercval which should not render clickable graph as well.

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Yeah, I agree that it is a bit confusing. Basically, the `refreshing` state describes the situation where only the graph is loading, but Top Stats are not changing. This happens when `interval` or `metric` is changed.

I thought this variable name would better describe what is actually happening. I don’t like the word “refreshing”, as the only meaning of that state is that new data for **only the graph** is being fetched. “Refreshing” isn’t telling us that.

I’ve tried getting rid of the third state and only keep “loading vs loaded”, but couldn’t find an easy solution with sensible time. I think let’s come back to it later, perhaps when moving the top stat metric querying logic to frontend.

For now, I just reverted the variable name to what it was before - `mainGraphRefreshing`. I agree that mixing the meaning of `loading` vs `refreshing` causes even more confusion.

Also, as topStatsLoadingState should never become `refreshing`, I changed the two checks for that condition.

https://github.com/plausible/analytics/pull/2662/commits/0704e5e909b06c87be4e2a719be49e8587639378

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Good point! It doesn't actually negate it since the `showLoader` is an OR condition. But I was able to make it look better here: https://github.com/plausible/analytics/pull/2662/commits/6bca0d951e3a8bf76b9ffa4a92d4fb219d333a39

vinibrsl (approved):

Issue #2258, opened 2022-09-22
Top pages list: CMD+Click causes a 404


peterhartree:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

1. CMD+click one of the pages in the list:
<img width="559" alt="image" src="https://user-images.githubusercontent.com/1412568/191687494-4c1fd53d-4333-497c-9d65-8c0966ac72e5.png">

2. I get a new tab with this URL:
<img width="507" alt="image" src="https://user-images.githubusercontent.com/1412568/191714819-28232dbb-41eb-4723-942c-98c88bf696e2.png">

### Expected behavior

I should get a new tab with this URL:

<img width="518" alt="image" src="https://user-images.githubusercontent.com/1412568/191714901-9a784e4b-2fd9-4dc3-ae8a-f64781396f95.png">

### Environment

```markdown
Chrome
```

PR #2673, opened 2023-02-14
Refactor: Use hooks instead of classes for DatePicker

ukutaht:
### Changes

Updates the `DatePicker` react component to use new hooks API instead of deprecated React Classes.

The idea is to slowly move over to hooks in general. This will allow us to upgrade react-router-dom from v5 -> v6 which should help us more easily solve https://github.com/plausible/analytics/issues/2258. It's also nice to keep dependencies up to date in general.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/datepicker.js
+++ b/assets/js/dashboard/datepicker.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from "react";
+import React, { Fragment, useState, useEffect, useCallback, useRef } from "react";
 import { withRouter } from "react-router-dom";
 import Flatpickr from "react-flatpickr";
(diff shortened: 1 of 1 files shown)


vinibrsl (approved):
Looks good!

There's possibly a bug with the key bindings, but it's not related to this pull request changes, because I can reproduce it in master. When pressing <kbd>c</kbd>(ustom) to open the date picker, I have to press it twice, otherwise it doesn't open.

[](https://user-images.githubusercontent.com/5093045/219126736-ed943bcb-fe24-4883-a346-a3c402b8fb5b.webm)

PR #2675, opened 2023-02-14
Reject unknown imported cities from queries

### LINKED ISSUE #1960: N/A instead of city names on All Time view with imported GA data
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

City names are not showing

### Expected behavior

Should see city names

### Screenshots

Screenshot 2022-06-16 at 17 27 10

### Environment

```markdown
- OS: MacOS
- Browser: Safari
- Browser Version: 15.5
```

vinibrsl:
### Changes

This commit fixes a bug where the city report returned `N/A` entries. The functions that build imported data queries were using SQL `COALESCE`, assuming city data is `NULL` when unknown, when actually its unknown value is `0`.

This commit addresses the problem using SQL `NULLIF` combined with the previous `COALESCE` call. With this change both `0` and `NULL` are treated as unknown.

Since 1cb07efe6dd800c906110858e55a6788cc2cdccf cities can be `NULL`, but previously we saved `0` as unknown.

Closes #1960

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -58,6 +58,7 @@ defmodule Plausible.Stats.Imported do
   # GA only has 'source'
   def merge_imported(q, _, _, "utm_source", _), do: q
 
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def merge_imported(q, site, query, property, metrics)
       when property in [
              "visit:source",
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/imported.ex
```diff
@@ -287,8 +290,8 @@ defmodule Plausible.Stats.Imported do
 
       :city ->
         q
-        |> select_merge([i, s], %{
-          city: fragment("coalesce(?, ?)", s.city, i.city)
+        |> select_merge([s, i], %{
```
vinibrsl, line 293: Bindings were in incorrect order :thisisfine:

ukutaht (approved):

PR #2677, opened 2023-02-15
Per-domain counters

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


test/plausible/ingestion/counters/buffer_test.exs
```diff
@@ -0,0 +1,41 @@
...
+             {bucket1, "metric2", "example.com", 1},
+             {bucket2, "metric1", "example.com", 1},
+             {bucket3, "metric2", "example.com", 1}
+           ] = aggregates
+
+    assert bucket3 - bucket2 == 60
+    assert bucket2 - bucket1 == 60
+  end
+
+  test "does not flush current bucket until its time passes", %{test: test} do
```
aerosol, line 28: This in turn makes the 10s flushing interval irrelevant, since the chunk is ready at least every 60 seconds. I'm going to change this.

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Plausible.IngestRepo.Migrations.CreateIngestCountersTable do
+  use Ecto.Migration
+
+  def change do
+    create_if_not_exists table(:ingest_counters,
+      engine: "MergeTree() ORDER BY (domain, toDate(event_timebucket))"
```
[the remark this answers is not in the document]
ukutaht: I also think this might be a good use-case for the [SummingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/summingmergetree#summingmergetree) table engine in Clickhouse. This should help reduce the number of rows we keep by collapsing rows that have the same `ORDER BY` expression

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
```diff
@@ -0,0 +1,17 @@
...
+                         ) do
+
+      #  XXX: store site identifier too?
```
ukutaht, line 9: Yes, I think we should store `site_id` as well

lib/plausible/ingestion/counters.ex
```diff
@@ -0,0 +1,91 @@
...
+      interval = Keyword.get(opts, :interval, @interval)
+
+      {:ok, {interval, buffer}}
+    else
+      :ignore
+    end
+  end
+
+  @impl true
+  def handle_cycle(buffer) do
```
ukutaht, line 55: Do you think it would be useful to measure how long it takes to flush the buffer with Promex?

lib/plausible/ingestion/counters/buffer.ex
```diff
@@ -0,0 +1,85 @@
...
+      bucket_fn: bucket_fn
+    }
+  end
+
+  @spec aggregate(t(), binary(), binary(), now :: DateTime.t()) :: t()
+  def aggregate(
+        %__MODULE__{buffer_name: buffer_name, bucket_fn: bucket_fn} = buffer,
+        metric,
+        domain,
+        now \\ DateTime.utc_now()
```
ukutaht, line 42: It's probably not an issue in the real world but I would feel slightly more comfortable if the `event_timebucket` was determined from the timestamp on the actual event, instead of a newly-created `utc_now()`.

test/plausible/ingestion/counters/buffer_test.exs
ukutaht, line 28: If I understand this correctly, the buffer will wait until the minute is over to flush a certain `event_timebucket`?

I think if we use SummingMergeTree (as suggested below) then we could also flush everything every 10 seconds with no problems. It might be good because instead of having one big insert every minute, the inserts would each be smaller and better spread out in time.

test/plausible/ingestion/counters_test.exs
```diff
@@ -0,0 +1,82 @@
...
+
+  defp emit_dropped_request() do
+    site = build(:site, domain: random_domain())
+
+    payload = %{
+      name: "pageview",
+      url: "http://#{site.domain}"
+    }
+
+    conn = build_conn(:post, "/api/events", payload)
```
ukutaht, line 41: The ingestion endpoint is `/api/event`, this is using `/api/events`. I guess this works because it bypasses the router completely?

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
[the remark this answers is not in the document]
ukutaht: It might be a good idea to ask Altinity for advice on this table schema.

ukutaht (commented):
❤️

Comments inline

test/plausible/ingestion/counters/buffer_test.exs
ukutaht: I have a suggestion in my review to flush every 10s but then use `SummingMergeTree` table engine in Clickhouse to aggregate data by 1-minute during background merges.

lib/plausible/ingestion/counters.ex
aerosol: Sure, I'll add some telemetry (size of the chunk + time taken to send), but note, I'm working on patching clickhousex so a fully async query is possible here (without waiting for an ack even in this case - I think that's reasonable?).

lib/plausible/ingestion/counters/buffer.ex
aerosol: Ah good idea

lib/plausible/ingestion/counters/buffer.ex
```diff
@@ -0,0 +1,85 @@
...
+        []
+
+      data ->
+        :ets.select_delete(buffer_name, match_specs_delete)
+        data
+    end
+  end
+
+  @spec minute_spiral(DateTime.t()) :: unix_timestamp()
+  def minute_spiral(now) do
```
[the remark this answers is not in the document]
aerosol: Ah, I'm not sure if this is established naming convention, but I remember folsom had a metric like this:

image

https://github.com/2600hz/erlang-folsom#spiral-meter

Happy to rename if this is too confusing

test/plausible/ingestion/counters/buffer_test.exs
aerosol: My only concern is the size of full minute payload to insert

aerosol: Cool, sounds like a plan, let me experiment with this

test/plausible/ingestion/counters_test.exs
aerosol: Yeah this doesn't go through any plugs really, it's a dummy conn really to build the event from

lib/plausible/ingestion/counters.ex
ukutaht: > without waiting for an ack even in this case - I think that's reasonable?

Yeah I don't think we're that worried about losing an insert here

test/plausible/ingestion/counters/buffer_test.exs
ukutaht: Yeah agreed

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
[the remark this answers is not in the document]
aerosol: > I also think this might be a good use-case for the [SummingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/summingmergetree#summingmergetree) table engine in Clickhouse. This should help reduce the number of rows we keep by collapsing rows that have the same ORDER BY expression

So given the following ordering key:

```
(domain, toDate(event_timebucket), metric, application, event_timebucket
```

the optimization here will be 1 row instead of (max) 6 rows in MergeTree (assuming 10s dump interval per application)? As in, the intention is to have Clickhouse sum every 1-minute bucket but still keep those metrics ordered by the app server?

I like the idea but I'm not sure what's the point of keeping `application` here tbh. We can debug/inspect uneven load balancing elsewhere. Is there anything else you'd like to track using the `application` column?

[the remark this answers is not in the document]
ukutaht: > the optimization here will be 1 row instead of (max) 6 rows in MergeTree (assuming 10s dump interval per application)?

Yeah exactly. Reduces the amount of data stored on disk. I supposed the way you flush the cache (only after the minute has passed) has the same effect but comes with the downside of doing one big insert per minute. Using SummingMergeTree allows to send INSERTs more often but still collapse the rows on disk once Clickhouse gets around to merging them.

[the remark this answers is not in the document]
aerosol: > has the same effect but comes with the downside of doing one big insert per minute

Right, but then the latest from Altinity is, the insert payload size is not really a concern - clickhouse doesn't care.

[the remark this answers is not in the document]
aerosol: > Yeah you're right, it's probably superfluous.

I'll remove it and include `site_id` - could be interesting to track changes one day

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
aerosol: What is the preferred column type for nullable identifier? 🤔 For some metrics there will be no site id available

test/plausible/ingestion/counters/buffer_test.exs
aerosol: Since there's no atomic "get and delete" in ets, I'll have to either keep an extra key counter (reserve before read/commit after read) or come up with something else so that no "unflushed" entries get deleted. This is the exact semantics implemented here btw: https://github.com/aerosol/airbag/blob/main/lib/airbag/buffer.ex 😅

INCOMING:

01:11:00
01:11:12
01:11:12
01:11:18

GET:

01:11:00
01:11:12
01:11:12
01:11:18

(dispatch to clickhouse)

(in the meantime more incoming)

DELETE:

01:11:00
01:11:12
01:11:12
01:11:18
01:11:18
01:11:18
01:11:19

ruslandoga: > come up with something else

Could two rotating ets table work?

- init with two ets tables, `tabA`, `tabB`
- set `state.tab = tabA`
- before flush, set `state.tab = tabB`
- flush `tabA`

aerosol: I've tried making airbag work with counters but it didn't feel like a good fit (ref https://github.com/aerosol/airbag/compare/store-counters?expand=1). I'm going to try 10s aggregates and 1m rounding dumps locally instead.

test/plausible/ingestion/counters_test.exs
aerosol: Fixed in the main PR https://github.com/plausible/analytics/pull/2693

test/plausible/ingestion/counters/buffer_test.exs
aerosol: We flush every 10s and turn the buckets into 1 minute ones on flush https://github.com/plausible/analytics/pull/2693

lib/plausible/ingestion/counters/buffer.ex
aerosol: Done in https://github.com/plausible/analytics/pull/2693

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
aerosol: Done in https://github.com/plausible/analytics/pull/2693

lib/plausible/ingestion/counters/buffer.ex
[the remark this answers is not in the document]
aerosol: Renamed in https://github.com/plausible/analytics/pull/2693

priv/ingest_repo/migrations/20230214114402_create_ingest_counters_table.exs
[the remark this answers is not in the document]
aerosol: Done in https://github.com/plausible/analytics/pull/2693

test/plausible/ingestion/counters/buffer_test.exs
aerosol: Addressed in https://github.com/plausible/analytics/pull/2693

PR #2683, opened 2023-02-15
Move CH migration from clickhouse_repo/ to ingest_repo/

vinibrsl:
This commit moves a migration that was created in `clickhouse_repo/` to its correct folder, `ingest_repo/`. The migration was created in 1cb07efe6dd800c906110858e55a6788cc2cdccf prior the read/write separation.


priv/repo/seeds.exs
```diff
@@ -62,29 +62,28 @@ geolocations = [
     city_geoname_id: 5_391_959
```
vinibrsl, line 62: I fixed an unused variable warning in this file. Please ignore :)

aerosol (approved):

PR #2686, opened 2023-02-16
Validate property key in breakdown API endpoint

ukutaht:
### Changes

Do not throw MatchClauseError when API user supplies a bad property name

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -9,7 +9,7 @@ defmodule Plausible.Stats.Breakdown do
 
   @event_metrics [:visitors, :pageviews, :events]
   @session_metrics [:visits, :bounce_rate, :visit_duration]
-  @event_props ["event:page", "event:page_match", "event:name"]
+  @event_props Plausible.Stats.Props.event_props()
 
   def breakdown(site, query, "event:goal" = property, metrics, pagination) do
     {event_goals, pageview_goals} =
--- a/lib/plausible/stats/props.ex
+++ b/lib/plausible/stats/props.ex
@@ -1,7 +1,7 @@
 defmodule Plausible.Stats.Props do
   use Plausible.ClickhouseRepo
   import Plausible.Stats.Base
-  @event_props ["event:page", "event:page_match", "event:name"]
+  @event_props ["event:page", "event:page_match", "event:name", "event:goal"]
   @session_props [
     "visit:source",
     "visit:country",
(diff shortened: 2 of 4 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -9,7 +9,7 @@ defmodule Plausible.Stats.Breakdown do
 
   @event_metrics [:visitors, :pageviews, :events]
   @session_metrics [:visits, :bounce_rate, :visit_duration]
-  @event_props ["event:page", "event:page_match", "event:name"]
+  @event_props Plausible.Stats.Props.event_props()
```
aerosol, line 12: chefs kiss 👌

lib/plausible/stats/props.ex
```diff
@@ -1,6 +1,33 @@
...
+    "visit:os_version",
+    "visit:browser",
+    "visit:browser_version"
+  ]
+
+  def event_props(), do: @event_props
+
+  def valid_prop?(prop) when prop in @event_props, do: true
+  def valid_prop?(prop) when prop in @session_props, do: true
+  def valid_prop?("event:props:" <> prop) when is_binary(prop), do: true
```
aerosol, line 29: This will match `"event:props:"` alone too. If you want to make sure that `prop` is non 0 length, it's probably best to use `byte_size` guard here.

```elixir
iex(1)> is_binary(<<>>)
true
iex(2)> byte_size(<<>>)
0
```

lib/plausible/stats/props.ex
```diff
@@ -1,6 +1,33 @@
...
+    "visit:os",
+    "visit:os_version",
+    "visit:browser",
+    "visit:browser_version"
+  ]
+
+  def event_props(), do: @event_props
+
+  def valid_prop?(prop) when prop in @event_props, do: true
+  def valid_prop?(prop) when prop in @session_props, do: true
```
aerosol, line 28:
```suggestion
  def valid_prop?(prop) when prop in @event_props ++ @session_props, do: true
```

aerosol: nitpick 🤏

aerosol (approved):

lib/plausible/stats/props.ex
ukutaht: I don't think this works

```
** (ArgumentError) invalid right argument for operator "in", it expects a compile-time proper list or compile-time range on the right side when used in guard expressions, got: @event_props ++ @session_props
```

aerosol: oops, nevermind me, it'd have to be done in another module attribute, disregard pls

lib/plausible/stats/props.ex
ukutaht: Thanks! Fixed in https://github.com/plausible/analytics/pull/2686/commits/b68cdfe79370b76b2ac831283358bacc278324be

PR #2687, opened 2023-02-16
Add visits metric under a feature flag

RobertJoonas:
### Changes

This PR adds a new `Visits` metric to the Top Stats row on the dashboard, as well as to the `visitors.csv` file in the CSV export zip. 

This is meant to be deployed under a feature flag so we can test it on production first. The feature can be enabled with a `visits_metric` flag.

### Tests

### Changelog & Documentation
Should be updated when the feature goes live for everyone (feature flag removed)

### Dark mode


assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -475,6 +475,7 @@ export default class VisitorGraph extends React.Component {
   fetchTopStatData() {
     api.get(`/api/stats/${encodeURIComponent(this.props.site.domain)}/top-stats`, this.props.query)
       .then((res) => {
+        res.top_stats = this.maybeRemoveVisitsMetric(res.top_stats)
```
ukutaht, line 478: Is there a good reason to do this on the frontend? I feel like not returning the `visits` metric from the backend would be easier, because the backend has access to the flags with no extra work.

ukutaht (commented):
Looks good overall. I think we should also address the `visits` metric as discussed in Basecamp

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: I did it on the frontend, because it seemed like it required writing less code. The `get "/:domain/top-stats"` route is currently not piped through `PlausibleWeb.AuthPlug` which makes the current user available in `conn.assigns`. And we'd need to look up the user to get the feature flag status, whereas this logic is all already there on the frontend.

ukutaht: Ah yes, makes sense!

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -95,7 +95,17 @@ defmodule PlausibleWeb.StatsController do
     site = conn.assigns[:site]
     query = Query.from(site, params) |> Filters.add_prefix()
 
-    metrics = [:visitors, :pageviews, :bounce_rate, :visit_duration]
+    visits_metric_enabled =
+      FunWithFlags.enabled?(:visits_metric, for: conn.assigns[:current_user]) ||
+        Mix.env() == :test
+
+    metrics =
+      if visits_metric_enabled && !query.filters["event:goal"] do
```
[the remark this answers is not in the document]
RobertJoonas: No. I thought we could already assume that the feature is enabled in tests. Do you think we should remove the `Mix.env() == :test` condition and add a test for the case where `visits` metric is disabled?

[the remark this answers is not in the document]
ukutaht: I meant more for the `query.filters["event:goal"]` clause here. Previously I think the metric list was the same regardless of that filter. Now there are two different sets of metrics depending on the `event:goal` filter which I thought could use a test.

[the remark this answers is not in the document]
RobertJoonas: Ah, I see. Yes, there is a test for a goal-filtered dashboard export already: https://github.com/plausible/analytics/blob/master/test/plausible_web/controllers/stats_controller_test.exs#L243-L248

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -353,9 +353,9 @@ defmodule PlausibleWeb.Api.StatsController do
 
     metrics =
       if query.filters["event:page"] do
-        [:visitors, :pageviews, :bounce_rate, :time_on_page, :sample_percent]
+        [:visitors, :pageviews, :bounce_rate, :time_on_page, :visits, :sample_percent]
       else
-        [:visitors, :pageviews, :bounce_rate, :visit_duration, :sample_percent]
+        [:visitors, :pageviews, :bounce_rate, :visit_duration, :visits, :sample_percent]
```
RobertJoonas, line 358: This would have made more sense to include in the very first commit of this PR (https://github.com/plausible/analytics/pull/2687/commits/14de861431591ac8e6d517f397382e179e974353) along with adding the `top_stats_entry`.

It was just a coincidence that the `visits` metric was returned before, as we were fetching it to be able to merge some imported data. The metric that we use internally for merging imported data is now renamed to `__internal_visits` (https://github.com/plausible/analytics/pull/2691) to avoid any further confusion.

ukutaht (approved):
LGTM 👍

PR #2692, opened 2023-02-20
Clickhouse migration: add ingest_counters table

aerosol:
### Changes

This PR adds ingest counter migration. We're going to use SummingMergeTree as discussed in https://github.com/plausible/analytics/pull/2677. Eventually the database will perform roll-ups on 1-minute event_buckets. `site_id` is made `Nullable` as it's not mandatory for all metrics (e.g. dropped events due to `not_found` won't have it).

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #2693, opened 2023-02-20
Implement ingest counters

aerosol:
### Changes

This PR is built on top of https://github.com/plausible/analytics/pull/2692 and will be rebased once the former is merged. **Do not merge yet.**

The commits in this PR are meant to be read in chronological order.

This change implements flushing internal metrics to Clickhouse. To have better understanding of the traffic we're receiving, we'll aggregate ingestion events into queryable 1-minute buckets of counters. Due to high cardinality those metrics are not suitable for grafana (cardinality is low from clickhouse perspective).

Internally those counters are aggregated within 10s buckets and flushed every 10s as well. 1 minute roll-ups are outsourced to Clickhouse (`SummingMergeTree`).

This PR also updates clickhousex to a branch (PR: https://github.com/plausible/clickhousex/pull/22) where a session-id parameter is assigned to each newly established connection, so that per-connection query-level settings can be invoked. Here, an [asynchronous insert](https://clickhouse.com/docs/en/optimize/asynchronous-inserts/) is attempted to minimize interference with main ingestion channel where "inserts per second" limitation may apply depending on horizontal scale. 

We'll take the opportunity to exercise async inserts here (the data collected is not critical and we can afford the experiment) to further assess if it's worth making the main ingestion path to follow.

@ukutaht I left the dirty work intact (https://github.com/plausible/analytics/pull/2677) so that there's a trace of our discussion without force pushes.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/ingestion/counters.ex
```diff
@@ -0,0 +1,112 @@
...
+        try do
+          IngestRepo.checkout(fn ->
+            IngestRepo.query!("SET async_insert = 1")
+            IngestRepo.query!("SET wait_for_async_insert = 0")
+            {_, _} = IngestRepo.insert_all(Record, records)
+            IngestRepo.query!("SET async_insert = 0")
+            IngestRepo.query!("SET wait_for_async_insert = 0")
+          end)
+        catch
+          _, thrown ->
```
vinibrsl, line 91: Do we need to restore `SET async_insert = 0; SET wait_for_async_insert = 0` in case of an exception, or `Ecto.Repo.checkout/2` already takes care of that?

lib/plausible/ingestion/counters/buffer.ex
```diff
@@ -0,0 +1,106 @@
+defmodule Plausible.Ingestion.Counters.Buffer do
+  @moduledoc """
+  A buffer aggregating counters for internal metrics, within 10 seconds time buckets.
+
+  See `Plausible.Ingestion.Counters` for integration.
+
+  Flushing is by default possible only once the 10s bucket is complete
```
ukutaht, line 7: I'm curious how `gen_cycle` handles app shutdowns. Ideally during a normal (graceful) shutdown it would flush the last buffer before exiting.

lib/plausible/ingestion/counters.ex
aerosol: I think I assumed that we always reconnect on every type of exception but it's not necessarily the case hm. Thanks for pointing that out, let me dig into this.

lib/plausible/ingestion/counters.ex
```diff
@@ -0,0 +1,112 @@
...
+
+    {:continue_hibernated, buffer}
+  end
+
+  @impl true
+  def handle_info(_msg, state) do
+    {:continue, state}
+  end
+
+  defp to_0_minute_datetime(unix_ts) when is_integer(unix_ts) do
```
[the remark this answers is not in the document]
aerosol: > Looks to me like the 1-minute rollups are currently achieved by truncating the timestamp here instead of Clickhouse doing any work with SummingMergeTree. 

We dump every ordering key at most 5 times per minute, every 10 seconds. So `SummingMergeTree` will do work at most 5 times. 

So it looks like this:

```mermaid
sequenceDiagram
	actor User
	participant GateKeeper
	participant Counters.Buffer
	participant Counters
	participant Clickhouse

	User->>GateKeeper: pageview foo.com
	User->>GateKeeper: pageview foo.com

	GateKeeper->>Counters.Buffer: buffered foo.com
	Note over GateKeeper: event time 10:11:12
	Note over Counters.Buffer: bucket 10:11:10

	Counters.Buffer->>Counters.Buffer: aggregate 
	Note over GateKeeper: event time 10:11:18
	Note over Counters.Buffer: bucket 10:11:10

	Counters.Buffer--)Counters.Buffer: aggregate 10:11:10 (2)

	User->>GateKeeper: pageview foo.com

	GateKeeper->>Counters.Buffer: buffered foo.com
	Note over GateKeeper: event time 10:11:29
	Note over Counters.Buffer: bucket 10:11:20

	Counters.Buffer--)Counters.Buffer: aggregate 10:11:20 (1)

	Counters->>Counters.Buffer: flush
	Counters.Buffer-->>Counters: 10:11:10 (2)

	Counters->>Clickhouse: insert 10:11:00 (2)

	Note over Counters: next cycle (10s)

	Counters->>Counters.Buffer: flush
	Counters.Buffer-->>Counters: 10:12:20 (1)

	Counters->>Clickhouse: insert 10:11:00 (1)

	Clickhouse--)Clickhouse: SummingMergeTree sum (x2)

	Note over Clickhouse: event_timebucket 10:11:00 value: 3


```

Does that make sense?

lib/plausible/ingestion/counters/buffer.ex
aerosol: Good call, I'll look into that.

lib/plausible/ingestion/counters.ex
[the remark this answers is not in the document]
ukutaht: Ah yes, sorry for the confusion. I was thrown off by the fact that the SummingMergeTree migration does not specify the `columns` parameter:

Screenshot 2023-02-22 at 10 50 00

But I think the default clause applies and is working there.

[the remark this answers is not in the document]
aerosol: Oh nice, I missed that setting. Wondering now if maybe it's good to be explicit about it? 🤔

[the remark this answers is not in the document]
ukutaht: [Comment: ](https://github.com/plausible/analytics/pull/2692/files#r1114020321)

lib/plausible/ingestion/counters.ex
aerosol: Discussed that with @ruslandoga and ended up doing https://github.com/plausible/analytics/pull/2693/commits/af4754a435f1aa79a6a7ee3d626f4a7017061e05

Basically transport errors will trigger a reconnect, but other errors such as clickhouse exception on "select * from doesnotexist" won't - this behavior is the same in postgrex for example.

`after` clause is best effort here. In case the connection is borked at the socket level we won't have to reset the state, since a reconnection is expected anyway. On random unexpected responses we'll assume we still got it checked out and usable.

lib/plausible/ingestion/counters.ex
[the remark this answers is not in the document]
ukutaht: was trying to link to this comment :) https://github.com/plausible/analytics/pull/2692#discussion_r1114020321

ukutaht (approved):

lib/plausible/ingestion/counters/buffer.ex
aerosol: @ukutaht:

  - https://github.com/aerosol/gen_cycle/pull/5
  - along with https://github.com/plausible/analytics/pull/2693/commits/32a485caed782c117f0fa8d37dfc8e25bb3792bb

aerosol: WYT?

vinibrsl (approved):

PR #2695, opened 2023-02-21
Upgrade erlang/elixir/alpine stack

aerosol:
### Changes

This PR performs the following platform upgrades:
  - Elixir: [1.14.3](https://github.com/elixir-lang/elixir/releases/tag/v1.14.3)
  - Erlang: [25.2.3](https://www.erlang.org/patches/otp-25.2.3)
  - Alpine: [3.17.0](https://www.alpinelinux.org/posts/Alpine-3.17.0-released.html)

Going to run it on staging first.
cc @ruslandoga 

For local environment `asdf reshim` can be handy.

### Tests

### Changelog

### Documentation

### Dark mode


.tool-versions
```diff
@@ -1,4 +1,5 @@
-elixir 1.13.4-otp-24
-erlang 24.3.3
+erlang 25.2.3
+elixir 1.14.3-otp-25
+erlang 25.2.3
```
ruslandoga, line 3: erlang seems duplicated here

Dockerfile
```diff
@@ -1,8 +1,8 @@
 # we can not use the pre-built tar because the distribution is
 # platform specific, it makes sense to build it in the docker
-
+nt
```
ruslandoga, line 3: nt?

ruslandoga:
I wonder if it makes sense to remove erlang from `apk get` on line 15?

aerosol:
> I wonder if it makes sense to remove erlang from apk add on line 15?

yes, I'll give it a go

ruslandoga:
Should we also update CI versions?

ruslandoga (approved):

ruslandoga:
Re CI failing, it wants a newline in an empty file. I think we can just delete it (since it's empty).

aerosol:
@ruslandoga I'm getting tons of errors locally on mix test. Could you check on your end while I'm looking closer?

aerosol:
I guess it's a consequence of 

```
 ** (exit) exited in: GenServer.call(Plausible.Session.WriteBuffer, :flush, :infinity)
         ** (EXIT) an exception was raised:
             ** (ArgumentError) :naive_datetime expects microseconds to be empty, got: ~N[2021-10-20 11:59:00.000000]

     Use `NaiveDateTime.truncate(naive_datetime, :second)` (available in Elixir v1.6+) to remove microseconds.
```

aerosol:
@ruslandoga it looks like we need to truncate it in gazillion of places don't we? 💀

ruslandoga:
yeah...

aerosol:
alright on it

aerosol:
On staging.

aerosol:
All looks good to me cc @cnkk

aerosol:
@ruslandoga would you mind doing some sanity check on your instance?

Dockerfile
```diff
@@ -48,7 +48,7 @@ COPY rel rel
 RUN mix release plausible
 
 # Main Docker Image
-FROM alpine:3.17.0
+FROM alpine:3.17.2@sha256:e2e16842c9b54d985bf1ef9242a313f36b856181f188de21313820e177002501
```
aerosol, line 51: this might cause issues as the build is done on https://github.com/plausible/analytics/pull/2695/commits/97ca3ad882badc80e5c5149bd7a29f4a88dab5c5#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R5 - I remember I had openssl problems with non matching releases

cnkk: Ack.
Hm, they already upgraded to the newer Alpine image(https://github.com/hexpm/bob/commit/df58cc4f483d93a7179e6ae7cd93a7622593182a), but it's not on Docker Hub yet

cnkk:
We should at least pin the main image to an exact digest (https://github.com/plausible/analytics/pull/2695/commits/b2465aae3743e3ff7b46cc6e589b8579cb7d26fd).

ruslandoga:
>  would you mind doing some sanity check on your instance?

Sure, will do. Is there a place I can fetch this image from? If not, I'll rebuild it tomorrow and deploy to my instance.

aerosol:
@ruslandoga would https://hub.docker.com/layers/plausible/analytics-private/pr-2695/images/sha256-7bea94579db39a0bcde36bb4c9136956df12f364a995e5e8d8cf9b0697aa7d73?context=explore work?

ruslandoga:
Yes! Deployed.

ukutaht (approved):

PR #2698, opened 2023-02-22
Fix graph element z-index

vinibrsl:
Previously, the graph element was overlapping with other UI elements such as the period input. This commit fixes this issue by setting the z-index of the graph element to 0, which moves it behind other elements in the stacking order.


RobertJoonas (approved):
🙏🏼

PR #2704, opened 2023-02-23
Implement "Year over Year" comparison mode

vinibrsl:
### Changes

This pull request adds support for multiple comparison modes, changes the comparison checkbox to a combobox, and implements the year over year / last year comparison mode. The feature is still behind a feature flag.

Screenshot 2023-02-23 at 12 56 07

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -27,7 +27,7 @@ const ComparisonInput = function({ site, query, history }) {
       <Menu.Item
         key={value}
         onClick={() => update(value)}
-        className="px-4 py-2 leading-tight hover:bg-gray-100 hover:text-gray-900 dark:hover:bg-gray-900 dark:hover:text-gray-100 flex hover:cursor-pointer">
+        className="px-4 py-2 leading-tight hover:bg-gray-100 dark:text-white hover:text-gray-900 dark:hover:bg-gray-900 dark:hover:text-gray-100 flex hover:cursor-pointer">
         <span className={labelClass}>{ label }</span>
       </Menu.Item>
     )
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -1,6 +1,14 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import { withRouter } from "react-router-dom";
 import { navigateToQuery } from './query'
+import { Menu, Transition } from '@headlessui/react'
+import { ChevronDownIcon } from '@heroicons/react/20/solid'
+import classNames from 'classnames'
+
+const COMPARISON_MODES = {
+  'previous_period': 'Previous period',
+  'last_year': 'Last year',
+}
 
 export const COMPARISON_DISABLED_PERIODS = ['realtime', 'all']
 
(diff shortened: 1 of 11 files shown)


assets/js/dashboard/comparison-input.js
```diff
@@ -1,21 +1,61 @@
...
+import React, { Fragment } from 'react'
 import { withRouter } from "react-router-dom";
 import { navigateToQuery } from './query'
+import { Menu, Transition } from '@headlessui/react'
+import { ChevronDownIcon } from '@heroicons/react/20/solid'
+import classNames from 'classnames'
+
+const COMPARISON_MODES = {
+  'previous_period': 'Previous period',
+  'last_year': 'Last year',
```
vinibrsl, line 10: I'd like some feedback on what we should call this comparison mode. I named it "Last year", but it's pretty common to call this "Year over year" too. Essentially it compares the current stats with last year stats, e.g. Jan 2021 compared to Jan 2020.

assets/js/dashboard/query.js
```diff
@@ -19,7 +19,7 @@ export function parseQuery(querystring, site) {
     period = '30d'
   }
 
-  let comparison = !!q.get('comparison')
+  let comparison = q.get('comparison')
```
vinibrsl, line 22: Previously the comparison query property was a boolean, on or off. Now it supports multiple modes and it's a string, e.g. `last_year`, `previous_period`.

assets/js/dashboard/comparison-input.js
metmarkosaric: nice! "Previous period" and "Year over year" are good i think

ukutaht: I think I prefer 'year over year' vs 'last year'

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -170,6 +166,25 @@ defmodule PlausibleWeb.Api.StatsController do
     nil
   end
 
+  defp query_comparison(comparison_mode, site, query, selected_metric) do
+    case comparison_mode do
+      "previous_period" ->
+        comparison_query = Query.shift_back(query, site)
+        Stats.timeseries(site, comparison_query, [selected_metric])
+
+      "last_year" ->
```
ukutaht, line 175: Hmm, I think this implementation is missing the incomplete period logic that we have in `Query.shift_back`.

The idea is that comparisons should compare the same number of months/weeks/days/hours. For example, if you're in the middle of the month and looking at 'this month', then we display the full month even though there's only data for the first 14 days. In this case we want the comparison to also only include 14 days from the comparison period.

We've started to add clauses to `shift_back` to account for this discrepancy. I think some are still missing, and that's what [this PR](https://github.com/plausible/analytics/pull/2628) is about.

I think similar special cases need to be added for the year over year comparison mode. When 'this month' only includes 14 days, we need to shift back one year but also make sure we only select 14 days to compare with.

Does that make sense? Happy to hop on a quick call to dive deeper

vinibrsl: Good catch! If I understand that correctly, we should not show comparisons for any future date in the graph, is that it? Even if comparison has data to show.

Screenshot 2023-03-01 at 12 22 40

We could reimplement what we have for shift_back to match the comparison logic. Or we could truncate the comparison plot to match the same number of items of the main plot. What do you think?

I'm up for pairing on this. We can talk more about it tomorrow in our call.

lib/plausible/stats/comparisons.ex
```diff
@@ -12,29 +12,44 @@ defmodule Plausible.Stats.Comparisons do
...
+  defp do_compare(site, source_query, "previous_period", _now) do
     Stats.Query.shift_back(source_query, site)
   end
 
-  defp do_compare(_site, source_query, "year_over_year") do
+  defp do_compare(_site, source_query, "year_over_year", now) do
     start_date = Date.add(source_query.date_range.first, -365)
-    end_date = Date.add(source_query.date_range.last, -365)
-    range = Date.range(start_date, end_date)
+    end_date = earliest(source_query.date_range.last, now) |> Date.add(-365)
```
ukutaht, line 42: elegant 👌

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -296,29 +298,41 @@ defmodule PlausibleWeb.Api.StatsController do
...
+  defp fetch_top_stats(site, %Query{filters: %{"event:goal" => _goal}} = query, comparison_mode) do
     total_q = Query.remove_event_filters(query, [:goal, :props])
-    prev_query = Query.shift_back(query, site)
-    prev_total_query = Query.shift_back(total_q, site)
+
+    {prev_converted_visitors, prev_completions} =
+      case Stats.Comparisons.compare(site, query, comparison_mode) do
+        {:ok, prev_query} ->
+          %{visitors: %{value: prev_converted_visitors}, events: %{value: prev_completions}} =
+            Stats.aggregate(site, prev_query, [:visitors, :events])
```
ukutaht, line 308: Not a concern for this PR but I think eventually we want to also support the different comparison modes in the API. In that case it would make sense to push the comparison logic down to `Stats.Aggregate` and `Stats.Timeseries` so the logic can be shared between the dashboard and the API.

ukutaht (approved):

PR #1659, opened 2022-02-09
Change screen_size to use browser user agent

eriknakata:
### Changes

Change how the screen size is calculated. The library UA inspector is already used by the project. So is possible use it to calculate the screen size. If not founded by the UA inspector the mechanism of calculate from screen_width is used as a kind of fallback. The existent function is still used to calculate Desktop and Laptop devices.

Original discussion: https://github.com/plausible/analytics/discussions/157

### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #2711, opened 2023-02-28
Use user-agent instead of screen_width to get device type

ukutaht:
### Changes

Follow-up for [#1659](https://github.com/plausible/analytics/pull/1659):

* More exhaustive matching for device types from matomo/device-detector
* Reverted a change to the UI labels for now. Not sure if that's desirable.

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl (approved):

lib/plausible/ingestion/event.ex
```diff
@@ -324,6 +311,54 @@ defmodule Plausible.Ingestion.Event do
     end
   end
 
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
+  defp screen_size(ua) do
+    case ua.device do
+      %UAInspector.Result.Device{type: "smartphone"} ->
```
aerosol, line 317: minor readability nitpick -- we could match like this

```elixir
%Device{type: t} when t in @mobile_types -> "Mobile"
%Device{type: t} when t in @tablet_types -> "Tablet"
%Device{type: t} when t in @desktop_types -> "Desktop"
%Device{type: t} -> nil
...
```

aerosol (approved):

lib/plausible/ingestion/event.ex
ukutaht: great suggestion, thanks!

https://github.com/plausible/analytics/pull/2711/commits/65aec6591a7771e76ff93a56a8a94afd6221b840

PR #2714, opened 2023-02-28
Add new `Pages / Visit` metric to Stats API Aggregate and Timeseries

RobertJoonas:
### Changes

This PR adds a new `pages_per_visit` metric to two endpoints in the Stats API:

* `api/v1/stats/aggregate`
* `api/v1/stats/timeseries`

Supporting the metric in the breakdown endpoint is out of scope.

> Reviewing commit by commit suggested. Comments in-line.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/aggregate.ex
+++ b/lib/plausible/stats/aggregate.ex
@@ -1,7 +1,7 @@
 defmodule Plausible.Stats.Aggregate do
   alias Plausible.Stats.Query
   use Plausible.ClickhouseRepo
-  import Plausible.Stats.{Base, Imported, Util}
+  import Plausible.Stats.{Base, Imported}
 
   @event_metrics [:visitors, :pageviews, :events, :sample_percent]
   @session_metrics [:visits, :bounce_rate, :visit_duration, :pages_per_visit, :sample_percent]
@@ -45,7 +45,6 @@ defmodule Plausible.Stats.Aggregate do
     |> select_session_metrics(metrics)
     |> merge_imported(site, query, :aggregate, metrics)
     |> ClickhouseRepo.one()
-    |> stringify_pages_per_visit()
   end
 
   defp aggregate_time_on_page(site, query) do
--- a/lib/plausible/stats/aggregate.ex
+++ b/lib/plausible/stats/aggregate.ex
@@ -109,9 +109,11 @@ defmodule Plausible.Stats.Aggregate do
   end
 
   defp round_and_wrap_value({metric, nil}), do: {metric, %{value: 0}}
+
   defp round_and_wrap_value({:pages_per_visit = metric, value}) do
     {metric, %{value: round(value * 100) / 100}}
   end
+
   defp round_and_wrap_value({metric, value}) do
     {metric, %{value: round(value)}}
   end
--- a/lib/plausible/stats/aggregate.ex
+++ b/lib/plausible/stats/aggregate.ex
@@ -111,7 +111,7 @@ defmodule Plausible.Stats.Aggregate do
   defp round_and_wrap_value({metric, nil}), do: {metric, %{value: 0}}
 
   defp round_and_wrap_value({:pages_per_visit = metric, value}) do
-    {metric, %{value: round(value * 100) / 100}}
+    {metric, %{value: Float.to_string(round(value * 100) / 100)}}
   end
 
   defp round_and_wrap_value({metric, value}) do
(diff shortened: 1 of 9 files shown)


lib/plausible/stats/aggregate.ex
```diff
@@ -109,4 +107,12 @@ defmodule Plausible.Stats.Aggregate do
     [[time_on_page]] = res.rows
     %{time_on_page: time_on_page}
   end
+
+  defp round_and_wrap_value({metric, nil}), do: {metric, %{value: 0}}
+  defp round_and_wrap_value({:pages_per_visit = metric, value}) do
+    {metric, %{value: round(value * 100) / 100}}
```
RobertJoonas, line 114: I'm now thinking, that maybe it's a good idea to convert this value into a string here already? Since we're not doing any further calculations with this metric, I think it would make sense to return a string straight from the Aggregate module. The same could be done for Timeseries.

Doing this, we could get rid of the [`stringify_float_values` function](https://github.com/plausible/analytics/blob/1ce25b3e9dd3f4bfeea95355f37397d09790861b/lib/plausible_web/controllers/api/external_stats_controller.ex#L262) which I was thinking to use right before returning the JSON response.

lib/plausible/stats/aggregate.ex
```diff
@@ -109,4 +107,14 @@ defmodule Plausible.Stats.Aggregate do
     [[time_on_page]] = res.rows
     %{time_on_page: time_on_page}
   end
+
+  defp round_and_wrap_value({metric, nil}), do: {metric, %{value: 0}}
+
+  defp round_and_wrap_value({:pages_per_visit = metric, value}) do
+    {metric, %{value: Float.to_string(round(value * 100) / 100)}}
```
ukutaht, line 114: Can this rounding be done in the database query instead? In other places we use `round()` clickhouse function instead of rounding in Elixir.

ukutaht: Personally I don't see much value in wrapping the `Float.to_string` either. It makes encoding and decoding more inconvenient but I'm not sure what we get from it.

lib/plausible/stats/base.ex
```diff
@@ -359,6 +359,16 @@ defmodule Plausible.Stats.Base do
     |> select_session_metrics(rest)
   end
 
+  def select_session_metrics(q, [:pages_per_visit | rest]) do
+    from(s in q,
+      select_merge: %{
+        pages_per_visit:
+          fragment("ifNotFinite(sum(? * ?) / sum(?), 0)", s.sign, s.pageviews, s.sign)
```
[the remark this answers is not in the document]
ukutaht: > My common sense tells me that sum(sign) should be more efficient, and still give the right number.

Yes I think that's right.

> Not sure if there's anything wrong with using sum(s.sign)? Perhaps change it here ^ too?

I don't think there's anything wrong, and I agree that it should be changed for the `visits` metric as well

lib/plausible/stats/aggregate.ex
RobertJoonas: > Can this rounding be done in the database query instead? In other places we use round() clickhouse function instead of rounding in Elixir.

Good point! Agreed.

> Personally I don't see much value in wrapping the Float.to_string either. It makes encoding and decoding more inconvenient but I'm not sure what we get from it.

@ukutaht are you saying that the API should return floats? As far I understand, a good practice is to return decimal numbers as strings in JSON, to provide a consistent numeric precision for different JSON implementations. https://stackoverflow.com/questions/35709595/why-would-you-use-a-string-in-json-to-represent-a-decimal-number

Once it's converted into a string, our backend does not have to decode it into float anymore - the API would just return it as a string. Furthermore, I don't think our frontend requires the value to be a float either. So I would leave that `Float.to_string` as is.

ukutaht: > Good point! Done.

I'm not seeing the change. Did you push it?

> @ukutaht are you saying that the API should return floats?

Yeah, that's my suggestion. I agree that floats shouldn't be used for money and other fields where precision is important. For a statistical number like `pages_per_visit` it doesn't really matter if a small error occurs in transit.

RobertJoonas: I just pushed some new changes. Sorry, I'm only now seeing your comment, after changing it in https://github.com/plausible/analytics/pull/2714/commits/34fee686b6bd6a96187c80e76bd3e7db9b4f499a. But your point makes sense, I will remove the stringifying.

RobertJoonas: Done here: https://github.com/plausible/analytics/pull/2714/commits/43721e7ae07602149f2c4f26199b6ef6c04a2c23.

ukutaht (approved):

PR #2728, opened 2023-03-06
Add the new `Views per Visit` metric to dashboard Top Stats and CSV export

RobertJoonas:
### Changes

* Add `Views per Visit` to dashboard Top Stats
  * Similarly to the `visits` metric, only shows in the Top Stats if:
    * the selected time period **is not** `realtime`
    * no goal filter is applied
* Add `Views per Visit` to dashboard CSV export
  * won't be included in the export if a goal filter is applied
* Re-order the Top Stats row
  * `Unique visitors` | `Total visits` | `Total pageviews` | `Views per visit` | `Bounce rate` | `Visit duration`
* Rename `Visits` to `Total visits` in the UI

> Reviewing commit-by-commit suggested

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -98,11 +98,25 @@ defmodule PlausibleWeb.StatsController do
     visits_metric_enabled =
       FunWithFlags.enabled?(:visits_metric, for: conn.assigns[:current_user])
 
+    views_per_visit_enabled =
```
ukutaht, line 101: Do we need a second feature flag here? Could reuse the `visits_metric` flag for this as well.

ukutaht (approved):
Looks good to me. One comment about the new flag inline.

lib/plausible_web/controllers/stats_controller.ex
RobertJoonas: You're right. There's no point of having two feature flags, when both metrics are supposed to go live together.

I'll make the change :)

PR #2731, opened 2023-03-08
Fix a bug in percent change calculation

RobertJoonas:
### Changes

This PR fixes a bug where the Top Stats crashed, trying to calculate percent difference for goal filtered data.   

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -427,6 +427,8 @@ defmodule PlausibleWeb.Api.StatsController do
     percent_change(old_count, new_count)
   end
 
+  defp percent_change(nil, _new_count), do: nil
+
   defp percent_change(old_count, new_count) do
     cond do
       old_count == 0 and new_count > 0 ->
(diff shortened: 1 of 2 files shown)


vinibrsl (approved):

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1007,7 +1009,7 @@ defmodule PlausibleWeb.Api.StatsController do
     end
   end
 
-  defp calculate_cr(nil, _converted_visitors), do: 100.0
```
aerosol, line 1010: Was this wrong? 🤔

aerosol: I'm not sure, just double checking. From no value to a number the change is 100 I guess?

RobertJoonas: >  From no value to a number the change is 100 I guess?

Actually, the current behaviour is that when there's no previous value to compare with (e.g. when `period=all`) then we do not display the comparison percents at all. In that case they are returned as `nil`.

PR #2732, opened 2023-03-08
another clickhouse adapter

ruslandoga:
### Changes

This PR makes a switch from [clickhousex](https://github.com/plausible/clickhousex) + [ecto_clickhouse](https://github.com/plausible/clickhouse_ecto) to [ch](https://github.com/ruslandoga/ch) + [chto.](https://github.com/ruslandoga/chto)

### Tests

### Changelog
- [x] This PR does not make a user-facing change (at least, that's the plan)

### Documentation

### Dark mode


config/runtime.exs
```diff
@@ -296,7 +296,7 @@ config :plausible, Plausible.AsyncInsertRepo,
   queue_interval: 2000,
   url: ch_db_url,
   pool_size: 1,
-  clickhouse_settings: [
+  settings: [
```
ruslandoga, line 299: This is just my preference, we can make it `opts[:clickhouse_settings]` in [here.](https://github.com/ruslandoga/ch/blob/64c8b349d3b7642256ee72f1556b82d75e57e603/lib/ch/connection.ex#L23)

lib/plausible/event/clickhouse_schema.ex
```diff
@@ -8,32 +8,32 @@ defmodule Plausible.ClickhouseEvent do
...
+    field :session_id, Ch.Types.UInt64
     field :timestamp, :naive_datetime
 
-    field :referrer, :string, default: ""
-    field :referrer_source, :string, default: ""
-    field :utm_medium, :string, default: ""
-    field :utm_source, :string, default: ""
-    field :utm_campaign, :string, default: ""
-    field :utm_content, :string, default: ""
-    field :utm_term, :string, default: ""
```
ruslandoga, line 21: I'm removing all defaults because in [RowBinary](https://clickhouse.com/docs/en/interfaces/formats/#rowbinary) we have to encode all columns to something, and the defaults [there](https://github.com/ruslandoga/ch/blob/64c8b349d3b7642256ee72f1556b82d75e57e603/lib/ch/row_binary.ex#L29-L164) are the same as here:

- zero length for `String` and `Array(T)`: `<<0>>` (strings and arrays are prefixed by their length: <<length::varint, data::bytes>>, so `<<0>>` means empty)
- zeroed strings for FixedString: `<<0::size(size * 8)>>` where size is the same as `FixedString(size)`
- `<<0::size-little>>` for all kinds of numbers, dates, and datetimes, booleans
- `<<1>>` for `Nullable(T)` to indicate a `null`

lib/plausible/session/cache_store.ex
```diff
@@ -53,41 +53,16 @@ defmodule Plausible.Session.CacheStore do
...
-        operating_system:
-          if(session.operating_system == "",
-            do: event.operating_system,
-            else: session.operating_system
-          ),
+        country_code: session.country_code || event.country_code,
+        subdivision1_code: session.subdivision1_code || event.subdivision1_code,
+        subdivision2_code: session.subdivision2_code || event.subdivision2_code,
+        city_geoname_id: session.city_geoname_id || event.city_geoname_id,
+        operating_system: session.operating_system || event.operating_system,
```
ruslandoga, line 60: And relying on `nil` by default in the schemas makes these fallbacks easier.

lib/plausible/stats/props.ex
```diff
@@ -51,7 +51,7 @@ defmodule Plausible.Stats.Props do
       nil ->
         ClickhouseRepo.all(
           from e in base_event_query(site, query),
-            inner_lateral_join: meta in fragment("meta as m"),
```
ruslandoga, line 54: `clickhouse_ecto` didn't alias fragments so `meta as m` worked fine, but `chto` aliases fragments and these part ended up as `meta as m as f0` which confused clickhouse.

config/runtime.exs
aerosol: this is fine

priv/ingest_repo/migrations/20200915070607_create_events_and_sessions.exs
```diff
@@ -30,15 +32,17 @@ defmodule Plausible.ClickhouseRepo.Migrations.CreateEventsAndSessions do
...
+                           options:
+                             "PARTITION BY toYYYYMM(start) ORDER BY (domain, toDate(start), user_id, session_id) SETTINGS index_granularity = 8192"
                          ) do
       add(:session_id, :UInt64)
       add(:sign, :Int8)
       add(:domain, :string)
       add(:user_id, :UInt64)
       add(:hostname, :string)
-      add(:is_bounce, :boolean)
+      add(:is_bounce, :UInt8)
```
ruslandoga, line 45: [`chto` uses `Bool` type for `:boolean`.](https://gist.github.com/ruslandoga/528a023fb436aed727ad784684a536af) This change is to keep the migrated database the same as before.

aerosol:
Deploying to staging

priv/ingest_repo/migrations/20201020083739_add_event_metadata.exs
```diff
@@ -3,7 +3,7 @@ defmodule Plausible.ClickhouseRepo.Migrations.AddEventMetadata do
 
   def change do
     alter table(:events) do
-      add :meta, {:nested, {{:key, :string}, {:value, :string}}}
+      add :meta, :"Nested(key String, value String)"
```
ruslandoga, line 6: `chto` drops support for all custom types like this `{:nested ...}` and instead pushes the user to hardcode the type they want.

priv/ingest_repo/migrations/20200915070607_create_events_and_sessions.exs
```diff
@@ -8,8 +8,10 @@ defmodule Plausible.ClickhouseRepo.Migrations.CreateEventsAndSessions do
 
   defp create_events() do
     create_if_not_exists table(:events,
-                           engine:
-                             "MergeTree() PARTITION BY toYYYYMM(timestamp) ORDER BY (domain, toDate(timestamp), user_id) SETTINGS index_granularity = 8192"
+                           primary_key: false,
```
ruslandoga, line 11: `clickhouse_ecto` [is dropping](https://github.com/plausible/clickhouse_ecto/blob/43126c020f07b097c55a81f68a906019fd061f29/lib/clickhouse_ecto/migration.ex#L31-L32) `:id` silently. `chto` crashes on the default id since it has a type that chto doesn't support (`id` or `bigserial` which don't work since ClickHouse doesn't have AUTOINCREMENT support).

test/support/test_utils.ex
```diff
@@ -73,28 +73,38 @@ defmodule Plausible.TestUtils do
...
       Enum.map(pageviews, fn pageview ->
-        Factory.build(:pageview, pageview) |> Map.from_struct() |> Map.delete(:__meta__)
+        Factory.build(:pageview, pageview)
+        |> Map.from_struct()
+        |> Map.delete(:__meta__)
+        |> update_in([:timestamp], &to_naive_truncate/1)
       end)
 
-    Plausible.IngestRepo.insert_all("events", pageviews)
+    Plausible.IngestRepo.insert_all(Plausible.ClickhouseEvent, pageviews)
```
ruslandoga, line 82: Because schemas are used instead of just tables, ecto checks for timestamps to be without microseconds, so they need to be truncated.

ruslandoga:
Closing in favour of https://github.com/plausible/analytics/pull/2733

PR #2736, opened 2023-03-08
Improve comparisons navigation

vinibrsl:
This pull request enhances the navigation for comparisons by implementing the following changes:

- Hides the comparison input behind the period picker, making the top bar less busy.
- Adds a key binding for the comparison input (assigned to <kbd>x</kbd>) alongside arrow navigation.
- Persists the selected comparison mode in Local Storage by domain, resulting in its persistence between refreshes.

Reviewing each commit separately may help :)

image

### DIFF
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -7,9 +7,12 @@ import classNames from 'classnames'
 import * as storage from './util/storage'
 
 const COMPARISON_MODES = {
+  'off': 'Disable comparison',
   'previous_period': 'Previous period',
   'year_over_year': 'Year over year',
 }
+
+const DEFAULT_COMPARISON_MODE = 'previous_period'
 
 export const COMPARISON_DISABLED_PERIODS = ['realtime', 'all']
 
@@ -26,15 +29,24 @@ const storeComparisonMode = function(domain, mode) {
   storage.setItem(`comparison_mode__${domain}`, mode)
 }
 
+export const isComparisonEnabled = function(mode) {
+  return mode && mode !== "off"
+}
+
 export const toggleComparisons = function(history, query, site) {
   if (!site.flags.comparisons) return
   if (COMPARISON_DISABLED_PERIODS.includes(query.period)) return
 
-  const defaultMode = getStoredComparisonMode(site.domain) || 'previous_period'
-  const toggle = query.comparison ? false : defaultMode
-  storeComparisonMode(site.domain, toggle)
+  if (isComparisonEnabled(query.comparison)) {
+    storeComparisonMode(site.domain, "off")
+    navigateToQuery(history, query, { comparison: "off" })
+  } else {
+    const storedMode = getStoredComparisonMode(site.domain)
+    const newMode = storedMode == "off" ? DEFAULT_COMPARISON_MODE : storedMode
 
-  navigateToQuery(history, query, { comparison: toggle })
+    storeComparisonMode(site.domain, newMode)
+    navigateToQuery(history, query, { comparison: newMode })
+  }
 }
 
 function DropdownItem({ label, value, isCurrentlySelected, updateMode }) {
@@ -55,7 +67,7 @@ function DropdownItem({ label, value, isCurrentlySelected, updateMode }) {
 const ComparisonInput = function({ site, query, history }) {
   if (!site.flags.comparisons) return null
   if (COMPARISON_DISABLED_PERIODS.includes(query.period)) return null
-  if (!query.comparison) return null
+  if (!isComparisonEnabled(query.comparison)) return null
 
   const updateMode = (key) => {
     storeComparisonMode(site.domain, key)
@@ -81,7 +93,6 @@ const ComparisonInput = function({ site, query, history }) {
               leaveFrom="transform opacity-100 scale-100"
               leaveTo="transform opacity-0 scale-95">
               <Menu.Items className="py-1 text-left origin-top-right absolute right-0 mt-2 w-56 rounded-md shadow-lg bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none z-10" static>
-                { DropdownItem({ label: "No comparison", value: false, isCurrentlySelected: !query.comparison, updateMode }) }
                 { Object.keys(COMPARISON_MODES).map((key) => DropdownItem({ label: COMPARISON_MODES[key], value: key, isCurrentlySelected: key == query.comparison, updateMode })) }
               </Menu.Items>
             </Transition>
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -24,28 +24,26 @@ function subscribeKeybinding(element) {
   }, [handleKeyPress])
 }
 
+function DropdownItem({ label, value, isCurrentlySelected, updateMode }) {
+  return (
+    <Menu.Item
+      key={value}
+      onClick={() => updateMode(value)}
+      disabled={isCurrentlySelected}>
+      {({ active }) => (
+        <button className={classNames("px-4 py-2 w-full text-left font-medium text-sm dark:text-white cursor-pointer", { "bg-gray-100 text-gray-900 dark:bg-gray-900 dark:text-gray-100": active, "font-bold": isCurrentlySelected })}>
+          { label }
+        </button>
+      )}
+    </Menu.Item>
+  )
+}
+
 const ComparisonInput = function({ site, query, history }) {
   if (!site.flags.comparisons) return null
   if (COMPARISON_DISABLED_PERIODS.includes(query.period)) return null
 
-  function update(key) {
-    navigateToQuery(history, query, { comparison: key })
-  }
-
-  function renderItem({ label, value, isCurrentlySelected }) {
-    return (
-      <Menu.Item
-        key={value}
-        onClick={() => update(value)}
-        disabled={isCurrentlySelected}>
-        {({ active }) => (
-          <button className={classNames("px-4 py-2 w-full text-left font-medium text-sm dark:text-white cursor-pointer", { "bg-gray-100 text-gray-900 dark:bg-gray-900 dark:text-gray-100": active, "font-bold": isCurrentlySelected })}>
-            { label }
-          </button>
-        )}
-      </Menu.Item>
-    )
-  }
+  const updateMode = (key) => navigateToQuery(history, query, { comparison: key })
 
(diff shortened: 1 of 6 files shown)


metmarkosaric:
looks much better i think @vinibrsl! what do you think about hiding the "Disabled" option from the menu when the comparison mode is not enabled? and how about perhaps renaming "Disabled" to something like "No comparison"?

vinibrsl:
> looks much better i think @vinibrsl! what do you think about hiding the "Disabled" option from the menu when the comparison mode is not enabled? and how about perhaps renaming "Disabled" to something like "No comparison"?

Good call @metmarkosaric! I renamed the option to "No comparison". Here's a preview:

[output.webm](https://user-images.githubusercontent.com/5093045/224023151-460fa9b0-f6eb-4897-83a2-2bbd9cab40aa.webm)

Sorry, I'm not sure I understand what you meant by "hiding the "Disabled" option from the menu when the comparison mode is not enabled".

vinibrsl:
@metmarkosaric I was discussing different options for this input with @ukutaht, and he suggested adding a toggle inside the date picker and showing the comparison input with some more details of what's being compared. I tried playing with his suggestion and here's a preview. I think not only this makes clearer the comparison dates, but also makes the top bar feel less bloated which was our concern.

Let me know your thoughts

[output.webm](https://user-images.githubusercontent.com/5093045/224176639-3e8c227c-1fa8-4dc5-ae18-f3d7b7261884.webm)

metmarkosaric:
@vinibrsl! looks great to me! is it possible to include the year too when comparing to a different year? would make it clearer what you're looking at when people share a screenshot etc

vinibrsl:
> @vinibrsl! looks great to me! is it possible to include the year too when comparing to a different year? would make it clearer what you're looking at when people share a screenshot etc

Good call! Here's a preview:

[output.webm](https://user-images.githubusercontent.com/5093045/224323525-38ff694b-7870-44d5-9a60-d35b681fb306.webm)

metmarkosaric:
@vinibrsl perfect!

assets/js/dashboard/historical.js
```diff
@@ -28,6 +28,8 @@ function Historical(props) {
     return null
   }
 
+  const [graphData, setGraphData] = React.useState(null)
```
vinibrsl, line 31: @ukutaht I'd like your feedback on this. The `Historical` component did not have access to `graphData`, and I needed this value so the comparison input knows the comparison result start and end dates to display this label:

image

`graphData` is part of the `VisitorGraph` component state, and this component is a child of `Historical`. I used `React.useState` and passed the `setGraphData` to the `VisitorGraph` as a prop. This function is called every time `graphData` has changed. By doing this, `Historical` now has access to `graphData` and is notified of its changes. I passed this `graphData` variable to `ComparisonInput` and now this component can react to changes in the `graphData` accordingly.

I'm not sure if that's the best way of sharing the `graphData` state, but that was the solution I found. Any thoughts?

assets/js/dashboard/datepicker.js
```diff
@@ -194,11 +195,24 @@ function DatePicker({query, site, history}) {
...
+      e: {date: formatISO(shiftDays(nowForSite(site), -1)), period: 'day'},
+      r: {period: 'realtime'},
+      w: {date: false, period: '7d'},
+      m: {date: false, period: 'month'},
+      y: {date: false, period: 'year'},
+      t: {date: false, period: '30d'},
+      s: {date: false, period: '6mo'},
+      l: {date: false, period: '12mo'},
+      a: {date: false, period: 'all'},
+    }
```
vinibrsl, line 209: I refactored these two arrays into a map with the possible keybindings.

ukutaht: I've wanted to do this for ages 🙏

assets/js/dashboard/comparison-input.js
```diff
@@ -12,51 +14,95 @@ const COMPARISON_MODES = {
...
+            <Transition
+              as={Fragment}
+              enter="transition ease-out duration-100"
+              enterFrom="transform opacity-0 scale-95"
+              enterTo="transform opacity-100 scale-100"
+              leave="transition ease-in duration-75"
+              leaveFrom="transform opacity-100 scale-100"
+              leaveTo="transform opacity-0 scale-95">
+              <Menu.Items className="py-1 text-left origin-top-right absolute right-0 mt-2 w-56 rounded-md shadow-lg bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none z-10" static>
+                { DropdownItem({ label: "No comparison", value: false, isCurrentlySelected: !query.comparison, updateMode }) }
```
ukutaht, line 98: I'd prefer label 'Disable comparison' instead of 'No comparison'

assets/js/dashboard/datepicker.js
```diff
@@ -352,6 +366,18 @@ function DatePicker({query, site, history}) {
...
             </div>
+            { !COMPARISON_DISABLED_PERIODS.includes(query.period) && site.flags.comparisons &&
+              <div className="py-1 date-option-group border-t border-gray-200 dark:border-gray-500">
+                <span
+                  onClick={() => {
+                    toggleComparisons(history, query, site)
+                    setOpen(false)
+                  }}
+                  className="px-4 py-2 text-sm leading-tight hover:bg-gray-100 dark:hover:bg-gray-900 hover:text-gray-900 dark:hover:text-gray-100 cursor-pointer flex items-center justify-between">
+                  Compare
```
ukutaht, line 377: When comparisons are enabled, I think this label should change to 'Disable comparison'.

ukutaht (commented):
Looks good overall 👍 Let me know what you think about changing the labels and removing dates from the comparison dropdown.

assets/js/dashboard/historical.js
vinibrsl: Yes, I also think the UI feels a bit glitchy after this change. That's probably because there are no loaders on the top bar - but I don't think adding is a good idea.

About the dates issue, I was thinking that could be solved by returning the `%Query{}` date range, which I guess would be correct?

Anyways, the screenshot you shared from Search Console looks good and I'll add it to the improvements after the release. Let's remove the dates from the input label :)

ukutaht: > About the dates issue, I was thinking that could be solved by returning the %Query{} date range, which I guess would be correct?

Yes, this would solve it most likely.

> Let's remove the dates from the input label :)

👍

vinibrsl: Removed in 9c370882c4299d13150df16ed643ffb72b9850f7

assets/js/dashboard/comparison-input.js
vinibrsl: Changed in cd54e4b27bc9486d0833662ab356a2a5cb0922ab

assets/js/dashboard/datepicker.js
vinibrsl: Changed in cd54e4b27bc9486d0833662ab356a2a5cb0922ab

assets/js/dashboard/util/date.js
```diff
@@ -49,8 +49,17 @@ export function formatDay(date) {
   }
 }
 
-export function formatDayShort(date) {
-  return `${date.getDate()} ${formatMonthShort(date)}`;
+export function formatDayShort(date, includeYear = false) {
```
ukutaht, line 52: Is this change still relevant?

assets/js/dashboard/util/date.js
```diff
@@ -49,8 +49,17 @@ export function formatDay(date) {
...
+  let formatted = `${date.getDate()} ${formatMonthShort(date)}`
+  if (includeYear) {
+    formatted += ` ${formatYearShort(date)}`
+  }
+
+  return formatted
+}
+
+export function formatYearShort(date) {
```
ukutaht, line 61: And this new function? Is it used?

assets/js/dashboard/comparison-input.js
```diff
@@ -4,59 +4,102 @@ import { navigateToQuery } from './query'
 import { Menu, Transition } from '@headlessui/react'
 import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import classNames from 'classnames'
+import * as storage from './util/storage'
 
 const COMPARISON_MODES = {
+  'off': 'Disable comparison',
```
ukutaht, line 10: Does this mean when comparisons are disabled it will add `comparison=off` to the URL?

I'm not a fan of the fact that then we will have two ways to represent the 'comparisons off' state: 
1. when the `comparison` query param is missing 
2. when the `comparison` query params is set to `off`

I feel like it would be cleaner to have a single representation of comparisons being disabled.

assets/js/dashboard/util/date.js
vinibrsl: Oops. Reverted in 23078e4d187a8ba8d5f60bcd43aead802a5e0f14

assets/js/dashboard/util/date.js
vinibrsl: 23078e4d187a8ba8d5f60bcd43aead802a5e0f14

assets/js/dashboard/comparison-input.js
vinibrsl: Thanks for bringing this up! To give this a bit more context, the `off` state was added as a solution to the following bug:

1. User turns comparison on, and URL is updated with `/dummy.site?comparison=previous_period`
2. User navigates to `/dummy.site` without any query params, and `comparison = previous_period` is loaded from Local Storage
3. User tries to disable comparison through the UI, but nothing happens because `navigateToQuery` with `{ comparison: false }` does not actually change the URL in this case

To solve this, I introduced this `off` state, which would force the URL to change.

Another solution to this problem could be to create a `updateComparison` function shared between `ComparisonInput` and `VisitorGraph` to notify each other of changes in the comparison state, without relying on the URL. I think that's similar to what we do when changing the metric state. Do you think that's a good solution?

ukutaht: Hmm yeah mixing query params and localStorage is a bit problematic...

> Another solution ... without relying on the URL

I do think it's important to keep the comparison in the URL so that if I grab a URL with comparisons enabled and send it to you, you'll see the same thing as I do.

> User navigates to /dummy.site without any query params, and comparison = previous_period is loaded from Local Storage

This also creates a mismatch where the URL isn't completely in sync with what's shown on the dashboard (url has no comparison but the UI is showing it). But I think we can live with it. Let's move forward as-is.

ukutaht (approved):

PR #2737, opened 2023-03-09
Added IPv6 Finch support (via the environment-variable FINCH_IPV6)

DKPillo:
### Background

We are working with [Hetzner Cloud](https://www.hetzner.com/cloud) and have a setup, where only the proxy server is available via IPv4. The Docker hosts used to run services like plausible are attached to the network via IPv6.

Now we run unto the problem that plausible was not able to connect to Google API to connect with Analytics/Search Console.

After some research I found the following issue on finch: [IPv6 Documentation](https://github.com/sneako/finch/issues/163)

### Changes

Updates `lib/plausible/application.ex` and enables IPv6 support for Finch by setting `transport_opts: [inet6: true]` when environment variable `FINCH_IPV6` is set.

### Tests

### Changelog

### Documentation
- [X] [Docs](https://github.com/plausible/docs) have been updated

### Dark mode

### DIFF
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -50,13 +50,13 @@ defmodule Plausible.Application do
   end
 
   defp finch_pool_config() do
-    finch_transport_opts = Application.get_env(:plausible, :finch_transport)
+    finch_conf = Application.get_env(:plausible, :finch)
     base_config = %{
       "https://icons.duckduckgo.com" => [
-        conn_opts: [transport_opts: finch_transport_opts]
+        conn_opts: [transport_opts: finch_conf[:transport_opts]]
       ],
       default: [
-        conn_opts: [transport_opts: finch_transport_opts]
+        conn_opts: [transport_opts: finch_conf[:transport_opts]]
       ]
     }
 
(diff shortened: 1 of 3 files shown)


DKPillo:
Merge Request for [Docs](https://github.com/plausible/docs/pull/365)

aerosol (approved):
Hey @DKPillo, looks good.

lib/plausible/application.ex
```diff
@@ -92,14 +95,18 @@ defmodule Plausible.Application do
...
+        |> Map.put(google_conf[:api_url], conn_opts: [transport_opts: finch_transport_opts()])
+        |> Map.put(google_conf[:reporting_api_url], conn_opts: [transport_opts: finch_transport_opts()])
 
       true ->
         pool_config
     end
   end
 
+  defp finch_transport_opts() do
+      maybe_ipv6 = if System.get_env("FINCH_IPV6"), do: [timeout: 15_000, inet6: true], else: [timeout: 15_000]
```
ruslandoga, line 107: It doesn't seem like there are any other `System` calls in this module. Maybe, to keep this style going, we could move `System.get_env("FINCH_IPV6")` to `config/runtime.exs` and do an `Application.get_env` call here instead?

aerosol: :facepalm: good call @ruslandoga, 100%

aerosol (changes requested):
https://github.com/plausible/analytics/pull/2737#discussion_r1141980454

DKPillo:
Hi @aerosol, @ruslandoga 

Thank you, for your feedback.
I have updated the code so that `plausible/application.ex` uses `Application.get_env` to load the transport_opts for finch.

Unfortunately it does not work as intended and I have some debugging to do.
I will update the thread as soon as I have a working solution again.

aerosol:
Hey @DKPillo,

Thanks for coming back to us on this.

> Unfortunately it does not work as intended and I have some debugging to do.

Could you elaborate a bit on what doesn't work exactly?

aerosol:
hi @DKPillo, just checking if you still need this or if you encounter any problems. Otherwise I'm going to close the issue.

DKPillo:
Hi @aerosol sorry for the bad response time. I did not find enough time besides work to fix this issue as I wanted to.

I ended up fixing my problem by configuring a NAT to allow the server to access resources via IPv4 network.

PR #2738, opened 2023-03-09
Bug fix - show `views_per_visit` when stats include imported data

RobertJoonas:
### Changes

The current behavior is that we don't display this metric at all if imported data is displayed.

This PR fixes the bug where the new `views_per_visit` metric is not displayed when imported data is shown on the dashboard. This adds a new function clause to select this metric as well, when imported stats are being merged with native data.

However, the new function simply fetches the metric from native Plausible tables (since this metric is not imported from GA). To make it clear to the user, the metric tooltip will display the message "_Based only on native data_", when the visible stats include GA data.

That means, for example, when we're viewing a period that only has GA data, the behavior will be this (see video):

https://user-images.githubusercontent.com/56999674/224135238-4b47eabe-bb89-41ee-94b7-972e7072bf9f.mov

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -45,7 +45,7 @@ export default class TopStats extends React.Component {
     statName = stat.value === 1 ? statName.slice(0, -1) : statName
 
     const { topStatData, lastLoadTimestamp } = this.props
-    const showingImported = topStatData && topStatData.imported_source && topStatData.with_imported
+    const showingImported = topStatData?.imported_source && topStatData?.with_imported
 
     return (
       <div>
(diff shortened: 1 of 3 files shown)


assets/js/dashboard/stats/graph/top-stats.js
```diff
@@ -44,10 +44,14 @@ export default class TopStats extends React.Component {
     let statName = stat.name.toLowerCase()
     statName = stat.value === 1 ? statName.slice(0, -1) : statName
 
+    const { topStatData, lastLoadTimestamp } = this.props
+    const showingImported = topStatData && topStatData.imported_source && topStatData.with_imported
```
vinibrsl, line 48: Minor suggestion, feel free to ignore :)

```suggestion
    const showingImported = topStatData?.imported_source?.with_imported
```

vinibrsl (approved):

assets/js/dashboard/stats/graph/top-stats.js
RobertJoonas: That wouldn't work, since `with_imported` is a property of `topStatData`, not `imported_source`, but I did use optional chaining here: https://github.com/plausible/analytics/pull/2738/commits/ab8601a2f31225c77ce179305ac09889c528294b

PR #2749, opened 2023-03-14
Support multiple filters - backend

ukutaht:
### Changes

Adds support for `member` filter type for UI filters.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -30,6 +30,7 @@ defmodule Plausible.Stats.Base do
     end
   end
 
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def query_events(site, query) do
     {first_datetime, last_datetime} = utc_boundaries(query, site)
 
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -52,13 +52,13 @@ defmodule Plausible.Stats.Base do
 
         {:matches_member, glob_exprs} ->
           page_regexes = Enum.map(glob_exprs, &page_regex/1)
-          from(e in q, where: fragment("multiMatchAny(?, array(?))", e.pathname, ^page_regexes))
+          from(e in q, where: fragment("multiMatchAny(?, ?)", e.pathname, ^page_regexes))
 
         {:not_matches_member, glob_exprs} ->
           page_regexes = Enum.map(glob_exprs, &page_regex/1)
 
           from(e in q,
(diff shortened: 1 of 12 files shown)


lib/plausible/stats/filters.ex
```diff
@@ -46,7 +46,7 @@ defmodule Plausible.Stats.Filters do
 
             Map.put(new_filters, "event:goal", filter)
 
-          name in (@visit_props ++ ["goal"]) ->
+          name in @visit_props ->
```
ukutaht, line 49: The `["goal"]` part of this match was unreachable code

lib/plausible/stats/base.ex
```diff
@@ -78,16 +82,28 @@ defmodule Plausible.Stats.Base do
...
+        {:member, clauses} ->
+          {events, pages} = Enum.split_with(clauses, fn {type, _} -> type == :event end)
+          events = Enum.map(events, fn {_, val} -> val end)
+          pages = Enum.map(pages, fn {_, val} -> val end)
+          from(e in q, where: e.pathname in ^pages or e.name in ^events)
+
+        {:not_member, clauses} ->
+          {events, pages} = Enum.split_with(clauses, fn {type, _} -> type == :event end)
+          events = Enum.map(events, fn {_, val} -> val end)
+          pages = Enum.map(pages, fn {_, val} -> val end)
```
vinibrsl, line 104: You can also use `group_by/3` to group and map at the same time.

```elixir
list = [{:event, 1}, {:event, 2}, {:page, 4}, {:page, 3}]
Enum.group_by(list, fn {k, _v} -> k end, fn {_k, v} -> v end)
# => %{event: [1, 2], page: [4, 3]}
```

vinibrsl (approved):
Looks good and well tested :)

aerosol (commented):
Just a general concern: should we limit the number of clauses applicable in a single query?

ukutaht:
> Just a general concern: should we limit the number of clauses applicable in a single query?

Good point. We've had support for multiple filters in the API for a while now with no issues. I think limiting to something like 10 would make sense in general, but for the dashboard it's simpler to apply that limit in the frontend rather than round-tripping to the server for the validation.

I'm tempted to leave it for now. We have to tweak the frontend filter selector to set a maximum number of options anyways (`prop_key` can only support one). So we can look at making that limit generic and applying a limit of 10 to all filters except for `prop_key` which will be limited to 1.

RobertJoonas (approved):
Looks good to me!

PR #2752, opened 2023-03-14
Stop treating page filter as an entry_page filter

RobertJoonas:
### Changes

This PR stops treating a page filter as an entry page filter when querying sessions. After this change, using a page filter will return all sessions that have visited the page at least once (like goal conversions).

### Tests

### Changelog

### Documentation

### Dark mode


test/plausible_web/controllers/CSVs/30d-filter-goal/visitors.csv
```diff
@@ -1,5 +1,5 @@
 date,visitors,pageviews,bounce_rate,visit_duration
```
ukutaht, line 1: Why did the bounce rate and visit_duration disappear from this export?

RobertJoonas: It's a good question to ask. I didn't pay close attention to it and removed it as just pure wrong behaviour at first, but this is currently a bug where the `sessions_timeseries` function totally ignores the goal filter.

CSV export is the only place where that bug can actually happen. This is because:

1. We don't allow displaying session metrics on the graph when a goal filter is applied (instead we show only conversions in top stats)
2. The Stats API doesn't allow querying session metrics with an `event:name` filter

However, in the CSV export, we show `[visitors, pageviews, bounce_rate, visit_duration]` instead of `[Unique visitors, Unique conversions, Total conversions and Conversion rate]` that are present in the dashboard, when fitlering by a goal. We should probably change that, and make the CSV export aligned with what the dashboard is showing

ukutaht: Ah yeah that makes sense.

> We should probably change that, and make the CSV export aligned with what the dashboard is showing

Agreed, I think that's probably the best way to deal with this. Should be a pretty small change, could you make a follow-up PR for that as well?

RobertJoonas: Sure thing! I'm on it

metmarkosaric: nr 1 sounds good to me @RobertJoonas!

ukutaht (approved):

PR #2753, opened 2023-03-14
Custom comparison mode

vinibrsl:
### Changes

This pull request implements the custom comparison mode using a date range.

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -103,13 +103,17 @@ const ComparisonInput = function({ site, query, history }) {
   }
 
   const calendar = React.useRef(null)
+
   const flatpickrOptions = {
     mode: 'range',
     showMonths: 1,
     maxDate: 'today',
+    minDate: parseUTCDate(site.statsBegin),
     animate: true,
     static: true,
-    onChange: ([from, to]) => updateMode("custom", formatISO(from), formatISO(to))
+    onChange: ([from, to]) => {
+      if (from && to) updateMode("custom", formatISO(from), formatISO(to))
+    }
   }
 
   return (
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -1,15 +1,18 @@
 import React, { Fragment } from 'react'
-import { withRouter } from "react-router-dom";
+import { withRouter } from 'react-router-dom'
 import { navigateToQuery } from './query'
 import { Menu, Transition } from '@headlessui/react'
 import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import classNames from 'classnames'
 import * as storage from './util/storage'
+import Flatpickr from 'react-flatpickr'
+import { formatISO, parseUTCDate, formatDayShort } from './util/date.js'
 
 const COMPARISON_MODES = {
   'off': 'Disable comparison',
   'previous_period': 'Previous period',
   'year_over_year': 'Year over year',
+  'custom': 'Custom period',
 }
 
 const DEFAULT_COMPARISON_MODE = 'previous_period'
@@ -26,6 +29,7 @@ export const getStoredComparisonMode = function(domain) {
 }
 
 const storeComparisonMode = function(domain, mode) {
+  if (mode == "custom") return
   storage.setItem(`comparison_mode__${domain}`, mode)
 }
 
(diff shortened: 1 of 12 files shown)


assets/js/dashboard/comparison-input.js
```diff
@@ -26,6 +29,7 @@ export const getStoredComparisonMode = function(domain) {
 }
 
 const storeComparisonMode = function(domain, mode) {
+  if (mode == "custom") return
```
vinibrsl, line 32: I decided not to store custom comparison mode in Local Storage. I feel like comparing with a custom date is very specific to a single dashboard session, and saving that between refreshes is not that useful. Let me know your thoughts.

assets/js/dashboard/comparison-input.js
```diff
@@ -49,17 +53,30 @@ export const toggleComparisons = function(history, query, site) {
   }
 }
 
-function DropdownItem({ label, value, isCurrentlySelected, updateMode }) {
+function DropdownItem({ label, value, isCurrentlySelected, updateMode, calendar }) {
+  const click = () => {
+    if (value == "custom") {
+      // https://github.com/flatpickr/flatpickr/issues/399#issuecomment-260007013
+      setTimeout(() => calendar.current.flatpickr.open(), 0)
```
vinibrsl, line 60: Without this `setTimeout` call, the calendar just doesn't open. A way around this is using the `setTimeout` function as suggested here: https://github.com/flatpickr/flatpickr/issues/399#issuecomment-260007013

assets/js/dashboard/comparison-input.js
```diff
@@ -49,17 +53,31 @@ export const toggleComparisons = function(history, query, site) {
   }
 }
 
-function DropdownItem({ label, value, isCurrentlySelected, updateMode }) {
+function DropdownItem({ label, value, isCurrentlySelected, updateMode, calendar }) {
+  const click = () => {
+    if (value == "custom") {
+      // https://github.com/flatpickr/flatpickr/issues/399#issuecomment-260007013
+      // FIXME: Use setState to prevent this issue
+      setTimeout(() => calendar.current.flatpickr.open(), 100)
```
ukutaht, line 61: Are you planning to address the race condition in this PR?

assets/js/dashboard/comparison-input.js
```diff
@@ -69,19 +87,42 @@ const ComparisonInput = function({ site, query, history }) {
...
+      const from = parseUTCDate(query.compare_from)
+      const to = parseUTCDate(query.compare_to)
+      return `${formatDayShort(from, false)} - ${formatDayShort(to, false)}`
+    } else {
+      return COMPARISON_MODES[query.comparison]
+    }
+  }
+
+  const calendar = React.useRef(null)
+  const flatpickrOptions = {
```
ukutaht, line 106: I think we should also limit the selection in the beginning of the site's creation date. Example in datepicker: https://github.com/plausible/analytics/blob/d0895efbddb0e4e3e6e0d36bf97c5b43302e2fca/assets/js/dashboard/datepicker.js#L394

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -140,8 +144,19 @@ defmodule PlausibleWeb.Api.StatsController do
     Enum.map(timeseries, fn row -> row[metric] || 0 end)
   end
 
-  defp label_timeseries(timeseries) do
-    Enum.map(timeseries, & &1.date)
+  defp label_timeseries(main_result, nil) do
+    Enum.map(main_result, & &1.date)
+  end
+
+  defp label_timeseries(main_result, comparison_result, empty \\ "") do
```
ukutaht, line 151: Is the final optional argument (`empty`) ever used? Could it be a module attribute instead?

ukutaht: I was also wondering if it's better to use a specific value for representing these blanks. Something like `'__blank__'`, for example. Things like `nil`, empty string, etc. can also be returned by some other part of the code by accident. If we use a specific value like `'__blank__'` we know for sure that it's supposed to be a blank label :)

assets/js/dashboard/comparison-input.js
vinibrsl: Good catch! Fixed in b1d8dbbc8914912a8c192225b9802cf4d31cf9f4. I also noticed that the datepicker.js implementation is a bit off. I fixed in that commit, but I'm not sure if that was intentional. Flatpickr minDate is inclusive, meaning that we don't need to shift back 1 day.

The live demo stats start date is set to Jan 23, 2019, but the date picker starts in January 21: https://plausible.io/plausible.io?period=custom&from=2019-01-21&to=2019-01-22

lib/plausible_web/controllers/api/stats_controller.ex
vinibrsl: Good call! Changed in 2f9a4fb0a6d748af6d99604211dffea2f9429ee5

ukutaht (approved):

PR #2733, opened 2023-03-08
Switch to new clickhouse adapter (ch/chto)

aerosol:
### Changes

This PR makes a switch from [clickhousex](https://github.com/plausible/clickhousex) + [ecto_clickhouse](https://github.com/plausible/clickhouse_ecto) to [ch](https://github.com/ruslandoga/ch) + [chto.](https://github.com/ruslandoga/chto)
Additionally it adapts the current codebase to work well with the new adapter.

### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #2762, opened 2023-03-16
Implement Numeric IDs migration

aerosol:
### Changes

First pass on Numeric IDs Clickhouse migration.
For now the PR contains the necessary SQL files + a coordination script.
I expect things to change going forward, but submitting this as an interim PR - no action required yet, should be completely transparent.

I am trying to work on an integration test on the side but it's currently painful due to clickhouse container having to call the host in order to retrieve domains lookup dictionary via HTTP. Awaiting clarity on clickhouse-postgres connectivity for that.

### Tests

### Changelog

### Documentation

### Dark mode


aerosol:
Rebased & adjusted for #2733

lib/plausible/data_migration/numeric_ids.ex
```diff
@@ -0,0 +1,139 @@
...
+        Application.get_env(:plausible, Plausible.IngestRepo)[:url]
+      )
+
+    max_threads =
+      "NUMERIC_IDS_MIGRATION_MAX_THREADS" |> System.get_env("16") |> String.to_integer()
+
+    # TBD: There's most likely a bug in Clickhouse defining Postgres dictionaries,
+    # we'll use a static URL for now
+    dict_url = Keyword.get(opts, :dict_url) || System.get_env("DOMAINS_DICT_URL") || ""
+    dict_password = Keyword.get(opts, :dict_url) || System.get_env("DOMAINS_DICT_PASSWORD") || ""
```
vinibrsl, line 26: This is fetching the `:dict_url` from `opts` in both variables. Should we change the second to `dict_password`? Or is the URL already authenticated somehow?

vinibrsl (approved):
Looks good! Awesome work, @aerosol.

lib/plausible/data_migration/numeric_ids.ex
aerosol: For our case the pass is already kept in the vault and provided via env vars, but good call, it's an omission on my behalf, thanks! 2823261f

PR #2768, opened 2023-03-20
Add regular ecto migration with numeric ids v2 schemas

aerosol:
### Changes

This work is on top of, and against #2762.

  Normally, for live environments the migration will be done via
  `DataMigration.NumericIDs` module (TBD). In which case `PASS_V2_SCHEMA_MIGRATION`
  environment variable needs to be set, to only make the standard migrate
  command write an entry into `schema_migrations` table.

  For tests, and entirely new self-hosted instances however, 
  we want to keep the ability of preparing the database without enforcing 
  any data migration.
  

### Tests

### Changelog

### Documentation

### Dark mode


aerosol:
Rebased against #2762 after #2733

priv/ingest_repo/migrations/20230320094327_create_v2_schemas.exs
```diff
@@ -0,0 +1,33 @@
...
+  we want to keep the ability of preparing the database without enforcing 
+  any data migration.
+  """
+
+  use Ecto.Migration
+
+  use Plausible.DataMigration, dir: "NumericIDs"
+
+  @cluster? false
+    @settings "SETTINGS index_granularity = 8192"
```
vinibrsl, line 18: Apparently `.exs` files are not being formatted neither checked. Not related to this pull request, but something for us to take a look.

```suggestion
  @settings "SETTINGS index_granularity = 8192"
```

vinibrsl (approved):

PR #2773, opened 2023-03-20
Support multiple filters - frontend

ukutaht:
### Changes

Implements the ability to select multiple filters which are automatically combined with logical OR.

https://user-images.githubusercontent.com/3731516/226345626-243c3ce6-539d-4b20-8920-6b6fd9e29145.mov

Unfortunately I could not get Headless UI to work properly with multiple selections. This PR contains a hand-rolled Combobox component from scratch. 

Depends on https://github.com/plausible/analytics/pull/2749
The diff will get smaller once the previous PR is merged.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -4,10 +4,28 @@ import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import debounce from 'debounce-promise'
 import classNames from 'classnames'
 
+function Option({isHighlighted, onClick, onMouseEnter, text}) {
+  const className = classNames('relative cursor-pointer select-none py-2 px-3', {
+    'text-gray-900 dark:text-gray-300': !isHighlighted,
+    'bg-indigo-600 text-white': isHighlighted,
+  })
+
+  return (
+    <li
+      className={className}
+      onClick={onClick}
+      onMouseEnter={onMouseEnter}
+    >
+      <span className="block truncate">{text}</span>
+    </li>
+  )
+}
+
 export default function PlausibleCombobox(props) {
   const [options, setOptions] = useState([])
   const [loading, setLoading] = useState(false)
   const [isOpen, setOpen] = useState(false);
+  const [input, setInput] = useState('');
   const [highlightedIndex, setHighlightedIndex] = useState(0);
   const searchRef = useRef(null);
   const containerRef = useRef(null);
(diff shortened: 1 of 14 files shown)


aerosol (commented):
I don't think this is a regression, so not sure if worth looking into right here. It crashes with the following input: `(/**`

image

and similarily for `(a|a)**`

assets/js/dashboard/components/combobox.js
```diff
@@ -1,106 +1,216 @@
...
+      >
+        <ul ref={listRef} className="z-50 absolute mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 text-base shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none sm:text-sm dark:bg-gray-900">
+          { loading && (
+            <div className="relative cursor-default select-none py-2 px-4 text-gray-700 dark:text-gray-300">
+              Loading options...
+            </div>
+          )}
+          { noMatchesFound && (
+            <div className="relative cursor-default select-none py-2 px-4 text-gray-700 dark:text-gray-300">
+              No matches found in the current dashboard. Try selecting a different time range or searching for something different
```
aerosol, line 184: For arbitrary non-existing values we show that for every filter but "Page". Is that intentional?

ukutaht: Yes. This is because the 'freeChoice' filters are only allowed for the page filter group. Since page filters support both wildcards and the `contains` type, it must be possible to filter by a value that doesn't strictly correspond to an existing site.

However, for other filters there is no option of wildcards or `contains` filter. In that case if there are no matches found we don't give the user a choice to filter by whatever value they've typed in.

RobertJoonas:
Found this one bug:

https://user-images.githubusercontent.com/56999674/227911789-ac454d4e-be3d-404f-ab3d-40a726119ccf.mov

RobertJoonas (approved):
Looks good to me in general, other than the one bug I found, that crashes the dashboard when an is_not filter type is selected with a value.

assets/js/dashboard/components/combobox.js
```diff
@@ -1,106 +1,225 @@
...
+    }
+  }
+
+  function isDisabled(option) {
+    return props.values.some((val) => val.value === option.value)
+  }
 
   function fetchOptions(query) {
     setLoading(true)
+    setOpen(true)
```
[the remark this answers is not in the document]
ukutaht: Yeah you're right, this one is not needed :)

PR #2777, opened 2023-03-21
Implement v2 events/sessions schema modules

aerosol:
### Changes

On top of, and against #2768 - this PR only adds new clickhouse schemas for the _v2 tables. Ref #2762 

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/clickhouse_session_v2.ex
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def changeset(session, attrs) do
+    session
+    |> cast(attrs, [
+      :hostname,
+      :site_id,
+      :entry_page,
+      :exit_page,
+      :fingerprint,
```
ukutaht, line 58: Just noticed this attribute which is unused AFAIK. Can be deleted.

ukutaht (approved):

lib/plausible/clickhouse_session_v2.ex
aerosol: Oh great, looks like even the changeset function can go from both session schemas 9f928615

PR #2780, opened 2023-03-21
Conditionally support switching between v1 and v2 clickhouse schemas

aerosol:
### Changes

This is on top of, ~~and against~~ https://github.com/plausible/analytics/pull/2777

This PR implements a new environment variable `V2_MIGRATION_DONE`, which, if set, makes the `Plausible.v2?()` check return true at runtime. Based on that check, the application decides whether to use `events`/`sessions` or `events_v2`/`sessions_v2` tables.

Once the migration from https://github.com/plausible/analytics/pull/2762 is live, an application restart is required with the said variable set - it acts like a feature flag, enabling the use of `events_v2` and `sessions_v2` tables.

Because we have to maintain two versions of the same code essentially, a separate CI check is added that will run another set of tests against the v2 switch, isolated. This is meant to assure that until the old code is removed, we are not introducing any regressions.

By the way, an implicit contract between multiple tests has been removed -- we used to have a ClickhouseSetup module run at `test_helper.exs` that would pre-populate clickhouse with some events that were further on assumed to exist by many tests. From now on, each test ensures its own, isolated setup.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/site/admin.ex
RobertJoonas, line 1: > In a separate changeset we will implement simplified
site transfer for when v2 migration is complete.
The new transfer will only rename the site domain in postgres
and keep track of the original site prior to the transfer
so we keep an ingestion grace period until the customers
redeploy their scripting.

I'm thinking that maybe we wouldn't need an "ingestion grace period". The instruction for changing a domain name could just be the following:

1. Make sure your Plausible snippet is using `data-site-id` instead of `data-domain` + verify that your stats are still tracked after making the change
2. In your site settings, simply change the name of your dashboard

For that we would have to support ingestion by both `site_id` and `domain` fields, but I guess that's pretty doable, right? With that approach we could get rid of the CRM transfer_data for good.

test/plausible_web/controllers/api/stats_controller/suggestions_test.exs
```diff
@@ -5,23 +5,37 @@ defmodule PlausibleWeb.Api.StatsController.SuggestionsTest do
     setup [:create_user, :log_in, :create_site]
 
     test "returns suggestions for pages without a query", %{conn: conn, site: site} do
+      populate_stats(site, [
+        build(:pageview, timestamp: ~N[2019-01-01 23:00:00], pathname: "/"),
+        build(:pageview, timestamp: ~N[2019-01-01 23:00:00], pathname: "/register"),
+        build(:pageview, timestamp: ~N[2019-01-01 23:00:01], pathname: "/contact"),
+        build(:pageview, timestamp: ~N[2019-01-01 23:00:01], pathname: "/irrelevant")
+      ])
+
```
RobertJoonas, line 14: Hmm, just wondering where were those stats populated before?

test/plausible/auth/auth_test.exs
```diff
@@ -18,7 +18,11 @@ defmodule Plausible.AuthTest do
 
     test "is true if user does have events" do
       user = insert(:user)
-      insert(:site, members: [user], domain: "test-site.com")
+      site = insert(:site, members: [user])
+
+      populate_stats(site, [
+        build(:pageview)
+      ])
```
RobertJoonas, line 25: Similar case to my previous question about where this data comes from... Is this because stats for `test-site.com` were previously populated by other tests and the test database is shared? And now we're generating a unique site name in each test, I believe?

test/workers/send_check_stats_emails_test.exs
```diff
@@ -25,6 +25,8 @@ defmodule Plausible.Workers.SendCheckStatsEmailsTest do
   test "does not send an email if the user has configured a weekly report" do
     user = insert(:user, inserted_at: days_ago(9), last_seen: days_ago(7))
     site = insert(:site, domain: "test-site.com", members: [user])
```
RobertJoonas, line 27: I guess the `domain: "test-site.com"` is no longer required here?

lib/plausible/site/admin.ex
aerosol: Nice thinking @RobertJoonas. We wanted to avoid any new APIs for now but, also to make it as painless as possible, support domains still as "external site ids" without leaking the internal identifiers on the outside. The grace period is only to improve the UX of transferring sites and get customers some extra time to redeploy their sites. Also we don't guarantee site id will be immutable - having it only known internally, we're free to recreate a site with the same domain but a new internal identifier, effectively leading to a fresh start (stats reset) for example.

test/plausible_web/controllers/api/stats_controller/suggestions_test.exs
aerosol: In `ClickhouseSetup` module mentioned in the PR description

test/plausible/auth/auth_test.exs
aerosol: Yes. There was an implicit contract executed in test_helper.exs - take a look there

test/workers/send_check_stats_emails_test.exs
aerosol: I guess not, doesn't hurt either

lib/plausible/site/admin.ex
RobertJoonas: Aah, I see! Thanks for explaining, this is brilliant! 💡

ukutaht (approved):

test/support/test_utils.ex
```diff
@@ -37,7 +37,7 @@ defmodule Plausible.TestUtils do
   def create_site(%{user: user}) do
     site =
       Factory.insert(:site,
-        domain: "test-site.com",
+        domain: "test-site-#{Base.encode16(:crypto.strong_rand_bytes(10), case: :lower)}.com",
```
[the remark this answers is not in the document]
aerosol: https://github.com/plausible/analytics/pull/2780/commits/8360ff47a6ff2a4c1ca6c67776261468f7998114

PR #2793, opened 2023-03-27
Update location dependency (faster startup)

aerosol:
### Changes

Ref https://github.com/plausible/location/pull/5

@vinibrsl I like to get your feedback on this, I know you're familiar with location lookups - a sanity check would be much appreciated, thanks!

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl:
I checked out your branch and tried importing from Google Analytics. The import was successful but I noticed it's a \*bit\* slower after the changes. This is probably related to [this change](https://github.com/plausible/location/commit/2c672d97f1ee872c5a4161c0945aafcb85ec8e0d#diff-5dd05df20699804442051125ca956ba3506f52943b2964ccd0fb147ac83e8c6fR58) - moving from an ETS lookup to a select.

I'm fine with merging this PR, but we should keep an eye on the Google Analytics Oban job queue. We only run one import at a time, and depending on the imports that we get, the queue can get busy. For the two GA sites I tested, the performance was just a bit slower, which is fine, but we never know what to expect from GA imports.

aerosol:
@vinibrsl thanks a lot, I'll have a look and profile it!

vinibrsl (approved):

PR #2797, opened 2023-03-28
Fix open state in combobox

ukutaht:
Fixes a small issue with combobox not opening when:
1. User has selected an option
2. Starts typing again


ukutaht:
Added a couple more improvements:
1. More streamlined UI for `contains` filter type. It failed to fetch suggestions when the `contains` filter type was used
2. Enable `contains` filter type for utm filter group

aerosol (approved):

PR #2802, opened 2023-03-29
Migration: enable domain changes (upcoming feature)

aerosol:
### Changes

Ref #1080 

First pass to migrate `sites` table.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20230328062644_allow_domain_change.exs
+++ b/priv/repo/migrations/20230328062644_allow_domain_change.exs
@@ -15,7 +15,7 @@ defmodule Plausible.Repo.Migrations.AllowDomainChange do
     BEGIN
       IF EXISTS (
         SELECT 1 FROM sites
-         WHERE NEW.domain = domain_changed_from
+         WHERE (NEW.domain = domain_changed_from AND NEW.id != id)
          OR (OLD IS NULL AND NEW.domain_changed_from = domain)
       ) THEN
         RAISE unique_violation USING CONSTRAINT = 'domain_change_disallowed';
--- a/priv/repo/migrations/20230328062644_allow_domain_change.exs
+++ b/priv/repo/migrations/20230328062644_allow_domain_change.exs
@@ -1,0 +1,49 @@
+defmodule Plausible.Repo.Migrations.AllowDomainChange do
+  use Ecto.Migration
+
+  def up do
+    alter table(:sites) do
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20230328062644_allow_domain_change.exs
```diff
@@ -0,0 +1,48 @@
...
+
+    create unique_index(:sites, :domain_changed_from)
+    create index(:sites, :domain_changed_at)
+
+    execute """
+    CREATE OR REPLACE FUNCTION check_domain() RETURNS TRIGGER AS $$
+    BEGIN
+      IF EXISTS (
+        SELECT 1 FROM sites
+        WHERE NEW.domain = domain_changed_from
```
ukutaht, line 18: I would test this but in a bit of a rush at the moment. I'll post now because I won't be able to test until tomorrow morning. Comments below might be completely incorrect because of this :)

I *think* it's still possible to get the DB into an inconsistent state like this:

```sql
INSERT INTO sites (domain) VALUES ('test-site.com');
INSERT INTO sites (domain, domain_changed_from) VALUES ('ignore.com', 'test-site.com');
```

Not sure if it's something you considered? I don't expect the app to do this but I can see myself messing it up while doing CS for a customer :) Would feel more comfortable if the database would refuse to get into an inconsistent state altogether. Something like 

```sql
BEGIN
       IF EXISTS (
         SELECT 1 FROM sites
         WHERE NEW.domain = domain_changed_from
         OR NEW.domain_changed_from = domain
         # NEW.domain = domain already handled by unique constrain on domain
         # NEW.domain_changed_from = domain_changed_from already handled by unique constrain on domain_changed_from
       ) THEN
         RAISE unique_violation USING CONSTRAINT = 'domain_change_disallowed';
       END IF;
       RETURN NEW;
     END;
```

aerosol: I pushed the wrong change by mistake but your suggestion is also incorrect in the end.
It'll prevent updating a single row - because the trigger query will return 1 for itself. ⌛

aerosol: Force pushed corrected version: https://github.com/plausible/analytics/pull/2802/commits/b2ff72dd19ca7c54c25b13f4e44015f9f4a9e79a

ukutaht (approved):

PR #2803, opened 2023-03-29
Allow domain change

aerosol:
### Changes

Ref https://github.com/plausible/analytics/discussions/1080

This PR implements the ability to change a site's domain keeping the collected stats intact.

We grant users with 72 hours transition period in which they're expected to update their API credentials and tracker scripts to point at the new domain. During that time, both old and new domains are accepted within incoming traffic and are made unavailable for registration.

This change will be only live after V2 migration is done (Ref #2762 #2780) hence the "feature flag" required for it is `V2_MIGRATION_DONE=1` variable set.

https://user-images.githubusercontent.com/173738/228543688-66901701-7026-4807-b800-5ea2e52bb564.mp4

https://user-images.githubusercontent.com/173738/228543470-ba820ac6-7c59-4149-9ed0-f7ef2bb48c8c.mp4

### Tests

### Changelog
  Updated

### Documentation
  https://github.com/plausible/docs/pull/380

### Dark mode

### DIFF
--- a/priv/repo/migrations/20230328062644_allow_domain_change.exs
+++ b/priv/repo/migrations/20230328062644_allow_domain_change.exs
@@ -15,7 +15,7 @@ defmodule Plausible.Repo.Migrations.AllowDomainChange do
     BEGIN
       IF EXISTS (
         SELECT 1 FROM sites
-         WHERE NEW.domain = domain_changed_from
+         WHERE (NEW.domain = domain_changed_from AND NEW.id != id)
          OR (OLD IS NULL AND NEW.domain_changed_from = domain)
       ) THEN
         RAISE unique_violation USING CONSTRAINT = 'domain_change_disallowed';
--- a/test/plausible/site/domain_test.exs
+++ b/test/plausible/site/domain_test.exs
@@ -38,6 +38,19 @@ defmodule Plausible.Site.DomainTest do
     assert error =~ "This domain cannot be registered"
   end
 
+  test "a single site's domain can be changed back and forth" do
+    site1 = insert(:site, domain: "foo.example.com")
+    site2 = insert(:site, domain: "baz.example.com")
+
+    assert {:ok, _} = Domain.change(site1, "bar.example.com")
+
+    assert {:error, _} = Domain.change(site2, "bar.example.com")
+    assert {:error, _} = Domain.change(site2, "foo.example.com")
+
+    assert {:ok, _} = Domain.change(Repo.reload!(site1), "foo.example.com")
+    assert {:ok, _} = Domain.change(Repo.reload!(site1), "bar.example.com")
+  end
+
   test "change info is cleared when the grace period expires" do
     site = insert(:site)
 


lib/plausible/site.ex
```diff
@@ -173,19 +195,17 @@ defmodule Plausible.Site do
     |> Timex.to_date()
   end
 
-  defp clean_domain(changeset) do
+  defp clean_domain(changeset, field \\ :domain) do
```
aerosol, line 198: This is unnecessary remnant, I'll remove

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -867,4 +872,53 @@ defmodule PlausibleWeb.SiteController do
         |> redirect(to: Routes.site_path(conn, :settings_general, site.domain))
     end
   end
+
+  def change_domain(conn, _params) do
+    if Plausible.v2?() do
+      site = conn.assigns[:site]
+
+      changeset = Plausible.Site.update_changeset(site)
```
vinibrsl, line 880: Nitpick, feel free to ignore:

```suggestion
      changeset = Plausible.Site.update_changeset(conn.assigns.site)
```

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -867,4 +872,53 @@ defmodule PlausibleWeb.SiteController do
...
+    else
+      render_error(conn, 404)
+    end
+  end
+
+  def change_domain_submit(conn, %{"site" => %{"domain" => new_domain}}) do
+    if Plausible.v2?() do
+      site = conn.assigns[:site]
+
+      case Plausible.Site.Domain.change(site, new_domain) do
```
vinibrsl, line 895:
```suggestion
      case Plausible.Site.Domain.change(conn.assigns.site, new_domain) do
```

priv/repo/migrations/20230328062644_allow_domain_change.exs
```diff
@@ -0,0 +1,49 @@
...
+    alter table(:sites) do
+      add(:domain_changed_from, :string, null: true)
+      add(:domain_changed_at, :naive_datetime, null: true)
+    end
+
+    create(unique_index(:sites, :domain_changed_from))
+    create(index(:sites, :domain_changed_at))
+
+    execute("""
+    CREATE OR REPLACE FUNCTION check_domain() RETURNS TRIGGER AS $$
```
vinibrsl, line 14: Would you mind documenting this database trigger behavior as well? I think the Plausible.Site.Domain moduledoc might be a good referencing point.

aerosol: Sure thing, good call!

aerosol: https://github.com/plausible/analytics/pull/2803/commits/622769b26196b6b1a579accb3fde4ae86bbfb1d6

vinibrsl (approved):

aerosol:
I've added a note to CHANGELOG via https://github.com/plausible/analytics/pull/2803/commits/e6a56b4f17e5fb0c40ee722033485f82e9a2eb96

test/plausible/site/cache_test.exs
```diff
@@ -106,6 +115,46 @@ defmodule Plausible.Site.CacheTest do
...
+
+      assert item_by_domain1 == item_by_domain2
+      assert item_by_domain1.domain == domain2
+      assert item_by_domain1.domain_changed_from == domain1
+
+      # domain_changed_from gets no longer tracked
+
+      {:ok, _} = Site.Domain.expire_change_transitions(-1)
+
+      # full refresh removes the stale entry
```
ukutaht, line 149: Can the small 30-second refresh also handle stale `domain_changed_from` fields?

aerosol: Nope, just as it doesn't detect deleted sites.

If `domain_changed_from` turns from something to `nil` back again, we have no trace of that old domain anymore. Which normally shouldn't be a problem because of the transition period - we'll only stop accepting traffic to the old domain a bit later than 72h.

ukutaht: Sounds good

aerosol:
@rphlo sure, we could add that. Do you mind describing your use case?

aerosol:
@rphlo Thanks for sharing your use case. I've added the API endpoint. Please note however that this feature requires a complex database migration to work, it is still in experimental phase and we don't have this documented for self-hosters yet.

And you are correct, in this implementation it's impossible to undo domain change without setting a temporary domain first, unless the 72h transition period ends.

aerosol:
With [67317b2](https://github.com/plausible/analytics/pull/2803/commits/67317b21a80249a774877d71c6c40862af51e119) repeated changes within the same site are allowed, no need for a temporary rename anymore.

PR #2814, opened 2023-04-03
Escape pipes in goal names

ukutaht:
### Changes

With the recent addition of the `member` filter type using pipes, goals that contain pipes don't always work. Demo of a case that's broken on master but fixed with this PR:

https://user-images.githubusercontent.com/3731516/229509412-3a240cf0-0949-4c06-8c93-a85029779a2d.mov

### Tests

### Changelog
- I don't think it's needed since it's a quick fix for something that was released very recently.

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/stats/filters_test.exs
+++ b/test/plausible/stats/filters_test.exs
@@ -76,12 +76,12 @@ defmodule Plausible.Stats.FiltersTest do
 
   describe "escaping pipe character" do
     test "in simple is filter" do
-      %{"goal" => "Foo \\| Bar"}
+      %{"goal" => ~S(Foo \| Bar)}
       |> assert_parsed(%{"event:goal" => {:is, {:event, "Foo | Bar"}}})
     end
 
     test "in member filter" do
-      %{"page" => "/|\\|"}
+      %{"page" => ~S(/|\|)}
       |> assert_parsed(%{"event:page" => {:member, ["/", "|"]}})
     end
   end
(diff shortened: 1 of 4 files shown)


test/plausible/stats/filters_test.exs
```diff
@@ -74,6 +74,18 @@ defmodule Plausible.Stats.FiltersTest do
     end
   end
 
+  describe "escaping pipe character" do
+    test "in simple is filter" do
+      %{"goal" => "Foo \\| Bar"}
```
vinibrsl, line 79:
```suggestion
      %{"goal" => ~S(Foo \| Bar)}
```

test/plausible/stats/filters_test.exs
```diff
@@ -74,6 +74,18 @@ defmodule Plausible.Stats.FiltersTest do
...
+      |> assert_parsed(%{"event:goal" => {:is, {:event, "Foo | Bar"}}})
+    end
+
+    test "in member filter" do
+      %{"page" => "/|\\|"}
```
vinibrsl, line 84:
```suggestion
      %{"page" => ~S(/|\|)}
```

vinibrsl (approved):
I left two minor suggestions up to you to include. Using the `~S` sigil makes it clearer to read strings that would need escaping when using `""`. For example:

```
iex(9)> ~S(Foo \| Bar) == "Foo \\| Bar"
true
```

ukutaht:
Thanks @vinibrsl! I made the changes in https://github.com/plausible/analytics/pull/2814/commits/0903a3ef7a6ebe72e57fbcb1be04496c328caceb

aerosol (approved):

PR #2822, opened 2023-04-05
Match day of the week feature for comparisons

vinibrsl:
### Changes

This pull request adds support for the match day of the week feature for comparisons.

When this option is enabled, the comparison query is shifted to start on the same day of the week as the source query, rather than on the exact same date. For example, if the source query starts on Sunday, January 1st, 2023 and the `year_over_year` comparison query is configured to match day of the week, it will be shifted to start on Sunday, January 2nd, 2022 instead of Saturday, January 1st, 2022.

Preview:

[output.webm](https://user-images.githubusercontent.com/5093045/231193472-c39b29a1-375f-4711-8b49-354ed90a5495.webm)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/api.js
+++ b/assets/js/dashboard/api.js
@@ -50,6 +50,7 @@ export function serializeQuery(query, extraQuery=[]) {
     queryObj.comparison = query.comparison
     queryObj.compare_from = query.compare_from ? formatISO(query.compare_from) : undefined
     queryObj.compare_to = query.compare_to ? formatISO(query.compare_to) : undefined
+    queryObj.match_day_of_week = query.match_day_of_week
   }
 
   Object.assign(queryObj, ...extraQuery)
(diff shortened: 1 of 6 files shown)


vinibrsl:
> I think 'match day per week' should be enabled by default, rather than opt-in as it seems to be. It's based on my assumption that most users prefer to match day of the week most of the time. Maybe it's wrong.

When I was working on this feature with @RobertJoonas, we had a discussion about enabling by default and I was worried it might be confusing for some users. For instance, when comparing yearly stats year over year, the default setting might show Jan 1 compared to Jan 2, which could be a bit confusing at first. Any thoughts on this, @metmarkosaric?

> How does 'match day of week' interact with custom comparison periods? I feel like if someone enters a specific date range they would not like for these dates to be changed by our comparison logic.

Good call! If you select a custom comparison period, the 'match day of the week' option will be hidden. The back-end also ignores this option with custom periods.

metmarkosaric:
i think that at least for "year over year" we should not match by day by default. most tools don't match by day so it might confuse people. if we enable it by default, it would be better to somehow display "matching by day" in the ui (but i don't know if there's space anywhere). it would then also be good to add something like "match by date" in the comparison drop down so it becomes clearer that you have two options (match by day or match by date)

vinibrsl:
I've applied your suggestions! This is now ready for review :)

ukutaht (approved):

PR #2828, opened 2023-04-06
Associate goals with sites, not domains

aerosol:
### Changes

With #2803, we've overlooked the fact that goals were being created with `domain` instead of site's foreign key. This PR addresses that. Requires a staged postgres migration partially introduced in #2833 first - only then I'll rebase this PR and re-enable goals creation.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -6,9 +6,12 @@ defimpl Jason.Encoder, for: Plausible.Goal do
         value.page_path -> :page
       end
 
+    domain = value.site.domain
+
     value
     |> Map.put(:goal_type, goal_type)
-    |> Map.take([:id, :domain, :goal_type, :event_name, :page_path])
+    |> Map.take([:id, :goal_type, :event_name, :page_path])
+    |> Map.put(:domain, domain)
     |> Jason.Encode.map(opts)
   end
 end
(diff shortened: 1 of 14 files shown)


lib/plausible/goal/schema.ex
```diff
@@ -8,7 +8,7 @@ defimpl Jason.Encoder, for: Plausible.Goal do
 
     value
     |> Map.put(:goal_type, goal_type)
-    |> Map.take([:id, :domain, :goal_type, :event_name, :page_path])
+    |> Map.take([:id, :site_id, :goal_type, :event_name, :page_path])
```
aerosol, line 11: this is probably a breaking change, I'll see if I can find a test for it

ukutaht: > @ukutaht how important do you think that is?

I don't know if anyone is using it but would like to stay on the safe side.

> There's a lot of indirection necessary now to keep the domain returned from the API

How about:
1.  Adding a `belongs_to :site, Plausible.Site` to goal schema.
2.  Make sure we call `Repo.preload(goal, :site)` before encoding to JSON.
3. In the encode function we can add `goal.site.domain` to the final result

aerosol: Yeah, that makes sense, I'll go with inner joins to prevent extra queries

aerosol: Done, force pushed with properly named commits

ukutaht (approved):

PR #2840, opened 2023-04-11
Add webhook alert from Checkly -> Instatus

ukutaht:
### Changes

Adds webhook to notify Instatus when a Checkly health check fails.


cnkk (approved):
lgtm!

The terraform plan output is correct. The unrelated change looks like an internal change on their side. This shouldn't change the current behaviour.

ukutaht:
@cnkk the alert channel wasn't actually hooked up to anything. I've tried attaching it to just one of the checks here: https://github.com/plausible/analytics/pull/2840/commits/e7e28ee74c02cef3d50677fb5d071283967e615f

aerosol:
@ukutaht @cnkk doing some minor cleanup with the PRs, just checking if you guys still want this in?

cnkk:
@aerosol we want to switch to instatus.com in the mid-term. I think there are not many steps left.
So, this PR is still relevant.

PR #2841, opened 2023-04-11
Prop filter modal

RobertJoonas:
### Changes

This PR adds a new filter modal component that allows filtering by custom pageview (and custom event) properties. The feature will be enabled with the already existing `custom_dimension_filter` feature flag.

### Tests

### Changelog

> Should be added once the feature flag is enabled for everyone

### Documentation

> Should be updated once the feature flag is enabled for everyone

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -169,8 +169,13 @@ export default function PlausibleCombobox(props) {
     'hidden': props.singleOption && props.values.length === 1
   })
 
+  const containerClass = classNames('relative w-full', {
+    [props.className]: !!props.className,
+    'opacity-20 cursor-default pointer-events-none': props.isDisabled
+  })
+
   return (
-    <div onKeyDown={onKeyDown} ref={containerRef} className={`relative w-full ${props.className || ''}`}>
+    <div onKeyDown={onKeyDown} ref={containerRef} className={containerClass}>
       <div onClick={toggleOpen} className={classNames('pl-2 pr-8 py-1 w-full dark:bg-gray-900 dark:text-gray-300 rounded-md shadow-sm border border-gray-300 dark:border-gray-700 focus-within:border-indigo-500 focus-within:ring-1 focus-within:ring-indigo-500', {'border-indigo-500 ring-1 ring-indigo-500': isOpen, '': !isOpen})}>
         { props.values.map((value) => {
             return (
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -159,7 +159,7 @@ export default function PlausibleCombobox(props) {
   const noMatchesFound = !loading && visibleOptions.length === 0
 
   return (
-    <div onKeyDown={onKeyDown} ref={containerRef} className="relative ml-2 w-full">
+    <div onKeyDown={onKeyDown} ref={containerRef} className={`relative w-full ${props.className || ''}`}>
       <div onClick={toggleOpen} className={classNames('pl-2 pr-8 py-1 w-full dark:bg-gray-900 dark:text-gray-300 rounded-md shadow-sm border border-gray-300 dark:border-gray-700 focus-within:border-indigo-500 focus-within:ring-1 focus-within:ring-indigo-500', {'border-indigo-500 ring-1 ring-indigo-500': isOpen, '': !isOpen})}>
         { props.values.map((value) => {
             return (
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -42,6 +42,7 @@ export default function PlausibleCombobox(props) {
   const [loading, setLoading] = useState(false)
   const [isOpen, setOpen] = useState(false);
   const [input, setInput] = useState('');
+  const [initialRender, setInitialRender] = useState(true)
   const [highlightedIndex, setHighlightedIndex] = useState(0);
   const searchRef = useRef(null);
   const containerRef = useRef(null);
@@ -74,7 +75,6 @@ export default function PlausibleCombobox(props) {
     if (e.key === 'Escape') {
       if (!isOpen || loading) return null
       setOpen(false)
-      searchRef.current?.focus()
       e.preventDefault()
     }
     if (e.key === 'ArrowDown') {
@@ -119,7 +119,6 @@ export default function PlausibleCombobox(props) {
   function toggleOpen() {
     if (!isOpen) {
       fetchOptions(input)
-      searchRef.current.focus()
     } else {
       setInput('')
       setOpen(false)
@@ -129,17 +128,20 @@ export default function PlausibleCombobox(props) {
   function selectOption(option) {
     if (isDisabled(option)) return
 
-    props.onSelect([...props.values, option])
+    if (props.singleOption) {
+      props.onSelect([option])
+    } else {
+      props.onSelect([...props.values, option])
+    }
+
     setOpen(false)
     setInput('')
-    searchRef.current.focus()
   }
 
   function removeOption(option, e) {
     e.stopPropagation()
     const newValues = props.values.filter((val) => val.value !== option.value)
     props.onSelect(newValues)
-    searchRef.current.focus()
     setOpen(false)
   }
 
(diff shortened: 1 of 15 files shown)


assets/js/dashboard/components/combobox.js
```diff
@@ -150,13 +152,22 @@ export default function PlausibleCombobox(props) {
     setOpen(false)
   })
 
+  useEffect(() => {
+    if (!initialRender) { searchRef.current.focus() }
+    setInitialRender(false)
+  }, [isOpen, props.values])
+
```
RobertJoonas, line 159: I used `useEffect` here because:

1) we want to hide the search-box element when the `prop_key` (the single option) is already selected, and ...
2) we cannot call `searchRef.current.focus()` on a hidden element.

`useEffect` will make sure we're calling the `focus()` function only after the component has been re-rendered without `display: hidden`

ukutaht: What actually happens if the `useEffect` is not used?

I tried locally and didn't see any problems with commenting out the `useEffect` and adding the `searchRef.current?.focus()` calls back to where they were. Maybe I didn't test properly?

assets/js/dashboard/stats/modals/prop-filter-modal.js
```diff
@@ -0,0 +1,147 @@
...
+import { parseQuery } from '../../query'
+import * as api from '../../api'
+import { apiPath, siteBasePath } from '../../util/url'
+import { toFilterQuery, parsePrefix } from '../../util/filters';
+
+function PropFilterModal(props) {
+  const query = parseQuery(props.location.search, props.site)
+  const [formState, setFormState] = useState(getFormState())
+
+  function getFormState() {
```
ukutaht, line 16: Can this be extracted into a separate function outside of `PropFIlterModal`? It doesn't look like it accesses any state or props.

assets/js/dashboard/stats/modals/prop-filter-modal.js
```diff
@@ -0,0 +1,147 @@
...
+    const rawValue = query.filters['props']
+    if (rawValue) {
+      const [[propKey, propVal]] = Object.entries(rawValue)
+      const {type, values} = parsePrefix(propVal)
```
ukutaht, line 20: This can be simplified somewhat: https://github.com/plausible/analytics/commit/785c16be6175ea620dc159494e31b6adfbf41035

With this change we don't need to `export parsePrefix` in `util/filters.js` either.

assets/js/dashboard/stats/modals/prop-filter-modal.js
```diff
@@ -0,0 +1,147 @@
...
+
+    return {
+      prop_key: null,
+      prop_value: { type: FILTER_TYPES.is, clauses: [] }
+    }
+  }
+
+  function fetchOptions(filter) {
+    return (input) => {
+      if (filter === 'prop_key') {
```
ukutaht, line 37: Since the two different `fetchOptions` have completely different implementations that share nothing, I think it's best to extra two functions:

```
function fetchPropKeyOptions() {
   // ...
}

function fetchPropValueOptions() {
   // ...
}
```

ukutaht (changes requested):
I've finished looking at the frontend changes. I will take a look at the backend separately.

Some comments inline. I also a couple generic comments:

1. The `PropFilterModal` is missing support for submitting the form with the `Enter` key
2. The text in `PropFilterModal` can overflow when props are long:
Screenshot 2023-04-12 at 12 05 10

lib/plausible/stats/props.ex
```diff
@@ -30,34 +31,30 @@ defmodule Plausible.Stats.Props do
...
-      nil ->
-        ClickhouseRepo.all(
-          from e in base_event_query(site, query),
-            inner_lateral_join: meta in fragment("meta"),
-            select: {e.name, meta.key},
-            distinct: true
+      _ ->
+        from(e in base_event_query(site, query),
+          inner_lateral_join: meta in fragment("meta"),
+          select: {e.name, meta.key},
```
ukutaht, line 47: Now that people can send pageview props I think we should also be able to show a prop breakdown for pageview goals. At the moment this `e.name` seems to prevent that because pageview goals don't work via custom event names.

Maybe this could be rewritten so that the prop breakdown is only fetched if there is a single `is` or `is_not` goal filter. And instead of grouping by `e.name` we could use the goal filter to fill the key of the returned map like `%{"Visit /page" => ["prop_1", "prop_2"]}`

IDK if this makes sense?

In addition, I don't really get why we return a map of `Goal name -> prop list` here. The breakdown is only ever shown for a single goal so it should be fine to just return a single prop list as well. If we don't have to return the goal name as the map key, it would simplify this query/function.

assets/js/dashboard/components/combobox.js
RobertJoonas: The problem I was trying to solve was the searchbar becoming focused after the selected `prop_key` has been removed. But you are right, at the moment, it’s not doing anything reasonable. It’s currently focusing on the `prop_value` searchbar, instead of the `prop_key` one which is not how I intended this to work.

This is the new solution that works as intended for focusing the `prop_key` input field: https://github.com/plausible/analytics/pull/2841/commits/a39f5973fbcb6bc6235156847f27aeefe44d4454.

RobertJoonas:
> 1. The `PropFilterModal` is missing support for submitting the form with the `Enter` key

True. Done in https://github.com/plausible/analytics/pull/2841/commits/e467cc0f6aaaef518d4714c3f9cf963eec96e992.

> 2. The text in `PropFilterModal` can overflow when props are long:

Made it work with `display: grid` in the end. Seems to work nicely now. This will break words into multiple lines if necessary. https://github.com/plausible/analytics/pull/2841/commits/0e6c461099f3eb3255753813970f98ea282ded02

image

assets/js/dashboard/stats/modals/prop-filter-modal.js
RobertJoonas: Good point! Done here: https://github.com/plausible/analytics/pull/2841/commits/a9b9d807c8ea2afb296860b30b787c9ff0f53259

assets/js/dashboard/stats/modals/prop-filter-modal.js
RobertJoonas: Yup. https://github.com/plausible/analytics/pull/2841/commits/a9b9d807c8ea2afb296860b30b787c9ff0f53259

assets/js/dashboard/stats/modals/prop-filter-modal.js
RobertJoonas: Agreed. https://github.com/plausible/analytics/pull/2841/commits/02137702b50c30b8bdcf4c2427401a85dd9aa2d8

lib/plausible/stats/props.ex
RobertJoonas: Makes sense!

I did the following in https://github.com/plausible/analytics/pull/2841/commits/bce455d357997c34d06123295243f7f6d58720b1:

* Separated the function that queries all properties for all custom event goals in one query (feature added by contributor)
* Changed the return type of the props function and made the function names more clear in general. The function used by our dashboard does not need the response to be a map of `goal_name -> prop_list` indeed. So it's now a list
* This new way, we can easily fetch props for a pageview goal too (the glob goal filter is applied in `base_event_query` already - we just select the `meta.key`s)

assets/js/dashboard/components/combobox.js
```diff
@@ -42,6 +42,7 @@ export default function PlausibleCombobox(props) {
   const [loading, setLoading] = useState(false)
   const [isOpen, setOpen] = useState(false);
   const [input, setInput] = useState('');
+  const [searchBoxHidden, setSearchBoxHidden] = useState(!!props.singleOption && props.values.length === 1)
```
ukutaht, line 45: What's the purpose of keeping this in `state`? As opposed to hiding/showing the input based on `!!props.singleOption && props.values.length === 1`?

lib/plausible/stats/custom_props.ex
```diff
@@ -0,0 +1,56 @@
...
+  properties for each event name.
+  """
+  def props_for_all_event_names(site, query) do
+    from(e in base_event_query(site, query),
+      inner_lateral_join: meta in fragment("meta"),
+      select: {e.name, meta.key},
+      distinct: true
+    )
+    |> ClickhouseRepo.all()
+    |> group_by_goal_name()
```
ukutaht, line 21: Would `|> Enum.into(%{})` work here as well?

ukutaht (approved):
A couple comments but it looks good overall.

lib/plausible/stats/custom_props.ex
RobertJoonas: Yeah, that works if we group by event name directly in the db query: https://github.com/plausible/analytics/pull/2841/commits/7a67b64cf899c956992499a31ab2c44d6f96caf1

PR #2848, opened 2023-04-13
Fix date picker shifting day bug

vinibrsl:
This commit fixes a bug where picking a specific date in custom range takes you to the stats of the day before.

This was caused because Flatpickr returns a Date instance, and we were trying to parse this Date instance into day.js using the `utc` function, assuming it was a ISO-8601 string.

This commit fixes it by casting Date into dayjs using the `dayjs` function instead of `utc`.

### DIFF
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -7,6 +7,7 @@ import classNames from 'classnames'
 import * as storage from './util/storage'
 import Flatpickr from 'react-flatpickr'
 import { formatISO, parseUTCDate, formatDateRange } from './util/date.js'
+import dayjs from 'dayjs'
 
 const COMPARISON_MODES = {
   'off': 'Disable comparison',
@@ -145,7 +146,11 @@ const ComparisonInput = function({ site, query, history }) {
     static: true,
     onClose: ([from, to], _dateStr, _instance) => {
       setUiMode("menu")
-      if (from && to) updateMode("custom", formatISO(parseUTCDate(from)), formatISO(parseUTCDate(to)))
+
+      if (from && to) {
+        [from, to] = [dayjs(from), dayjs(to)]
+        updateMode("custom", formatISO(from), formatISO(to))
+      }
     }
   }
 
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -6,7 +6,7 @@ import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import classNames from 'classnames'
 import * as storage from './util/storage'
 import Flatpickr from 'react-flatpickr'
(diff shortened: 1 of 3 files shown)


assets/js/dashboard/datepicker.js
```diff
@@ -389,13 +374,13 @@ function DatePicker({query, site, history}) {
...
               maxDate: 'today',
-              minDate: parseUTCDate(site.statsBegin),
+              minDate: site.statsBegin,
               showMonths: 1,
               static: true,
               animate: true}}
             ref={calendar}
             className="invisible"
-            onChange={setCustomDate}
+            onClose={setCustomDate}
```
aerosol, line 383: what's the reason behind changing the event here?

aerosol (approved):

assets/js/dashboard/comparison-input.js
```diff
@@ -6,7 +6,8 @@ import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import classNames from 'classnames'
 import * as storage from './util/storage'
 import Flatpickr from 'react-flatpickr'
-import { formatISO, parseUTCDate, formatDateRange } from './util/date.js'
+import { formatISO, formatDateRange } from './util/date.js'
+import dayjs from 'dayjs'
```
ukutaht, line 10: I think it's best to hide the `dayjs` usage and always access it through `./util/date`. In `util/date` we use it in UTC mode and I think it would be quite confusing to not use it in utc mode elsewhere.

ukutaht (changes requested):

assets/js/dashboard/comparison-input.js
vinibrsl: I agree we should not call `dayjs` directly, and I moved it to `./util/date.js` in 41aa32176b11e1681c8720f1c23a46e5af62343f.

The UTC mode is a tricky one, because in the context of the date picker, the selected dates should be interpreted in the local timezone, and that's why I'm not using `dayjs.utc()`. When using that, the dates get sometimes shifted back 1 day. For now, I decided to move those direct calls to `dayjs` to `./util/date.js` in a function named `parseNaiveDate`.

assets/js/dashboard/datepicker.js
vinibrsl: Good question! I was getting multiple `onChange` events when I was developing the date picker for comparisons. As I changed the custom date callback function to somewhat similar what we have for comparisons, I also changed the event to match what we use there. `onClose` gets called only when the modal is closed, and sends the two dates. `onChange` is called whenever a date changes in the date picker.

assets/js/dashboard/comparison-input.js
ukutaht: Thanks, makes sense.

PR #2865, opened 2023-04-21
Make self-hosted data migration easier

ruslandoga:
### Changes

This PR makes self-hosted migration a bit easier.

- Defaulting to v2 allows users to skip setting `V2_MIGRATION_DONE` env var. It also forces them to migrate since until they do, dashboard wouldn't show pre-v2-startup data.
- Allowing `N` prompt answer defaults and custom prompt messages provides more info to the users when they are dropping `events_v2` and `sessions_v2` tables. If they are running migrations some time after starting the new Plausible version, they'd have some data in those tables (due to the change above) and we are telling them exactly how much and defaulting to N for the answer. I'm keeping the ability to drop v2 tables in case someone botches the migration on first attempt and wants to try again (in which case they do want to drop v2 tables).

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/data_migration/numeric_ids.ex
```diff
@@ -90,8 +89,28 @@ defmodule Plausible.DataMigration.NumericIDs do
...
 
-    {:ok, _} = run_sql_fn.("drop-events-v2", cluster?: cluster?)
-    {:ok, _} = run_sql_fn.("drop-sessions-v2", cluster?: cluster?)
+    drop_v2_extra_opts = fn table ->
+      case @repo.query("select count(*) from {$0:Identifier}", [table]) do
+        {:ok, %{rows: [[count]]}} when count > 0 ->
+          [
+            prompt_message: "The table contains #{count} rows. Execute?",
+            prompt_default_choice: :no
+          ]
```
ruslandoga, line 98: Screenshot 2023-04-21 at 10 47 45

ruslandoga:
I also made [a PR that removes v1](https://github.com/ruslandoga/analytics/pull/90) but thought that's it's too much and possibly incomplete or worse, contains errors.

priv/data_migrations/NumericIDs/sql/insert-into-tmp-events-v2.sql.eex
```diff
@@ -10,7 +10,7 @@ INSERT INTO tmp_events_v2 (
   transferred_from
 )
 SELECT
-  (SELECT site_id from domains_lookup WHERE domain = domain LIMIT 1),
```
ruslandoga, line 13: Subqueries in ClickHouse don't seem to have access to parent queries so this subselect didn't work as expected: it always returned the first site id.

config/.env.dev
```diff
@@ -17,4 +17,3 @@ GOOGLE_CLIENT_SECRET=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
 
 PROMEX_DISABLED=false
 SITE_DEFAULT_INGEST_THRESHOLD=1000000
-V2_MIGRATION_DONE=1
```
aerosol, line 20: Maybe then we can also remove the additional CI test run that's running for V1 - if you look at the GitHub workflow for PRs. I'm not entirely sure if that's not too early, but probably not.

lib/plausible/data_migration.ex
```diff
@@ -12,20 +12,43 @@ defmodule Plausible.DataMigration do
       @dir dir
       @repo repo
 
-      def run_sql_confirm(name, assigns \\ []) do
-        query = unwrap_with_io(name, assigns)
+      def run_sql_confirm(name, options \\ []) do
```
aerosol, line 15: :raised_hands:

priv/data_migrations/NumericIDs/sql/insert-into-tmp-sessions-v2.sql.eex
```diff
@@ -6,12 +6,13 @@ operating_system_version, subdivision1_code, subdivision2_code, city_geoname_id,
...
-  (SELECT site_id from domains_lookup WHERE domain = domain LIMIT 1),
+  site_id,
   session_id, sign, user_id, hostname, timestamp, start, is_bounce, entry_page, 
   exit_page, pageviews, events, duration, referrer, referrer_source, country_code, 
   screen_size, operating_system, browser, utm_medium, utm_source, utm_campaign, browser_version, 
   operating_system_version, subdivision1_code, subdivision2_code, city_geoname_id, utm_content, 
   utm_term, transferred_from, entry_meta.key, entry_meta.value
 FROM
   sessions
+INNER JOIN domains_lookup ON domains_lookup.domain = sessions.domain
```
aerosol, line 17: Great, that should work for smaller datasets!

aerosol (approved):

priv/data_migrations/NumericIDs/sql/insert-into-tmp-sessions-v2.sql.eex
ruslandoga: Ah, I was wondering why you chose a different method. Is there a performance cost to a join like this?

aerosol: I think there is a notion of joins being discouraged if you can avoid them. I didn't want to risk it in the initial implementation.

priv/data_migrations/NumericIDs/sql/insert-into-tmp-events-v2.sql.eex
aerosol: Oh I see. I freaked out a little, but then I remembered we used a dict lookup in the first implementation.

config/.env.dev
aerosol: Great!

priv/data_migrations/NumericIDs/sql/insert-into-tmp-events-v2.sql.eex
ruslandoga: I ran the dict version and it worked great :) The only problem was setting the dict up...

PR #2871, opened 2023-04-25
Prevent overwriting with_imported when comparing

### LINKED ISSUE #2870: %-comparison is way off - not considering imported GA data?
### Discussed in https://github.com/plausible/analytics/discussions/2867

<div type='discussions-op-text'>

<sup>Originally posted by **Swiss-Mac-User** April 23, 2023</sup>
The comparison vs previous period of my site are way off, e.g. when using „Current month“ view. Is this a bug in correlation with GA data import, where also native Plausible data has been tracked ?

I started tracking with Plausible last month (March) and it recorded „natively“ 550 pageviews. All the other data was imported from Google Analytics to keep the historical data.

It seems to me, that the data comparison vs last period (increase / decrease) does not take the imported data into account…

* Current month: 2.9k pageviews
* Last month: 4.1k pageviews
* Comparison: should be „-34%“
* _Instead it shows as_: „+450%“? (That’s the 550 pageviews natively recorded by Plausible in the last period)

IMG_2158
IMG_2159
</div>

vinibrsl:
This commit fixes a bug where comparisons would not work with imported data. This is because comparisons copies the source query and modifies the dates only, where it should reevaluate if there is imported data for the new comparison date range.

Closes #2870

### DIFF
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -53,13 +53,24 @@ defmodule Plausible.Stats.Comparisons do
       it will be shifted to start on Sunday, January 2nd, 2022 instead of
       January 1st. Defaults to false.
 
+    * `:include_imported?` - includes imported data in the generated comparison
+      query if there is imported data. Defaults to true.
+
   """
   def compare(%Plausible.Site{} = site, %Stats.Query{} = source_query, mode, opts \\ []) do
-    if valid_mode?(source_query, mode) do
-      opts = Keyword.put_new(opts, :now, Timex.now(site.timezone))
-      do_compare(source_query, mode, opts)
+    opts =
+      opts
+      |> Keyword.put_new(:now, Timex.now(site.timezone))
+      |> Keyword.put_new(:match_day_of_week?, false)
+      |> Keyword.put_new(:include_imported?, true)
+
+    with true <- valid_mode?(source_query, mode),
+         {:ok, comparison_query} <- do_compare(source_query, mode, opts),
+         comparison_query <- maybe_include_imported(comparison_query, site, opts) do
+      {:ok, comparison_query}
     else
-      {:error, :not_supported}
+      false -> {:error, :not_supported}
+      {:error, cause} -> {:error, cause}
     end
   end
 
@@ -113,7 +124,7 @@ defmodule Plausible.Stats.Comparisons do
   end
 
   defp maybe_match_day_of_week(comparison_query, source_query, opts) do
-    if Keyword.get(opts, :match_day_of_week?, false) do
+    if Keyword.fetch!(opts, :match_day_of_week?) do
       day_to_match = Date.day_of_week(source_query.date_range.first)
 
       new_first =
@@ -156,6 +167,11 @@ defmodule Plausible.Stats.Comparisons do
     days_to_subtract = if days_to_subtract > 0, do: days_to_subtract, else: days_to_subtract + 7
 
     Date.add(date, -days_to_subtract)
+  end
+
+  defp maybe_include_imported(query, site, opts) do
+    include_imported? = Stats.Query.include_imported?(query, site, opts[:include_imported?])
+    %Stats.Query{query | include_imported: include_imported?}
   end
 
   @spec valid_mode?(Stats.Query.t(), mode()) :: boolean()
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -64,14 +64,10 @@ defmodule Plausible.Stats.Comparisons do
       |> Keyword.put_new(:match_day_of_week?, false)
       |> Keyword.put_new(:include_imported?, true)
 
-    with true <- valid_mode?(source_query, mode),
+    with :ok <- validate_mode(source_query, mode),
          {:ok, comparison_query} <- do_compare(source_query, mode, opts),
-         comparison_query <- maybe_include_imported(comparison_query, site, opts) do
-      {:ok, comparison_query}
-    else
-      false -> {:error, :not_supported}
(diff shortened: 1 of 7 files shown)


test/support/test_utils.ex
```diff
@@ -43,10 +43,10 @@ defmodule Plausible.TestUtils do
     {:ok, site: site}
   end
 
-  def add_imported_data(%{site: site}) do
+  def add_imported_data(%{site: site}, end_at \\ Timex.today()) do
```
aerosol, line 46: This is unused in your implementation I think? I think import's end date was important in reproducing the bug, let me check

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -829,5 +829,60 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
       assert length(labels) == length(comparison_labels)
       assert "__blank__" == List.last(labels)
     end
+
+    test "compares imported data and native data together", %{conn: conn, site: site} do
```
aerosol, line 833: we'll need similar for top-stats too

vinibrsl: Oops! On it!

aerosol: porting those tests to `master` doesn't make them fail

lib/plausible/stats/comparisons.ex
```diff
@@ -53,13 +53,24 @@ defmodule Plausible.Stats.Comparisons do
...
-    if valid_mode?(source_query, mode) do
-      opts = Keyword.put_new(opts, :now, Timex.now(site.timezone))
-      do_compare(source_query, mode, opts)
+    opts =
+      opts
+      |> Keyword.put_new(:now, Timex.now(site.timezone))
+      |> Keyword.put_new(:match_day_of_week?, false)
+      |> Keyword.put_new(:include_imported?, true)
+
+    with true <- valid_mode?(source_query, mode),
```
aerosol, line 67: maybe make this return `:ok | {:error, :not_supported}` so that only a single else clause is then handled?

priv/repo/seeds.exs
```diff
@@ -100,3 +106,52 @@ Enum.flat_map(-720..0, fn day_index ->
...
+          visitors: Enum.random(1..20),
+          pageviews: Enum.random(1..20),
+          exits: Enum.random(1..20),
+          time_on_page: Enum.random(1..100)
+        )
+      ]
+    end)
+  end)
+
+Plausible.TestUtils.populate_stats(site, imported_stats_up_to_previous_month)
```
aerosol, line 155: My takeaway from this is, we should have a plain interface for calling the import pipeline for test purposes, it's currently scattered between controllers and the oban job and it wasn't easy to figure out what function composition was needed in order to create successful import state. Maybe one for a boring day.

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -829,5 +829,60 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+        get(
+          conn,
+          "/api/stats/#{site.domain}/main-graph?period=year&date=2021-01-01&with_imported=false&comparison=year_over_year&interval=month"
+        )
+
+      assert %{
+               "plot" => plot,
+               "comparison_plot" => comparison_plot,
+               "imported_source" => "Google Analytics"
+             } = json_response(conn, 200)
```
aerosol, line 882: we'll have to assert the `with_imported` return value here - otherwise the G icon is inactive when current period has no imported data. I took a stab at it with https://github.com/plausible/analytics/blob/fix-comparison-with-imported-stats-in-another-period/lib/plausible_web/controllers/api/stats_controller.ex#L211-L212

lib/plausible/stats/comparisons.ex
```diff
@@ -158,6 +169,11 @@ defmodule Plausible.Stats.Comparisons do
     Date.add(date, -days_to_subtract)
   end
 
+  defp maybe_include_imported(query, site, opts) do
+    include_imported? = Stats.Query.include_imported?(query, site, opts[:include_imported?])
```
aerosol, line 173: 👍

aerosol (changes requested):
I think there's still something wrong with it, freshly seeded it shows no chart:

image

http://localhost:8000/dummy.site?comparison=off&period=month

edit:

only GA stats are visible

image

http://localhost:8000/dummy.site?comparison=off&period=month&with_imported=true&date=2023-03-25

vinibrsl:
> I think there's still something wrong with it, freshly seeded it shows no chart:
> 
> image
> 
> http://localhost:8000/dummy.site?comparison=off&period=month
> 
> edit:
> 
> only GA stats are visible
> 
> image
> 
> http://localhost:8000/dummy.site?comparison=off&period=month&with_imported=true&date=2023-03-25

I'm a bit confused by this too, but I'm getting this same behavior on master. Do you get it too?

vinibrsl:
> I think there's still something wrong with it, freshly seeded it shows no chart:
> 
> image
> 
> http://localhost:8000/dummy.site?comparison=off&period=month
> 
> edit:
> 
> only GA stats are visible
> 
> image
> 
> http://localhost:8000/dummy.site?comparison=off&period=month&with_imported=true&date=2023-03-25

I changed the seeds in 44e2efb7ba3dfc71999ee6ce0083ecc618236d89 to make sure native data is not overlapping with imported data. My guess is that this was causing this inconsistent state. Let me know if that solved it for your dataset, please :)

aerosol (approved):

PR #2873, opened 2023-04-25
Prepare Comparisons for release

vinibrsl:
This pull request removes the comparisons feature flag, making it visible to all users. It also removes percentages from top stats in default view, leaving those for the comparison view only.


aerosol (approved):
:ship: it

aerosol (approved):

PR #2879, opened 2023-04-26
fix qemu arm64 builds

ruslandoga:
### Changes

This PR adds `ERL_FLAGS=+JPperf true` to public image builds to stop it from segfaulting in qemu when building for arm64. More info for why it works: https://github.com/erlang/otp/pull/6340 (once v26 is released, we can switch to `+JMsingle bool`)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/.github/workflows/build-public-images.yml
+++ b/.github/workflows/build-public-images.yml
@@ -43,6 +43,7 @@ jobs:
           cache-to: type=gha,mode=max
           build-args: |
             BUILD_METADATA=${{ steps.meta.outputs.json }}
+            ERL_FLAGS=+JPperf true
 
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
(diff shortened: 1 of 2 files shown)


ruslandoga:
I've tagged this commit `v2.0.0-rc.0` to check if the build is fixed.

cnkk:
> Is it possible to pin to multi-arch image? Seems not.
Correct, this won't work. Images built from the same Dockerfile and base image but with different architectures will have a different sha256 hashsum.


> But maybe we can provide sha in build args like ALPINE_SHA_arm64=... and ALPIEN_SHA_amd64=c0d488a8... and then in the Dockerfile

The problem with moving the sha256 hash to a variable is, that we'll loose the dependabot auto updates for the image.
IMO you can continue with removing the fixed hash and we will try to re-add this in the future somehow (at least for production images).

cnkk (approved):

aerosol (approved):

PR #2893, opened 2023-05-02
Fix conversions top stats bug when comparing

vinibrsl:
This commit fixes a bug where unique conversions and the unique visitors of the previous period in comparison view were always the same.

Fixes https://github.com/plausible/analytics/discussions/532#discussioncomment-5760187

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -328,35 +328,34 @@ defmodule PlausibleWeb.Api.StatsController do
   end
 
   defp fetch_top_stats(site, %Query{filters: %{"event:goal" => _goal}} = query, comparison_query) do
-    total_q = Query.remove_event_filters(query, [:goal, :props])
-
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -328,36 +328,35 @@ defmodule PlausibleWeb.Api.StatsController do
...
+    } = Stats.aggregate(site, totals_query, [:visitors])
 
     %{
       visitors: %{value: converted_visitors},
       events: %{value: completions}
     } = Stats.aggregate(site, query, [:visitors, :events])
 
+    {prev_unique_visitors, prev_converted_visitors, prev_completions} =
+      if comparison_query do
+        totals_comparison_query = Query.remove_event_filters(comparison_query, [:goal, :props])
```
vinibrsl, line 344: This was missing from the current implementation when building top stats comparisons with a goal filter.

ukutaht (approved):

PR #2894, opened 2023-05-03
Postgres migration to add `allowed_event_props` field to `site`

RobertJoonas:
### Changes

This PR only adds a new migration file that adds a field needed to stop showing garbage properties on the dashboard.

For all sites that have this field with the default value of `[]`, we will show all properties that exist in Clickhouse. However, if at least one element exists in that array, we show only the props that are included in that array.

**EDIT: I think in the end it makes sense to keep the default value for this field as `null`.**

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20230503094245_add_event_prop_allowlist_to_site.exs
+++ b/priv/repo/migrations/20230503094245_add_event_prop_allowlist_to_site.exs
@@ -3,7 +3,7 @@ defmodule Plausible.Repo.Migrations.AddEventPropAllowlistToSite do
 
   def change do
     alter table("sites") do
-      add :allowed_event_props, {:array, :string}, default: []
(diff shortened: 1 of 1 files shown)


ukutaht (approved):

RobertJoonas:
@ukutaht what do you think about leaving this column nullable instead. As I understand (after having discussed this with @vinibrsl in our last call), running this migration would lock the `sites` table for some time, because all the ~155k records would have to be updated (setting `allowed_custom_props` to `[]`).

I also think it's clearer that all props should be allowed if this value is `nil`. And if for whatever reason we'd want to block all custom props for a site, this way we could.  https://github.com/plausible/analytics/pull/2894/commits/48403ee00b0533f2bbf90894a94fc0833f96e1ec

aerosol (approved):

PR #2904, opened 2023-05-05
Limit ingestion payload size

vinibrsl:
### Changes

This pull request adds constraints on user input values during ingestion. The purpose of this PR is to filter out unrealistic or invalid inputs. More comments about each limit inline. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -47,6 +47,7 @@ defmodule Plausible.Ingestion.Request do
         |> put_hostname()
         |> put_user_agent(conn)
         |> put_request_params(request_body)
+        |> put_props(request_body)
         |> put_pathname()
         |> put_query_params()
         |> map_domains(request_body)
@@ -90,8 +91,7 @@ defmodule Plausible.Ingestion.Request do
       changeset,
       event_name: request_body["n"] || request_body["name"],
       referrer: request_body["r"] || request_body["referrer"],
-      hash_mode: request_body["h"] || request_body["hashMode"],
-      props: parse_props(request_body)
+      hash_mode: request_body["h"] || request_body["hashMode"]
     )
   end
 
@@ -153,42 +153,56 @@ defmodule Plausible.Ingestion.Request do
     Changeset.put_change(changeset, :hostname, sanitize_hostname(host))
   end
 
-  defp parse_props(%{} = request_body) do
-    raw_props =
-      request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"]
+  defp put_props(changeset, %{} = request_body) do
+    props =
+      (request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"])
+      |> decode_props_or_fallback()
+      |> Enum.reject(fn {_k, v} -> is_nil(v) || is_list(v) || is_map(v) || v == "" end)
+      |> Map.new()
 
-    case decode_raw_props(raw_props) do
-      {:ok, parsed_json} ->
-        parsed_json
-        |> Enum.filter(&valid_prop_value?/1)
-        |> Map.new()
+    changeset
+    |> Changeset.put_change(:props, props)
+    |> validate_props()
+  end
 
-      _error ->
-        %{}
+  defp decode_props_or_fallback(raw) do
+    with raw when is_binary(raw) <- raw,
+         {:ok, %{} = decoded} <- Jason.decode(raw) do
+      decoded
+    else
+      already_a_map when is_map(already_a_map) -> already_a_map
+      {:ok, _list_or_other} -> %{}
+      {:error, _decode_error} -> %{}
+      _any -> %{}
     end
   end
 
-  defp decode_raw_props(props) when is_map(props), do: {:ok, props}
+  @max_props 50
+  @max_prop_key_length 300
+  @max_prop_value_length 2000
+  defp validate_props(changeset) do
+    props = Changeset.get_field(changeset, :props)
 
-  defp decode_raw_props(raw_json) when is_binary(raw_json) do
-    case Jason.decode(raw_json) do
-      {:ok, parsed_props} when is_map(parsed_props) ->
-        {:ok, parsed_props}
+    props_too_large? =
+      Enum.any?(props, fn
+        {key, _value} when byte_size(key) > @max_prop_key_length -> true
+        {_key, value} when byte_size(value) > @max_prop_value_length -> true
+        _prop -> false
+      end)
 
-      _ ->
-        :not_a_map
-    end
-  end
+    cond do
+      Enum.count(props) > @max_props ->
+        Changeset.add_error(changeset, :props, "should not have more than #{@max_props} items")
 
-  defp decode_raw_props(_), do: :bad_format
+      props_too_large? ->
+        Changeset.add_error(
+          changeset,
+          :props,
+          "keys should have at most #{@max_prop_key_length} chars and values #{@max_prop_value_length} chars"
+        )
 
-  defp valid_prop_value?({key, value}) do
-    case {key, value} do
-      {_key, ""} -> false
-      {_key, nil} -> false
-      {_key, value} when is_list(value) -> false
-      {_key, value} when is_map(value) -> false
-      {_key, _value} -> true
+      true ->
+        changeset
     end
   end
 
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -58,6 +58,7 @@ defmodule Plausible.Ingestion.Request do
         ])
         |> Changeset.validate_length(:pathname, max: 2000)
         |> Changeset.validate_length(:referrer, max: 2000)
+        |> Changeset.validate_length(:event_name, max: 300)
         |> Changeset.apply_action(nil)
 
       {:error, :invalid_json} ->
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -12,7 +12,7 @@ defmodule Plausible.Ingestion.Request do
     field :remote_ip, :string
     field :user_agent, :string
     field :event_name, :string
-    field :uri, :string
+    field :uri, :map
     field :hostname, :string
     field :referrer, :string
     field :domains, {:array, :string}
@@ -129,15 +129,17 @@ defmodule Plausible.Ingestion.Request do
     end
   end
 
+  @max_url_size 2_000
   @disallowed_schemes ~w(data)
   defp put_uri(changeset, %{} = request_body) do
     with url when is_binary(url) <- request_body["u"] || request_body["url"],
+         url when byte_size(url) <= @max_url_size <- url,
          %URI{} = uri when uri.scheme not in @disallowed_schemes <- URI.parse(url) do
       Changeset.put_change(changeset, :uri, uri)
     else
       nil -> Changeset.add_error(changeset, :url, "is required")
       %URI{} -> Changeset.add_error(changeset, :url, "scheme is not allowed")
-      _ -> Changeset.add_error(changeset, :url, "must be a string")
+      _ -> Changeset.add_error(changeset, :url, "must be a valid url")
     end
   end
 
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -57,6 +57,7 @@ defmodule Plausible.Ingestion.Request do
           :timestamp
         ])
         |> Changeset.validate_length(:pathname, max: 2000)
+        |> Changeset.validate_length(:referrer, max: 2000)
         |> Changeset.apply_action(nil)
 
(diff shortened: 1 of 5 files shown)


lib/plausible/ingestion/request.ex
```diff
@@ -57,6 +58,8 @@ defmodule Plausible.Ingestion.Request do
           :timestamp
         ])
         |> Changeset.validate_length(:pathname, max: 2000)
+        |> Changeset.validate_length(:referrer, max: 2000)
```
vinibrsl, line 61: 2,000 chars is the reasonable limit most software imply for URLs (https://stackoverflow.com/a/417184)

lib/plausible/ingestion/request.ex
```diff
@@ -127,15 +129,17 @@ defmodule Plausible.Ingestion.Request do
     end
   end
 
+  @max_url_size 2_000
   @disallowed_schemes ~w(data)
   defp put_uri(changeset, %{} = request_body) do
     with url when is_binary(url) <- request_body["u"] || request_body["url"],
+         url when byte_size(url) <= @max_url_size <- url,
```
vinibrsl, line 136: I set a limit on how long the `url` parameter can be before parsing into `%URI{}`. This also means that any extra stuff tagged onto the URL (like UTM tags) can't be longer than 2,000 characters altogether.

lib/plausible/ingestion/request.ex
```diff
@@ -149,42 +153,56 @@ defmodule Plausible.Ingestion.Request do
     Changeset.put_change(changeset, :hostname, sanitize_hostname(host))
   end
 
-  defp parse_props(%{} = request_body) do
-    raw_props =
-      request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"]
+  defp put_props(changeset, %{} = request_body) do
```
vinibrsl, line 156: I decided to move things around here to make clear what's strict validation that will trigger an error, and what's not. When we fail to parse the props JSON string, we don't drop the event, but fallback to an empty map. But when the props have more than 50 items, we return HTTP 400 and drop the event.

lib/plausible/ingestion/request.ex
```diff
@@ -149,42 +153,56 @@ defmodule Plausible.Ingestion.Request do
...
-  defp valid_prop_value?({key, value}) do
-    case {key, value} do
-      {_key, ""} -> false
-      {_key, nil} -> false
-      {_key, value} when is_list(value) -> false
-      {_key, value} when is_map(value) -> false
-      {_key, _value} -> true
+  @max_props 50
+  @max_prop_key_length 300
+  @max_prop_value_length 2000
```
vinibrsl, line 182: Those numbers were arbitrary, and I would like feedback on whether we should be less strict.

For reference, I chose 300 characters to limit prop key lengths because that's the limit for the event name too. And 2000 characters for prop values so you can fit a URL into it if needed.

lib/plausible/ingestion/request.ex
ukutaht: > But when the props have more than 50 items, we return HTTP 400 and drop the event.

We have some customers sending more than 50 custom props at the moment. We'd need to reach out to them about this new limit before applying it.

lib/plausible/ingestion/request.ex
```diff
@@ -57,6 +58,8 @@ defmodule Plausible.Ingestion.Request do
           :timestamp
         ])
         |> Changeset.validate_length(:pathname, max: 2000)
+        |> Changeset.validate_length(:referrer, max: 2000)
+        |> Changeset.validate_length(:event_name, max: 300)
```
ukutaht, line 62: I think this limit can be much lower. GA puts it at 40 characters only. @aerosol and I seemed to agree that 100 or 128 would be a reasonable limit for event name.

The prod database has some events longer than 128 but they all look like complete garbage so I think they're mostly from broken integrations.

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -35,6 +35,23 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert pageview.pathname == "/"
     end
 
+    test "validates event name length", %{conn: conn, site: site} do
+      params = %{
+        domain: site.domain,
+        name: for(_ <- 1..500, do: "a", into: ""),
```
aerosol, line 41:
```suggestion
        name: String.duplicate("a", 500),
```

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -250,6 +267,23 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert pageview.referrer_source == "Facebook"
     end
 
+    test "limits referrer string size", %{conn: conn, site: site} do
+      params = %{
+        name: "pageview",
+        url: "http://gigride.live/",
+        referrer: "https://facebook.com/#{for _ <- 1..2500, do: "a", into: ""}",
```
aerosol, line 274:
```suggestion
        referrer: "https://facebook.com/" <> String.duplicate("a", 2500),
```

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -414,6 +448,28 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert pageview.utm_campaign == "video_story"
     end
 
+    test "validates url length with query params", %{conn: conn, site: site} do
+      long_string = for _ <- 1..500, do: "a", into: ""
```
aerosol, line 452:
```suggestion
      long_string = String.duplicate("a", 500)
```

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -565,6 +621,58 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert Map.get(event, :"meta.value") == ["true", "12"]
     end
 
+    test "validates props key length", %{conn: conn, site: site} do
+      long_string = for _ <- 1..500, do: "a", into: ""
```
aerosol, line 625:
```suggestion
      long_string = String.duplicate("a", 500)
```

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -565,6 +621,58 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
...
+
+      assert %{
+               "errors" => %{
+                 "props" => ["keys should have at most 300 chars and values 2000 chars"]
+               }
+             } = json_response(conn, 400)
+    end
+
+    test "validates props value length", %{conn: conn, site: site} do
+      long_string = for _ <- 1..2500, do: "a", into: ""
```
aerosol, line 644:
```suggestion
      long_string = String.duplicate("a", 2500)
```

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -35,6 +35,23 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert pageview.pathname == "/"
     end
 
+    test "validates event name length", %{conn: conn, site: site} do
```
aerosol, line 38: 🤔 It feels like those tests should be in `Plausible.Ingestion.RequestTest`, we don't need to go to end to end anymore thanks to the Request abstraction. WYT?

lib/plausible/ingestion/request.ex
```diff
@@ -149,42 +153,56 @@ defmodule Plausible.Ingestion.Request do
...
+      props_too_large? ->
+        Changeset.add_error(
+          changeset,
+          :props,
+          "keys should have at most #{@max_prop_key_length} chars and values #{@max_prop_value_length} chars"
+        )
+
+      true ->
+        changeset
     end
```
aerosol, line 206: I think we can make it more compact by leveraging pattern matching and guards over `cond` and more efficient by only checking props length when the map size is valid. What do you think about:

```suggestion
defp validate_props(changeset) do
    case Changeset.get_field(changeset, :props) do
      props when map_size(props) > @max_props ->
        Changeset.add_error(changeset, :props, "should not have more than #{@max_props} items")

      props ->
        Enum.reduce_while(props, changeset, fn
          {key, value}, changeset
          when byte_size(key) > @max_prop_key_length or
                 byte_size(value) > @max_prop_value_length ->
            {:halt,
             Changeset.add_error(
               changeset,
               :props,
               "keys should have at most #{@max_prop_key_length} chars and values #{@max_prop_value_length} chars"
             )}

          _, changeset ->
            {:cont, changeset}
        end)
    end
```

lib/plausible/ingestion/request.ex
aerosol: Real max is over 10k 😱.

Does the referrer is guaranteed to be a valid URI? I think not and that makes it hard to sanitize it properly leading to potentially confusing data display.

I'm seeing entries like `r.search.yahoo.com/cbclk2/longEncodedGarbage` -- not sure if maybe dropping the referrer alone in such cases would be better than trying to massage it.

lib/plausible/ingestion/request.ex
aerosol: Yeah I think around 120 is sane for custom goals - we should probably validate the form input too then.

aerosol (commented):
Added some comments inline and minor code nitpicks

lib/plausible/ingestion/request.ex
aerosol: We also need to keep in mind that we're counting bytes here, not graphemes, so the "at most ... chars" error message might be confusing?

```
iex(1)> s = "我"
"我"
iex(2)> String.length(s)
1
iex(3)> byte_size(s)
3
```

test/plausible_web/controllers/api/external_controller_test.exs
vinibrsl: I agree! Added in 352bf019cdf46c426e7cfcad86a00ebe4754e7b4. Do you think we should remove them from ExternalControllerTest or can we keep in both modules as they are testing different interfaces?

lib/plausible/ingestion/request.ex
vinibrsl: Yes, it makes the pathname validation obsolete. Removed in 4b84e846aee4fd62aec85462058d97bdcbe4219b.

lib/plausible/ingestion/request.ex
vinibrsl: Good call! Changed in ea508c1cd80af9673f077f2b49d30028dadead5a

lib/plausible/ingestion/request.ex
vinibrsl: Good one! I replaced chars with bytes in the error message. Changed in 26ac00d8.

test/plausible_web/controllers/api/external_controller_test.exs
aerosol: I think it's redundant, and the higher we get, the less tests we should maintain there. If the contract of "changeset includes error message on invalid input" is tested, we have no reason to believe this will be any different than the behaviours we already test at the controller level.

image

lib/plausible/ingestion/request.ex
vinibrsl: We've collected 780 events with referrer length > 2000 this year. Most of these events had a referrer length < 2500 though. Do you think it would be reasonable to increase the limit from 2000 to 2500 for both pathnames and referrers?

The main purpose of this pull request is to drop malicious/non-sense input, so increasing the limit to 2500 should be fine. Let me know what you think.

lib/plausible/ingestion/request.ex
vinibrsl: > EDIT: To be clear I think the limit makes sense. GA has a limit of 27 custom dimensions.

So we're fine to set the limit without notice?

test/plausible_web/controllers/api/external_controller_test.exs
vinibrsl: Makes sense. Thank you! Reverted in ca171a18.

lib/plausible/ingestion/request.ex
aerosol: Got it. I think 2500 as a hard cap is generous enough. WYT @ukutaht?

aerosol (approved):

lib/plausible/ingestion/request.ex
vinibrsl: I changed the implementation to truncate the referrer at 2000 chars. 4045ab25b3536fa9321dc7345ab26b1cd979042f.

vinibrsl:
I emailed customers sending more than 30 props to understand their use of custom properties, and to warn our intention of adding this limit. In the meantime, we can release this pull request if you agree. I removed the prop count validation (522c84477eb551a8bee90bc5a17030a01309a4e2) for now, and I'll push to a separate pull request after we get input from customers.

ukutaht (approved):

PR #2907, opened 2023-05-09
Filter out garbage props by `site.allowed_event_props`

RobertJoonas:
### Changes

This PR builds on top of https://github.com/plausible/analytics/pull/2894 that contains the migration that should be run on production **before this PR is deployed**.

This PR changes the behavior of returning custom properties for:

* property name suggestions
image

* goal conversions:
image

The properties returned will now be filtered by the custom allowlist, that can be defined for each site in the CRM. For example, if we were to set the allowlist to `author, logged_in`, the above screenshots would only contain those property names - nothing else.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/admin.ex
+++ b/lib/plausible/site/admin.ex
@@ -41,7 +41,14 @@ defmodule Plausible.SiteAdmin do
       public: nil,
       owner: %{value: &get_owner_email/1},
       other_members: %{value: &get_other_members/1},
-      allowed_event_props: %{value: &Enum.join(&1.allowed_event_props, ", ")},
+      allowed_event_props: %{
+        value: fn site ->
+          case site.allowed_event_props do
+            nil -> ""
+            list -> Enum.join(list, ", ")
+          end
+        end
+      },
       limits: %{
         value: fn site ->
           case site.ingest_rate_limit_threshold do
@@ -74,7 +81,9 @@ defmodule Plausible.SiteAdmin do
   def set_allowed_props_for_site(_conn, [site], params) do
     props_list =
       case params["props"] do
-        "" -> []
+        "" ->
+          nil
+
         props ->
           props
           |> String.trim()
@@ -83,6 +92,7 @@ defmodule Plausible.SiteAdmin do
 
     Plausible.Site.set_allowed_event_props(site, props_list)
     |> Repo.update!()
+
     :ok
   end
 
--- a/lib/plausible/site/admin.ex
+++ b/lib/plausible/site/admin.ex
@@ -41,6 +41,7 @@ defmodule Plausible.SiteAdmin do
       public: nil,
       owner: %{value: &get_owner_email/1},
       other_members: %{value: &get_other_members/1},
+      allowed_event_props: %{value: &Enum.join(&1.allowed_event_props, ", ")},
       limits: %{
         value: fn site ->
           case site.ingest_rate_limit_threshold do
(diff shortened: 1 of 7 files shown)


lib/plausible/stats/filter_suggestions.ex
```diff
@@ -145,6 +145,7 @@ defmodule Plausible.Stats.FilterSuggestions do
       )
 
     ClickhouseRepo.all(q)
+    |> Plausible.Stats.CustomProps.drop_garbage_props(site)
```
aerosol, line 148: Would it be possible to exclude garbage at the query level?

lib/plausible/site/admin.ex
```diff
@@ -53,6 +61,45 @@ defmodule Plausible.SiteAdmin do
...
+        ],
+        name: "Allow only these custom properties",
+        action: &set_allowed_props_for_site/3
+      }
+    ]
+  end
+
+  def set_allowed_props_for_site(_conn, [site], params) do
+    props_list =
+      case params["props"] do
```
aerosol, line 83:
```suggestion
      case String.trim(params["props"]) do
```

lib/plausible/site/admin.ex
```diff
@@ -53,6 +61,45 @@ defmodule Plausible.SiteAdmin do
...
+        "" ->
+          nil
+
+        props ->
+          props
+          |> String.trim()
+          |> String.split(~r/\s*,\s*/)
```
aerosol, line 90:
```suggestion
          |> String.split(",", trim: true)
```

lib/plausible/site/admin.ex
RobertJoonas: https://github.com/plausible/analytics/pull/2907/commits/1e00c264b6f8dc920ce111de605a943ee90a054d

lib/plausible/site/admin.ex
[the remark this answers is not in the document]
RobertJoonas: Good one! I made it look nicer as per your suggestions here: https://github.com/plausible/analytics/pull/2907/commits/1e00c264b6f8dc920ce111de605a943ee90a054d.

But left the regex part in there because the `trim: true` will not trim individual list elements.

```
iex(1)> String.split("one,      two", ",", trim: true)    
["one", "      two"]
```

lib/plausible/stats/filter_suggestions.ex
RobertJoonas: @aerosol, your recommendation looks and works good! https://github.com/plausible/analytics/pull/2907/commits/963f8b5eabef3bf634831b35c8d3c0c8a9c6055f

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -331,6 +331,65 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
...
+         %{
+           conn: conn,
+           site: site
+         } do
+      site =
+        site
+        |> Plausible.Site.set_allowed_event_props(["author"])
+        |> Plausible.Repo.update!()
+
+      populate_stats(site, [
```
ukutaht, line 344: This test is missing an event with `author` property. Currently you can replace the implementation with:

```elixir
  def maybe_allowed_props_only(q, allowed_props) when is_list(allowed_props) do
     from [_e, m] in q, where: false
end
```

And the tests will pass, although the implementation is incorrect.

ukutaht (approved):
One comment about a test, but otherwise looks good.

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
RobertJoonas: Good catch! https://github.com/plausible/analytics/pull/2907/commits/1d710d7bd3f37591ee43aca64f05f9aac773cc78

ukutaht (approved):

PR #2868, opened 2023-04-24
Clean up after V2 migration

aerosol:
This PR removes all the leftovers and alternative code branching after v2 migration.

The self-hosted release is being drafted at:

https://github.com/plausible/hosting/issues/68

Refs:
  - https://github.com/plausible/analytics/pull/2865
  - https://github.com/plausible/analytics/pull/2825
  - https://github.com/plausible/analytics/pull/2780

### Changes

Please describe the changes made in the pull request here.

Below you'll find a checklist. For each item on the list, check one option and delete the other.

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #2908, opened 2023-05-09
Drop events/sessions tables after V2 migration

aerosol:
### Changes

This PR adds a housekeeping migration to drop the unused tables ref: #2868 
#2825 #2777  

### Tests

### Changelog

### Documentation

### Dark mode


ruslandoga:
I wonder if it could be done a bit later, after full `v2.0` self-hosted release. Otherwise it might be possible we'd release a version that does the drop automatically for a user who's going from v1.5 to, for example, v2.0.1 or similar.

aerosol:
@ruslandoga are there any blockers for releasing the final v2.0 now? Besides I think we can add an appropriate warning note for anything that comes after v2.0.

ruslandoga:
It seems like there aren't any problems with v2.0.0-rc.0 so it's probably OK to release it as v2.0.0.

PR #2913, opened 2023-05-11
Simplify Phoenix error template

vinibrsl:


lib/plausible_web/templates/layout/error.html.eex
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
```
ukutaht, line 1: AFAICT within this template we're rendering two inner templates:
* lib/plausible_web/templates/error/error.html.eex
* lib/plausible_web/templates/error/server_error.html.eex

The first one also includes the whole HTML layout code. I think that should be adjusted to only contain what's needed in the `<%= @inner_content %>` of this layout.

lib/plausible_web/controllers/helpers.ex
```diff
@@ -6,13 +6,16 @@ defmodule PlausibleWeb.ControllerHelpers do
     conn
     |> put_status(status)
     |> put_view(PlausibleWeb.ErrorView)
-    |> render("#{status}.html", layout: false, message: message)
+    |> render("#{status}.html", message: message, layout: error_layout())
```
vinibrsl, line 9: This module renders errors without the need of exceptions, for example when a controller wants to explicitly render a 404. Phoenix only automatically sets the layout when there is an exception, and for this case I needed to set the layout "manually".

lib/plausible_web/templates/layout/error.html.eex
vinibrsl: You're right! I pushed a change removing the HTML boilerplate as we're using a layout now.

vinibrsl:
I still need to fix one test that's failing.

lib/plausible_web/controllers/helpers.ex
```diff
@@ -6,13 +6,16 @@ defmodule PlausibleWeb.ControllerHelpers do
...
   def render_error(conn, status) do
     conn
     |> put_status(status)
     |> put_view(PlausibleWeb.ErrorView)
-    |> render("#{status}.html", layout: false)
+    |> render("#{status}.html", layout: error_layout())
   end
+
+  defp error_layout,
+    do: Application.get_env(:plausible, PlausibleWeb.Endpoint)[:render_errors][:layout]
```
vinibrsl, line 20: This module renders errors without the need of exceptions, for example when a controller wants to explicitly render a 404. Phoenix automatically sets the layout only when there is an exception, and for this case I needed to set the layout "manually".

ukutaht (approved):

PR #2919, opened 2023-05-11
Support `with_imported=true` in Stats API aggregate

RobertJoonas:
### Changes

This PR fixes a bug where a Stats API query responded with a 500 error message, because the original query was fetching stats without imported data, but the comparison query tried to fetch stats with imported data.

It also adds support for querying imported data from the Stats API aggregate endpoint. Adding the `with_imported` option to other endpoints (breakdown and timeseries is out of scope of this PR).

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/stats/comparisons_test.exs
+++ b/test/plausible/stats/comparisons_test.exs
@@ -205,7 +205,7 @@ defmodule Plausible.Stats.ComparisonsTest do
   describe "include_imported" do
     setup [:create_user, :create_new_site, :add_imported_data]
 
-    test "defaults to source_query.include_imported if not specified by opts", %{site: site} do
+    test "defaults to source_query.include_imported", %{site: site} do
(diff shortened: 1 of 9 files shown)


test/plausible/stats/comparisons_test.exs
```diff
@@ -200,4 +201,16 @@ defmodule Plausible.Stats.ComparisonsTest do
                Comparisons.compare(site, query, "custom", from: "2022-05-30", to: "2022-05-25")
     end
   end
+
+  describe "include_imported" do
+    setup [:create_user, :create_new_site, :add_imported_data]
+
+    test "defaults to source_query.include_imported if not specified by opts", %{site: site} do
```
vinibrsl, line 208: Since we don't support that in opts anymore:

```suggestion
    test "defaults to source_query.include_imported", %{site: site} do
```

vinibrsl (approved):

PR #2931, opened 2023-05-16
Add unique constraint to api keys

ukutaht:
### Changes

Adds unique constraint to `api_keys` on the `[:user_id, :key_hash]` field combination. This protects from duplicate API keys being created. The UI provides appropriate feedback:

Screenshot 2023-05-16 at 16 15 44

### DIFF
--- a/priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
+++ b/priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
@@ -2,6 +2,6 @@ defmodule Plausible.Repo.Migrations.AddUniqueIndexToApiKeys do
   use Ecto.Migration
 
   def change do
-    create unique_index(:api_keys, [:user_id, :key_hash])
+    create unique_index(:api_keys, :key_hash)
   end
 end
--- a/priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
+++ b/priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
@@ -1,0 +1,7 @@
+defmodule Plausible.Repo.Migrations.AddUniqueIndexToApiKeys do
+  use Ecto.Migration
+
+  def change do
+    create unique_index(:api_keys, [:user_id, :key_hash])
+  end
+end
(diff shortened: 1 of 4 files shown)


test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -709,6 +709,49 @@ defmodule PlausibleWeb.AuthControllerTest do
...
+
+    test "cannot create a duplicate API key", %{conn: conn, user: user} do
+      site = insert(:site)
+      insert(:site_membership, site: site, user: user, role: "owner")
+
+      conn =
+        post(conn, "/settings/api-keys", %{
+          "api_key" => %{
+            "user_id" => user.id,
+            "name" => "all your code are belong to us",
```
vinibrsl, line 738: lol

vinibrsl (approved):

priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Plausible.Repo.Migrations.AddUniqueIndexToApiKeys do
+  use Ecto.Migration
+
+  def change do
+    create unique_index(:api_keys, [:user_id, :key_hash])
```
vinibrsl, line 5: We have some dupes on the database that need to be deleted before creating the index:

```sql
select user_id, key_hash from api_keys group by user_id, key_hash having count(*) > 1
```

vinibrsl (changes requested):

priv/repo/migrations/20230516131041_add_unique_index_to_api_keys.exs
ukutaht: Good call, I've changed the index in https://github.com/plausible/analytics/pull/2931/commits/f07dcd84a32d4b5f275bc1af96e1b5fd99aff2e6

Do you think removing duplicates should be part of the migration? Or a manual fix on the prod database?

vinibrsl: I think it would be best to change the database manually. I haven't checked that, but if we have duplicate entries for different user IDs, we'll have to review each case individually.

vinibrsl (approved):

PR #2942, opened 2023-05-17
Escape domain when creating URL for favicon

vangberg:
### Changes

Domains can include slashes, e.g. I have created a site with the "domain" "vangberg.github.io/adventures-in-bioimaging". I am not sure if you are supposed to be able to do that, but it does seem to work.

One thing that does not work, is that the favicon is broken on "My sites" (but not on the stats page for the site itself):

image

The culprit is that the "domain" is not being escaped before being used in the URL for the image tag. This is done in other parts of the codebase, e.g. in the `SiteSwitcher` component:

 https://github.com/plausible/analytics/blob/f29d07a49e91ecfe0aa87de1b4b5c011e26c722a/assets/js/dashboard/site-switcher.js#L11

This pull requests adds URL escaping to the domain in the site index template.

### Tests
- [x] This PR does not require tests (I think?)

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 - Add error message in case a transfer to an invited (but not joined) user is requested plausible/analytics#2651
 - Fix bug with [showing property breakdown with a prop filter](https://github.com/plausible/analytics/issues/1789)
 - Fix bug when combining goal and prop filters plausible/analytics#2654
+- Fix broken favicons when domain includes a slash
 
 ### Changed
 - Treat page filter as entry page filter for `bounce_rate`
--- a/lib/plausible_web/plugs/favicon.ex
+++ b/lib/plausible_web/plugs/favicon.ex
@@ -80,11 +80,11 @@ defmodule PlausibleWeb.Favicon do
 
   """
   def call(conn, favicon_domains: favicon_domains) do
-    case conn.path_info do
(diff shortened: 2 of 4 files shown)


RobertJoonas:
Thanks for the contribution, @vangberg! That's a nice discovery! :)

Your changes would work nicely. However, I see we're already doing some URI decoding [here](https://github.com/plausible/analytics/blob/master/lib/plausible_web/plugs/favicon.ex#L88). Due to an additional forward slash, the pattern is currently not match on line 87, but we could use `conn.request_path` rather than `conn.path_info` and match with strings instead. That would solve the issue as well. I'm thinking:

```elixir
case conn.request_path do
    "/favicon/sources/placeholder" ->
        # ...
    "/favicon/sources/" <> source ->
        # ...
    _ ->
        # ...
end
```

This way we could keep the templates as they are. Also, would be nice to cover this with a test in `favicon_test.exs` :)

Please let me know if you're up for implementing those changes! 

Thanks again!  🙏🏼

vangberg:
> Please let me know if you're up for implementing those changes!

Unfortunately, I am afraid I have exhausted my Elixir abilities already 😄

RobertJoonas:
> Unfortunately, I am afraid I have exhausted my Elixir abilities already

Hehe, no worries! I just committed those changes myself :)

ukutaht (approved):

PR #2946, opened 2023-05-18
collapse clickhouse migrations

ruslandoga:
### Changes

This is an alternative approach to #2945 which completely removes v1 events and sessions migrations. I think it's a bit cleaner in that no v1 tables are created in tests / new setups, and nothing is automatically dropped. The downside is that it breaks continuity, but on the other hand, we can still see the previous migrations in git history.

### Tests

### Changelog

### Documentation

### Dark mode


priv/ingest_repo/structure.sql
```diff
@@ -1,10 +1,12 @@
-CREATE TABLE plausible_events_db.sessions
+CREATE TABLE plausible_events_db.sessions_v2
```
ruslandoga, line 1: I've also updated the dump to make sure I haven't removed too much.

```
$ mix ecto.dump --repo Plausible.IngestRepo
```

ruslandoga:
This might need to be revisited with the v2.1.0 release. If the user updates to v2.1.0 from v1, they would lose their data.

PR #2948, opened 2023-05-19
Revenue tracking: Add currency field to goal creation

vinibrsl:
This commit adds a currency field to the goals form. Goals that have a currency set are now revenue goals, and are cached with sites to later be used during ingestion. This is hidden under a feature flag for now.


config/runtime.exs
```diff
@@ -320,6 +320,9 @@ config :plausible, Plausible.ImportDeletionRepo,
   transport_opts: ch_transport_opts,
   pool_size: 1
 
+config :ex_money,
+  open_exchange_rates_app_id: get_var_from_path_or_env(config_dir, "OPEN_EXCHANGE_RATES_APP_ID")
```
vinibrsl, line 324: The application should work normally without this environment variable, but currency conversion won't work in runtime. Not used now, but I'll make sure to set this before deploying.

lib/plausible/site/cache.ex
vinibrsl, line 1: I'd appreciate your feedback on this file, @aerosol. I added the `sites.revenue_goals` association to the sites cache.

vinibrsl, line 1: I'd appreciate your feedback on this file, @aerosol. I added the sites.revenue_goals association to the sites cache.

priv/repo/migrations/20230417092745_add_monetary_value_to_goals.exs
vinibrsl, line 1: Once approved, I'll extract this migration to another pull request so it's merged and deployed first.

priv/repo/migrations/20230417092745_add_monetary_value_to_goals.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.Repo.Migrations.AddMonetaryValueToGoals do
+  use Ecto.Migration
+
+  def change do
+    alter table(:goals) do
+      add :currency, :string
```
ukutaht, line 6: Does the currency code have a fixed length we could use in this field?

ukutaht (approved):

priv/repo/migrations/20230417092745_add_monetary_value_to_goals.exs
vinibrsl: Yes! Good call. Can we please continue in this pull request? https://github.com/plausible/analytics/pull/2955/files

PR #2949, opened 2023-05-19
Revenue tracking: Add revenue tracking fields to events

vinibrsl:
This commit adds 4 fields to the ClickHouse events_v2 table:

* `revenue_source_amount` and `revenue_source_currency` store revenue in the original currency sent during ingestion

* `revenue_reporting_amount` and `revenue_reporting_currency` store revenue in a common currency to perform calculations, and this currency is defined by the user when setting up the goal

### DIFF
--- a/priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
+++ b/priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
@@ -1,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.AddRevenueToEvents do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
+      add :revenue_reporting_amount, :"Decimal64(4)"
+      add :revenue_reporting_currency, :"LowCardinality(FixedString(3))"
+
+      add :revenue_source_amount, :"Decimal64(4)"
+      add :revenue_source_currency, :"LowCardinality(FixedString(3))"
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.AddRevenueToEvents do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
+      add :revenue_reporting_amount, :"Decimal64(4)"
```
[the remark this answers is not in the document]
ruslandoga: @ukutaht the only downside to decimals I know of is that they are slower than ints (operations on them). Physically they are stored as ints.

priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
```diff
@@ -0,0 +1,13 @@
...
+      add :revenue_reporting_currency, :"LowCardinality(FixedString(3))"
```
ukutaht, line 7: > For small fixed values using of LowCardinality is usually inefficient, because ClickHouse stores a numeric index for each row. As a result:
> 
> * Disk space usage can rise.
> * RAM consumption can be higher, depending on a dictionary size.
> * Some functions can work slower due to extra coding/encoding operations.
> * Merge times in MergeTree-engine tables can grow due to all the reasons described above.
> 

Source: https://clickhouse.com/docs/en/operations/settings/settings#allow_suspicious_low_cardinality_types

I think just `FixedString(3)` would be better here.

ukutaht: Basically the rule seems to be that if the String is less than 8 bytes, LowCardinality will only hurt.

ukutaht (changes requested):

priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
vinibrsl: I didn't know that. Thanks! Changed in 3136e4d5f45200370c584405577ef216abfbc11a

vinibrsl:
Closed in favor of the discussions in #2957. I'll reopen when we decide on the database type.

PR #2957, opened 2023-05-23
Revenue tracking: Ingestion and breakdown queries

vinibrsl:
This pull request adds revenue data to ingestion, saving `revenue_source_amount`, `revenue_source_currency`, `revenue_reporting_amount` and `revenue_reporting_currency` to events.

Goals are stored in sites cache, and if the client sends revenue data and there's a matching goal, revenue data is stored. If needed, the amount is also converted into the goal's preferred currency using the Open Exchange Rates API.

More comments inline.


lib/plausible/ingestion/request.ex
```diff
@@ -209,6 +199,52 @@ defmodule Plausible.Ingestion.Request do
...
+      decoded
+    else
+      already_a_map when is_map(already_a_map) -> already_a_map
+      {:ok, _list_or_other} -> %{}
+      {:error, _decode_error} -> %{}
+      _any -> %{}
+    end
+  end
+
+  defp parse_decimal(value) do
```
vinibrsl, line 239: Revenue fields can take strings `"312.12"`, floats `312.12` and integers `312`.

lib/plausible/ingestion/request.ex
```diff
@@ -209,6 +199,52 @@ defmodule Plausible.Ingestion.Request do
...
+    with true <- currency in @valid_currencies,
+         {%Decimal{} = amount, _rest} <- parse_decimal(amount),
+         %Money{} = amount <- Money.new(currency, amount) do
+      Changeset.put_change(changeset, :revenue_source, amount)
+    else
+      false ->
+        Changeset.add_error(changeset, :revenue_source, "currency is not supported or invalid")
+
+      _parse_error ->
+        Changeset.add_error(changeset, :revenue_source, "must be a valid number")
```
vinibrsl, line 223: When the request passes a revenue object, validations are strict, meaning that events are dropped with invalid numbers or currencies.

lib/plausible/ingestion/event.ex
```diff
@@ -206,6 +207,28 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_props(%__MODULE__{} = event), do: event
 
+  defp put_revenue(%__MODULE__{request: %{revenue_source: %Money{} = revenue_source}} = event) do
+    site = Plausible.Site.Cache.get(event.domain)
```
ukutaht, line 211: Could we look up the site once instead of doing another lookup here?

I'm thinking that the `GateKeeper.check/1` function could return `{:allow, site}` instead of `{:allow, site_id}`. Then the `process_unless_stopped/1` could send the site as an optional argument to the pipeline steps.

I'm leaning this way because I feel like over time we'll be using the site config a lot during ingestion and it's better to always have it ready to go instead of having to do lookups each time.

lib/plausible/ingestion/request.ex
ukutaht: I'm not sure about this. Are there any problems with accepting the event with revenue fields as `nil`?

lib/plausible/ingestion/request.ex
```diff
@@ -209,6 +199,52 @@ defmodule Plausible.Ingestion.Request do
...
+    end
+  end
+
+  defp decode_json_or_fallback(raw) do
+    with raw when is_binary(raw) <- raw,
+         {:ok, %{} = decoded} <- Jason.decode(raw) do
+      decoded
+    else
+      already_a_map when is_map(already_a_map) -> already_a_map
+      {:ok, _list_or_other} -> %{}
```
ukutaht, line 233: I suppose the two tuple cases are redundant here since the last clause will always match.

lib/plausible/ingestion/request.ex
```diff
@@ -209,6 +199,52 @@ defmodule Plausible.Ingestion.Request do
...
+      {:error, _decode_error} -> %{}
+      _any -> %{}
+    end
+  end
+
+  defp parse_decimal(value) do
+    case value do
+      value when is_binary(value) -> Decimal.parse(value)
+      value when is_float(value) -> {Decimal.from_float(value), nil}
```
ukutaht, line 242: What's the purpose of return a tuple here? The second value seems to be always `nil` and ignored.

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -671,6 +671,91 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
...
+        domain: site.domain,
+        revenue: %{amount: 10.2, currency: "USD"}
+      }
+
+      insert(:goal, event_name: "Payment", currency: "BRL", site: site)
+
+      assert %{status: 202} = post(conn, "/api/event", params)
+      assert %{revenue_reporting_amount: amount} = get_event(site)
+
+      assert Decimal.equal?(Decimal.new("7.14"), amount)
```
ukutaht, line 687: Where does the exchange rate for this come from? Does the test need to change when exchange rates move in the future?

vinibrsl: Exchange rates in tests are fixed and we define it here: https://github.com/plausible/analytics/blob/master/test/support/exchange_rate_mock.ex

lib/plausible/ingestion/request.ex
vinibrsl: This is to mimic the same return type of `Decimal.parse/1`, used in the first clause of this `case`. This is to maintain the signature of the `Decimal.parse/1` function return type. It returns a "rest" of what it could not parse, e.g.

```elixir
Decimal.parse("12.4x2")
#=> {Decimal.new("12.4"), "x2"}
```

I did it like this to avoid nesting a second `case` here, and I decided to use the same return type as `Decimal.parse/1`.

lib/plausible/ingestion/request.ex
vinibrsl: The reasoning behind dropping invalid events is to avoid bringing averages down when doing calculations with revenue data. AFAIK decimals can't be `NULL` on ClickHouse, and can't be `Nullable()` either, so we can't differentiate real zeroes from invalid/null data.

lib/plausible/ingestion/request.ex
vinibrsl: You're right! Changed in e4d0182b

lib/plausible/ingestion/event.ex
vinibrsl: I really like your suggestion! Can I work on this separately in a new PR after we merge the current one? This function only runs when there's revenue data in the event request, so it's ok to merge this and update GateKeeper right after.

lib/plausible/money_without_currency.ex
```diff
@@ -0,0 +1,90 @@
...
+    :error
+  end
+
+  @doc """
+  Dumps a %Decimal{} to the database as integer with a fixed precision of
+  #{@exponent} digits.
+  """
+  def dump(value)
+
+  def dump(%Decimal{} = decimal) do
```
vinibrsl, line 71: @ukutaht you had suggested me to use `Money.to_integer_exp/2` here. However, at this stage, we don't have the `%Money{}` struct available. In this particular case, we can only use `%Decimal{}` because Ecto expects the `cast/1` function to return the expected form of `Plausible.MoneyWithoutCurrency`, which is `%Decimal{}`.

If I were to return `%Money{}`, currency codes would not be available at times since it's stored in a separate field. As a con, I had to implement `to_integer_exp/2` myself for `%Decimal{}`.

lib/plausible/ingestion/request.ex
ruslandoga: Sorry to butt in, but decimals can be nullable in ClickHouse. I don't know about the goals of this feature, but having nullable decimals could potentially simplify the queries compared the current approach that seems to require a `where amount < @max_unit64` and exponent fiddling `amount > ^to_uint64(money)` to be usable.

vinibrsl: Thanks for the heads up, @ruslandoga! Not butting in at all, your feedback is always welcome. I discussed about this with Uku yesterday, and there are some reasons why we wanted to use UInt64 instead of Decimal64 or Nullable(Decimal64):

- It seems like ClickHouse does not recommend using Nullable due to the extra storage it needs
- Decimal64 does not have an unsigned variant, and storing negative amounts is out of the scope of the feature
- Using an integer gives us the flexibility of using a dynamic exponent if we ever need that for non-fiat currencies, for example
- To solve the problem of differentiating missing values, we decided to use max_uint64, and use a custom Ecto type to cast that accordingly

This commit changes the type from Decimal64 to UInt64 https://github.com/plausible/analytics/pull/2957/commits/3c583c8b0df3fa2f81176474de6695e1ef2501fc. Let me know what you think.

ruslandoga: Ah, great points!

lib/plausible/stats/base.ex
```diff
@@ -277,6 +277,34 @@ defmodule Plausible.Stats.Base do
...
 
+  def select_event_metrics(q, [:total_revenue | rest]) do
+    from(e in q,
+      select_merge: %{
+        total_revenue:
+          fragment(
+            "sumIf(?, not empty(?)) * any(_sample_factor)",
+            e.revenue_reporting_amount,
+            e.revenue_reporting_currency
+          )
```
vinibrsl, line 288: `Decimal64(3)`'s default value is `0.000`, meaning that we need a way to differentiate missing values from actual zeroes. I used the currency to check revenue is present in events for sums and averages, e.g. `sumIf(revenue_reporting_amount, not empty(revenue_reporting_currency))`

lib/plausible/stats/base.ex
```diff
@@ -277,6 +277,34 @@ defmodule Plausible.Stats.Base do
     |> select_event_metrics(rest)
   end
 
+  def select_event_metrics(q, [:total_revenue | rest]) do
+    from(e in q,
+      select_merge: %{
+        total_revenue:
+          fragment(
```
vinibrsl, line 284: Even though we're working with `Decimal64(3)` on ClickHouse, this query returns an Elixir float. I was expecting a `%Decimal{}` struct. Is that because of the `sumIf` function? @ruslandoga

lib/plausible/ingestion/event.ex
```diff
@@ -206,6 +207,38 @@ defmodule Plausible.Ingestion.Event do
...
+      matching_goal.currency == revenue_source.currency ->
+        update_attrs(event, %{
+          revenue_source_amount: Money.to_decimal(revenue_source),
+          revenue_source_currency: to_string(revenue_source.currency),
+          revenue_reporting_amount: Money.to_decimal(revenue_source),
+          revenue_reporting_currency: to_string(revenue_source.currency)
+        })
+
+      matching_goal.currency != revenue_source.currency ->
+        converted = Money.to_currency!(revenue_source, matching_goal.currency)
```
vinibrsl, line 229: This is where currency conversion happens. The source amount is saved for debugging purposes if we ever need it.

lib/plausible/ingestion/event.ex
```diff
@@ -206,6 +207,38 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_props(%__MODULE__{} = event), do: event
 
+  defp put_revenue(%__MODULE__{request: %{revenue_source: %Money{} = revenue_source}} = event) do
+    revenue_goals = Plausible.Site.Cache.get(event.domain).revenue_goals || []
```
vinibrsl, line 211: TODO in a separate pull request: call site cache only once and pass it over to ingestion pipeline functions.

lib/plausible/stats/base.ex
ruslandoga: Not sure, but `sum` doesn't do any type conversions according to

```sql
mac3.local :) select sum(number) as s, toTypeName(s) from (select toDecimal64(number / 3, 3) as number from system.numbers limit 5);

SELECT
    sum(number) AS s,
    toTypeName(s)
FROM
(
    SELECT toDecimal64(number / 3, 3) AS number
    FROM system.numbers
    LIMIT 5
)

Query id: 53e52fa6-0fb5-4628-9cf7-325fbad2657a

┌─────s─┬─toTypeName(sum(number))─┐
│ 3.332 │ Decimal(38, 3)          │
└───────┴─────────────────────────┘

1 row in set. Elapsed: 0.004 sec.
```

priv/ingest_repo/migrations/20230417104025_add_revenue_to_events.exs
vinibrsl, line 1: Once approved, I'll push that in a separate PR for deploying the migration only first.

lib/plausible/stats/base.ex
ruslandoga: I'm sure I'm missing a lot of context here. But I just wanted to note that a built-in nullable type, with all it's downsides, might be about the same in performance as `-If` prefixed operators (`sumIf`, `avgIf`, etc.).

lib/plausible/stats/base.ex
ruslandoga: Some updates:

The query that is generated in test `test/plausible_web/controllers/api/stats_controller/conversions_test.exs:228` is 

```sql
SELECT
  e0."name",
  toUInt64(round(uniq(e0."user_id") * any(_sample_factor))),
  toUInt64(round(count(*) * any(_sample_factor))),
  ifNotFinite(avgIf(e0."revenue_reporting_amount", not empty(e0."revenue_reporting_currency")) * any(_sample_factor), 0.0),
  sumIf(e0."revenue_reporting_amount", not empty(e0."revenue_reporting_currency")) * any(_sample_factor)
FROM "events_v2" AS e0 sample 20000000
WHERE (e0."site_id" = {$0:Int64}) AND
  (
    (e0."timestamp" >= {$1:DateTime}) AND
    (e0."timestamp" < {$2:DateTime})
  ) AND
  (e0."name" IN ({$3:String}))
GROUP BY e0."name"
ORDER BY
  uniq(e0."user_id") DESC,
  e0."name"
LIMIT {$4:Int64}
OFFSET {$5:Int64}
```
I've tried the functions applied to the columns one by one and here's what I'm getting:

```elixir
iex> Ch.query!(conn, "select sum(number) from (select toDecimal64(number, 3) as number from system.numbers limit 5)").rows
[[Decimal.new("10.000")]]

iex> Ch.query!(conn, "select sumIf(number, number > 3) from (select toDecimal64(number, 3) as number from system.numbers limit 5)").rows
[[Decimal.new("4.000")]]

iex> Ch.query!(conn, "select avgIf(number, number > 3) from (select toDecimal64(number, 3) as number from system.numbers limit 5)").rows
[[4.0]]

iex> Ch.query!(conn, "select sumIf(number, number > 3) * any(20000000) from (select toDecimal64(number, 3) as number from system.numbers limit 5)").rows
[[Decimal.new("80000000.000")]]
```

vinibrsl: Ohhhh I feel so dumb now. Thank you! I added a `toDecimal64(..., 3)` to the query and it works perfectly!

vinibrsl:
I rebased the branch and reorganized the commits after the decimal type changes. I believe it's easier to review now. cc/ @ukutaht

ukutaht (approved):
One comment - I don't think we have to do it in the context of this PR but I think `total_revenue` and `average_revenue` should become first-class metrics in their own right. At the moment they're magically added to the response in `Plausible.Stats.Breakdown` but I think the end state we want to get to is that the caller (either from React or Plausible.Api.StatsController) specifies the metrics to query.

PR #2958, opened 2023-05-23
Allow setting sample_threshold in dashboard API queries

RobertJoonas:
### Changes

This PR does not change behaviour much, only:

* refactors the Query module
* whenever a `sample_threshold=<value>` query param is present in an internal dashboard API query, we will cast that string value into an integer and use that as the sample threshold. As opposed to blowing up like before, when trying to use the `string` type as a sample threshold

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -163,7 +163,13 @@ defmodule Plausible.Stats.Query do
   end
 
   defp put_sample_threshold(query, params) do
-    sample_threshold = Map.get(params, "sample_threshold", @default_sample_threshold)
+    sample_threshold =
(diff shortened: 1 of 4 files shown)


lib/plausible/stats/query.ex
```diff
@@ -13,76 +13,65 @@ defmodule Plausible.Stats.Query do
 
   @type t :: %__MODULE__{}
 
-  def from(site, %{"period" => "realtime"} = params) do
+  def from(site, params) do
```
ukutaht, line 16: Beautiful stuff, thanks for this

ukutaht (approved):

PR #2970, opened 2023-05-24
Allow optional IPv6 for clickhouse repo

rissson:
### Changes

Adds optional support for connect to clickhouse over ipv6. Same as #1661, but for clickhouse.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -282,7 +282,7 @@ maybe_ch_ipv6 =
 ch_transport_opts = [
   keepalive: true,
   show_econnreset: true,
-  inet6: maybe_ch_ipv6,
+  inet6: maybe_ch_ipv6
 ]
 
 config :plausible, Plausible.ClickhouseRepo,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -275,9 +275,14 @@ config :plausible, :google,
   reporting_api_url: "https://analyticsreporting.googleapis.com",
   max_buffer_size: get_int_from_path_or_env(config_dir, "GOOGLE_MAX_BUFFER_SIZE", 10_000)
 
+maybe_ch_ipv6 =
+  get_var_from_path_or_env(config_dir, "ECTO_CH_IPV6", "false")
+  |> String.to_existing_atom()
+
 ch_transport_opts = [
   keepalive: true,
-  show_econnreset: true
+  show_econnreset: true,
+  inet6: maybe_ch_ipv6,
 ]
 
 config :plausible, Plausible.ClickhouseRepo,


rissson:
This ought to replace the ungodly `sed -i 's/ch_transport_opts = \[/ch_transport_opts = [ inet6: true ,/' /app/releases/0.0.1/runtime.exs` I have in my deployment currently...

config/runtime.exs
```diff
@@ -275,9 +275,12 @@ config :plausible, :google,
   reporting_api_url: "https://analyticsreporting.googleapis.com",
   max_buffer_size: get_int_from_path_or_env(config_dir, "GOOGLE_MAX_BUFFER_SIZE", 10_000)
 
+maybe_ch_ipv6 = System.get_env("ECTO_CH_IPV6", false)
+
 ch_transport_opts = [
   keepalive: true,
   show_econnreset: true
+  inet6: maybe_ch_ipv6,
```
ruslandoga, line 283: Even though anything truthy would work: https://github.com/elixir-mint/mint/blob/5bb90a9a9aa4dfb96f5f13b4480142081422088e/lib/mint/core/transport/tcp.ex#L29, I think it'd be a bit cleaner if we ensure `maybe_ch_ipv6` is a boolean. `System.get_env("ECTO_CH_IPV6", false)` returns a string if `ECTO_CH_IPV6` is set.

ruslandoga:
```
ch_inet6? = case get_int_from_path_or_env(config_dir, "ECTO_CH_IPV6") do
  "true" -> true
  "false" -> false
  nil -> false
end

ch_transport_opts = [
   keepalive: true,
   show_econnreset: true,
   inet6: ch_ipv6?
 ]
```

config/runtime.exs
```diff
@@ -275,9 +275,12 @@ config :plausible, :google,
   reporting_api_url: "https://analyticsreporting.googleapis.com",
   max_buffer_size: get_int_from_path_or_env(config_dir, "GOOGLE_MAX_BUFFER_SIZE", 10_000)
 
+maybe_ch_ipv6 = System.get_env("ECTO_CH_IPV6", false)
+
 ch_transport_opts = [
   keepalive: true,
   show_econnreset: true
```
ruslandoga, line 282: missing a comma

config/runtime.exs
```diff
@@ -275,9 +275,12 @@ config :plausible, :google,
   reporting_api_url: "https://analyticsreporting.googleapis.com",
   max_buffer_size: get_int_from_path_or_env(config_dir, "GOOGLE_MAX_BUFFER_SIZE", 10_000)
 
+maybe_ch_ipv6 = System.get_env("ECTO_CH_IPV6", false)
```
ruslandoga, line 278: we should probably use `get_int_from_path_or_env(config_dir, "ECTO_CH_IPV6")` to stay consistent

config/runtime.exs
ukutaht: So far we've used:

```elixir
get_var_from_path_or_env(config_dir, "<VAR>", "false")
  |> String.to_existing_atom()
```

For booleans in config.

A better approach would be to extract a helper function in helpers/config.exs like:

```elixir
  def get_bool_from_path_or_env(config_dir, var_name, default \\ nil) do
   var = get_var_from_path_or_env(config_dir, var_name)

    case var do
      "true" -> true
       "false" -> false
       _ -> default
    end
  end

```

rissson:
I've never actually written in this language before, so which options do you want me to use?

ruslandoga:
I think

```elixir
get_var_from_path_or_env(config_dir, "ECTO_CH_IPV6", "false")
|> String.to_existing_atom()
```

that Uku suggested would be the most consistent. Let's go with that.

rissson:
Got it, thanks for the pointers!

ruslandoga (approved):

rissson:
Ah I see this missed 2.0, shame. Was anything else needed?

ruslandoga:
@rissson `v2.0` hasn't been officially released yet: https://github.com/plausible/analytics/pull/3030

aerosol (approved):

PR #2975, opened 2023-05-25
Fix event name validation

ukutaht:
### Changes

Currently when an event is sent with a `name` field that is not a String, we run into an Ecto error:

```
CaseClauseError anonymous fn/5 in Ecto.Changeset.validate_length/3
```

This PR makes 2 changes:
1. Accept numeric event names (like `404`)2. 
2. Return validation error if event name is not a String or Integer

### DIFF
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -1,3 +1,16 @@
+defmodule Plausible.Ecto.EventName do
+  use Ecto.Type
+  def type, do: :string
+
+  def cast(val) when is_binary(val), do: {:ok, val}
+  def cast(val) when is_integer(val), do: {:ok, Integer.to_string(val)}
+
+  # Everything else is a failure though
+  def cast(_), do: :error
+  def load(val), do: {:ok, val}
+  def dump(val), do: {:ok, val}
+end
+
 defmodule Plausible.Ingestion.Request do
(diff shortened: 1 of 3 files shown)


lib/plausible/ingestion/request.ex
```diff
@@ -13,12 +26,12 @@ defmodule Plausible.Ingestion.Request do
...
     field :remote_ip, :string
     field :user_agent, :string
-    field :event_name, :string
+    field :event_name, Plausible.Ecto.EventName
     field :uri, :map
     field :hostname, :string
     field :referrer, :string
     field :domains, {:array, :string}
-    field :hash_mode, :string
+    field :hash_mode, :integer
```
ukutaht, line 34: This type was wrong, down below we do:

```elixir
if hash_mode == 1 && uri.fragment do
    pathname <> "#" <> URI.decode(uri.fragment)
else
    pathname
end
```

So it must be an integer. It worked before because `Ecto.Changeset.change` does not cast types whereas `Ecto.Changeset.cast` does.

aerosol (approved):
LGTM, nice fix. I think the tests should be in Plausible.Ingestion.RequestTest though. No need to go full end to end here.

ukutaht:
Good call, I've moved the tests 👍

vinibrsl (approved):

PR #2985, opened 2023-05-29
Enable sampling in the Stats API

ukutaht:
We're currently hardcoding `sample_threshold=infinity` in the Stats API. This PR removes the hardcoding so it can be configured, and applies a standard sampling threshold of 20m.


vinibrsl (approved):
Looks good. I'm just curious where the "standard sampling threshold of 20m" you mentioned is defined. Is that a ClickHouse default?

aerosol:
@vinibrsl found it, it comes from https://github.com/plausible/analytics/blob/master/lib/plausible/stats/query.ex#L6

ukutaht:
> @vinibrsl found it, it comes from https://github.com/plausible/analytics/blob/master/lib/plausible/stats/query.ex#L6

Correct :)

PR #2986, opened 2023-05-30
Dockerfile and co updates

ruslandoga:
### Changes

This PR explores possible changes to the current docker build setup.

### Tests

### Changelog

### Documentation

### Dark mode


Dockerfile
```diff
@@ -59,21 +59,20 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
...
-RUN chown -R plausibleuser:plausibleuser /app
 USER plausibleuser
-WORKDIR /app
-ENV LISTEN_IP=0.0.0.0
+
+COPY --from=buildcontainer --chown=plausibleuser:plausibleuser /app/_build/prod/rel/plausible /app
+COPY --chown=plausibleuser:plausibleuser ./rel/docker-entrypoint.sh /entrypoint.sh
+
 ENTRYPOINT ["/entrypoint.sh"]
-EXPOSE 8000
```
ruslandoga, line 78: It's not always true since port is configurable via env.

Dockerfile
```diff
@@ -59,21 +59,20 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
 ARG BUILD_METADATA={}
 ENV BUILD_METADATA=$BUILD_METADATA
 ENV LANG=C.UTF-8
+ENV LISTEN_IP=0.0.0.0
```
ruslandoga, line 62: LISTEN_IP can probably be removed but that would require a change to runtime.exs

Dockerfile
```diff
@@ -59,21 +59,20 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
...
-COPY ./rel/docker-entrypoint.sh /entrypoint.sh
+WORKDIR /app
 
-RUN chmod a+x /entrypoint.sh && \
-  adduser -h /app -u 1000 -s /bin/sh -D plausibleuser
+RUN adduser -u 1000 -s /bin/sh -D plausibleuser && \
+  chown plausibleuser /app
 
-COPY --from=buildcontainer /app/_build/prod/rel/plausible /app
-RUN chown -R plausibleuser:plausibleuser /app
```
ruslandoga, line 73: <details><summary>Copying the release and then changing the owner creates duplicate layers for the release (~90MB)</summary>

```console
$ docker history plausible:master
IMAGE          CREATED       CREATED BY                                      SIZE      COMMENT
9970282ff1ca   2 hours ago   CMD ["run"]                                     0B        buildkit.dockerfile.v0
<missing>      2 hours ago   EXPOSE map[8000/tcp:{}]                         0B        buildkit.dockerfile.v0
<missing>      2 hours ago   ENTRYPOINT ["/entrypoint.sh"]                   0B        buildkit.dockerfile.v0
<missing>      2 hours ago   ENV LISTEN_IP=0.0.0.0                           0B        buildkit.dockerfile.v0
<missing>      2 hours ago   WORKDIR /app                                    0B        buildkit.dockerfile.v0
<missing>      2 hours ago   USER plausibleuser                              0B        buildkit.dockerfile.v0
<missing>      2 hours ago   RUN |1 BUILD_METADATA={} /bin/sh -c chown -R…   79.6MB    buildkit.dockerfile.v0
<missing>      2 hours ago   COPY /app/_build/prod/rel/plausible /app # b…   79.6MB    buildkit.dockerfile.v0
<missing>      2 hours ago   RUN |1 BUILD_METADATA={} /bin/sh -c chmod a+…   4.86kB    buildkit.dockerfile.v0
<missing>      2 hours ago   COPY ./rel/docker-entrypoint.sh /entrypoint.…   169B      buildkit.dockerfile.v0
<missing>      2 hours ago   RUN |1 BUILD_METADATA={} /bin/sh -c apk add …   5.03MB    buildkit.dockerfile.v0
<missing>      2 hours ago   RUN |1 BUILD_METADATA={} /bin/sh -c apk upgr…   361kB     buildkit.dockerfile.v0
<missing>      2 hours ago   ENV LANG=C.UTF-8                                0B        buildkit.dockerfile.v0
<missing>      2 hours ago   ENV BUILD_METADATA={}                           0B        buildkit.dockerfile.v0
<missing>      2 hours ago   ARG BUILD_METADATA={}                           0B        buildkit.dockerfile.v0
<missing>      2 hours ago   LABEL maintainer=plausible.io <hello@plausib…   0B        buildkit.dockerfile.v0
<missing>      2 weeks ago   /bin/sh -c #(nop)  CMD ["/bin/sh"]              0B
<missing>      2 weeks ago   /bin/sh -c #(nop) ADD file:df7fccc3453b6ec14…   7.73MB
```

Dockerfile
```diff
@@ -58,22 +58,21 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
 
 ARG BUILD_METADATA={}
 ENV BUILD_METADATA=$BUILD_METADATA
-ENV LANG=C.UTF-8
+ENV LANG=en_US.UTF-8 LANGUAGE=en_US:en LC_ALL=en_US.UTF-8
```
ruslandoga, line 61: I copied it from ClickHouse and `mix phx.gen.release --docker`, but not sure if it matters to Plausible. I remember there were warnings in Elixir during start-up about locales being setup incorrectly sometimes.

Here's what `mix phx.gen.release --docker` does (note that it uses debian):

```dockerfile
RUN apt-get update -y && apt-get install -y libstdc++6 openssl libncurses5 locales \
  && apt-get clean && rm -f /var/lib/apt/lists/*_*

# Set the locale
RUN sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen && locale-gen

ENV LANG en_US.UTF-8
ENV LANGUAGE en_US:en
ENV LC_ALL en_US.UTF-8
```

config/runtime.exs
```diff
@@ -7,6 +7,7 @@ if config_env() in [:dev, :test] do
 end
 
 config_dir = System.get_env("CONFIG_DIR", "/run/secrets")
+storage_dir = get_var_from_path_or_env(config_dir, "STORAGE_DIR", System.tmp_dir!())
```
ruslandoga, line 10: To make the above work we need to move all writing outside of `/app`, for example to `/tmp`.

config/runtime.exs
```diff
@@ -527,6 +529,7 @@ geo_opts =
       [
         license_key: maxmind_license_key,
         edition: maxmind_edition,
+        database_cache_file: maxmind_database_cache_file,
```
ruslandoga, line 532: Or stop writing completely (unless the user provides the path) like here. If `maxmind_database_cache_file`, `no_cache` option is used in `locus`.

config/runtime.exs
```diff
@@ -582,9 +585,7 @@ else
     traces_exporter: :none
 end
 
-config :tzdata,
-       :data_dir,
-       get_var_from_path_or_env(config_dir, "STORAGE_DIR", Application.app_dir(:tzdata, "priv"))
+config :tzdata, :data_dir, Path.join(storage_dir, "plausible_tzdata_data")
```
ruslandoga, line 588: I've added namespacing here to avoid collisions.

Dockerfile
```diff
@@ -58,22 +58,20 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
...
-RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates
+RUN adduser -u 1000 -s /bin/sh -D plausibleuser
 
-COPY ./rel/docker-entrypoint.sh /entrypoint.sh
+RUN apk upgrade --no-cache && \
+  apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates
 
-RUN chmod a+x /entrypoint.sh && \
-  adduser -h /app -u 1000 -s /bin/sh -D plausibleuser
+COPY --from=buildcontainer --chmod=a+rX /app/_build/prod/rel/plausible /app
```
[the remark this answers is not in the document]
ruslandoga: From https://cookbook.openshift.org/users-and-role-based-access-control/why-do-my-applications-run-as-a-random-user-id.html

> To ensure container images are portable and will work with container deployment platforms which enforce additional separation between applications by overriding the user ID, images and the applications they contain, should be designed to be able to be run as an arbitrary user ID.

Dockerfile
```diff
@@ -58,22 +58,20 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
...
 ENV BUILD_METADATA=$BUILD_METADATA
-ENV LANG=C.UTF-8
+ENV LANG=en_US.UTF-8 LANGUAGE=en_US:en LC_ALL=en_US.UTF-8
+ENV LISTEN_IP=0.0.0.0
 
-RUN apk upgrade --no-cache
+WORKDIR /app
 
-RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates
+RUN adduser -u 1000 -s /bin/sh -D plausibleuser
```
ruslandoga, line 66: - [ ] maybe system user? (like in [imgproxy](https://github.com/imgproxy/imgproxy/blob/master/docker/Dockerfile) and [ClickHouse](https://github.com/ClickHouse/ClickHouse/blob/master/docker/server/Dockerfile.alpine))

```Dockerfile
RUN addgroup -S plausible \
  && adduser -S -u 999 -g plausible plausible

USER 999
```

cnkk: For the sake of accuracy, system user sounds better

config/runtime.exs
cnkk: Sounds good to me.
After implementing this, we could use [tmpfs mounts](https://docs.docker.com/storage/tmpfs/) when running the container to prevent writing data into the container's writable layer.

cnkk:
Thanks! Looks good in general.

For deployment, we should start with deploying to staging and test there first.

config/runtime.exs
ruslandoga: The two dependencies that write to `/app/lib/.../<dependency>/priv` do it for caching reasons:

- [`tzdata`](https://github.com/lau/tzdata#data-directory-and-releases) caches timezone info
- [`locus`](https://github.com/g-andrade/locus/tree/master) caches geolocation databases

So I think it might make sense to actually persist this data across restarts.

Dockerfile
ruslandoga: Reverted in https://github.com/plausible/analytics/pull/2986/commits/58b1ad4f146ed8f2ca706f77c75676cfa30f68b7

Dockerfile
ruslandoga: Reverted in https://github.com/plausible/analytics/pull/2986/commits/58b1ad4f146ed8f2ca706f77c75676cfa30f68b7

Dockerfile
ruslandoga: @cnkk should we have a system user with uid < 1000 then or still >= 10000?

From https://www.redhat.com/sysadmin/user-account-gid-uid

> The theory behind this arbitrary assignment is that anything below 1000 is reserved for system accounts, services, and other special accounts, and regular user UIDs and GIDs stay above 1000.

ruslandoga: I've set it to 999 for now: https://github.com/plausible/analytics/pull/2986/commits/2ea125c3ed7cfca6fcbb806b6a7c8eab767cbe90

```console
$ docker compose exec plausible ash
$ id
uid=999(plausible) gid=65533(nogroup) groups=65533(nogroup)
```

I wonder why the group is not set.

ruslandoga: Seems to be a busybox thing, `-G` is used for groups.

Dockerfile
```diff
@@ -60,18 +60,17 @@ ARG BUILD_METADATA={}
 ENV BUILD_METADATA=$BUILD_METADATA
 ENV LANG=C.UTF-8
 
-RUN apk upgrade --no-cache
+RUN addgroup -S plausible && \
+  adduser -S -H -u 999 -G plausible plausible -g 'Plausible Analytics'
```
ruslandoga, line 64: `plausible:x:999:101:Plausible Analytics:/home/plausible:/sbin/nologin`

ruslandoga:
```
-h DIR		Home directory
-g GECOS	GECOS field
-s SHELL	Login shell
-G GRP		Group
-S		Create a system user
-D		Don't assign a password
-H		Don't create home directory
-u UID		User id
-k SKEL		Skeleton directory (/etc/skel)
```

Dockerfile
ruslandoga: Maybe we can use `777` for all ids, it's still "system" by convention. And it should be lucky.

cnkk: We probably put too much thoughts into it :D
plausible:nogroup should be fine though.

In our use case, the plausible container runs on isolated hosts, and there is only one non-sensitive file mounted into the container.
Selfhosters will most likely use volumes.
All in all, as long as we are running as a non-root user, it should be fine.

cnkk:
> Thanks! Looks good in general.
> 
> For deployment, we should start with deploying to staging and test there first.

@ruslandoga could you update your fork, so we can test this on staging?

config/runtime.exs
```diff
@@ -15,6 +15,7 @@ if config_env() == :small_test do
 end
 
 config_dir = System.get_env("CONFIG_DIR", "/run/secrets")
+storage_dir = get_var_from_path_or_env(config_dir, "STORAGE_DIR", System.tmp_dir!())
```
ruslandoga, line 18: I'll prepare a separate PR with STORAGE_DIR changes.

cnkk: Let's just use `PERSISTENT_CACHE_DIR`?

ruslandoga: Done in https://github.com/plausible/analytics/pull/2986/commits/abcedbe39b9cb2d5f056f3524b7c743ac1395f12

.github/workflows/build-public-images.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           build-args: |
             MIX_ENV=small
             BUILD_METADATA=${{ steps.meta.outputs.json }}
-            ERL_FLAGS=+JPperf true
+            ERL_FLAGS=+JMsingle true
```
ruslandoga, line 51: it should be ready now https://github.com/erlang/otp/pull/6340

ruslandoga: Seems to work, I built an arm64 image and it boots successfully on arm64 mac.

ruslandoga:
> I'm going to run it for a few days on my instance and report back if anything bad happens.

Everything seems OK!

config/runtime.exs
```diff
@@ -229,7 +229,7 @@ ip_geolocation_db = get_var_from_path_or_env(config_dir, "IP_GEOLOCATION_DB", ge
 geonames_source_file = get_var_from_path_or_env(config_dir, "GEONAMES_SOURCE_FILE")
 maxmind_license_key = get_var_from_path_or_env(config_dir, "MAXMIND_LICENSE_KEY")
 maxmind_edition = get_var_from_path_or_env(config_dir, "MAXMIND_EDITION", "GeoLite2-City")
-maxmind_cache_dir = get_var_from_path_or_env(config_dir, "PERSISTENT_CACHE_DIR")
+persistent_cache_dir = get_var_from_path_or_env(config_dir, "PERSISTENT_CACHE_DIR")
```
ruslandoga, line 232: Maybe this could fallback to `/tmp`?

```elixir
persistent_cache_dir = get_var_from_path_or_env(config_dir, "PERSISTENT_CACHE_DIR") || System.tmp_dir!()
```

config/runtime.exs
```diff
@@ -672,9 +672,7 @@ else
     traces_exporter: :none
 end
 
-config :tzdata,
-       :data_dir,
-       get_var_from_path_or_env(config_dir, "STORAGE_DIR", Application.app_dir(:tzdata, "priv"))
+config :tzdata, :data_dir, persistent_cache_dir || Path.join(System.tmp_dir!(), "tzdata_data")
```
ruslandoga, line 675: Should it be scoped to `tzdata_data` dir in both cases?

```elixir
config :tzdata, :data_dir, Path.join(persistent_cache_dir || System.tmp_dir!(), "tzdata_data")
```

config/runtime.exs
cnkk: Sounds good

ruslandoga:
Closing in favor of #3811

PR #2996, opened 2023-06-01
remove custom_dimension_filter feature and views_per_visit_metric flags

ruslandoga:
### Changes

This PR enables custom dimension filter for self-hosters.

Relevant discussion: https://github.com/plausible/analytics/discussions/2984

This PR is based on https://github.com/plausible/analytics/pull/2898

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/filters.js
+++ b/assets/js/dashboard/filters.js
@@ -64,13 +64,6 @@ function renderDropdownFilter(site, history, [key, value], query) {
           <span className="inline-block w-full truncate">{filterText(key, value, query)}</span>
           <PencilSquareIcon className="w-4 h-4 ml-1 cursor-pointer group-hover:text-indigo-700 dark:group-hover:text-indigo-500" />
         </Link>
-        <div
-          onClick={(e) => {e.preventDefault()}}
-          className="group flex w-full justify-between items-center"
-          style={{ width: 'calc(100% - 1.5rem)' }}
-        >
-          <span className="inline-block w-full truncate">{filterText(key, value, query)}</span>
-        </div>
         <b title={`Remove filter: ${formattedFilters[key]}`} className="ml-2 cursor-pointer hover:text-indigo-700 dark:hover:text-indigo-500" onClick={() => removeFilter(key, history, query)}>
           <XMarkIcon className="w-4 h-4" />
         </b>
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -326,7 +326,6 @@ defmodule PlausibleWeb.StatsController do
 
   defp get_flags(user) do
     %{
-      custom_dimension_filter: FunWithFlags.enabled?(:custom_dimension_filter, for: user),
       views_per_visit_metric: FunWithFlags.enabled?(:views_per_visit_metric, for: user)
```
ruslandoga, line 329: Can this be enabled as well?

https://github.com/plausible/analytics/discussions/131#discussioncomment-6052701

ukutaht:
@RobertJoonas could you take a look? I think you're most familiar with this stuff

assets/js/dashboard/filters.js
```diff
@@ -83,10 +79,6 @@ function renderDropdownFilter(site, history, [key, value], query) {
   )
 }
 
-function shouldLinkToFilterModal(site, key) {
-  return key !== 'props' || site.flags.custom_dimension_filter
```
aerosol, line 87: I lack context here but sanity check -- shouldn't we retain the `key !== 'props'` condition?

aerosol (approved):

assets/js/dashboard/filters.js
ruslandoga: I lack context here too. My thinking was, if `custom_dimension_filter` is enabled for all, then it's always `true`, so even if `key !== 'props'` returns `false`, the end result is still `true`.

aerosol:
@ruslandoga could you resolve the CHANGELOG conflict please?

assets/js/dashboard/filters.js
RobertJoonas: The `key !== 'props'` check makes sure that if `key` is anything other than 'props' (e.g. 'page') then it should return `true`, no matter the flag status. We can remove that too as we don't need any special handling for props anymore, all of the options should be links.

lib/plausible_web/controllers/stats_controller.ex
RobertJoonas: > Can views_per_visit_metric be enabled as well?

@ruslandoga, the `views_per_visit_metric` flag is actually redundant. It's a duplicate that can be safely deleted now.

We used a `visits_metric` flag for both `visits` and `views_per_visits` metrics. And that flag was removed in https://github.com/plausible/analytics/pull/2898, so in the current master the `views_per_visit` metric is enabled for everyone.

RobertJoonas (approved):
Looks good to me!

assets/js/dashboard/filters.js
```diff
@@ -55,26 +55,22 @@ function renderDropdownFilter(site, history, [key, value], query) {
...
+        <Link
+          title={`Edit filter: ${formattedFilters[key]}`}
+          to={{ pathname: `/${encodeURIComponent(site.domain)}/filter/${filterGroupForFilter(key)}`, search: window.location.search }}
+          className="group flex w-full justify-between items-center"
+          style={{ width: 'calc(100% - 1.5rem)' }}
+        >
+          <span className="inline-block w-full truncate">{filterText(key, value, query)}</span>
+          <PencilSquareIcon className="w-4 h-4 ml-1 cursor-pointer group-hover:text-indigo-700 dark:group-hover:text-indigo-500" />
+        </Link>
+        <div
```
ruslandoga, line 67: Oops, this shouldn't be included, since it was using a `!shouldLinkToFilterModal(site, key)` check.

ruslandoga: Fixed in https://github.com/plausible/analytics/pull/2996/commits/bc7a61640bb77f8520ba6219a98021eaa7120408

lib/plausible_web/controllers/stats_controller.ex
ruslandoga: PR into this branch: https://github.com/plausible/analytics/pull/3001

RobertJoonas: @ruslandoga nice! I've approved :)

RobertJoonas (approved):

PR #3005, opened 2023-06-05
UI groundwork: Conversions to Behaviors

RobertJoonas:
### Changes

This PR builds some foundation for the new Funnels and Properties sections. Depends on [this migration PR](https://github.com/plausible/analytics/pull/2988). The idea is that on a fresh dashboard the bottom section will be displayed by default with 3 tab selections: Conversions, Funnels, and Properties. Conversions will be the default tab as it is right now. 

### Tracking progress

- [x] Merge, deploy, and migrate https://github.com/plausible/analytics/pull/2988 
- [x] Add the site schema change (add the three new fields)
- [x] `Conversions` &rarr; `Behaviours`- separate three tabs, of which the two new tabs (Funnels and Properties) are hidden behind a feature flag 
- [x] Add a "setup or hide" notice for each tab
- [x] Add a switch to Site Settings > Goals tab to enable/disable the Goal Conversions section
- [x] Optimize for light mode
- [x] Limit seeing the setup notice only for logged in users with an `admin` or `owner` role (disable for shared links)
- [x] Add a note next to the goal creation form about having to send custom events to Plausible

### Tests

### Changelog

### Documentation

### Dark mode


RobertJoonas:
@metmarkosaric sounds good! Something like this?

https://github.com/plausible/analytics/assets/56999674/62a51d4a-fcba-46b6-8e0a-29246eadb794

metmarkosaric:
looks nice @RobertJoonas! what do you think about allowing people to click to hide and it automatically closes the message without any need to confirm it?

RobertJoonas:
@metmarkosaric 

> looks nice @RobertJoonas! what do you think about allowing people to click to hide and it automatically closes the message without any need to confirm it?

This action also hides the entire Goal Conversions section (not only the message), and in order to get it back, you have to go to the site settings. I think it's important to make that clear for the user...

Mobile users would not see the tooltip, so I think the only other option is to say that in the message itself:

"Hide this section by clicking on the icon on the top right. You can make it visible again in your site settings later."

lib/plausible_web/router.ex
```diff
@@ -117,6 +117,8 @@ defmodule PlausibleWeb.Router do
     post "/paddle/webhook", Api.PaddleController, :webhook
 
     get "/:domain/status", Api.InternalController, :domain_status
+    get "/:domain/disable-feature", Api.InternalController, :disable_feature
```
ukutaht, line 120: Since this actually writes something to the database, I don't think `GET` is the appropriate http verb here. `PUT` is probably better.

ukutaht (approved):
Looks good. Just one comment about an http verb

aerosol:
image

A little verbose perhaps, but I'm wondering if something like "Show Goal Conversions panel on the dashboard" would make the intent easier to understand.

metmarkosaric:
yeah, that is clearer. "Show goals in the dashboard" is a shorter version

RobertJoonas:
I forgot to do that before, but here are also the screenshots of the goal creation UI with the new note about having to set up custom events:

image
image

For pageview goals, there is no message displayed:

image

@metmarkosaric @ukutaht your styling suggestions are welcome :)

metmarkosaric:
looks good to me @RobertJoonas!

ukutaht (approved):

PR #3023, opened 2023-06-13
Funnels backend core

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/funnels.ex
```diff
@@ -0,0 +1,229 @@
...
+    |> Ecto.Changeset.apply_changes()
+  end
+
+  @spec list(Plausible.Site.t()) :: [Funnel.t()]
+  def list(%Plausible.Site{id: site_id}) do
+    Repo.all(
+      from f in Funnel,
+        inner_join: steps in assoc(f, :steps),
+        where: f.site_id == ^site_id,
+        select: %{name: f.name, id: f.id, steps_count: count(steps)},
```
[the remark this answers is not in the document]
aerosol: Ah, thanks, good call - e99f61c9

lib/plausible/funnels.ex
```diff
@@ -0,0 +1,229 @@
...
+          steps: {steps, goal: goal}
+        ]
+
+    funnel = Repo.one(q)
+
+    if funnel do
+      {:ok, funnel}
+    else
+      {:error, "Funnel not found"}
+    end
```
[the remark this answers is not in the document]
aerosol: Sure, why not, there's only one caller thus far, and the `get` interface is overexposed currently. One could argue that no funnel is always an error you'd want to match on, so perhaps an atom reason would be more applicable here. 

But, do you mean something like: 242081e5 ?

lib/plausible/goals.ex
```diff
@@ -47,26 +50,101 @@ defmodule Plausible.Goals do
...
+        query
+      end
+
     query
     |> Repo.all()
     |> Enum.map(&maybe_trim/1)
   end
 
+  def by_id!(site, id) do
+    Repo.one(
```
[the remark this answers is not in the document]
aerosol: I think this is some remnant, thanks for spotting it. Removed via 5f3bd291

lib/plausible/funnels.ex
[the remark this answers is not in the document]
vinibrsl: That's what I meant. Thank you!

lib/plausible/funnel.ex
```diff
@@ -0,0 +1,76 @@
...
+  Each step references a goal (either a Custom Event or Visit)
+  - see: `Plausible.Goal`.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Plausible.Funnel.Step
+
+  defmacro min_steps() do
```
ukutaht, line 22: I usually create a normal function to access a module variable. Is the macro an optimization to 'inline' or avoid the function call at runtime?

lib/plausible/funnels.ex
```diff
@@ -0,0 +1,229 @@
...
+    case get(site.id, funnel_id) do
+      %Funnel{} = funnel_definition ->
+        evaluate(query, funnel_definition, site)
+
+      nil ->
+        {:error, :funnel_not_found}
+    end
+  end
+
+  def evaluate(query, funnel_definition, site) do
```
ukutaht, line 104: We currently have 3 main entry points to query analytics data from Clickhouse:
1. Plausible.Stats.Aggregate
2. Plausible.Stats.Breakdown
3. Plausible.Stats.Timeseries

These operations are exposed via defdelegate in `Plausible.Stats`

```elixir
Plausible.Stats.aggregate(site, query, metrics)
Plausible.Stats.breakdown(site, query, property, metrics)
Plausible.Stats.timeseries(site, query, metrics)
```

I think the funnel constitutes a new 'query type` and I would propose to follow the same pattern for Clickhouse funnel data access:

```elixir
# lib/plausible/stats/funnel.ex
defmodule Plausible.Stats.Funnel do
  def funnel(site, query, funnel) do
     # Body of current Funnels.evaluate/3
  end
end

# lib/plausible/stats/mod.ex
defmodule Plausible.Stats do
    # previous delegates
    defdelegate funnel(site, query, funnel), to: Plausible.Stats.Funnel
end

# Use it like so
Plausible.Stats.funnel(site, query, funnel)
```

My comment only applies to the Clickhouse query building (i.e. the `evaluate/3` function). I think all the other functions that manage Funnels in Postgres should stay in this context module.

lib/plausible/funnels.ex
```diff
@@ -0,0 +1,229 @@
+defmodule Plausible.Funnels do
```
ukutaht, line 1: Missing tests for this module?

lib/plausible/funnels.ex
ukutaht: This reminds me – I don't think it's necessary for the initial public release but soon after we'll have to add funnels to the Stats API as well.

lib/plausible/funnel.ex
aerosol: A macro can be used in guards ([it is, later on](https://github.com/plausible/analytics/pull/3023/files#diff-5e4e6d6ec0bdd4f07ac0ead61420e716556a749b1670b6e4b4bf0facd25e4a0dR23)). I'm not sure now but I recall some issues with Phoenix Live Component stuff (the ~H sigil IIRC).

lib/plausible/funnels.ex
aerosol: OMG totally, I must've lost them extracting from the wip branch.

aerosol: cfbb7cd1 and cfbb7cd1

lib/plausible/funnels.ex
aerosol: Like [this](3738bd37)? I took the liberty of moving the `mod.ex` file, so it conforms to Elixir conventions.

ukutaht: Yes, perfect :)

ukutaht (approved):

PR #3027, opened 2023-06-13
Pass cached site struct down the ingestion pipeline

vinibrsl:
Revenue goals need the cached site struct during ingestion to get the goals name and currency. This cache lookup is not necessary as `GateKeeper.check/1`, which is called first in the ingestion pipeline, could already return the site struct from the cache.

This commit changes `GateKeeper.check/1` to return the site struct instead of the site ID. Moreover, this commit changes the ingestion pipeline to avoid calling the sites cache twice.

Related: https://github.com/plausible/analytics/pull/2957#discussion_r1203921549

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -10,7 +10,7 @@ defmodule Plausible.Ingestion.Event do
   alias Plausible.Site.GateKeeper
 
   defstruct domain: nil,
-            site_id: nil,
+            site: nil,
             clickhouse_event_attrs: %{},
             clickhouse_event: nil,
             dropped?: false,
@@ -27,7 +27,7 @@ defmodule Plausible.Ingestion.Event do
 
   @type t() :: %__MODULE__{
           domain: String.t() | nil,
-          site_id: pos_integer() | nil,
+          site: %Plausible.Site{} | nil,
           clickhouse_event_attrs: map(),
           clickhouse_event: %ClickhouseEventV2{} | nil,
           dropped?: boolean(),
@@ -45,10 +45,10 @@ defmodule Plausible.Ingestion.Event do
       else
         Enum.reduce(domains, [], fn domain, acc ->
           case GateKeeper.check(domain) do
-            {:allow, site_id} ->
+            {:allow, site} ->
               processed =
                 domain
-                |> new(site_id, request)
+                |> new(site, request)
                 |> process_unless_dropped(pipeline())
 
               [processed | acc]
@@ -120,8 +120,8 @@ defmodule Plausible.Ingestion.Event do
     struct!(__MODULE__, domain: domain, request: request)
   end
 
-  defp new(domain, site_id, request) do
-    struct!(__MODULE__, domain: domain, site_id: site_id, request: request)
+  defp new(domain, site, request) do
+    struct!(__MODULE__, domain: domain, site: site, request: request)
   end
 
   defp drop(%__MODULE__{} = event, reason, attrs \\ []) do
@@ -163,7 +163,7 @@ defmodule Plausible.Ingestion.Event do
   defp put_basic_info(%__MODULE__{} = event) do
     update_attrs(event, %{
       domain: event.domain,
-      site_id: event.site_id,
+      site_id: event.site.id,
       timestamp: event.request.timestamp,
       name: event.request.event_name,
       hostname: event.request.hostname,
@@ -208,7 +208,7 @@ defmodule Plausible.Ingestion.Event do
   defp put_props(%__MODULE__{} = event), do: event
 
   defp put_revenue(%__MODULE__{request: %{revenue_source: %Money{} = revenue_source}} = event) do
-    revenue_goals = Plausible.Site.Cache.get(event.domain).revenue_goals || []
+    revenue_goals = event.site.revenue_goals || []
 
     matching_goal =
       Enum.find(revenue_goals, &(&1.event_name == event.clickhouse_event_attrs.name))
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -208,10 +208,8 @@ defmodule Plausible.Ingestion.Event do
   defp put_props(%__MODULE__{} = event), do: event
 
   defp put_revenue(%__MODULE__{request: %{revenue_source: %Money{} = revenue_source}} = event) do
-    revenue_goals = event.site.revenue_goals || []
-
     matching_goal =
-      Enum.find(revenue_goals, &(&1.event_name == event.clickhouse_event_attrs.name))
+      Enum.find(event.site.revenue_goals, &(&1.event_name == event.clickhouse_event_attrs.name))
 
     cond do
       is_nil(matching_goal) ->
--- a/lib/plausible/goals.ex
+++ b/lib/plausible/goals.ex
@@ -5,12 +5,35 @@ defmodule Plausible.Goals do
 
   use Plausible.Funnel
 
-  def create(site, params) do
+  @doc """
+  Creates a Goal for a site.
+
+  If the created goal is a revenue goal, it sets site.updated_at to be
+  refreshed by the sites cache, as revenue goals are used during ingestion.
+  """
+  def create(site, params, now \\ DateTime.utc_now()) do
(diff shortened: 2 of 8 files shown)


ukutaht (approved):

lib/plausible/ingestion/event.ex
```diff
@@ -208,7 +208,7 @@ defmodule Plausible.Ingestion.Event do
   defp put_props(%__MODULE__{} = event), do: event
 
   defp put_revenue(%__MODULE__{request: %{revenue_source: %Money{} = revenue_source}} = event) do
-    revenue_goals = Plausible.Site.Cache.get(event.domain).revenue_goals || []
+    revenue_goals = event.site.revenue_goals || []
```
aerosol, line 211: I don't think this `|| []` fallback is necessary here. The early preload should provide it, no?

aerosol (changes requested):
I think there are some issues with how revenue goals are cached via `Plausible.Cache` which results with inconsistent state that's going to be hard to debug until the full refresh kicks in.

Consider this test:

```elixir
    test "cache caches revenue goals with revent refresh", %{test: test} do
      {:ok, _} =
        Supervisor.start_link([{Cache, [cache_name: test, child_id: :test_cache_caches_id]}],
          strategy: :one_for_one,
          name: Test.Supervisor.Cache
        )

      yesterday = DateTime.utc_now() |> DateTime.add(-1 * 60 * 60 * 24)

      # the site was added yesterday so full refresh will pick it up
      %{id: site_id} = site = insert(:site, domain: "site1.example.com", updated_at: yesterday)
      # the goal was added yesterday so full refresh will pick it up
      insert(:goal, site: site, event_name: "Purchase", currency: :BRL, updated_at: yesterday)
      # this goal is added "just now"
      insert(:goal, site: site, event_name: "Add to Cart", currency: :USD)
      # and this one does not matter
      insert(:goal, site: site, event_name: "Click", currency: nil)

      # at this point, we have 3 goals associated with the cached struct
      :ok = Cache.refresh_all(cache_name: test)

      # the goal was added 70 seconds ago so partial refresh should pick it up and merge with the rest of goals, but the merging does not take place intelligently
      insert(:goal,
        site: site,
        event_name: "Purchase2",
        currency: :BRL,
        updated_at: DateTime.utc_now() |> DateTime.add(-70)
      )

      :ok = Cache.refresh_updated_recently(cache_name: test)

      assert %Site{from_cache?: true, id: ^site_id, revenue_goals: cached_goals} =
               Cache.get("site1.example.com", force?: true, cache_name: test)

      # instead, this fails with Purchase goal missing from the list
      assert [
               %Goal{event_name: "Add to Cart", currency: :USD},
               %Goal{event_name: "Purchase", currency: :BRL},
               %Goal{event_name: "Purchase2", currency: :BRL}
             ] = Enum.sort_by(cached_goals, & &1.event_name)
    end
```

We'd need to implement proper merging with deeper lookups for nested fields if the current way of querying is something we must keep (what I mean by that, I'd try to experiment with fragments/subqueries, to see if it's any better than the left join scan). 

If we choose to do it that way, there is a [test](https://github.com/plausible/analytics/blob/cached-site-ingestion-pipeline/test/plausible/site/cache_test.exs#L335) that should help assessing the performance impact of merging alone.

Excluding the revenue goals age clause from recently updated cache warmer could work too, if we're okay with revenue goals being refreshed only on full sweeps or events such as domain change.

We could also try modifying the age clause to account for any currency having goal instead of filtering.

Either way it feels like this is specific enough to be covered with an extra test (under `Plausible.Site.CacheTest`).

While at it, you could change the child id [here](https://github.com/plausible/analytics/blob/fd158539657c5e0733f4bd58c3c607c40d43fc86/test/plausible/site/cache_test.exs#L58), because it makes the test prone to race conditions as seen here: https://github.com/plausible/analytics/actions/runs/5264345107/jobs/9515511593?pr=3029

Side note, we should be mindful of cache refresh times. The optimization done here for the single cache lookup is okay, but those lookups are cheap as opposed to potentially quirky left joins - what we can observe already since the new join was introduced via r2023.05.24-117:

image

lib/plausible/ingestion/event.ex
vinibrsl: Thanks, fixed in 0969381

vinibrsl:
Thank you, @aerosol, for catching this issue and writing this test case, @aerosol!

Revenue goals are now required during ingestion, and insertions are ideally available on the cache. I tried some solutions but they would bloat the cache query even more, so I went with an alternative fix.

d537bde removes the revenue goal `ago` condition from the cache query, leaving only the site timestamp as a filter. However, in order for revenue goals to be available, I had to update `sites.updated_at` whenever a new revenue goal is created. Deletions and updates are fine to be miss, but goal insertions are required by ingestion.

aerosol (approved):

ukutaht:
> I think there are some issues with how revenue goals are cached via `Plausible.Cache` which results with inconsistent state that's going to be hard to debug until the full refresh kicks in.
> 
> Consider this test:
> 
> ```elixir
>     test "cache caches revenue goals with revent refresh", %{test: test} do
>       {:ok, _} =
>         Supervisor.start_link([{Cache, [cache_name: test, child_id: :test_cache_caches_id]}],
>           strategy: :one_for_one,
>           name: Test.Supervisor.Cache
>         )
> 
>       yesterday = DateTime.utc_now() |> DateTime.add(-1 * 60 * 60 * 24)
> 
>       # the site was added yesterday so full refresh will pick it up
>       %{id: site_id} = site = insert(:site, domain: "site1.example.com", updated_at: yesterday)
>       # the goal was added yesterday so full refresh will pick it up
>       insert(:goal, site: site, event_name: "Purchase", currency: :BRL, updated_at: yesterday)
>       # this goal is added "just now"
>       insert(:goal, site: site, event_name: "Add to Cart", currency: :USD)
>       # and this one does not matter
>       insert(:goal, site: site, event_name: "Click", currency: nil)
> 
>       # at this point, we have 3 goals associated with the cached struct
>       :ok = Cache.refresh_all(cache_name: test)
> 
>       # the goal was added 70 seconds ago so partial refresh should pick it up and merge with the rest of goals, but the merging does not take place intelligently
>       insert(:goal,
>         site: site,
>         event_name: "Purchase2",
>         currency: :BRL,
>         updated_at: DateTime.utc_now() |> DateTime.add(-70)
>       )
> 
>       :ok = Cache.refresh_updated_recently(cache_name: test)
> 
>       assert %Site{from_cache?: true, id: ^site_id, revenue_goals: cached_goals} =
>                Cache.get("site1.example.com", force?: true, cache_name: test)
> 
>       # instead, this fails with Purchase goal missing from the list
>       assert [
>                %Goal{event_name: "Add to Cart", currency: :USD},
>                %Goal{event_name: "Purchase", currency: :BRL},
>                %Goal{event_name: "Purchase2", currency: :BRL}
>              ] = Enum.sort_by(cached_goals, & &1.event_name)
>     end
> ```
> 
> We'd need to implement proper merging with deeper lookups for nested fields if the current way of querying is something we must keep (what I mean by that, I'd try to experiment with fragments/subqueries, to see if it's any better than the left join scan).
> 
> If we choose to do it that way, there is a [test](https://github.com/plausible/analytics/blob/cached-site-ingestion-pipeline/test/plausible/site/cache_test.exs#L335) that should help assessing the performance impact of merging alone.
> 
> Excluding the revenue goals age clause from recently updated cache warmer could work too, if we're okay with revenue goals being refreshed only on full sweeps or events such as domain change.
> 
> We could also try modifying the age clause to account for any currency having goal instead of filtering.
> 
> Either way it feels like this is specific enough to be covered with an extra test (under `Plausible.Site.CacheTest`).
> 
> While at it, you could change the child id [here](https://github.com/plausible/analytics/blob/fd158539657c5e0733f4bd58c3c607c40d43fc86/test/plausible/site/cache_test.exs#L58), because it makes the test prone to race conditions as seen here: https://github.com/plausible/analytics/actions/runs/5264345107/jobs/9515511593?pr=3029
> 
> Side note, we should be mindful of cache refresh times. The optimization done here for the single cache lookup is okay, but those lookups are cheap as opposed to potentially quirky left joins - what we can observe already since the new join was introduced via r2023.05.24-117:
> 
> image

What do you think about adding an alert for when cache refresh takes more than N seconds?

PR #3040, opened 2023-06-15
Remove revenue metrics from pageview goal query

vinibrsl:
This commit fixes a bug where performing a breakdown by goal with revenue metrics would return data for both event goals and pageview goals. However, revenue data is only applicable for event goals as they hold currency data.

To fix the bug, this commit removes revenue metrics from the pageview breakdown query.

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -7,25 +7,26 @@ defmodule Plausible.Stats.Breakdown do
   @no_ref "Direct / None"
   @not_set "(not set)"
 
-  @event_metrics [:visitors, :pageviews, :events]
+  @event_metrics [:visitors, :pageviews, :events, :average_revenue, :total_revenue]
   @session_metrics [:visits, :bounce_rate, :visit_duration]
+  @revenue_metrics [:average_revenue, :total_revenue]
   @event_props Plausible.Stats.Props.event_props()
 
   def breakdown(site, query, "event:goal" = property, metrics, pagination) do
     site = Plausible.Repo.preload(site, :goals)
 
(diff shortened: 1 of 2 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -54,7 +55,7 @@ defmodule Plausible.Stats.Breakdown do
             goal: fragment("concat('Visit ', ?[index])", ^page_exprs)
           }
         )
-        |> select_event_metrics(metrics)
+        |> select_event_metrics(metrics -- @revenue_metrics)
```
vinibrsl, line 58: The actual fix is this. This branch only runs when there are pageview goals, and for pageview goals it doesn't make sense to query revenue data. Only event goals support revenue tracking.

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -275,36 +275,48 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
...
+                 "total_conversions" => 2,
+                 "total_revenue" => nil,
+                 "unique_conversions" => 2
+               },
+               %{
+                 "average_revenue" => nil,
                  "conversion_rate" => 33.3,
-                 "average_revenue" => %{"long" => "€10.00", "short" => "€10.0"},
-                 "total_revenue" => %{"long" => "€10.00", "short" => "€10.0"}
+                 "name" => "Visit /checkout",
```
vinibrsl, line 313: The actual change in this test is that `"Visit /checkout"`, a pageview metric, was returning a revenue data, and now it returns `nil`.

PR #3053, opened 2023-06-19
Set with_imported=false in realtime mode

vinibrsl:
This commit fixes a bug where timeseries queries - such as the main graph - would fail with imported data in realtime mode. Realtime mode `date` field is not actually a date, but minutes from now. This would cause the imported tables join to fail with this error:

```
(Ch.Error Code: 53. DB::Exception: Can't infer common type for joined
columns: date: Int64 at left, s1.date: Date at right. There is no
supertype for types Int64, Date because some of them are
Date/Date32/DateTime/DateTime64 and some of them are not.
(TYPE_MISMATCH)
```

This commit removes imported data from realtime queries, as it doesn't make sense to include it. Imported data does not have time precision, and would only appear in the first day the data was imported anyways.


RobertJoonas (approved):
Looks good to me 👍

PR #3057, opened 2023-06-19
Pin alpine versions to 3.18.0

aerosol:
This reverts commit c19a51efb0dbe2da0f8dede8e3d5df167bd7fba1.

@cnkk we have two alpine image references in our Dockerfile and only one got updated. This may lead to unpredictable build oddities.

### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # platform specific, it makes sense to build it in the docker
 
 #### Builder
-FROM hexpm/elixir:1.14.3-erlang-25.2.3-alpine-3.17.0 as buildcontainer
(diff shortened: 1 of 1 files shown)


aerosol:
actually even with the revert they're diverged already

cnkk (approved):
Good point. I missed this earlier

aerosol:
Ah great, thank you for looking into it. 

This PR was a knee-jerk reaction to some random errors (of unknown then source) and I only remembered we've had issues with Erlang's SSL in the past due to a imprecise bump.

PR #3059, opened 2023-06-19
Add revenue metrics to top stats

vinibrsl:
### Changes

This PR adds revenue data to top stats. Average and total are displayed when filtering by a revenue goal (or many if they have the same currency set).

Screenshot 2023-06-19 at 18 50 50

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/aggregate.ex
+++ b/lib/plausible/stats/aggregate.ex
@@ -3,10 +3,24 @@ defmodule Plausible.Stats.Aggregate do
   use Plausible.ClickhouseRepo
   import Plausible.Stats.{Base, Imported, Util}
 
-  @event_metrics [:visitors, :pageviews, :events, :sample_percent]
+  @event_metrics [
+    :visitors,
+    :pageviews,
+    :events,
+    :sample_percent,
+    :average_revenue,
+    :total_revenue
+  ]
   @session_metrics [:visits, :bounce_rate, :visit_duration, :views_per_visit, :sample_percent]
+  @revenue_metrics [:average_revenue, :total_revenue]
 
   def aggregate(site, query, metrics) do
+    # Aggregating revenue data works only for same currency goals. If the query
+    # is filtered by goals with different currencies, for example, one USD and
+    # other EUR, revenue metrics are dropped.
+    currency = get_revenue_tracking_currency(site, query)
+    metrics = if currency, do: metrics, else: metrics -- @revenue_metrics
+
     event_metrics = Enum.filter(metrics, &(&1 in @event_metrics))
     event_task = fn -> aggregate_events(site, query, event_metrics) end
     session_metrics = Enum.filter(metrics, &(&1 in @session_metrics))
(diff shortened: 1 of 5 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -329,54 +329,61 @@ defmodule PlausibleWeb.Api.StatsController do
     {stats, 100}
   end
 
-  defp fetch_top_stats(site, %Query{filters: %{"event:goal" => _goal}} = query, comparison_query) do
-    totals_query = Query.remove_event_filters(query, [:goal, :props])
+  defp fetch_top_stats(site, %Query{filters: %{"event:goal" => _}} = query, comparison_query) do
```
vinibrsl, line 332: Revenue metrics are not always present, so I had to refactor this function before adding revenue metrics. If it's easier to review:

1. Refactor: a8adc012aa5bf4f6d6b3738d5f1fba6980889b80
2. Revenue metrics: 1f74419f11497e53e872822b56aeca92d14053e1

lib/plausible/stats/aggregate.ex
```diff
@@ -3,10 +3,24 @@ defmodule Plausible.Stats.Aggregate do
...
+  ]
   @session_metrics [:visits, :bounce_rate, :visit_duration, :views_per_visit, :sample_percent]
+  @revenue_metrics [:average_revenue, :total_revenue]
 
   def aggregate(site, query, metrics) do
+    # Aggregating revenue data works only for same currency goals. If the query
+    # is filtered by goals with different currencies, for example, one USD and
+    # other EUR, revenue metrics are dropped.
+    currency = get_revenue_tracking_currency(site, query)
+    metrics = if currency, do: metrics, else: metrics -- @revenue_metrics
```
ukutaht, line 22: API design: in case when currency cannot be resolved, I'm not sure which is better:
1. Don't return the field at all (current implementation)
2. Return the field as `null`

Thoughts @aerosol ?

lib/plausible/stats/aggregate.ex
```diff
@@ -134,4 +149,38 @@ defmodule Plausible.Stats.Aggregate do
...
+
+  defp get_revenue_tracking_currency(site, query) do
+    goal_filters =
+      case query.filters do
+        %{"event:goal" => {:is, {_, goal_name}}} -> [goal_name]
+        %{"event:goal" => {:member, list}} -> Enum.map(list, fn {_, goal_name} -> goal_name end)
+        _any -> []
+      end
+
+    revenue_goals = Plausible.Site.Cache.get(site.domain).revenue_goals
```
ukutaht, line 161: We are planning to separate the workloads of ingestion and app. I think ingestion will be running the `Site.Cache` but app will not. So this call here is crossing a boundary it should not cross.

As a solution, I think it's better to look up revenue goals in the controller, in case revenue metrics are requested. And use Repo directly instead of going through the Site.Cache

ukutaht (approved):

PR #3062, opened 2023-06-20
Postgres upgrade

cnkk:


aerosol (approved):

.github/workflows/elixir.yml
cnkk, line 1: We're using v15 on prod now.

ruslandoga:
I think it still might make sense to use postgres:15-alpine in Makefile's postgres target. Even if coincides with prod right now.

PR #3066, opened 2023-06-21
Implement Funnels view on dashboard

aerosol:
### Changes

The PR implements funnel presentation on the dashboard.

https://github.com/plausible/analytics/assets/173738/eed523b7-99f9-4e10-917f-48c4ec1d9fc2

https://github.com/plausible/analytics/assets/173738/01f92428-d81d-49da-97cd-42f0ad4c958e

https://github.com/plausible/analytics/assets/173738/ed294e6f-9aeb-4405-b590-7fe442516305

The following additions have been made:
  - Funnel picker on the dashboard, renders either the funnel chart or a simplified bar-based version on mobile.
  - Funnel evaluation result (and in consequence the react API) has been extended with extra properties to make step-by-step tooltips informative.
  - Querying funnels is now done aggregating `user_id` rather than `session_id` so we can track the journey across multiple sessions (within the rotating 24h window). Need @ukutaht's ack this - comment inline.

Kindly asking for review \cc @RobertJoonas, @vinibrsl 

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated (wip at https://github.com/plausible/docs/pull/407)

### Dark mode


lib/plausible/stats/funnel.ex
```diff
@@ -33,13 +33,39 @@ defmodule Plausible.Stats.Funnel do
...
+       entering_visitors: visitors_at_first_step,
+       entering_visitors_percentage: percentage(visitors_at_first_step, all_visitors),
+       never_entering_visitors: all_visitors - visitors_at_first_step,
+       never_entering_visitors_percentage:
+         percentage(all_visitors - visitors_at_first_step, all_visitors)
      }}
   end
 
+  defp query_all_visitors(query, _goal) do
+    # condition = goal_condition(goal)
```
aerosol, line 57: @ukutaht see my comment on basecamp - I think this must not belong here, otherwise we'll only count users who have effectively entered the funnel and stopped their journey at step 1

aerosol: fixed.

assets/js/dashboard/stats/behaviours/funnel.js
```diff
@@ -246,12 +263,56 @@ export default function Funnel(props) {
...
+
+          <span className="font-medium dark:text-gray-200 w-20 text-right" tooltip={step.visitors.toLocaleString()}>
+            {numberFormatter(step.visitors)}
+          </span>
+        </div>
+      </>
+    )
+  }
+
+  const renderBars = (funnel) => {
```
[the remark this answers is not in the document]
aerosol: fixed.

assets/js/dashboard/stats/behaviours/funnel-tooltip.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
```
ukutaht, line 2: For the graph tooltip, the decision was made to not use ReactDOMServer in order to save 5KB on file size: https://github.com/plausible/analytics/commit/092e29500bebde0603b8b1f510f8b2ca80af6888

If we bring in the dependency at this point, it's worth refactoring the other tooltip to use it as well.

lib/plausible/stats/funnel.ex
ukutaht: Yes I think the goal condition is not needed.

For the query itself, I think we want:
* Make sure it applies the `SAMPLE` hint to Clickhouse (I'm not sure if it does at the moment)
* Use `uniq(user_id)` over `count(DISTINCT ?)`. DISTINCT will calculate the exact value but `uniq()` uses [HLL](https://en.wikipedia.org/wiki/HyperLogLog) to provide a probabilistic answer with less compute resources. We always use `uniq()` over `DISTINCT` in other queries as well.

lib/plausible/stats/funnel.ex
```diff
@@ -33,41 +33,68 @@ defmodule Plausible.Stats.Funnel do
...
+        select: fragment("count(DISTINCT ?)", e.user_id)
+        # where: ^condition
+      )
+
+    ClickhouseRepo.one(q_events_count)
+  end
+
   defp query_funnel(query, funnel_definition) do
     q_events =
       from(e in query,
```
ukutaht, line 71: As with the other query, we should make sure this applies sampling correctly.

ukutaht: Could the 'not entered' number be calculated from this query by removing the condition `having: fragment("step > 0")` ? Maybe the second query is not needed after all.

assets/js/dashboard/stats/behaviours/funnel-tooltip.js
aerosol: I've added it only so that I can use JSX + formatting of the embedded code. Otherwise caught myself leaving mismatched tags and such. I'll remove if this is an issue.

lib/plausible/stats/funnel.ex
aerosol: It's a subquery, so will have to become a second query anyway if we like to execute outside of the funnel itself. Here, we group by user_id instead of aggregating the count alone. Or have I misunderstood?

assets/js/dashboard/stats/behaviours/funnel-tooltip.js
aerosol: https://github.com/plausible/analytics/pull/3066/commits/423b6fe162b63590afc984b92f79c03703cc883f

lib/plausible/stats/funnel.ex
aerosol: Oh I see what you mean now, let me try this

aerosol: @ukutaht e07011db ✨ kudos for spotting that one

PR #3069, opened 2023-06-22
Add notice to revenue goals UI

vinibrsl:
This pull request changes the UI for revenue goals, replaces the select box with a toggle, and adds a notice. It also polishes up the goal creation UI in general to have the same font sizes and weights compared to the rest of the page.

https://github.com/plausible/analytics/assets/5093045/d43b8d88-b8e3-470b-b33d-e620e94fee55

Thanks to @RobertJoonas for pairing up with me on this one!

### DIFF
--- a/lib/plausible_web/templates/site/new_goal.html.heex
+++ b/lib/plausible_web/templates/site/new_goal.html.heex
@@ -18,7 +18,7 @@
         <%= error_tag f, :event_name %>
       </div>
 
-      <div class="mt-6 space-y-3" x-data="<%= Jason.encode!(%{active: !!Ecto.Changeset.get_field(@changeset, :currency), currency: Ecto.Changeset.get_field(@changeset, :currency)}) %>">
+      <div class="mt-6 space-y-3" x-data={Jason.encode!(%{active: !!Ecto.Changeset.get_field(@changeset, :currency), currency: Ecto.Changeset.get_field(@changeset, :currency)})}>
         <div class="flex items-center w-max cursor-pointer" x-on:click="active = !active; currency = ''">
           <button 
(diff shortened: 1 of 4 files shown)


lib/plausible_web/templates/site/new_goal.html.heex
```diff
@@ -0,0 +1,124 @@
...
+          placeholder: "Signup"
+        ) %>
+        <%= error_tag(f, :event_name) %>
+      </div>
+
+      <div
+        class="mt-6 space-y-3"
+        x-data={
+          Jason.encode!(%{
+            active: !!Ecto.Changeset.get_field(@changeset, :currency),
```
ukutaht, line 40: Won't these always be `nil`? What's the purpose of accessing the changeset here?

lib/plausible_web/templates/site/new_goal.html.heex
```diff
@@ -0,0 +1,124 @@
...
+            during the private preview. Pricing will be announced soon. See
+            examples and learn more in <a
+              class="font-medium text-yellow underline hover:text-yellow-600"
+              href="https://plausible.io/docs/ecommerce-revenue-tracking"
+            >our docs</a>.
+          </p>
+        </div>
+
+        <div x-show="active">
+          <%= label(f, :currency, "Reporting currency", class: "hidden") %>
```
ukutaht, line 83: Is the hidden label necessary/useful?

vinibrsl: I kept the label for accessibility, but in fact, [I just read](https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements) that visually hidden elements are ignored by screen readers. So I replaced the `<label>` for `aria-label` in d3f5aa1d.

PR #3078, opened 2023-06-26
Make the ListReport component more flexible

RobertJoonas:
### Changes

* Make the ListReport React component more configurable so that we could use it for other reports (such as Sources, Goal Conversions, and Prop Breakdown).
* Bugfix - on the live demo, when you set the period to realtime, and then apply a goal filter, then we'll keep fetching the old data (without the goal filter applied) on every consequent realtime tick.
* In the devices reports, keep `visitors` and `%` as separate columns, rather than 
image
--->
image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/historical.js
+++ b/assets/js/dashboard/historical.js
@@ -12,6 +12,7 @@ import Devices from './stats/devices'
 import Behaviours from './stats/behaviours'
 import ComparisonInput from './comparison-input'
 import { withPinnedHeader } from './pinned-header-hoc';
+import { statsBoxClass } from '.';
 
 function Historical(props) {
   const tooltipBoundary = React.useRef(null)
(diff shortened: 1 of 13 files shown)


aerosol:
I have some general feedback on this: I wanted to use this for funnels mobile view but the complexity was a bit discouraging. What I dislike about the current shape of the component is that it tightly couples the API async calls - I believe it should be "dumb", in the sense that it's the component user should be the entity providing the input data, preferably with all the details on where the data is coming from, handled on their end too. E.g. if I have already fetched some data from the API, massaged it, I would like the ability to plot it directly, without the ListReport enforcing making some API calls. Does that make sense? Not sure if this currently doable.

RobertJoonas:
@aerosol

> I have some general feedback on this: I wanted to use this for funnels mobile view but the complexity was a bit discouraging.

I do agree about the complexity being overwhelming - the component is responsible for so many things... But since ListReport is currently used by 9 different reports, and I believe in the future, it should also be used in Conversions, Props, Sources, Referrers, Google search keywords + all the 5 UTM reports, I think it makes sense to encapsulate everything in a single component.

> What I dislike about the current shape of the component is that it tightly couples the API async calls - I believe it should be "dumb", in the sense that it's the component user should be the entity providing the input data, preferably with all the details on where the data is coming from, handled on their end too. E.g. if I have already fetched some data from the API, massaged it, I would like the ability to plot it directly, without the ListReport enforcing making some API calls. Does that make sense? Not sure if this currently doable.

One of the core functionalities of ListReport is that it updates itself on every "tick" (see `realtime-update-timer.js`). If we want to feed the data to ListReport and make it "dumb", then we'd have to specify that realtime update logic in every place where we call out a `<ListReport />`.

Alternatively, we could make the realtime update functionality optional, which sounds like a better option to me.

@aerosol, I'm actually not sure about whether the funnels mobile view should be using ListReport 🤔 IDK, to me it actually sounds a little bit like a different use case.

aerosol:
I think what I was trying to say is that the separation of concerns could've been better in such way that ListReport composes List and Report components maybe, so the List bit could be reused with greater freedom. Not a big deal, especially considering that it works for all cases except mine 😀

PR #3097, opened 2023-06-29
Evaluate ephemeral funnels before saving

aerosol:
### Changes

This PR introduces funnel evaluation during funnel creation, so that it's easier to get a grasp of whether the funnel has been configured properly. The funnel form will now show data similar to the funnel chart with every step selection change updating the result. 

By the way, percentage compaction (i.e. `0` instead of `0.00`) was moved server-side, so that it's consistent across the whole stack. Primitives used for percentage calculation are available in the API anyway, in case clients want to use their own representation.

https://github.com/plausible/analytics/assets/173738/398a8e89-3135-47e1-8d11-b6420464e77f

Dark mode:

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/funnel.ex
+++ b/lib/plausible/stats/funnel.ex
@@ -169,14 +169,22 @@ defmodule Plausible.Stats.Funnel do
   end
 
   defp percentage(x, y) when x in [0, nil] or y in [0, nil] do
-    "0.00"
+    "0"
   end
 
   defp percentage(x, y) do
-    x
-    |> Decimal.div(y)
-    |> Decimal.mult(100)
-    |> Decimal.round(2)
-    |> Decimal.to_string()
+    result =
+      x
+      |> Decimal.div(y)
+      |> Decimal.mult(100)
+      |> Decimal.round(2)
+      |> Decimal.to_string()
+
+    case result do
+      <<compact::binary-size(1), ".00">> -> compact
+      <<compact::binary-size(2), ".00">> -> compact
+      <<compact::binary-size(3), ".00">> -> compact
+      decimal -> decimal
+    end
   end
 end
--- a/lib/plausible_web/live/funnel_settings/form.ex
+++ b/lib/plausible_web/live/funnel_settings/form.ex
@@ -31,14 +31,15 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
        form: to_form(Plausible.Funnels.create_changeset(site, "", [])),
        goals: goals,
        site: site,
-       selections_made: Map.new()
+       selections_made: Map.new(),
+       evaluation_result: nil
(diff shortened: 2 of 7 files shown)


lib/plausible_web/live/funnel_settings/form.ex
```diff
@@ -240,12 +286,55 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
...
+    {:noreply, assign(socket, evaluation_result: evaluate_funnel(socket))}
+  end
+
+  defp evaluate_funnel(socket) do
+    if map_size(socket.assigns.selections_made) >= Funnel.min_steps() do
+      site = socket.assigns.site
+
+      steps =
+        socket.assigns.selections_made
+        |> Enum.sort_by(&elem(&1, 0))
```
RobertJoonas: So the order of the `selections_made` map can change? I removed the `Enum.sort_by(&elem(&1, 0))` clause and all tests still passed. Maybe we can cover that with a test?

lib/plausible/stats/funnel.ex
```diff
@@ -169,14 +169,22 @@ defmodule Plausible.Stats.Funnel do
...
+      |> Decimal.mult(100)
+      |> Decimal.round(2)
+      |> Decimal.to_string()
+
+    case result do
+      <<compact::binary-size(1), ".00">> -> compact
+      <<compact::binary-size(2), ".00">> -> compact
+      <<compact::binary-size(3), ".00">> -> compact
+      decimal -> decimal
+    end
```
RobertJoonas, line 188: Nice! I think this could even be a public util function, not only used by funnels, but throughout the project for the `CR` and `%` metrics as well (just a note, doesn't have to be in scope of this PR of course)

RobertJoonas (approved):
Nice work @aerosol! I played around with the UI a little bit, and it looks super user-friendly. Left two comments inline. Overall it looks good to me! :)

lib/plausible/stats/funnel.ex
aerosol: Cool. It's either this or check the remainder. Note that I specifically return strings from the API layer, trying to avoid floats in JSON.

lib/plausible_web/live/funnel_settings/form.ex
aerosol: Maps do not have order - see https://hexdocs.pm/elixir/1.12/Map.html - This is standard hygiene. I can't reliably test this with a unit test, because it's unspecified how to create a map that will break this.

RobertJoonas: Ah, I see. Makes sense!

vinibrsl (approved):

Issue #2982, opened 2023-05-27
All time view not working correctly


LuD-up:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

Hello,

For a few days now, the "All time" view has been blocked at "Month" only.

https://github.com/plausible/analytics/assets/77074386/5d3f643c-e8d9-4ccc-993f-23fd6418a965

### Expected behavior

Normally, the tab changes over the selected duration.

### Environment

```markdown
- OS: MacOS/iOS
- Browser: Safari
- Browser Version: 16.4
```

ukutaht:
Sorry I don't fully understand the report. What do you expect to see in the dashboard and what do you see instead?

LuD-up:
I can't get the Days/Weeks views in All time, only Months. It happen only on the "All time" view, everything else is working correctly.

metmarkosaric:
thanks for reporting @LuD-up! i'm able to reproduce this too. in all time view, it's not possible to change the interval from the default month to week or day. when you click on a different interval, it goes back the the default monthly interval. doesn't look to be happening on other time ranges

toniengelhardt:
Would also love to see this fixed. 

A temporary workaround is to use custom range and then select the first available date and the last.

kekkodance:
I get this now. Latest version

ruslandoga:
@kekkodance 👋

Note that the current `latest` self-hosted release might not contain all bug fixes that have been merged.

Issue #2713, opened 2023-02-28
Web page frozen when displaying "all time" data 


mquandalle:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

Since a few days I can't display my stats page on Firefox, the page is loading, my fan starts spinning and I some point I have to kill it.

### Expected behavior

No infinite loop, the page should load.

### Screenshots

![Peek 28-02-2023 16-13](https://user-images.githubusercontent.com/1730702/221896148-3bd95ada-3804-4a3e-9ab8-40d6c587cb8b.gif)

### Environment

```markdown
- OS: Linux PopOS
- Browser: Firefox
- Browser Version: 109.0.1
```

metmarkosaric:
thanks for reporting that @mquandalle! it seems related to changing the interval to daily as on default view it works fine. if you clear your browser cache, you should be able to see the default view again at least. we'll take a look at what's going on

mquandalle:
Yes it seems related to the daily view. But even if I reload with clearing the cache (<kbd>Ctrl</kbd> + <kbd>Maj</kbd> + <kbd>R</kbd>) it still stays on the daily view and can't load the page (so I'm stuck I can't change to the monthly view as I cannot interact with the app). Also this is a regression as this view worked before.

metmarkosaric:
could you try to clear the cookies and site date? then you can log back in and you should be able to view the default view again

![Screenshot from 2023-02-28 16-53-06](https://user-images.githubusercontent.com/34340819/221906702-b0dad1c5-b86a-495f-b99f-3ee712065705.png)

mquandalle:
Yes, this works as a workaround, thanks

a-alhusaini:
@metmarkosaric Is the page in question a liveview? This sort of error can easily occur when the liveview is doing way too much work (Like enumerating a list of users)

Also, I am new to the codebase. Could you link me to the file that renders that page so I can take a look?

metmarkosaric:
this has now been fixed. the default interval is now set to monthly on very long date ranges including those manually selected using custom range and it's no longer possible to choose daily interval on such long date ranges when there's so much data that it would end up freezing the browser

PR #3110, opened 2023-07-05
Fix the interval selection in the all time view

krisbuist:
Previously, the interval was always overwritten with the default value, which is determined based on the age of the site. It should not be overwritten when the interval is explicitly passed by the user

Fixes issue #2982 

### Changes

The way the query is build for the given params has been edited so that the default interval for the "all" period is only set when there is no interval in the passed params

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
 - Fix bug when using multiple [wildcard goal filters](https://github.com/plausible/analytics/pull/3015)
 - Fix a bug where realtime would fail with imported data
 - Fix a bug where the country name was not shown when [filtering through the map](https://github.com/plausible/analytics/issues/3086)
+- Fix [broken interval selection](https://github.com/plausible/analytics/issues/3086) in the all time view
 
 ### Changed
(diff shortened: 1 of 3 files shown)


krisbuist:
Not sure if this will cause issues (again?) like described in #2713 because the daily view of all time can be quite hefty to load.

vinibrsl (approved):
Thank you, @krisbuist! There's still the performance issue with all time + daily interval, but this could happen with custom periods or even the year period. Let's track that in a separate issue and merge your fix. Thanks again! ✨

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
 - Fix bug when using multiple [wildcard goal filters](https://github.com/plausible/analytics/pull/3015)
 - Fix a bug where realtime would fail with imported data
 - Fix a bug where the country name was not shown when [filtering through the map](https://github.com/plausible/analytics/issues/3086)
+- Fix [broken interval selection](https://github.com/plausible/analytics/issues/3086) in the all time view
```
ruslandoga, line 43: This should've probably gone under `Unreleased`

PR #3111, opened 2023-07-06
New private API to return custom prop values

RobertJoonas:
### New private API route - `/:domain/custom-prop-values/:prop_key`

This PR adds a new private API route (used by the React dashboard) to return custom prop values for a given `prop_key`. 

We already have a very similar API route (`"/:domain/property/:prop_name"`) that also returns a custom prop breakdown BUT expects a goal filter. The new API however, is able to return custom props with and without it. Depending on whether there is a goal filter applied, the returned metrics differ:

* with goal filter: `visitors`, `events`, `CR`
* without goal filter: `visitors`, `events`, `%`

### Define two different CR definitions

This PR also makes a clear distinction between two different CR definitions that we are currently using (calling them `A` and `B`). In the future, we want to define CR as a single metric, but this is a bigger project that should be tackled separately from everything else.

### Percentage precision

This PR also increases the precision of the `%` metric to one decimal place. Consequently, changing the return type from `integer` to `float`

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1265,26 +1285,35 @@ defmodule PlausibleWeb.Api.StatsController do
...
+    Enum.map(stat_list, fn stat ->
+      Map.put(stat, :percentage, Float.round(stat[:visitors] / total * 100, 1))
     end)
   end
 
-  defp maybe_add_cr([], _site, _query, _pagination, _key_name, _filter_name), do: []
+  # maybe_add_cr_a will add a CR metric with definition A.
+  #  A: unique_conversions with P / total_unique_visitors
+  # where P is an arbitrary property with an arbitrary value
+  defp maybe_add_cr_a(list, total_unique_visitors, query) do
```
aerosol, line 1301: This is very hard to understand 😅

- the naming should be more clear. `a` and `b` are vague and trying to compare the comments doesn't help in understanding the difference.
- avoid names like `list_*` - perhaps guards/pattern matches are better here, e.g. instead of writing `(list...` try `([_ | _] = breakdown_result...` - it's an idiom for non-empty list

Let's look at the similarities:
- both functions only run if there's "event:goal" filter applied and the breakdown result is not empty
- thus we can drop the `maybe_` prefix, it doesn't help understanding the preconditions

I would consider something like:

```elixir
defp add_visitors_conversion_rate(breakdown_props, query, total_unique_visitors, opts \\ [])

defp add_visitors_conversion_rate([_ | _] = breakdown_result, query, total_unique_visitors, []) when is_map_key(query.filters, "event:goal") do
  Enum.map(breakdown_result, %{&1 | conversion_rate: calculate_cr(total_unique_visitors, prop.visitors)})
end

defp add_visitors_conversion_rate([_ | _] = breakdown_result, query, total_unique_visitors, [_ | _] = opts) when is_map_key(query.filters, "event:goal") do
  pagination = Keyword.fetch!(opts, :pagination)
  only = Keyword.fetch!(opts, :only) # e.g. "visit:source"
  key = Map.fetch!(%{
    "visit:os" => :os,
    # ... can be a module attribute if need be
  }, only)
 
  new_query = 
    query 
    |> Query.put_filter(only, {:member, Enum.map(breakdown_result, & &1.key)}) 
    |> Query.remove_event_filters([:goal, :props])

  sub_breakdown = Stats.breakdown(new_query, only, [:visitors], pagination)

  Enum.map(breakdown_result, fn item ->
    sub_item = Enum.find(sub_breakdown, & &1.key == item.key) 
    %{item | total_visitors: sub_item.visitors, conversion_rate: calculate_cr(sub_item.visitors, item.visitors)} 
  end)
end

defp add_visitors_conversion_rate(breakdown_result, _, _, _) do
  # pokemon catch
  breakdown_result
end
```

I haven't run this code but hopefully it's a bit more concise.

Note that I'm mostly using `map.foo` (and `{Map,Keyword}.fetch!`) instead of `map[:foo]` to crash early on any `nil`s I assume should not be there. This way we'll detect bugs early.

Minor: `===` in elixir is unnecessary, unless you want strict integer vs float comparison.

PS: similar could be done to `maybe_add_percentages`.

aerosol (changes requested):
I had hard time wrapping my head around the cr logic and added a suggestion that hopefully improves readability, not sure. WYT?

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Thanks @aerosol, for the thorough feedback!

> the naming should be more clear. a and b are vague and trying to compare the comments doesn't help in understanding the difference.

Totally agree. Picked my brain hard, but couldn't come up with a reasonable naming that would indicate how the CR is actually being calculated. In fact, I spent several hours trying to understand the nature of these metrics and the reasoning behind those calculations... It's far from perfect, and as mentioned in the PR description, in the future we want to stick to a single CR definition which will make both the code, and the data model easier to understand for both us and the users.

Actually, the reason I went with something as dumb as `a` and `b` is just to make it clear that we're dealing with **two very different definitions** of CR. And I thought comments in code would make the difference clear right away. Thanks for noting that's not the case - added examples to both comments. Let me know what you think, and if that makes it clearer.

My first instinct was also to merge these two CR definitions into a single `add_cr` function, but having learned about the difference, I felt like it made more sense to totally separate them, and even name them differently (in case we need to reference these definitions elsewhere in code, or simply in our dicsussions - makes it easier to understand what we mean).

> Note that I'm mostly using map.foo (and {Map,Keyword}.fetch!) instead of map[:foo] to crash early on any nils I assume should not be there. This way we'll detect bugs early.

> Minor: === in elixir is unnecessary, unless you want strict integer vs float comparison.

> PS: similar could be done to maybe_add_percentages.

> avoid names like list_* - perhaps guards/pattern matches are better here, e.g. instead of writing (list... try ([_ | _] = breakdown_result... - it's an idiom for non-empty list

> thus we can drop the maybe_ prefix, it doesn't help understanding the preconditions

All good points, addressed in https://github.com/plausible/analytics/pull/3111/commits/6da66bbd797fe3e0d3da336bd75889615523ada2

RobertJoonas: @aerosol we reviewed in a call with @ukutaht and he also thought that creating the `a` and `b` abstraction in code is not worth it at this point. The `add_cr_a` function was only called in one place, so we agreed that it makes sense to do the following:

1) remove the `add_cr_a` function and inline the code directly into the `custom_prop_values` controller action function. It's also inlined in the `prop_breakdown` function but that will be removed anyway, once the FF is flipped.
2) rename `add_cr_b` &rarr; `add_cr`

I also removed the comments as it looked like they were not really helpful. https://github.com/plausible/analytics/pull/3111/commits/c1a8ff4515b1c193bd793148b79b7bf10fb66ac2

aerosol (approved):

PR #3112, opened 2023-07-06
Plot revenue metrics on main graph

vinibrsl:
### Changes

This pull request plots revenue metrics on main graph. Previously, revenue metrics were not clickable from top stats. Here's a preview:

Screenshot 2023-07-06 at 21 32 35

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -90,10 +90,13 @@ export default class TopStats extends React.Component {
   }
 
   canMetricBeGraphed(stat) {
-    const isTotalUniqueVisitors = this.props.query.filters.goal && stat.name === 'Unique visitors'
-    const isKnownMetric = Object.keys(METRIC_MAPPING).includes(stat.name)
+    const goalFilteredMetrics = ['Unique visitors', 'Average revenue', 'Total revenue']
 
-    return isKnownMetric && !isTotalUniqueVisitors
+    if (goalFilteredMetrics.includes(stat.name)) {
+      return Boolean(this.props.query.filters.goal)
+    } else {
+      return Object.keys(METRIC_MAPPING).includes(stat.name)
+    }
   }
 
   maybeUpdateMetric(stat) {
--- a/assets/js/dashboard/stats/graph/graph-util.js
+++ b/assets/js/dashboard/stats/graph/graph-util.js
@@ -10,6 +10,8 @@ export const METRIC_MAPPING = {
   'Total visits': 'visits',
   'Bounce rate': 'bounce_rate',
   'Unique conversions': 'conversions',
+  'Average revenue': 'average_revenue',
+  'Total revenue': 'total_revenue',
 }
 
 export const METRIC_LABELS = {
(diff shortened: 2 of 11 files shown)


assets/js/dashboard/stats/graph/top-stats.js
```diff
@@ -90,10 +90,13 @@ export default class TopStats extends React.Component {
   }
 
   canMetricBeGraphed(stat) {
-    const isTotalUniqueVisitors = this.props.query.filters.goal && stat.name === 'Unique visitors'
-    const isKnownMetric = Object.keys(METRIC_MAPPING).includes(stat.name)
+    const goalFilteredMetrics = ['Unique visitors', 'Average revenue', 'Total revenue']
```
RobertJoonas, line 93: This seems to be a typo. `Unique visitors` should not be graphed in the goal filter view. Also, without a goal filter, this currently blocks graphing the `Unique visitors` metric.

I believe on line 93 `Unique visitors` should be replaced with `Unique conversions`, but best to double check

vinibrsl: Good catch! I refactored the function to make it clearer. From what I understood, exceptions are:
- Revenue metrics **should only** be graphed when filtering by goal
- Unique visitors **should not** be graphed when filtering by goal
Changed in 431df232. Let me know what you think.

RobertJoonas: > Revenue metrics should only be graphed when filtering by goal

Is there a situation where we want to display revenue metrics, but disallow graphing them? My understanding is that the condition of *displaying* and *graphing* the metrics should be equivalent. And assuming that, can't we just add both revenue metrics into `METRIC_MAPPING` without having to touch this function at all? 🤔

> Unique visitors should not be graphed when filtering by goal

Yup. It's not graphable in a goal-filtered view because it has nothing to do with the goal filter itself. You'll see the same number without a goal filter, where you can also graph it. Preventing this from being graphed is is just to avoid confusion I believe.

ukutaht (approved):

assets/js/dashboard/stats/graph/top-stats.js
ukutaht: The `Unique Visitors` metric in goal view is a bit confusing, both in code and for customers at the moment. It's special cased all the time to make sure you can't graph it, remove the filter, etc.

@metmarkosaric and I would both prefer to remove that metric, and show this information in the tooltip for `Conversion rate` instead. When hovering on the conversion rate, it could show something like:

```
1% conversion rate
16 conversions / 1.5k total visitors
```

Do you think it would also simplify the code for the top stats section with a goal filter?

vinibrsl: You're right, @RobertJoonas! We actually don't even need to change this `canMetricBeGraphed` function at all for revenue goals. I don't know why I was overthinking this. I reverted the changes in 74ef14b2cf8a3960b96d4e661b7a6c840d490695.

As for unique visitors, I like your idea with the tooltip. However, I don't think it simplifies the code. Top stats displays every metric the backend returns, and for this specific case, we'd need the metric to be returned but displayed only in the tooltip, not as a top stat metric. Do you think we can tackle separately from this change? @ukutaht @RobertJoonas

PR #3114, opened 2023-07-07
Upgrade CI runner

cnkk:


.github/workflows/elixir.yml
```diff
@@ -18,6 +18,9 @@ jobs:
     services:
       postgres:
         image: postgres:${{ matrix.postgres_version }}
+        credentials:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
aerosol, line 22: @cnkk are those necessary we're wondering?

https://github.com/plausible/analytics/pull/3179#issuecomment-1647872131

cnkk: @aerosol [unfortunately yes. ](https://buildjet.com/for-github-actions/docs/guides/docker#docker-rate-limiting-and-authentication)

aerosol: Do you know how can we make those vars optional at least? So that PRs submitted off forks (such as dependabot) have at least the chance to run, even if rate limited?

PR #3121, opened 2023-07-10
refactor LazyLoader and fix bug

RobertJoonas:
### Changes

* Refactors LazyLoader to a function component using https://www.npmjs.com/package/react-intersection-observer
* Fixes a bug where the currently used `LazyLoader` did not call the `onVisible` function on page refresh when being scrolled down on the dashboard (screencast of the bug below). Note that the bug happened for me only on Chrome.

https://github.com/plausible/analytics/assets/56999674/3b9e6b2b-ddc8-47fc-ad98-446dad2cd35d

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/behaviours/conversions.js
+++ b/assets/js/dashboard/stats/behaviours/conversions.js
@@ -40,7 +40,7 @@ export default class Conversions extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (this.props.query !== prevProps.query) {
-      const height = this.htmlNode.current.element.offsetHeight
+      const height = this.htmlNode.current.offsetHeight
       this.setState({ loading: true, goals: null, prevHeight: height })
       this.fetchConversions()
     }
@@ -108,9 +108,11 @@ export default class Conversions extends React.Component {
 
   renderConversions() {
     return (
-      <LazyLoader ref={this.htmlNode} style={{ minHeight: '132px', height: this.state.prevHeight ?? 'auto' }} onVisible={this.onVisible}>
-        {this.renderInner()}
-      </LazyLoader>
+      <div ref={this.htmlNode} style={{ minHeight: '132px', height: this.state.prevHeight ?? 'auto' }} >
+        <LazyLoader onVisible={this.onVisible}>
+          {this.renderInner()}
+        </LazyLoader>
+      </div>
     )
   }
 
--- a/assets/js/dashboard/stats/sources/referrer-list.js
+++ b/assets/js/dashboard/stats/sources/referrer-list.js
@@ -163,8 +163,8 @@ export default class Referrers extends React.Component {
 
   render() {
     return (
-      <div>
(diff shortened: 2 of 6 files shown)


assets/js/dashboard/stats/behaviours/conversions.js
```diff
@@ -40,7 +40,7 @@ export default class Conversions extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (this.props.query !== prevProps.query) {
-      const height = this.htmlNode.current.element.offsetHeight
+      const height = this.htmlNode.current.offsetHeight
```
RobertJoonas, line 43: This is because I put the `ref` (this.htmlNode) on a parent `div` instead. Previously, since the ref was given as a prop to a class component, the actual HTML element was accessed through `current.element`. Not anymore.

assets/js/dashboard/stats/sources/referrer-list.js
```diff
@@ -163,8 +163,8 @@ export default class Referrers extends React.Component {
 
   render() {
     return (
-      <div>
-        <LazyLoader onVisible={this.onVisible} className="flex flex-col flex-grow">
+      <div className="flex flex-col flex-grow">
+        <LazyLoader onVisible={this.onVisible}>
```
ukutaht, line 167: The previous LazyLoader would take the classname as a prop and use it on the div that it renders. This avoids HTML pollution, meaning we avoid unnecessary extra nesting of `div`s in HTML. It's not the biggest problem in the world but a good hygene practice (see https://elisehe.in/2023/03/27/minimal-html-in-design-systems).

Is there anything preventing this new component of using the same strategy to avoid unnecessary nesting? Or is the nesting required for the interaction-observer to work correctly?

RobertJoonas: There was only one place where I added a new `<div>` and it was for the `ref` attribute, which was necessary (because the ref attr cannot be given to a fn component in React). In other places, the parent divs already existed (I'm not sure if needed though), and so I moved the classnames and styles into those divs.

> Is there anything preventing this new component of using the same strategy to avoid unnecessary nesting? Or is the nesting required for the interaction-observer to work correctly?

Nope, we can also apply styles and classnames through to the component itself, and I believe we should be able to remove some of those parent divs as well then

aerosol (approved):
I checked out the PR and tested the funnels panel, looked fine.

assets/js/dashboard/stats/sources/referrer-list.js
RobertJoonas: Looking at the places of where we might want to get rid of extra HTML (referrer-list.js and source-list.js) - they should both be using ListReport in the future anyway, so perhaps let's just leave it as it is for now... Still readable and doesn't look the worst. If we need to avoid HTML pollution in the future, we can add the classnames back to the LazyLoader component. What do you think @ukutaht?

Issue #2991, opened 2023-05-31
Search in Filters is broken


marekzelezny:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

Searching for a value of a property doesn't work. It shows only (none) option. 

It seems there is a problem with XHR props undefined as shown on screenshot.

<img width="654" alt="Snímek obrazovky 2023-05-31 v 12 31 40" src="https://github.com/plausible/analytics/assets/6258792/7e733f3d-b6db-4c2b-b5cc-116285f78794">

### Expected behavior

Get the value matching the query

### Environment

```markdown
- OS: MacOS Ventura
- Browser: Microsoft Edge
- Browser Version: 113.0.1774.57
```

metmarkosaric:
thanks for reporting @marekzelezny! i can reproduce that too. we'll see to fix it before the public release of properties

RobertJoonas:
Closing this as completed in https://github.com/plausible/analytics/pull/3126

PR #3126, opened 2023-07-11
List props frontend

RobertJoonas:
### Changes

_Reviewing commit-by-commit recommended_

* Add a new Properties section into Behaviors - only visible when:
  * The `props` FF is enabled
  * The current site has an `allowed_event_list` field containing at least one property
* Fix a bug https://github.com/plausible/analytics/issues/2991
* Changes the Combobox search-select component (example screencast below):
  * Single-option mode is now "non-clearable" (i.e. once you select a value, you cannot clear the input field anymore)
  * Do not display values in the dropdown that are already picked (the current behavior is to gray those out and make them unclickable)
* Get rid of the `%` sign when it's already displayed in the column header (in Devices, Screen Sizes, Operating Systems, and Properties) 
image

https://github.com/plausible/analytics/assets/56999674/b96f31e7-3d7e-4330-8d10-9ba284ebdff7

### Tests

### Changelog
- [x] This PR does not make a user-facing change YET

### Documentation
- [x] This change does not need a documentation update YET

### Dark mode


assets/js/dashboard/util/realtime-update-timer.js
```diff
@@ -1,4 +1,4 @@
-const THIRTY_SECONDS = 30000
+const THIRTY_SECONDS = 5000
```
vinibrsl, line 1: 😅 oops!

RobertJoonas: 😬Nice catch!

vinibrsl (approved):

PR #3144, opened 2023-07-14
fix(tracker): handle `pageshow` events from bfcache

ryanccn:
### Changes

Adds a `pageshow` event handler to the tracker in order to handle navigations via the [back/forward cache](https://web.dev/bfcache/#how-bfcache-affects-analytics-and-performance-measurement)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -162,6 +162,10 @@
     } else {
       page()
     }
+
+    document.addEventListener('pageshow', (ev) => {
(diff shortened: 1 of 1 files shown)


RobertJoonas:
Sorry for a late reply @ryanccn, and thank you for the contribution 🙏🏼

We've tackled the issue here https://github.com/plausible/analytics/pull/4656. Before releasing it to the customers though, we will do some testing on our own website.

PR #3153, opened 2023-07-18
Refactor: Use ListReport component in Sources

RobertJoonas:
### Changes

This PR uses the new more flexible ListReport component in the following "Source" reports:

* All Sources (the `All` tab)
* All UTMs
* referrer URL drilldown (i.e. when filtered by a source)

Also, this changes the `/referrers` endpoint response format and excludes `total_visitors` which was not used anywhere annymore.

This PR **does not attempt** to use ListReport in Search Terms, because of several differences in the behavior:

* list entries do not act as filter links
* only metric we ever show is visitors
* has to know how to render errors such as "Google account not linked" or "Cannot show search terms with a goal filter"

### Tests
- [x] Automated tests have been ~~added~~ modified 

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/reports/list.js
+++ b/assets/js/dashboard/stats/reports/list.js
@@ -7,6 +7,7 @@ import FadeIn from '../../fade-in'
 import MoreLink from '../more-link'
 import Bar from '../bar'
 import LazyLoader from '../../components/lazy-loader'
+import classNames from 'classnames'
 
 const MAX_ITEMS = 9
 const MIN_HEIGHT = 380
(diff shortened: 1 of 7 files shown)


assets/js/dashboard/stats/reports/list.js
```diff
@@ -15,10 +16,29 @@ const ROW_GAP_HEIGHT = 4
...
+      <Link
+        to={{search: filterQuery.toString()}}
+        onClick={onClick}
+        className={className}
+        >
+        { children }
+      </Link>
+    )
+  } else {
+    return <span class={className}>{ children }</span>
```
aerosol, line 35:
```suggestion
    return <span className={className}>{ children }</span>
```

RobertJoonas: ✅Thanks!

aerosol (approved):
Nice diffstat!

PR #3162, opened 2023-07-19
Expose 9009 for native clickhouse protocol (dev mode)

aerosol:
### Changes

The port we normally expose is HTTP, I'd like to use the native one for development.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ install: ## Run the initial setup
 server: ## Start the web server
 	mix phx.server
 
-CH_FLAGS ?= --detach -p 8123:8123 --ulimit nofile=262144:262144 --name plausible_clickhouse
+CH_FLAGS ?= --detach -p 8123:8123 -p 9009:9000 --ulimit nofile=262144:262144 --name plausible_clickhouse
 
 clickhouse: ## Start a container with a recent version of clickhouse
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ install: ## Run the initial setup
 server: ## Start the web server
 	mix phx.server
 
-CH_FLAGS ?= --detach -p 8123:8123 --ulimit nofile=262144:262144 --name plausible_clickhouse
+CH_FLAGS ?= --detach -p 8123:8123 -p 9009:9000 --ulimit nofile=262144:262144 --name plausible_clickhouse
```
ruslandoga, line 18: 👋

I wonder why not `9000:9000`?

ruslandoga (approved):

Makefile
aerosol: Yeah no reason really apart from me being scared of 9000 being more likely taken by some other service (such as Sentry?)

aerosol: I'm happy to go with 9000 and change it in a different container, if there's a conflict. WYT?

ruslandoga: I think `9000` by default is a bit easier to work with as it wouldn't require passing extra options for `clickhouse client`.

PR #3167, opened 2023-07-20
Props into CSV export

RobertJoonas:
### Changes

This PR starts including a new CSV file in the export zip - `custom_props.csv`. This file will only be included in the zip, if the `props` feature flag is enabled for the user triggering the export.

There's also the existing `prop_breakdown.csv` file which will be kept until the `props` feature is globally enabled. This will be a breaking change.

### Tests

### Changelog
- [x] This PR does not require a changelog update **yet**

### Documentation
- [x] This change does not need a documentation update **yet**

### Dark mode

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -178,7 +178,7 @@ defmodule Plausible.Sites do
     )
   end
 
-  def set_allowed_event_props(site, props) do
+  def set_allowed_event_props!(site, props) do
     Ecto.Changeset.change(site, allowed_event_props: props)
     |> Repo.update!()
   end
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -177,4 +177,9 @@ defmodule Plausible.Sites do
         where: sm.role == :owner
     )
   end
+
+  def set_allowed_event_props(site, props) do
+    Ecto.Changeset.change(site, allowed_event_props: props)
+    |> Repo.update!()
+  end
 end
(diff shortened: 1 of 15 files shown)


lib/plausible/sites.ex
```diff
@@ -177,4 +177,9 @@ defmodule Plausible.Sites do
         where: sm.role == :owner
     )
   end
+
+  def set_allowed_event_props(site, props) do
```
vinibrsl, line 181:
```suggestion
  def set_allowed_event_props!(site, props) do
```

test/plausible_web/controllers/stats_controller_test.exs
```diff
@@ -296,6 +359,35 @@ defmodule PlausibleWeb.StatsControllerTest do
...
+
+      filters = Jason.encode!(%{goal: "Newsletter Signup"})
+      conn = get(conn, "/" <> site.domain <> "/export?period=day&filters=#{filters}")
+
+      {:ok, zip} = :zip.unzip(response(conn, 200), [:memory])
+
+      {_filename, result} =
+        Enum.find(zip, fn {filename, _data} -> filename == 'custom_props.csv' end)
+
+      assert parse_csv(result) == [
```
vinibrsl, line 384: The test function `assert_zip` asserts the contents of the files returned by the API, but does not assert which files were returned. Maybe we can have a follow up PR with that?

vinibrsl (approved):

lib/plausible/sites.ex
RobertJoonas: ✅ https://github.com/plausible/analytics/pull/3167/commits/2ecc763f969c51877b63d871c3e64795e047ccf9

PR #3169, opened 2023-07-20
Add creatable option to ComboBox

vinibrsl:
This commit changes the ComboBox component to allow a `creatable` option. This option enables users to create new options along with choosing existing options.

Preview:

https://github.com/plausible/analytics/assets/5093045/985bf2e5-5b10-42d5-bb2b-e3d08c322c94

### DIFF
--- a/lib/plausible_web/live/components/combo_box.ex
+++ b/lib/plausible_web/live/components/combo_box.ex
@@ -51,7 +51,6 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
   attr(:class, :string, default: "")
   attr(:required, :boolean, default: false)
   attr(:creatable, :boolean, default: false)
-  attr(:input, :string, default: "")
 
   def render(assigns) do
     ~H"""
@@ -102,7 +101,7 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
         suggestions={@suggestions}
         target={@myself}
         creatable={@creatable}
-        input={@input}
+        display_value={@display_value}
       />
     </div>
     """
@@ -136,7 +135,7 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
   attr(:suggest_mod, :atom, required: true)
   attr(:target, :any)
   attr(:creatable, :boolean, required: true)
-  attr(:input, :string, required: true)
+  attr(:display_value, :string, required: true)
 
   def dropdown(assigns) do
     ~H"""
@@ -164,10 +163,10 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
       />
 
       <.option
-        :if={@creatable && String.length(@input) > 0}
+        :if={@creatable && String.length(@display_value) > 0}
         idx={length(@suggestions)}
-        submit_value={@input}
-        display_value={@input}
+        submit_value={@display_value}
+        display_value={@display_value}
         target={@target}
         ref={@ref}
         creatable
@@ -248,7 +247,7 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
 
     socket =
       if socket.assigns[:creatable] do
-        assign(socket, display_value: input, submit_value: input, input: input)
+        assign(socket, display_value: input, submit_value: input)
       else
         socket
       end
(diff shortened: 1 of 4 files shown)


lib/plausible_web/live/components/combo_box.ex
```diff
@@ -48,12 +48,16 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
   attr(:submit_value, :string, default: "")
   attr(:suggest_mod, :atom, required: true)
   attr(:suggestions_limit, :integer)
+  attr(:class, :string, default: "")
+  attr(:required, :boolean, default: false)
```
aerosol, line 52: I think those can be handled with a global - https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-global-attributes ?

lib/plausible_web/live/components/combo_box.ex
```diff
@@ -48,12 +48,16 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
...
+  attr(:creatable, :boolean, default: false)
+  attr(:input, :string, default: "")
```
aerosol, line 54: Could we use `display_value` down the line instead? This feels a bit redundant.

lib/plausible_web/live/components/combo_box.ex
```diff
@@ -217,6 +246,13 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
     input = params[target]
     input_len = input |> String.trim() |> String.length()
 
+    socket =
+      if socket.assigns[:creatable] do
```
aerosol, line 250: Isn't this always present? We could use the dot syntax here.

aerosol (approved):
Nice! Nitpicks inline

aerosol:
Oh and please check if selecting via keyboard works (arrow keys + enter). This has been quirky and is difficult to test.

lib/plausible_web/live/components/combo_box.ex
vinibrsl: I also thought so, but I got `KeyError` in tests. I'm not sure, but I'm guessing this is something with how [`Phoenix.LiveViewTest.render_component/3`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#render_component/3) works. The reason I think that is because I had some problems using global attributes with that macro. `@rest` contained `socket`, `cid` and other unexpected keys.

```
** (KeyError) key :creatable not found in: %{__changed__: %{}, flash: %{}, id: "test-component", myself: %Phoenix.LiveComponent.CID{cid: 1}, options: [{1, "Option 1"}, {2, "Option 2"}, {3, "Option 3"}, {4, "Option 4"}, {5, "Option 5"}, {6, "Option 6"}, {7, "Option 7"}, {8, "Option 8"}, {9, "Option 9"}, {10, "Option 10"}, {11, "Option 11"}, {12, "Option 12"}, {13, "Option 13"}, {14, "Option 14"}, {15, "Option 15"}, {16, "Option 16"}, {17, "Option 17"}, {18, "Option 18"}, {19, "Option 19"}, {20, "Option 20"}], submit_name: "some_submit_name", suggest_mod: PlausibleWeb.Live.Components.ComboBoxTest.SampleView.SampleSuggest, suggestions: [{1, "Option 1"}, {2, "Option 2"}, {3, "Option 3"}, {4, "Option 4"}, {5, "Option 5"}, {6, "Option 6"}, {7, "Option 7"}], suggestions_limit: 7}
```

lib/plausible_web/live/components/combo_box.ex
vinibrsl: I had some issues with tests and global attributes. For some reason `@rest` contained `socket`, `cid` and other unexpected keys.

But in any case, I don't think a global attribute would be that helpful in this case because we have two elements. `class` should be added to the outer `<div>`, but `required` to the `<input>`. Let me know if you think it's worth it.

lib/plausible_web/live/components/combo_box.ex
vinibrsl: Good call! Changed in 1a1c3614f4bca04ca72f80ed5f3000e263bdee30

vinibrsl:
> Oh and please check if selecting via keyboard works (arrow keys + enter). This has been quirky and is difficult to test.

Selecting via keyboard is working, but I also pushed an improvement with scrolling (48330ce1290f871969f5067674875f92dbcc3585). Previously when using the keyboard the browser would scroll even if it had space. Now it only scrolls when needed. Here's the before/after:

| Before                                                                                          | After                                                                                           |
|-------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|
|  |  |

PR #3172, opened 2023-07-21
Refine ComboBox.StaticSearch

vinibrsl:
### Changes

This commit makes static search more strict by rejecting matches with a score less than `0.6`. Here's an example of suggestion that was matching with a `0.5` score that should not be suggested.

```
iex> String.jaro_distance("author", "uaa")
0.5
```

This makes the suggestion list smaller and more reasonable.

Co-authored by: Robert Joonas <robertjoonas16@gmail.com>

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/components/combo_box.ex
+++ b/lib/plausible_web/live/components/combo_box.ex
@@ -252,16 +252,15 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
         socket
       end
 
-    if input_len > 0 do
-      suggestions =
-        input
-        |> suggest_mod.suggest(options)
-        |> Enum.take(suggestions_limit(socket.assigns))
+    suggestions =
+      if input_len > 0 do
+        suggest_mod.suggest(input, options)
+      else
+        options
+      end
+      |> Enum.take(suggestions_limit(socket.assigns))
 
-      {:noreply, assign(socket, %{suggestions: suggestions})}
-    else
-      {:noreply, assign(socket, %{suggestions: options})}
-    end
+    {:noreply, assign(socket, %{suggestions: suggestions})}
   end
 
   defp do_select(socket, submit_value, display_value) do
--- a/lib/plausible_web/live/components/combo_box.ex
+++ b/lib/plausible_web/live/components/combo_box.ex
@@ -260,7 +260,7 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
 
       {:noreply, assign(socket, %{suggestions: suggestions})}
     else
-      {:noreply, socket}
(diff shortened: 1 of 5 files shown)


lib/plausible_web/live/components/combo_box.ex
```diff
@@ -260,7 +260,7 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
 
       {:noreply, assign(socket, %{suggestions: suggestions})}
     else
-      {:noreply, socket}
+      {:noreply, assign(socket, %{suggestions: options})}
```
vinibrsl, line 263: This fixes a bug where searching and then deleting the search input would not show all options again. Related test: https://github.com/plausible/analytics/pull/3172/files#diff-390c19e2a314b162d56c997680346e7c3d5cc72d6a5dfe393b8bfe126e320a7c

aerosol: Hm, nice catch. Maybe limit the suggestions as we do it in initial render: https://github.com/plausible/analytics/blob/c13df05ea956738fe59061b1d659e5a27c69676a/lib/plausible_web/live/components/combo_box.ex#L37

lib/plausible_web/live/components/combo_box/static_search.ex
```diff
@@ -11,25 +11,27 @@ defmodule PlausibleWeb.Live.Components.ComboBox.StaticSearch do
...
-      :desc
-    )
+    |> Enum.map(fn {_, value} = option -> {option, weight(value, input)} end)
+    |> Enum.reject(fn {_option, weight} -> weight < 0.6 end)
+    |> Enum.sort_by(fn {_option, weight} -> weight end, :desc)
+    |> Enum.map(fn {option, _weight} -> option end)
+  end
+
+  defp weight(value, input) do
+    cond do
```
aerosol, line 22: I think `case` is generally preferred 

```elixir
case {value, input} do
  {^value, ^value} -> 3
  {_, _} when byte_size(input) > byte_size(value) -> 0
  {_, _} -> ...
end
```

I'm not sure why I was explicitly calling `to_string()` here. Perhaps it was some remnant from trying to use `String.Chars` protocol implemented for the `Goal` struct.

I think we can safely assume that `input` is always a string and there's no need for this casting anymore (clause three).

aerosol (approved):

lib/plausible_web/live/components/combo_box.ex
vinibrsl: Good call! Changed in b00ba1d6ebc3f693981230b451616130cdde2440

aerosol (approved):

PR #3180, opened 2023-07-24
inspect request in sentry context

ruslandoga:
### Changes

Fixes https://github.com/plausible/analytics/discussions/3166

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/external_controller.ex
+++ b/lib/plausible_web/controllers/api/external_controller.ex
@@ -12,7 +12,7 @@ defmodule PlausibleWeb.Api.ExternalController do
 
   def event(conn, _params) do
     with {:ok, request} <- Ingestion.Request.build(conn),
-         _ <- Sentry.Context.set_extra_context(%{request: request}) do
+         _ <- Sentry.Context.set_extra_context(%{request: inspect(request)}) do
       case Ingestion.Event.build_and_buffer(request) do
         {:ok, %{dropped: [], buffered: _buffered}} ->
           conn


lib/plausible_web/controllers/api/external_controller.ex
```diff
@@ -12,7 +12,7 @@ defmodule PlausibleWeb.Api.ExternalController do
 
   def event(conn, _params) do
     with {:ok, request} <- Ingestion.Request.build(conn),
-         _ <- Sentry.Context.set_extra_context(%{request: request}) do
+         _ <- Sentry.Context.set_extra_context(%{request: inspect(request)}) do
```
ruslandoga, line 15: Nitpick: we can use `_ = Sentry.Context.set_extra_context(%{request: inspect(request)})` here since the result doesn't matter. And then since `with` has only one case in pipeline and `else` is handled, we can switch to `case` to make it clearer :)

aerosol: Agreed, `case` fits here much better. Would you like to include that change?

ruslandoga: I'd prefer to do it in a separate PR to make fix fix clear to anyone who checks the changelog :)

ruslandoga: Oops, I forgot to update changelog!

aerosol:
Welp, looks like we've not been getting much sentry reports.
However the requests are filtered:

image

aerosol:
Awesome, while at it could you make sure this: https://github.com/plausible/analytics/blob/master/lib/plausible/ingestion/event.ex#L376 is fingerprinted in Sentry.EventFilter please? There's quite a lot of it suddenly :)

PR #3188, opened 2023-07-25
remove dockerhub credentials from pull

cnkk:
I previously introduced using the dockerhub credentials when pulling images from dockerhub. But this leads to problems when creating PRs from a fork.
And it is not so clear if we really need to use the dockerhub creds, as it seems that there were no problems before.


cnkk:
@aerosol @ruslandoga
I removed the usage of dockerhub creds. The rate limiting was probably only an issue while I was testing the different GitHub runner types.
(I also totally missed the PR issue for forks)

Let's see if this is going to be an issue again.

PR #3191, opened 2023-07-25
Create props settings page

vinibrsl:
### Changes

This pull request adds a new settings section for props. Here's a preview:

https://github.com/plausible/analytics/assets/5093045/9d823d75-8879-4d42-a8c1-a16a75dbb2b6

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/components/combo_box.ex
```diff
@@ -58,10 +58,11 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
       id={"input-picker-main-#{@id}"}
       class={@class}
       x-data={"window.suggestionsDropdown('#{@id}')"}
-      x-on:keydown.arrow-up="focusPrev"
-      x-on:keydown.arrow-down="focusNext"
-      x-on:keydown.enter="select()"
+      x-on:keydown.arrow-up.prevent="focusPrev"
+      x-on:keydown.arrow-down.prevent="focusNext"
+      x-on:keydown.enter.prevent="select"
```
vinibrsl, line 63: The browser default behaviour on a `input[type="text"]` field is to move the cursor from the start to the end of the line when pressing arrow up/down. I added `.prevent` to bypass this - arrow down should open up the dropdown.

priv/repo/migrations/20230724131709_change_allowed_event_props_type.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Plausible.Repo.Migrations.ChangeAllowedEventPropsType do
+  use Ecto.Migration
+
+  def change do
+    alter table("sites") do
+      modify :allowed_event_props, {:array, :"varchar(300)"},
```
vinibrsl, line 6: 300 is the same limit applied to prop keys during ingestion.

lib/plausible_web/live/components/combo_box.ex
aerosol: Nice!

priv/repo/migrations/20230724131709_change_allowed_event_props_type.exs
aerosol, line 6: An attempt of adding long-ish property crashes the connection. And could be truncated perhaps to fit in view.

image

lib/plausible_web/templates/site/settings_props.html.heex
```diff
@@ -0,0 +1,53 @@
+<section class="shadow bg-white dark:bg-gray-800 sm:rounded-md sm:overflow-hidden py-6 px-4 sm:p-6">
+  <header class="w-full flex">
+    <span class="flex-1">
+      <h1 class="text-lg leading-6 font-medium text-gray-900 dark:text-gray-100">
+        Properties
+      </h1>
+
+      <h2 class="mt-1 text-sm leading-5 text-gray-500 dark:text-gray-200">
+        Attach custom properties to events
```
aerosol, line 9: It's probably just me, feel free to disregard. It's confusing, because here I'm supposed to "attach properties to events" but then I can "import properties from existing events". So they were already attached?

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+      <form id="props-form" phx-submit="allow" class="flex space-x-2">
+        <.live_component
+          id={:prop_input}
+          submit_name="prop"
+          class="flex-1"
+          module={ComboBox}
+          suggest_mod={ComboBox.StaticSearch}
+          options={build_options(@suggestions, @allowed_event_props)}
+          required
+          creatable
```
aerosol, line 44: image

Looks like it's possible to add empty props

lib/plausible_web/views/layout_view.ex
```diff
@@ -30,6 +30,9 @@ defmodule PlausibleWeb.LayoutView do
       if Plausible.Funnels.enabled_for?(conn.assigns[:current_user]) do
         [key: "Funnels", value: "funnels"]
       end,
+      if Plausible.Props.enabled_for?(conn.assigns[:current_user]) do
+        [key: "Properties", value: "properties"]
```
aerosol, line 34: We should totally put some emoji on the newly released items 🎁

:D

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+  def handle_event("disallow", %{"prop" => prop}, socket) do
+    {:ok, site} =
+      socket.assigns.current_user_id
+      |> get_site(socket.assigns.domain)
+      |> Plausible.Props.disallow(prop)
+
+    {:noreply, assign(socket, allowed_event_props: site.allowed_event_props)}
+  end
+
+  def handle_event("auto-import", _params, socket) do
```
aerosol, line 126: Have you checked the following scenarios:
  - auto-import from a set of over 300 properties (should there be any feedback that you've reached the limit then?)
  - add manually/auto-import 1 property, then gather more events and auto-import 300
 
?

aerosol: image

aerosol (changes requested):
Looks nice! I found some edge cases - comments inline

lib/plausible/ingestion/request.ex
```diff
@@ -200,8 +200,8 @@ defmodule Plausible.Ingestion.Request do
     |> validate_props()
   end
 
-  @max_prop_key_length 300
-  @max_prop_value_length 2000
+  @max_prop_key_length Plausible.Props.max_prop_key_length()
+  @max_prop_value_length Plausible.Props.max_prop_value_length()
```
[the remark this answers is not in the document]
vinibrsl: Yes! `lib/plausible/ingestion/request.ex` is recompiled when changing `lib/plausible/props.ex`.

lib/plausible_web/templates/site/settings_props.html.heex
```diff
@@ -0,0 +1,53 @@
...
+        href="https://plausible.io/docs/custom-pageview-props"
+        target="_blank"
+        rel="noreferrer"
+      >our docs</.link>.
+    </PlausibleWeb.Components.Generic.notice>
+
+    <%= live_render(@conn, PlausibleWeb.Live.PropsSettings,
+      id: "props-form",
+      session: %{"site_id" => @site.id, "domain" => @site.domain},
+      router: PlausibleWeb.Router
```
aerosol, line 50: This option does nothing. I'll remove it from funnel's live_render too.

lib/plausible/props.ex
```diff
@@ -0,0 +1,97 @@
...
+
+    site
+    |> changeset(new_props)
+    |> Plausible.Repo.update()
+  end
+
+  @spec disallow(Plausible.Site.t(), prop()) ::
+          {:ok, Plausible.Site.t()} | {:error, Ecto.Changeset.t()}
+  def disallow(site, prop) do
+    allowed_event_props = site.allowed_event_props || []
```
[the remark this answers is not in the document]
vinibrsl: I initially thought of this, but we still need to make distinction between empty lists and nils. For compatibility, there are dashboard queries that keep the "old" custom props behavior when `nil`, e.g. https://github.com/plausible/analytics/blob/master/lib/plausible/stats/custom_props.ex#L48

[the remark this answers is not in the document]
aerosol: Thanks for clarification.

lib/plausible_web/templates/site/settings_props.html.heex
vinibrsl: Removed in 285645ad396e608a75dfe4621c6b36df73ce3830

lib/plausible/props.ex
```diff
@@ -0,0 +1,97 @@
...
+      from uk in subquery(unnested_keys),
+        where: uk.key not in ^allowed_event_props,
+        group_by: uk.key,
+        select: uk.key,
+        order_by: {:desc, count(uk.key)},
+        limit: ^limit
+    )
+  end
+
+  def enabled_for?(%Plausible.Auth.User{} = user) do
```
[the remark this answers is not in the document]
vinibrsl: Good call! Changed in 587030b8

lib/plausible_web/live/props_settings.ex
vinibrsl: Oops! Should be fixed in 1d7bf7dc

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+    {:ok, site} =
+      socket.assigns.current_user_id
+      |> get_site(socket.assigns.domain)
```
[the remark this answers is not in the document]
vinibrsl: Good call! Changed in c6e15a43.

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+        <button id="allow" type="submit" class="button">+ Add</button>
+      </form>
+
+      <button
+        :if={length(@suggestions) > 0}
+        title="Use this to import any existing properties from your past events into your settings. This allows you to set up properties without having to manually enter each item."
+        class="mt-1 text-sm hover:underline text-indigo-600 dark:text-indigo-400"
+        phx-click="auto-import"
+      >
+        Or auto-import properties from your events
```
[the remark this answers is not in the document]
vinibrsl: Good catch! eec4c777 fixes the bug and adds a test case

priv/repo/migrations/20230724131709_change_allowed_event_props_type.exs
vinibrsl: Good catch! 74e3f7906a28cb0e19f9945f137a7d8cd9541745 limits the dropdown width to the input element.

https://github.com/plausible/analytics/assets/5093045/5bac8cc2-4432-4ad7-9b45-db3a65416546

lib/plausible_web/live/components/combo_box.ex
```diff
@@ -58,10 +58,11 @@ defmodule PlausibleWeb.Live.Components.ComboBox do
       id={"input-picker-main-#{@id}"}
       class={@class}
       x-data={"window.suggestionsDropdown('#{@id}')"}
-      x-on:keydown.arrow-up="focusPrev"
-      x-on:keydown.arrow-down="focusNext"
-      x-on:keydown.enter="select()"
+      x-on:keydown.arrow-up.prevent="focusPrev"
```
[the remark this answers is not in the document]
vinibrsl: Good call! I also think the first <kbd>Enter</kbd> should select from the dropdown, while the second <kbd>Enter</kbd> should submit the form. I initially tried doing this, but I had problems with the page focus, so I decided to go with the "easiest but not terrible" (tm) solution which is <kbd>Enter</kbd> interacts with the combo box dropdown only. I still want to try making this work, but do you think it's okay to file it as improvement for later?

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+              <span class="flex-1 truncate font-medium text-sm text-gray-800 dark:text-gray-200">
+                <%= prop %>
+              </span>
+              <button
+                phx-click="disallow"
+                phx-value-prop={prop}
+                class="w-4 h-4 text-red-600 hover:text-red-700"
+                aria-label={"Remove #{prop} property"}
+              >
+                <svg
```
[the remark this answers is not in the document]
vinibrsl: This button permanently removes the prop from the allow list. Removed props often reappear in the suggestions list because there are events with that prop name on ClickHouse. Do you think adding a confirmation prompt is enough in this case?

lib/plausible_web/live/components/combo_box.ex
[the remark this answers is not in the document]
aerosol: > but do you think it's okay to file it as improvement for later?

Sure thing! Maybe check with others too, I'd consider removing the "Add" button and saving the choice upon selection.

lib/plausible_web/live/props_settings.ex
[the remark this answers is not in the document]
aerosol: > Do you think adding a confirmation prompt is enough in this case?

Or a tooltip? I don't have a strong opinion on that one, just wanted to flag it for your consideration.

priv/repo/migrations/20230724131709_change_allowed_event_props_type.exs
aerosol: Nice!

lib/plausible_web/live/props_settings.ex
```diff
@@ -0,0 +1,153 @@
...
+            No properties configured for this site yet
+          </p>
+        <% end %>
+      </div>
+    </div>
+    """
+  end
+
+  def handle_event("allow", %{"prop" => prop}, socket) do
+    {:ok, site} = Plausible.Props.allow(socket.assigns.site, prop)
```
aerosol, line 114: Cool. I think what's left is handling the changeset error cases. 

I'm also wondering is there a reason for internal interfaces using allow/disallow terminology but no similar phrasing in the UI. Was this intentional?

lib/plausible_web/live/props_settings.ex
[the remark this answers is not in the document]
vinibrsl: What do you think, @RobertJoonas?

lib/plausible_web/live/props_settings.ex
vinibrsl: > Cool. I think what's left is handling the changeset error cases.

Changeset error cases added in 814ce44.

> I'm also wondering is there a reason for internal interfaces using allow/disallow terminology but no similar phrasing in the UI. Was this intentional?

Thanks for pointing that out! When I started working on this, the `sites.allowed_event_props` field was already in the database, so I decided to follow its terminology: allowed -> allow/disallow.

However, when designing the UI, I felt those terms might be confusing for our users. Instead of focusing on what’s allowed or not, I wanted users to understand the actual workflow: users first create properties in the dashboard using the settings page, and then send them via the API. If they don't, the properties won't appear on the dashboard. And just to clarify, this allow list only affects ClickHouse queries and not the ingestion process as we don't want to drop any data.

lib/plausible_web/live/props_settings.ex
vinibrsl: Good catch! 814ce440f3b811b45e57265a030d515fd53d2521 adds changeset validations and a test case covering this.

lib/plausible_web/templates/site/settings_props.html.heex
vinibrsl: I see what you mean and I also think it's not the best subtitle. Do you think "Track custom properties from events" is a better name?

lib/plausible_web/live/props_settings.ex
aerosol: Oh that makes sense, thank you. Would you mind leaving that explanation somewhere in the code comments? I think this is pretty much the context I was missing.

lib/plausible_web/templates/site/settings_props.html.heex
aerosol: > "Track custom properties from events"

Yeah that sounds much better IMHO. Paging @metmarkosaric

lib/plausible_web/live/props_settings.ex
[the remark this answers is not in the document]
RobertJoonas: > What do you think, @RobertJoonas?

Since configuring props is only a UI thing and it doesn't affect any actual data, I think confirmation popups would be annoying and redundant. Also, personally I don't see much value in tooltips.

But I do agree with what @aerosol said in another comment here:

> It's probably just me, feel free to disregard. It's confusing, because here I'm supposed to "attach properties to events" but then I can "import properties from existing events". So they were already attached?

I feel like "Attaching" custom properties to events is a separate and a much more complex part of this feature. To start sending props to Plausible, people need to dig deeper into docs.

Instead of "_Attach custom properties to events_", how about something like:

"_You can attach custom properties to all your events to analyze metrics we don't track automatically. Check out our documentation about sending custom properties to Plausible. In order for the properties to show up on your dashboard, you need to explicitly add them in the below menu first._"

Also, what do you think about changing the auto-import link to something like:

Already sent custom properties to Plausible? [Import from existing events](https://dontclick.me/

aerosol (approved):
Nice work! 👏

I'm only wondering if there should be a user interface to "remove all", in case people set up many props and want to start from scratch.

I was trying to do that from the shell, to test the UI, and it took me a while to realise that `disallow` doesn't accept `prop_or_props` (it silently fell back to no change). 

Maybe something for another iteration.

lib/plausible_web/live/props_settings.ex
RobertJoonas: > I'm also wondering is there a reason for internal interfaces using allow/disallow terminology but no similar phrasing in the UI. Was this intentional?

Good point... I think the reason I went with the `allowed_event_props` field name was that we were working on "filtering out garbage props" at the time. Since I believe this field is not used by anyone else other than ourselves, now is probably a good time to change the name. How about just `site.event_props`?

RobertJoonas: Although from the backend perspective it does make sense that the props are "allowed" to show up on the dashboard. Dunno, it's fine by me to keep the current name too...

lib/plausible_web/templates/site/settings_props.html.heex
RobertJoonas: Same convo going on in two places, so dropping [a link](https://github.com/plausible/analytics/pull/3191#discussion_r1276121008) here to my other comment in an above discussion (/cc @metmarkosaric).

Do you guys think a text like:

> You can attach custom properties to all your events to analyze metrics we don't track automatically. Check out our documentation about sending custom properties to Plausible. In order for the properties to show up on your dashboard, you need to explicitly add them in the below menu first.

would be too long?

lib/plausible_web/live/props_settings.ex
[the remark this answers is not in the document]
metmarkosaric: @RobertJoonas how about this for the subheading:

"Attach custom properties when sending a pageview or an event to create custom metrics."

we do something like this in goals section and it's worth adding here too to minimize confusion:

"Custom properties are not tracked by default - you have to configure them on your site to be sent to Plausible. See examples and learn more in our docs."

about the auto-import: does it import any property already sent to us both on the event and pageview level or only custom event level? I think best not to talk about it as "import" as it confuses things. perhaps this could work for it:

"Already sending custom properties? Click to show all existing properties"

> Since configuring props is only a UI thing and it doesn't affect any actual data, I think confirmation popups for removing the props from the list would be annoying and redundant. Also, personally I don't see much value in tooltips.

I see we already do a confirmation in the goals section when trying to delete a goal so best to keep it consistent with what we already do

it's always a good idea to keep the user interface as close as possible and consistent with what we already do other relevant places. in this case the new props settings section basically has the same purpose as the existing goals settings section. both these sections do the same for the user: they allow them to configure a specific section of the dashboard and add or remove specific entries from being shown in the dashboard. so the more consistent they are the easier it is to understand and use them

lib/plausible_web/templates/site/settings_props.html.heex
metmarkosaric: @RobertJoonas I responded to the other comment here https://github.com/plausible/analytics/pull/3191#discussion_r1276252588

PR #3196, opened 2023-07-26
Props details view

RobertJoonas:
### Changes

This PR adds a "details" view for custom property value breakdown.

https://github.com/plausible/analytics/assets/56999674/fdeca9f6-26fb-4898-bf3d-4436e923c009

Also fixed two bugs that surfaced along the way:

* Fix percentage metric calculation (make it account for pagination)
* Stop returning (none) values after the first page of results in prop-breakdown (documented in changelog)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -1194,12 +1194,12 @@ defmodule PlausibleWeb.Api.StatsController do
   end
 
   defp breakdown_custom_prop_values(site, %{"prop_key" => prop_key} = params) do
-    query = Query.from(site, params) |> Filters.add_prefix()
+    query =
+      Query.from(site, params)
+      |> Filters.add_prefix()
+      |> Map.put(:include_imported, false)
+
     pagination = parse_pagination(params)
-
-    total_q = Query.remove_event_filters(query, [:goal, :props])
-
-    %{:visitors => %{value: total_unique_visitors}} = Stats.aggregate(site, total_q, [:visitors])
 
     prefixed_prop = "event:props:" <> prop_key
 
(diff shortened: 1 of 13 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1322,20 +1322,18 @@ defmodule PlausibleWeb.Api.StatsController do
...
 
-  defp add_percentages([_ | _] = breakdown_result, query)
+  defp add_percentages([_ | _] = breakdown_result, site, query)
        when not is_map_key(query.filters, "event:goal") do
-    total = Enum.reduce(breakdown_result, 0, fn %{visitors: count}, total -> total + count end)
-    do_add_percentages(breakdown_result, total)
-  end
-
-  defp add_percentages(breakdown_result, _), do: breakdown_result
+    %{visitors: %{value: total_visitors}} = Stats.aggregate(site, query, [:visitors])
```
RobertJoonas, line 1327: Instead of summing up the breakdown results, I'm using this call to `Stats.aggregate` here to fetch the total number of visitors with the same `query` struct that's used for breaking down the results.

This means that imported data is also included in `total_visitors`, which I think is expected because `browsers`, `locations`, `OSs`, etc,  are also imported from GA.

However, custom props are not - so it would probably make sense to exclude `imported_visitors` from the total there. On the other hand, I like the fact that the metric can be easily explained: calculated as `A / B`, where:

A: the number of visitors that completed an event with this property,
B: total `unique_visitors` - i.e. the number also showing in the Top Stats (which also includes imported data)

vinibrsl (approved):
Looks good to me!

PR #3209, opened 2023-07-29
Revenue props

RobertJoonas:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/reports/list.js
```diff
@@ -139,6 +140,20 @@ export default function ListReport(props) {
     return () => { document.removeEventListener('tick', fetchData) }
   }, [props.keyLabel, props.query, visible]);
 
+  // returns a filtered `metrics` list. Since currently, the backend can return different
+  // metrics based on filters and existing data, this function validates that the metrics
+  // we want to display are actually there in the API response.
+  function validatedMetrics() {
```
vinibrsl, line 146: I think validate gives me more the idea of verifying user input, rather than filtering a list based on the API output. What do you think of renaming this function to something like `filterMetricsFromApi` or `getAvailableMetrics`?

assets/js/dashboard/stats/reports/list.js
```diff
@@ -139,6 +140,20 @@ export default function ListReport(props) {
...
+    return metrics.filter((metric) => {
+      return state.list.some((listItem) => listItem[metric.name] != null)
+    })
+  }
+
+  function hiddenOnMobileClass(i) {
+    if (i > MOBILE_MAX_METRIC_COLS) { return 'hidden md:block'}
+    return ''
```
vinibrsl, line 154: Feel free to ignore, but I think the early return in this case makes it harder to read.

```suggestion
    if (i > MOBILE_MAX_METRIC_COLS) {
      return 'hidden md:block'
    } else {
      return ''
    }
```

You could also classNames and format the whole class name (including `text-right`):

```js
return classNames('text-right', {'hidden md:block': i > MOBILE_MAX_METRIC_COLS})
```

assets/js/dashboard/stats/reports/list.js
```diff
@@ -159,8 +174,19 @@ export default function ListReport(props) {
...
+      return (
+        <div
+          key={metric.name}
+          className={`text-right ${hiddenOnMobileClass(i)}`}
+          style={{minWidth: colMinWidth}}
+        >
+            { metricLabelFor(metric, props.query) }
+        </div>
+      )
     })
```
vinibrsl, line 190: I believe the second element of the map inner function is the index. In this case you wouldn't need to declare `i` or that extra `i++`

```suggestion
    const metricLabels = validatedMetrics().map((metric, i) => {
      return (
        <div
          key={metric.name}
          className={`text-right ${hiddenOnMobileClass(i)}`}
          style={{minWidth: colMinWidth}}
        >
            { metricLabelFor(metric, props.query) }
        </div>
      )
    })
```

assets/js/dashboard/stats/reports/list.js
```diff
@@ -241,9 +267,15 @@ export default function ListReport(props) {
   }
 
   function renderMetricValuesFor(listItem) {
-    return metrics.map((metric) => {
+    let i = 0
```
vinibrsl, line 270: Same suggestion of removing that extra `i`, and using the `map` second argument.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1206,18 +1206,29 @@ defmodule PlausibleWeb.Api.StatsController do
   end
 
   defp breakdown_custom_prop_values(site, %{"prop_key" => prop_key} = params) do
+    site = Plausible.Repo.preload(site, :goals)
```
vinibrsl, line 1209: I think we can remove this preload here. The breakdown function does that in a separate query internally.

```suggestion
```

assets/js/dashboard/stats/reports/list.js
RobertJoonas: Ah, cool! Good to know, but in the end I had to do it differently, because I realised we have to decide per metric, which ones to show or not. Currently, in the Goal Conversions section, on mobile view we only show two metrics - `Uniques` (`visitors`) and `CR`. And since we want to remove the `Total` (`events`) on mobile, we cannot rely on the order anymore. https://github.com/plausible/analytics/pull/3209/commits/96c11c5417afed6f478fcc2a1a148fb4b122986a

RobertJoonas: That actually feels like a better and more explicit way of handling this problem, so I think we're good :)

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: ✅ https://github.com/plausible/analytics/pull/3209/commits/11e63332c50e494f0cf9c0158e329498e3033047

assets/js/dashboard/stats/reports/list.js
RobertJoonas: Sure, a standard if-else block probably reads easier indeed (✅ in https://github.com/plausible/analytics/pull/3209/commits/8ba23b25779e12388225c851c9ad44c7f6e0e984). But I'd prefer to keep it in a separate function that describes the problem that we're solving so that it's clear that this logic is required in two places (column headers and column body).

```js
classNames('text-right', {'hidden md:block': i > MOBILE_MAX_METRIC_COLS})
```

would be nice, but I can't think of a good function name in this scenario 😃

vinibrsl (approved):

PR #3212, opened 2023-07-31
Fix pageviews filtered by page breakdown

aerosol:
### Changes

:warning: I'm not sure if this is the correct solution. Tests pass (for scenarios with less than `SAMPLE` events though), but I could use a review from someone familiar with the query building mechanics.

In my understanding, the problem was that, when filtering by page (so `events_v2` property) and asking for `pageviews` during the session breakdown, the latter was taken from `sessions_v2` table, thus bypassing the filter applied to the joined table.

Because we've already had a query joining the events (built via `filter_converted_sessions` function), we're now also calculating the sum of event pageviews, if and only if, the metric we're trying to retrieve is `pageviews`. I tried to mimic the sampling seen elsewhere, but I'm not quite sure if this has been done correctly.

I couldn't find a way to avoid some gymnastics with it -- the result is stored in an "internal" property, later on, conditionally swapped and removed from the results -- not to clash with `pageviews` taken from sessions.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/util.ex
```diff
@@ -6,21 +6,26 @@ defmodule Plausible.Stats.Util do
...
-  is needed to calculate these from imported data. This function removes that metric
-  from all entries in the results list.
+  Drops all the atom keys starting with two underscores.
+
+  Currently private fields are:
+    * `__internal_visits` is fetched when querying bounce rate and visit duration, as it
+       is needed to calculate these from imported data. This function removes that metric
+       from all entries in the results list.
+    * `__event_pageviews` is calculated for sessions when it's the event pageviews
+      (and not session pageviews) that need to be summed in breakdowns.
```
RobertJoonas: What do you think about following the same naming convention and prefix internally used metrics with `__internal`? Also, maybe worth giving a bit more context on what the metrics are internally used for.

```
* `__internal_events` is used to count the events from the `events`
  table in queries that are breaking down on session level properties
  and filtering by event level properties at the same time. In this case,
  we cannot use the denormalised `events` and `pageviews` fields
  from the sessions table, since they cannot account for event level
  filters.
* `__internal_pageviews` is used for the exact same purpose as
  internal_events, but for `pageviews`.
```

RobertJoonas: What do you think about following the same naming convention and prefix internally used metrics with `__internal`? Also, maybe worth giving a bit more context on what the metrics are internally used for. Here's what I'm thinking:

```
* `__internal_events` is used to count the events from the `events`
  table in queries that are breaking down on session level properties
  and filtering by event level properties at the same time. In this case,
  we cannot use the denormalised `events` and `pageviews` fields
  from the sessions table, since they cannot account for event level
  filters.
* `__internal_pageviews` is used for the exact same purpose as
  internal_events, but for `pageviews`.
```

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
```diff
@@ -1943,4 +1943,89 @@ defmodule PlausibleWeb.Api.ExternalStatsController.BreakdownTest do
...
+          )
+        ])
+
+        conn =
+          get(conn, "/api/v1/stats/breakdown", %{
+            "site_id" => site.domain,
+            "period" => "day",
+            "date" => "2021-01-01",
+            "property" => "visit:#{unquote(property)}",
+            "metrics" => "pageviews",
```
RobertJoonas, line 2015: I believe the problem that this PR addresses also applies for the `events` metric. For instance you can try changing this line to `"metrics" => "events"` and line 2024 to `"events" => 2` - the test will still fail.

lib/plausible/stats/base.ex
```diff
@@ -464,20 +472,42 @@ defmodule Plausible.Stats.Base do
     end
   end
 
-  def filter_converted_sessions(db_query, site, query) do
+  def filter_converted_sessions(db_query, site, query, count_event_pageviews \\ false) do
```
RobertJoonas: How about `count_event_pageviews` -> `count_event_metrics` here?

lib/plausible/stats/base.ex
```diff
@@ -464,20 +472,42 @@ defmodule Plausible.Stats.Base do
...
     if Query.has_event_filters?(query) do
-      converted_sessions =
-        from(e in query_events(site, query),
-          select: %{
-            session_id: fragment("DISTINCT ?", e.session_id),
-            _sample_factor: fragment("_sample_factor")
+      if count_event_pageviews do
+        converted_sessions =
```
RobertJoonas, line 478: The name `converted_sessions` doesn't describe the problem we're solving very well here. How about `event_filtered_sessions` instead?

RobertJoonas (commented):
Wow, this stuff is really complicated indeed...

The overall approach looks good to me. Nice work putting this together @aerosol! But I believe there are ways to make this a bit more understandable in code, and we also need to address the same problem for the `events` metric. Comments inline

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
aerosol: Oh that's a great catch!

lib/plausible/stats/base.ex
aerosol: Very nice!

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
aerosol: This is a bit puzzling. What should "events" count?

aerosol: Just total count(*) without any conditions on event.name, right?

aerosol:
@RobertJoonas thanks for a thorough review, addressed all your comments, I think 😅

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
aerosol: https://github.com/plausible/analytics/pull/3212/commits/bb350472618d06e385c54a6e7a86f0e0dd4df93e

lib/plausible/stats/base.ex
aerosol: https://github.com/plausible/analytics/pull/3212/commits/a227d519dd0070a1c592cdba0084a1a7d48b48fc

lib/plausible/stats/util.ex
aerosol: done

PR #3220, opened 2023-08-02
[pg migration] Delete account with multiple subscriptions

aerosol:
### Changes

There were some cascading deletes missing in postgres schema, this PR takes care of that via a migration. Additionally, all subscription preloads are now done through an unified interface to minimize the risk of preloading a subscription that's not latest (explicit ordering is provided).

BTW CRM user deletion is fixed too.

I think it's fine to deploy this migration with the code in one shot.

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl (approved):
Thanks for this! Looks good.

PR #3226, opened 2023-08-03
UI tweaks to Properties and Conversions

RobertJoonas:
### Changes

* Add special handling for `url` breakdown when filtering on `404`, `File Download` or `Outbound Link: Click` goals. This will rename the Goal Conversions section accordingly, and display a breakdown by the url. It also links to a Modal view where more results can be seen.
* For reports showing the details view with the exact same data already displayed on the dashboard list (i.e. `Conversions`, `Properties`, and all the three new Goal `url` breakdown sections), rename `DETAILS` -> `MORE` and use a different icon. Also, the `MORE` link will not be displayed when there are less than 9 entries.

### Tests

### Changelog
- [x] not yet

### Documentation
- [x] not yet

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/behaviours/goal-conversions.js
+++ b/assets/js/dashboard/stats/behaviours/goal-conversions.js
@@ -5,22 +5,20 @@ import { CR_METRIC } from "../reports/metrics"
 import * as url from "../../util/url"
 import * as api from "../../api"
 
+const SPECIAL_GOALS = {
+  '404': {title: '404 Pages', prop: 'path'},
+  'Outbound Link: Click': {title: 'Outbound Links', prop: 'url'},
+  'File Download': {title: 'File Downloads', prop: 'url'}
+}
+
 export function specialTitleWhenGoalFilter(query, defaultTitle) {
-  switch (query.filters.goal) {
-    case '404':
-      return '404 Pages'
-    case 'Outbound Link: Click':
-      return 'Outbound Links'
-    case 'File Download':
-      return 'File Downloads'
-    default:
-      return defaultTitle
-  }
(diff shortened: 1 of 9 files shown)


assets/js/dashboard/stats/behaviours/goal-conversions.js
```diff
@@ -0,0 +1,56 @@
+import React from "react"
+import Conversions from './conversions'
+import ListReport from "../reports/list"
+import { CR_METRIC } from "../reports/metrics"
+import * as url from "../../util/url"
+import * as api from "../../api"
+
+export function specialTitleWhenGoalFilter(query, defaultTitle) {
+  switch (query.filters.goal) {
```
vinibrsl, line 9: What do you think of replacing this switch case with a constant that maps goal name and title? That could be used both here and in line 51. Here's an example:

```javascript
const SPECIAL_GOALS = {
  '404': '404 Pages',
  'Outbound Link: Click': 'Outbound Links',
  'File Download': 'File Downloads'
}

// Line 8
export function specialTitleWhenGoalFilter(query, defaultTitle) {
  return SPECIAL_GOALS[query.filter.goal] || defaultTitle
}

// Line 51
if (SPECIAL_GOALS[query.filters.goal]) {
```

I think this can help to DRY those special goal cases.

vinibrsl (approved):

assets/js/dashboard/stats/behaviours/goal-conversions.js
vinibrsl, line 9: I also think we should not let users create event goals with those special names. I have a PR for this: https://github.com/plausible/analytics/pull/3227

vinibrsl:
Related: https://github.com/plausible/docs/pull/418

PR #3228, opened 2023-08-03
Remove url and path prop keys from suggestions

vinibrsl:
This commit prevents the `url` and `path` prop keys from being suggested by the props settings page. These props are used internally for file downloads and outbound link clicks, and don't need to be manually allowed.

### DIFF
--- a/lib/plausible/props.ex
+++ b/lib/plausible/props.ex
@@ -71,6 +71,11 @@ defmodule Plausible.Props do
     allow(site, props_to_allow)
   end
 
+  # List of props to be ignored from suggestions. For example, `url` is used both
+  # for file downloads and outbound links, and it doesn't make sense to suggest
+  # users to allow this prop key.
+  @internal_props ["url"]
+
   @spec suggest_keys_to_allow(Plausible.Site.t(), non_neg_integer()) :: [String.t()]
   @doc """
   Queries the events table to fetch the #{@max_props} most frequent prop keys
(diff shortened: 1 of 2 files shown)


aerosol:
Even if it's not provided explicitly by the user when tracking outbound clicks, I'm not sure why are we trying to hide it. It's such a common name to use. 
  - What if the user sends a custom event with an `url` property? Will this change render it unusable?
  - How do we let them know they should rename it? 
  - What's the harm in keeping it in suggestions as it is right now?

vinibrsl:
> Even if it's not provided explicitly by the user when tracking outbound clicks, I'm not sure why are we trying to hide it. It's such a common name to use.
> 
> * What if the user sends a custom event with an `url` property? Will this change render it unusable?
> * How do we let them know they should rename it?
> * What's the harm in keeping it in suggestions as it is right now?

Good questions! We've identified two "internal" property keys that don't make sense to show up in suggestions: `url` and `path`. Although they're common names, I believe it doesn't make sense for users to track url and path info using props, as we already track those?

> What if the user sends a custom event with an `url` property? Will this change render it unusable? How do we let them know they should rename it?

They can still add `url` and `path` to the allow list if they need to. With this change, those special prop keys won't be suggested in the combo box to avoid confusion. A next PR will make sure to include `url` and `path` in queries as if they were in the allow list, as they are internal prop keys we always want to query.

> What's the harm in keeping it in suggestions as it is right now?

My thinking is that suggesting the user to allow `url` and `path` can be confusing. Those are internal and should be invisible to them.

RobertJoonas (approved):
👍

PR #3242, opened 2023-08-07
Remove Props FF

RobertJoonas:
### Changes

This is the final PR to clean up everything related to gradually releasing the new Custom Properties feature

* Removes the `props` feature flag and makes it a core feature
* Removes the following backend (internal API) functionalities:
  * the old `/property/:prop_name` endpoint
  * the old `prop_breakdown.csv` file (⚠️breaking change)
  * returning `prop_names` from the `/conversions` endpoint
* Removes the existing `Goal Conversions` UI (called `DeprecatedConversions` React component - since the whole component was replaced with a new one)
* Remove everything related to `allowed_event_props` from the CRM

### Tests

### Changelog

### Documentation

### Dark mode


vinibrsl (approved):
Good job!

PR #3251, opened 2023-08-09
Fix breakdown pagination when filtering by goals

aerosol:
### Changes

This PR addresses a crash on CR calculation when filtering by goals and using the details view.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -1399,8 +1399,9 @@ defmodule PlausibleWeb.Api.StatsController do
       |> Query.put_filter(filter_name, {:member, items})
       |> Query.remove_event_filters([:goal, :props])
 
-    # Because we are doing yet another query with a different filter, we should only keep
-    # the limit and keep the offset default to retrieve the first page always.
+    # Here, we're always only interested in the first page of results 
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1399,6 +1399,10 @@ defmodule PlausibleWeb.Api.StatsController do
       |> Query.put_filter(filter_name, {:member, items})
       |> Query.remove_event_filters([:goal, :props])
 
+    # Because we are doing yet another query with a different filter, we should only keep
```
RobertJoonas, line 1402: I got a bit confused by this comment 😃

What do you think about explaining this through the `:member` filter instead? I think that makes it a bit easier to understand. How about:

"Here, we're always only interested in the first page of results - the `:member` filter makes sure that the results always match with the items in the given `breakdown_results` list"

RobertJoonas (approved):
Nice catch! Just one comment about a comment, otherwise looks good to me :)

lib/plausible_web/controllers/api/stats_controller.ex
aerosol: Of course, df80c48f

PR #3266, opened 2023-08-15
Ignore unkown device type

ukutaht:
### Changes

The ua_inspector library has many device types that it reports. We have our own grouping rules to categorize them. For example, "smartphone", "wearable", and "portable media player" are  categorised as "Mobile" in Plausible.

Originally I added a fallback clause that reports any unknown device type to Sentry so that we can add it to our grouping rules if necessary. I wasn't sure if our rules were exhaustive or not, and the library might introduce new ones over time.

Turns out the library sometimes returns "unknown" as the device type. It is not helpful to report this to Sentry since we will never add a grouping rule for it. Since this gets triggered a lot, it's better to ignore the "unknown" device type completely which this PR does.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -372,6 +372,9 @@ defmodule Plausible.Ingestion.Event do
       %Device{type: t} when t in @desktop_types ->
         "Desktop"
 
+      %Device{type: "unknown"} ->
+        nil
+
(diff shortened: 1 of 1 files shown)


aerosol (approved):
Nice, we were considering completely removing that log message

cnkk (approved):

PR #3270, opened 2023-08-16
Inflate sampled results in Stats.Funnel.funnel/3

ukutaht:
### Changes

Our base Clickhouse queries are annotated with the [SAMPLE](https://clickhouse.com/docs/en/sql-reference/statements/select/sample) hint in order to reduce DB load on queries that scan over 20 million rows.

Clickhouse will not automatically inflate the results according to how much sampling was done. Instead, it returns a virtual row called `_sample_factor` which tells us exactly how much sampling was done to satisfy the query. I think it's left up to the user in case you want to do something crazy like run multiple subqueries from different tables, and joining the results for example. In some cases only the user (=query author) can tell how and when to inflate the results again for correct results.

Our use-case is mostly pretty straight-forward. It can get a bit complicated when we JOIN events and sessions but this is not relevant for this PR. This PR just adds the manual result inflation to the funnels query. It was missing in the initial implementation which causes bad results for very large dashboards.

~~For now, we don't have a way to test sampling and the correctness of results for very large result sets. We pretty much have to remember to do it right. The PR is in draft because I want to spend some time to see if I can set the `sample_thershold` really low and get a test to fail with the wrong implementation.~~

I tried testing sampling with really small numbers but it did not work. From Clickhouse docs:

> Since the minimum unit for data reading is one granule (its size is set by the index_granularity setting), it makes sense to set a sample that is much larger than the size of the granule.

It seems that in order for sampling to kick in, the amount of data you need to scan is more than the 'granule size' which in our case is 8192. Should we add some tests that create more than 10k records to test sampling?

### Tests

### Changelog

### Documentation

### Dark mode


aerosol:
> Should we add some tests that create more than 10k records to test sampling?

I think similar problem may exist at https://github.com/plausible/analytics/pull/3212 

I'm not sure if joining an unsampled query on a sampled one (there, select from `sessions_v2` comes with a `SAMPLE` clause, but joined `events_v2` subquery does not - I couldn't figure out if the sampling is necessary in the subquery). It's unclear to me whether the outer query `_sample_size` should be used perhaps?

Seeding a large dataset for tests seems reasonable. If this turns out slow we can tag the test(s) with `@tag :slow` and they will be skipped in local dev.

aerosol:
With more distribution for user_ids (random u64) the test seems a bit better. 
We should make some rough assertion (e.g. >40k and <50k - although with sampling threshold of 20k I'm getting over 50k visitors 🤔) and tag it with `:slow` so it doesn't run on local env.

test/plausible/funnels_test.exs
```diff
@@ -321,5 +321,35 @@ defmodule Plausible.FunnelsTest do
...
+          ]
+        end)
+
+      populate_stats(site, stats)
+
+      query =
+        Plausible.Stats.Query.from(site, %{"period" => "all", "sample_threshold" => "10000"})
+
+      {:ok, funnel_data} = Stats.funnel(site, query, funnel.id)
+      assert_in_delta funnel_data[:all_visitors], 50_000, 1000
```
aerosol, line 352: whoa TIL

ukutaht:
> although with sampling threshold of 20k I'm getting over 50k visitors 🤔

How many are you getting? As long as it's not something wildly different it should be ok. I think sampling can both over and underestimate depending on the data. I got 50,472 with a sampling threshold of 10k locally.

test/plausible/funnels_test.exs
```diff
@@ -334,7 +334,7 @@ defmodule Plausible.FunnelsTest do
       stats =
         1..50_000
         |> Enum.flat_map(fn n ->
-          user_id = SipHash.hash!("0123456789ABCDEF", Integer.to_string(n))
+          user_id = SipHash.hash("0123456789ABCDEF", :crypto.strong_rand_bytes(64))
```
aerosol, line 337: there will be unused `n` here sorry

PR #3274, opened 2023-08-17
Reconfigure session cookie

aerosol:
### Changes

Rename session cookie, set secure flag, regenerate signing salt.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/endpoint.ex
```diff
@@ -4,65 +4,70 @@ defmodule PlausibleWeb.Endpoint do
...
     # domain added dynamically via RuntimeSessionAdapter, see below
   ]
 
   # Serve at "/" the static files from "priv/static" directory.
   #
   # You should set gzip to true if you are running phx.digest
   # when deploying your static files in production.
-  plug PlausibleWeb.Tracker
-  plug PlausibleWeb.Favicon
+  plug(PlausibleWeb.Tracker)
```
aerosol, line 20: Sorry for those, there's a race condition in elixir-ls that makes the formatter kick in somewhat unpredictably. Until we upgrade Elixir we can't use the [fix](https://github.com/elixir-lsp/elixir-ls/pull/890). I'm tired of manually reverting those all the time.

lib/plausible_web/endpoint.ex
```diff
@@ -4,65 +4,70 @@ defmodule PlausibleWeb.Endpoint do
...
     store: :cookie,
-    key: "_plausible_key",
-    signing_salt: "3IL0ob4k",
+    key: "_plausible_session",
+    signing_salt: "I45i0SKHEku2f3tJh6y4v8gztrb/eG5KGCOe/o/AwFb7VHeuvDOn7AAq6KsdmOFM",
     # 5 years, this is super long but the SlidingSessionTimeout will log people out if they don't return for 2 weeks
     max_age: 60 * 60 * 24 * 365 * 5,
-    extra: "SameSite=Lax"
+    extra: "SameSite=Lax",
+    secure: true
```
ukutaht, line 12: @ruslandoga FYI we're changing the session cookie to be `secure: true` for production needs. This makes sessions not work over http.

For self-hosting our docs instruct people to set things up over http first and then add an ssl-terminating proxy as an optional extra. What do you think would be the most reasonable approach for self-hosting? Omit the secure flag by default to keep the current behaviour and make it a config option?

ukutaht (approved):

lib/plausible_web/endpoint.ex
ruslandoga: @ukutaht 👋

Since it's a Plug option, I assume it's set at compile-time, so it might not work in `config/runtime.exs` + `plausible-config.env` 

Those who run their websites over HTTPS have to use HTTPS in Plausible too, otherwise Chrome (and probably others too) won't allow the `tracker.js` to load over HTTP. So I think enforcing HTTPS makes sense.

ruslandoga: It could look something like this: https://github.com/plausible/analytics/pull/3277

aerosol: Guys, secure cookies don't enforce ssl in any way

aerosol: uh nevermind. I guess it only works for localhost over plain http.

config/runtime.exs
```diff
@@ -254,7 +271,9 @@ config :plausible, PlausibleWeb.Endpoint,
     protocol_options: [max_request_line_length: 8192, max_header_value_length: 8192]
   ],
   secret_key_base: secret_key_base,
-  websocket_url: websocket_url
+  websocket_url: websocket_url,
+  secure_cookie: secure_cookie,
```
ruslandoga, line 275: Just FYI, @aerosol @ukutaht it seems like `secure: true` is set automatically when `endpoint[:scheme]` is `https`

ruslandoga: https://github.com/elixir-plug/plug/blob/7b859ef85ede3154ff39249f0310064654d896e3/lib/plug/conn.ex#L1641-L1642

Docs: https://github.com/elixir-plug/plug/blob/7b859ef85ede3154ff39249f0310064654d896e3/lib/plug/conn.ex#L1594-L1595

ruslandoga: And since in Plausible `scheme` is read from `BASE_URL`, it seems like everything should work automatically, 

`BASE_URL=https://` -> `secure: true`
`BASE_URL=http://` -> `secure: false`

aerosol: Our base url is https but the cookie isn't marked secure

config/runtime.exs
```diff
@@ -232,6 +232,23 @@ if byte_size(websocket_url) > 0 and
...
+    config_dir
+    |> get_var_from_path_or_env("SECURE_COOKIE", "false")
+    |> String.to_existing_atom()
+  else
+    # in Plausible Cloud we always use `secure: true` in Plug.Session cookie opts
+    true
+  end
+
+cookie_key =
+  if env = get_var_from_path_or_env(config_dir, "ENVIRONMENT") do
```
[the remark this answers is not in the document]
ruslandoga: Sorry, this "suggested change" broke the config, there's the fix: https://github.com/plausible/analytics/pull/3278/files

config/runtime.exs
```diff
@@ -232,6 +232,18 @@ if byte_size(websocket_url) > 0 and
...
+  if is_selfhost do
+    config_dir
+    |> get_var_from_path_or_env("SECURE_COOKIE", "false")
+    |> String.to_existing_atom()
+  else
+    # in Plausible Cloud we always use `secure: true` in Plug.Session cookie opts
+    true
+  end
+
+cookie_key = "_plausible_session_#{env}"
```
aerosol, line 245: @ruslandoga this doesn't work sadly, the cookie set is "_plausible_session"

aerosol: The Plug.Session's init must be called with already patched options

aerosol: Perhaps we could revert back to RuntimeSessionAdapter, I don't mind the repetition tbh.
Also found this in sentry:

image

WYT?

aerosol: Pushed the changes, I'll confirm on staging

config/runtime.exs
```diff
@@ -232,6 +232,11 @@ if byte_size(websocket_url) > 0 and
   """
 end
 
+secure_cookie =
+  config_dir
+  |> get_var_from_path_or_env("SECURE_COOKIE", if(is_selfhost, do: "false", else: "true"))
```
aerosol, line 237: we'll sometimes need this set to false on dev, while working with fake domains e.g. plausible.local

config/runtime.exs
aerosol: The Sentry error is fine, the websocket connection was attempted at an old node that hasn't rotated yet

aerosol: Yeah I can't test this now because GitHub Actions are unavailable, but I still don't think this will help because of plug inits being compile-time. What a mess.

aerosol: Doesn't work still. On staging the cookie is _plausible_, local looks fine. 🤦

aerosol: It only worked on dev because of `config :phoenix, :plug_init_mode, :runtime`

lib/plausible_web/live_socket.ex
```diff
@@ -0,0 +1,15 @@
+defmodule PlausibleWeb.LiveSocket do
+  use Phoenix.LiveView.Socket
+  require Logger
+
+  def connect(params, %Phoenix.Socket{} = socket, connect_info) do
```
aerosol, line 5: https://github.com/phoenixframework/phoenix_live_view/commit/a825e6ba825c96d7fc98d27a2e8899aaa2b079f7

> "https://github.com/phoenixframework/phoenix_live_view/commit/a825e6ba825c96d7fc98d27a2e8899aaa2b079f7"

aerosol:
@ukutaht please have another look

ukutaht (approved):

PR #3277, opened 2023-08-18
Allow configure secure cookie

ruslandoga:
### Changes

Continues from https://github.com/plausible/analytics/pull/3274/files#r1298169062

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/endpoint.ex
```diff
@@ -57,7 +57,7 @@ defmodule PlausibleWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
 
-  plug(PlausibleWeb.Plugs.RuntimeSessionAdapter, @session_options)
+  plug(:runtime_session, Plug.Session.init(@session_options))
```
ruslandoga, line 60: Not sure if this is a good idea, but using a function plug we keep the logic in one place (here).

lib/plausible_web/endpoint.ex
```diff
@@ -75,11 +75,33 @@ defmodule PlausibleWeb.Endpoint do
     |> Keyword.fetch!(:websocket_url)
   end
 
-  def patch_session_opts() do
+  def secure_cookie? do
+    Application.fetch_env!(:plausible, :secure_cookie)
```
ruslandoga, line 79: Since it's called on each request, we might be able to cache it into `persistent_term`. It's probably plenty fast as is though.

aerosol:
@ruslandoga LGTM, could you also make the session key concatenate `ENVIRONMENT` runtime var? So that we end up with different cookie name for staging.

config/runtime.exs
```diff
@@ -232,14 +232,24 @@ if byte_size(websocket_url) > 0 and
   """
 end
 
+secure_cookie =
+  if is_selfhost do
+    config_dir
+    |> get_var_from_path_or_env("SECURE_COOKIE", "false")
+    |> String.to_existing_atom()
+  else
+    _cloud_version = true
```
aerosol, line 241: a comment would do a better job since this is unused anyway?

config/runtime.exs
```diff
@@ -232,6 +232,23 @@ if byte_size(websocket_url) > 0 and
...
+    # in Plausible Cloud we always use `secure: true`
+    true
+  end
+
+cookie_key =
+  if env = get_var_from_path_or_env("ENVIRONMENT") do
+    "_plausible_session_#{env}"
+  else
+    "_plausible_session"
+  end
```
ruslandoga, line 250: @aerosol like this?

lib/plausible_web/endpoint.ex
```diff
@@ -8,9 +8,9 @@ defmodule PlausibleWeb.Endpoint do
     signing_salt: "I45i0SKHEku2f3tJh6y4v8gztrb/eG5KGCOe/o/AwFb7VHeuvDOn7AAq6KsdmOFM",
```
ruslandoga, line 8: I'm keeping `key: "_plausible_session"` since Plug.Session.init fails at compile-time without it.

PR #3290, opened 2023-08-22
Dogfooding improvements

ukutaht:
### Changes

Start tracking dashboard pages using anonymized URLs:
1. `/:dashboard` for normal dashboard loads. 
2. `/share/:dashboard` for shared links.
3. `/:dashboard/settings/**` for the different settings sections

Adds custom event `Signup via invitation` for registrations that come through invitations, which is a different flow.

Started using tagged-events for our normal register page. Could not use it for the `Signup via invitation` flow because I AFAICT I cannot override the native `url` property using `tagged-events`. I thought about adding this capability to tagged-events but there may be an issue with people already using `plausible-event-url=custom://url.com` which would end up as a custom prop. I'm not sure how to deal with this, and I don't think we need to deal with this issue now (cc @RobertJoonas, just letting you know since you've worked on this the most)

When a dashboard is rendered as an embedded dashboard (`@conn.assigns[:embedded] == true`), then the events are sent not to `plausible.io` but `embed.plausible.io` instead.

Also started to introduce the label 'dogfooding' in code for when we track plausible.io using plausible.io. We used to call it 'tracking' but that's very generic and could refer to anything.


ukutaht:
NOTE: do we want to track embeds? Probably not.

lib/plausible_web/views/layout_view.ex
```diff
@@ -13,6 +13,36 @@ defmodule PlausibleWeb.LayoutView do
...
+  def dogfood_domain(conn) do
+    if conn.assigns[:embedded] do
+      "embed." <> base_domain()
+    else
+      base_domain()
+    end
+  end
+
+  def dogfood_page_url(id), do: PlausibleWeb.Endpoint.url() <> dogfood_page_path(id)
+  def dogfood_page_path(:dashboard), do: "/:dashboard"
```
aerosol, line 33: Any advantage of keeping those here instead of passing paths directly?
Do we want to crash on unknown identifiers?

lib/plausible_web/views/layout_view.ex
```diff
@@ -13,6 +13,36 @@ defmodule PlausibleWeb.LayoutView do
...
+
+  def dogfood_domain(conn) do
+    if conn.assigns[:embedded] do
+      "embed." <> base_domain()
+    else
+      base_domain()
+    end
+  end
+
+  def dogfood_page_url(id), do: PlausibleWeb.Endpoint.url() <> dogfood_page_path(id)
```
aerosol, line 32: Path.join?

aerosol (approved):

lib/plausible_web/templates/layout/_tracking.html.eex
```diff
@@ -1,9 +1,19 @@
 <%= if !Application.get_env(:plausible, :is_selfhost) && !@conn.assigns[:skip_plausible_tracking] do %>
```
aerosol, line 1: Do we still need `:skip_plausible_tracking` ?

ukutaht: Yes there are still many uses of `skip_plausible_tracking`. I did not convert all of them yet, there were just so many.

lib/plausible_web/views/layout_view.ex
ukutaht: When I was doing shared links I found that it was really error prone to manually fiddle with the paths. It's rendered in 3 different controller actions. When I wanted to change the path it was very easy to get it wrong because you have make 3 strings match exactly which are 100 lines apart in the file. 

Then again I think shared links is the only case where this is relevant. I could pull it up as a module constant in the controller that deals with shared links, and just pass the path directly in all other cases. WDYT?

ukutaht:
@aerosol A few updates:
1. Converted the `_tracking.html` layout to heex
1. I've pulled in `Phoenix.VerifiedRoutes`
2. I wanted to inline the dogfood page paths using `unverified_url(conn, ~p"/my/random/url")`. I assumed I need to use `unverified_url` since the paths used are not real paths that can be resolved by the router.
3. For some reason I get no compile errors when using `url(~p"/my/random/url"` ... A bit confused by this, I would hope that it fails. I'll see if there's any migration doc where I've missed a step

aerosol:
I've never used verified routes before, I'll take a look.

aerosol:
@ukutaht it's the catch-all route that fscks it up:

`get "/:domain/*path", StatsController, :stats`

it basically makes any route valid. Uncommenting it works:

```
Compiling 14 files (.ex)
warning: no route path for PlausibleWeb.Router matches ~p"/WRONGregister/invitation/:invitation_id"
  lib/plausible_web/controllers/auth_controller.ex:105: PlausibleWeb.AuthController.register_from_invitation_form/2
```

Plus, the choice of using colons in dogfood URL identifiers makes it again a bit tricky, because they're evaluated like path placeholders, thus:

```
~p"/register/invitation/:invitation_id"
```

is as valid as:

```
~p"/register/invitation/123123123"
```

I don't think it makes sense to use `VerifiedRoutes` in this setup.

ukutaht:
@aerosol I've just inlined the paths

PR #3293, opened 2023-08-24
Improve goal settings UX

aerosol:
### Changes

This PR improves the Goal Settings page UX.

~~:warning: do not merge before https://github.com/plausible/analytics/pull/3292 - I'll need to rebase then.~~ :heavy_check_mark: 

### Highlights:

- goal form is a modal
- combobox/autocompletion for pageview and currency fields
- goal type displayed in the goals list
- live search for goals list
- minor layout adjustments (e.g. long goal names are properly displayed now)

### Hints for reviewers:

- the most substantial commit is https://github.com/plausible/analytics/pull/3293/commits/fa5ed72ccec0c68bfa412ff09b3f46863dac83e6, going through the test descriptions first should be helpful

### Demo

(a little bit outdated already, please run the branch locally to see it in action)

https://github.com/plausible/analytics/assets/173738/8a3c4d5e-6996-4db8-8f6d-d5765ccd99a6

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -0,0 +1,274 @@
...
+      ]}
+      id="pageview-tab"
+      phx-click="switch-tab"
+    >
+      Pageview
+    </a>
+    """
+  end
+
+  def handle_event("switch-tab", _params, socket) do
```
ukutaht, line 241: On staging, tab switching feels off to me due to the roundtrip to the server. Would a little AlpineJS magic work here to make tab switching a client-only interaction? https://codebase-frontend-library.github.io/codebase-4/docs/8-alpinejs-components/alpinejs-tabs/

aerosol: Yes, absolutely, it was there before so I'll try to bring it back

lib/plausible_web/live/goal_settings.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  alias Plausible.{Sites, Goals}
+
+  def mount(
+        _params,
+        %{"site_id" => _site_id, "domain" => domain, "current_user_id" => user_id},
+        socket
+      ) do
+    site =
+      if Plausible.Auth.is_super_admin?(user_id) do
```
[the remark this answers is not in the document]
aerosol: https://github.com/plausible/analytics/pull/3293/commits/5c9f58e78fe8ec258d763a549e64c6580f440b65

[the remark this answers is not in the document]
aerosol: hold up that's not all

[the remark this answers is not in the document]
aerosol: 2acac253c

lib/plausible_web/live/goal_settings/form.ex
aerosol: I think it's due to large suggestions payload for currencies. I like to try initialising it asynchronously, WYT?

aerosol: https://github.com/plausible/analytics/pull/3293/commits/5252a620ba845424fcfc59725bfb48a4b6ee5833 and https://github.com/plausible/analytics/pull/3293/commits/42f0060369c1fb8e9018cdb9b40d62b34819a083 on their way to stage

ukutaht (commented):
@aerosol Nice, the modal/tabs do feel more snappy than before. I think the latency is definitely good in Europe. Probably acceptable-ish in the US as well. I agree that optimizing any further would sacrifice testability. Long-term we should probably look into the possibility of putting some app servers in the US to make these interactions feel really good (@cnkk 😉)

There's no changed line here to comment on the exact place but:
1. The `ComboBox` text input should be using `phx-debounce` instead of `phx-throttle`. Throttle has the issue that by typing fast (and stopping before 300ms is up) it's possible to get into a state where the dropdown results do not match the search term the user typed. Debounce will wait for the user to stop typing, and fire the search function at that point. It is definitely the 'correct' operation here as opposed to throttle. [Hexdocs description](https://hexdocs.pm/phoenix_live_view/bindings.html#rate-limiting-events-with-debounce-and-throttle) of the difference is not the best. I love [this interactive explanation](https://redd.one/blog/debounce-vs-throttle) much more 😊
2. I'm not sure what the purpose of the `async` option is.. I didn't dig into it too much but when I changed `phx-throttle` to `phx-debounce` locally, I didn't see any difference when using `async={true}` vs `async={false}` on the combobox. 
3. For some reason a 200ms debounce (which we use on the dashboard search inputs) feels much better to me than 300ms. The former feels instant whereas the latter has a noticeable gap when I stop typing. Let's go with 200 unless there are reasons not to?

ukutaht:
It also looks like the Search input has a higher z-index than the dimming overlay:

Screenshot 2023-09-01 at 12 04 31

aerosol:
> It also looks like the Search input has a higher z-index than the dimming overlay:

Ah right, I spotted that by the way of Custom Events UI and fixed there. Backported change: https://github.com/plausible/analytics/pull/3293/commits/6fde00d5e06b1f5263878acf754a48a8e83c5c53 

> Debounce will wait for the user to stop typing, and fire the search function at that point. It is definitely the 'correct' operation here as opposed to throttle. [Hexdocs description](https://hexdocs.pm/phoenix_live_view/bindings.html#rate-limiting-events-with-debounce-and-throttle) of the difference is not the best. I love [this interactive explanation](https://redd.one/blog/debounce-vs-throttle) much more

Thanks for this, I was struggling to understand the difference from the docs.

https://github.com/plausible/analytics/pull/3293/commits/edd15746009913b3986b93008c8779e7ef0de552

aerosol:
> I'm not sure what the purpose of the async option is.. I didn't dig into it too much but when I changed phx-throttle to phx-debounce locally, I didn't see any difference when using async={true} vs async={false} on the combobox.

Basically the form will show up before loading suggestions. Otherwise the clickhouse query returning URLs for page paths is blocking and so is transferring the payloads (also for currencies). The gains are noticeable with latency sim.

ukutaht:
> Basically the form will show up before loading suggestions. Otherwise the clickhouse query returning URLs for page paths is blocking and so is transferring the payloads (also for currencies). The gains are noticeable with latency sim.

Ah cool, that makes sense.

IDK, something still feels off about the ComboBox API but I can't put my finger on it. Let's merge this and I'll keep staring at it for a while to see if I get any ideas.

ukutaht (approved):

ukutaht:
> IDK, something still feels off about the ComboBox API but I can't put my finger on it

I think I realized what it is: It doesn't make much sense to me to make have both `async` and `sync` behaviours _in this component_. Every (debounced) keystroke will make a roundtrip to the server, so I think it should always be considered `async`. 

For a true synchronous search experience (like the non-async examples [here](https://react-select.com/home)), I think we'd need to ship the options to the frontend and use JS for filtering with a very low debounce value. This would be a completely different way of doing things and I don't see a need for it at the moment. Mayyybe it would be nice for currencies (and timezones in the future) but not worth the effort IMO.

Honestly this programming model is breaking my brain a bit but I'm enjoying it 😀

aerosol:
What I meant by async here is a separate `Task` is spawned whenever we (re)apply the suggestions function - regardless if it's done for the first render or based on user input. The connection is persistent and asynchronous by nature, it's just we're not blocking the main process.

ukutaht:
> What I meant by async here is a separate `Task` is spawned whenever we (re)apply the suggestions function - regardless if it's done for the first render or based on user input. The connection is persistent and asynchronous by nature, it's just we're not blocking the main process.

What's the advantage of using `async={false}` over `async={true}`? What use-cases is it better for and why?

aerosol:
async=false when the suggestions payload is small or the external resource used for retrieving suggestions is not expected to impose any latency - you don't care then and it's easier to test, no need to wait till the message arrives.

async=true when the payload is larger and/or the database is involved, to display the form as soon as possible and have it updated in the background. at the expense of sleeps in tests.

aerosol:
@ukutaht please see the latest two commits: https://github.com/plausible/analytics/pull/3293/commits/95b5c92092c41ba36c02d177b34fecbcfc7a8069 and https://github.com/plausible/analytics/pull/3293/commits/143982972ada984d886a9abd3eb283432f66f715, per @metmarkosaric's suggestions

aerosol:
@ukutaht 

> Modal scroll behaviour makes the form impossible to complete on screens with less height:

https://github.com/plausible/analytics/pull/3293/commits/1af1dcae1f0b02e2360fa10802f7bb9dc27eab4c

ukutaht:
>  it's easier to test

I was missing that, thanks.

PR #3305, opened 2023-08-29
Implement limits for team members

vinibrsl:
This pull request implements the team member limit, adds it to the Usage & Limits section and applies the limit when inviting new members.

I'll tackle ownership transfers in a separate pull request, as we need to consider not only team members but the total of sites as well.

### DIFF
--- a/lib/plausible/billing/quota.ex
+++ b/lib/plausible/billing/quota.ex
@@ -3,6 +3,7 @@ defmodule Plausible.Billing.Quota do
   This module provides functions to work with plans usage and limits.
   """
 
+  import Ecto.Query
   alias Plausible.Billing.Plans
 
   @limit_sites_since ~D[2021-05-05]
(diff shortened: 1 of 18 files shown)


lib/plausible/billing/quota.ex
```diff
@@ -79,6 +80,57 @@ defmodule Plausible.Billing.Quota do
...
+    end
+  end
+
+  @spec team_member_usage(Plausible.Auth.User.t()) :: integer()
+  @doc """
+  Returns the total count of team members and pending invitations associated
+  with the user's sites.
+  """
+  def team_member_usage(user) do
+    owned_sites_query =
```
ukutaht, line 105: It should be possible to get the same data from `site_memberships` directly without having to use a JOIN: 

`SELECT site_id from site_memberships WHERE user_id = {id} and role = 'owner'`

Our schema has a foreign key constraint on `site_memberships.site_id` so the result can be trusted.

lib/plausible/billing/quota.ex
```diff
@@ -79,6 +80,57 @@ defmodule Plausible.Billing.Quota do
...
+      from s in Plausible.Site,
+        inner_join: sm in assoc(s, :memberships),
+        where: sm.role == :owner and sm.user_id == ^user.id,
+        select: %{site_id: s.id}
+
+    team_members_query =
```
ukutaht, line 111: I think this can query `site_memberships` directly as well without having to go through `sites`.

lib/plausible/billing/quota.ex
```diff
@@ -79,6 +80,57 @@ defmodule Plausible.Billing.Quota do
...
+        on: i.site_id == os.site_id,
+        select: %{email: i.email},
+        union_all: ^team_members_query
+
+    query =
+      from itm in subquery(invitations_and_team_members_query),
+        where: itm.email != ^user.email,
+        select: count(itm.email, :distinct)
+
+    Plausible.Repo.one(query)
```
ukutaht, line 131: All together, it's a beautiful query 😊 gotta love SQL

lib/plausible/sites.ex
```diff
@@ -62,14 +63,24 @@ defmodule Plausible.Sites do
         else: :ok
     end
 
+    check_limit = fn site ->
```
ukutaht, line 66: Let's wrap all the queries in this function in a DB transaction to make sure the state stays consistent

ukutaht (approved):

PR #3323, opened 2023-09-05
Funnel Settings UI tweaks (to match Custom Props/Goals)

aerosol:
### Changes

:warning: This PR is on top of https://github.com/plausible/analytics/pull/3322 - please review/merge that one first so the base is changed.

The changes make the Funnels Settings similar to Goals and Custom Props so that:
  - funnels list can be filtered
  - funnels can be added via a modal
  - add funnel button is placed on top

Additional changes include:
  - making sure the ephemeral funnel evaluation is actually computed against a `month` period, instead of `all`
  - `StaticSearch` module is extended/tested with extra options (`to_string` and `weight_threshold`)

## Screenshots

image

image

image

image

image

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/funnel_settings/form.ex
```diff
@@ -39,119 +39,109 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
...
+              </label>
+
+              <div :for={step_idx <- @step_ids} class="flex mb-3 mt-3">
+                <div class="w-2/5 flex-1">
+                  <.live_component
+                    submit_name="funnel[steps][][goal_id]"
+                    module={PlausibleWeb.Live.Components.ComboBox}
+                    suggest_fun={&PlausibleWeb.Live.Components.ComboBox.StaticSearch.suggest/2}
+                    id={"step-#{step_idx}"}
+                    options={reject_alrady_selected("step-#{step_idx}", @goals, @selections_made)}
```
zoldar, line 88:
```suggestion
                    options={reject_already_selected("step-#{step_idx}", @goals, @selections_made)}
```

perhaps a good opportunity to fix that typo here an in the original definition below? 😄

zoldar (approved):
Aside from ESC behavior in suggestion input (which is already fixed in #3322), LGTM ✨

lib/plausible_web/live/funnel_settings/form.ex
aerosol: done, thanks

PR #3325, opened 2023-09-05
Render keybinding hints in the range picker

aerosol:
There was a regression in #3321 - some keybinding
hints stopped appearing due to exact string matching.


ukutaht:
Definitely better than before!

PR #3340, opened 2023-09-12
Allow admins to initiate ownership transfer from the CRM

ukutaht:
### Changes

Adds an action in the CRM to initiate ownership transfers. Uses the `Sites.invite` function that @vinibrsl recently extracted.

### Tests

### Changelog

### Documentation

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -93,8 +93,7 @@ defmodule Plausible.Sites do
     end
   end
 
-  defp check_invitation_permissions(site, inviter, requested_role, check_permissions: false),
-    do: :ok
+  defp check_invitation_permissions(_, _, _, check_permissions: false), do: :ok
 
   defp check_invitation_permissions(site, inviter, requested_role, _) do
     required_roles = if requested_role == :owner, do: [:owner], else: [:admin, :owner]
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -46,6 +46,14 @@ defmodule Plausible.Sites do
     end
   end
 
+  def bulk_transfer_ownership(sites, inviter, invitee_email, opts \\ []) do
+    Repo.transaction(fn ->
+      for site <- sites do
+        do_invite(site, inviter, invitee_email, :owner, opts)
+      end
+    end)
+  end
+
   @spec invite(Site.t(), Plausible.Auth.User.t(), String.t(), atom()) ::
           {:ok, Plausible.Auth.Invitation.t()}
           | {:error, Ecto.Changeset.t()}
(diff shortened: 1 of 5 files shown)


test/plausible/site/admin_test.exs
```diff
@@ -0,0 +1,57 @@
+defmodule Plausible.Site.AdminTest do
+  use Plausible.DataCase, async: true
+  use Bamboo.Test
+
+  setup do
+    admin_user = insert(:user)
+    conn = %Plug.Conn{assigns: %{current_user: admin_user}}
```
ukutaht, line 7: The `SiteAdmin` module is not a true controller but rather some type of service module that fits into Kaffy's framework. I didn't think the whole `ConnTest` harness was necessary here but let me know if you think otherwise.

test/plausible/site/admin_test.exs
```diff
@@ -0,0 +1,57 @@
...
+    action = Plausible.SiteAdmin.list_actions(conn)[:transfer_ownership][:action]
```
ukutaht, line 8: This is the only way I could access the action through the public interface of `SiteAdmin`. I can also make the `defp transfer_ownership` public instead to make the tests more readable?

aerosol: yeah this reads a bit abusive :D

I think it makes sense to extract the logic to something like `Sites.transfer_all_ownerships(inviter, invitee)` context interface:
 - test it in `SitesTest`
 - don't test the CRM controller (it's going away anyway)

aerosol (approved):

test/plausible/site/admin_test.exs
ukutaht: https://github.com/plausible/analytics/pull/3340/commits/4c4068d6b65fbb9c08399160f023b4e9c0375060

I left the tests in SiteAdmin for good measure, albeit a bit slimmer than before. They can be deleted when Kaffy goes away.

lib/plausible/sites.ex
```diff
@@ -92,7 +92,11 @@ defmodule Plausible.Sites do
       from m in Plausible.Site.Membership,
         where: m.user_id == ^inviter.id and m.site_id == ^site.id and m.role in ^required_roles
 
-    if Repo.exists?(membership_query), do: :ok, else: {:error, :forbidden}
+    cond do
+      Plausible.Auth.is_super_admin?(inviter) -> :ok
+      Repo.exists?(membership_query) -> :ok
+      true -> {:error, :forbidden}
+    end
```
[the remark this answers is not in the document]
ukutaht: Thanks. I would make the change but as per @aerosol's suggestion we don't check for superadmin role anymore https://github.com/plausible/analytics/pull/3340/commits/4c4068d6b65fbb9c08399160f023b4e9c0375060

lib/plausible/sites.ex
```diff
@@ -85,18 +93,17 @@ defmodule Plausible.Sites do
     end
   end
 
-  defp check_invitation_permissions(site, inviter, requested_role) do
+  defp check_invitation_permissions(site, inviter, requested_role, check_permissions: false),
```
aerosol, line 96: You could use `check_permissions? = Keyword.get(opts, :check_permissions?, true)` otherwise this will only match a single element list.

test/plausible/site/sites_test.exs
```diff
@@ -209,6 +209,38 @@ defmodule Plausible.SitesTest do
     end
   end
 
+  describe "bulk_transfer_ownership/4" do
+    test "initiates ownership transfer for multiple sites in one action" do
+      admin_user = insert(:user)
```
zoldar, line 214: I suppose this test fails due to existing admin user missing membership as in https://github.com/plausible/analytics/pull/3340/files#diff-2e507b8d42a73d8173d32d241881a56b0a1b90639dd885c7a2fb81b3980e39d2R37. Either this, or passing the option skipping membership check. Ideally, it would be good to test both scenarios but I'm of course not insisting 😅

PR #3343, opened 2023-09-14
Enforce goals unique

aerosol:
### Changes

~~:warning: draft till I figure out how to deal with goals that belong to funnels~~ #3361 

This PR sets up two unique constraints so that goals can be considered unique within their site scope. 

Technically duplicates are not a problem in terms of stats presentation, but we've observed some crashes coming from `Plasusible.Goals.find_or_create` since it expects only one result.

For Wordpress plugin integration we like this to work without surprises.

Because we already have some duplicates on production, an extra query to remove the dupes is needed.

Here we're doing this by deleting everything except the `min(id)` from a `GROUP BY goal_unique_tuple` query. I've tested this locally but since this is a destructive operation I'd like extra eyes on it please - cc @ukutaht @zoldar 

image

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20230914071245_goals_unique.exs
+++ b/priv/repo/migrations/20230914071245_goals_unique.exs
@@ -15,7 +15,7 @@ defmodule Plausible.Repo.Migrations.GoalsUnique do
       GROUP BY
       (site_id,
         CASE
-        WHEN page_path IS NOT NULL THEN goals.page_path
+        WHEN page_path IS NOT NULL THEN page_path
         WHEN event_name IS NOT NULL THEN event_name
         END )
     );
--- a/priv/repo/migrations/20230914071245_goals_unique.exs
+++ b/priv/repo/migrations/20230914071245_goals_unique.exs
@@ -8,19 +8,23 @@ defmodule Plausible.Repo.Migrations.GoalsUnique do
     execute """
     DELETE
     FROM
-    goals
+    goals g
     WHERE
-    id NOT IN (
+    g.id NOT IN (
       SELECT
-      min(id)
+      min(g2.id)
       FROM
-      goals
+      goals g2
       GROUP BY
-      (site_id,
+      (g2.site_id,
         CASE
-        WHEN page_path IS NOT NULL THEN page_path
-        WHEN event_name IS NOT NULL THEN event_name
+        WHEN g2.page_path IS NOT NULL THEN g2.page_path
+        WHEN g2.event_name IS NOT NULL THEN g2.event_name
         END )
+    )
+    AND g.id NOT IN (
+      SELECT fs.goal_id
+      FROM funnel_steps fs
     );
     """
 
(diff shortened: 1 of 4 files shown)


priv/repo/migrations/20230914071245_goals_unique.exs
```diff
@@ -0,0 +1,43 @@
...
+      GROUP BY
+      (site_id,
+        CASE
+        WHEN page_path IS NOT NULL THEN page_path
+        WHEN event_name IS NOT NULL THEN event_name
+        END )
+    );
+    """
+
+    create(
```
zoldar, line 24: Perhaps it's unnecessary here due to not that big data volume but have you considered adding the indexes with `concurrently: true`? 🤔 

Ideally, the destructive query would be run in a separate, preceeding migration. The indexes could be added in one go, however perhaps it would be good to drop wrapping transactions with:

```elixir
@disable_ddl_transaction true
@disable_migration_lock true
```

(BTW about the second option, TIL that it's now possible to switch to PG advisory lock for instead of disabling it completely https://hexdocs.pm/ecto_sql/Ecto.Migration.html#index/3-adding-dropping-indexes-concurrently).

priv/repo/migrations/20230914071245_goals_unique.exs
```diff
@@ -0,0 +1,43 @@
...
+    id NOT IN (
+      SELECT
+      min(id)
+      FROM
+      goals
+      GROUP BY
+      (site_id,
+        CASE
+        WHEN page_path IS NOT NULL THEN page_path
+        WHEN event_name IS NOT NULL THEN event_name
```
zoldar, line 19: Sorry, just to be sure - it should be unlikely that there are different table entries with same exact value for `page_path` and `event_name` (for the same site) respectively, right? 😅 (I'm assuming the fields are mutually exclusive already).

aerosol: mutually exclusive yeah

aerosol: At least after https://github.com/plausible/analytics/pull/3357 lolsob

priv/repo/migrations/20230914071245_goals_unique.exs
aerosol: It's a small dataset, https://github.com/plausible/analytics/pull/3343/commits/f95aa640c768bb525a9b9415c6f6772951f118ab should be fine

priv/repo/migrations/20230914071245_goals_unique.exs
```diff
@@ -0,0 +1,46 @@
...
+        END )
```
zoldar, line 23:
```suggestion
    DELETE
    FROM
    goals g
    WHERE
    g.id NOT IN (
      SELECT
      min(g2.id)
      FROM
      goals g2
      GROUP BY
      (g2.site_id,
        CASE
        WHEN g2.page_path IS NOT NULL THEN g2.page_path
        WHEN g2.event_name IS NOT NULL THEN g2.event_name
        END )
    )
    AND g.id NOT IN (
      SELECT fs.goal_id
      FROM funnel_steps fs
```

maybe, for more safety? 😅

aerosol: Thank you :heart:

zoldar (approved):
Tested sequence of deduplication and migration locally and it seems to behave as expected ✨

PR #3356, opened 2023-09-18
Fix Add Property button font weight

aerosol:
This sets the Add Property button font-weight to the ones seen in Add Goals and Add Funnels.


zoldar (approved):
🚀

PR #3357, opened 2023-09-18
[migration] Data fix: rewrite goals for which `event_name` and `page_path` both exist

aerosol:
### Changes

See https://github.com/plausible/analytics/commit/118795c1c86d9f45b819b935c69f1da1bafa7ab9 for explanation.

We've decided to favour `event_name` over `page_path`.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20230918084301_fix_broken_goals.exs
+++ b/priv/repo/migrations/20230918084301_fix_broken_goals.exs
@@ -1,0 +1,25 @@
+defmodule Plausible.Repo.Migrations.FixBrokenGoals do
+  use Ecto.Migration
+
+  def up do
+    execute """
+    UPDATE goals SET page_path = NULL WHERE page_path IS NOT NULL AND event_name IS NOT NULL
+    """
+
+    execute("""
+     ALTER TABLE goals
+     ADD CONSTRAINT check_event_name_or_page_path
+     CHECK (
+       (event_name IS NOT NULL AND page_path IS NULL) OR
+       (event_name IS NULL AND page_path IS NOT NULL)
+     );
+    """)
+  end
+
+  def down do
+    execute """
+    ALTER TABLE goals
+    DROP CONSTRAINT check_event_name_or_page_path;
+    """
+  end
+end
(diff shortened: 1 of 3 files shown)


priv/repo/migrations/20230918084301_fix_broken_goals.exs
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    execute("""
+     ALTER TABLE goals
+     ADD CONSTRAINT check_event_name_or_page_path
+     CHECK (
+       (event_name IS NOT NULL AND page_path IS NULL) OR
+       (event_name IS NULL AND page_path IS NOT NULL)
+     );
+    """)
```
zoldar, line 16: Not sure if this is going to be a problem, but we might want to add `NOT VALID` modifier at the end of `ADD CONSTRAINT` to avoid exclusive lock on the whole table when running that DDL query (as it has to scan the whole table to enforce the constraint on existing entries). By that point we already know it's fulfilled for existing entries - the UPDATE query is run right before it. WDYT?

priv/repo/migrations/20230918084301_fix_broken_goals.exs
```diff
@@ -0,0 +1,25 @@
+defmodule Plausible.Repo.Migrations.FixBrokenGoals do
+  use Ecto.Migration
```
zoldar, line 2: I think it would be good to drop transaction wrappings here, like it was mentioned in https://github.com/plausible/analytics/pull/3343#discussion_r1325677531

priv/repo/migrations/20230918084301_fix_broken_goals.exs
aerosol: Ah yeah, good point, data integrity is not endangered once the update is done.

ukutaht (approved):

priv/repo/migrations/20230918084301_fix_broken_goals.exs
aerosol: Done

PR #3358, opened 2023-09-18
Adding a note about WordPress plugin

metmarkosaric:
after getting another support email about this, i wanted to see if there was a nice looking way to include a quick link to the WordPress plugin and integration guides on the screen where we show the snippet. tried to do something but cannot test how it looks. 

it's way better experience when people use our plugin on WordPress rather than the manual snippet as the plugin reduces the number of cases where the tracking doesn't work correctly due to performance optimization plugins etc


ukutaht:
Made some changes for the layout, how does this look?

Screenshot 2023-09-18 at 16 06 27

ukutaht:
I think it's best to make the links open in a new tab so people are not removed from the onboarding flow

metmarkosaric:
@ukutaht nice, thanks! both suggestions sound good to me

lib/plausible/billing/subscription.ex
```diff
@@ -48,7 +48,8 @@ defmodule Plausible.Billing.Subscription do
     %__MODULE__{
       paddle_plan_id: "free_10k",
       status: "active",
-      next_bill_amount: "0"
+      next_bill_amount: "0",
+      currency_code: "EUR"
```
ukutaht, line 52: For testing the link on `user_settings.html` template

ukutaht (commented):
This PR adds some useful links to the onboarding flow.

Extra changes by me:
* I added a `styled_link` component to unify the styles for blue links.
* Used the `styled_link` component in some places where I found blue links in the UI. They'll have a consistent look along with hover and dark mode styles.

aerosol (approved):
Very nice and tidy ✨

PR #3363, opened 2023-09-22
Fix the timeseries for the 7d period with hour interval

krisbuist:
When bucketing on hour, the current day needs to be taken into account for as well. The query executed on the database already contained the correct results, but they were omitted because too few buckets were selected

### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/timeseries.ex
```diff
@@ -87,12 +87,12 @@ defmodule Plausible.Stats.Timeseries do
...
+  defp buckets(%Query{interval: "hour", period: period} = query) do
     n_buckets =
-      if query.date_range.first == query.date_range.last do
-        @full_day_in_hours
-      else
-        Timex.diff(query.date_range.last, query.date_range.first, :hours)
+      cond do
+        query.date_range.first == query.date_range.last -> @full_day_in_hours
+        #      period == "7d" -> Timex.diff(query.date_range.last |> Timex.shift(days: 1), query.date_range.first, :hours)
+        true -> Timex.diff(query.date_range.last, query.date_range.first, :hours)
```
ruslandoga, line 95: The new approach seems to be equivalent to the old one, no?

ruslandoga: Maybe `# period == "7d" ->` clause needs to be uncommented?

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -765,6 +765,194 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+               "2021-09-12 15:00:00",
+               "2021-09-12 16:00:00",
+               "2021-09-12 17:00:00",
+               "2021-09-12 18:00:00",
+               "2021-09-12 19:00:00",
+               "2021-09-12 20:00:00",
+               "2021-09-12 21:00:00",
+               "2021-09-12 22:00:00",
+               "2021-09-12 23:00:00",
+               "2021-09-13 00:00:00"
```
ruslandoga, line 951: `"2021-09-13 01:00:00"` and others still don't seem to be included

krisbuist: I am sorry, I wanted to make sure I had a failing test first, that I accidentally commented out the actual fix.
Whoops

lib/plausible/stats/timeseries.ex
krisbuist: Yeah, that was totally my mistake :) Thx!

krisbuist:
I'm wondering if this is actually a relevant fix or that the previous behaviour was intentional.
For me, the behaviour from this PR feels more intuitive, so that the "7d" period also shows the current day. But I might be overlooking something

PR #3370, opened 2023-09-25
Implement Plugins API Token schema

aerosol:
### Changes

This PR implements basic Plugins API tokens schema + context module. 
Plugin API tokens are passwords initially used for the Wordpress plugin authentication (TBA).

This PR contains the relevant postgres migration. It is safe to run at any time, the code is not exposed via any publicly available interfaces yet.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/plugins/api/tokens.ex
```diff
@@ -0,0 +1,41 @@
...
+  alias Plausible.Site
+  alias Plausible.Repo
+
+  import Ecto.Query
+
+  @spec create(Site.t(), String.t()) ::
+          {:ok, Token.t(), String.t()} | {:error, Ecto.Changeset.t()}
+  def create(%Site{} = site, description) do
+    with %{raw: raw, hash: hash} <- Token.generate(),
+         changeset <- Token.insert_changeset(site, %{token_hash: hash, description: description}),
```
zoldar, line 16: Given `token_hash` is not coming from user input, I'd consider providing it as an explicit argument and setting explicit guard for non-empty string just in case. At the same time though, given this changeset is only used internally by the service, it's fine as is as well 👍

lib/plausible/plugins/api/tokens.ex
```diff
@@ -0,0 +1,41 @@
...
+  @spec find(String.t(), String.t()) :: {:ok, Token.t()} | {:error, :not_found}
+  def find(domain, raw) do
+    found =
+      Repo.one(
+        from(t in Token,
+          inner_join: s in Site,
+          on: s.id == t.site_id,
+          where: t.token_hash == ^Token.hash(raw),
+          where: s.domain == ^domain or s.domain_changed_from == ^domain,
+          select: t
```
zoldar, line 31:
```suggestion
          where: s.domain == ^domain or s.domain_changed_from == ^domain
```

can be skipped as its implicitly selected. Unless that explicitness is intentional, then feel free to ignore 😅

priv/repo/migrations/20230925072840_plugins_api_tokens.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Plausible.Repo.Migrations.PluginsAPITokens do
+  use Ecto.Migration
+
+  def change do
+    create table("plugins_api_tokens", primary_key: false) do
+      add(:id, :uuid, primary_key: true)
+      add(:site_id, references(:sites, on_delete: :delete_all), null: false)
+      add(:token_hash, :bytea, null: false)
```
zoldar, line 8: Just out of curiosity, what's the reason for using byte column instead of text (the hash is still ASCII)? Is it about space savings?

zoldar (approved):
LGTM ✨ left 2 optional suggestions and one question

priv/repo/migrations/20230925072840_plugins_api_tokens.exs
aerosol: SHA digest is not UTF-8 encoded and it's never meant to be readable by humans anyway

zoldar: Ahh, right, I've missed the fact that base encoded rand bytes are for original token, not after hashing 🙈

lib/plausible/plugins/api/tokens.ex
aerosol: https://github.com/plausible/analytics/pull/3370/commits/21cb369d35673a34403ee28221075f480582548f

lib/plausible/plugins/api/tokens.ex
aerosol: Done via https://github.com/plausible/analytics/pull/3370/commits/21cb369d35673a34403ee28221075f480582548f - also I started storing a hint, i.e. last four bytes of the raw token, in case we like to present it to the users.

lib/plausible/plugins/api/tokens.ex
zoldar: 👍

PR #3376, opened 2023-09-27
Fix and refactor invitation logic

zoldar:
This PR refactors and fixes a number of issues with invitation flow:

- invitation accepting, rejecting and removing logic moved to dedicated "service" modules
- accepting logic significantly refactored for readability and clearer sepration between ownership transfer and invite flows
- fixed a bug where it was possible to accept transfer ownership on an owner, degrading them to admin and leaving site without an owner
- fix a bug where it was possible to degrade existing member's role with an invite
- fix grace period end handling in invitation acceptance due to a missing relationship preload in SiteLocker
- improve naming and test coverage of surrounding logic
- prevent issuing transfer ownership on owner's email at the time of issuing the invitation
- make acceptance of invitation and transfer ownerships in problematic cases graceful, without degrading roles and corrupting data integrity (can happen when users update their emails between issuing the invitation and accepting it)
- do not lock sites on transfer ownership on selfhosted instances
- prevent ability to remove invitation from another site on service logic level
- prevent ability to reject invitation by anyone else than invitee

### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -73,12 +73,14 @@ defmodule Plausible.Billing do
     end
   end
 
-  @spec needs_to_upgrade?(Plausible.Auth.User.t()) ::
+  @spec check_needs_to_upgrade(Plausible.Auth.User.t()) ::
           {true, :no_trial | :no_active_subscription | :grace_period_ended | nil}
           | {false, nil}
-  def needs_to_upgrade?(%Plausible.Auth.User{trial_expiry_date: nil}), do: {true, :no_trial}
+  def check_needs_to_upgrade(%Plausible.Auth.User{trial_expiry_date: nil}) do
+    {true, :no_trial}
+  end
 
-  def needs_to_upgrade?(user) do
+  def check_needs_to_upgrade(user) do
     user = Plausible.Users.with_subscription(user)
     trial_is_over = Timex.before?(user.trial_expiry_date, Timex.today())
     subscription_active = subscription_is_active?(user.subscription)
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -83,7 +83,7 @@ defmodule Plausible.Billing do
     cond do
       trial_is_over && !subscription_active -> {true, :no_active_subscription}
       Plausible.Auth.GracePeriod.expired?(user) -> {true, :grace_period_ended}
-      true -> false
+      true -> {false, nil}
     end
   end
 
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -73,6 +73,9 @@ defmodule Plausible.Billing do
     end
   end
 
+  @spec needs_to_upgrade?(Plausible.Auth.User.t()) ::
+          {true, :no_trial | :no_active_subscription | :grace_period_ended | nil}
+          | {false, nil}
   def needs_to_upgrade?(%Plausible.Auth.User{trial_expiry_date: nil}), do: {true, :no_trial}
 
   def needs_to_upgrade?(user) do
@@ -260,8 +263,30 @@ defmodule Plausible.Billing do
     end
   end
 
-  defp check_lock_status(user) do
-    Plausible.Billing.SiteLocker.check_sites_for(user)
+  def paddle_api(), do: Application.fetch_env!(:plausible, :paddle_api)
+
+  defp active_subscription_query(user_id) do
+    from(s in Subscription,
+      where: s.user_id == ^user_id and s.status == "active",
+      order_by: [desc: s.inserted_at],
+      limit: 1
+    )
+  end
+
+  defp after_subscription_update(subscription) do
+    user =
+      Plausible.Auth.User
+      |> Repo.get!(subscription.user_id)
+      |> Map.put(:subscription, subscription)
+
+    user
+    |> maybe_remove_grace_period()
+    |> update_lock_status()
+    |> maybe_adjust_api_key_limits()
+  end
+
+  defp update_lock_status(user) do
+    Plausible.Billing.SiteLocker.update_sites_for(user)
     user
   end
 
@@ -279,26 +304,5 @@ defmodule Plausible.Billing do
     end
 
     user
-  end
-
-  def paddle_api(), do: Application.fetch_env!(:plausible, :paddle_api)
-
-  defp active_subscription_query(user_id) do
-    from s in Subscription,
-      where: s.user_id == ^user_id and s.status == "active",
-      order_by: [desc: s.inserted_at],
-      limit: 1
-  end
-
-  defp after_subscription_update(subscription) do
-    user =
-      Plausible.Auth.User
-      |> Repo.get!(subscription.user_id)
-      |> Map.put(:subscription, subscription)
-
-    user
-    |> maybe_remove_grace_period()
-    |> check_lock_status()
-    |> maybe_adjust_api_key_limits()
   end
 end
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -283,13 +283,8 @@ defmodule Plausible.Billing do
 
     user
     |> maybe_remove_grace_period()
-    |> update_lock_status()
+    |> tap(&Plausible.Billing.SiteLocker.update_sites_for/1)
     |> maybe_adjust_api_key_limits()
-  end
-
-  defp update_lock_status(user) do
-    Plausible.Billing.SiteLocker.update_sites_for(user)
-    user
   end
 
   defp maybe_adjust_api_key_limits(user) do
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -74,10 +74,10 @@ defmodule Plausible.Billing do
   end
 
   @spec check_needs_to_upgrade(Plausible.Auth.User.t()) ::
-          {true, :no_trial | :no_active_subscription | :grace_period_ended | nil}
-          | {false, nil}
+          {:needs_to_upgrade, :no_trial | :no_active_subscription | :grace_period_ended}
+          | :no_upgrade_needed
   def check_needs_to_upgrade(%Plausible.Auth.User{trial_expiry_date: nil}) do
-    {true, :no_trial}
+    {:needs_to_upgrade, :no_trial}
   end
 
   def check_needs_to_upgrade(user) do
@@ -86,9 +86,9 @@ defmodule Plausible.Billing do
     subscription_active = subscription_is_active?(user.subscription)
 
     cond do
-      trial_is_over && !subscription_active -> {true, :no_active_subscription}
-      Plausible.Auth.GracePeriod.expired?(user) -> {true, :grace_period_ended}
-      true -> {false, nil}
+      trial_is_over && !subscription_active -> {:needs_to_upgrade, :no_active_subscription}
+      Plausible.Auth.GracePeriod.expired?(user) -> {:needs_to_upgrade, :grace_period_ended}
+      true -> :no_upgrade_needed
     end
   end
 
--- a/lib/plausible/billing/site_locker.ex
+++ b/lib/plausible/billing/site_locker.ex
@@ -1,7 +1,13 @@
 defmodule Plausible.Billing.SiteLocker do
   use Plausible.Repo
 
-  def check_sites_for(user, opts \\ []) do
+  @type update_opt() :: {:send_email?, boolean()}
+
+  @type lock_reason() :: :grace_period_ended_now | :grace_period_ended_already
+
+  @spec update_sites_for(Plausible.Auth.User.t(), [update_opt()]) ::
+          {:locked, lock_reason()} | {:unlocked, nil}
+  def update_sites_for(user, opts \\ []) do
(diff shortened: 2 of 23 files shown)


zoldar:
I have ran some rudimentary testing locally but if anyone feels like taking it for a spin themselves, I have deployed that PR to staging at, available at https://staging.plausible.io/login 😄

lib/plausible/site/memberships/accept_invitation.ex
```diff
@@ -0,0 +1,163 @@
...
+      multi =
+        if invitation.role == :owner do
+          add_and_transfer_ownership(invitation, membership, user, selfhost?)
+        else
+          add(invitation, membership, user)
+        end
+
+      case Repo.transaction(multi) do
+        {:ok, changes} ->
+          if changes[:site_locker] == {:locked, :grace_period_ended_now} do
```
zoldar, line 42: This is done to keep side effects out of transaction and assure things like email submission are done at most once. I'm not super keen on exposing internals like this and there are ways to address this (I've even made a small library in the past for seamlessly dispatching, modifying or discaring side effects in such contexts), but this can do for now.

lib/plausible/site/memberships/accept_invitation.ex
```diff
@@ -0,0 +1,112 @@
...
+      membership = get_or_create_membership(invitation, user)
+
+      multi =
+        Multi.new()
+        |> maybe_prepare_ownership_transfer(invitation, user)
+        |> Multi.insert_or_update(:membership, membership)
+        |> Multi.delete(:invitation, invitation)
+        |> Multi.run(:site_locker, fn _, %{user: updated_user} ->
+          {:ok,
+           Billing.SiteLocker.check_sites_for(Repo.reload!(updated_user), send_email?: false)}
```
ukutaht: It took me a long time to rebuild the context for this check. I *think* the idea is: If I'm on a trial, and someone transfers ownership of a site to me, it should get locked immediately so that I am prompted to upgrade my subcription. This prevents a loophole where I could register a new user every 30 days and keep transferring the ownership of a site indefinitely without paying.

I'm reviewing commit-by-commit so maybe this is outdated but my suggestion here would be to make the code reflect the business logic more directly:

1. The check is probably more appropriate in the `maybe_prepare_ownership_transfer` function because it only matters in case we're dealing with an ownership transfer.
2. Stop relying on the indirect logic of ending the trial, calling `SiteLocker.check_sites_for`, which calls `Billing.needs_to_upgrade?` which is always true if trial has been ended. Refactor to lock the site directly in case it's of an ownership transfer.
3. Instead of sending two emails: `ownership_transfer_accepted` and  `dashboard_locked`, add the upgrade details to the first one and stop sending the second.

test/plausible/site/memberships/accept_invitation_test.exs
```diff
@@ -231,5 +231,46 @@ defmodule Plausible.Site.Memberships.AcceptInvitationTest do
       assert Repo.reload!(new_owner).trial_expiry_date == Date.add(Date.utc_today(), -1)
       assert Repo.reload!(site).locked
     end
+
+    test "ends grace period and sends an email about it if new owner is in grace period" do
```
ukutaht, line 282: Darn it. My previous comment about locking the sites directly did not factor in this case. Happy to leave it as-is.

test/plausible_web/controllers/invitation_controller_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule PlausibleWeb.Site.InvitationControllerTest do
-  use PlausibleWeb.ConnCase
+  use PlausibleWeb.ConnCase, async: true
```
ukutaht, line 2: ❤️👏

ukutaht (approved):
Looks great @zoldar. Was a bit of a journey for me to remind all these code paths and what they're for 😀

lib/plausible/site/memberships/accept_invitation.ex
ukutaht: After further review about the emails: I think we should just remove the [grace period end email](https://github.com/plausible/analytics/blob/1dca70a380bc48296045a514b8283639858851d8/lib/plausible/site/memberships/accept_invitation.ex#L43) from this flow.

If you check the copy of that email, it is very specific to the flow where the user exceeds their plan limits:
1. User goes over their limits for 2 billing intervals in a row
2. [Gets notified about it over email](https://github.com/plausible/analytics/blob/1dca70a380bc48296045a514b8283639858851d8/lib/workers/check_usage.ex#L92-L108), and grace period is started
3. After 7 days, grace period ends
4. Nightly job (`site_locker`) calls `SiteLocker` to lock their dashboards and sends an email to notify the user about being locked out

Since an ownership transfer is accepted in the UI and we warn the customer about the effects directly [here](https://github.com/plausible/analytics/blob/1dca70a380bc48296045a514b8283639858851d8/lib/plausible_web/templates/site/index.html.eex#L162), I think it's OK to not send any email about it.

lib/plausible/billing/billing.ex
```diff
@@ -83,7 +83,7 @@ defmodule Plausible.Billing do
     cond do
       trial_is_over && !subscription_active -> {true, :no_active_subscription}
       Plausible.Auth.GracePeriod.expired?(user) -> {true, :grace_period_ended}
-      true -> false
+      true -> {false, nil}
```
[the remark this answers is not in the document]
zoldar: It's a bit of a personal preference. Regular result tuple shape is more convenient to match against in various contexts. For instance if somebody is only interested in the second value (reason), they can just rely on a single "{_, reason}" tuple. Otherwise calling code has to always have another branch for a different pattern for completeness and to avoid crashing, even if both branches could follow the same logic. It's a bit like handling function returning single type and a union of types in static languages - the other one is more expressive but more cumbersome to handle.

lib/plausible/billing/site_locker.ex
```diff
@@ -1,33 +1,47 @@
...
 
-      {true, _} ->
+      {true, reason} ->
         set_lock_status_for(user, true)
+        {:locked, reason}
 
-      _ ->
+      {false, _} ->
         set_lock_status_for(user, false)
+        {:unlocked, nil}
```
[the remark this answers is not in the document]
zoldar: Ditto about reasoning in https://github.com/plausible/analytics/pull/3376#discussion_r1340342430

test/plausible/site/memberships/accept_invitation_test.exs
ukutaht: >  It's a pretty narrow edge case (perhaps that's why missing preload hasn't caused any significant issues in production 😅 )

Yeah, I wouldn't be surprised if it has never happened on prod.

lib/plausible/billing/billing.ex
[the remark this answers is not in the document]
ukutaht: Interesting. My personal preference has always been the more expressive union type. I don't like bikeshedding over stylistic preference so let's just forget it 😀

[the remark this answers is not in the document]
zoldar: On a second thought, that was rather shoddy justification 😂 Unions of course are great. And tuple shape argument falls flat in cases like `:ok | {:error, ...}` or `{:ok, ..} | :error`, which can be perfectly fine (well, perhaps mixing `{:error, _}` and `{:error, _, _}` can be seen as a bit more legit case of inconvenient mismatch - I usually go with `{:error, {_, _}}` in such cases, but again, no hard feelings on it). 

What actually irked me about this particular case, is a return type mixing boolean with tuples. I have seen similar cases where what begun life as a plain boolean, turned into tuple/boolean and wreaked havoc due to some parts of the code still treating it as boolean (truthiness lets it slide). And one of such instances was actually billing code haha.

I'll revise that again and probably make it return an atoms like `{:upgrade, ...} | :no_upgrade`, or something to that effect.

lib/plausible/billing/site_locker.ex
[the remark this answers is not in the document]
zoldar: It doesn't make sense really, and was a bit of a knee-jerk reaction to https://github.com/plausible/analytics/pull/3376#discussion_r1340942962. I'll revert that change.

lib/plausible/billing/billing.ex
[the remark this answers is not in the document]
ukutaht: > I have seen similar cases where what begun life as a plain boolean turned into tuple/boolean and wreaked havoc due to some parts of the code still treating it as boolean (truthiness lets it slide)

Oh yeah, I can see how easily that could happen

> I'll revise that again and probably make it return an atom like {:upgrade, ...} | :no_upgrade, or something to that effect.

I like that!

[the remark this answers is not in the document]
zoldar: https://github.com/plausible/analytics/pull/3376/commits/c8e899ac71a7c9d39eb1c45209779062c4554f61

lib/plausible/billing/site_locker.ex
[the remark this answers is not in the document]
zoldar: https://github.com/plausible/analytics/pull/3376/commits/4a16714b5ab68d836df65f2f96dfbdf7c27422e5

lib/plausible/billing/billing.ex
```diff
@@ -73,17 +73,22 @@ defmodule Plausible.Billing do
...
     user = Plausible.Users.with_subscription(user)
     trial_is_over = Timex.before?(user.trial_expiry_date, Timex.today())
     subscription_active = subscription_is_active?(user.subscription)
 
     cond do
-      trial_is_over && !subscription_active -> {true, :no_active_subscription}
-      Plausible.Auth.GracePeriod.expired?(user) -> {true, :grace_period_ended}
-      true -> false
+      trial_is_over && !subscription_active -> {:needs_to_upgrade, :no_active_subscription}
+      Plausible.Auth.GracePeriod.expired?(user) -> {:needs_to_upgrade, :grace_period_ended}
```
aerosol, line 90: feel free to disregard -- I remember you've already discussed it, but could be a struct as well - descriptive and easy to match against:

```elixir
def subscription_state(user) do
  %SubscriptionState{
    trial_over?: ...,
    active?: ...,
    needs_upgrade?: ...,
    needs_upgrade_reason: ...,
    grace_period_over?: ...
  }
end
```

lib/plausible/site/memberships/accept_invitation.ex
```diff
@@ -0,0 +1,163 @@
+defmodule Plausible.Site.Memberships.AcceptInvitation do
+  @moduledoc """
+  Service for accepting invitations, including ownership transfers.
+
+  Accepting invitation accounts for the fact that it's possible
+  that accepting user has an existing membership for the site and
+  acts permissively to not unnecesarily disrupt the flow while
```
aerosol, line 7:
```suggestion
  acts permissively to not unnecessarily disrupt the flow while
```

test/plausible/site/memberships/accept_invitation_test.exs
```diff
@@ -0,0 +1,323 @@
...
+
+      assert existing_membership.id == new_membership.id
+      assert existing_membership.user_id == new_membership.user_id
+      assert existing_membership.site_id == new_membership.site_id
+      assert existing_membership.role == new_membership.role
+      assert new_membership.role == :admin
+      refute Repo.reload(invitation)
+    end
+
+    test "returns an error on non-existent inviation" do
```
aerosol, line 85:
```suggestion
    test "returns an error on non-existent invitation" do
```

aerosol (approved):

lib/plausible/billing/billing.ex
zoldar: Sounds cool, but it would overload meaning of a subscription (as a concrete entity) which is just one of the variables in overall user's state that is evaluated here. I'd keep it as is for now, if you don't mind.

lib/plausible/site/memberships/accept_invitation.ex
ukutaht: Oh yeah sorry. [The comment](https://github.com/plausible/analytics/pull/3376#discussion_r1340230951) referred just to the behaviour of ending the grace period in case the user has one.

>  letting them know by email wouldn't hurt

I think it would make sense to send *some* email to the user. But the one that gets sent currently is written for the normal grace period flow. The email starts with  _'Last week we sent a reminder that..'_  because normally we give people 7 days to upgrade which we call the 'grace period'. But when someone accepts an ownership transfer it might happen 5 minutes after the grace period was started. It also says _'As we haven't heard from you, we've now locked your stats'_ - stating the wrong reason why the dashboards got locked. 

So the email copy is inaccurate and misleading. Besides, we don't notify the new owner in any other states like 'active trial' which is way more common. It's rare for ownership to be transferred to someone who happens to be on a grace period.

I agree it wouldn't hurt to let the new owner know via email what has happened as well. But to do it right we would need to come up with a new email copy and send it for all new owner states. Ownership transfers are rare and they seem to work fine without email notifications. The severity/priority doesn't justify the time investment IMO.

ukutaht: So to be clear: Let's remove the current (bad) email and leave it there.

zoldar: All fair points, sorry, I have not fully registered that copy is so misleading in this case 🙈

PR #3377, opened 2023-09-27
Fix dogfooding

ukutaht:
### Changes

1. Removes the `skip_plausible_tracking` assign from 2 pages that should be tracked
2. Starts sending our own events to `ingest.plausible.io` for further testing of that endpoint

### DIFF
--- a/lib/plausible_web/views/layout_view.ex
+++ b/lib/plausible_web/views/layout_view.ex
@@ -29,6 +29,16 @@ defmodule PlausibleWeb.LayoutView do
     end
   end
 
+  @doc """
+  Temporary override to do more testing of the new ingest.plausible.io endpoint for accepting events. In staging and locally
+  will fall back to staging.plausible.io/api/event and localhost:8000/api/event respectively.
+  """
+  def dogfood_api_destination() do
+    if Application.get_env(:plausible, :environment) == "prod" do
+      "https://ingest.plausible.io/api/event"
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


lib/plausible_web/views/layout_view.ex
```diff
@@ -29,6 +29,16 @@ defmodule PlausibleWeb.LayoutView do
     end
   end
 
+  @doc """
+  Temporary override to do more testing of the new ingest.plausible.io endpoint for accepting events. In staging and locally
+  will fall back to staging.plausible.io/api/event and localhost:8000/api/event respectively.
+  """
+  def dogfood_api_destination() do
+    if Application.get_env(:plausible, :environment) == "prod" do
+      "https://ingest.plausible.io/api/event"
```
zoldar, line 38: Maybe I'm talking nonsense 😅 but - won't this break self-hosted instances? 🤔

zoldar: Ah, nevermind, it shouldn, though it might trigger rogue event submissions if somebody else deploys that version (but these should be safely discarded, I think? 😅 ).

zoldar (approved):

lib/plausible_web/views/layout_view.ex
zoldar: Agh, on third look, this all is behind selfhost flag, don't mind me 🤡

PR #3373, opened 2023-09-26
Implement Plugins API authorization

aerosol:
### Changes

On top of #3370 

This PR implements a Plug for Plugins API authorization.
Credentials are carried over in a Basic Auth header consisting of site domain and the Plugins API Token (#3370).

### Tests
- [x] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #3378, opened 2023-09-27
OpenAPI: first pass on Plugins API - Shared Links

aerosol:
### Changes

This PR implements Shared Links resource exposed via the new Plugins API (work in progress, initially aiming at new Wordpress plugin integration).

The API implements [OpenAPI Specification 3](https://swagger.io/specification/) (via https://hexdocs.pm/open_api_spex/readme.html) for the benefit of easier 3rd party client integration with stronger contract enforcement.

The Plugins API is served off a new router (`PlausibleWeb.Plugins.API.Router`) that requests coming to `/api/plugins` are forwarded to. This is only for the sake of readability, the TCP endpoint remains the same for now.

The [schema](https://gist.github.com/aerosol/197d3d37bccc6c3c55af1e4572efb32d) is served from `/api/plugins/spec/openapi`, read by the swagger UI at `/api/plugins/spec/swagger-ui`. Based on the schema, clients should be able to [generate their implementations](https://openapi-generator.tech/) and keep them verified against the contract definition. At later stage, we'll look into generating and publishing documentation from it, for now the swagger UI should be good enough.

At later stage, incorporating schema generation into our CI flow will help us detect any breaking changes (we'll be able to compare the base schema to the PR submitted for example). 

API authentication is done via tokens, carried over in the `authorization` header, implemented via #3373 and #3370.

The UI for provisioning authentication tokens will be done in a separate PR. Currently the `Plausible.Plugins.API.Tokens.create/2` function can be used to play with the API locally.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/plugins/api/context/shared_links.ex
```diff
@@ -0,0 +1,54 @@
+defmodule PlausibleWeb.Plugins.API.Context.SharedLinks do
```
zoldar, line 1: Perhaps that module could be put in core namespace, under `Plausible`?

priv/repo/migrations/20230914071245_goals_unique.exs
```diff
@@ -18,8 +18,8 @@ defmodule Plausible.Repo.Migrations.GoalsUnique do
       GROUP BY
       (g2.site_id,
         CASE
-        WHEN g2.page_path IS NOT NULL THEN g2.page_path
-        WHEN g2.event_name IS NOT NULL THEN g2.event_name
+        WHEN g2.page_path IS NOT NULL THEN 'Page: ' || g2.page_path
+        WHEN g2.event_name IS NOT NULL THEN 'Event: ' || g2.event_name
```
aerosol, line 22: On staging we've had goals equally named but of different types, so that's another edge case to be addressed. Ref #3343

lib/plausible_web/plugins/api/errors.ex
```diff
@@ -12,6 +12,20 @@ defmodule PlausibleWeb.Plugins.API.Errors do
     |> error(:unauthorized, "Plugins API: unauthorized")
   end
 
+  @spec internal_server_error(Plug.Conn.t()) :: Plug.Conn.t()
```
zoldar, line 15: Do we intend to user that error somewhere? It would be good to have a well formed page like this as a default maybe, especially for UI. [API controllers seem to implement something rudimentary already at least](https://github.com/plausible/analytics/blob/777b4b3741197428616e708e916647bbc6d0b395/lib/plausible_web/controllers/api/stats_controller.ex#L1282-L1284).

priv/repo/migrations/20230914071244_fix_broken_goals.exs
aerosol, line 1: We ran this on prod in correct order, but merging was a bit careless and reversed it

test/plausible_web/plugins/api/controllers/shared_links_test.exs
```diff
@@ -0,0 +1,238 @@
...
+        |> json_response(200)
+        |> assert_schema("SharedLink.ListResponse", spec())
+
+      assert resp.data == []
+      assert resp.meta.pagination.has_next_page == false
+      assert resp.meta.pagination.has_prev_page == false
+      assert resp.meta.pagination.links == %{}
+    end
+
+    test "returns an shared links list with pagination", %{
```
ukutaht, line 193:
```suggestion
    test "returns a shared links list with pagination", %{
```

test/support/plugins_api_case.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  setup tags do
+    :ok = Ecto.Adapters.SQL.Sandbox.checkout(Plausible.Repo)
+
+    unless tags[:async] do
+      Ecto.Adapters.SQL.Sandbox.mode(Plausible.Repo, {:shared, self()})
+    end
+
+    # randomize client ip to avoid accidentally hitting
+    # rate limiting during tests
```
ukutaht, line 43: Is there rate limiting in place? I can't see any..

ukutaht (approved):
As you said - OpenAPI specs are a bit verbose to define but looks good to me! Love the `assert_spec` helper in tests :)

test/support/plugins_api_case.ex
aerosol: Ah you're right there's none currently. I've carelessly copied that bit from our events controller test case, removed via 155e23843 - will submit a separate PR for it.

lib/plausible_web/plugins/api/context/shared_links.ex
aerosol: ec9637ebd

zoldar: Affected aliases might need updating still 🙈 https://github.com/plausible/analytics/actions/runs/6377286453/job/17305659271?pr=3378

aerosol: Right right 😅

lib/plausible_web/plugins/api/errors.ex
aerosol: I don't think we want to leak error/exception details, especially for UI - unless I misunderstood. We even track an issue where there's too many details returned from the Stats API the user can't do anything about anyway.

Good point though, I'm not sure how can we even use that one since it's the same endpoint as everything else, sadly, and it's already configured to use `PlausibleWeb.ErrorView` so a true response to an exception will look like this:

image

I think I'll remove that function for now, WYT?

aerosol: Like this: b63217a76

aerosol: WYT @zoldar?

zoldar: Looks good 👍  And I haven't meant explicitly exposing exceptions - just handling these errors with a well formed response, like you said (had a pause when looking at that myself but assumed code being in the open anyway, it might have been intentional; I wasn't sure and forgot to ask).

lib/plausible/plugins/api/tokens.ex
```diff
@@ -10,9 +10,9 @@ defmodule Plausible.Plugins.API.Tokens do
   import Ecto.Query
 
   @spec create(Site.t(), String.t()) ::
-          {:ok, Token.t(), String.t()} | {:error, Ecto.Changeset.t()}
-  def create(%Site{} = site, description) do
-    with generated_token <- Token.generate(),
+          {:ok, Token.t(), String.t(), String.t()} | {:error, Ecto.Changeset.t()}
+  def create(%Site{} = site, description, generate \\ Token.generate()) do
+    with generated_token <- generate,
```
zoldar, line 15: hmm maybe just name the argument `generated_token`?

aerosol: nice! b7f65fec9

PR #3387, opened 2023-10-02
Handle missing or expired token in password reset action and LV gracefully

zoldar:
### Changes

This change addresses two problems:

* controller action crashing on missing "token" param - it's handled gracefully
  now and will not pollute Sentry anymore with http://sentry.plausible.io/organizations/sentry/issues/4319
* LiveView receives email extracted from token on initial page load instead
  of reverifying token on every re-mount (which can happen when somebody
  leaves form open for an extended period of time; rare but happens and
  needlessly pollutes Sentry as well)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/reset_password_form.ex
+++ b/lib/plausible_web/live/reset_password_form.ex
@@ -11,11 +11,9 @@ defmodule PlausibleWeb.Live.ResetPasswordForm do
   alias Plausible.Auth
   alias Plausible.Repo
 
-  def mount(_params, %{"reset_token" => reset_token}, socket) do
+  def mount(_params, %{"email" => email}, socket) do
     socket =
       assign_new(socket, :user, fn ->
-        # by that point token should be already verified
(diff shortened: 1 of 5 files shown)


lib/plausible_web/live/reset_password_form.ex
```diff
@@ -60,7 +57,6 @@ defmodule PlausibleWeb.Live.ResetPasswordForm do
           class="transition bg-gray-100 dark:bg-gray-900 outline-none appearance-none border border-transparent rounded w-full p-2 text-gray-700 dark:text-gray-300 leading-normal appearance-none focus:outline-none focus:bg-white dark:focus:bg-gray-800 focus:border-gray-300 dark:focus:border-gray-500"
         />
       </div>
-      <input name="token" type="hidden" value={@reset_token} />
```
zoldar, line 63: This field was unnecessary either way, as POST action is only doing logout of existing session and a redirect to the login page with a relevant flash message.

aerosol (approved):
🚀

PR #3396, opened 2023-10-04
Plugins API: 2nd pass with Goals resource + SharedLinks schema changes

aerosol:
### Changes

First and foremost, this PR implements the `Goals` resource in the Plugins API.

---

:information_source: Similarly to #3378, while reviewing, you can play with the API locally by visiting [http://localhost:8000/api/plugins/spec/swagger-ui](http://localhost:8000/api/plugins/spec/swagger-ui). After seeding the database(s), the dev token for `dummy.site` will be `plausible-plugin-dev-seed-token` - use that to authorize your calls in the SwaggerUI.

The examples displayed are expected to work out of the box.

---

The `goals` resource exposes endpoints for:

  - goal creation any type of currently supported goals, either a single object or in bulk (this is however limited - a single request can contain maximum number of `Funnel.max_steps()` goals, future proof - for when we incorporate funnels in the API). Attempt to create an existing goal will fetch the existing instance (so it's effectively "get or create"). 
     - one exception to this is trying to create a Revenue Goal with a different currency symbol than the existing one - in such case, the API will return a descriptive error
  - goal listing (either single instance by ID or paginated list)
  - goal deletion (single instance only)

Goal creation and deletion requests are idempotent. 

Because the API distinguishes three main goal types (`Revenue`, `Pageview`, `CustomEvent`), the UI was slightly changed by the way. In the goal settings list we'll no longer see:

```
Purchase
Revenue Goal (EUR)
```

but:

```
Purchase (EUR)
Revenue Goal
```

This is aligned with how we cast pageview goals to strings (`"Visit /..."`).

---

Furthermore, the `SharedLink` OpenAPI schema was changed to closely resemble what has been implemented for goals. Objects are now clearly annotated, without the vague `data` attribute etc. I've also changed the creation endpoint route from `post` to `put`, to put emphasis on the idempotent behavior. 

There's some room for further refactoring of the `Plausible.Goals` module that I've given up for now -- I'm not super happy with the `Plausible.API.Goals` intersection, feels like the core contextual interfaces could use cleaner delegation paths, but the sheer amount of changes felt inappropriate for this iteration.

One thing that's missing from this PR is automatic enabling goals visibility on the dashboard, in case it's turned off. I'd like to hold off with that till the feature gates are implemented (for billing extensions) - so I can have better understanding of what's the best approach there.

Once live, this should be sufficient for the Wordpress Plugin 1.4.0 scope integration (next up on our end: auth tokens provisioning).

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/goals.ex
```diff
@@ -32,12 +32,54 @@ defmodule Plausible.Goals do
...
-      {:error, _failed_operation, failed_value, _changes_so_far} -> {:error, failed_value}
+      {:ok, %{goal: goal}} ->
+        {:ok, Repo.preload(goal, :site)}
+
+      {:error, _failed_operation, failed_value, _changes_so_far} ->
+        {:error, failed_value}
+    end
+  end
+
+  def find_or_create(site, %{
```
aerosol, line 43: This will enable exposing Revenue Goals creation in the Sites API too; to be addressed with tests in a separate PR.

aerosol:
CI failure unrelated, flaky test, tracked separately

lib/plausible/goals.ex
```diff
@@ -11,100 +11,92 @@ defmodule Plausible.Goals do
...
-        query
-      end
-
-    query =
-      if opts[:preload_funnels?] do
-        from(g in query,
-          left_join: assoc(g, :funnels),
-          group_by: g.id,
-          preload: [:funnels]
-        )
```
zoldar, line 100: FYI I have got rid of duplicate left join + group_by when doing the upsert refactor. (( 👀 👀 ))

aerosol: Beautiful 🤩

lib/plausible/plugins/api/goals.ex
```diff
@@ -0,0 +1,90 @@
...
+  def get(site, column, value) when column in [:event_name, :page_path] and is_binary(value) do
+    site
+    |> get_query()
+    |> where([g], field(g, ^column) == ^value)
+    |> Repo.one()
+  end
+
+  defp get_query(site) do
+    from g in Plausible.Goal,
+      inner_join: assoc(g, :site),
```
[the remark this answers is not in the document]
aerosol: https://github.com/plausible/analytics/pull/3396/commits/c173c9a008ceb79f9cff6bfebbb478b6f00debae

lib/plausible/goals.ex
aerosol: I still wanted to keep funnels preloading to include funnel metadata in the API later on, but I guess this can be done after upsert

zoldar: It's still here at https://github.com/plausible/analytics/pull/3396/files#diff-e4efb7868527fcba6528cc63067165514f2d9131fd85a665df2ec7cc470f6f00R91-R99

aerosol: That's `for_site` not the get or create call

aerosol: Nevermind, I see it wasn't there before 🫠

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -0,0 +1,544 @@
...
+
+      resp =
+        conn
+        |> json_response(422)
+        |> assert_schema("UnprocessableEntityError", spec())
+
+      assert [%{detail: "currency: is invalid"}] = resp.errors
+    end
+
+    test "edge case - revenue goals exists under the same name and different currency", %{
```
zoldar, line 144:
```suggestion
    test "edge case - revenue goal exists under the same name and different currency", %{
```

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -0,0 +1,544 @@
...
+      |> json_response(201)
+      |> assert_schema("Goal.ListResponse", spec())
+
+      initial_conn
+      |> put(url, %{goals: payload})
+      |> json_response(201)
+      |> assert_schema("Goal.ListResponse", spec())
+    end
+
+    test "edge case - revenue goals exists under the same name and different currency", %{
```
zoldar, line 323:
```suggestion
    test "edge case - revenue goals exist under the same name and different currency", %{
```

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -0,0 +1,544 @@
...
+        conn
+        |> authenticate(site.domain, token)
+        |> get(url)
+        |> json_response(422)
+        |> assert_schema("UnprocessableEntityError", spec())
+
+      assert %{errors: [%{detail: "Invalid integer. Got: string"}]} = resp
+    end
+
+    test "retrieve revenue goal by ID", %{conn: conn, site: site, token: token} do
```
zoldar, line 367:
```suggestion
    test "retrieves revenue goal by ID", %{conn: conn, site: site, token: token} do
```

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -0,0 +1,544 @@
...
+        |> json_response(200)
+        |> assert_schema("Goal", spec())
+        |> assert_schema("Goal.Revenue", spec())
+
+      assert resp.goal.id == goal.id
+      assert resp.goal_type == "Goal.Revenue"
+      assert resp.goal.display_name == "Purchase (EUR)"
+    end
+
+    test "retrieve pageview goal by ID", %{conn: conn, site: site, token: token} do
```
zoldar, line 386:
```suggestion
    test "retrieves pageview goal by ID", %{conn: conn, site: site, token: token} do
```

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -0,0 +1,544 @@
...
+        |> json_response(200)
+        |> assert_schema("Goal", spec())
+        |> assert_schema("Goal.Pageview", spec())
+
+      assert resp.goal.id == goal.id
+      assert resp.goal_type == "Goal.Pageview"
+      assert resp.goal.display_name == "Visit /checkout"
+    end
+
+    test "retrieve custom event goal by ID", %{conn: conn, site: site, token: token} do
```
zoldar, line 404:
```suggestion
    test "retrieves custom event goal by ID", %{conn: conn, site: site, token: token} do
```

zoldar (approved):
LGTM ✨ left only some minor corrections, like fixes of test names

lib/plausible/plugins/api/goals.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  @spec get(Plausible.Site.t(), pos_integer()) :: nil | Plausible.Goal.t()
+  def get(site, id) when is_integer(id) do
+    site
+    |> get_query()
+    |> where([g], g.id == ^id)
+    |> Repo.one()
+  end
+
+  @spec get(Plausible.Site.t(), String.t() | atom(), String.t()) :: nil | Plausible.Goal.t()
```
ukutaht, line 41: Sorry I'm a bit lost, what is this used for?

ukutaht (approved):

lib/plausible/plugins/api/goals.ex
aerosol: Thanks for noticing! The git log here doesn't show it, but I've started the implementation here with some prematurely added functions (I was expecting to use that clause in the "get or create" flow). Removed via 4bfb208bf

PR #3397, opened 2023-10-04
Refactor enterprise plan upgrade and change-plan actions

RobertJoonas:
### Changes

> NOTE: Reviewing commit-by-commit recommended

This PR is a refactor and does not change any behaviour for the users.

The objective of this PR is to make the code more understandable and get rid of the puzzling redirects between the different `/billing/**` routes. 

This PR replaces the two enterprise plan management routes:

* `/billing/upgrade/enterprise/:plan_id`
* `/billing/change-plan/enterprise/:plan_id`

with a single `/billing/upgrade-to-enterprise-plan` route that handles all the different cases in a single HEEx template:

* Upgrade for the first time (paddle JS checkout to provide billing details etc)
* Change enterprise plan (linking to the `/billing/change-plan/preview/:plan_id` route to confirm plan change)
* rendering a message to contact us when the user is already subscribed to the latest enterprise plan

This PR also builds extensive test coverage for the new page with different subscription status scenarios that were not tested for the old routes.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/billing.ex
```diff
@@ -183,12 +189,104 @@ defmodule PlausibleWeb.Components.Billing do
...
+    """
+  end
+
+  defp present_limit(enterprise_plan, key) do
+    enterprise_plan
+    |> Map.fetch!(key)
+    |> PlausibleWeb.StatsView.large_number_format()
+  end
+
+  def format_price(%Money{} = money, opts \\ []) do
```
aerosol, line 214: In case of Paddle unavailability we'll get a `nil` here via `{:error, :api_error}`. Shouldn't we persist/cache the last known value and make sure price formatting always receives `%Money{}`?

lib/plausible/billing/subscriptions.ex
```diff
@@ -19,4 +19,10 @@ defmodule Plausible.Billing.Subscriptions do
 
     cancelled? && expired?
   end
+
+  def resumable?(nil), do: false
+
+  def resumable?(%Subscription{status: status}) do
+    status in ["active", "past_due", "paused"]
```
aerosol, line 26: Consider providing macros to return all the individual status values to avoid compile-time typos.

```elixir
defmodule Plausible.Billing.Subscription.Status do
  defmacro __using__(_opts) do
    quote do
      require Plausible.Billing.Subscription.Status
      alias Subscription.Status
    end
  end

  defmacro active() do
    "active"
  end 
end
```

This should make it possible to pattern match still? 🤔

zoldar: Either this, or consider changing `status` field to [`Ecto.Enum`](https://hexdocs.pm/ecto/Ecto.Enum.html). It's a non-breaking change to the underlying DB, as internal status representation would remain as the same strings, while on Ecto schema level those would become atoms. Then `Ecto.Enum.mappings/2` and `Ecto.Enum.values/2` allow to retrieve the values and can be wrapped in convenient schema function(s).

lib/plausible/billing/subscription/status.ex
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  defmacro active(), do: :active
+  defmacro past_due(), do: :past_due
+  defmacro paused(), do: :paused
+  defmacro deleted(), do: :deleted
+
+  def valid_statuses() do
+    [active(), past_due(), paused(), deleted()]
+  end
```
zoldar, line 50: Absolutely nothing wrong with what's here, but there's a potential to make things a bit more DRY within the module 😄 

```suggestion
  @statuses [:active, :past_due, :paused, :deleted]
  
  @type status() :: unquote(Enum.reduce(@features, &{:|, [], [&1, &2]}))

  defmacro __using__(_opts) do
    quote do
      require Plausible.Billing.Subscription.Status
      alias Plausible.Billing.Subscription
    end
  end

  for status <- @statuses do
    defmacro unquote(status)(), do: unquote(status)
  end

  def valid_statuses() do
    @statuses
  end
```

lib/mix/tasks/cancel_subscription.ex
```diff
@@ -9,15 +9,16 @@ defmodule Mix.Tasks.CancelSubscription do
   use Mix.Task
   use Plausible.Repo
   alias Plausible.{Repo, Billing.Subscription}
+  use Plausible.Billing.Subscription.Status
```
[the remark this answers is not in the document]
RobertJoonas: Good point. I've made the aliases and requires more consistent in [this commit](https://github.com/plausible/analytics/pull/3397/commits/122e2f26a31e371ede11f222342a559ebdf79982). Also, removed the macro used only for hiding the alias.

lib/plausible/billing/plans.ex
```diff
@@ -134,6 +134,20 @@ defmodule Plausible.Billing.Plans do
...
+      Repo.one!(
+        from(e in EnterprisePlan,
+          where: e.user_id == ^user.id,
+          order_by: [desc: e.inserted_at],
+          limit: 1
+        )
+      )
+
+    enterprise_plan
+    |> Map.put(:price_per_interval, get_price_for(enterprise_plan))
```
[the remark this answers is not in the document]
RobertJoonas: Agreed. Done in https://github.com/plausible/analytics/pull/3397/commits/7fd60cae6ee6960255b2233086d517787dfcd672

lib/plausible/billing/subscription/status.ex
RobertJoonas: I like that suggestion. Done in https://github.com/plausible/analytics/pull/3397/commits/1bb8a39fdbe72f67806b04db40d75601743d984f

lib/plausible_web/controllers/billing_controller.ex
```diff
@@ -50,54 +48,64 @@ defmodule PlausibleWeb.BillingController do
...
+
   def upgrade_success(conn, _params) do
     render(conn, "upgrade_success.html", layout: {PlausibleWeb.LayoutView, "focus.html"})
   end
 
   def change_plan_form(conn, _params) do
-    user =
-      conn.assigns[:current_user]
-      |> Repo.preload(:enterprise_plan)
+    user = conn.assigns[:current_user]
```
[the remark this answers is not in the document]
RobertJoonas: Good point. On these billing paths we should be able to safely assume that a user is in the `assigns`. I changed it only for the two new routes since we're planning to get rid of the old ones anyway: https://github.com/plausible/analytics/pull/3397/commits/2c6ad252e600b3fce56cfdb5c339acd517b1fa92

lib/plausible_web/templates/auth/user_settings.html.heex
```diff
@@ -61,16 +63,16 @@
...
-        conn={@conn}
         business_tier={FunWithFlags.enabled?(:business_tier, for: @user)}
       />
       <div
         class="h-32 px-2 py-4 my-4 text-center bg-gray-100 rounded dark:bg-gray-900"
         style="width: 11.75rem;"
       >
         <h4 class="font-black dark:text-gray-100">Next bill amount</h4>
-        <%= if @subscription && @subscription.status in ["active", "past_due"] do %>
+        <%= if @subscription && @subscription.status in [Plausible.Billing.Subscription.Status.active(), Plausible.Billing.Subscription.Status.past_due()] do %>
```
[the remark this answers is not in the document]
RobertJoonas: Thanks for the suggestion, but I'm putting that on hold for now. This could definitely be improved, but as we just discussed with @vinibrsl it's time to move forward and merge this :)

PR #3401, opened 2023-10-05
Feature gates

vinibrsl:
This pull request implements limits to funnels, revenue goals and custom props based on the site owner plan. It extends the current "premium feature" notice to account for the new plans, trials and the on-going private preview. Stats API is not in the context of this pull request, but will be implemented likewise.

Reviewing commit-by-commit may help :)

Screenshot 2023-10-06 at 12 00 35

### DIFF
--- a/lib/plausible/billing/feature.ex
+++ b/lib/plausible/billing/feature.ex
@@ -66,7 +66,7 @@ defmodule Plausible.Billing.Feature do
 
   @doc false
   defmacro __using__(opts \\ []) do
-    quote location: :keep, bind_quoted: [opts: opts] do
+    quote location: :keep do
       @behaviour Plausible.Billing.Feature
       alias Plausible.Billing.Quota
 


lib/plausible/billing/quota.ex
```diff
@@ -134,6 +129,63 @@ defmodule Plausible.Billing.Quota do
     Plausible.Repo.one(query)
   end
 
+  @spec extra_features_usage(Plausible.Auth.User.t()) :: [atom()]
```
aerosol, line 132: I can't see that being used anywhere; in what context do we need to query all sites and all features?

lib/plausible/billing/quota.ex
```diff
@@ -134,6 +129,85 @@ defmodule Plausible.Billing.Quota do
...
+  defp check_extra_feature_usage({feature, query}, acc) do
+    if Plausible.Repo.exists?(query), do: [feature | acc], else: acc
+  end
+
+  @all_features [:props, :revenue_goals, :funnels]
+  @doc """
+  Returns a list of extra features the user can use. Trial users have the
+  ability to use all features during their trial.
+  """
+  def extra_features_limit(user) do
```
aerosol, line 174: `_limit` is a bit confusing here, it makes me think of usage counters. Perhaps "availability" would be better?

lib/plausible/billing/quota.ex
```diff
@@ -134,6 +129,85 @@ defmodule Plausible.Billing.Quota do
...
+        on: g.site_id == os.site_id,
+        where: not is_nil(g.currency)
+
+    queries = [
+      props: props_usage_query,
+      funnels: funnels_usage_query,
+      revenue_goals: revenue_goals_usage
+    ]
+
+    Plausible.Repo.transaction(fn ->
```
aerosol, line 160: Any reason for the transaction here?

lib/plausible/billing/plans.ex
```diff
@@ -31,6 +31,8 @@ defmodule Plausible.Billing.Plans do
   alias Plausible.Billing.{Subscription, Plan, EnterprisePlan}
   alias Plausible.Auth.User
 
+  @available_features ["props", "revenue_goals", "funnels"]
+
```
zoldar, line 35: I see this getting duplicated across a number of places. Perhaps it could be reworked into:

```elixir
@features [:props, :revenue_goals, :funnels]

# input processing can then be done via `Map.fetch(@feature_mappings, feature_str)` or the like
@feature_mappings Enum.into(@features, %{}, &{to_string(&1) => &1})

# that maps to `@type feature() :: :props | ....` and can be reused as a type wherever it's used as a public function input
@type feature() :: unquote(Enum.reduce(@features, &{:|, [], [&1, &2]}))

@spec features() :: [feature()]
def features do
  @features
end

File: lib/plausible/billing/quota.ex (hunk 2 above)
Comment (zoldar) [reply lib/plausible/billing/quota.ex:174]: Or maybe call it just `available_features` ?

File: test/support/factory.ex [hunk 5]
```diff
@@ -21,14 +21,21 @@ defmodule Plausible.Factory do
     }
   end
 
-  def site_factory do
+  def site_factory(attrs) do
```
RobertJoonas, line 24: Related to my other comment about extracting `create_user/1` and `create_site/1` into the setup block from individual tests. Maybe it would be cleaner to keep the low-level factory functions unaware of the schema associations and let `Plausible.TestUtils` handle that?

ukutaht:
by @aerosol 
> One thing I feel we could try to encapsulate is the `Feature` context itself. Wish we had some interface that makes it easier to extend this further as we add more features, something like:
> 
> WYT?

I like it. I thought of it when reviewing this controller action: https://github.com/plausible/analytics/blob/536f0d5b594ee71ec82491cb5dc90de5d85c0808/lib/plausible_web/controllers/site_controller.ex#L124-L160

Along with https://github.com/plausible/analytics/blob/536f0d5b594ee71ec82491cb5dc90de5d85c0808/lib/plausible/sites.ex#L305-L327
 
I wonder if the _Feature_ context could also help with the toggling code if we add `Goals` as a feature (never needs to be gated).

aerosol:
I don't think only gated features can exist in that context. As long as the underlying behaviour is common, "what's the availability?" ("always for all" is a valid answer), "toggle visbility" etc. We could think of an abstraction that captures those aspects.

Toggling conversions visibility however, means both revenue and non-revenue goals, but the former come with other constraints. So I wonder if keeping a single flag for both is the way to go. By the way of that PR you linked, IIRC I suggested keeping a schema-less db column rather than several boolean columns, perhaps that'd give us more flexibility in handling the quirks going forward. Maybe?

lib/plausible/billing/quota.ex
```diff
@@ -134,6 +129,85 @@ defmodule Plausible.Billing.Quota do
...
+  ability to use all features during their trial.
+  """
+  def extra_features_limit(user) do
+    user = Plausible.Users.with_subscription(user)
+
+    case Plans.get_subscription_plan(user.subscription) do
+      %Plausible.Billing.EnterprisePlan{} -> @all_features
+      %Plausible.Billing.Plan{extra_features: extra_features} -> extra_features
+      :free_10k -> []
+      nil -> @all_features
```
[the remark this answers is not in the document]
zoldar: I have created a small PR with a migration enforcing that constraint: https://github.com/plausible/analytics/pull/3403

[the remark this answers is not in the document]
aerosol: Merged, I guess a rebase is needed here.

lib/plausible_web/templates/site/settings_props.html.heex
```diff
@@ -1,15 +1,10 @@
...
-      href="https://plausible.io/docs/custom-props/introduction"
-      target="_blank"
-      rel="noreferrer"
-    >our docs</.link>.
-  </PlausibleWeb.Components.Generic.notice>
+  <PlausibleWeb.Components.Billing.extra_feature_notice
+    site={@site}
+    current_user={@current_user}
+    feature={:props}
+    grandfathered?
```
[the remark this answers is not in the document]
aerosol: > ps: boolean fields can be shortened in heex like that:

Oh, thanks, TIL! I couldn't find that in the docs.

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+    * `:display_name` - human-readable display name of the feature
+
+    * `:toggle_field` - the field in the %Plausible.Site{} schema that toggles
+    the feature. If `nil` or not set, toggle/2 silently returns `:ok`
+
+    * `:extra_feature` - an atom representing the feature name in the plan JSON
+    file (see also Plausible.Billing.Plan). If `nil` or not set,
+    `check_availability/1` silently returns `:ok`
+
+  Functions defined by `__using__` can be overriden if needed.
```
zoldar, line 21:
```suggestion
  Functions defined by `__using__` can be overridden if needed.
```

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+  @callback check_availability(Plausible.Site.t() | Plausible.Auth.User.t()) ::
+              :ok | {:error, :upgrade_required}
+
+  @features [
+    Plausible.Billing.Feature.Funnels,
+    Plausible.Billing.Feature.Goals,
+    Plausible.Billing.Feature.Props,
+    Plausible.Billing.Feature.RevenueGoals
+  ]
+  @doc """
```
zoldar, line 59:
```suggestion
  ]

  @doc """
```

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+  end
+
+  @doc false
+  defmacro __using__(opts \\ []) do
+    quote location: :keep, bind_quoted: [opts: opts] do
+      @behaviour Plausible.Billing.Feature
+      alias Plausible.Billing.Quota
+
+      @impl true
+      def display_name, do: Keyword.get(unquote(opts), :display_name)
```
zoldar, line 73: Doesn't using `unquote` on `opts` which are already bind_quoted defeat the purpose? 🤔 The idea is to only evaluate it once in the macro - `Keyword.get(opts, :display_name)` should work. Or was there a problem with it? It's not a big deal though as `opts` does not contain any computed expressions.

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+
+      @impl true
+      def toggle_field, do: Keyword.get(unquote(opts), :toggle_field)
```
zoldar, line 76: Ditto on unquoting something that is bind_quoted already

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+      @impl true
+      def check_availability(site_or_user)
+
+      def check_availability(%Plausible.Site{} = site) do
+        site = Plausible.Repo.preload(site, :owner)
+        check_availability(site.owner)
+      end
+
+      def check_availability(%Plausible.Auth.User{} = user) do
+        extra_feature = Keyword.get(unquote(opts), :extra_feature)
```
zoldar, line 96: Ditto on bind_quote

lib/plausible/billing/feature.ex
```diff
@@ -0,0 +1,155 @@
...
+        with key when not is_nil(key) <- toggle_field(),
+             :ok <- check_availability(site) do
+          override = Keyword.get(opts, :override)
+          toggle = if is_boolean(override), do: override, else: !Map.fetch!(site, toggle_field())
+
+          site
+          |> Ecto.Changeset.change(%{toggle_field() => toggle})
+          |> Plausible.Repo.update()
+        else
+          nil = _feature_not_toggable -> :ok
```
zoldar, line 117:
```suggestion
          nil = _feature_not_togglable -> :ok
```

zoldar (approved):
Only some minor typo fixes left ✨

PR #3404, opened 2023-10-10
Use default Phoenix.LiveView.Socket

ukutaht:
### Changes

Stops using our custom `PlausibleWeb.LiveSocket` in favour of the default `Phoenix.LiveView.Socket`.

Why?

We implemented a custom socket behaviour to fix a redirect loop which was happening due to cookies being placed on both `domain=plausible.io` and `domain=.plausible.io`. This effectively meant that HTTP and Websockets were sometimes reading different cookies.

We gave up on trying to work with that cookie setup and instead [renamed the session cookie](https://github.com/plausible/analytics/pull/3274) to start with a blank slate. Now cookies are always placed on `domain=.plausible.io` and HTTP and websocket see the same thing.

The default behaviour of `Phoenix.LiveView.Socket` is: if it fails to connect the websocket due to session being `nil`, it falls back to a full HTTP refresh. This makes sense now that cookies are set properly.


aerosol (approved):

ukutaht:
Added a commit that simplifies the `Plug.Session` setup as suggested here: https://github.com/elixir-plug/plug/issues/1176

lib/plausible_web/endpoint.ex
```diff
@@ -57,14 +64,7 @@ defmodule PlausibleWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
 
-  plug(PlausibleWeb.Plugs.RuntimeSessionAdapter, @session_options)
-
-  socket("/live", PlausibleWeb.LiveSocket,
-    websocket: [
-      check_origin: true,
```
aerosol, line 64: We should be checking the origin on websocket connections, it's currently missed.

PR #3411, opened 2023-10-11
Implement Stats API feature gate

vinibrsl:
This pull request implements a feature gate for the Stats API. Users that don't have access to the feature will see a notice in site settings and the API responds with Payment Required (HTTP 402).

I had to make some changes to make the new Feature context work without a %Site{}. Previously the extra feature notice required a %Site{} in order to check the owner plan. However, not every feature is scoped by site, for example the Stats API. (see also 638f6f8).

### DIFF
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -132,7 +132,8 @@ defmodule Plausible.Auth do
     params = %{name: name, user_id: user.id, key: key}
     changeset = Auth.ApiKey.changeset(%Auth.ApiKey{}, params)
 
-    Repo.insert(changeset)
+    with :ok <- Plausible.Billing.Feature.StatsAPI.check_availability(user),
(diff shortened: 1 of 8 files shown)


lib/plausible/auth/auth.ex
```diff
@@ -144,4 +145,22 @@ defmodule Plausible.Auth do
       {0, _} -> {:error, :not_found}
     end
   end
+
+  @spec find_api_key(String.t()) :: {:ok, Auth.ApiKey.t()} | {:error, :invalid_api_key}
+  def find_api_key(raw_key) do
```
RobertJoonas, line 150: I'd prefer to keep the argument name as `token` as it was, or `bearer_token` instead. The caller also passes a variable named `token` to this function.

RobertJoonas (approved):
Minor suggestion about a function argument name, but otherwise looks good to go 👍

PR #3424, opened 2023-10-13
Apply feature gates to dashboard queries

vinibrsl:
This pull request limits dashboard queries depending on the features available to the user. For example, users should only be able to filter by prop if the site owner has access to the feature. Similarly the CSV export should be filtered to include only reports the site owner has access to.


RobertJoonas (approved):
Looks good to me overall! 👍 Left two suggestions inline

PR #3430, opened 2023-10-17
update erlang version

RobertJoonas:
### Changes

Updates the Erlang version to one that is a supported release (25.3.2.7) in `kerl`, the Erlang installer.


ukutaht:
Make sure to also upgrade the [Dockerfile](https://github.com/plausible/analytics/blob/master/Dockerfile) with the new version.

aerosol:
image


Try bumping the cache version on CI - note that it's shared across all the PRs so it's going to break stuff, until this one is merged and everyone rebase against it, I think.

aerosol:
@RobertJoonas can I take this over? I'm upgrading Elixir and major Erlang bump is in the books for it. I'll ping you in the PR so you can check if you're good on macos. How does that sound?

PR #3440, opened 2023-10-18
Plugins API: handle Revenue Goals creation wrt Business Tier

aerosol:
### Changes

This PR enables error handling on revenue goal creation constrained by subscription/business tier rules.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugins/api/controllers/goals.ex
+++ b/lib/plausible_web/plugins/api/controllers/goals.ex
@@ -58,11 +58,7 @@ defmodule PlausibleWeb.Plugins.API.Controllers.Goals do
         |> render("goal.json", goal: goal, authorized_site: site)
 
       {:error, :upgrade_required} ->
-        Errors.error(
-          conn,
-          402,
-          "#{Plausible.Billing.Feature.RevenueGoals.display_name()} is part of the Plausible Business plan. To get access to this feature, please upgrade your account."
-        )
+        payment_required(conn)
 
       {:error, changeset} ->
         Errors.error(conn, 422, changeset)
@@ -84,6 +80,9 @@ defmodule PlausibleWeb.Plugins.API.Controllers.Goals do
         |> put_view(Views.Goal)
         |> put_status(:created)
         |> render("index.json", goals: goals, authorized_site: site)
+
+      {:error, :upgrade_required} ->
+        payment_required(conn)
 
       {:error, changeset} ->
         Errors.error(conn, 422, changeset)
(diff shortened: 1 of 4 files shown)


test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -44,6 +44,35 @@ defmodule PlausibleWeb.Plugins.API.Controllers.GoalsTest do
...
+  describe "buisness tier" do
+    test "fails on revenue goal creation attempt with insufficient plan", %{
+      site: site,
+      token: token,
+      conn: conn
+    } do
+      site = Plausible.Repo.preload(site, :owner)
+      FunWithFlags.enable(:business_tier, for_actor: site.owner)
+
+      insert(:subscription, paddle_plan_id: "free_10k", user: site.owner)
```
aerosol, line 56: @vinibrsl is there a better way of preparing this data?

lib/plausible_web/plugins/api/controllers/goals.ex
```diff
@@ -56,6 +57,13 @@ defmodule PlausibleWeb.Plugins.API.Controllers.Goals do
         |> put_resp_header("location", goals_url(base_uri(), :get, goal.id))
         |> render("goal.json", goal: goal, authorized_site: site)
 
+      {:error, :upgrade_required} ->
```
zoldar, line 60: Shouldn't that be handled for `create` on multiple `goals` (action clause below)?

zoldar: There's also another spot, `PlausibleWeb.Live.GoalSettings.Form` (`handle_event` for `save-goal`) - I'm not sure, maybe it's handled by some plug earlier in the pipeline here, maybe?

aerosol: thanks! da5253b41

aerosol: > There's also another spot, PlausibleWeb.Live.GoalSettings.Form (handle_event for save-goal) - I'm not sure, maybe it's handled by some plug earlier in the pipeline here?

Ah right. It's annoying that's not a changeset error, but then I guess it's harder to distinguish whether it's subscription related.

aerosol: I think we can ignore it? Looks like the form is disabled.

image

image

zoldar (approved):

PR #3444, opened 2023-10-19
Plan benefits

RobertJoonas:
### Changes

This PR starts listing the different plan benefits (including limits and supported features) on the upgrade page.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/billing/plans.ex
```diff
@@ -43,6 +43,7 @@ defmodule Plausible.Billing.Plans do
     cond do
       Application.get_env(:plausible, :environment) == "dev" -> plans_sandbox()
       !owned_plan -> if v4_available, do: @plans_v4, else: @plans_v3
+      owned_plan.kind == :business -> @plans_v4
```
RobertJoonas, line 46: This check here makes sure that even when the user is on the new "duplicate" v3 Business plan, we will only allow them to downgrade to Growth v4 (and not v3).

To summarize, migrating from:

* v3 Growth &rarr; v3 Business is possible ✅
* v3 Growth &rarr; v4 Business is impossible ⛔️
* v3 Business &rarr; v4 Growth is possible ✅
* v3 Business &rarr; v3 Growth is impossible ⛔️

lib/plausible_web/live/choose_plan.ex
```diff
@@ -620,6 +632,77 @@ defmodule PlausibleWeb.Live.ChoosePlan do
     PlausibleWeb.StatsView.large_number_format(volume)
   end
 
+  defp growth_benefits(plan) do
```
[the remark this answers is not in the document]
RobertJoonas: I like that suggestion! But I agree that we can consider it a post-release todo item.

lib/plausible/billing/plan.ex
```diff
@@ -0,0 +1,93 @@
...
+            volume: String.t(),
+            monthly_cost: Money.t() | nil,
+            yearly_cost: Money.t() | nil,
+            monthly_product_id: String.t() | nil,
+            yearly_product_id: String.t() | nil,
+            features: [atom()]
+          }
+          | :enterprise
+
+  def new(params, file_name) when is_map(params) do
```
ukutaht, line 24: This is probably my OOP background talking but I always expect a function called `new` to be super basic and not do any work/validation/transformations.

IMO good options here would be:
1. A function `new` that takes a map with all the fields pre-processed already
2. A function `from_json!` that takes the result of `File.read` and does all the rest
3. A function `from_file!` that takes the filename and does everything including `File.read`

I used the exclamation mark for the last two because there's plenty of opportunity for both to crash loudly.

ukutaht (approved):
One naming comment, otherwise looks good. The code to list benefits is a bit complex but I can't think of a way to simplify.

lib/plausible/billing/plan.ex
RobertJoonas: I see what you mean. I wanted to separate the "building of a plan" into the Plan module context. With functions like `from_json!` and `from_file!` we'd still need to Enum.map the list of plans, so I figured a function dealing with `raw_params` (i.e. decoded json map) would be more convenient.

I just pushed two commits to rename the function `new` to `build!`, still keeping the `new` function around that only casts the params map into a struct, and using exclamation marks where applicable to indicate that the function will crash with unexpected input.

PR #3445, opened 2023-10-19
Check for limits and features used by site before transferring ownership

vinibrsl:
This pull request adds extra checks when transferring ownership to ensure the new owner subscription plan can accommodate the new site. The previous and new owner have to coordinate with CS to have the new owner downsize their sites, or have the old owner turn off premium features. From an implementation point we'll keep it simple - no ability to accept a transfer if it doesn't fit with your current subscription plan.

### DIFF
--- a/lib/plausible/site/admin.ex
+++ b/lib/plausible/site/admin.ex
@@ -82,10 +82,11 @@ defmodule Plausible.SiteAdmin do
 
     if new_owner do
       {:ok, _} =
-        Plausible.Site.Memberships.CreateInvitation.bulk_transfer_ownership(
+        Plausible.Site.Memberships.bulk_create_invitation(
           sites,
           inviter,
           new_owner.email,
+          :owner,
           check_permissions: false
         )
 
@@ -103,10 +104,7 @@ defmodule Plausible.SiteAdmin do
     new_owner = Plausible.Auth.find_user_by(email: email)
 
     if new_owner do
-      case Plausible.Site.Memberships.CreateInvitation.bulk_transfer_ownership_direct(
-             sites,
-             new_owner
-           ) do
+      case Plausible.Site.Memberships.bulk_transfer_ownership_direct(sites, new_owner) do
         {:ok, _} -> :ok
         {:error, :transfer_to_self} -> {:error, "User is already an owner of one of the sites"}
       end
--- a/lib/plausible/site/memberships/create_invitation.ex
+++ b/lib/plausible/site/memberships/create_invitation.ex
@@ -17,7 +17,7 @@ defmodule Plausible.Site.Memberships.CreateInvitation do
           | :forbidden
           | :upgrade_required
 
-  @spec invite(Site.t(), User.t(), String.t(), atom()) ::
+  @spec create_invitation(Site.t(), User.t(), String.t(), atom()) ::
           {:ok, Invitation.t()} | {:error, invite_error()}
   @doc """
   Invites a new team member to the given site. Returns a
@@ -30,7 +30,7 @@ defmodule Plausible.Site.Memberships.CreateInvitation do
   If the new team member role is `:owner`, this function handles the invitation
   as an ownership transfer and requires the inviter to be the owner of the site.
   """
-  def invite(site, inviter, invitee_email, role) do
+  def create_invitation(site, inviter, invitee_email, role) do
     Plausible.Repo.transaction(fn ->
       do_invite(site, inviter, invitee_email, role)
     end)
@@ -52,12 +52,12 @@ defmodule Plausible.Site.Memberships.CreateInvitation do
     end)
   end
 
-  @spec bulk_transfer_ownership([Site.t()], User.t(), String.t(), Keyword.t()) ::
+  @spec bulk_create_invitation([Site.t()], User.t(), String.t(), atom(), Keyword.t()) ::
           {:ok, [Invitation.t()]} | {:error, invite_error()}
-  def bulk_transfer_ownership(sites, inviter, invitee_email, opts \\ []) do
+  def bulk_create_invitation(sites, inviter, invitee_email, role, opts \\ []) do
     Plausible.Repo.transaction(fn ->
       for site <- sites do
-        do_invite(site, inviter, invitee_email, :owner, opts)
+        do_invite(site, inviter, invitee_email, role, opts)
       end
     end)
   end
--- a/lib/plausible_web/controllers/site/membership_controller.ex
+++ b/lib/plausible_web/controllers/site/membership_controller.ex
@@ -39,7 +39,7 @@ defmodule PlausibleWeb.Site.MembershipController do
     site_domain = conn.assigns[:site].domain
     site = Sites.get_for_user!(conn.assigns[:current_user].id, site_domain)
 
-    case Memberships.CreateInvitation.invite(site, conn.assigns.current_user, email, role) do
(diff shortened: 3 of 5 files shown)


lib/plausible/sites.ex
```diff
@@ -142,11 +145,42 @@ defmodule Plausible.Sites do
     Plausible.Mailer.send(email)
   end
 
-  defp ensure_transfer_valid(site, invitee, :owner) do
-    if invitee && role(invitee.id, site) == :owner do
-      {:error, :transfer_to_self}
-    else
-      :ok
+  defp within_team_member_limit_after_transfer?(site, new_owner) do
```
RobertJoonas, line 148: Not sure if this is possible, but how about wrapping the ownership transfer in an Ecto transaction and rolling it back in case any limits were exceeded? I believe this would help us keep a leash on the code complexity here and stick to the existing functions for checking the limits. WDYT?

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -111,6 +111,16 @@ defmodule PlausibleWeb.Site.MembershipController do
         |> put_flash(:error, "Can't transfer ownership to existing owner")
         |> redirect(to: Routes.site_path(conn, :settings_people, site.domain))
 
+      {:error, :upgrade_required} ->
+        conn
+        |> put_flash(:ttl, :timer.seconds(5))
+        |> put_flash(:error_title, "Transfer error")
+        |> put_flash(
+          :error,
+          "The site you're trying to transfer exceeds the invitee's subscription plan. To proceed, please contact us at hello@plausible.io for further assistance."
```
RobertJoonas, line 120: I'd consider returning a more explicit error message to the user about exactly what limits would be exceeded with the transfer. In code, we know exactly what went wrong, but if the customer contacts our support, it will probably take us a while to figure it out manually.

lib/plausible/sites.ex
zoldar: Maybe it would be a good opportunity to extract invite creation to a dedicated service module, like https://github.com/plausible/analytics/blob/e9251aebfaae44001bf13cf4bc049b989e94f974/lib/plausible/site/memberships/accept_invitation.ex ?

lib/plausible/billing/quota.ex
```diff
@@ -108,27 +108,36 @@ defmodule Plausible.Billing.Quota do
...
+        select: %{email: u.email, site_id: os.site_id, role: sm.role}
 
-    invitations_and_team_members_query =
+    team_members_and_invitations_query =
       from i in Plausible.Auth.Invitation,
         inner_join: os in subquery(owned_sites_query(user)),
         on: i.site_id == os.site_id,
-        where: i.role != :owner,
-        select: %{email: i.email},
+        select: %{email: i.email, site_id: i.site_id, role: i.role},
```
zoldar, line 127: Should role be treated uniquely given the same user and site combination? 🤔 Even if there happen to be invites on a different role, accepting it will modify existing membership, not create a new one.

lib/plausible/sites.ex
```diff
@@ -142,11 +145,42 @@ defmodule Plausible.Sites do
...
+    features_to_check = [
+      Plausible.Billing.Feature.Props,
+      Plausible.Billing.Feature.RevenueGoals,
+      Plausible.Billing.Feature.Funnels
+    ]
+
+    Enum.all?(features_to_check, fn feature ->
+      if feature.enabled?(site), do: feature.check_availability(new_owner) == :ok, else: true
+    end)
+  end
```
zoldar, line 175: Perhaps that could be abstraced in `Plausible.Billing.Features` to something like `Features.available?(site, new_owner, [Feature.Props, Feature.RevenueGoals, Feature.Funnels])`?

lib/plausible/billing/quota.ex
```diff
@@ -108,27 +108,35 @@ defmodule Plausible.Billing.Quota do
...
+        where: sm.role != :owner,
+        select: u.email
+
+    from i in Plausible.Auth.Invitation,
+      inner_join: os in subquery(owned_sites_query),
+      on: i.site_id == os.site_id,
+      where: i.role != :owner,
+      select: i.email,
+      union: ^team_members_query,
+      distinct: true
```
zoldar, line 139: AFAIK this `distinct` is redundant as `union` will already make results unique (`union_all` does not).

lib/plausible/site/admin.ex
```diff
@@ -100,7 +103,10 @@ defmodule Plausible.SiteAdmin do
     new_owner = Plausible.Auth.find_user_by(email: email)
 
     if new_owner do
-      case Plausible.Sites.bulk_transfer_ownership_direct(sites, new_owner) do
+      case Plausible.Site.Memberships.CreateInvitation.bulk_transfer_ownership_direct(
```
zoldar, line 106: Service functions could be delegated via https://github.com/plausible/analytics/blob/72df112589019859d51807ae00a5e74e24369155/lib/plausible/site/memberships.ex#L12-L15, treating `Plausible.Site.Memberships` as a context, sort of 😅

lib/plausible/site/memberships/create_invitation.ex
```diff
@@ -0,0 +1,181 @@
+defmodule Plausible.Site.Memberships.CreateInvitation do
```
zoldar, line 1: 👏

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -39,7 +39,7 @@ defmodule PlausibleWeb.Site.MembershipController do
     site_domain = conn.assigns[:site].domain
     site = Sites.get_for_user!(conn.assigns[:current_user].id, site_domain)
 
-    case Sites.invite(site, conn.assigns.current_user, email, role) do
+    case Memberships.CreateInvitation.invite(site, conn.assigns.current_user, email, role) do
```
zoldar, line 42: Ditto about delegating via `Plausible.Site.Memberships`

lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -98,7 +98,7 @@ defmodule PlausibleWeb.Site.MembershipController do
     site_domain = conn.assigns[:site].domain
     site = Sites.get_for_user!(conn.assigns[:current_user].id, site_domain)
 
-    case Sites.invite(site, conn.assigns.current_user, email, :owner) do
+    case Memberships.CreateInvitation.invite(site, conn.assigns.current_user, email, :owner) do
```
zoldar, line 101: Ditto about delegating via `Plausible.Site.Memberships`

test/plausible/site/sites_test.exs
```diff
@@ -68,310 +67,6 @@ defmodule Plausible.SitesTest do
     end
   end
 
-  describe "invite/4" do
```
zoldar, line 71: Some tests strictly focused on result handling in controller (redirects, flash messages) for success cases and each kind of error case (when possible), would still be nice to have here IMHO.

zoldar (approved):
Nice work ✨  Left some optional 💅  feedback

test/plausible/site/sites_test.exs
zoldar: Yeah, sorry, I meant testing this thin layer of controller logic _on top_ of service 😅 I know it's usually not too complex but still, nice to have confirmation that it's integrated correctly. But no worries, coverage of service logic is the most importatnt thing here for sure, and that's ensured here 👍

PR #3454, opened 2023-10-23
Upgrade Erlang/Elixir stack

aerosol:
### Changes

A joint co-op with @zoldar 
This PR bumps Erlang to 26.1.2 and Elixir to 1.15.7

The enormous diffstat is only due to rewriting exvcr casettes to JSON with newlines, don't let that freak you out.

some deps updated, notably exvcr was removed, it doesn't play well with otp26 and we weren't using it sensibly anyway - @vinibrsl and I were planning to get rid of it in favor of Mox long time ago; the time has come.

public suffix library updated, also ingest handles ipv4 hostnames properly now.
sessions will likely reset for some (the hash will be different for domains like xxx.netlify.app etc.).

Locus can be now configured to cache maxmind licensed db, useful in case maxmind is unavailable during app startup. We have a persistent volume mounted to keep that.

More tests tagged as `:slow` to speed up local feedback loop.

### Tests

### Changelog

### Documentation

### Dark mode


test/plausible/google/import_integration_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Plausible.Google.Api.ImportIntegrationTest do
+  use Plausible.DataCase, async: false
```
zoldar, line 2: I'm not 100% sure but I think this test suite could be run async (at least mox should be safe to run in that mode and I don't see us doing any env patching here)? It would slitghtly shorten CI run.

zoldar (approved):

config/runtime.exs
```diff
@@ -164,10 +164,11 @@ ip_geolocation_db = get_var_from_path_or_env(config_dir, "IP_GEOLOCATION_DB", ge
 geonames_source_file = get_var_from_path_or_env(config_dir, "GEONAMES_SOURCE_FILE")
 maxmind_license_key = get_var_from_path_or_env(config_dir, "MAXMIND_LICENSE_KEY")
 maxmind_edition = get_var_from_path_or_env(config_dir, "MAXMIND_EDITION", "GeoLite2-City")
+maxmind_cache_dir = get_var_from_path_or_env(config_dir, "PERSISTENT_CACHE_DIR")
```
ruslandoga, line 167: Maybe `STORAGE_DIR` could be reused? We can add some locus-specific path to avoid collisions.

config/runtime.exs
```diff
@@ -560,6 +561,7 @@ geo_opts =
       [
         license_key: maxmind_license_key,
         edition: maxmind_edition,
+        cache_dir: maxmind_cache_dir,
```
ruslandoga, line 564: I think it would need to be optional, in case the env var is not set.

config/runtime.exs
ruslandoga: Oops, sorry. There is no `STORAGE_DIR`, it's something I added in a separate PR https://github.com/plausible/analytics/pull/2986/files#r1209884699

config/runtime.exs
ruslandoga: Ah, it's checked in `Plasuible.Geo`.

ruslandoga (approved):

config/runtime.exs
aerosol: Feel free to rebase/merge that one

.tool-versions
```diff
@@ -1,4 +1,2 @@
-erlang 25.2.3
-elixir 1.14.3-otp-25
-nodejs 18.17.1
```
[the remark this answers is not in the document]
aerosol: Ah right, hose are installed separately on build. Let's see if people have issues building locally.

test/plausible/google/import_integration_test.exs
aerosol: Not sure, the import buffer is involved

.tool-versions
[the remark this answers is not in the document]
aerosol: I'm going to add latest node, it's important to have uniform assets building locally; python is optional though, for git-hook only

test/plausible/google/import_integration_test.exs
aerosol: looks good, did several runs of 97f126344

.tool-versions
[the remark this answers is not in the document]
aerosol: ab9af4fc1

ukutaht (approved):
Looks good to me 👍

PR #3461, opened 2023-10-25
Restrict subscribing to a plan when exceeding its limits + warning for losing feature access

RobertJoonas:
### Changes

* On the new upgrade page, disable the upgrade/change-plan link and render an error message beneath it saying "Your usage exceeds this plan", if the user's usage exceeds any of the following limits:
  * monthly pageview limit (based on the last 30 days)
  * team member limit
  * site limit

* On the new upgrade page, use the built-in browser confirmation popup for displaying a warning about losing access to features.

* Also, on the API level, restrict subscribing to a plan that does not accommodate usage (also based on all the three limits mentioned above)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -1,7 +1,7 @@
 defmodule Plausible.Billing do
   use Plausible.Repo
   require Plausible.Billing.Subscription.Status
-  alias Plausible.Billing.Subscription
+  alias Plausible.Billing.{Subscription, Plans, Quota}
 
   @spec active_subscription_for(integer()) :: Subscription.t() | nil
   def active_subscription_for(user_id) do
@@ -40,6 +40,13 @@ defmodule Plausible.Billing do
   def change_plan(user, new_plan_id) do
     subscription = active_subscription_for(user.id)
 
+    case Quota.ensure_can_subscribe_to_plan(user, Plans.find(new_plan_id)) do
+      :ok -> do_change_plan(subscription, new_plan_id)
+      error -> error
+    end
+  end
+
+  defp do_change_plan(subscription, new_plan_id) do
     res =
       paddle_api().update_subscription(subscription.paddle_subscription_id, %{
         plan_id: new_plan_id
--- a/lib/plausible/billing/quota.ex
+++ b/lib/plausible/billing/quota.ex
@@ -166,6 +166,21 @@ defmodule Plausible.Billing.Quota do
     end)
   end
 
+  def ensure_can_subscribe_to_plan(user, %Plan{} = plan) do
+    usage = %{
+      monthly_pageviews: monthly_pageview_usage(user),
+      team_members: team_member_usage(user),
+      sites: site_usage(user)
+    }
+
+    case exceeded_limits(usage, plan) do
+      [] -> :ok
+      exceeded_limits -> {:error, %{exceeded_limits: exceeded_limits}}
+    end
+  end
+
+  def ensure_can_subscribe_to_plan(_user, nil), do: :ok
+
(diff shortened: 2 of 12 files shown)


lib/plausible/billing/quota.ex
```diff
@@ -162,10 +162,36 @@ defmodule Plausible.Billing.Quota do
     ]
 
     Enum.reduce(queries, [], fn {feature, query}, acc ->
-      if Plausible.Repo.exists?(query), do: [feature | acc], else: acc
+      if Plausible.Repo.exists?(query), do: acc ++ [feature], else: acc
```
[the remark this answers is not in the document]
RobertJoonas: This is actually just a personal preference thing - I figured it's easier to follow in code, if the order of the listed features stays the same around the codebase. E.g. the features returned from `Feature.list()`.

lib/plausible_web/live/choose_plan.ex
```diff
@@ -23,7 +23,12 @@ defmodule PlausibleWeb.Live.ChoosePlan do
         Users.with_subscription(user_id)
       end)
       |> assign_new(:usage, fn %{user: user} ->
-        Quota.monthly_pageview_usage(user)
+        %{
+          monthly_pageviews: Quota.monthly_pageview_usage(user),
+          team_members: Quota.team_member_usage(user),
+          sites: Quota.site_usage(user),
+          features: Quota.features_usage(user)
+        }
```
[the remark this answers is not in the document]
RobertJoonas: Good call! https://github.com/plausible/analytics/pull/3461/commits/9fb695c20def4f3489db66bf1a9cdfe97d06224d

lib/plausible/billing/billing.ex
```diff
@@ -39,7 +39,13 @@ defmodule Plausible.Billing do
 
   def change_plan(user, new_plan_id) do
     subscription = active_subscription_for(user.id)
+    plan = Plans.find(new_plan_id)
```
zoldar, line 42: I'm assuming missing plan is meant to be a pass-through case for custom subscriptions that are only defined at Paddel and that's why `Quote.ensure_can_subscribe_to_plan` returns `:ok` for `plan` = `nil` right?

lib/plausible/billing/quota.ex
```diff
@@ -162,10 +177,30 @@ defmodule Plausible.Billing.Quota do
...
+
+  def ensure_can_subscribe_to_plan(_user, nil), do: :ok
+
+  def exceeded_limits(usage, %Plan{} = plan) do
+    for {usage_field, limit_field} <- [
+          {:monthly_pageviews, :monthly_pageview_limit},
+          {:team_members, :team_member_limit},
+          {:sites, :site_limit}
+        ],
+        !within_limit?(Map.get(usage, usage_field), Map.get(plan, limit_field)) do
```
zoldar, line 199:
```suggestion
        !within_limit?(Map.fetch!(usage, usage_field), Map.fetch!(plan, limit_field)) do
```

just in case, to not let it fail silently and cause issues down the line perhaps?

zoldar (approved):

lib/plausible/billing/billing.ex
RobertJoonas: Exactly. in case `plan = nil` we didn't find a "regular" plan corresponding to this plan_id, so that means it's an enterprise plan.

lib/plausible/billing/quota.ex
RobertJoonas: Sorry, just merged a sec before seeing your comment. It's a fair point. I can make a note and change in the next PR perhaps

PR #3463, opened 2023-10-26
Implement new sites view

zoldar:
### Changes

This PR includes two additions to the /sites view:

1. Ability to live search sites list by domain
2. Render a small plot of 24h visits for each site card, with 48h change indicator

To accomplish this, the /sites route was moved to LiveView entirely. The page handles defunct websocket state mostly gracefully (search and pagination will be still usable in case LV is unavailable).

### Tests

### Changelog

### Documentation

### Dark mode


aerosol (approved):
lgtm but I'm biased :D

ukutaht:
Reviewing purely the UI, haven't looked at the code yet.

1. The graph is beautiful but the colours feel a little bit off to me.

Current:
Screenshot 2023-10-30 at 17 27 56

I tried aligning this with the main graph in the dashboard on localhost. Grabbed the gradient color from [here](https://github.com/plausible/analytics/blob/8cc7bce689b04c5a3e346f57da8aa0e4cc0878b2/assets/js/dashboard/stats/graph/graph-util.js#L91-L92) and the line color from [here](https://github.com/plausible/analytics/blob/8cc7bce689b04c5a3e346f57da8aa0e4cc0878b2/assets/js/dashboard/stats/graph/graph-util.js#L81).  I also made the line use `strokeWidth="2.6"`  instead of 3 to make it lighter. The result looks a bit more aligned with the dashboard IMO:
Screenshot 2023-10-30 at 17 28 16
Screenshot 2023-10-30 at 17 31 49


2. On the previous screenshots it's clear that the line gets cut off at the top, at least with the data I have (1 visitor in one hour in the last 24h)
3. Another inconsistency with the dashboard - on the dashboard the comparison arrow is shown in green but the number and percentage sign are normal text color. On /sites the number and percentage sign are also green. I think it would make sense to use the same styling on /sites as in the dashboard.

ukutaht:
On the search input:

I think the search input looks better without any shadow here. 



With shadow:

Screenshot 2023-10-30 at 17 45 00

Without shadow:
Screenshot 2023-10-30 at 17 44 53

Autofocusing on the search input feels quite unexpected - I checked every site with a search box I could find and not one of them uses autofocus on the search box. Besides just being unexpected, it breaks some basic default behaviours like being able to use arrow down key to scroll down. What's the reason for that?

ukutaht:
I think the search box looks better without any shadow applied as well - the site boxes are by far the most important information on the screen and removing the shadow from search helps highlight them.

zoldar:
> I tried aligning this with the main graph in the dashboard on localhost. Grabbed the gradient color from [here](https://github.com/plausible/analytics/blob/8cc7bce689b04c5a3e346f57da8aa0e4cc0878b2/assets/js/dashboard/stats/graph/graph-util.js#L91-L92) and the line color from [here](https://github.com/plausible/analytics/blob/8cc7bce689b04c5a3e346f57da8aa0e4cc0878b2/assets/js/dashboard/stats/graph/graph-util.js#L81). I also made the line use strokeWidth="2.6" instead of 3 to make it lighter. The result looks a bit more aligned with the dashboard IMO:

Colors and stroke width got adjusted according to suggestions.

> Another inconsistency with the dashboard - on the dashboard the comparison arrow is shown in green but the number and percentage sign are normal text color. On /sites the number and percentage sign are also green. I think it would make sense to use the same styling on /sites as in the dashboard.

Fixed, now styling in /sites is following what's in the dashboard.

> Autofocusing on the search input feels quite unexpected - I checked every site with a search box I could find and not one of them uses autofocus on the search box. Besides just being unexpected, it breaks some basic default behaviours like being able to use arrow down key to scroll down. What's the reason for using autofocus here?

> Ah. We forgot to add / key handling.

Autofocus got removed and typing "/" now focuses on search. While focused, pressing ESC will remove focus _and_ reset search.

> I think the search box looks better without any shadow applied as well - the site boxes are by far the most important information on the screen and removing the shadow from search helps highlight them.

The shadow got removed completely.

aerosol:
>     2. On the previous screenshots it's clear that the line gets cut off at the top, at least with the data I have (1 visitor in one hour in the last 24h)


@ukutaht can you still replicate that?

image

ukutaht:
@zoldar @aerosol looks great!

I cannot replicate the cut-off SVG anymore.

One more thing - when no results are found, it looks quite bare. A message like 'no results found, please search for something else' would be helpful.

Screenshot 2023-10-31 at 11 01 14


I'll take a look at the code now

aerosol:
@ukutaht we're working on some refinements, will be pushed shortly

lib/plausible/stats/clickhouse.ex
```diff
@@ -213,6 +215,98 @@ defmodule Plausible.Stats.Clickhouse do
     |> Enum.into(%{})
   end
 
+  @spec last_24h_visitors_hourly_intervals([Plausible.Site.t()], NaiveDateTime.t()) :: map()
+  def last_24h_visitors_hourly_intervals(sites, now \\ NaiveDateTime.utc_now())
```
ukutaht, line 219: I can't find any references to the old `Plausible.Stats.Clickhouse.last_24h_visitors/1`. Can be removed?

lib/plausible/stats/clickhouse.ex
```diff
@@ -213,6 +215,98 @@ defmodule Plausible.Stats.Clickhouse do
...
+        group_by: [e.site_id, fragment("toStartOfHour(timestamp)")],
+        order_by: [e.site_id, fragment("toStartOfHour(timestamp)")]
+      )
+
+    result =
+      current_query
+      |> ClickhouseRepo.all()
+      |> Enum.group_by(& &1.site_id)
+      |> Enum.map(fn {site_id, entries} ->
+        visitors =
```
ukutaht, line 277: If a single `user_id` visits the site at `15:59` and `16:00` they will be double-counted with this method. Failing test to illustrate the point:

```elixir
test "calculates uniques correctly across hour boundaries" do
  fixed_now = ~N[2023-10-26 10:00:15]
  site = insert(:site)

  user_id = 111

  populate_stats(site, [
    build(:pageview, user_id: user_id, timestamp: ~N[2023-10-25 15:59:00]),
    build(:pageview, user_id: user_id, timestamp: ~N[2023-10-25 16:00:00])
  ])

  result = Clickhouse.last_24h_visitors_hourly_intervals([site], fixed_now)[site.domain]
  assert result[:visitors] == 1
end
```

lib/plausible_web/live/sites.ex
```diff
@@ -0,0 +1,490 @@
...
+      ("/sites?" <> URI.encode_query(Map.take(params, ["filter_text"])))
+      |> URI.new!()
+
+    socket =
+      socket
+      |> assign(:uri, uri)
+      |> assign(:filter_text, params["filter_text"] || "")
+      |> assign(:params, params)
+      |> assign(:user, Repo.get!(Auth.User, user_id))
+      |> load_sites()
```
ukutaht, line 35: Trying to be mindful of Clickhouse SELECTs here since it's our main bottleneck. I think it's worth only loading sites on mount if `connected?(socket)` condition is true, and coming up with a nice loading state when `!connected?(socket)` to reduce the amount of SELECTs on Clickhouse.

This page probably doesn't stand for much load compared to the dashboard and API. But I also don't see a reason to be wasteful here.

lib/plausible_web/live/sites.ex
```diff
@@ -0,0 +1,490 @@
...
+        >
+          <div class="bg-white dark:bg-gray-800 px-4 pt-5 pb-4 sm:p-6 sm:pb-4">
+            <div class="hidden sm:block absolute top-0 right-0 pt-4 pr-4">
+              <button
+                x-on:click="invitationOpen = false"
+                class="bg-white dark:bg-gray-800 rounded-md text-gray-400 dark:text-gray-500 hover:text-gray-500 dark:hover:text-gray-400 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500"
+              >
+                <span class="sr-only">Close</span>
+                <!-- Heroicon name: outline/x -->
+                <svg
```
ukutaht, line 312: Opportunity to use `Heroicons` library here instead of inlined `svg`?

lib/plausible_web/live/sites.ex
```diff
@@ -0,0 +1,490 @@
...
+          <input
+            type="text"
+            name="filter_text"
+            id="filter-text"
+            phx-debounce={200}
+            class="pl-8 dark:bg-gray-900 dark:text-gray-300 focus:ring-indigo-500 focus:border-indigo-500 block w-full sm:text-sm border-gray-300 dark:border-gray-500 rounded-md dark:bg-gray-800"
+            placeholder="Type / to search sites"
+            value={@filter_text}
+            x-ref="filter_text"
+            x-on:keydown.escape="$refs.filter_text.blur(); $refs.reset_filter.dispatchEvent(new Event('click', {bubbles: true, cancelable: true}));"
```
ukutaht: Painful that [Phoenix.LiveView.JS](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.JS.html) doesn't have a `blur()` function. I think everything here could be accomplished with plain `LiveView.JS` if it did.

lib/plausible_web/live/sites.ex
zoldar: We specifically wanted the site to be functional even if websocket is broken for whatever reason. Currently, when WSS socket failure is emulated by putting `if connected?(socket), do: raise "boom"`, it's still possible to see the sites, paginate and even filter. 

Each filter text or page change will trigger reloading sites either way. Nerverthless, if we want to avoid selecting from stats on "dead" load, perhaps we could make charts render conditionally, and skip loading them in disconnected state but leave the rest working still? WYT? \cc @aerosol

lib/plausible/stats/clickhouse.ex
zoldar: Removed in https://github.com/plausible/analytics/pull/3463/commits/b8ca70596e927bc3504792504878f153182fb86d

lib/plausible/stats/clickhouse.ex
zoldar: I have made an attempt to address this in https://github.com/plausible/analytics/pull/3463/commits/d6a8071372c3fe6bac75ddc0be9a8b3ff7ace169. While it works, I'm a bit concerned about grouping in the base query? 🙈  It follows the partitioning key order but skips timestamp and appends `_sample_factor` because otherwise aggregate won't let it pass. But maybe that's ok (I'll run checks against prod DB in a bit anyway).

lib/plausible_web/live/sites.ex
zoldar: 👍 https://github.com/plausible/analytics/pull/3463/commits/05bcb796161837a3a33c9011e830cdb307e6a5e4

lib/plausible_web/live/sites.ex
aerosol: I think we could do the `connected?` check for graphs specifically (that's the only ClickHouse select AFAIU), but indeed having a nice fallback to LV unavailability is IMHO sensible on that page, given its importance. Worst case we'll have fully functional site list without graphs rendered (with a blank placeholder to avoid jumpy layout when the socket connects). How does that sound?

lib/plausible/stats/clickhouse.ex
aerosol: I think we should remove `ORDER BY` (added in https://github.com/plausible/analytics/pull/3463/commits/a6be5e302662b3a90f68da0e3fdfa6ff271c168e) - the difference is significant. For a list of 6 large(st) sites - with order by:

```
147 rows in set. Elapsed: 0.490 sec. Processed 20.37 million rows, 407.31 MB (41.58 million rows/s., 831.55 MB/s.)
```

Without:

```
147 rows in set. Elapsed: 0.164 sec. Processed 20.37 million rows, 407.31 MB (123.93 million rows/s., 2.48 GB/s.)
```

lib/plausible_web/live/sites.ex
aerosol: 3992cbaef

lib/plausible/stats/clickhouse.ex
aerosol: 2a2f3db63

aerosol:
cc @zoldar @ukutaht switch to offset-based pagination: https://github.com/plausible/analytics/pull/3463/commits/f44396ee61e65a6b0933e10ffd9553f023736253

lib/plausible/stats/clickhouse.ex
```diff
@@ -235,12 +237,23 @@ defmodule Plausible.Stats.Clickhouse do
       |> ClickhouseRepo.all()
       |> Map.new()
 
-    current_query =
+    current_base_query =
```
ukutaht, line 244: With this change, the query doesn't handle hourly uniqueness correctly anymore. Failing test to illustrate:

```elixir
test "another one" do
    fixed_now = ~N[2023-10-26 10:00:15]
    site = insert(:site)

    user_id = 111

    populate_stats(site, [
      build(:pageview, timestamp: ~N[2023-10-25 13:59:00]),
      build(:pageview, timestamp: ~N[2023-10-25 13:58:00]),
      build(:pageview, user_id: user_id, timestamp: ~N[2023-10-25 15:00:00]),
      build(:pageview, user_id: user_id, timestamp: ~N[2023-10-25 16:00:00])
    ])

    %{
      change: 100,
      visitors: 3,
      intervals: [
        %{interval: ~N[2023-10-25 11:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 12:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 13:00:00], visitors: 2},
        %{interval: ~N[2023-10-25 14:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 15:00:00], visitors: 1},
        %{interval: ~N[2023-10-25 16:00:00], visitors: 1},
        %{interval: ~N[2023-10-25 17:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 18:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 19:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 20:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 21:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 22:00:00], visitors: 0},
        %{interval: ~N[2023-10-25 23:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 00:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 01:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 02:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 03:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 04:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 05:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 06:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 07:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 08:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 09:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 10:00:00], visitors: 0},
        %{interval: ~N[2023-10-26 11:00:00], visitors: 0}
      ]
    } = Clickhouse.last_24h_visitors_hourly_intervals([site], fixed_now)[site.domain]
  end
```

aerosol: in your example, should the total `visitors` be still `3`?

aerosol: So `visitors` is the number of unique sessions across 24h, but `intervals[].visitors` should be the number of unique user_ids per interval?

aerosol: https://github.com/plausible/analytics/blob/feat/new-sites/test/plausible/stats/clickhouse_test.exs#L162-L174 is this test good?

aerosol: Not sure if that's possible with a single query, let me look at dashboard intervals implementation

ukutaht: >  So visitors is the number of unique sessions across 24h, but intervals[].visitors should be the number of unique user_ids per interval?

Exactly. So if a single visitor has pageviews on 15:59 and 16:00, the results should be:
15:00 - 1 visitor
16:00 - 1 visitor

but overall for 24h - 1 visitor

In the dashboard these are two separate queries - `aggregate` vs `timeseries`.

In order to make it a single query I think some trickery is needed, like joining with a subquery that calculates the totals or something.

aerosol: OK, need to run some experiments then. I think worst case, another option would be to use the regular Query/Stats interface and load each site's plot one by one, asynchronously. I was thinking of leveraging browser cache, so that each site card requests something like `/plots/{sitehash}.svg`, but @zoldar pointed out it we must be mindful of people wanting to see their traffic overview as it happens. Would be nice to make this view actually live update itself over time. 🤔

ukutaht: This is my attempt at the subquery+join trickery I mentioned:

```elixir
 total_q =
      from(
        e in "events_v2",
        hints: [sample: 20_000_000],
        where: e.site_id in ^Map.keys(site_id_to_domain_mapping),
        where: e.timestamp >= ^NaiveDateTime.add(now, -24, :hour),
        where: e.timestamp <= ^now,
        select: %{
          site_id: e.site_id,
          total_visitors: uniq(e.user_id)
        },
        group_by: e.site_id
      )

    new_current_q =
      from(
        e in "events_v2",
        hints: [sample: 20_000_000],
        join: total_q in subquery(total_q),
        on: e.site_id == total_q.site_id,
        where: e.site_id in ^Map.keys(site_id_to_domain_mapping),
        where: e.timestamp >= ^NaiveDateTime.add(now, -24, :hour),
        where: e.timestamp <= ^now,
        select: %{
          site_id: e.site_id,
          interval: fragment("toStartOfHour(timestamp)"),
          visitors: uniq(e.user_id),
          total: total_q.total_visitors
        },
        group_by: [e.site_id, fragment("toStartOfHour(timestamp)")],
        order_by: [e.site_id, fragment("toStartOfHour(timestamp)")]
      )
```

Logic seems sound to me - the `total` field would be duplicated for each interval but that wouldn't really be a problem.  Unfortunately this errors with

```
  ** (Ch.Error) Code: 47. DB::Exception: Unknown column: total_visitors, there are only columns site_id, toStartOfHour(timestamp), uniq(user_id), any(_sample_factor), multiply(uniq(user_id), any(_sample_factor)), round(multiply(uniq(user_id), any(_sample_factor))), toUInt64(round(multiply(uniq(user_id), any(_sample_factor)))). (UNKNOWN_IDENTIFIER) (version 23.5.3.1)
```

And I have no idea why.

ruslandoga: @ukutaht sorry to butt in. But I was checking email and saw this comment so... I think setting `allow_experimental_analyzer: 1` might enable better errors in such cases, `Plausible.ClickhouseRepo.all(query, settings: [allow_experimental_analyzer: 1])`, which would probably show that `total_q.total_visitors` needs to be aggregated.

aerosol: Thanks @ruslandoga @ukutaht!

It works, I'm going to run on it large sets to see how it performs.

aerosol: Ran 76d09b459 on a list of 6 fairly large sites:

```sql
SELECT e0."site_id",
toStartOfHour(timestamp),
toUInt64(round(uniq(e0."user_id") * any(_sample_factor))),
any(total_visitors)
FROM "events_v2" AS e0 sample 20000000
INNER JOIN
(SELECT se0."site_id" AS "site_id",
toUInt64(round(uniq(user_id) * any(_sample_factor))) AS "total_visitors"
FROM "events_v2" AS se0 sample 20000000
WHERE (se0."site_id" IN (...))
AND (se0."timestamp" >= '2023-10-29 10:00:00')
AND (se0."timestamp" <= '2023-10-30 10:00:00')
GROUP BY se0."site_id") AS s1 ON e0."site_id" = s1."site_id"
WHERE (e0."site_id" IN (...))
AND (e0."timestamp" >= '2023-10-29 10:00:00')
AND (e0."timestamp" <= '2023-10-30 10:00:00')
GROUP BY e0."site_id", toStartOfHour(timestamp)
ORDER BY e0."site_id", toStartOfHour(timestamp)
```

```
149 rows in set. Elapsed: 0.278 sec. Processed 40.73 million rows, 814.61 MB (146.53 million rows/s., 2.93 GB/s.)
```

The inner query alone:

```
6 rows in set. Elapsed: 0.159 sec. Processed 20.37 million rows, 407.31 MB (127.73 million rows/s., 2.55 GB/s.)
```

The outer query without inner:

```
149 rows in set. Elapsed: 0.113 sec. Processed 20.37 million rows, 407.31 MB (179.94 million rows/s., 3.60 GB/s.)
```

aerosol: I think we should keep track of `/sites` rendering time (including the 2nd mount) in grafana to understand any UX impact

aerosol: Enabled LiveView telemetry via 990060505:


```
$ curl -s localhost:8000/metrics | grep live
plausible_prom_ex_application_dependency_info{modules="59",name="phoenix_live_view",version="0.19.4"} 1
plausible_prom_ex_application_dependency_info{modules="4",name="phoenix_live_reload",version="1.4.1"} 1
# HELP plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds The time it takes for the live view to complete the handle_event callback.
# TYPE plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds histogram
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="10"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="100"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="500"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="1000"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="2500"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="5000"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="10000"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form",le="+Inf"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_sum{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form"} 0.292294
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_count{action="",event="validate",module="PlausibleWeb.Live.FunnelSettings.Form"} 3
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="10"} 0
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="100"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="500"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="1000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="2500"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="5000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="10000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form",le="+Inf"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_sum{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form"} 25.880778
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_count{action="",event="save",module="PlausibleWeb.Live.FunnelSettings.Form"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="10"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="100"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="500"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="1000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="2500"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="5000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="10000"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_bucket{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings",le="+Inf"} 1
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_sum{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings"} 0.034275
plausible_prom_ex_phoenix_live_view_handle_event_duration_milliseconds_count{action="",event="add-funnel",module="PlausibleWeb.Live.FunnelSettings"} 1
# HELP plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds The time it takes for the live view to complete the mount callback.
# TYPE plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds histogram
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="10"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="100"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="500"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="1000"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="2500"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="5000"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="10000"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="index",module="PlausibleWeb.Live.Sites",le="+Inf"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_sum{action="index",module="PlausibleWeb.Live.Sites"} 9.440883
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_count{action="index",module="PlausibleWeb.Live.Sites"} 5
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="10"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="100"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="500"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="1000"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="2500"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="5000"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="10000"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings.Form",le="+Inf"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_sum{action="",module="PlausibleWeb.Live.FunnelSettings.Form"} 9.357895
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_count{action="",module="PlausibleWeb.Live.FunnelSettings.Form"} 1
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="10"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="100"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="500"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="1000"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="2500"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="5000"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="10000"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_bucket{action="",module="PlausibleWeb.Live.FunnelSettings",le="+Inf"} 2
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_sum{action="",module="PlausibleWeb.Live.FunnelSettings"} 14.118652999999998
plausible_prom_ex_phoenix_live_view_mount_duration_milliseconds_count{action="",module="PlausibleWeb.Live.FunnelSettings"} 2
```

ukutaht: Thanks @ruslandoga ❤️

> I think we should keep track of /sites rendering time (including the 2nd mount) in grafana to understand any UX impact

Agreed.

The stats from the queries make sense - with the JOIN we're scanning double the data. I think there's a way to rewrite it using window functions so that it only has to scan the disk once. Something similar to a standard cumulative sum query but using `uniqState` + `uniqMerge`. The last interval would then contain the 'cumulative uniques' for the whole period. But I don't think we need to go that deep into this rabbit-hole for now :) This PR is ready to IMO

ukutaht (approved):

aerosol:
I fixed the viewbox cut-off:

Was able to repro with the following markup:

```html
<div class="pl-8 mt-2 flex items-center justify-between">
  <span class="text-gray-600 dark:text-gray-400 text-sm truncate">
    <svg viewBox="0 0 500 51" class="chart w-full mb-2">
  <defs>
    <clipPath id="gradient-cut-off-6e37808f-79b8-4e63-b59d-2c9ca9ff3712">
      <polyline points="0,51 0,35 20,11 40,10 60,6 80,0 100,1 120,2 140,7 160,8 180,11 200,11 220,10 240,16 260,23 280,24 300,31 320,32 340,36 360,36 380,37 400,35 420,28 440,15 460,8 480,17 500,50 500,51 0,51"></polyline>
    </clipPath>
  </defs>
  <rect x="0" y="1" width="520" height="50" fill="url(#chart-gradient-cut-off)" clip-path="url(#gradient-cut-off-6e37808f-79b8-4e63-b59d-2c9ca9ff3712)"></rect>
  <polyline fill="none" stroke="rgba(101,116,205)" stroke-width="2.6" points="0,35 20,11 40,10 60,6 80,0 100,1 120,2 140,7 160,8 180,11 200,11 220,10 240,16 260,23 280,24 300,31 320,32 340,36 360,36 380,37 400,35 420,28 440,15 460,8 480,17 500,50"></polyline>
</svg>
    <div class="flex justify-between items-center">
      <p>
        <span class="text-gray-800 dark:text-gray-200">
          <b>11.5k</b>
          visitor<span>s</span> in last 24h
        </span>
      </p>

      <p class="dark:text-gray-100">
  
  <span class="font-semibold text-red-400">↓</span>
  1%
</p>
    </div>
  </span>
</div>
```

image

After:

image

Via: 411f71259

Turns out the viewbox needs negative y-index.

PR #3464, opened 2023-10-26
Prepare business tier for release 

vinibrsl:
⚠️ Please do not merge!

This pull request changes limits for trials, keeping the old limits for users that registered before Oct 30. This date can be changed depending on when we release the business tier. I also changed the private preview notice to:

> Business plans are now live! If you wish to continue using <%= @feature_mod.display_name() %>, please upgrade your subscription to the Plausible Business plan.

Some tests were added/changed to comply with the user registration date. These tests can be removed after current trials end (in ~30 days).

Before merging, we need to set the `premium_features_private_preview` feature flag on. It can be switched off after 7 days as we previously defined. After switching off, the private preview ends, and users will have to subscribe to Business to use funnels, revenue goals, etc.

### DIFF
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -31,7 +31,7 @@ defmodule PlausibleWeb.Components.Billing do
 
         private_preview? && display_upgrade_link? && growth? ->
           ~H"""
-          Business plans are now live! If you wish to continue using <%= @feature_mod.display_name() %>, please
+          Business plans are now live! The private preview of <%= @feature_mod.display_name() %> for Plausible Growth plans ends <%= private_preview_days_remaining() %>. If you wish to continue using this feature, please
           <.link class="underline" href={Routes.billing_path(PlausibleWeb.Endpoint, :upgrade)}>
             upgrade your subscription
           </.link> to the Plausible Business plan.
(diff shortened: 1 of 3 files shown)


test/plausible/billing/quota_test.exs
```diff
@@ -359,8 +375,23 @@ defmodule Plausible.Billing.QuotaTest do
     end
 
     test "returns 5 when user in on trial" do
```
RobertJoonas, line 377:
```suggestion
    test "returns 3 when user in on trial" do
```

RobertJoonas (commented):
Added one minor typo fix suggestion inline but other than that looks good to me . Holding off on approving this though, to prevent this from being merged too early :)

lib/plausible_web/components/billing.ex
```diff
@@ -16,21 +16,27 @@ defmodule PlausibleWeb.Components.Billing do
...
-
-        private_preview? ->
-          "#{assigns.feature_mod.display_name()} is an upcoming premium functionality that's free-to-use during the private preview. Pricing will be announced soon."
-
         Plausible.Billing.on_trial?(assigns.billable_user) ->
           "#{assigns.feature_mod.display_name()} is part of the Plausible Business plan. You can access it during your trial, but you'll need to subscribe to the Business plan to retain access after the trial ends."
 
+        private_preview? && display_upgrade_link? && growth? ->
+          ~H"""
+          Business plans are now live! If you wish to continue using <%= @feature_mod.display_name() %>, please
```
RobertJoonas, line 34: I wonder if it would be useful for the user to see the countdown from 7 days? To me this message would give the idea that I don't have the premium feature access anymore. However they would still keep access. They just wouldn't know for how long... @metmarkosaric what do you think - should we display a message here like "_The preview of Revenue Goals for Plausible Growth plans will end in 7 days._" and the days would start counting down.

PR #3471, opened 2023-10-31
Warn user about Stats API when downgrading

vinibrsl:


test/plausible/billing/quota_test.exs
```diff
@@ -426,6 +426,13 @@ defmodule Plausible.Billing.QuotaTest do
       assert [RevenueGoals] == Quota.features_usage(user)
     end
 
+    test "returns [StatsAPI] when user uses revenue goals" do
```
RobertJoonas, line 429:
```suggestion
    test "returns [StatsAPI] when user has a stats api key" do
```

RobertJoonas (approved):
Just one typo fix change suggested but looks good otherwise 👍

PR #3472, opened 2023-11-01
Fix dark mode race condition

ukutaht:
### Changes

Ensure app.js is loaded before dashboard.js - this is assumed in the code for dark mode to work correctly in dashboard components.

PS - is there a nicer way to do this instead of an assign?

### DIFF
--- a/lib/plausible_web/templates/layout/app.html.eex
+++ b/lib/plausible_web/templates/layout/app.html.eex
@@ -33,5 +33,8 @@
       <%= render("_footer.html", assigns) %>
     <% end %>
     <script type="text/javascript" data-pref="<%= @conn.assigns[:theme] || (@conn.assigns[:current_user] && @conn.assigns[:current_user].theme) %>"  src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
+    <%= if @conn.assigns[:load_dashboard_js] do %>
+      <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/dashboard.js") %>"></script>
+    <% end %>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


aerosol (approved):
This works with the initial render but I think we should also handle system changes:


https://github.com/plausible/analytics/assets/173738/fa217b7e-bcc2-47d3-93c8-30337616e539

in a similar way it's done in apply-theme.

There, however, I'm getting warning about 

```javascript
window.matchMedia('(prefers-color-scheme: dark)').addListener(reapplyTheme);
```

being deprecated in favor of:

```javascript
window.matchMedia('(prefers-color-scheme: dark)').addEventListener("change", reapplyTheme);
```

lib/plausible_web/templates/layout/app.html.eex
```diff
@@ -33,5 +33,8 @@
       <%= render("_footer.html", assigns) %>
     <% end %>
     <script type="text/javascript" data-pref="<%= @conn.assigns[:theme] || (@conn.assigns[:current_user] && @conn.assigns[:current_user].theme) %>"  src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
+    <%= if @conn.assigns[:load_dashboard_js] do %>
```
aerosol, line 36: You can write just `@load_dashboard_js` I think?

ukutaht: That's going to 500 when `@load_dashboard_js` is not assigned at all :(

ukutaht:
@aerosol I fixed the deprecation warning.

Responding to system theme changing never worked in map.js I believe. Choosing to ignore this for now to get the fix live quickly. I think the right solution here is creating a react context for the theme (it's not only used in map.js) but it's too much work for this quick fix.

PR #3473, opened 2023-11-02
Fixing the broken GSC settings link in the dashboard

metmarkosaric:


aerosol (approved):
oh wow, thanks @metmarkosaric :sweat:

PR #3474, opened 2023-11-02
Extract button component

ukutaht:
### Changes

Extracts a common component for primary buttons to make them consistent across the app. Eventually we should be able to remove the `.button` css class.

I've tried using it in as many places as I could. Probably missed a few places as well :)


RobertJoonas (approved):
Looks good to me! 👍

PR #3490, opened 2023-11-06
Fix site card y-padding

aerosol:
image

Right: before
Left: after


aerosol:
Will be fixed elsewhere via https://github.com/plausible/analytics/pull/3469/commits/fe6b42393009931a544de553e7d7abb00528a2d9

PR #3504, opened 2023-11-08
Fixing Business Tier related bugs

RobertJoonas:
### Changes

Keeping this open here until there are either more changes or something more critical to deploy.

### DIFF
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -50,7 +50,7 @@ defmodule PlausibleWeb.Components.Billing do
       not has_access? ->
         ~H"""
         <.notice class="rounded-t-md rounded-b-none" size={@size} {@rest}>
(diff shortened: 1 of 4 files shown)


lib/plausible_web/components/billing.ex
```diff
@@ -80,6 +80,18 @@ defmodule PlausibleWeb.Components.Billing do
 
   # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def limit_exceeded_notice(assigns) do
+    ~H"""
+    <.notice {@rest}>
+      <%= account_label(@current_user, @billable_user) %> is limited to <%= @limit %> <%= @resource %>. To increase this limit,
```
zoldar, line 85: We are assuming there won't be limits to singular resource, correct?

zoldar (approved):

lib/plausible_web/components/billing.ex
vinibrsl: Correct :)

PR #3510, opened 2023-11-11
Notice across the app about Funnels and Revenue goals private preview end

RobertJoonas:
### Changes
 

* Change the upgrade CTA to say "contact us to upgrade your subscription" instead of "contact us about your enterprise plan" (which is not a good message for Business subscriptions who've hit some limit)
* Implement a `dismissable_id` option to `PlausibleWeb.Components.Generic.notice` which saves the user's preference in their localStorage that a notice with this `dismissable_id` has been dismissed.
* show a notice about losing access to Funnels and Revenue goals (to the site owners that are using either of these features) across the app (i.e. on `/settings`, `/sites`, dashboard, and site settings pages)
* Since we don't want to display the same notice twice, also remove the current private preview end notices from the relevant sections of site settings (i.e. the `PlausibleWeb.Components.Billing.premium_feature_notice` component)
* Fix a bug applying classnames to the `notice` component

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -318,7 +318,7 @@ defmodule PlausibleWeb.Components.Billing do
 
         _free_10k_or_enterprise_or_growth ->
           used_features = Plausible.Billing.Quota.features_usage(assigns.user)
-          for f_mod <- [Funnels, RevenueGoals], f_mod in used_features, do: f_mod
+          Enum.filter([Funnels, RevenueGoals], &(&1 in used_features))
       end
 
     assigns = assign(assigns, :features_to_lose, features_to_lose)
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -113,7 +113,7 @@ defmodule PlausibleWeb.Components.Billing do
         """
 
       true ->
-        ~H"please contact support@plausible.io about the Enterprise plan"
+        ~H"please contact hello@plausible.io to upgrade your subscription"
     end
   end
 
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -330,7 +330,7 @@ defmodule PlausibleWeb.Components.Billing do
     >
       <.notice
         class="rounded-t-md rounded-b-none"
-        dismissable_id="premium_features_private_preview_end"
+        dismissable_id={"premium_features_private_preview_end__#{@user.id}"}
       >
         Business plans are now live! The private preview of <%= PlausibleWeb.TextHelpers.pretty_join(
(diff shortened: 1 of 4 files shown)


lib/plausible_web/components/generic.ex
```diff
@@ -219,4 +238,21 @@ defmodule PlausibleWeb.Components.Generic do
       ["w-4 h-4"]
     end
   end
+
+  attr(:class, :string, default: "")
+
+  defp x_icon(assigns) do
```
ukutaht, line 244: How about `x-mark` from https://heroicons.com/?

RobertJoonas: @ukutaht yeah, that's the same icon 😅 I copied the SVG and pasted into this util function

zoldar: We already have `heroicons` dependency, so maybe we could use it instead? Like here https://github.com/plausible/analytics/blob/0c2e6b8751fc89bbb1dd40ec89d978df15ca7dbf/lib/plausible_web/live/sites.ex#L434

lib/plausible_web/components/billing.ex
```diff
@@ -317,6 +305,48 @@ defmodule PlausibleWeb.Components.Billing do
...
+      case Plans.get_subscription_plan(user.subscription) do
+        nil ->
+          []
+
+        %Plan{kind: :business} ->
+          []
+
+        _free_10k_or_enterprise_or_growth ->
+          used_features = Plausible.Billing.Quota.features_usage(assigns.user)
+          for f_mod <- [Funnels, RevenueGoals], f_mod in used_features, do: f_mod
```
zoldar, line 321: More of a matter of preference perhaps, but I personally got a little pause trying to sort out what's going on here 😅 Maybe this would communicate intent a bit clearer?

```suggestion
          Enum.filter([Funnels, RevenueGoals], &(&1 in used_features))
```

zoldar (approved):

lib/plausible_web/components/generic.ex
RobertJoonas: Ah, totally missed that, thanks for noting! 👍

RobertJoonas: https://github.com/plausible/analytics/pull/3510/commits/2d9f6fd6f1bf11af209b69858b362c537cf6851b

lib/plausible_web/components/billing.ex
RobertJoonas: Oh yeah I totally overdid this... Looks much better with `Enum.filter`, thanks! :) https://github.com/plausible/analytics/pull/3510/commits/20f4a2df97c5201ed114c8f7e61a595aec915276

PR #3513, opened 2023-11-13
Fix bug showing invitation popups

aerosol:
Apparently using the `<p>` element breaks x-show/styles cascading. Swapping it with `<div>` renders `x-show` correctly including children elements.

Before:

image

image

After:

image

image


zoldar (approved):

ukutaht:
🤷‍♂️

PR #3524, opened 2023-11-16
Limit grandfathering to *active* subscribers and trials

RobertJoonas:
### Grandfathering

* Stop offering grandfathered plans to subscribers who cancelled and let their subscription expire
  * These users will also see a warning (see the first screenshot below)
* Stop offering grandfathered plans to trials who didn't subscribe during, or within the next 10 days after the trial
  * Obviously only applies to those who started the trial **before** the business tiers release
  * The 10 days limit is silent - they won't be notified about having only 10 days to subscribe to get the grandfathered benefits

### Cancelled subscription notices

The current notice looks like this and it's only displayed on the account settings page:

image
 

**With this PR:**

* Move the notice across the app (same scope as the "subscription paused/past-due" notices -`_notice.html`)
* Note about grandfathering as mentioned above
* Change the styling by re-using the `Generic.notice` component (added a configurable color theme option for that)

image
image

### Bonus

Use `Generic.notice` in `subscription_past_due_notice` and `subscription_past_due_notice` for more consistent styling and less code.

Before:
image
image

After:
image
image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/plans.ex
+++ b/lib/plausible/billing/plans.ex
@@ -38,14 +38,14 @@ defmodule Plausible.Billing.Plans do
   still choose from old plans.
   """
   # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
-  def growth_plans_for(%User{} = user) do
+  def growth_plans_for(%User{} = user, now \\ Timex.now()) do
     user = Plausible.Users.with_subscription(user)
     v4_available = FunWithFlags.enabled?(:business_tier, for: user)
     owned_plan = get_regular_plan(user.subscription)
 
     cond do
       Application.get_env(:plausible, :environment) == "dev" -> @sandbox_plans
-      is_nil(owned_plan) && Timex.before?(user.inserted_at, @business_tier_launch) -> @plans_v3
+      is_nil(owned_plan) && grandfathered_trial?(user.trial_expiry_date, now) -> @plans_v3
       is_nil(owned_plan) && v4_available -> @plans_v4
       is_nil(owned_plan) -> @plans_v3
       owned_plan.kind == :business -> @plans_v4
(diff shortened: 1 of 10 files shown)


lib/plausible/billing/plans.ex
```diff
@@ -57,19 +58,36 @@ defmodule Plausible.Billing.Plans do
...
+
+  def grandfathered_trial?(trial_expiry, now) do
+    trial_start = Timex.shift(trial_expiry, days: -30)
+
+    joined_before_business_tiers = Timex.before?(trial_start, @business_tier_launch)
+    trial_active_or_expired_less_than_10d_ago = Timex.diff(now, trial_expiry, :days) <= 10
+
+    joined_before_business_tiers && trial_active_or_expired_less_than_10d_ago
+  end
+
```
vinibrsl, line 90: Is this function used outside this module? If not, I'd suggest leaving it private.

lib/plausible_web/components/generic.ex
```diff
@@ -58,28 +73,31 @@ defmodule PlausibleWeb.Components.Generic do
...
   attr(:rest, :global)
   slot(:inner_block)
 
   def notice(assigns) do
+    assigns = assign(assigns, :theme, Map.fetch!(@notice_themes, assigns.theme))
+
     ~H"""
     <div id={@dismissable_id} class={@dismissable_id && "hidden"}>
-      <div class={"rounded-md bg-yellow-50 dark:bg-yellow-100 p-4 relative #{@class}"} {@rest}>
+      <div class={"rounded-md #{@theme.bg} p-4 relative #{@class}"} {@rest}>
```
vinibrsl, line 89: Minor suggestion: by using an array, it avoids that extra space in the element class list when a custom class is not present, e.g. `rounded-md p-4 relative `

```suggestion
      <div class={["rounded-md p-4 relative", @theme.bg, @class]} {@rest}>
```

vinibrsl (approved):
Looks good :)

lib/plausible/billing/plans.ex
RobertJoonas: Good catch, thanks! https://github.com/plausible/analytics/pull/3524/commits/577bf195b39199e3423a8a5a5f6e4174fc8fd661

PR #3537, opened 2023-11-20
Include extra in docker image

aerosol:


zoldar (approved):
👍 though let's double-check it on staging

PR #3540, opened 2023-11-21
Changes to the emails as discussed

metmarkosaric:
for @RobertJoonas to review and edit. btw I've used the "Plausible.Billing.upgrade_route_for" link that we use in some other emails to make the call to action a bit more clear and to link directly to the upgrade page


metmarkosaric:
@RobertJoonas in the over_limit email, I think we should also hide these two sections from those above our standard limits as they are not relevant:
 
1. "You can easily adjust your subscription through our self-serve platform. The new charge will be prorated to reflect the amount you have already paid and the time until your current subscription is supposed to expire." 
2. "You can upgrade your subscription tier by clicking the 'Change plan' link."

and in the dashboard_locked email, these two sentences: 

1. "You can upgrade your subscription tier by clicking the 'Change plan' link."
2. "The new charge will be prorated to reflect the amount you have already paid and the time until your current subscription is supposed to expire."

ukutaht (approved):

PR #3541, opened 2023-11-21
Implement UI for 2FA setup and verification

zoldar:
### Changes

This PR implements controller level logic and UI views for 2FA.

The setting section is hidden behind a feature flag so that the visibility of the feature can be enabled any time after production deployment.

### Description of changes

When implemening UI, we have stuck to standard views as LV wouldn't improve user experience in any meaningful way.

The most important part is implementation of controller level logic for 2FA. Most of it is in `AuthController` and `PlausibleWeb.TwoFactor`. I have spiked a refactor extracting all things related to settings and user management from `AuthController` (creating another one, called `UserController`) but decided against doing it now as that would be too many things changing at once and 2FA implementation on its own is pretty big anyway.

The controller actions lean on `Auth.TOTP` API but they also layer logic related to session and cookie management on top. The moment the user with 2FA enabled logs in with their email and password, a dedicated session cookie is created for them (`session_2fa`), which records the fact that user has authenticated with first factor already but they haven't yet completed the second one. Only after the second factor is completed, the proper user session is setup on 2FA session is discarded. The lifetime of 2FA session is strictly limited to 5 minutes. If user does not verify their second factor within that timeframe, they will have to start over. This mechanism is meant to minimize risk of leaving half-authenticated session for extended period of time. The 2FA session is also invalidated the moment the user enters auth views outside second factor verification, including login. The session is stored in an encrypted cookie - a numerical user ID is stored there, so data is treated as sensitive.

Another 2FA aspect managed by auth controller is remembering trusted device. When a user checks "Trust this device for 30 days" when verifying the second factor (applies only to TOTP codes, not recovery ones), a separate cookie is created with expiry time 30 days in the future. This cookie is encrypted as well and contains TOTP token set at the time when the cookie was created. If the cookie is set and token inside it matches token stored for user, 2FA validation step is skipped after login even if 2FA is enabled. The cookie is cleared when 2FA is disabled or password is reset. It's cleared also when verifying 2FA without trusting (for instance, when running 2FA verification against another account in the same browser without setting trust checkbox).

### TODO before deployment

- [x] extract migration adding `totp_token` and run it before deploying the rest

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -267,6 +309,13 @@ defmodule PlausibleWeb.AuthController do
     end
   end
 
+  defp check_user_rate_limit(user) do
+    case Hammer.check_rate("login:user:#{user.id}", @login_interval, @login_limit) do
```
aerosol, line 313: cc @ruslandoga in context of #3571

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -298,17 +340,196 @@ defmodule PlausibleWeb.AuthController do
...
+        |> put_flash(:error, "Incorrect password provided")
+        |> redirect(to: Routes.auth_path(conn, :user_settings) <> "#setup-2fa")
+    end
+  end
+
+  def generate_2fa_recovery_codes(conn, %{"password" => password}) do
+    case Auth.TOTP.generate_recovery_codes(conn.assigns.current_user, password) do
+      {:ok, codes} ->
+        conn
+        |> put_flash(:success, "New recovery codes generated")
```
aerosol, line 411:
```suggestion
        |> put_flash(:success, "New Recovery Codes generated")
```

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -241,8 +268,22 @@ defmodule PlausibleWeb.AuthController do
...
+    redirect(conn, to: Routes.auth_path(conn, :login_form))
+  end
+
+  defp set_user_session_and_redirect(conn, user) do
+    login_dest = get_session(conn, :login_dest) || Routes.site_path(conn, :index)
+
+    conn
+    |> set_user_session(user)
+    |> put_session(:login_dest, nil)
+    |> redirect(to: login_dest)
```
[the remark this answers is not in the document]
zoldar: https://github.com/plausible/analytics/pull/3541/commits/88a25f63044b345b55e482125c07cc72eb00713c and https://github.com/plausible/analytics/pull/3541/commits/9690c509c79e048db16eb11a722c344be06bdec2

lib/plausible_web/components/generic.ex
```diff
@@ -248,6 +248,44 @@ defmodule PlausibleWeb.Components.Generic do
     end
   end
 
+  attr :text, :string, required: true
+  attr :scale, :integer, default: 4
+
+  def qr_code(assigns) do
```
ukutaht, line 254: Maybe move this to `Components.TwoFactor`? Unlikely we'll need QR codes outside of 2FA context.

lib/plausible_web/two_factor.ex
```diff
@@ -0,0 +1,80 @@
+defmodule PlausibleWeb.TwoFactor do
```
ukutaht, line 1: This module could use a more descriptive name like `TwoFactor.Session` or `TwoFactor.Web`.

lib/plausible_web/two_factor.ex
```diff
@@ -0,0 +1,80 @@
...
+  @remember_2fa_cookie "remember_2fa"
+  @remember_2fa_days 30
+  @remember_2fa_seconds @remember_2fa_days * 24 * 60 * 60
+
+  @session_2fa_cookie "session_2fa"
+  @session_2fa_seconds 5 * 60
+
+  @spec set_2fa_user(Plug.Conn.t(), Auth.User.t()) :: Plug.Conn.t()
+  def set_2fa_user(conn, %Auth.User{} = user) do
+    put_resp_cookie(conn, @session_2fa_cookie, %{current_2fa_user_id: user.id},
```
ukutaht, line 20: Hmm could we use the existing `Plug.Session` to store this data instead of a new dedicated cookie?

ukutaht (approved):

lib/plausible_web/two_factor.ex
zoldar: https://github.com/plausible/analytics/pull/3541/commits/e546ab1778118ff5405574068597f3db0a20f9ee

lib/plausible_web/components/generic.ex
zoldar: https://github.com/plausible/analytics/pull/3541/commits/be027e3d3b33b2ad9fd12ee3386715cac337868f

lib/plausible_web/two_factor/session.ex
```diff
@@ -0,0 +1,80 @@
...
+  @remember_2fa_cookie "remember_2fa"
+  @remember_2fa_days 30
+  @remember_2fa_seconds @remember_2fa_days * 24 * 60 * 60
+
+  @session_2fa_cookie "session_2fa"
+  @session_2fa_seconds 5 * 60
+
+  @spec set_2fa_user(Plug.Conn.t(), Auth.User.t()) :: Plug.Conn.t()
+  def set_2fa_user(conn, %Auth.User{} = user) do
+    put_resp_cookie(conn, @session_2fa_cookie, %{current_2fa_user_id: user.id},
```
ukutaht, line 20: @aerosol won't we have the same issue here as we did with the session cookie earlier this year?

Namely, leaving the `domain` blank will mean that this cookie is not available in LiveView.

ukutaht: We also started setting `secure: true` on the session cookie this summer. Probably makes sense to do it for these cookies as well

zoldar: > @aerosol won't we have the same issue here as we did with the session cookie earlier this year?

FWIW we do not use that cookie in LV. It's only read in the context of code verification views.

> Namely, leaving the domain blank will mean that this cookie is not available in LiveView.

> We also started setting secure: true on the session cookie this summer. Probably makes sense to do it for these cookies as well

I'll see if setting both, `domain` and `secure` (via `:secure_cookie` app env config flag) for the added cookies does not have any adverse effects.

zoldar: Okay, now `domain` and `secure` are set for both new cookies. I have tested the changes on staging and everything seems to be working as expected. I have actually caught one more inconsistency between attributes sent when creating and clearing the 2FA session cookie when doing that change, but it did not seem to affect anything in practice.

aerosol: I don't think the other cookie will ever be problematic for liveview - it's not the one that's being used for liveview session after all? Either way, this is a sensible change I think.

PR #3546, opened 2023-11-22
Small build updates

aerosol:


.credo.exs
```diff
@@ -0,0 +1,205 @@
...
+          {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
+          {Credo.Check.Readability.VariableNames, []},
+          {Credo.Check.Readability.WithSingleClause, []},
+
+          #
+          ## Refactoring Opportunities
+          #
+          {Credo.Check.Refactor.Apply, []},
+          {Credo.Check.Refactor.CondStatements, []},
+          {Credo.Check.Refactor.CyclomaticComplexity, []},
```
ukutaht, line 119: I think every time this is triggered we choose to ignore it. I propose we just disable the check globally instead.

assets/js/dashboard/stats/modals/conversions.js
```diff
@@ -1,15 +1,26 @@
...
 import numberFormatter from '../../util/number-formatter'
-import {parseQuery} from '../../query'
+import { parseQuery } from '../../query'
 import { escapeFilterValue } from '../../util/filters'
 
+/*global BUILD_EXTRA*/
+/*global require*/
+function maybeRequire() {
+  if (BUILD_EXTRA) {
+    return require('../../extra/money')
```
ukutaht, line 16: Can this function:
1. Take the path as an argument
2. Be extracted and reused instead of re-defining it in every file that uses it?

vinibrsl (approved):

lib/plausible_web/router.ex
```diff
@@ -125,16 +125,18 @@ defmodule PlausibleWeb.Router do
...
-    pipe_through [:public_api, PlausibleWeb.AuthorizeSitesApiPlug]
-
-    post "/", ExternalSitesController, :create_site
-    put "/shared-links", ExternalSitesController, :find_or_create_shared_link
-    put "/goals", ExternalSitesController, :find_or_create_goal
-    delete "/goals/:goal_id", ExternalSitesController, :delete_goal
-    get "/:site_id", ExternalSitesController, :get_site
-    put "/:site_id", ExternalSitesController, :update_site
-    delete "/:site_id", ExternalSitesController, :delete_site
+  on_full_build do
```
ukutaht, line 128: Can we move `PlausibleWeb.AuthorizeSitesApiPlug` and `ExternalSitesController` to the `extra/` folder as well?

assets/js/dashboard/stats/modals/conversions.js
aerosol: No. if the function takes an argument, the browser spits "dynamic require is not supported" or something along those lines

.credo.exs
aerosol: cool, will follow-up

lib/plausible_web/router.ex
aerosol: will do!

aerosol: See https://github.com/plausible/analytics/pull/3547 - btw `ExternalSitesController` was already moved

.credo.exs
aerosol: https://github.com/plausible/analytics/pull/3547

PR #3559, opened 2023-11-27
Remove custom domains support + update build options

aerosol:


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -363,14 +363,11 @@ defmodule PlausibleWeb.StatsController do
...
-    is_or_nil =
-      if Application.get_env(:plausible, :is_selfhost) do
-        if type = Plausible.Geo.database_type() do
-          String.starts_with?(type, "DBIP")
-        end
-      end
-
-    !!is_or_nil
+    on_full_build do
+      true
```
ukutaht, line 367: I think this is supposed to be `false` in full build

lib/plausible_web/live/register_form.ex
```diff
@@ -162,7 +162,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
         <% end %>
 
         <% submit_text =
-          if @is_selfhost or @invitation do
+          if not full_build?() or @invitation do
```
ukutaht, line 165: Is `small_build?` also available? Would be easier to read here than the negation of `full_build?`.

aerosol: https://github.com/plausible/analytics/pull/3559/commits/0b15a095c84ae811474e364f21043ba13ae71753

lib/plausible_web/controllers/stats_controller.ex
aerosol: https://github.com/plausible/analytics/pull/3559/commits/71ec0a429c3d25db72c9e5745810182959220954

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -363,14 +363,11 @@ defmodule PlausibleWeb.StatsController do
...
-        if type = Plausible.Geo.database_type() do
-          String.starts_with?(type, "DBIP")
-        end
-      end
-
-    !!is_or_nil
+    on_full_build do
+      false
+    else
+      String.starts_with?(Plausible.Geo.database_type(), "DBIP")
```
zoldar, line 369: Are we sure `Plausible.Geo.database_type()` is going to be always non-nil for self hosted instances? AFAIU it can return `nil` under some circumstances.

lib/plausible_web/plugs/first_launch_plug.ex
```diff
@@ -3,6 +3,17 @@ defmodule PlausibleWeb.FirstLaunchPlug do
   Redirects first-launch users to registration page.
   """
 
+  defmodule Test do
```
zoldar, line 6: Maybe we could put that module definition under `test/support`? Though on the other hand, maybe not as it's going to make code harder to follow 🤔

lib/plausible_web/views/layout_view.ex
```diff
@@ -83,9 +83,7 @@ defmodule PlausibleWeb.LayoutView do
       [key: "Custom Properties", value: "properties"],
       [key: "Integrations", value: "integrations"],
       [key: "Email Reports", value: "email-reports"],
-      if !is_selfhost() && conn.assigns[:site].custom_domain do
-        [key: "Custom domain", value: "custom-domain"]
-      end,
+      # XXX
```
zoldar, line 86: Is that a placeholder for something or can we get rid of it?

priv/repo/migrations/20231127132321_remove_custom_domains.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Plausible.Repo.Migrations.RemoveCustomDomains do
+  use Ecto.Migration
+
+  def change do
+    drop table(:custom_domains)
+  end
+end
```
zoldar, line 7: Just a note that we might want to run this migration _before_ deploying the rest of the code? 🤔

test/plausible/release_test.exs
```diff
@@ -1,35 +1,43 @@
...
-    assert Plausible.Repo.aggregate("plans", :count) == 54
+    stdout =
+      capture_io(fn ->
+        Release.dump_plans()
+      end)
+
+    assert stdout =~ "Loading plausible.."
+    assert stdout =~ "Starting dependencies.."
+    assert stdout =~ "Starting repos.."
+    assert stdout =~ "Inserted 54 plans"
```
zoldar, line 41: 👍

zoldar (approved):
✨ Left a couple questions and nits (most important one about migration, but maybe it's safe to deploy at once?).

priv/repo/migrations/20231127132321_remove_custom_domains.exs
aerosol: Why? This PR removes all the preloads/queries, so the table will be unused.

lib/plausible_web/plugs/first_launch_plug.ex
aerosol: Yeah, I found FirstLaunchPlug/MaybeDisableRegistration extremely hard to follow in general - this is why I prefer to have it here. Saves yet another round of greps.

priv/repo/migrations/20231127132321_remove_custom_domains.exs
zoldar: Yes it does but we deploy code first, run migration second. Or am I missing somnething?

aerosol: If we drop table before deploying the code, the existing code will query a non-existing table and will crash. 

If we deploy the code, the table becomes unused and the migration is safe to be executed at any time. Am I missing something?

extra/lib/plausible/stats/sampling.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Plausible.Stats.Sampling do
+  @moduledoc """
+  Sampling related functions
+  """
+  @default_sample_threshold 20_000_000
+
+  import Ecto.Query
+
+  @spec add_query_hint(Ecto.Query.t(), Plausible.Stats.Query.t()) :: Ecto.Query.t()
```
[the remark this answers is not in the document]
aerosol: I think multiple specs are allowed, but I'll change it

priv/repo/migrations/20231127132321_remove_custom_domains.exs
zoldar: Oh, right, I dunno why, it was reversed in my mind for some reason 🤡  You are absolutely right.

extra/lib/plausible/stats/sampling.ex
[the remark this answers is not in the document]
zoldar: Nono, you are right, they can be overloaded

https://hexdocs.pm/elixir/1.15/typespecs.html#defining-a-specification

lib/plausible_web/views/layout_view.ex
aerosol: https://github.com/plausible/analytics/pull/3559/commits/00b68779edf8ca9f1d573638aa563f79964fbf6d

lib/plausible_web/controllers/stats_controller.ex
aerosol: https://github.com/plausible/analytics/pull/3559/commits/fba7bb86c5a3e0f91b2ae998bf279db42946bc7b

PR #3560, opened 2023-11-27
Fix domains that start with UTF character

ukutaht:
### Changes

Currently the whole settings are is broken when the `site.domain` starts with a special character like ä.

The issue seems to be Phoenix itself being a bit too eager in a security check: https://github.com/phoenixframework/phoenix/issues/5508

Unfortunately it hasn't been fixed yet in Phoenix. The fix is to use `redirect(external: )` rather than `redirect(to: )` since the former will skip the problematic security check.

### Tests

### Changelog

### Documentation

### Dark mode


zoldar:
I'm wondering if it wouldn't be better to encode and decode the domain using IDNA encoding (aka punycode) instead 🤔 

```elixir
iex(5)> :idna.encode("zaęółć.pl")
~c"xn--za-6ja4f2d1l.pl"
iex(6)> :idna.encode("zaęółć.pl") |> :idna.decode |> to_string
"zaęółć.pl"
```

For reference:
- https://en.wikipedia.org/wiki/Punycode
- https://github.com/benoitc/erlang-idna

ukutaht:
> I'm wondering if it wouldn't be better to encode and decode the domain using IDNA encoding (aka punycode) instead 🤔

What I like about the current approach is that the domain name is displayed in the most 'native' format for the user.

Screenshot 2023-11-28 at 12 11 36


Also we need to encode `/` as part of the domain. I don't think punycode supports that?

zoldar (approved):

aerosol:
I went ahead trying to change the domain from ąęćż.pl to ćććąęćż.pl and upon submitting the domain change form got this:

image

Same with Reset Stats in the Danger Zone.

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -535,7 +535,7 @@ defmodule PlausibleWeb.SiteController do
       :success,
       "Removed #{recipient} as a recipient for the monthly report"
     )
-    |> redirect(to: "/#{URI.encode_www_form(site.domain)}/settings/email-reports")
+    |> redirect(external: "/#{URI.encode_www_form(site.domain)}/settings/email-reports")
```
aerosol, line 538: While at it we could wrap all in `Routes.*_path` helpers?

test/support/factory.ex
```diff
@@ -24,7 +24,8 @@ defmodule Plausible.Factory do
   end
 
   def site_factory(attrs) do
-    domain = sequence(:domain, &"example-#{&1}.com")
+    # The é exercises unicode support in domain names
+    domain = sequence(:domain, &"éxample-#{&1}.com")
```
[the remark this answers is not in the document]
ukutaht: Good point. Fixed in https://github.com/plausible/analytics/pull/3560/commits/c2b126098cfe582c2f5637941debbf7c6b93dbb9

lib/plausible_web/controllers/site_controller.ex
ukutaht: https://github.com/plausible/analytics/pull/3560/commits/506334b283d19f723c82999a492f595b9e580f8a

ukutaht:
@aerosol Thanks for the review. Fixes for the remaining errors:

https://github.com/plausible/analytics/pull/3560/commits/92b6a772331386d7152c10875c93580b9fdeda68
https://github.com/plausible/analytics/pull/3560/commits/cbfb2c7fbdf1cff7fd58bfe9e6aa1d9d32e690a1

PR #3564, opened 2023-11-28
Define a better monthly pageview usage

RobertJoonas:
### Changes

This PR defines a new structure for monthly pageview usage, which can be measured in either billing cycles, or for the last 30 days. Also in this PR:

- [x] Start preventing customers from subscribing to plans that do not accommodate their current pageview usage
- [x] Display monthly pageview usage information in account settings per billing cycles if the user has an active subscription
image

- [x] add the same monthly pageview usage information to CRM

### Tests

### Changelog

### Documentation

### Dark mode
- [x] The UI has been tested in both light/darkmode

### DIFF
--- a/lib/plausible/auth/user.ex
+++ b/lib/plausible/auth/user.ex
@@ -30,6 +30,9 @@ defmodule Plausible.Auth.User do
     field :theme, Ecto.Enum, values: [:system, :light, :dark]
     field :email_verified, :boolean
     field :previous_email, :string
+
+    # A field only used as a manual override - allow subscribing
+    # to any plan, even when exceeding its pageview limit
     field :allow_next_upgrade_override, :boolean
 
     # Fields for TOTP authentication. See `Plausible.Auth.TOTP`.


lib/plausible/billing/quota.ex
```diff
@@ -259,32 +336,45 @@ defmodule Plausible.Billing.Quota do
...
+
+  defp exceeds_monthly_pageview_limit?(%User{allow_next_upgrade_override: true}, _, _) do
+    false
+  end
+
+  defp exceeds_monthly_pageview_limit?(_user, plan, usage) do
+    case usage.monthly_pageviews do
+      %{last_30_days: %{total: total}} ->
+        limit = ceil(plan.monthly_pageview_limit * 1.15)
+        total > 13_000 && !within_limit?(total, limit)
```
vinibrsl, line 371: Maybe we could extract this `13_000` into a constant that explains what it means in this context?

lib/plausible/auth/user.ex
```diff
@@ -30,6 +30,7 @@ defmodule Plausible.Auth.User do
     field :theme, Ecto.Enum, values: [:system, :light, :dark]
     field :email_verified, :boolean
     field :previous_email, :string
+    field :allow_next_upgrade_override, :boolean
```
vinibrsl, line 33: What do you think of adding a comment here to document what this field does, or link to a function `@doc` that explains it?

vinibrsl (approved):

lib/plausible/billing/quota.ex
RobertJoonas: I agree that it's not really clear what the `13_000` stands for. I didn't extract a constant, but instead extracted a more descriptive function in https://github.com/plausible/analytics/pull/3564/commits/b62971c21232f4fc13a4086732d00858a04bd36c which will hopefully have the same effect of making it more readable :)

lib/plausible/auth/user.ex
RobertJoonas: Makes sense - https://github.com/plausible/analytics/pull/3564/commits/04caa6561ce76a70278be1f651e1e66b3461f0e0

lib/plausible/billing/quota.ex
vinibrsl: Perfect!

ukutaht (approved):

PR #3567, opened 2023-11-29
Update dockerfile for build options

aerosol:


Dockerfile
```diff
@@ -59,6 +61,8 @@ LABEL maintainer="plausible.io <hello@plausible.io>"
 ARG BUILD_METADATA={}
 ENV BUILD_METADATA=$BUILD_METADATA
 ENV LANG=C.UTF-8
+ARG MIX_ENV=prod
```
ukutaht, line 64: Why is MIX_ENV set to `small` in the build container and `prod` in the app container? Shouldn't it be the same in both containers?

aerosol:
Staging built fine

zoldar (approved):

PR #3568, opened 2023-11-29
Match choose plan styling with landing page

vinibrsl:
This pull request changes the choose plan styling with the landing page. Here's a preview:

Screenshot 2023-11-29 at 11 44 37

Related: https://github.com/plausible/website/pull/507

### DIFF
--- a/lib/plausible_web/live/choose_plan.ex
+++ b/lib/plausible_web/live/choose_plan.ex
@@ -100,7 +100,7 @@ defmodule PlausibleWeb.Live.ChoosePlan do
               else: "Upgrade your account" %>
           </p>
         </div>
-        <div class="mt-12 flex flex-col gap-4 lg:flex-row lg:gap-8 items-center lg:items-baseline">
+        <div class="mt-12 flex flex-col gap-8 lg:flex-row items-center lg:items-baseline">
           <.interval_picker selected_interval={@selected_interval} />
           <.slider_output volume={@selected_volume} available_volumes={@available_volumes} />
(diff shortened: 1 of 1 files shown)


lib/plausible_web/live/choose_plan.ex
```diff
@@ -217,33 +218,49 @@ defmodule PlausibleWeb.Live.ChoosePlan do
...
+            value={@selected_index}
+          />
+          <output
+            class="absolute bottom-[35px] py-[4px] px-[12px] -translate-x-1/2 rounded-md text-white bg-indigo-600 position text-xs font-medium"
+            style={@bubble_position}
+          >
+            <%= format_volume(@selected_volume, @available_volumes) %>
+          </output>
+        </div>
+        <span class="text-xs font-medium text-gray-600">10M+</span>
```
RobertJoonas, line 262: For v1 and v2 plans, volumes like 20M and 50M are also available:

image

Therefore we should not hardcode the 10m+ label at the end of the slider, but read it dynamically based on the available volumes instead

RobertJoonas (approved):
Nice! Looks much better than before 👍

Leaving a couple of styling suggestion here + one comment about a small bug inline.

In dark mode, I'd change the `10k` and `10m+` labels at the edges of the slider to a more brigther color.

image

Also, I think on mobile, it would look better with a bit more space between the slider label and the text here:
image

lib/plausible_web/live/choose_plan.ex
vinibrsl: Good catch!

PR #3577, opened 2023-11-30
add a new crm usage route for admins

RobertJoonas:
### Changes

Since the `calculate_usage` action didn't work out due to some internal Kaffy error that only surfaced on production (and I couldn't figure out the cause for that in a reasonable amount of time), I decided to implement a new admin route for displaying usage information and linking to it from the users index page in the CRM - looks cleaner and more readable like that anyway.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/auth/user_admin.ex
+++ b/lib/plausible/auth/user_admin.ex
@@ -30,6 +30,7 @@ defmodule Plausible.Auth.UserAdmin do
       trial_expiry_date: %{name: "Trial expiry", value: &format_date(&1.trial_expiry_date)},
       subscription_plan: %{value: &subscription_plan/1},
       subscription_status: %{value: &subscription_status/1},
+      usage: %{value: &usage_link/1},
       grace_period: %{value: &grace_period_status/1}
     ]
   end
@@ -43,10 +44,6 @@ defmodule Plausible.Auth.UserAdmin do
       lock: %{
         name: "Lock",
         action: fn _, user -> lock(user) end
-      },
-      calculate_usage: %{
-        name: "Calculate usage",
-        action: fn _, user -> calculate_usage(user) end
       }
     ]
   end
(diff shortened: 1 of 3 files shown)


lib/plausible_web/router.ex
```diff
@@ -72,6 +72,13 @@ defmodule PlausibleWeb.Router do
       pipe_through: [PlausibleWeb.Plugs.NoRobots, PlausibleWeb.CRMAuthPlug]
   end
 
+  on_full_build do
+    scope "/crm", PlausibleWeb do
+      pipe_through :flags
+      get "/auth/user/:user_id/usage", AdminController, :usage
```
vinibrsl, line 78: What do you think of adding a test ensuring this route is not available for non-admin accounts? We don't need to fully test it as it's an internal thing, but I'd recommend testing the authorization part.

lib/plausible/auth/user_admin.ex
```diff
@@ -70,42 +67,6 @@ defmodule Plausible.Auth.UserAdmin do
...
-    msg = """
-    TOTAL PAGEVIEWS (limit: #{pageview_limit})
-    #{@separator}
-    #{pageview_usage}
-    #{@separator}
-    SITES (#{site_usage} / #{site_limit}) #{@separator}
-    TEAM MEMBERS (#{team_member_usage} / #{team_member_limit})
-    """
-
-    {:error, user, msg}
```
zoldar, line 106: I know this is about to be replace but I'm wondering - was it supposed to always return error tuple like this?

RobertJoonas: @zoldar, yes 😅 Couldn't figure out how else to hack it with the Kaffy admin tool... I found the error message in a resource action the only way to return this information to the client.

lib/plausible_web/router.ex
RobertJoonas: Makes sense! ✅ https://github.com/plausible/analytics/pull/3577/commits/a116c8c45e40683979d8a7446d1bd63c0f5e3a77

zoldar (approved):

PR #3593, opened 2023-12-05
Plugins API exentsions (custom props, bulk goal delete, goal creation => ListResponse always)

aerosol:
### Changes

This PR introduces the following changes to Plugins API:

- Goal creation no longer results with polymorphic response; as this is part of OpenAPI v3 specs and client generators won't deal well with `oneOf` annotated-responses (ref https://github.com/OpenAPITools/openapi-generator/issues/10514) - instead, every time a `Goal.ListResponse` is returned, regardless of the input shape
- New endpoint `PUT /custom_props` allowing enabling custom props for a site (either single or in bulk)
  - available only for Business Tier clients, otherwise a PaymentRequired is returned
- New endpoint `DELETE /custom_props` allowing disabling custom prop(s) for a site (again, either a single or in bulk)
- New `DELETE /goals` endpoint - allowing goal deletion in bulk (a list of Goal IDs must be provided in JSON-encoded body)

Updated schema: https://gist.github.com/aerosol/8c37c50336ac2a4b96fead3d63d916d4
Diff: https://gist.github.com/aerosol/8c37c50336ac2a4b96fead3d63d916d4/revisions?diff=split&w=

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugins/api/controllers/goals.ex
+++ b/lib/plausible_web/plugins/api/controllers/goals.ex
@@ -23,13 +23,13 @@ defmodule PlausibleWeb.Plugins.API.Controllers.Goals do
       ) do
     site = conn.assigns.authorized_site
 
-    goals =
+    goal_or_goals =
       case body_params do
         %{goals: goals} -> goals
-        %{goal: _} = single_goal -> List.wrap(single_goal)
+        %{goal: _} = single_goal -> single_goal
       end
 
-    case API.Goals.create(site, goals) do
+    case API.Goals.create(site, goal_or_goals) do
       {:ok, goals} ->
         location_headers = Enum.map(goals, &{"location", goals_url(base_uri(), :get, &1.id)})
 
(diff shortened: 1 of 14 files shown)


lib/plausible_web/plugins/api/controllers/goals.ex
```diff
@@ -4,72 +4,30 @@ defmodule PlausibleWeb.Plugins.API.Controllers.Goals do
...
-        Errors.error(conn, 422, changeset)
-    end
-  end
-
-  def create(
-        %{private: %{open_api_spex: %{body_params: %{goals: goals}}}} = conn,
-        _params
-      ) do
-    site = conn.assigns.authorized_site
+    goals =
```
zoldar, line 26: Perhaps we could err on the side of consistency and bind to `goal_or_goals` instead, just like in case of custom props? `Api.Goals.create` does list wrapping implicitly anyway.

zoldar (approved):
✨

lib/plausible_web/plugins/api/controllers/goals.ex
aerosol: Nice 493b8c356

ukutaht (approved):

PR #3598, opened 2023-12-06
Improve subscription status checking code

vinibrsl:
This commit improves the subscription status checking in code, and creates convinience functions to work with it, including nil-checking and in?/2 function.


vinibrsl:
Sounds good, @zoldar! I applied your suggestion in bcc3927d50f79b2a0dc133ca33d6f138bf60c338 :)

PR #3600, opened 2023-12-07
Prevent invited users without sites from subscribing

RobertJoonas:
### Changes

This PR only contains one addition to the upgrade page - not allowing users with `trial_expiry_date=nil` to upgrade. Those are the users that were invited as an admin or a viewer to another site in Plausible, but don't own any sites themselves.

Preview:
image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/components/billing.ex
+++ b/lib/plausible_web/components/billing.ex
@@ -447,7 +447,7 @@ defmodule PlausibleWeb.Components.Billing do
 
   def subscription_paused_notice(assigns), do: ~H""
 
-  def upgrade_ineligible_notice(%{user: %User{trial_expiry_date: nil}} = assigns) do
+  def upgrade_ineligible_notice(assigns) do
     ~H"""
     <aside id="upgrade-eligible-notice" class="pb-6">
       <PlausibleWeb.Components.Generic.notice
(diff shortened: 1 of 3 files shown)


zoldar:
Are we sure `trial_expiry_date` being `nil` is a reliable indicator of user not owning any sites? I conducted a quick test and deleting the only site the user owns does not clear trial expiry. So if a user creates an account on PA, gets invited to another site and then subsequently removes their only owned site, they will still be falsely classified as legible for upgrade. It does not seem to be entirely implausible (sic!) scenario. Am I missing something? 🤔

RobertJoonas:
> Are we sure `trial_expiry_date` being `nil` is a reliable indicator of user not owning any sites? I conducted a quick test and deleting the only site the user owns does not clear trial expiry. So if a user creates an account on PA, gets invited to another site and then subsequently removes their only owned site, they will still be falsely classified as legible for upgrade. It does not seem to be entirely implausible (sic!) scenario. Am I missing something? 🤔

Good point, @zoldar! Adding the first site will start the trial, and if the user removes all of their sites, they'll be able to upgrade. The condition of not owning any sites is more wholesome and also includes the case where `trial_expiry_date = nil` in any case. I changed this in https://github.com/plausible/analytics/pull/3600/commits/0180139fcbd408e9cb2cf2a4b9ecc259aed54f35, thanks for the suggestion! :)

ukutaht (approved):

zoldar (approved):

PR #3613, opened 2023-12-11
dump buffer on invalid insert

ruslandoga:
This PR adds a temporary measure to check what's going on with invalid arrays in inserts.

### DIFF
--- a/lib/plausible/ingestion/write_buffer.ex
+++ b/lib/plausible/ingestion/write_buffer.ex
@@ -104,7 +104,7 @@ defmodule Plausible.Ingestion.WriteBuffer do
     _ ->
       path =
         Path.join(
-          Application.get_env(:plausible, :persistent_cache_dir, System.tmp_dir!()),
+          Application.get_env(:plausible, :persistent_cache_dir) || System.tmp_dir!(),
           "insert_error.dump"
         )
 
--- a/lib/plausible/ingestion/write_buffer.ex
+++ b/lib/plausible/ingestion/write_buffer.ex
@@ -100,6 +100,15 @@ defmodule Plausible.Ingestion.WriteBuffer do
         Logger.info("Flushing #{buffer_size} byte(s) RowBinary from #{name}")
         IngestRepo.query!(insert_sql, [header | buffer], insert_opts)
     end
+  rescue
+    e ->
+      path =
+        Path.join(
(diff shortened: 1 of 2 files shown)


lib/plausible/ingestion/write_buffer.ex
```diff
@@ -100,6 +100,15 @@ defmodule Plausible.Ingestion.WriteBuffer do
         Logger.info("Flushing #{buffer_size} byte(s) RowBinary from #{name}")
         IngestRepo.query!(insert_sql, [header | buffer], insert_opts)
     end
+  rescue
+    e ->
+      path =
+        Path.join(
+          System.tmp_dir!(),
```
aerosol, line 107: let's use PERSISTENT_CACHE_DIR here with always the same file name so subsequent failures overwrite rather than pile up? One should be enough

lib/plausible/ingestion/write_buffer.ex
```diff
@@ -100,6 +100,15 @@ defmodule Plausible.Ingestion.WriteBuffer do
...
+          "#{System.unique_integer([:positive])}-#{Exception.message(e)}.dump"
+        )
+
+      File.write!(path, state.buffer)
```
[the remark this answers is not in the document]
ruslandoga: I tried 

```elixir
Plausible.Event.WriteBuffer.insert(%Plausible.ClickhouseEventV2{"meta.key": [], "meta.value": ["a"]})
```

and it resulted in this file

```
998-Code: 190. DB::Exception: Elements 'meta.key' and 'meta.value' of Nested data structure 'meta' (Array columns) have different array sizes. (SIZES_OF_ARRAYS_DONT_MATCH) (version 23.3.7.5 (official build))?.dump
```

with the contents (passed through hexdump)

```
00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000030  00 00 00 00 01 01 61 01  00 00 00 01 00 00 00 00  |......a.........|
00000040
```

lib/plausible/ingestion/write_buffer.ex
ruslandoga: https://github.com/plausible/analytics/pull/3613/commits/2af86e0f0b6490140a7e927cb2a537edc10baae7

PR #3617, opened 2023-12-12
Protect against custom props payloads breaking event insertions

aerosol:
### Changes

This patch puts the shields up against custom props payloads ending up as uneven arrays insertions.

```
Elements 'meta.key' and 'meta.value' of Nested data structure 'meta' (Array columns) have different array sizes. (SIZES_OF_ARRAYS_DONT_MATCH)
```

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -214,15 +214,11 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_props(%__MODULE__{request: %{props: %{} = props}} = event) do
     # defensive: ensuring the keys/values are always in the same order
-    kvs =
-      Enum.reduce(props, %{keys: [], values: []}, fn
-        {k, v}, %{keys: ks, values: vs} ->
-          %{keys: [k | ks], values: [to_string(v) | vs]}
-      end)
+    {keys, values} = Enum.unzip(props)
 
     update_attrs(event, %{
-      "meta.key": Enum.reverse(kvs.keys),
-      "meta.value": Enum.reverse(kvs.values)
+      "meta.key": keys,
+      "meta.value": Enum.map(values, &to_string/1)
     })
   end
 
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -218,7 +218,7 @@ defmodule Plausible.Ingestion.Event do
 
     update_attrs(event, %{
       "meta.key": keys,
-      "meta.value": Enum.map(values, &to_string/1)
+      "meta.value": values
     })
   end
 
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -213,9 +213,16 @@ defmodule Plausible.Ingestion.Event do
   end
 
   defp put_props(%__MODULE__{request: %{props: %{} = props}} = event) do
+    # defensive: ensuring the keys/values are always in the same order
+    kvs =
+      Enum.reduce(props, %{keys: [], values: []}, fn
+        {k, v}, %{keys: ks, values: vs} ->
+          %{keys: [k | ks], values: [to_string(v) | vs]}
+      end)
+
     update_attrs(event, %{
-      "meta.key": Map.keys(props),
-      "meta.value": Enum.map(props, fn {_, v} -> to_string(v) end)
+      "meta.key": Enum.reverse(kvs.keys),
+      "meta.value": Enum.reverse(kvs.values)
     })
   end
 
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -206,20 +206,21 @@ defmodule Plausible.Ingestion.Request do
     props =
       (request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"])
       |> Plausible.Helpers.JSON.decode_or_fallback()
-      |> Enum.filter(fn {k, v} ->
-        non_empty_key = is_binary(k) and byte_size(k) > 0
-
-        valid_value =
-          (is_binary(v) and byte_size(v) > 0) or is_number(v) or (is_atom(v) and not is_nil(v))
-
-        non_empty_key and valid_value
-      end)
+      |> Enum.reduce([], &filter_bad_props/2)
       |> Enum.take(@max_props)
       |> Map.new()
 
     changeset
     |> Changeset.put_change(:props, props)
     |> validate_props()
+  end
+
+  defp filter_bad_props({k, v}, acc) do
+    cond do
+      Enum.any?([k, v], &(is_list(&1) or is_map(&1))) -> acc
+      Enum.any?([k, v], &(String.trim_leading(to_string(&1)) == "")) -> acc
+      true -> [{to_string(k), to_string(v)} | acc]
+    end
   end
 
   @max_prop_key_length Plausible.Props.max_prop_key_length()
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -206,7 +206,14 @@ defmodule Plausible.Ingestion.Request do
     props =
       (request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"])
       |> Plausible.Helpers.JSON.decode_or_fallback()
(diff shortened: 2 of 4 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -213,9 +213,16 @@ defmodule Plausible.Ingestion.Event do
...
+      Enum.reduce(props, %{keys: [], values: []}, fn
+        {k, v}, %{keys: ks, values: vs} ->
+          %{keys: [k | ks], values: [to_string(v) | vs]}
+      end)
+
     update_attrs(event, %{
-      "meta.key": Map.keys(props),
-      "meta.value": Enum.map(props, fn {_, v} -> to_string(v) end)
+      "meta.key": Enum.reverse(kvs.keys),
+      "meta.value": Enum.reverse(kvs.values)
```
ruslandoga, line 225:
```elixir
{keys, values} = Enum.unzip(props)
{keys, Enum.map(values, &Kernel.to_string/1)
```
?

aerosol: https://github.com/plausible/analytics/pull/3617/commits/a4fca2fadcbf470df3c31f26e087f800d534a51d

ukutaht (approved):

lib/plausible/ingestion/request.ex
```diff
@@ -206,7 +206,14 @@ defmodule Plausible.Ingestion.Request do
...
       (request_body["m"] || request_body["meta"] || request_body["p"] || request_body["props"])
       |> Plausible.Helpers.JSON.decode_or_fallback()
-      |> Enum.reject(fn {_k, v} -> is_nil(v) || is_list(v) || is_map(v) || v == "" end)
+      |> Enum.filter(fn {k, v} ->
+        non_empty_key = is_binary(k) and byte_size(k) > 0
+
+        valid_value =
+          (is_binary(v) and byte_size(v) > 0) or is_number(v) or (is_atom(v) and not is_nil(v))
+
+        non_empty_key and valid_value
```
ruslandoga, line 215: Was it Ecto.Changeset.cast that was removing the values from arrays?

aerosol: No, it was the `reject` that wasn't exhaustive - ignoring the keys, `nil: false` would end up as `%{"" => false}`

ruslandoga: This would mean that if `empty_keys` option is changed in `ClickhouseEventV2.new`, this filter would need to be changed as well. For example, by default Ecto removed whitespaced values too, `"  "`

ruslandoga: Here's what happens to the failing test in master:

```elixir
# attrs before cast
%{
  name: "Signup",
  timestamp: ~N[2023-12-12 10:09:20],
  domain: "é-0.example.com",
  hostname: "gigride.live",
  referrer: nil,
  pathname: "/",
  site_id: 136845,
  user_id: 14937685462052691295,
  "meta.key": ["", "good"],
  referrer_source: nil,
  utm_campaign: nil,
  utm_content: nil,
  utm_medium: nil,
  utm_source: nil,
  utm_term: nil,
  "meta.value": ["false", "true"]
}
# after cast
#Ecto.Changeset<
  action: nil,
  changes: %{
    name: "Signup",
    timestamp: ~N[2023-12-12 10:09:20],
    hostname: "gigride.live",
    pathname: "/",
    site_id: 136845,
    user_id: 14937685462052691295,
    "meta.key": ["good"],
    "meta.value": ["false", "true"]
  },
  errors: [],
  data: #Plausible.ClickhouseEventV2<>,
  valid?: true
>
# after ClickhouseEventV2.new
%Plausible.ClickhouseEventV2{
  __meta__: #Ecto.Schema.Metadata<:built, "events_v2">,
  name: "Signup",
  site_id: 136845,
  hostname: "gigride.live",
  pathname: "/",
  user_id: 14937685462052691295,
  session_id: 13558915368958867678,
  timestamp: ~N[2023-12-12 10:09:20],
  referrer: nil,
  referrer_source: nil,
  utm_medium: nil,
  utm_source: nil,
  utm_campaign: nil,
  utm_content: nil,
  utm_term: nil,
  country_code: nil,
  subdivision1_code: nil,
  subdivision2_code: nil,
  city_geoname_id: nil,
  screen_size: nil,
  operating_system: nil,
  operating_system_version: nil,
  browser: nil,
  browser_version: nil,
  "meta.key": ["good"],
  "meta.value": ["false", "true"],
  revenue_source_amount: nil,
  revenue_source_currency: nil,
  revenue_reporting_amount: nil,
  revenue_reporting_currency: nil,
  transferred_from: nil
}
```

aerosol: Can you elaborate? What `empty_keys`? Not sure what are you suggesting.

ruslandoga: Note that attrs has valid values in `meta` columns, it might be valid app-wise, but not something ClickHouse would complain about. So I think maybe `ClickHouseEventV2new` needs to be changed as well.

ruslandoga: https://github.com/plausible/analytics/blob/26809bfd2ef7da34bcaf85605a12482520f571dd/lib/plausible/clickhouse_event_v2.ex#L82

ruslandoga: I'm suggesting making a "single source of truth" for Ecto casting and for Request filtering.

aerosol: And that source of truth would be what exactly? I'm sorry I don't follow at all. Could you suggest a patch and tests that you have in mind? Note we have some sepration there already in `Ingestion.RequestTest` and `Ingestion.EventTest` if that helps.

ruslandoga: I think to future-proof, adding a whitespace key to props test would be enough.

```elixir
params = %{
        name: "Signup",
        url: "http://gigride.live/",
        domain: site.domain,
        props: %{
          false: nil,
          "    " => "also filtered",
          nil: false,
          good: true
        }
      }
```

aerosol: Maybe I'm dense but I don't see the point of keeping `ClickHouseEventV2.empty_values()`at all, looks like some remnant from https://github.com/plausible/analytics/commit/5c759dbfed576d58dac9b1a55fee30ada2211318#diff-a120b2feb579aca881694510480f6ff31c1f4cce9c3ae1eeae4fd9b1c139f4cfR57 - the `nil` there is redundant.

What do you think about 445a34a75?

ruslandoga: ❤️

ruslandoga (approved):

lib/plausible/ingestion/request.ex
aerosol: Reasoning: this is custom props specific, no point in generalizing it, we definitely don't care about empty keys nor values in this case, existing tests already show it - I don't know about other fields. Relying on https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L906 alone is clearly not enough in this case since we need to guarantee parity. Thus, `Request` seems like a good place to do all the pre-processing. Happy to reiterate if you think it can be addressed differently, but since both `meta.key` and `meta.value` are separate schema fields, I don't see how to apply Ecto casting cleanly here.

ruslandoga: >  Thus, Request seems like a good place to do all the pre-processing. 

I was just worried that the casting and Request pre-processing would be different. Note that omitting `empty_values` in `cast` would use the default `Ecto.Changeset.empty_values`

aerosol: @ruslandoga 

>  But note that omitting empty_values in cast doesn't make Ecto keep all values, instead it uses Ecto.Changeset.empty_values

which is why I mentioned https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L906 which in fact is called by `Ecto.Changeset.empty_values`. 

@ukutaht do you remember the intention behind https://github.com/plausible/analytics/commit/5c759dbfed576d58dac9b1a55fee30ada2211318#diff-a120b2feb579aca881694510480f6ff31c1f4cce9c3ae1eeae4fd9b1c139f4cfR57 ? It essentially allows empty strings such as `"      "` - can you think of a field where this is necessary to support?

PR #3622, opened 2023-12-13
add session collapse order test

ruslandoga:
### Changes

This PR adds tests to ensure the correct session collapse order. It should prevent issues similar to https://github.com/plausible/analytics/pull/3511#discussion_r1423382636

### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #3624, opened 2023-12-13
Randomize user_id in session collapse tests

aerosol:
### Changes

This is a blind stab at https://github.com/plausible/analytics/actions/runs/7192271123/job/19588394230 introduced via #3622 - to prevent `Cachex` lookups clash.
### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/session/cache_store_test.exs
+++ b/test/plausible/session/cache_store_test.exs
@@ -1,5 +1,5 @@
 defmodule Plausible.Session.CacheStoreTest do
-  use Plausible.DataCase
+  use Plausible.DataCase, asyc: true
   alias Plausible.Session.CacheStore
 
   defmodule FakeBuffer do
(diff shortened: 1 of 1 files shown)


test/plausible/session/cache_store_test.exs
```diff
@@ -123,8 +123,16 @@ defmodule Plausible.Session.CacheStoreTest do
...
+    defp random_user_id() do
+      SipHash.hash!(:crypto.strong_rand_bytes(16), Ecto.UUID.generate())
+    end
+
     test "across parts" do
-      e = build(:event, name: "pageview")
+      e =
+        build(:event,
+          name: "pageview",
+          user_id: random_user_id()
```
aerosol, line 134: eh, this is factory-provided

aerosol:
invalid, events have user_id based on UUID already

aerosol:
@ruslandoga I'm thinking that maybe `site_id: Enum.random(1000..10_000),` in `event_factory` is not enough for this to be reliable, but on the other hand that kind of error should be reproducible with `--seed` 🤔

ruslandoga:
Ah, I assumed `site_id`s point to real site ids in PostgreSQL. I'll update the tests to create real sites and use their ids. `Enum.random(1000..10_000)` might be replaced with `System.unique_integer([:positive])`

aerosol:
@ruslandoga I have a repro, hold on, inserting site IDs doesn't guarantee in-cache uniqueness

aerosol:
https://github.com/plausible/analytics/pull/3626/files

PR #3626, opened 2023-12-13
Fix flaky test

aerosol:
### Changes

The test flakiness introduced via #3622 was reproducible locally by running the `across parts` tests hundred times with a macro generator. This patch no longer leads to failures.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/session/cache_store_test.exs
+++ b/test/plausible/session/cache_store_test.exs
@@ -118,13 +118,20 @@ defmodule Plausible.Session.CacheStoreTest do
   end
 
   describe "collapse order" do
+    defp new_site_id() do
+      [[site_id]] =
+        Plausible.ClickhouseRepo.query!("select max(site_id) + rand() from sessions_v2 FINAL").rows
+
+      site_id
+    end
+
     defp flush(events) do
       for e <- events, do: CacheStore.on_event(e, nil)
       Plausible.Session.WriteBuffer.flush()
     end
 
     test "across parts" do
-      e = build(:event, name: "pageview")
+      e = build(:event, name: "pageview", site_id: new_site_id())
 
       flush([%{e | pathname: "/"}])
       flush([%{e | pathname: "/exit"}])
@@ -138,7 +145,7 @@ defmodule Plausible.Session.CacheStoreTest do
     end
 
     test "within parts" do
-      e = build(:event, name: "pageview")
+      e = build(:event, name: "pageview", site_id: new_site_id())
 
       flush([
         %{e | pathname: "/"},
@@ -154,7 +161,7 @@ defmodule Plausible.Session.CacheStoreTest do
     end
 
     test "across and within parts" do
-      e = build(:event, name: "pageview")
+      e = build(:event, name: "pageview", site_id: new_site_id())
 
       flush([
         %{e | pathname: "/"},


test/plausible/session/cache_store_test.exs
```diff
@@ -118,13 +118,20 @@ defmodule Plausible.Session.CacheStoreTest do
   end
 
   describe "collapse order" do
+    defp new_site_id() do
+      [[site_id]] =
+        Plausible.ClickhouseRepo.query!("select max(site_id) + rand() from sessions_v2 FINAL").rows
```
aerosol, line 123: I guess `max(site_id) + 1` works just as good. Not sure.

ruslandoga: Can it be part of the fixtures? Is the query expensive?

ruslandoga: Otherwise we can use the same approach as all the other tests use, a custom large int for site_id.

aerosol: > Can it be part of the fixtures? Is the query expensive?

I'll check :+1: 

> Otherwise we can use the same approach as all the other tests use, a custom large int for site_id.

What do you mean? Which approach? Random ints we currently use are clearly prone to clashing.

ruslandoga: `build(:event, name: "pageview", site_id = 123784652187346)`

aerosol: > Can it be part of the fixtures? Is the query expensive?

```
Finished in 41.2 seconds (13.7s async, 27.4s sync)
9 doctests, 1937 tests, 0 failures, 29 excluded
```

vs 

```
Finished in 116.0 seconds (69.3s async, 46.7s sync)
9 doctests, 1937 tests, 2 failures, 29 excluded
```

Clearly not worth it, this workaround is again, specific to that test.

ruslandoga: I guess it's fine if it works, it's juts a test after all.

Another idea I had was to use `System.unique_integer([:positive])` but since it can overlap with `sites.id` from PostgreSQL, I'd read the current sequence value for `sites` and add a million (a number large enough to cover all sites inserted into PostgreSQL during the test run) and then add `System.unique_integer([:positive])` to ensure uniqueness in ClickHouse-only tests.

ruslandoga:
```elixir
# test_helpers.exs
[[min_sites_id]] = Plausible.Repo.query!("select nextvar('sites_id_seq')").rows
Application.put_env(:plausible, :tests, min_sites_id: min_sites_id)
# fixtures
def event do
  # ...
  site_id: Keyword.fetch!(Application.fetch_env!(:plausible, :tests), :min_sites_id) + 1_000_000 + System.unique_integer([:positive])
  # ...
end
```

ruslandoga: I don't like it either though :)

ruslandoga (approved):

test/plausible/session/cache_store_test.exs
aerosol: application env is another shared state we're suffering from, I'm gonna just go ahead and merge it

PR #3631, opened 2023-12-13
Fix `class` in `styled_link` and remove existing uses of `class`

RobertJoonas:
### Changes

Make it possible to add custom classes to the `styled_link` component. The class attribute was given in several cases, but it was not used by the component.


zoldar (approved):
✨

PR #3632, opened 2023-12-13
Remove business tier feature flag

vinibrsl:
This commit removes all code branches related to the business tier feature flag, as we're not flipping this flag off anymore. It also removes unused routes, e.g. /billing/change-plan and /billing/upgrade


RobertJoonas (approved):
Looks good to me! 👍 We also don't need several billing templates anymore. I'll push a commit (+ 2 other small commits) in a sec, so you can review those and merge if everything looks good :)

PR #3636, opened 2023-12-14
Remove onClick handler from details link in ListReport

ukutaht:
### Changes

Fixes issue where the UI would unexpectedly tab-switch when the user clicks on details.

### Tests

### Changelog
- [x] No changelog entry needed

### Documentation

### Dark mode


zoldar:
Is it expected that properties tab is opened when clicking goal directly from the list but not when clicked from the list in the modal?

https://github.com/plausible/analytics/assets/588351/97f710a8-e1e5-4b1c-a05c-0f8f58cfea3c

metmarkosaric:
@zoldar not intentionally. It's just not something we've had a chance to build yet. We should also ideally open the most popular property by default so people don't need to manually select it after clicking on a goal

zoldar (approved):
:sparkles:

PR #3650, opened 2023-12-22
Add a note about site settings being available on the locked screen

metmarkosaric:
some people are not aware that they can access site settings when the dashboard is locked. i've tried to add that to the note here as it should improve the user experience and minimize the number of people reaching out to support


zoldar (approved):

zoldar:
I have adjusted the styling and converted the view to use our predefined components for links:

image

image

image

image

image

image

RobertJoonas (approved):
Looks nice!

PR #3656, opened 2023-12-28
Add `custom_props.csv` back to CSV export for Growth plans

RobertJoonas:
### Changes

This PR fixes a bug where `custom_props.csv` file was not included in the CSV export for Growth plans using the `url` or `path` props (e.g. with outbound link or file download goals). These props should not be considered a premium feature.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -1214,21 +1214,32 @@ defmodule PlausibleWeb.Api.StatsController do
 
     prop_names = Plausible.Stats.CustomProps.fetch_prop_names(site, query)
 
-    values =
-      prop_names
-      |> Enum.map(fn prop_key ->
-        breakdown_custom_prop_values(site, Map.put(params, "prop_key", prop_key))
-        |> Enum.map(&Map.put(&1, :property, prop_key))
-        |> transform_keys(%{:name => :value})
-      end)
-      |> Enum.concat()
+    prop_names =
+      if Plausible.Billing.Feature.Props.enabled?(site) do
+        prop_names
+      else
+        prop_names |> Enum.filter(& &1 in Plausible.Props.internal_keys())
+      end
 
-    percent_or_cr =
-      if query.filters["event:goal"],
-        do: :conversion_rate,
-        else: :percentage
+    if prop_names == [] do
+      nil
+    else
+      values =
+        prop_names
+        |> Enum.map(fn prop_key ->
+          breakdown_custom_prop_values(site, Map.put(params, "prop_key", prop_key))
+          |> Enum.map(&Map.put(&1, :property, prop_key))
+          |> transform_keys(%{:name => :value})
+        end)
+        |> Enum.concat()
 
-    to_csv(values, [:property, :value, :visitors, :events, percent_or_cr])
+      percent_or_cr =
+        if query.filters["event:goal"],
+          do: :conversion_rate,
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1214,21 +1214,32 @@ defmodule PlausibleWeb.Api.StatsController do
...
+        prop_names |> Enum.filter(&(&1 in Plausible.Props.internal_keys()))
+      end
 
-    percent_or_cr =
-      if query.filters["event:goal"],
-        do: :conversion_rate,
-        else: :percentage
+    if prop_names == [] do
+      nil
+    else
```
aerosol, line 1226:
```suggestion
   if not Enum.empty?(prop_names) do
```

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1214,21 +1214,32 @@ defmodule PlausibleWeb.Api.StatsController do
...
-        |> Enum.map(&Map.put(&1, :property, prop_key))
-        |> transform_keys(%{:name => :value})
-      end)
-      |> Enum.concat()
+    prop_names =
+      if Plausible.Billing.Feature.Props.enabled?(site) do
+        prop_names
+      else
+        prop_names |> Enum.filter(&(&1 in Plausible.Props.internal_keys()))
+      end
```
aerosol, line 1222: so `url` and `path` are available regardless of the plan?

aerosol (approved):

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1214,21 +1214,32 @@ defmodule PlausibleWeb.Api.StatsController do
...
-        breakdown_custom_prop_values(site, Map.put(params, "prop_key", prop_key))
-        |> Enum.map(&Map.put(&1, :property, prop_key))
-        |> transform_keys(%{:name => :value})
-      end)
-      |> Enum.concat()
+    prop_names =
+      if Plausible.Billing.Feature.Props.enabled?(site) do
+        prop_names
+      else
+        prop_names |> Enum.filter(&(&1 in Plausible.Props.internal_keys()))
```
zoldar, line 1221: A bit of an aside, but perhaps `Plausible.Props.internal_keys()` could be renamed to `Plausible.Props.internal_props()`?

zoldar (approved):
✨

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: I think in the props module, the scope and the fact that we're speaking of props is already clear. I usually avoid writing module names into the functions themselves as well - in this case, we'd then write "props" two times.

Hence, I'd stick to either `Props.internal()` or leave it as is - the `keys` word is currently there to make it clear that we're talking of `prop_key` i.e. the `meta.key` field in the database.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Yes, exactly.

lib/plausible_web/controllers/api/stats_controller.ex
zoldar: Makes sense, we can leave it as is then 👍

PR #3666, opened 2024-01-04
Always enable e-mail verification on full build

aerosol:
### Changes

There is no point in ever disabling it on full build.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/auth/user.ex
+++ b/lib/plausible/auth/user.ex
@@ -67,7 +67,7 @@ defmodule Plausible.Auth.User do
     |> validate_password_strength()
     |> hash_password()
     |> start_trial()
-    |> change_email_verification_status()
+    |> set_email_verification_status()
     |> unique_constraint(:email)
   end
 
@@ -85,7 +85,7 @@ defmodule Plausible.Auth.User do
     |> validate_email_changed()
     |> check_password()
     |> unique_constraint(:email)
-    |> change_email_verification_status()
+    |> set_email_verification_status()
     |> put_change(:previous_email, user.email)
   end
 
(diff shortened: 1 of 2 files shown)


lib/plausible/auth/user.ex
```diff
@@ -259,11 +259,13 @@ defmodule Plausible.Auth.User do
...
-  defp set_email_verified(user) do
-    if Keyword.fetch!(Application.get_env(:plausible, :selfhost), :enable_email_verification) do
+  defp change_email_verification_status(user) do
+    on_full_build do
       change(user, email_verified: false)
     else
-      change(user, email_verified: true)
+      selfhosted_config = Application.get_env(:plausible, :selfhost)
+      enable? = Keyword.fetch!(selfhosted_config, :enable_email_verification)
+      change(user, email_verified: !enable?)
```
zoldar, line 268: `ENABLE_EMAIL_VERIFICATION` env var is casted to an (existing) atom which is expected to be either `true` or `false`, so I think we can rephrase that as:

```suggestion
      must_verify? = Keyword.fetch!(selfhosted_config, :enable_email_verification)
      change(user, email_verified: not must_verify?)
```

WDYT?

lib/plausible/auth/user.ex
```diff
@@ -66,8 +66,8 @@ defmodule Plausible.Auth.User do
     |> validate_confirmation(:password, required: true)
     |> validate_password_strength()
     |> hash_password()
-    |> start_trial
-    |> set_email_verified
+    |> start_trial()
+    |> change_email_verification_status()
```
zoldar, line 70: As this is a provisioning action, perhaps `set_email_verification_status` sounds better?

zoldar (approved):

lib/plausible/auth/user.ex
aerosol: https://github.com/plausible/analytics/pull/3666/commits/bcd3e0b6f07d5566b015dae5724003a2f486886b

PR #3661, opened 2024-01-01
Don't leak internal server errors, use support hash

aerosol:
### Changes

This PR:

- makes internal server error details inaccessible to end users
- instead, a support hash is generated and presented to the user in case of server errors caught; it's also stored in the sentry error context for further identification

image

image

image

image

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [x] The UI has been tested both in dark and light mode
- [ ] This PR does not change the UI

PR #3668, opened 2024-01-04
Use tags rather than context to annotate support hashes in Sentry

aerosol:
Correction to #3661 

`set_tags_context` is more applicable here, as we can search by it. Per Sentry moduledoc:

```
  Merges new fields into the `:tags` context, specific to the current process.

  This is used to set fields which should display when looking at a specific
  instance of an error. These fields can also be used to search and filter on.
```

### Tests

### Changelog

### Documentation

### Dark mode


zoldar (approved):
✨

PR #3670, opened 2024-01-06
Limit Custom Properties access in Stats API

RobertJoonas:
### Changes

Starting from the v4 generation of subscription plans, Stats API and Custom Properties are both considered premium features. However, there are two possible cases where the user might have access to Stats API, but not Custom Properties.

1. Enterprise plans with the Stats API feature allowed, but props disallowed
2. An invited user on trial / business plan /grandfathered plan is querying with their own API key, but the site owner does not have access to props

These cases in mind, extra checks are required to ensure that the queried breakdown property, or the used filters do not contain custom properties that the site owner does not have access to. Internal prop keys such as `url` and `path` should always be allowed.

 This PR also build a single function `allowed_for` that acts as a gateway to custom properties, and documents the difficult aspects of why some sites have `allowed_event_props` set to `nil` and so on.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/external_stats_controller.ex
+++ b/lib/plausible_web/controllers/api/external_stats_controller.ex
@@ -11,12 +11,13 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
   end
 
   def aggregate(conn, params) do
-    site = conn.assigns[:site]
+    site = conn.assigns[:site] |> Repo.preload(:owner)
 
     with :ok <- validate_period(params),
          :ok <- validate_date(params),
          query <- Query.from(site, params),
-         {:ok, metrics} <- parse_and_validate_metrics(params, nil, query) do
+         {:ok, metrics} <- parse_and_validate_metrics(params, nil, query),
+         :ok <- ensure_custom_props_access(site, query) do
       results =
         if params["compare"] == "previous_period" do
           {:ok, prev_query} = Plausible.Stats.Comparisons.compare(site, query, "previous_period")
@@ -43,22 +44,20 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
 
       json(conn, %{results: results})
     else
-      {:error, msg} ->
-        conn
-        |> put_status(400)
-        |> json(%{error: msg})
+      err_tuple -> send_json_error_response(conn, err_tuple)
     end
   end
 
   def breakdown(conn, params) do
-    site = conn.assigns[:site]
+    site = conn.assigns[:site] |> Repo.preload(:owner)
 
     with :ok <- validate_period(params),
          :ok <- validate_date(params),
          {:ok, property} <- validate_property(params),
          query <- Query.from(site, params),
          {:ok, metrics} <- parse_and_validate_metrics(params, property, query),
-         {:ok, limit} <- validate_or_default_limit(params) do
+         {:ok, limit} <- validate_or_default_limit(params),
+         :ok <- ensure_custom_props_access(site, query, property) do
       page = String.to_integer(Map.get(params, "page", "1"))
       results = Plausible.Stats.breakdown(site, query, property, metrics, {limit, page})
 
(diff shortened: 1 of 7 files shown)


lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -11,12 +11,13 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
   end
 
   def aggregate(conn, params) do
-    site = conn.assigns[:site]
+    site = conn.assigns[:site] |> Repo.preload(:owner)
```
zoldar, line 14:
```suggestion
    site = Repo.preload(conn.assigns.site, :owner)
```

perhaps, because `site` assign is guaranteed by `AuthorizeStatsApiPlug`?

lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -43,22 +44,20 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
...
-        conn
-        |> put_status(400)
-        |> json(%{error: msg})
+      err_tuple -> send_json_error_response(conn, err_tuple)
     end
   end
 
   def breakdown(conn, params) do
-    site = conn.assigns[:site]
+    site = conn.assigns[:site] |> Repo.preload(:owner)
```
zoldar, line 52: Ditto about

```suggestion
    site = Repo.preload(conn.assigns.site, :owner)
```

lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -172,21 +197,19 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
   end
 
   def timeseries(conn, params) do
-    site = conn.assigns[:site]
+    site = conn.assigns[:site] |> Repo.preload(:owner)
```
zoldar, line 200:
```suggestion
    site = Repo.preload(conn.assigns.site, :owner)
```

test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
```diff
@@ -1,10 +1,48 @@
 defmodule PlausibleWeb.Api.ExternalStatsController.AggregateTest do
   use PlausibleWeb.ConnCase
   import Plausible.TestUtils
+  alias Plausible.Billing.Feature
 
   setup [:create_user, :create_new_site, :create_api_key, :use_api_key]
   @user_id 123
 
+  describe "feature access" do
```
zoldar, line 9: Perhaps it would be good to add at least two extra test cases where `allowed_event_props` are set and filtered/queried property is in and not in that list?

zoldar (approved):
✨

test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
RobertJoonas: Hmm, I think these tests would be redundant since the private `ensure_custom_props_access` function in `ExternalStatsController` is asking for allowed props with the `bypass_setup?` option set to `true`. That option is solely meant to ignore whatever is in the `allowed_event_props` list. Or am I missing the point behind your suggestion? 🤔

lib/plausible_web/controllers/api/external_stats_controller.ex
RobertJoonas: Makes sense, thanks! https://github.com/plausible/analytics/pull/3670/commits/f6092be85b15e166e38a5a9019939bf31c9f2824

test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
zoldar: Yeah, fair point 👍

PR #3683, opened 2024-01-11
Prevent upgrades that would leave the customer locked

RobertJoonas:
### Changes

This PR fixes two cases where a customer can subscribe to a plan, and their account will still remain locked:

1. The `allowance_required` on their grace period is more than what they subscribe to. The fix to this is removing the `allowance_required` field from the grace period. We don't need that anymore because we're now preventing upgrades to plans that don't accommodate the user.
2. We are currently allowing cancelled and paused subscriptions to subscribe to a plan when exceeding its pageview limit by 30% (for 10k plans) or 15% (for all other plans).

    Once these customers upgrade, we'll start measuring their pageview usage in billing cycles (defined by the last payment date). When that happens, it may be true that the customer exceeded their pageview limit for the last two billing cycles which will leave the customer locked.

    * last cycle might be exceeded since we allow to go 15% over the limit, and the site locker only allows 10%

    * penultimate cycle might be exceeded since the customer might have still sent events that we recorded when their subscription was not active

    To make the upgrades impossible in these scenarios, we can make sure that the last cycle is not exceeded when the upgrade happens. We can do that by only allowing them to go over the limit by 10%.
    
    So to conclude, the fix is only applying the 30% / 15% allowance margin to trials, because they cannot have exceeded the penultimate cycle.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/billing/plan_box.ex
```diff
@@ -218,6 +207,28 @@ defmodule PlausibleWeb.Components.Billing.PlanBox do
     """
   end
 
+  defp usage_within_plan_limits?(%{usage: usage, user: user, plan_to_render: plan}) do
+    trial_active_or_ended_recently? =
+      Timex.diff(Timex.today(), user.trial_expiry_date, :days) <= 10
```
RobertJoonas, line 212: In the PR description I mentioned that we'll only be giving the bigger margins to trials. However...

I figured that, instead of checking whether the user is on trial, we should allow a bigger margin up to 10 days after the trial expired to give them some time to upgrade to the plan that they want.

While this makes it theoretically possible that they will be locked after the upgrade (because the first 10 days will then belong to the penultimate cycle, and the usage might be exceeded in those 10 days) - it's extremely unlikely, and we also have manual overrides in the CRM when that happens.

lib/plausible/billing/quota.ex
```diff
@@ -421,9 +425,14 @@ defmodule Plausible.Billing.Quota do
     end
   end
 
-  defp pageview_limit_with_margin(%{monthly_pageview_limit: limit}) do
-    allowance_margin = if limit == 10_000, do: 0.3, else: 0.15
-    ceil(limit * (1 + allowance_margin))
+  defp pageview_limit_with_margin(%{monthly_pageview_limit: limit}, opts) do
+    margin =
+      case Keyword.get(opts, :pageview_allowance_margin) do
```
aerosol, line 430:
```elixir
margin = Keyword.get(opts, :pageview_allowance_margin, @pageview_allowance_margin)
```

should suffice, the likeliness of a non-nil non-float is low?

test/plausible_web/live/choose_plan_test.exs
```diff
@@ -231,6 +231,75 @@ defmodule PlausibleWeb.Live.ChoosePlanTest do
...
+         %{conn: conn, site: site} do
+      generate_usage_for(site, 115_000)
+
+      {:ok, lv, _doc} = get_liveview(conn)
+      doc = set_slider(lv, "100k")
+
+      refute class_of_element(doc, @growth_checkout_button) =~ "pointer-events-none"
+      refute class_of_element(doc, @business_checkout_button) =~ "pointer-events-none"
+
+      generate_usage_for(site, 1)
```
aerosol, line 247: Could you elaborate on the part of the test starting here? It's not like we're resetting the usage, but appending a single event - was that intention here?

lib/plausible/auth/grace_period.ex
```diff
@@ -17,43 +17,39 @@ defmodule Plausible.Auth.GracePeriod do
 
   @type t() :: %__MODULE__{
           end_date: Date.t() | nil,
-          allowance_required: non_neg_integer(),
```
aerosol, line 20: Should we migrate the existing records in a clean-up step?

lib/plausible_web/components/billing/plan_box.ex
zoldar, line 212: While it should not be possible to run this function against a user without `trial_expiry_date` filled, perhaps we should still explicitly account for it? Now `{:error, :invalid_date} <= 10` will evaluable to `false`, which might be okay but also accidental.

test/plausible_web/live/choose_plan_test.exs
zoldar: AFAIU it's about checking a conditions where plan is locked on checkout when it's past 15% margin of pageview allowance (115k vs 115k+1) - so it makes sense to me at least 😅

zoldar (approved):

test/plausible_web/live/choose_plan_test.exs
RobertJoonas: Yes, it's like @zoldar said - the intention of `generate_usage_for(site, 1)` is adding one more event so that the usage allowance margin be exceeded, meaning that the user cannot subscribe to this plan anymore.

lib/plausible/auth/grace_period.ex
RobertJoonas: I don't think that's necessary as we're most likely never dealing with this field again, and it can remain a stale key in the grace period (`map` type) for existing records.

The purpose of `grace_period.allowance_required` was to keep the user locked if they don't upgrade "high enough". We're now preventing such upgrades in the first place which is surely a better experience for customers. So we will never need to read this key again, and I don't think we'll ever need to go back to doing things the old way.

lib/plausible_web/components/billing/plan_box.ex
RobertJoonas: Hmm, yes. Would be nice to use something like `Timex.diff!/3` here but that doesn't exist. @zoldar what do you think about just adding a pattern match clause at the start of this function to assert that it's a `%Date{}` struct? 

```elixir
defp usage_within_plan_limits?(%{usage: usage, user: user, plan_to_render: plan}) do
  %Date{} = user.trial_expiry_date

  trial_active_or_ended_recently? =
    Timex.diff(Timex.today(), user.trial_expiry_date, :days) <= 10
  
  # ...
end
```

test/plausible_web/live/choose_plan_test.exs
aerosol: oh crap, I haven't noticed the `refute` morphing into `assert` 🤦

lib/plausible_web/components/billing/plan_box.ex
zoldar: Nothing better comes to my mind now 👍

RobertJoonas: done in https://github.com/plausible/analytics/pull/3683/commits/69a61531a0a44d329fb8ade6e6a2d92152b2f6da

PR #3686, opened 2024-01-14
Fix weekly interval bug

RobertJoonas:
### Changes

This PR fixes a bug where the main graph weekly interval query sometimes returns 0 visitors for the first week, even when there were visitors.

The bug can be reproduced on any dashboard by looking at the main graph, when the following conditions are met:

1. The site timezone is ahead of UTC
2. The first day of the selected period is not a Monday
3. The `week` interval option is selected

This happened because the start of the month defining the first week was missing the UTC -> site.timezone conversion.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
@@ -688,6 +688,31 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
              }
     end
 
+    test "returns stats for the first week of the month when site.timezone is ahead of UTC", %{
+      conn: conn,
+      site: site
+    } do
+      site =
+        site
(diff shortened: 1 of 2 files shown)


test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -688,6 +688,31 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
              }
     end
 
+    test "returns stats for the first week of the month when site.timezone is ahead of UTC", %{
```
macobo, line 691:
```suggestion
    test "returns stats for the first week of the month when site timezone is ahead of UTC", %{
```

macobo (approved):
Nice simple fix! :partying_face: (hi!)

aerosol (approved):

zoldar (approved):
✨

PR #3687, opened 2024-01-15
Support `contains`/`matches` for custom properties

macobo:
### Changes

This PR a filtering with `contains`/`matches` for custom properties, e.g. "Filter where (custom property) url contains foo"

image

First PR :raised_hands: 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ### Added
+- Allow filtering with `contains`/`matches` operator for custom properties
 - Add `referrers.csv` to CSV export
 - Add a new Properties section in the dashboard to break down by custom properties
 - Add `custom_props.csv` to CSV export (almost the same as the old `prop_breakdown.csv`, but has different column headers, and includes props for pageviews too, not only custom events)
--- a/assets/js/dashboard/stats/modals/prop-filter-modal.js
+++ b/assets/js/dashboard/stats/modals/prop-filter-modal.js
@@ -40,11 +40,15 @@ function PropFilterModal(props) {
 
   const fetchPropValueOptions = useCallback(() => {
     return (input) => {
+      if (formState.prop_value?.type === FILTER_TYPES.contains) {
+        return Promise.resolve([])
+      }
+
       const propKey = formState.prop_key?.value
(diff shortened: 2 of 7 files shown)


aerosol:
There's something wrong with re-applying the filter when modifying the custom props breakdown query:

https://github.com/plausible/analytics/assets/173738/9e7940d0-eb90-4a98-9e56-618ac75cddd6

aerosol:
On a freshly seeded db (you can repro via `mix ecto.reset`), Total Visits doubles when a non-existing property value is supplied to the query:

https://github.com/plausible/analytics/assets/173738/52327b2a-ebcf-4415-bfea-2cdc24ec9b88

I thought it was due to seed data being nonsense, but modifying the test you've added, also returns extra rows that shouldn't be included?

```diff
diff --git a/test/plausible_web/controllers/api/stats_controller/pages_test.exs b/test/plausible_web/controllers/api/stats_controller/pages_test.exs
index 339a4bab3..ba9aa0fed 100644
--- a/test/plausible_web/controllers/api/stats_controller/pages_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/pages_test.exs
@@ -103,7 +103,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
         build(:pageview, pathname: "/5")
       ])
 
-      filters = Jason.encode!(%{props: %{"prop" => "~bar"}})
+      filters = Jason.encode!(%{props: %{"prop" => "~bar|pppp"}})
       conn = get(conn, "/api/stats/#{site.domain}/pages?period=day&filters=#{filters}")
 
       assert json_response(conn, 200) == [
```

macobo:
@aerosol thank you for the great feedback. I've updated the branch.

(1) Regarding breakdowns, this was a legit bug in the code, see fix in https://github.com/plausible/analytics/pull/3687/commits/43772bc1c950b8957ce1a8cb39fe1a00f0626b16

(2) seems to be about how regex is internally escaped. I hoped that merging https://github.com/plausible/analytics/pull/3634 would fix this, but it did not. Given that this behavior already exists on master, I'll fix this in a separate PR, [filed it in my worklog](https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/6952398131).

aerosol (approved):

PR #3689, opened 2024-01-15
Prioritize X-Plausible-IP header over everything else

cnkk:
Since we already need to replicate this logic in our load balancer setup for session stickiness, we now move the responsibility completely to the load balancers.

The correct IP address is now fully looked up by the load balancer and passed to the backends as `X-Plausible-IP` header.

### DIFF
--- a/test/plausible_web/controllers/api/external_controller_test.exs
+++ b/test/plausible_web/controllers/api/external_controller_test.exs
@@ -970,7 +970,7 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       params = %{
         name: "pageview",
         domain: site.domain,
-        url: "http://gigride.live/"
+        url: "http://example.com/"
       }
 
       conn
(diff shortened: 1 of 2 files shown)


test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -963,6 +963,29 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
...
     end
 
+    test "prioritizes x-plausible-ip header over everything else", %{
+      conn: conn,
+      site: site
+    } do
+      params = %{
+        name: "pageview",
+        domain: site.domain,
+        url: "http://gigride.live/"
```
aerosol, line 973:
```suggestion
        url: "http://example.com/"
```

dunno what that previous URL was, but we should probably stop using it 😅

aerosol (approved):

test/plausible_web/controllers/api/external_controller_test.exs
aerosol: Ref https://github.com/plausible/analytics/pull/3690

PR #3698, opened 2024-01-17
Update ua_inspector database

ukutaht:
### Tests

### Changelog

### Documentation

### Dark mode


aerosol (changes requested):
In order to actually update, don't we have to pin a SHA from https://github.com/matomo-org/device-detector/commits/master/ in https://github.com/plausible/analytics/blob/master/config/config.exs#L39-L41 ?

Ref: https://github.com/plausible/analytics/pull/3263

ukutaht:
Ah thanks, I was confused why the data files didn't update with mix ua_inspector.download

I bumped the version

aerosol (approved):

PR #3704, opened 2024-01-21
Improve query tracing

RobertJoonas:
### Changes

This PR will fix the unintentional behaviour of `event:goal` breakdown property being overwritten by `event:name` in the OpenTelemetry span.

Also, as we'll want to have some visibility on how many customers are currently using the `event:goal` property in filters, I added `filter_keys` into the query trace. I ran the following Stats API query on staging:

```
https://staging.plausible.io/api/v1/stats/breakdown
  ?site_id=plausible.io
  &property=event:goal
  &metrics=visitors,pageviews,events
  &filters=event:page==/register;visit:browser==Chrome;event:goal==Signup
```

And the fields were reported as follows (the joined string of `breakdown_metrics` and `filter_keys` are always in alphabetical order):

image

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/breakdown.ex
```diff
@@ -16,14 +16,16 @@ defmodule Plausible.Stats.Breakdown do
...
+
+  def breakdown(site, query, "event:goal" = property, metrics, pagination, opts) do
     site = Plausible.Repo.preload(site, :goals)
 
     {event_goals, pageview_goals} = Enum.split_with(site.goals, & &1.event_name)
     events = Enum.map(event_goals, & &1.event_name)
     event_query = %Query{query | filters: Map.put(query.filters, "event:name", {:member, events})}
 
-    trace(query, property, metrics)
+    trace(query, property, metrics, opts)
```
zoldar, line 28: IMHO checking `skip_tracing` option here directly and not calling `trace` when it's set would be better in terms of code readbility. Please feel free to ignore though.

zoldar (approved):

PR #3705, opened 2024-01-21
Add ReleaseRuntimeProvider for exs config at runtime

Kloenk:
### Changes

This lets the user define elixir code as config.

I plane to use this in my approach for OIDC as with this the config can easily define functions to apply roles base on claims in the token.

### Tests

### Changelog
At this time the config is just an extra way to configure, so the user does not yet have to be aware of it.

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/config/release_runtime_provider.ex
+++ b/lib/plausible/config/release_runtime_provider.ex
@@ -1,0 +1,40 @@
+defmodule Plausible.Config.ReleaseRuntimeProvider do
+  @moduledoc false
+  @behaviour Config.Provider
+
+  @impl Config.Provider
+  def init(opts), do: opts
+
+  @impl Config.Provider
+  def load(config, opts) do
+    config_path =
+      opts[:config_path] || System.get_env("PLAUSIBLE_CONFIG_PATH") ||
+        "/etc/plausible/config.exs"
+
+    with_runtime_config =
+      if File.exists?(config_path) do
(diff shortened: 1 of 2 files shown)


ruslandoga:
👋 @Kloenk 

I think this should be possible already by mounting a custom `config/runtime.exs`.

Kloenk:
As I don't use docker and my package gets installed into a immutable place (nix) this is less convenient. So the approach in this PR would be favorable for me, but I could work around it as well.

ruslandoga:
Right, I think allowing extra config might be a useful escape hatch for the configurations not covered by the current env vars (like https://github.com/plausible/analytics/discussions/3799). I explored a bit and it seems like it isn't necessary to implement a custom config reader. The default one could be reused and there are multiple ways to do it, some hackier than others.

### Hacky way one

We can check for an env var like `EXTRA_CONFIG_PATH` and import the config at this location conditionally.

For this we'd add the following at the end of the current [config/runtime.exs](https://github.com/plausible/analytics/blob/master/config/runtime.exs)
```elixir
if extra_config_path = System.get_env("EXTRA_CONFIG_PATH") do
  Process.put({Config, :imports}, []) # this is the hacky part
  import_config extra_config_path
end
```

ruslandoga:
I'm going to close this as https://github.com/plausible/analytics/pull/3906 has been merged

PR #3708, opened 2024-01-22
Remove ARRAY JOINs

macobo:
### Changes

This PR refactors the backend code to use `meta.value[indexOf(meta.key, 'custom_property')]` over array joins. Doing so has a few benefits:

1. It allows us to more easily add support for multiple custom property filters.
2. Performance - tests showed ~20% performance benefit for some production queries (see [basecamp](https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/6946026076) for an example)

Note that ARRAY JOINs were retained in places where it still makes sense - e.g. `Plausible.Stats.CustomProps#fetch_prop_names`.

Based on top of https://github.com/plausible/analytics/pull/3687 as that introduced new LEFT JOINs.

### DIFF
--- a/lib/plausible/stats/fragments.ex
+++ b/lib/plausible/stats/fragments.ex
@@ -97,6 +97,46 @@ defmodule Plausible.Stats.Fragments do
     end
   end
 
+  @doc """
+  Returns whether a key (usually property) exists under `meta.key` array or similar.
+
+  This macro is used for operating on custom properties.
+
+  ## Examples
+
+  `has_key(e, :meta, "some_property_name")` expands to SQL `has(meta.key, "some_property_name")`
+  """
+  defmacro has_key(table, meta_column, key) do
+    quote do
+      fragment(
+        "has(?, ?)",
+        field(unquote(table), unquote(String.to_atom("#{meta_column}.key"))),
+        unquote(key)
+      )
+    end
+  end
+
+  @doc """
+  Returns value of a key (usually property) under `meta.value` array or similar.
+
+  This macro is used for operating on custom properties.
+  Callsites should also check whether key exists first in SQL via `has_key` macro.
+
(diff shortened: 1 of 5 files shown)


lib/plausible/stats/fragments.ex
```diff
@@ -97,6 +97,46 @@ defmodule Plausible.Stats.Fragments do
...
+
+  ## Examples
+
+  `has_key(e, :meta, "some_property_name")` expands to SQL `has(meta.key, "some_property_name")`
+  """
+  defmacro has_key(table, meta_column, key) do
+    quote do
+      fragment(
+        "has(?, ?)",
+        field(unquote(table), unquote(String.to_atom("#{meta_column}.key"))),
```
aerosol, line 113: Always use `String.to_exiting_atom` - atoms are not garbage collected

lib/plausible/stats/fragments.ex
```diff
@@ -97,6 +97,46 @@ defmodule Plausible.Stats.Fragments do
...
+
+  ## Examples
+
+  `get_by_key(e, :meta, "some_property_name")` expands to SQL `meta.value[indexOf(meta.key, "some_property")]`
+  """
+  defmacro get_by_key(table, meta_column, key) do
+    quote do
+      fragment(
+        "?[indexOf(?, ?)]",
+        field(unquote(table), unquote(String.to_atom("#{meta_column}.value"))),
```
aerosol, line 133: ditto, existing atoms. If the only input is `:meta` or `:entry_meta` maybe best to be explicit about it and make the conversion handle those two cases only:

```
defp meta_column("meta"), do: :"meta.value"
defp meta_column("entry_meta"), do: :"entry_meta.value"
```

aerosol (approved):

PR #3709, opened 2024-01-22
Github Actions: Run tests on all PRs

macobo:
Chained PRs (like https://github.com/plausible/analytics/pull/3708) don't currently run tests. This PR fixes that.

### DIFF
--- a/.github/workflows/build-private-images.yml
+++ b/.github/workflows/build-private-images.yml
@@ -5,7 +5,6 @@ on:
     branches: [master, stable]
     tags: ['r*']
   pull_request:
-    branches: [master]
     types: [synchronize, labeled]
 
 concurrency:
--- a/.github/workflows/terraform-e2e.yml
+++ b/.github/workflows/terraform-e2e.yml
@@ -4,11 +4,9 @@ on:
   push:
     branches:
       - master
-    paths: 
-      - 'test/e2e/**.tf' 
+    paths:
+      - 'test/e2e/**.tf'
   pull_request:
-    branches:
-      - master
     paths:
       - 'test/e2e/**.tf'
 
@@ -43,7 +41,7 @@ jobs:
       - name: Terraform Init
         id: init
         run: terraform init
-      
+
       - name: Terraform Validate
         id: validate
         run: terraform validate -no-color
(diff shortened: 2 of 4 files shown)


aerosol:
Any impact on PRs submitted from forks having no access to secrets? I can't recall what's the current situation on this. cc @cnkk

macobo:
Valid question - there is 0 impact on this. 

Quick search showed these resources (unrelated to this PR):
- https://stackoverflow.com/questions/76952023/how-to-make-github-actions-safely-access-secrets-for-prs-created-from-forks
- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories

.github/workflows/terraform-e2e.yml
```diff
@@ -4,11 +4,9 @@ on:
   push:
     branches:
       - master
-    paths: 
-      - 'test/e2e/**.tf' 
+    paths:
+      - 'test/e2e/**.tf'
   pull_request:
-    branches:
```
cnkk, line 10: I'd suggest to still keep this for this workflow

macobo: I don't mind - mind giving a bit of context why?

.github/workflows/build-private-images.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches: [master, stable]
     tags: ['r*']
   pull_request:
-    branches: [master]
```
aerosol, line 8: Why would we build private images for non-master merges automatically? Not only this is what effectively goes live but also don't we hit dockerhub too often then?

.github/workflows/terraform-e2e.yml
cnkk: Ah nevermind!
I just want to avoid to run this workflow on every PR, but we already do this by scoping to path `test/e2e/**.tf`

.github/workflows/build-private-images.yml
aerosol: Unless I also misunderstood the yaml haha, disregard in that case

cnkk (approved):

PR #3715, opened 2024-01-22
Store referrers from android apps

ukutaht:
### Changes

Overall task is figuring out if we can separate the Google Discover referral source from Google (the search engine). Some sources say that the referrer `android-app://com.google.android.googlequicksearchbox` is Google Discover but other sources say different things. I want to check our actual data to see which bloggers are most correct.

We currently only store referral sources with `http` and `https` protocol in the URI. With this change we'll also be storing referrers with `android-app://` protocol. This will allow me to correlate this referrer data with proprietary numbers from Google Search Console to see if the number of visitors is similar.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/external_controller_test.exs
+++ b/test/plausible_web/controllers/api/external_controller_test.exs
@@ -444,6 +444,26 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert session.referrer_source == "indiehackers.com"
     end
 
+    test "if the referrer is not http, https, or android it is ignored", %{conn: conn, site: site} do
+      params = %{
+        name: "pageview",
+        url: "http://example.com/",
+        referrer: "ftp://wat",
+        domain: site.domain
+      }
+
+      conn =
+        conn
+        |> put_req_header("user-agent", @user_agent)
+        |> post("/api/event", params)
+
+      pageview = get_event(site)
+
+      assert response(conn, 202) == "ok"
+      assert pageview.referrer_source == ""
+      assert pageview.referrer == ""
+    end
+
     test "stores referrer from android app", %{conn: conn, site: site} do
       params = %{
         name: "pageview",
--- a/test/plausible_web/controllers/api/external_controller_test.exs
+++ b/test/plausible_web/controllers/api/external_controller_test.exs
@@ -480,7 +480,8 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       session = get_created_session(site)
 
       assert response(conn, 202) == "ok"
-      assert session.referrer_source == "some.android.app"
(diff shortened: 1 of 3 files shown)


test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -434,11 +434,11 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       assert pageview.referrer_source == "indiehackers.com"
     end
 
-    test "if the referrer is not http or https, it is ignored", %{conn: conn, site: site} do
```
macobo, line 437: It should be worth keeping this test, just with a random protocol?

macobo:
> Some sources say that the referrer android-app://com.google.android.googlequicksearchbox is Google Discover but other sources say different things. I want to check our actual data to see which bloggers are most correct.

Could you link to these sources for future reference?

test/plausible_web/controllers/api/external_controller_test.exs
```diff
@@ -450,7 +450,7 @@ defmodule PlausibleWeb.Api.ExternalControllerTest do
       pageview = get_event(site)
 
       assert response(conn, 202) == "ok"
-      assert pageview.referrer_source == ""
+      assert pageview.referrer_source == "some.android.app"
```
macobo, line 453: It seems we should we start storing the referrer_protocol as well - otherwise impossible to distinguish them from web traffic.

ukutaht:
This one says it's definitely `android-app://com.google.android.googlequicksearchbox`: https://www.cueblocks.com/blog/how-to-optimize-track-google-discover-in-google-analytics/

This one says it's more complicated: https://valentin.app/discover-tracking.html


> It seems we should we start storing the referrer_protocol as well - otherwise impossible to distinguish them from web traffic.



I was on the fence about maybe storing these with the protocol. I've added it now, although it's in the same `referrer` field instead of a separate field for the protocol

macobo (approved):

macobo:
Didn't have a quick way of testing this locally - so take my approval with a grain of salt. The tests speak for themselves though :+1:

zoldar (approved):
Just to double-check, I'm assuming imports are out of scope of this change, for now at least, right?

https://github.com/plausible/analytics/blob/9c2c1ee4232158b7a27de87550ffadec9e2b8b88/lib/plausible/imported/universal_analytics.ex#L234-L238

ukutaht:
> Just to double-check, I'm assuming imports are out of scope of this change, for now at least, right?

Correct. This change is irrelevant for imports.

ukutaht:
More context for future reference: https://github.com/plausible/analytics/discussions/1337

PR #3716, opened 2024-01-23
Generalize Cache/Warmers - extract only specifics to `Site.Cache`

aerosol:
### Changes

This PR is an attempt at generalizing `Plausible.Site.Cache`. For the upcoming work (such as block/allow lists for the "Shields" feature), we like to keep the opportunity to implement parallel posrgres caches without too much duplication.

The setup has been simplified, so it's now more clear by looking at the supervision tree what caches are started and where the details are implemented. We're still able to swap out the underlying cache implementation when the time comes (will be done in a single place).

A general `Plausible.Cache` behavior has been extracted, so that any similarly acting, periodic database fetchers can collect results retrievable by key in a similar fashion.

The change is transparent from user perspective.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/cache.ex
```diff
@@ -0,0 +1,210 @@
...
+  Even though the Cachex process is started, cache access is disabled
+  during tests via the `:plausible, #{__MODULE__}, enabled: bool()` application env key.
+  This can be overridden on case by case basis, using the child specs options.
+
+  The `base_db_query/0` callback is used to generate the base query that is 
+  executed on every cache refresh. 
+
+  There are two modes of refresh operation: `:all` and `:updated_recently`;
+  the former will invoke the query as is and clear all the existing entries,
+  while the later will attempt to limit the query to only the records that
```
aerosol, line 23:
```suggestion
  while the latter will attempt to limit the query to only the records that 
```

lib/plausible/cache.ex
```diff
@@ -0,0 +1,210 @@
+defmodule Plausible.Cache do
+  @moduledoc """
+  Caching interface specific for Plausible. Usage:
+
+      use Plausible.Cache
+
+      # Implement the callbacks required.
+      # Optionally override `unwrap_cache_keys/1`.
+      # Populate the cache with `Plausible.Cache.Warmer`
```
aerosol, line 9:
```suggestion
      # Populate the cache with `Plausible.Cache.Warmer`.
```

lib/plausible/cache/warmer.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Plausible.Cache.Warmer do
+  @moduledoc """
+  A periodic cache warmer.
+
+  Child specification options available:
+
+    * `cache_impl` - module expected to implement `Plausible.Cache` behaviour
+    * `interval` - the number of milliseconds for each warm-up cycle
+    * `cache_name` - defaults to cache_impl.name() but can be overriden for testing
```
aerosol, line 9:
```suggestion
    * `cache_name` - defaults to cache_impl.name() but can be overridden for testing
```

ukutaht (approved):

PR #3718, opened 2024-01-23
Add `event:goal` property to Stats API

RobertJoonas:
### Changes

_Reviewing commit by commit highly recommended._

This PR builds "official" support for the `event:goal` property in Stats API, both in filters, and as a breakdown property. I'm stressing "official" since the majority of functionality was always there and accessible, just not documented and missing several tests.

#### Breaking change

This PR stops the `event:goal` property breakdown from returning `prop_names` used for each goal. Since custom properties are not considered a part of the goals feature anymore, we've decided to remove this coupling between goals and props.

#### Filtering

This PR also adds Stats API support for `:matches`, `:member`, and `:matches_member` goal filter types. All the functionality already exists in the lower level queries - it's only a matter of parsing the filter expression in the `Query.from` function.

### Tests

### Changelog

### Documentation
- [x] PR is waiting (https://github.com/plausible/docs/pull/460)

### Dark mode


test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
```diff
@@ -116,22 +116,26 @@ defmodule PlausibleWeb.Api.ExternalStatsController.AggregateTest do
...
+          get(conn, "/api/v1/stats/aggregate", %{
+            "site_id" => site.domain,
+            "period" => "30d",
+            "metrics" => "pageviews,visit_duration",
+            "filters" => "#{prop}==some_value"
+          })
+
+        assert json_response(conn, 400) == %{
+                 "error" =>
+                   "Session metric `visit_duration` cannot be queried when using a filter on `#{prop}`."
```
ukutaht, line 138: I'm not sure why this limitation was added but I do know that customers have asked for the ability to see visit duration when using a goal filter.

Out of scope for this PR but can you think of any reasons to keep this limitation? If not I'd like to start allowing this

lib/plausible/stats/breakdown.ex
```diff
@@ -436,24 +436,6 @@ defmodule Plausible.Stats.Breakdown do
     )
   end
 
-  defp do_group_by(
```
ukutaht, line 439: Do you have the context why this was added and what allowed us to remove it? I'm just curious

lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -271,6 +274,37 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
 
   defp validate_interval(_), do: :ok
 
+  defp validate_goal_filter(site, %{"event:goal" => goal_filter}) do
+    configured_goals = Plausible.Goals.for_site(site) |> Enum.map(&to_string/1)
+
+    goals_in_filter =
+      case goal_filter do
```
ukutaht, line 286: Can probably be rewritten more concisely using [List.wrap/1](https://hexdocs.pm/elixir/1.12/List.html#wrap/1)

ukutaht: `Tuple.elem(goal_filter, 1) |> List.wrap`

ukutaht (commented):
Overall I like the choices made here 👍

test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
RobertJoonas: Hmm, yes, I agree. We could display visit duration with a goal filter. Currently it's not available in the dashboard either so I guess this could be added to both Stats API and the dashboard at the same time.

lib/plausible/stats/breakdown.ex
RobertJoonas: I looked at the [commit](https://github.com/plausible/analytics/commit/7d1b1214e2#diff-44853387040800d84a97394436f012261cfeec8e6174de537c09735b8359d2aeR299-R312) that added this function, but didn't find any reference of the `do_group_by` function call with the `event:page_match` property.

I suspect that this was used as a pageview goal breakdown property in the Stats API at some point, since the properties were not validated back then - the only thing that was validated was that the breakdown_property existed:

image

Anyway, this is definitely not an allowed property in the API anymore, and safe to remove :)

lib/plausible_web/controllers/api/external_stats_controller.ex
RobertJoonas: Nice, thanks! https://github.com/plausible/analytics/pull/3718/commits/bbd002cda4345bfbcdf7984a2db10fe99b086b68 Ended up pattern matching in the function definition instead of `Tuple.elem` 🙂

ukutaht (approved):

PR #3655, opened 2023-12-28
Skip breakdown of imported pages with pageviews=0

aerosol:
### Changes

Similar to https://github.com/plausible/analytics/pull/3123 - apparently some imported pages google sends us look like: 

```
┌─visitors─┬─pageviews─┬─exits─┬─time_on_page─┐
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
│        1 │         0 │     0 │            0 │
└──────────┴───────────┴───────┴──────────────┘
```

It's questionable whether we should import those records at all, but that's a matter addressable separately. Regardless, we should exclude those from breakdowns to avoid crashing the exit rate calculation via division by zero.

### Tests
- [x] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #3720, opened 2024-01-24
Work around add_percentages errors due to race condition

macobo:
Basecamp issue: https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/6964655047

We occasionally seem to get divided-by-zero errors with add_percentages function. As this seems to mainly happen with low-traffic sites on realtime graphs, I believe it's caused by a race condition where `total_visitors` (which is calculated later) returns 0 but the previous calculations returned results.

Not fully confident in this.

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -1328,10 +1328,16 @@ defmodule PlausibleWeb.Api.StatsController do
        when not is_map_key(query.filters, "event:goal") do
     %{visitors: %{value: total_visitors}} = Stats.aggregate(site, query, [:visitors])
 
-    breakdown_result
-    |> Enum.map(fn stat ->
-      Map.put(stat, :percentage, Float.round(stat.visitors / total_visitors * 100, 1))
-    end)
+    # :HACK: With realtime graphs, total_visitors might report as 0 while there are stats
+    #   As division by 0 would result in errors, just don't include percentages
+    if total_visitors == 0 do
+      breakdown_result
+    else
+      breakdown_result
+      |> Enum.map(fn stat ->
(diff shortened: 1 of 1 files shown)


aerosol:
Don't think that's a race condition, most likely it's some data error; I think fixing it here is treating the symptom not the cause. PRs addressing the same error in different contexts:
  - https://github.com/plausible/analytics/pull/3655
  - https://github.com/plausible/analytics/pull/3123

I think we should consider doing those on case by case basis to at least understand what's going on.

aerosol:
If you're right and it's a race, shouldn't we be able to reproduce that state with a test though?

aerosol:
@macobo I'll try writing a test for this

aerosol:
I think your hypothesis is correct, the second aggregate call on 30m+1s will return 0 after all. I'm wondering if we could make sure the second call is made with the same time boundaries -- we could carry "now" in `Query` struct for example and calculate the offsets based on that. This way the actual ClickHouse query `Plausible.Query` evaluates to would be idempotent. WYT?

macobo:
I think that would be a better fix - at least until we start allowing users to send data with timestamps from the past.

aerosol:
> at least until we start allowing users to send data with timestamps from the past

can you elaborate on that? what do you mean exactly - i.e. how would that work and is that an actual plan?

macobo:
Discussed this on basecamp - historical data coming in shouldn't be a problem. As such, fixing 'now' to be the same across all queries being made would be the ideal fix.

Updated the original issue with the proposed fix and made it a free-for-all. https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/6970466136

ukutaht:
> start allowing users to send data with timestamps from the past
> how would that work and is that an actual plan?

for context, eventually we will need to be able to ingest historical data. This is required for integrating with other systems, most importantly Segment. But it's pretty far down the list at the moment

PR #3721, opened 2024-01-24
Support using matches/contains for most filters

macobo:
### Changes

This PR adds support for filters like `Browser contains Ch`.

image

The backend already supported this, only needed to add frontend support.

I've excluded 'goal' and 'screen' filters from this as (IMO) these don't make sense to use contains with.

This also refactors the behavior on Devices/OS/Sources sections not to 'zoom in' to specifics unless filtering on a single fixed value.


metmarkosaric:
Nice! I think if contains is already supported for goals in the backend, then no reason not to add it to the dashboard too. We've had some requests for this. Mostly to add "is not" to goals but also some requests to add "contains"

macobo:
I tested adding support for `is_not` for goals - sadly blows up as the backend doesn't support _that_ (or contains). From a users perspective I'm not sure what the semantics of this are either.

I think it's worth prioritizing more filters for goals separately.

ukutaht:
Looks good but at the moment we should also exclude `location` filters from supporting contains. Since they are stored in the database as IDs, searching by name does not work.

ukutaht (approved):

PR #3723, opened 2024-01-24
Make subsequent query boundaries idempotent

aerosol:
This should prevent time-sensitive race conditions on subsequent `Plausible.Stats.*` calls for the same query. Not sure if there are more places where now() is
dynamic relative to query life cycle, but it's a start.

Ref https://github.com/plausible/analytics/pull/3720 - thanks @macobo for bringing it up!

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -16,8 +16,9 @@ defmodule Plausible.Stats.Query do
   @type t :: %__MODULE__{}
 
   def from(site, params) do
-    now = NaiveDateTime.utc_now() 
-          |> NaiveDateTime.truncate(:second)
+    now =
+      NaiveDateTime.utc_now()
+      |> NaiveDateTime.truncate(:second)
 
     query =
       __MODULE__
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -7,7 +7,8 @@ defmodule Plausible.Stats.Query do
             filters: %{},
             sample_threshold: 20_000_000,
             imported_data_requested: false,
-            include_imported: false
+            include_imported: false,
+            now: nil
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{FilterParser, Interval}
@@ -15,8 +16,13 @@ defmodule Plausible.Stats.Query do
   @type t :: %__MODULE__{}
 
   def from(site, params) do
+    now = NaiveDateTime.utc_now() 
+          |> NaiveDateTime.truncate(:second)
+
     query =
-      query_by_period(site, params)
+      __MODULE__
+      |> struct!(now: now)
+      |> put_period(site, params)
       |> put_interval(params)
       |> put_parsed_filters(params)
       |> put_imported_opts(site, params)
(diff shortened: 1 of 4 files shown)


test/plausible/stats/query_test.exs
```diff
@@ -15,6 +15,19 @@ defmodule Plausible.Stats.QueryTest do
...
   end
 
+  @tag :slow
+  test "keeps current timestamp so that utc_boundaries don't depend on time passing by", %{
+    site: site
+  } do
+    q1 = Query.from(site, %{"period" => "realtime"})
+    q2 = Query.from(site, %{"period" => "30m"})
+    boundaries1 = Plausible.Stats.Base.utc_boundaries(q1, site)
+    boundaries2 = Plausible.Stats.Base.utc_boundaries(q2, site)
```
aerosol, line 25: it's a bit silly but we never test those functions in isolation, everything's tested from behind controllers 😕

lib/plausible/stats/query.ex
```diff
@@ -7,16 +7,23 @@ defmodule Plausible.Stats.Query do
...
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{FilterParser, Interval}
 
   @type t :: %__MODULE__{}
 
   def from(site, params) do
+    now =
+      NaiveDateTime.utc_now()
+      |> NaiveDateTime.truncate(:second)
```
wojtekmach, line 21: since Elixir 1.15 you can also:

```suggestion
    now = NaiveDateTime.utc_now(:second)
```

aerosol: Aw thanks!

ukutaht:
Super impressed with both the root cause analysis and the proposed solution in this PR

lib/plausible/stats/query.ex
```diff
@@ -220,9 +216,10 @@ defmodule Plausible.Stats.Query do
   defp put_imported_opts(query, site, params) do
     requested? = params["with_imported"] == "true"
 
-    query
```
macobo, line 223: Not-blocking elixir question: What is the advantage of structs vs maps here?

test/plausible/stats/query_test.exs
```diff
@@ -15,6 +15,19 @@ defmodule Plausible.Stats.QueryTest do
     {:ok, site: site, user: user}
   end
```
macobo, line 17: I think a unit test that checks whether `Query.from` has .now defined and what it is would be worthwhile.

macobo (approved):
Nice clear fix - learned a lot just reading it! One small question and one small nit under comments, nothing blocking.

lib/plausible/stats/query.ex
aerosol: structs _are_ maps, just with a `__struct__` key:

```elixir
iex(1)> IO.inspect(%URI{}, structs: false)
%{
  port: nil,
  scheme: nil,
  path: nil,
  host: nil,
  __struct__: URI,
  query: nil,
  fragment: nil,
  authority: nil,
  userinfo: nil
}
```


There's not much difference here:

```
  • when building a struct, as in struct!(SomeStruct, key: :value), it is
    equivalent to %SomeStruct{key: :value} and therefore this function will
    check if every given key-value belongs to the struct. If the struct is
    enforcing any key via @enforce_keys, those will be enforced as well;
  • when updating a struct, as in struct!(%SomeStruct{}, key: :value), it
    is equivalent to %SomeStruct{struct | key: :value} and therefore this
    function will check if every given key-value belongs to the struct.
    However, updating structs does not enforce keys, as keys are enforced only
    when building;
```

There's `struct/2` too (without the bang) that will discard unknown keys:

```elixir
iex(2)> struct(URI, hum: 1, host: "foo")
%URI{
  scheme: nil,
  userinfo: nil,
  host: "foo",
  port: nil,
  path: nil,
  query: nil,
  fragment: nil
}
iex(3)> struct!(URI, hum: 1, host: "foo")
** (KeyError) key :hum not found
    (elixir 1.16.0) anonymous fn/2 in URI.__struct__/1
    (elixir 1.16.0) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir 1.16.0) lib/kernel.ex:2433: Kernel.struct!/2
    iex:5: (file)
```

Mostly a personal preference, the function call here arguably reads less bloated at the cost of compile time check I suppose?

test/plausible/stats/query_test.exs
aerosol: 4eaf8a031

PR #3740, opened 2024-01-31
Bugfix: available features for expired trials, no subscriptions

aerosol:
### Changes

This PR fixes a bug where expired trials with no subscriptions would still have access to all the business tier features. I did a minor code re-org, reading commit by commit should help. The actual bugfix is https://github.com/plausible/analytics/pull/3740/commits/7365e58636efa80c90b79fa9903dd4dff105e2c3
On top of that, I've changed the `on_trial?` implementation so it always returns `true` for `small` builds.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/users.ex
+++ b/lib/plausible/users.ex
@@ -8,8 +8,24 @@ defmodule Plausible.Users do
   import Ecto.Query
 
   alias Plausible.Auth
-  alias Plausible.Billing.Subscription
+  alias Plausible.Billing.{Subscription, Subscriptions}
   alias Plausible.Repo
+
+  on_full_build do
+    def on_trial?(%Auth.User{trial_expiry_date: nil}), do: false
+
+    def on_trial?(user) do
+      user = with_subscription(user)
+      not Subscriptions.active?(user.subscription) && trial_days_left(user) >= 0
+    end
+  else
+    def on_trial?(_), do: false
+  end
+
+  @spec trial_days_left(Auth.User.t()) :: integer()
+  def trial_days_left(user) do
+    Timex.diff(user.trial_expiry_date, Timex.today(), :days)
+  end
 
   @spec update_accept_traffic_until(Auth.User.t()) :: Auth.User.t()
   def update_accept_traffic_until(user) do
@@ -24,7 +40,7 @@ defmodule Plausible.Users do
       user = with_subscription(user)
 
       cond do
-        Plausible.Billing.on_trial?(user) ->
+        Plausible.Users.on_trial?(user) ->
           Timex.shift(user.trial_expiry_date,
             days: Auth.User.trial_accept_traffic_until_offset_days()
           )
@@ -86,7 +102,7 @@ defmodule Plausible.Users do
   end
 
   defp last_subscription_query(user_id) do
-    from(subscription in Plausible.Billing.Subscription,
+    from(subscription in Subscription,
       where: subscription.user_id == ^user_id,
       order_by: [desc: subscription.inserted_at],
       limit: 1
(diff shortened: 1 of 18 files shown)


test/plausible/billing/quota_test.exs
```diff
@@ -499,6 +499,11 @@ defmodule Plausible.Billing.QuotaTest do
   end
 
   describe "allowed_features_for/1" do
+    test "users with expired trials have no access to subscription features" do
+      user = insert(:user, trial_expiry_date: ~D[2023-01-01])
+      assert [] == Quota.allowed_features_for(user)
```
aerosol, line 504: @RobertJoonas I'm not sure if we should return `[Goals]` here - but I guess not. We don't seem to be checking if that feature the user has access to anyway.

lib/plausible/users.ex
```diff
@@ -11,6 +11,23 @@ defmodule Plausible.Users do
...
+  @spec on_trial?(Auth.User.t()) :: boolean()
+  on_full_build do
+    def on_trial?(%Auth.User{trial_expiry_date: nil}), do: false
+
+    def on_trial?(user) do
+      user = with_subscription(user)
+      not Plausible.Billing.Subscriptions.active?(user.subscription) && trial_days_left(user) >= 0
+    end
+  else
+    def on_trial?(_), do: true
```
aerosol, line 23: `small` build users can be treated as always on trial when this function is called (the actual "extra" features the small build doesn't ship with are disabled anyway)

test/plausible/billing/quota_test.exs
aerosol: Should we also behave differently here in case the user's trial_expiry_date is nil? I saw some conditionals like that.

RobertJoonas: Now that I start to think about it, I do think we should return `[Goals]` for expired trials at this stage, just to keep things consistent.

So far, we've always considered Goals a "free" feature. If you look up the `Plausible.Billing.Feature.Goals` module def, you'll notice the `free: true` field there which basically overrides the `feature_mod.check_availability` function to always return `true`.

Around the codebase we're using both `allowed_features_for/1` and `Goals.check_availability/1`. The latter will always return `true`, which is why I think we should also return `[Goals]` for expired trials from `allowed_features_for/1`.

I don't think it's a problem that expired trials can configure new goals in their dashboard. True, the labels look a bit weird if we're saying that they can't use revenue goals because it requires a subscription, etc... But they'll see plenty of CTA's for upgrade around the app, so I'd say we can keep that problem for another day :)

RobertJoonas: > Should we also behave differently here in case the user's trial_expiry_date is nil? I saw some conditionals like that.

I don't think that's relevant in terms of checking feature access. We mainly do that check because we don't want to let those users to upgrade, as they are "invited users". Meaning that they have access to some dashboard but they're not the ones responsible for billing.

RobertJoonas: What's more, I think we should remove the `free` option from the Feature context and move it to the Quota context instead. E.g. a module attribute like `@free_features [Goals]`. That way it'll be a single place where the free features are defined.

RobertJoonas (approved):
Recommended a change, but otherwise looks good to go! Thanks 🙏🏼

test/plausible/billing/quota_test.exs
aerosol: fc1d178bf but 

> What's more, I think we should remove the free option from the Feature context and move it to the Quota context instead. E.g. a module attribute like @free_features [Goals]. That way it'll be a single place where the free features are defined.

That makes sense, however I like to not push this in this PR, looks like a more involved change.

aerosol: Filed a backlog item

PR #3724, opened 2024-01-24
Add multiple imports per site

zoldar:
### Changes

This PR will be eventually have migrations extracted from it and merged first in order to safely deploy the rest of the code.

This is a significant refactor of imported stats logic to open up the possibility of a) having more than one import per site b) having imports from more than one source (beyond our current Google Analytics integration).

The most important addition is `Plausible.Imported.Importer` behaviour that defines a common interface for running all imports in the system. As all import jobs run through it, they are globally serialized to keep a better control over system load.

The `Importer` behaviour will be implemented for any future import sources.

All `imported_*` CH tables are now extended with an additional column, `import_id`, which is used for scoping imported statistics queries only to imports that are complete (not in progress or failed).

There's `Plausible.Imported.SiteImport` PG schema now, which holds information about each import with a reference to the site. Its ID is what is put in CH tables' `import_id` column.

In order to support soon-to-be legacy import, we still support transparently fetching data from `site.imported_data`, with the assumption that legacy imports have import ID of `0` (default column value in CH after migration).

The `Plausible.Imported.UniversalAnalytics` importer implementation in its current shape maintains `site.imported_data` contents to ensure import UI in its current shape is supported. Some of that logic will go away once we implemented proper import UI, replacing the existing one tightly bound with Google integration in "Integrations" site section (there's a stub built on top of this branch for this purpose in https://github.com/plausible/analytics/pull/3727).

Module docs of `Plausible.Imported` and `Plausible.Imported.Importer` might be a good starting point for review.

### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #3743, opened 2024-02-01
safe(r) Oban telemetry

ruslandoga:
### Changes

This PR makes ObanErrorReporter safer by processing the potentially fallible events in ~~an unlinked process.~~ a try/catch block.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/oban_error_reporter.ex
+++ b/lib/oban_error_reporter.ex
@@ -1,5 +1,12 @@
 defmodule ObanErrorReporter do
-  def handle_event([:oban, :job, :exception], measure, %{job: job} = meta, _) do
+  def handle_event(name, measurements, metadata, _) do
+    # handling telemetry event in an unlinked process
+    # to avoid handler detachment in the case of an error
+    # see https://hexdocs.pm/telemetry/telemetry.html#attach/4
+    Task.start(fn -> handle_event(name, measurements, metadata) end)
+  end
+
+  defp handle_event([:oban, :job, :exception], measure, %{job: job} = meta) do
     extra =
       job
       |> Map.take([:id, :args, :meta, :queue, :worker])
@@ -10,13 +17,13 @@ defmodule ObanErrorReporter do
     Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
   end
 
-  def handle_event([:oban, :notifier, :exception], _timing, meta, _) do
+  defp handle_event([:oban, :notifier, :exception], _timing, meta) do
     extra = Map.take(meta, ~w(channel payload)a)
 
     Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
   end
 
-  def handle_event([:oban, :plugin, :exception], _timing, meta, _) do
+  defp handle_event([:oban, :plugin, :exception], _timing, meta) do
     extra = Map.take(meta, ~w(plugin)a)
 
     Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
--- a/lib/oban_error_reporter.ex
+++ b/lib/oban_error_reporter.ex
@@ -1,9 +1,17 @@
 defmodule ObanErrorReporter do
+  require Logger
+
   def handle_event(name, measurements, metadata, _) do
-    # handling telemetry event in an unlinked process
+    # handling telemetry event in a try/catch block
     # to avoid handler detachment in the case of an error
     # see https://hexdocs.pm/telemetry/telemetry.html#attach/4
-    Task.start(fn -> handle_event(name, measurements, metadata) end)
+    try do
+      handle_event(name, measurements, metadata)
+    catch
+      kind, reason ->
+        message = Exception.format(kind, reason, __STACKTRACE__)
+        Logger.error(message)
+    end
   end
 
   defp handle_event([:oban, :job, :exception], measure, %{job: job} = meta) do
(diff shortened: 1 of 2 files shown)


lib/oban_error_reporter.ex
```diff
@@ -1,5 +1,12 @@
 defmodule ObanErrorReporter do
-  def handle_event([:oban, :job, :exception], measure, %{job: job} = meta, _) do
+  def handle_event(name, measurements, metadata, _) do
+    # handling telemetry event in an unlinked process
+    # to avoid handler detachment in the case of an error
+    # see https://hexdocs.pm/telemetry/telemetry.html#attach/4
```
ruslandoga, line 5: > If the function fails (raises, exits or throws) then the handler is removed and a failure event is emitted.

lib/oban_error_reporter.ex
```diff
@@ -1,5 +1,12 @@
...
+    Task.start(fn -> handle_event(name, measurements, metadata) end)
```
ruslandoga, line 6: We can also do `try/catch`

lib/oban_error_reporter.ex
aerosol: I think we've even seen that on prod, however the stacktrace was very cryptic

lib/oban_error_reporter.ex
aerosol: I think it'd be cleaner to go with try/catch since the intention is to do exactly that and not order parallel work. Maybe even log the error/exception then? 

That aside, what are we expecting to fail here exactly?

zoldar: Chiming in, because that's something @ruslandoga has touched on when we discussed imports 😄 

Now not much can failk, except _maybe_ Sentry call going belly up for some reason.

But with https://github.com/plausible/analytics/pull/3724/files#diff-353dfdb1a4143c31e52afe9b8bbda92b66b526712a91b075df869908fd86de35, there will be extra CH and PG database calls which might be more likely to fail (for instance when DB connection has a hiccup).

ruslandoga: > That aside, what are we expecting to fail here exactly?

It's very unlikely, and not exactly something I'd expect to happen, but both Repo (in `on_job_exception/1`) and Sentry calls can potentially fail.

test/workers/oban_error_reporter_test.exs
```diff
@@ -0,0 +1,33 @@
...
+        &ObanErrorReporter.handle_event/4,
+        %{}
+      )
+
+      on_exit(fn -> :ok = :telemetry.detach("oban-errors-test") end)
+    end
+
+    @tag :capture_log
+    test "doesn't detach on failure" do
+      :telemetry.execute([:oban, :job, :exception], %{}, %{job: %{}})
```
ruslandoga, line 18: Not sure why, but I had it another handler detached in the first test run.

```
19:16:19.416 [error] Handler "Elixir.OpentelemetryOban.JobHandler.job_exception" has failed and has been detached. Class=:error
Reason=:function_clause
Stacktrace=[
  {OpentelemetryOban.JobHandler, :handle_job_exception,
   [[:oban, :job, :exception], %{}, %{job: %{}}, []],
   [file: ~c"lib/opentelemetry_oban/job_handler.ex", line: 87]},
  {:telemetry, :"-execute/3-fun-0-", 4,
   [
     file: ~c"/Users/x/Developer/plausible/analytics/deps/telemetry/src/telemetry.erl",
     line: 160
   ]},
  {:lists, :foreach_1, 2, [file: ~c"lists.erl", line: 1686]},
  {ObanErrorReporterTest, :"test handle_event/4 doesn't detach on failure", 1,
   [file: ~c"test/workers/oban_error_reporter_test.exs", line: 17]},
  {ExUnit.Runner, :exec_test, 2, [file: ~c"lib/ex_unit/runner.ex", line: 472]},
  {:timer, :tc, 2, [file: ~c"timer.erl", line: 270]},
  {ExUnit.Runner, :"-spawn_test_monitor/4-fun-1-", 6,
   [file: ~c"lib/ex_unit/runner.ex", line: 394]}
]
```

Will try to reproduce.

ruslandoga: Ah, of course. The payload is invalid and the other handlers are also receiving these events from tests, and they get detached on failure as expected.

test/workers/oban_error_reporter_test.exs
```diff
@@ -0,0 +1,43 @@
...
+      on_exit(fn -> :ok = :telemetry.detach("oban-errors-test") end)
+    end
+
+    @tag :capture_log
+    test "doesn't detach on failure" do
+      :ok =
+        :telemetry.execute(
+          [:oban, :job, :exception],
+          _bad_measurements = %{},
+          _bad_metadata = %{job: :bad_job}
```
ruslandoga, line 22: A more realistic way to test it would be to pass an invalid `site_id`.

aerosol: Even more realistic would be to trigger job exception from within some worker; I recall Oban has bumped minor version whilst changing the telemetry format.

aerosol (approved):

lib/oban_error_reporter.ex
```diff
@@ -1,5 +1,20 @@
 defmodule ObanErrorReporter do
-  def handle_event([:oban, :job, :exception], measure, %{job: job} = meta, _) do
+  require Logger
+
+  def handle_event(name, measurements, metadata, _) do
```
aerosol, line 4: Just one more thing, shouldn't we have a catch-all clause for the remaining telemetry such as:

```
[:oban, :job, :start]
[:oban, :job, :stop]
```

?

aerosol: Ah nevermind, we attach the handler explicltyl

PR #3747, opened 2024-02-01
Improve Stats API upgrade required error message

RobertJoonas:
### Changes

We recently fixed a bug and took away Stats API (and other features) access from users without a trial or an active subscription. Before we made that change, users might have been using API keys of admin accounts that didn't have an active subscription. Now that we've finally blocked the access, those API keys do not have access anymore, and the current error message is pretty intimidating -

> Stats API is part of the Plausible Business plan. To get access to this feature, please upgrade your account.

Upgrading the owner account will not fix the issue if you keep using the admin key. This PR is a quick fix to improve the communication and tell users about the real problem.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugs/authorize_stats_api.ex
+++ b/lib/plausible_web/plugs/authorize_stats_api.ex
@@ -45,7 +45,7 @@ defmodule PlausibleWeb.AuthorizeStatsApiPlug do
       {:error, :upgrade_required} ->
         H.payment_required(
           conn,
-          "#{Plausible.Billing.Feature.StatsAPI.display_name()} is part of the Plausible Business plan. To get access to this feature, please upgrade your account."
+          "The account that owns this API key does not have access to Stats API. Please make sure you're using the API key of a subscriber account and that the subscription plan includes Stats API"
         )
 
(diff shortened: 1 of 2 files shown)


lib/plausible_web/plugs/authorize_stats_api.ex
```diff
@@ -45,7 +45,7 @@ defmodule PlausibleWeb.AuthorizeStatsApiPlug do
       {:error, :upgrade_required} ->
         H.payment_required(
           conn,
-          "#{Plausible.Billing.Feature.StatsAPI.display_name()} is part of the Plausible Business plan. To get access to this feature, please upgrade your account."
+          "The account that owns this API key does not have access to Stats API. Please make sure you're using the API key of a subscriber account and that the subscription plan includes Stats API"
```
zoldar, line 48:
```suggestion
          "The account that owns this API key does not have access to Stats API. Please make sure you're using the API key of the subscriber account and that the subscription plan includes Stats API"
```
?

zoldar (approved):

PR #3763, opened 2024-02-07
Include metrics for Plugins API router

aerosol:
It's a separate one, so PromEx must know about it.


aerosol:
welp https://github.com/akoutmos/prom_ex/issues/224

PR #3767, opened 2024-02-08
Merge `Plugins.API.Router` into main one

aerosol:
In order to get grafana metrics reported
See: https://github.com/akoutmos/prom_ex/issues/224


macobo (approved):

lib/plausible_web/views/error_view.ex
```diff
@@ -2,7 +2,7 @@ defmodule PlausibleWeb.ErrorView do
   use Plausible
   use PlausibleWeb, :view
 
-  def render("500.json", %{conn: %{private: %{PlausibleWeb.Plugins.API.Router => _}}}) do
+  def render("500.json", %{conn: %{assigns: %{plugins_api: true}}}) do
```
zoldar, line 5: This change will skip rendering this clause of "500.json" for /spec/* endpoints. Is that intentional?

zoldar (approved):

PR #3772, opened 2024-02-09
Add migration for minmax indexes on session timestamps

macobo:
### Changes

Related: https://github.com/plausible/analytics/pull/3770

This PR adds a minmax index for sessions_v2.timestamp column. This will make it cheaper to calculate queries like `SELECT * FROM sessions_v2 WHERE start >= ? and timestamp < ?` as clickhouse will know what parts of the data it can skip.

Operationally not sure how we can run these migrations - dear reviewer if you do know, ping me and let's run/document together!

### DIFF
--- a/lib/plausible/migration_utils.ex
+++ b/lib/plausible/migration_utils.ex
@@ -1,0 +1,12 @@
+defmodule Plausible.MigrationUtils do
+  @moduledoc """
+  Base module for to use in Clickhouse migrations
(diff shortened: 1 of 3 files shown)


ukutaht (approved):

macobo:
Found the docs: https://kb.plausible.io/engineering/deployments/migrations :raised_hands:

PR #3777, opened 2024-02-12
Avoid changing Custom properties tab automatically every time filter changes

macobo:
Follow-up to https://github.com/plausible/analytics/pull/3719

The previous behavior was predicated on:
- Allowing a single custom property filter
- Allowing breaking down only by the chosen custom property filter

Now these restrictions are removed the previous auto-switching just becomes annoying


RobertJoonas (approved):
👍

PR #3781, opened 2024-02-12
Unify percentage change for CR and bounce_rate

RobertJoonas:
### Changes

1. Calculate `conversion_rate` change by subtracting the percentages (similar to bounce rate), not taking percentages of percentages (involving Top Stats in the dashboard and the `conversion_rate` metric in Stats API aggregate with `compare=previous_period`)

    This is how the percentage change is reported on the dashboard currently:
    Pasted Graphic
    
    Since with CR we're talking about percentages from 0-100, that change should be -1.4% instead. That should be consistent with the percentage change of `bounce_rate`
    
    48 vs  49 bounce rate

2. Use the same percentage change between the dashboard API and Stats API
    
    * The Stats API percentage change for bounce rate is currently calculated by taking percentages of percentages. This PR makes it consistent with the dashboard.
    * Exactly the same for `conversion_rate` change - not live yet, but this PR intends to fix that before we publish the docs

### Tests
- [x] Automated tests have been modified

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/external_stats_controller.ex
+++ b/lib/plausible_web/controllers/api/external_stats_controller.ex
@@ -2,7 +2,7 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
   use PlausibleWeb, :controller
   use Plausible.Repo
   use PlausibleWeb.Plugs.ErrorHandler
-  alias Plausible.Stats.Query
+  alias Plausible.Stats.{Query, Compare, Comparisons}
 
   @metrics [
     :visitors,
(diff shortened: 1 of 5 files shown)


lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -44,12 +44,9 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
 
           Enum.map(curr_result, fn {metric, %{value: current_val}} ->
             %{value: prev_val} = prev_result[metric]
+            change = Compare.calculate_change(metric, prev_val, current_val)
```
zoldar, line 47: Is it guaranteed that only `:conversion_rate` and `:bounce_rate` metrics will be passed here?

RobertJoonas: Any other metrics can be passed as well, in which case the change will be considered a "percent change". It's just that `:bounce_rate` and `:conversion_rate` have separate function clauses.

zoldar: Oh damn it, I have missed the other clauses that are visible after expanding the diff 😂 . 👍  all good

zoldar (approved):

PR #3782, opened 2024-02-13
Refactor: Exclude selected prop keys in API instead of in FE

macobo:
Solves feedback from original PR: https://github.com/plausible/analytics/pull/3719#discussion_r1466583893

The FE approach would run into problems if you have >25 custom props, so excluding in the BE allows avoiding problems

### DIFF
--- a/assets/js/dashboard/stats/modals/prop-filter-row.js
+++ b/assets/js/dashboard/stats/modals/prop-filter-row.js
@@ -23,7 +23,8 @@ function PropFilterRow({
 }) {
   function fetchPropKeyOptions() {
     return (input) => {
-      return api.get(apiPath(site, "/suggestions/prop_key"), query, { q: input.trim() })
+      const exclude = selectedPropKeys.map((key) => key.value)
+      return api.get(apiPath(site, "/suggestions/prop_key"), query, { q: input.trim(), exclude: JSON.stringify(exclude) })
     }
   }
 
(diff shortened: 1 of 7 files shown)


assets/js/dashboard/stats/modals/prop-filter-row.js
```diff
@@ -23,7 +23,8 @@ function PropFilterRow({
 }) {
   function fetchPropKeyOptions() {
     return (input) => {
-      return api.get(apiPath(site, "/suggestions/prop_key"), query, { q: input.trim() })
+      const exclude = selectedPropKeys.map((key) => key.value)
+      return api.get(apiPath(site, "/suggestions/prop_key"), query, { q: input.trim(), exclude: JSON.stringify(exclude) })
```
RobertJoonas, line 27: I think instead of specifying an additional `exclude` parameter, we should use the `formState` to update the query object sent to the backend instead.

Right now, if you do the following:

0. setup: send two events on localhost:
* `mix send_pageview --props "{\"one\":\"1\"}"`
* `mix send_pageview --props "{\"two\":\"2\"}"`
* configure these props in Site Settings > Custom Properties

1. Open the prop filter modal and apply the filter "Property `one` is `1`
2. Click on the applied filter label on top of the dashboard to open the filter modal again
3. Click "+ Add another"
4. Remove the already applied one clicking the red trash icon
5. Try to select a `prop_key`

After step 5, `two` will not be returned anymore, because the suggestions request will be unaware of what is happening in the `formState`.

Now there's a small blocker here I believe - how do you update the query with the current `formState` if a `prop_key` is selected, but the `prop_value` corresponding to it is empty?

Perhaps we could solve this by disabling (or hiding) the "+ Add another" button when any of the rows is incomplete? Then, the situation where a prop_key is hanging in the formstate without a prop_value would become impossible

PR #3786, opened 2024-02-14
Experimental session count

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -138,7 +138,7 @@ defmodule Plausible.Stats.Base do
       from(
         s in "sessions_v2",
         where: s.site_id == ^site.id,
-        where: s.start >= ^first_datetime and s.start < ^last_datetime
+        where: s.timestamp >= ^first_datetime and s.start < ^last_datetime
       )
 
     on_full_build do
--- a/lib/plausible/stats/base.ex
+++ b/lib/plausible/stats/base.ex
@@ -132,14 +132,18 @@ defmodule Plausible.Stats.Base do
   }
 
   def query_sessions(site, query) do
-    {first_datetime, last_datetime} = utc_boundaries(query, site)
+    {first_datetime, last_datetime} =
+      utc_boundaries(query, site)
+
+    q = from(s in "sessions_v2", where: s.site_id == ^site.id)
 
     sessions_q =
-      from(
-        s in "sessions_v2",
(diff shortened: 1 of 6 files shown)


lib/plausible/stats/base.ex
```diff
@@ -132,14 +132,18 @@ defmodule Plausible.Stats.Base do
...
+    q = from(s in "sessions_v2", where: s.site_id == ^site.id)
 
     sessions_q =
-      from(
-        s in "sessions_v2",
-        where: s.site_id == ^site.id,
-        where: s.start >= ^first_datetime and s.start < ^last_datetime
-      )
+      if FunWithFlags.enabled?(:experimental_session_count, for: site) or
+           query.experimental_session_count? do
```
macobo, line 142: Q: do we need this in params/query? If only for tests, could we enable the flag for that specific test and cut down on param-passing?

macobo:
Problem with this flag is we have only a single test for it being on. As such I'd be really careful for turning it on for anything but testing.

Also, related RFC: https://3.basecamp.com/5308029/buckets/35611491/messages/7085680123

lib/plausible/stats/base.ex
aerosol: The intention was for the parameter to be passed down the whole pipeline so we can evaluate the solution for customers who require the fix, before enabling the feature flag

aerosol:
> Problem with this flag is we have only a single test for it being on. As such I'd be really careful for turning it on for anything but testing.

```diff
diff --git a/test/test_helper.exs b/test/test_helper.exs
index 7cbd300df..efed6e079 100644
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -2,6 +2,7 @@
 Mox.defmock(Plausible.HTTPClient.Mock, for: Plausible.HTTPClient.Interface)
 Application.ensure_all_started(:double)
 FunWithFlags.enable(:window_time_on_page)
+FunWithFlags.enable(:experimental_session_count)
 FunWithFlags.enable(:shields)
 Ecto.Adapters.SQL.Sandbox.mode(Plausible.Repo, :manual)
```

and `mix test` runs fine.

lib/plausible/stats/base.ex
aerosol: Also, feature flags refresh in 15-minute cycles, each node separately.

PR #3795, opened 2024-02-19
add csv importer

ruslandoga:
### Changes

This PR implements the CSV importer from S3.

### Tests

### Changelog

### Documentation

### Dark mode


config/.env.test
```diff
@@ -15,3 +15,11 @@ IP_GEOLOCATION_DB=test/priv/GeoLite2-City-Test.mmdb
...
 GOOGLE_CLIENT_ID=fake_client_id
 GOOGLE_CLIENT_SECRET=fake_client_secret
+
+S3_DISABLED=false
+S3_ACCESS_KEY_ID=minioadmin
+S3_SECRET_ACCESS_KEY=minioadmin
+S3_REGION=us-east-1
+S3_ENDPOINT=http://localhost:9000
+S3_IMPORTS_BUCKET=imports
+S3_HOST_FOR_CLICKHOUSE=172.19.0.4
```
ruslandoga, line 25: I'll need to figure this one out for tests

config/runtime.exs
```diff
@@ -701,3 +701,69 @@ if not is_selfhost do
...
+
+  config :ex_aws,
+    http_client: Plausible.S3.Client,
+    access_key_id: s3_env_value.("S3_ACCESS_KEY_ID"),
+    secret_access_key: s3_env_value.("S3_SECRET_ACCESS_KEY"),
+    region: s3_env_value.("S3_REGION")
+
+  %URI{scheme: s3_scheme, host: s3_host, port: s3_port} = URI.parse(s3_env_value.("S3_ENDPOINT"))
+
+  config :ex_aws, :s3,
```
ruslandoga, line 761: `ex_aws` is used just as the config store and URL builder in CSVImporter.

lib/plausible/imported/csv_importer.ex
```diff
@@ -16,10 +16,77 @@ defmodule Plausible.Imported.CSVImporter do
...
+      table = Path.rootname(filename)
+
+      s3_structure = input_structure(table)
+      s3_url = Plausible.S3.import_clickhouse_url(s3_path)
+
+      statement =
+        """
+        INSERT INTO {table:Identifier} \
+        SELECT {site_id:UInt64} AS site_id, {import_id:UInt64} AS import_id, * \
+        FROM s3({s3_url:String},{s3_access_key_id:String},{s3_secret_access_key:String},{s3_format:String},{s3_structure:String})\
```
ruslandoga, line 47: The actual downloading and importing is done by ClickHouse.

config/runtime.exs
```diff
@@ -701,3 +701,69 @@ if not is_selfhost do
...
+  s3_env =
+    Enum.map(s3_env, fn var ->
+      Map.put(var, :value, get_var_from_path_or_env(config_dir, var.name))
+    end)
+
+  s3_missing_env = Enum.filter(s3_env, &is_nil(&1.value))
+
+  unless s3_missing_env == [] do
+    raise """
+    Missing S3 configuration. Please set #{s3_missing_env |> Enum.map(& &1.name) |> Enum.join(", ")} environment variable(s):
```
ruslandoga, line 743: TODO config test

config/.env.test
ruslandoga: I'll use 172.17.0.1 to access MinIO via the Docker host from ClickHouse.

lib/plausible/imported/csv_importer.ex
```diff
@@ -16,10 +16,99 @@ defmodule Plausible.Imported.CSVImporter do
...
+    {:ok, ch} =
+      Plausible.IngestRepo.config()
+      |> Keyword.replace!(:pool_size, 1)
+      |> Ch.start_link()
+
+    ranges =
+      Enum.map(uploads, fn upload ->
+        %{"filename" => filename, "s3_path" => s3_path} = upload
+
+        ".csv" = Path.extname(filename)
```
ruslandoga, line 38: I'll revisit this after implementing imports. I think we'd do this sort of check there. If not, I'll make it into an ArgumentError to cancel the job early.

ruslandoga:
I'll need to add some documentation but the implementation itself and the tests are ready for a review :)

lib/plausible/imported/csv_importer.ex
```diff
@@ -16,10 +16,99 @@ defmodule Plausible.Imported.CSVImporter do
...
+      end)
+
+    {:ok,
+     %{
+       start_date: Enum.min_by(ranges, & &1.first, Date).first,
+       end_date: Enum.max_by(ranges, & &1.last, Date).last
+     }}
+  rescue
+    # we are cancelling on any argument or ClickHouse errors
+    e in [ArgumentError, Ch.Error] ->
```
ruslandoga, line 82: transient error would be `Mint.TransportError`

test/plausible/imported/csv_importer_test.exs
```diff
@@ -0,0 +1,391 @@
...
+  use Plausible.DataCase, async: true
+
+  doctest Plausible.Imported.CSVImporter, import: true
+
+  @moduletag :minio
+
+  # uses https://min.io
+  # docker run -d --rm -p 9000:9000 -p 9001:9001 --name minio minio/minio server /data --console-address ":9001"
+  # docker exec minio mc alias set local http://localhost:9000 minioadmin minioadmin
+  # docker exec minio mc mb local/imports
```
ruslandoga, line 11: These comments describe how to run MinIO locally for these tests. I'm using Docker for both MinIO and ClickHouse. ClickHouse accesses MinIO through Docker host (172.17.0.1).

zoldar (commented):
Looks good 👍 We'll most likely need to refinement because import date range will have to be known up front, but it can be tackled in a follow-up.

test/plausible/imported/csv_importer_test.exs
```diff
@@ -0,0 +1,391 @@
+defmodule Plausible.Imported.CSVImporterTest do
+  use Plausible.DataCase, async: true
+
+  doctest Plausible.Imported.CSVImporter, import: true
+
+  @moduletag :minio
+
+  # uses https://min.io
```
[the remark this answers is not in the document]
ruslandoga: For the tests @cnkk suggested using [testcontainers](https://github.com/testcontainers/testcontainers-elixir/tree/main?tab=readme-ov-file#exunit-tests)

But for development it definitely would be useful to update Makefile as well. I'll do it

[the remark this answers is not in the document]
ruslandoga: I've added Makefile changes in https://github.com/plausible/analytics/pull/3795/commits/eb905f3d93be8c6875704b1df77a74f68ee3bdad but testcontainers seems a bit more difficult. For some reason the previous approach of accessing MinIO via Docker host doesn't work anymore locally but keeps working in the CI.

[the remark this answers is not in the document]
ruslandoga: Since the CI tests pass, I think this can be left for later. I don't think anyone would be running `minio` tests other than me anyway :)

test/plausible/imported/csv_importer_test.exs
```diff
@@ -0,0 +1,408 @@
...
+
+      assert {:discard, message} = Plausible.Workers.ImportAnalytics.perform(job)
+      assert message =~ "CANNOT_PARSE_INPUT_ASSERTION_FAILED"
+
+      assert %SiteImport{id: import_id, source: :csv, status: :failed} =
+               Repo.get_by!(SiteImport, site_id: site.id)
+
+      # ensure no browser left behind
+      imported_browsers_q = from b in "imported_browsers", where: b.import_id == ^import_id
+      assert Plausible.ClickhouseRepo.aggregate(imported_browsers_q, :count) == 0
```
ruslandoga, line 400: Seems like the deletion is async? https://github.com/plausible/analytics/actions/runs/8031716053/job/21940336515?pr=3795

zoldar: Yes, deletions are async. Tests running purging directly use `eventually` via `await_clickhouse_count`, like here https://github.com/plausible/analytics/blob/master/test/plausible/purge_test.exs#L83-L89

ruslandoga: Would it make sense to make them sync?

zoldar: AFAICT, ansynchronicity shouldn't have any adverse effects as long as we don't use shards. See https://github.com/plausible/analytics/blob/1b4d81c11a4de05f1e65d5907f05c29d67b3dd6f/lib/plausible/purge.ex#L22-L31.

ruslandoga: Right. Sorry, I skipped these docs... 😓

According to https://clickhouse.com/docs/en/sql-reference/statements/alter#mutations

> Mutations are totally ordered by their creation order and are applied to each part in that order. Mutations are also partially ordered with INSERT INTO queries: data that was inserted into the table before the mutation was submitted will be mutated and data that was inserted after that will not be mutated. 

My concern about successful imports being mutated is invalid.

ruslandoga: I'll update the test to use `eventually`. Or maybe I can remove it completely since it's not testing CSVImporter but rather the underlying behaviour?

zoldar: No strong preferences here. We can leave it with `eventually`, like you said

ruslandoga: I've updated the test https://github.com/plausible/analytics/pull/3795/commits/ab4f9c0351b234b1dcfefcbac6aa6dfe6e509992

I think it's ready now.

mix.exs
```diff
@@ -135,7 +135,11 @@ defmodule Plausible.MixProject do
       {:esbuild, "~> 0.7", runtime: Mix.env() in [:dev, :small_dev]},
       {:tailwind, "~> 0.2.0", runtime: Mix.env() in [:dev, :small_dev]},
       {:ex_json_logger, "~> 1.4.0"},
-      {:ecto_network, "~> 1.5.0"}
+      {:ecto_network, "~> 1.5.0"},
+      {:ex_aws, "~> 2.5"},
+      {:ex_aws_s3, "~> 2.5"},
+      {:sweet_xml, "~> 0.7.4"},
```
zoldar, line 141: Is this explicit dependency declaration intentional?

test/plausible/imported/csv_importer_test.exs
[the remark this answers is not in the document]
zoldar: Hmm, at least for me, tests work fine locally with `testcontainer` based setup provided I setup CH via `make clickhouse`.

zoldar (approved):
✨ 

Left some minor stuff.

mix.exs
ruslandoga: yes, since sweet_xml is optional in ex_aws https://hex.pm/packages/ex_aws

ruslandoga: or do you mean ex_aws? it can be omitted since it's required in ex_aws_s3

zoldar: no no, it's fine

PR #3800, opened 2024-02-20
Stop writing session properties into events table

macobo:
### Changes

Part of [the project to improve sessions and events table schemas](https://3.basecamp.com/5308029/buckets/35611491/messages/7061875211).

After this PR:
- Sessions properties (which were never queried) are no longer written to events table
- In tests, one should use `session_referrer: 'Google'` when creating events and want to create session properties.

~~Note that the test change is currently lenient to avoid breaking peoples work-in-progress. I'll go and update rest of the tests down the line.~~

~~This change is best reviewed commit-by-commit as most of it is repetetive test changes.~~


test/plausible/stats/clickhouse_test.exs
```diff
@@ -228,22 +228,22 @@ defmodule Plausible.Stats.ClickhouseTest do
       populate_stats(site, [
         build(:pageview,
           pathname: "/",
-          referrer_source: "Twitter"
+          session: %{referrer_source: "Twitter"}
```
macobo, line 231: Review from chat with Uku: Make this `session_referrer_source`

macobo:
Marking as draft temporarily since work around testing harness is required.

lib/plausible/ingestion/event.ex
```diff
@@ -13,6 +13,7 @@ defmodule Plausible.Ingestion.Event do
   defstruct domain: nil,
             site: nil,
             clickhouse_event_attrs: %{},
+            clickhouse_session_attrs: %{},
```
aerosol, line 16: Wondering if, to be strict in enforcing possible keys here, we could make those both internal structs, preferably with defaults where applicable. Would make it clearer of what goes where then and we could also get rid of all the `Map.get` calls and use the dot-access syntax with confidence.

Or, even consider building and updating the respective Ecto schema structs for both tables directly here. It feels like there's a lot of attribute mangling going on - quite hard to keep track of that.

test/support/test_utils.ex
```diff
@@ -82,9 +82,8 @@ defmodule Plausible.TestUtils do
           |> Map.put(:site_id, pageview.site.id)
 
         Factory.build(:pageview, pageview)
-        |> Map.from_struct()
-        |> Map.delete(:__meta__)
         |> update_in([:timestamp], &to_naive_truncate/1)
+        |> Map.delete(:_factory_event)
```
aerosol, line 86: That doesn't look too happy. Help me understand what's going on with that field. Is that temporary?

test/support/test_utils.ex
```diff
@@ -196,14 +194,42 @@ defmodule Plausible.TestUtils do
...
+        session_id = Plausible.Session.CacheStore.on_event(event, session_params, nil)
         Map.put(sessions, {event.site_id, event.user_id}, session_id)
       end)
 
     Enum.each(events, fn event ->
-      event = Map.put(event, :session_id, sessions[{event.site_id, event.user_id}])
-      Plausible.Event.WriteBuffer.insert(event)
+      clickhouse_event =
+        %Plausible.ClickhouseEventV2{}
+        |> Map.merge(event)
```
aerosol, line 204: This is not a valid `%Plausible.ClickhouseEventV2` struct anymore. I would seriously consider doing the changes properly and have people rebase their PRs if this is the price we're paying.

aerosol (commented):
Find it very hard to reason about all the random map juggling. I understand we want make other's lives easier (are we?) but this change already touches tons of tests, so why not go with full force?

test/support/test_utils.ex
macobo: populate_events function needs to disambiguate between imported stats and events. Previously it could do so by pattern matching on %Plausible.ClickhouseEventV2 but out factory no longer returns that, instead returning a plain struct with some metadata attached to it to still be able to be matched against.

Note that this while messy doesn't leak outside factory/test_utils, but happy to change to a better option.

Is there a more preferred pattern for doing this? Another option I considered was creating a two-field struct in testUtils but it seemed like the more complicated option.

test/support/test_utils.ex
aerosol: I mean the resulting map has fields that don't belong to the struct definition. So it works kind of accidentally, because all structs are maps, but hard to predict how error prone is that.

```patch
diff --git a/test/support/test_utils.ex b/test/support/test_utils.ex
index 812a7f7ef..6c162296c 100644
--- a/test/support/test_utils.ex
+++ b/test/support/test_utils.ex
@@ -200,9 +200,12 @@ defmodule Plausible.TestUtils do
 
     Enum.each(events, fn event ->
       clickhouse_event =
-        %Plausible.ClickhouseEventV2{}
-        |> Map.merge(event)
-        |> Map.put(:session_id, sessions[{event.site_id, event.user_id}])
+        struct!(
+          Plausible.ClickhouseEventV2,
+          Map.merge(event, %{
+            session_id: sessions[{event.site_id, event.user_id}]
+          })
+        )
 
       Plausible.Event.WriteBuffer.insert(clickhouse_event)
     end)

```

This makes everything explode :/

lib/plausible/ingestion/event.ex
macobo: Let me reason this through: 

>  to be strict in enforcing possible keys here, we could make those both internal structs, preferably with defaults where applicable. 

By creating a new struct, we'd be duplicating the schema file making future migrations/changes more painful.

> Or, even consider building and updating the respective Ecto schema structs for both tables directly here. 

Isn't this what this file already is doing, in a pipeline?

> Would make it clearer of what goes where then and we could also get rid of all the Map.get calls and use the dot-access syntax with confidence.

Agreed on this - this is noisy, especially when the map crosses the module boundary and gets passed to sessions cache store. I'm lacking the proper elixir experience to see what a better solution here is.

Would you be up for pairing on this?

macobo:
> Find it very hard to reason about all the random map juggling. 

Which map juggling do you mean? Refactoring of test utils/factories, ingestion/event.ex changes or the many test changes?

On the test changes: Most of the test changes in this PR are not needed and can easily be separated out. 

On one hand I want to avoid breaking peoples PRs, which might in turn break master. On the other, I wanted to provide proper examples on how to use the factories in the 'new way' so people start using it in their newest PRs.

Once this has been in for a ~week and larger in-progress PRs are merged I was planning to go back and refactor the rest of the tests.

aerosol:
Apologies for the tone in this PR, let's pair on this and see if we can clean it up a little.

lib/plausible/ingestion/event.ex
aerosol: > Isn't this what this file already is doing, in a pipeline?

not quite, we're building the `Event` struct with arbitrary `clickhouse_event_attrs` - that only in the end becomes `ClickhouseEventV2`. I'm thinking that maybe this could be `ClickhouseEvent` from the start. And similarly, `ClickhouseSessionV2` instead of `clickhouse_session_attrs`. Currently we're registering a session, not knowing if the `ClickhouseEventV2` changeset is valid. Perhaps we can improve that and drop the event early?

aerosol: Ah I'm talking nonsense, there is a event validation step prior to session registration. So I guess we should do similar for ClickhouseSessionV2.

aerosol (approved):
Let's merge and I'll submit testing-related refactoring on top

PR #3802, opened 2024-02-20
Add data migration for moving sessions_v2 table to VersionedCollapsingMergeTree

macobo:
This PR adds a data migration migration move sessions_v2 table to VersionedCollapsingMergeTree. This will allow us to avoid rare data corruption issues around sessions.

Part of [the project to improve sessions and events table schemas](https://3.basecamp.com/5308029/buckets/35611491/messages/7061875211).

The migration will swap out the engine for a new one and leave behind a backup table that will need to be cleaned up manually. We don't need to worry about replication or data ingestion - the queries at work here were instant, which will mean very minimal data loss (<1s)

This has been run both locally and on staging. Once successfully run on production, will add a normal migration to do this on all environments.

## How to verify data

This query can help compare the main and temporary tables against each other.

```sql
SELECT main._partition_id, tmp.count, main.count
FROM (
  SELECT _partition_id, count() AS count
  FROM sessions_v2_tmp_versioned
  GROUP BY _partition_id
) AS tmp
FULL OUTER JOIN (
  SELECT _partition_id, count() AS count
  FROM sessions_v2
  GROUP BY _partition_id
) AS main
ON (tmp._partition_id == main._partition_id)
ORDER BY main._partition_id
```

### DIFF
--- a/lib/plausible/data_migration/versioned_sessions.ex
+++ b/lib/plausible/data_migration/versioned_sessions.ex
@@ -6,6 +6,10 @@ defmodule Plausible.DataMigration.VersionedSessions do
   use Plausible.DataMigration, dir: "VersionedSessions", repo: Plausible.IngestRepo
 
   @suffix_format "{YYYY}{0M}{0D}{h24}{m}{s}"
+  @versioned_table_engines [
+    "ReplicatedVersionedCollapsingMergeTree",
+    "VersionedCollapsingMergeTree"
+  ]
 
   def run(opts \\ []) do
     run_exchange? = Keyword.get(opts, :run_exchange?, true)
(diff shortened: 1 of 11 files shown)


lib/plausible/data_migration/versioned_sessions.ex
```diff
@@ -0,0 +1,79 @@
...
+    "ReplicatedVersionedCollapsingMergeTree",
+    "VersionedCollapsingMergeTree"
+  ]
+
+  def run(opts \\ []) do
+    run_exchange? = Keyword.get(opts, :run_exchange?, true)
+
+    unique_suffix = Timex.now() |> Timex.format!(@suffix_format)
+
+    cluster? =
```
ukutaht, line 19: Generalize this so we don't have to write it for a third or fourth time?

ukutaht (approved):

PR #3803, opened 2024-02-20
Migration to improve compression options of sessions_v2 and events_v2 tables

macobo:
Part of https://3.basecamp.com/5308029/buckets/35611491/messages/7061875211 

Based on work in https://3.basecamp.com/5308029/buckets/35611491/messages/6949640880#__recording_7015576045

Note that the improved compression options only apply for new incoming data - to get the full benefits, running an `OPTIMIZE TABLE` afterwards is needed (but expensive computationally)

After merging this, planning to:
1. Do a before-and-after of compression ratios
2. Run OPTIMIZE TABLE for both tables during a quiet period

### DIFF
--- a/priv/ingest_repo/migrations/20240220123656_create_sessions_events_compression_options.exs
+++ b/priv/ingest_repo/migrations/20240220123656_create_sessions_events_compression_options.exs
@@ -1,0 +1,45 @@
+defmodule Plausible.IngestRepo.Migrations.CreateSessionsEventsCompressionOptions do
+  @moduledoc """
+  Improves compression options for events_v2 and sessions_v2 tables.
+
+  It's suggested to run OPTIMIZE both tables after this to get full compression wins afterwards
+  """
+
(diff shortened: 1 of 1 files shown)


cnkk (approved):

priv/ingest_repo/migrations/20240220123656_create_sessions_events_compression_options.exs
```diff
@@ -0,0 +1,45 @@
+defmodule Plausible.IngestRepo.Migrations.CreateSessionsEventsCompressionOptions do
+  @moduledoc """
+  Improves compression options for events_v2 and sessions_v2 tables.
+
+  It's suggested to run OPTIMIZE both tables after this to get full compression wins afterwards
```
ukutaht, line 5: Can we wrap this in an ecto migration like https://github.com/plausible/analytics/pull/3809 for self-hosters as well?

PR #3805, opened 2024-02-20
Reapply conversion rate into Stats API + bugfixes

RobertJoonas:
### Changes

This PR reapplies two reverted merge commits from https://github.com/plausible/analytics/pull/3739 and https://github.com/plausible/analytics/pull/3781 which already found their way to prod for half a day.

The reason I decided to revert those was that the CSV export started failing more often with a Sentry error we've seen before, and I decided to investigate that. Also, due to the uncertainty with the conversion rate bug in general, it was a safe play to keep it out of the public API at that point.

The way we are currently adding the CR into the browser version breakdown is broken at the moment - it doesn't crash in most cases, but it doesn't return correct conversion rate either. The recent deployment just made it crash.

The fix introduced in this PR is not pretty, and in fact, although it's much less likely, there are still ways for it to fail. If that happens though, instead of a crashing, we'll be giving back both `conversion_rate` and `total_visitors` as `"N/A"`, and reporting a message to Sentry with additional debugging details.

Longer term fix to this problem involves rebuilding the database query construction logic and asking CR from ClickHouse queries directly. That's out of scope of this PR.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -3,9 +3,9 @@ defmodule Plausible.Stats.Breakdown do
   use Plausible
   use Plausible.Stats.Fragments
 
-  import Plausible.Stats.{Base, Imported, Util}
+  import Plausible.Stats.{Base, Imported}
   require OpenTelemetry.Tracer, as: Tracer
-  alias Plausible.Stats.Query
+  alias Plausible.Stats.{Query, Util}
 
   @no_ref "Direct / None"
   @not_set "(not set)"
@@ -16,7 +16,12 @@ defmodule Plausible.Stats.Breakdown do
 
   @event_metrics [:visitors, :pageviews, :events] ++ @revenue_metrics
 
-  @event_props Plausible.Stats.Props.event_props()
+  # These metrics can be asked from the `breakdown/5` function,
+  # but they are different from regular metrics such as `visitors`,
+  # or `bounce_rate` - we cannot currently "select them" directly in
+  # the db queries. Instead, we need to artificially append them to
+  # the breakdown results later on.
+  @computed_metrics [:conversion_rate, :total_visitors]
 
   def breakdown(site, query, property, metrics, pagination, opts \\ [])
 
@@ -29,15 +34,22 @@ defmodule Plausible.Stats.Breakdown do
 
     if !Keyword.get(opts, :skip_tracing), do: trace(query, property, metrics)
 
+    {revenue_goals, metrics} =
+      if full_build?() && Plausible.Billing.Feature.RevenueGoals.enabled?(site) do
+        revenue_goals = Enum.filter(event_goals, &Plausible.Goal.Revenue.revenue?/1)
+        metrics = if Enum.empty?(revenue_goals), do: metrics -- @revenue_metrics, else: metrics
+
+        {revenue_goals, metrics}
+      else
+        {nil, metrics -- @revenue_metrics}
+      end
+
+    metrics_to_select = Util.maybe_add_visitors_metric(metrics) -- @computed_metrics
+
     event_results =
       if Enum.any?(event_goals) do
-        revenue_goals =
-          on_full_build do
-            Enum.filter(event_goals, &Plausible.Goal.Revenue.revenue?/1)
-          end
-
         site
-        |> breakdown(event_query, "event:name", metrics, pagination, skip_tracing: true)
+        |> breakdown(event_query, "event:name", metrics_to_select, pagination, skip_tracing: true)
         |> transform_keys(%{name: :goal})
         |> cast_revenue_metrics_to_money(revenue_goals)
       else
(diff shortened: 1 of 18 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -693,24 +693,61 @@ defmodule Plausible.Stats.Breakdown do
     # Here, we're always only interested in the first page of results
     # - the :member filter makes sure that the results always match with
     # the items in the given breakdown_results list
-    pagination = {length(items), 1}
+    page = 1
+
+    # For browser_versions we need to fetch a lot more entries than the
```
ukutaht, line 698: 🤮

Sorry, not a reflection of your work, I'm just sad about how fundamentally fucked this whole thing is. This must have been hard to debug. For now I agree we should move forward with this workaround.

lib/plausible/stats/breakdown.ex
```diff
@@ -693,24 +693,61 @@ defmodule Plausible.Stats.Breakdown do
...
+    # pagination limit. This is because many browsers can correspond to
+    # a single version number and we need to make sure that the results
+    # without goal filter will include all combinations of browsers and
+    # their versions that were present in the `breakdown_results` array.
+    {pagination_limit, find_match_fn} =
+      if property_atom == :browser_version do
+        pagination_limit = min(elem(pagination, 0) * 10, 3_000)
+
+        find_match_fn = fn total, conversion ->
+          total[:browser_version] == conversion[:browser_version] &&
```
ukutaht, line 708: Do we need a special case for OS versions, too?

lib/plausible/stats/breakdown.ex
```diff
@@ -693,24 +693,61 @@ defmodule Plausible.Stats.Breakdown do
...
-        Enum.find(res_without_goal, fn s ->
-          Map.fetch!(s, property_atom) == Map.fetch!(item, property_atom)
-        end)
+      without_goal = Enum.find(res_without_goal, &find_match_fn.(&1, item))
+
+      {conversion_rate, total_visitors} =
+        if without_goal do
+          {Util.calculate_cr(without_goal.visitors, item.visitors), without_goal.visitors}
+        else
+          Sentry.capture_message(
```
ukutaht, line 731: Nice!

ukutaht (commented):
Not approving atm because I'm unsure whether OS versions need the same treatment?

> Longer term fix to this problem involves rebuilding the database query construction logic and asking CR from ClickHouse queries directly. That's out of scope of this PR.

This makes sense to me 👍 I've thought about this before and it should be possible to do it all in one query since Clickouse is scanning the whole column anyways. Something like:

```sql
SELECT 
  countIf(name = 'Signup') AS conversions, 
  count(*) AS visitors ,
  conversions / visitors AS conversion_rate
FROM 
  events_v2 AS e
GROUP BY
  e.browser, e.browser_version
```

Not sure if you were thinking along the same lines. This would require substantial rewriting of core query building logic. But I think it's worth it because it solves pagination issues PLUS it allows in the future to allow ordering their reports by conversion_rate which is a must-have (see https://feedback.plausible.io/206). Also avoids an unnecessary extra query.

lib/plausible/stats/breakdown.ex
```diff
@@ -693,24 +693,61 @@ defmodule Plausible.Stats.Breakdown do
...
+            "Unable to find a conversion_rate divisor from a breakdown response",
+            extra: %{
+              domain: site.domain,
+              query: query,
+              property: property,
+              pagination: pagination,
+              item_not_found: item
```
zoldar, line 738: I think it's better to wrap each of those passed values in `inspect(...)` as some might blow up when attempting to convert them to string (tuples for instance). The inspect form also preserves more information, usually.

lib/plausible/stats/breakdown.ex
RobertJoonas: Totally. There was a lot of frustration involved in building this 😀... Can't wait to get rid of `add_cr/5` 🤞

lib/plausible/stats/breakdown.ex
RobertJoonas: OS versions are not treated the same way. Yet... If you look at the [do_group_by function clause for the `os_version` property](https://github.com/plausible/analytics/blob/23b9032148b69ae27544ebd43223a6fbfdf9441e/lib/plausible/stats/breakdown.ex#L588), unlike two function clauses below for browser versions, there's no os in the group by clause.

Currently, only versions (e.g. 14, 11, 20.04) are returned from the breakdown, and many OS's can fall under the same entry. So in this PR, there's no special handling required for OS versions.

However, I have another [PR](https://github.com/plausible/analytics/pull/3806) open for adding os in the os version breakdown which will add a special case in the `add_cr` function too.

zoldar (approved):

PR #3809, opened 2024-02-22
Ecto migration to move sessions_v2 to VersionedCollapsingMergeTree

macobo:
This migration will noop in staging/production as it already has been run. It also leaves behind a backup table that initially takes no extra space but will need to be cleaned up manually.

@ruslandoga Mind if I hand off handling the self-hosted side of this to you from here?

### DIFF
--- a/priv/ingest_repo/migrations/20240222082911_sessions_v2_versioned_collapsing_merge_tree.exs
+++ b/priv/ingest_repo/migrations/20240222082911_sessions_v2_versioned_collapsing_merge_tree.exs
@@ -1,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.SessionsV2VersionedCollapsingMergeTree do
+  use Ecto.Migration
+
+  def up do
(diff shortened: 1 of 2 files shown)


ruslandoga:
@macobo sure!

ruslandoga (approved):

priv/ingest_repo/migrations/20240222082911_sessions_v2_versioned_collapsing_merge_tree.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.SessionsV2VersionedCollapsingMergeTree do
+  use Ecto.Migration
+
+  def up do
+    Plausible.DataMigration.VersionedSessions.run(run_exchange?: true)
```
ruslandoga, line 5: Worked perfectly on my instance. I'll add the notes about `sessions_v2_tmp_versioned` and `OPTIMIZE TABLE` for `events_v2` and `sessions_v2` in the next release candidate. Thank you!

PR #3811, opened 2024-02-22
Dockerfile and co updates

ruslandoga:
In place of #2986


macobo:
Please revert https://github.com/plausible/analytics/pull/3873 once this is merged - this helps fix a bug with timezone handling that PR works around in an ugly way.

cnkk (approved):

PR #3828, opened 2024-02-26
Shield: Country Rules

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/shields/country_rules.ex
```diff
@@ -0,0 +1,244 @@
...
+          <.form
+            :let={f}
+            for={@form}
+            phx-submit="save-country-rule"
+            phx-target={@myself}
+            class="max-w-md w-full mx-auto bg-white dark:bg-gray-800"
+          >
+            <h2 class="text-xl font-black dark:text-gray-100 mb-8">Add Country to Block List</h2>
+
+            <.live_component
```
zoldar, line 77: It's not a huge deal, but when suggestions appear after typing a couple characters and I reflexively hit Escape, the suggestions are closed and are not shown even if I start typing again - until I either press Down arrow with input in focus or reopen the modal. This might be desirable behavior when, for instance, adding page view goals which might be different from suggestions (?) but here, one has to pick one of the selections to add anything - there's no falling back on custom input. We could revisit this in a follow-up adding an option to only clear input without setting closed state on suggestions list perhaps?

lib/plausible_web/live/shields/country_rules.ex
```diff
@@ -0,0 +1,244 @@
...
+        socket =
+          socket
+          |> Modal.close("country-rule-form-modal")
+          |> assign(
+            form: new_form(),
+            country_rules: country_rules,
+            country_rules_count: socket.assigns.country_rules_count + 1
+          )
+
+        send_update(PlausibleWeb.Live.Components.ComboBox,
```
zoldar, line 193: Perhaps we could add  a short clarifying comment for the purpose of this change (resetting input state on subsequent modal reopenings)? 😅

priv/repo/migrations/20240221122626_shield_country_rules.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Plausible.Repo.Migrations.ShieldCountryRules do
+  use Ecto.Migration
+
+  def change do
```
zoldar, line 4: Just a note that we need to extract this migration and release it separately before the rest.

lib/plausible/shields.ex
```diff
@@ -5,38 +5,87 @@ defmodule Plausible.Shields do
...
       result =
-        if count_ip_rules(site_id) >= @maximum_ip_rules do
+        if count(schema, site_id) >= max do
           changeset =
-            %Shield.IPRule{}
-            |> Shield.IPRule.changeset(Map.put(params, "site_id", site_id))
-            |> Ecto.Changeset.add_error(:inet, "maximum reached")
+            schema
+            |> struct()
+            |> schema.changeset(Map.put(params, "site_id", site_id))
```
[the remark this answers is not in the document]
aerosol: How about 734906018 ?

lib/plausible_web/live/shields/country_rules.ex
aerosol: 48636617f

lib/plausible/shields.ex
```diff
@@ -5,38 +5,87 @@ defmodule Plausible.Shields do
...
   end
 
-  @spec add_ip_rule(Plausible.Site.t() | non_neg_integer(), map()) ::
-          {:ok, Shield.IPRule.t()} | {:error, Ecto.Changeset.t()}
-  def add_ip_rule(site_id, params) when is_integer(site_id) do
+  defp add(schema, site_id, {field, max}, params) when is_integer(site_id) do
     Repo.transaction(fn ->
       result =
-        if count_ip_rules(site_id) >= @maximum_ip_rules do
+        if count(schema, site_id) >= max do
```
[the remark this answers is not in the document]
aerosol: How else would you enforce the exact limit other than a constraint check on the db?

lib/plausible/shields.ex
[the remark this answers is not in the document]
zoldar: Given we never pass `site_id` as a raw string (from URL params for instance) - I think it's fine 👍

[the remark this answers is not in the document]
aerosol: Yeah at this point we trust the site_id is authorized on mount

lib/plausible_web/live/shields/country_rules.ex
aerosol: 8f56ecbf1 should fix it 😅

lib/plausible/shields.ex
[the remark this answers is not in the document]
aerosol: > very fringe cases

yeah, I thought it's ok-effort, not the end of the world if we register 31 rather than 30. serialized queries go up to max only.

lib/plausible_web/live/shields/country_rules.ex
```diff
@@ -0,0 +1,244 @@
...
+    </section>
+    """
+  end
+
+  def handle_event("save-country-rule", %{"country_rule" => params}, socket) do
+    user = socket.assigns.current_user
+
+    case Shields.add_country_rule(
+           socket.assigns.site.id,
+           Map.put(params, "added_by", "#{user.name} <#{user.email}>")
```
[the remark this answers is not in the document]
aerosol: I dislike the params patching for internal data, fully agree. Let me see what I can do. And yeah the idea was to set some historically immutable trace. This is far from proper auditing though, and anyone can  delete rules still...

lib/plausible_web/live/shields/country_rules.ex
zoldar: Yep, works well 👍

lib/plausible_web/live/shields/country_rules.ex
```diff
@@ -0,0 +1,245 @@
...
+            <tbody>
+              <%= for rule <- @country_rules do %>
+                <% country = Location.Country.get_country(rule.country_code) %>
+                <tr class="text-gray-900 dark:text-gray-100">
+                  <td class="px-6 py-4 text-sm font-medium">
+                    <div class="flex items-center">
+                      <span
+                        id={"country-#{rule.id}"}
+                        class="mr-4 cursor-help border-b border-dotted border-gray-400"
+                        title={"Added at #{rule.updated_at} by #{rule.added_by}"}
```
macobo, line 138: This seems to be an UTC timestamp, not the users local time, will this cause confusion?

lib/plausible_web/live/shields/country_rules.ex
```diff
@@ -0,0 +1,245 @@
...
+              + Add Country
+            </PlausibleWeb.Components.Generic.button>
+          </div>
+          <PlausibleWeb.Components.Generic.notice
+            :if={@country_rules_count >= Shields.maximum_country_rules()}
+            class="mt-4"
+            title="Maximum number of countries reached"
+          >
+            <p>
+              You've reached the maximum number of countries you can block. Please remove one before adding another.
```
macobo, line 62: Let's include the max number in this message.

test/plausible_web/live/shields/countries_test.exs
```diff
@@ -0,0 +1,127 @@
...
+      assert resp =~ "Estonia"
+
+      assert element_exists?(
+               resp,
+               ~s/button[phx-click="remove-country-rule"][phx-value-rule-id="#{r1.id}"]#remove-country-rule-#{r1.id}/
+             )
+
+      assert element_exists?(
+               resp,
+               ~s/button[phx-click="remove-country-rule"][phx-value-rule-id="#{r2.id}"]#remove-country-rule-#{r2.id}/
```
macobo, line 40: This is clunky. Idea: Should we add a custom attribute to this element, e.g. `data-test-remove-rule="123"`? IMO this would make tests much easier to write and follow and would allow for separation of concerns.

lib/plausible/stats/breakdown.ex
```diff
@@ -34,14 +34,20 @@ defmodule Plausible.Stats.Breakdown do
 
     if !Keyword.get(opts, :skip_tracing), do: trace(query, property, metrics)
 
+    no_revenue = {nil, metrics -- @revenue_metrics}
+
     {revenue_goals, metrics} =
-      if full_build?() && Plausible.Billing.Feature.RevenueGoals.enabled?(site) do
-        revenue_goals = Enum.filter(event_goals, &Plausible.Goal.Revenue.revenue?/1)
-        metrics = if Enum.empty?(revenue_goals), do: metrics -- @revenue_metrics, else: metrics
+      on_full_build do
```
macobo, line 40: Unrelated refactoring?

lib/plausible/ingestion/event.ex
```diff
@@ -226,6 +228,22 @@ defmodule Plausible.Ingestion.Event do
...
+  defp drop_shield_rule_country(
+         %__MODULE__{domain: domain, clickhouse_event_attrs: %{country_code: cc}} = event
+       )
+       when is_binary(domain) and is_binary(cc) do
+    case Plausible.Shield.CountryRuleCache.get({domain, String.upcase(cc)}) do
+      %Plausible.Shield.CountryRule{action: :deny} ->
+        drop(event, :site_country_blocklist)
+
+      _ ->
+        :lookup_failed
```
macobo, line 240: Q: Why is this symbol here?

lib/plausible/shield/country_rule_cache.ex
```diff
@@ -0,0 +1,65 @@
...
+      inner_join: s in assoc(rule, :site),
+      select: {
+        s.domain,
+        s.domain_changed_from,
+        %{struct(rule, ^@cached_schema_fields) | from_cache?: true}
+      }
+  end
+
+  @impl true
+  def get_from_source({domain, country_code}) do
```
macobo, line 41: Q: Does this mean we load all country rules one-by-one as the process boots? If so, why not load all of them at once - avoids extra load during the boot process and this list _needs_ to fit into memory anyways.

lib/plausible/stats/breakdown.ex
aerosol: https://github.com/plausible/analytics/commit/d74b1d5e6044a2fbdac119353c8a7bb9549f6374 introduced compilation warnings on small build that for some reason didn't flip the CI

lib/plausible/ingestion/event.ex
aerosol: lol, thanks for spotting it. some remnant. will :fire:

lib/plausible/shield/country_rule_cache.ex
aerosol: >   When Cache is disabled via application env, the `get/1` function
  falls back to pure database lookups (implemented via `get_from_source/1` callback. 
  This should help with introducing cached lookups in existing code, 
  so that no existing tests should break.


Basically it's a test fallback/pass-through, so you don't have to set up cache in unrelated tests.

macobo: Answering myself after exploring the code more: Nope. We seem to load them all at once.

lib/plausible_web/live/shields/country_rules.ex
aerosol: Nice, I'll use site's timezone.

lib/plausible_web/live/shields/country_rules.ex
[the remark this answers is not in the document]
aerosol: b83497e17

lib/plausible_web/live/shields/country_rules.ex
aerosol: Oh wait, this only renders a date

aerosol: 316817540

test/plausible_web/live/shields/countries_test.exs
aerosol: Only for the sake of test selectors? I think using `phx-` attributes is meaningful because it implies LV picks them up.

lib/plausible/ingestion/event.ex
aerosol: ef5b20e24

lib/plausible_web/live/shields/country_rules.ex
aerosol: 1d9537580

test/plausible_web/live/shields/countries_test.exs
aerosol: adeb5f4a7

macobo (approved):
LGTM.

Something to consider as a follow-up: It might be worth adding the VPN nation as a separate entity on top of the countries dropdown.

PR #3830, opened 2024-02-26
Query imported data for views_per_visit

RobertJoonas:
### Changes

Currently, the `views_per_visit` metric is only queried from native stats. This limitation is arbitrary, since imported data has  both `visits` and `pageviews` needed to calculate the metric.

The motivation behind doing it now is that we could make `with_imported=true` by default in Stats API. Otherwise, we'd need to exclude imported data whenever this metric is queried, or worse, only exclude imported data for this metric without a notice.

On the dashboard we have a tooltip, which is not ideal either as on mobile you wouldn't see it:
image

This PR removes that arbitrary limitation and starts including imported data in the calculation of `views_per_visit`.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -424,6 +424,16 @@ defmodule Plausible.Stats.Imported do
     |> select_imported_metrics(rest)
   end
 
+  defp select_imported_metrics(q, [:views_per_visit | rest]) do
+    q
+    |> where([i], i.pageviews > 0)
+    |> select_merge([i], %{
+      pageviews: sum(i.pageviews),
+      __internal_visits: sum(i.visits)
+    })
+    |> select_imported_metrics(rest)
+  end
+
   defp select_imported_metrics(q, [_ | rest]) do
     q
     |> select_imported_metrics(rest)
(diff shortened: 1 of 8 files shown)


lib/plausible/stats/imported.ex
```diff
@@ -424,6 +424,16 @@ defmodule Plausible.Stats.Imported do
     |> select_imported_metrics(rest)
   end
 
+  defp select_imported_metrics(q, [:views_per_visit | rest]) do
+    q
+    |> where([i], i.pageviews > 0)
```
RobertJoonas, line 429: @aerosol wanted to get your eyes on this one. I need to select pageviews here in order to calculate `views_per_visit` later in `select_joined_metrics/2`. I noticed that we're adding a `where` clause for `i.pageviews > 0` when selecting pageviews, so my logical assumption was that we should do it here too.

As I've understood, it's supposed to prevent crashes on broken GA data and ignoring the rows that are invalid (i.e. pageviews is 0 but visitors is >0).

My question is, why do we do that when selecting metrics - why not in the `merge_imported` function, before selecting the metrics. We already have a visitors > 0 condition there.

aerosol: Pardon me if I'm talking nonsense - I think not every query built with this module will have `pageviews` column available, since the "select" is built dynamically. 

Only if you select specific metrics, a sum of pageviews will be calculated. Thus, if you apply `where: i.pageviews > 0` to the `imported_q` in `merge_imported/5` (is that what you meant?), you'll likely confuse postgres. I mean I checked out your branch, applied the change I think you were suggesting and ran tests, it's all red.

You can see the original test at: https://github.com/plausible/analytics/pull/3655/files

RobertJoonas: > I think not every query built with this module will have pageviews column available

Oh yes, that's it. Thanks!

PR #3836, opened 2024-02-27
CSV exports (no UI)

ruslandoga:
### Changes

This PR adds CSV exports logic without any web or UI changes.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/exports.ex
```diff
@@ -0,0 +1,418 @@
...
+  import Ecto.Query
+
+  # TODO time_on_page
+  # TODO visits spanning multiple days deduplication
+  # TODO sampling
+  # TODO export_visitors_q only can use just sessions_v2?
+
+  # @sample to_string(2_000_000)
+
+  def export_import do
```
ruslandoga, line 15: I'm using this function just for development. Eventually I'll make it into a test.

lib/plausible/exports.ex
```diff
@@ -0,0 +1,418 @@
+defmodule Plausible.Exports do
+  @moduledoc """
+  Contains functions to export data for events and sessions as Zip archives.
+  """
+
+  import Ecto.Query
+
+  # TODO time_on_page
```
ruslandoga, line 8: This is the last metrics that needs work.

lib/plausible/exports.ex
ruslandoga: No tests for now.

lib/workers/export_csv.ex
```diff
@@ -0,0 +1,57 @@
+defmodule Plausible.Workers.ExportCSV do
```
ruslandoga, line 1: I'm going to open a new PR which would add a UI action on the exports button that starts this job.

mix.lock
```diff
@@ -11,7 +11,7 @@
...
-  "ch": {:hex, :ch, "0.2.4", "d510fbb5542d009f7c5b00bb1ecab73307b6066d9fb9b220600257d462cba67f", [:mix], [{:db_connection, "~> 2.0", [hex: :db_connection, repo: "hexpm", optional: false]}, {:decimal, "~> 2.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:ecto, "~> 3.5", [hex: :ecto, repo: "hexpm", optional: true]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}, {:mint, "~> 1.0", [hex: :mint, repo: "hexpm", optional: false]}], "hexpm", "8f065d15aaf912ae8da56c9ca5298fb2d1a09108d006de589bcf8c2b39a7e2bb"},
+  "ch": {:hex, :ch, "0.2.5", "b8d70689951bd14c8c8791dc72cdc957ba489ceae723e79cf1a91d95b6b855ae", [:mix], [{:db_connection, "~> 2.0", [hex: :db_connection, repo: "hexpm", optional: false]}, {:decimal, "~> 2.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:ecto, "~> 3.5", [hex: :ecto, repo: "hexpm", optional: true]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}, {:mint, "~> 1.0", [hex: :mint, repo: "hexpm", optional: false]}], "hexpm", "97de104c8f513a23c6d673da37741f68ae743f6cdb654b96a728d382e2fba4de"},
```
ruslandoga, line 14: This version improves streams in Ch.

lib/plausible/exports.ex
```diff
@@ -0,0 +1,345 @@
...
+      select: [
+        date(e.timestamp),
+        selected_as(fragment("any(?)", e.hostname), :hostname),
+        selected_as(e.pathname, :page),
+        visitors(e),
+        selected_as(
+          fragment("toUInt64(round(countIf(?='pageview')*any(_sample_factor)))", e.name),
+          :pageviews
+        ),
+        # NOTE: are exits pageviews or any events?
```
zoldar, line 186: https://github.com/plausible/analytics/pull/3870 might be relevant here. Currently, exit_page is set on all events, not only pageviews but that seems to be incorrect and leads to wrong calculation of other derivative metrics under some circumstances. We'll most likely switch to recording exit_pages only on events that are strictly pageviews.

lib/workers/export_csv.ex
ruslandoga: https://github.com/plausible/analytics/pull/3875

lib/plausible/exports.ex
```diff
@@ -0,0 +1,345 @@
...
+        # NOTE: no smapling is used right now
+        # hints: ["SAMPLE", unsafe_fragment(^@sample)],
+        where: s.site_id == ^site_id,
+        group_by: selected_as(:date),
+        select: %{
+          date: date(s.start),
+          bounces: bounces(s),
+          visits: visits(s),
+          visit_duration: visit_duration(s)
+          # NOTE: can we use just sessions_v2 table in this query? sum(pageviews) and visitors(s)?
```
zoldar, line 84: While investigating issue that led to https://github.com/plausible/analytics/pull/3870, we have found cases where `session.pageviews` is inconsistent with actual related events - session had `0` for for `pageviews` but there was a `pageview` event associated with it, along with numerous other custom events. That might have been some one-off issue though as session updates don't do anything fancy here https://github.com/plausible/analytics/blob/83a46fb47e96f51f058ad7d3315b0d218c331c6c/lib/plausible/session/cache_store.ex#L53-L55. It might have been caused by cached session not being in sync with what's stored in the DB at the time of the update but can't tell for sure yet.

lib/plausible/exports.ex
```diff
@@ -0,0 +1,345 @@
...
+      )
+    end
+  end
+
+  @spec export_visitors_q(pos_integer) :: Ecto.Query.t()
+  def export_visitors_q(site_id) do
+    visitors_sessions_q =
+      from s in "sessions_v2",
+        # NOTE: no smapling is used right now
+        # hints: ["SAMPLE", unsafe_fragment(^@sample)],
```
zoldar, line 76: Is sampling skipped because of visit duration calculation or just for better accuracy of exports in general?

lib/plausible/exports.ex
```diff
@@ -0,0 +1,345 @@
...
+      ]
+  end
+
+  @spec export_pages_q(pos_integer) :: Ecto.Query.t()
+  def export_pages_q(site_id) do
+    window_q =
+      from e in "events_v2",
+        # NOTE: no smapling is used right now
+        # hints: ["SAMPLE", unsafe_fragment(^@sample)],
+        where: e.site_id == ^site_id,
```
zoldar, line 158: I'm wondering it that shouldn't be limited to pageview events 🤔 If there's a custom event setup that for instance records custom events like "scrolled to the bottom", they might skew the time on page metric calculated like this. That's probably something that should be considered in the actual calculation logic at https://github.com/plausible/analytics/blob/83a46fb47e96f51f058ad7d3315b0d218c331c6c/lib/plausible/stats/breakdown.ex#L340 first. It's also possible that I'm reading it completely wrong 😅

lib/workers/export_csv.ex
```diff
@@ -0,0 +1,57 @@
...
+        ch,
+        fn conn ->
+          conn
+          |> Plausible.Exports.stream_archive(
+            Plausible.Exports.export_queries(site_id, extname: ".csv"),
+            format: "CSVWithNames"
+          )
+          |> Plausible.S3.export_upload_multipart(s3_path)
+        end,
+        timeout: :infinity
```
zoldar, line 37: Eventually, we will probably want _some_ upper time limit on this 😅

lib/plausible/s3.ex
```diff
@@ -3,6 +3,64 @@ defmodule Plausible.S3 do
...
+  @spec export_upload_multipart(Enumerable.t(), Path.t()) :: :uri_string.uri_string()
+  def export_upload_multipart(stream, s3_path) do
+    config = ExAws.Config.new(:s3)
+    bucket = config(:exports_bucket)
+
+    stream
+    # 5 MiB is the smallest chunk size AWS S3 supports
+    |> chunk_into_parts(5 * 1024 * 1024)
+    |> ExAws.S3.upload(bucket, s3_path,
+      content_disposition: ~s|attachment; filename="Plausible.zip"|,
```
zoldar, line 26: Eventually, it probably would be good to make the downloaded filename a bit more unique - we could at least embed current timestamp in it (ideally, a sanitized domain name the export is for would be nice too). But that's something to flesh out later on.

zoldar (approved):
Any chance to add at least one general integration test for the import routine? There was a helper function for running import/export cycle which could potentially be turned into one, I recall? Other than this, IMHO we could go ahead with this and work on further refinements in follow-ups as it's going to be hidden either way.

lib/plausible/exports.ex
ruslandoga: At first it was to get better accuracy... I re-enable it for full builds.

ruslandoga: https://github.com/plausible/analytics/pull/3836/commits/565d4a95a4ff9cfc30171937fcc076558f96ba99

ruslandoga:
I'm going to add some tests next.

test/plausible/imported/csv_importer_test.exs
```diff
@@ -401,6 +397,157 @@ defmodule Plausible.Imported.CSVImporterTest do
     end
   end
 
+  describe "export -> import" do
```
ruslandoga, line 400: I've added a basic e2e test here and I'll add more later :)

test/plausible/exports_test.exs
```diff
@@ -0,0 +1,154 @@
...
+    end
+  end
+
+  describe "stream_archive/3" do
+    setup do
+      ch_config = Keyword.replace!(Plausible.ClickhouseRepo.config(), :pool_size, 1)
+
+      tmp_dir =
+        Path.join(
+          System.tmp_dir!(),
```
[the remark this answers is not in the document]
ruslandoga: Done in https://github.com/plausible/analytics/pull/3836/commits/f351b24b1c4b63b4cbc067bd24169c9ea740263a

I also removed cleanups on exit to make tests clearer. ExUnit doesn't do cleanups by itself, but this tmp dir is local to the project so it's easy to delete if it becomes a problem for the developer.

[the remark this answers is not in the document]
ruslandoga: @zoldar thank you for telling me about this, I didn't know about this super handy feature!

zoldar (approved):

lib/plausible/s3.ex
ruslandoga: I've made some attempts at a better filename in https://github.com/plausible/analytics/pull/3921

PR #3844, opened 2024-02-28
Mutation to populate event session columns

macobo:
This PR introduces a new data migration `PopulateEventSessionColumns` which will populate the events table session property columns. These columns will be used to speed up queries by avoiding joins.

Depends on https://github.com/plausible/analytics/pull/3864 and https://github.com/plausible/analytics/pull/3865. Part of [the project to improve sessions and events table schemas](https://3.basecamp.com/5308029/buckets/35611491/messages/7061875211)

Tested locally and on staging.

## Operations 
To run this, use `./bin/plausible rpc "Plausible.DataMigration.PopulateEventSessionColumns.run"`

The script will loop, monitoring the in-progress mutation, exiting when it's done. 

To limit partition range being updated, use `./bin/plausible rpc "Plausible.DataMigration.PopulateEventSessionColumns.run(partition_range: ['202301', '202305'])"

Killing the script will not cancel the mutation. This can be done via `./bin/plausible rpc "Plausible.DataMigration.PopulateEventSessionColumns.kill"`

### Plan on plausible.io

We'll run this migration during a quiet period, e.g. a weekend, limiting partition range. 

If needed, will tune `max_replicated_mutations_in_queue` merge tree setting limiting mutation parallelism and/or kill the migration.

## Constaints

1. Users should ensure they have enough free space for the mutation (exact amount TBD, but ~30% extra disk space is wise)

## Notes

We might need to store information on whether this migration has been run for self-hosted users. Will clarify this once we have querying ready.


config/runtime.exs
```diff
@@ -406,7 +406,8 @@ config :plausible, Plausible.IngestRepo,
   max_buffer_size: ch_max_buffer_size,
   pool_size: ingest_pool_size,
   settings: [
-    materialized_views_ignore_errors: 1
+    materialized_views_ignore_errors: 1,
+    allow_nondeterministic_mutations: 1
```
aerosol, line 410: Would you mind explaining this one? How does it affect ingestion? If it's only needed for the migration would it make sense to consider a separate connection with this setting?

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,158 @@
...
+    Logger.info("Mutations seem done, cleaning up!")
+    {:ok, _} = run_sql("drop-sessions-dictionary", cluster?: cluster?)
+  end
+
+  def kill(opts \\ []) do
+    cluster? = Plausible.MigrationUtils.clustered_table?("events_v2")
+
+    report_progress(opts)
+
+    Logger.info("Killing running mutations")
```
aerosol, line 45: Please check if Logger works via an RPC call, I fear the output may be empty due to group leader. Otherwise, writing to stdout should be fine.

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,158 @@
+defmodule Plausible.DataMigration.PopulateEventSessionColumns do
```
aerosol, line 1: Will self-hosters need to run this?

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,158 @@
...
+  def run(opts \\ []) do
+    cluster? = Plausible.MigrationUtils.clustered_table?("sessions_v2")
+
+    {:ok, _} =
+      run_sql("create-sessions-dictionary",
+        cluster?: cluster?,
+        dictionary_connection_params: get_dictionary_connection_params(opts)
+      )
+
+    {partitions, _, _, _} = get_partitions(opts)
```
aerosol, line 26: Are you going to pass some partition range manually? How is  it determined?

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  # See https://clickhouse.com/docs/en/sql-reference/dictionaries#clickhouse for context
+  def get_dictionary_connection_params(opts \\ []) do
+    dictionary_connection_string = Keyword.get(opts, :dictionary_connection_string)
+
+    if dictionary_connection_string do
+      dictionary_connection_string
+    else
+      uri =
+        Application.get_env(:plausible, Plausible.IngestRepo) |> Keyword.get(:url) |> URI.parse()
```
aerosol, line 130: You can use `IngestRepo.config()` to avoid parsing the connection string

priv/data_migrations/PopulateEventSessionColumns/sql/create-sessions-dictionary.sql.eex
```diff
@@ -0,0 +1,26 @@
...
+    screen_size String,
+    operating_system String,
+    operating_system_version String,
+    browser String,
+    browser_version String
+)
+PRIMARY KEY site_id, session_id
+SOURCE(CLICKHOUSE(TABLE sessions_v2 <%= @dictionary_connection_params %>))
+LIFETIME(600000)
+LAYOUT(complex_key_cache(size_in_cells 1000000 max_threads_for_updates 6 allow_read_expired_keys 1))
```
aerosol, line 26: Would you mind explaining the rationale behind the numbers here? How do you determine `LIFETIME`, `size_in_cells` and `max_threads_for_updates`?

aerosol (commented):
Lots of interesting stuff in there. What steps did you undertake to test-drive this?

lib/plausible/data_migration/populate_event_session_columns.ex
ruslandoga: Since logging is async, it might not have enough time to flush.

```console
$ docker compose exec plausible bin/plausible rpc 'Logger.bare_log(:error, "hello")'
$ docker compose exec plausible bin/plausible rpc 'IO.puts "hm"'
hm
$ docker compose exec plausible bin/plausible rpc 'Logger.bare_log(:error, "hello"); :timer.sleep(2000)'

10:57:21.340 [error] hello
$ docker compose logs plausible
plausible-1  | Loading plausible..
plausible-1  | Starting dependencies..
plausible-1  | Starting repos..
plausible-1  | create Plausible.Repo database if it doesn't exist
plausible-1  | create Plausible.IngestRepo database if it doesn't exist
plausible-1  | Creation of Db successful!
plausible-1  | Loading plausible..
plausible-1  | Starting dependencies..
plausible-1  | Starting repos..
plausible-1  | Running migrations for Elixir.Plausible.Repo
plausible-1  | Running migrations for Elixir.Plausible.IngestRepo
plausible-1  | Migrations successful!
```

config/runtime.exs
ruslandoga: Settings can also be provided per query.

```elixir
IngesetRepo.query!(some_sql, params, settings: [allow_nondeterministic_mutations: 1])
```

lib/plausible/data_migration/populate_event_session_columns.ex
ruslandoga: > These columns will be used to speed up queries by avoiding joins.

I think self-hosters would appreciate this :)

macobo: They will need to run this.

We likely will need to set a flag _somewhere_ that will indicate whether this has been run or not. Not sure where yet - will figure it out together with the querying. :)

lib/plausible/data_migration/populate_event_session_columns.ex
macobo: On self-hosted, users will run this manually and likely do all in one go. On cloud, I'll be running this a few partitions to get a sense of impact on query speed. Given partitions are something like `202301` I'll likely start with a single partition and include more as I gain confidence on the larger data set.

priv/data_migrations/PopulateEventSessionColumns/sql/create-sessions-dictionary.sql.eex
macobo: Great question. Note that all these numbers are subject to change based on what we're seeing this when running this migration.

The primary source for this is https://clickhouse.com/docs/en/sql-reference/dictionaries

**Lifetime**

> For cache dictionaries, the expiration [lifetime](https://clickhouse.com/docs/en/sql-reference/dictionaries#dictionary-updates) of data in the cache can be set. If more time than lifetime has passed since loading the data in a cell, the cell’s value is not used and key becomes expired. The key is re-requested the next time it needs to be used. This behaviour can be configured with setting allow_read_expired_keys.

Picking 600000 here as an arbitrarily high value.

**size_in_cells**

> The size of the cache, in number of cells. Rounded up to a power of two.

This determines the size of the cache and thereby how much memory we allow the migration to consume. It's hard to say how big each cell will end up being, but overcounting it should be up to 1000000 * 1KiB = 1GiB. 

A larger cache won't necessarily win much in performance of the migration since all the events of a session should be packed relatively tight in the clickhouse dataset.

This should be enough for both self-hosted and cloud, though further tuning might be needed on cloud.

**max_threads_for_updates**

Since our dictionary is backed by a clickhouse table, this will determine how many reads we'll issue at it in parallel. Limiting it since it'll affect other traffic.

Note the dictionary will batch reads together, fetching a batch of sessions at a time.

6 seems like a reasonable number that _should_ work for both self-hosted and cloud.

lib/plausible/data_migration/populate_event_session_columns.ex
macobo: Thoughts whether I should change it to puts? Using logger mainly for outputting the date here :)

macobo:
> What steps did you undertake to test-drive this?

Ran this pretty much as-is on both my dev instance and staging.

lib/plausible/data_migration/populate_event_session_columns.ex
macobo: Wow, that simplifies this significantly! Thanks!

priv/data_migrations/PopulateEventSessionColumns/sql/create-sessions-dictionary.sql.eex
aerosol: > 1000000

> This determines the size of the cache and thereby how much memory

> This should be enough for both self-hosted and cloud, though further tuning might be needed on cloud.

Will that still work for people on modest VPSes where 1GiB might not be available?

> Note that all these numbers are subject to change based on what we're seeing this when running this migration.

Ah cool, were you planning to allow overriding those with input parameters?

lib/plausible/data_migration/populate_event_session_columns.ex
ruslandoga: I think we can check if the migration has been run during app startup and maybe just raise if it hasn't.

config/runtime.exs
ruslandoga: Settings in config.exs are just settings per query specified before-hand. The adapter just merges the two.

macobo: Confirmed it works. The reason I suspected it didn't is that [ALTER TABLE UPDATE](https://clickhouse.com/docs/en/sql-reference/statements/alter/update) doesn't allow setting settings explicitly as SELECT does.

priv/data_migrations/PopulateEventSessionColumns/sql/create-sessions-dictionary.sql.eex
macobo: I wasn't but this was easy enough to solve. Done. :)

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,170 @@
...
+      IO.puts("  Out of #{length(parts)} parts, #{total_size}")
+      IO.puts("  Running for #{format_duration(running_for)}")
+
+      if progress > 0 do
+        estimated_time_left = round(running_for / progress / 100 - running_for)
+        IO.puts("  Estimated #{progress}% done, #{format_duration(estimated_time_left)} left")
+      end
+
+      if latest_fail_reason do
+        Logger.warning(
```
aerosol, line 125: missed a spot?

lib/plausible/data_migration/populate_event_session_columns.ex
```diff
@@ -0,0 +1,170 @@
...
+  defp dictionary_config(opts) do
+    @default_dictionary_config
+    |> Map.merge(Keyword.get(opts, :dictionary_config, %{}))
+  end
+
+  # See https://clickhouse.com/docs/en/sql-reference/dictionaries#clickhouse for context
+  defp dictionary_connection_params() do
+    Plausible.IngestRepo.config()
+    |> Enum.map(fn config ->
+      case config do
```
aerosol, line 145: Just a stylistic note - completely non-issue, you can pattern match in the function head, e.g.

```elixir
Enum.map_(fn 
  {:database, db} -> ..
  {:username, u} -> ..
  _ -> ..
end)    
```

Question - will ClickHouse use some (here: incorrect) `default` credentials if not supplied, even when the credentials are already in the connection string?

macobo: > you can pattern match in the function head, e.g.

That's really cool!

> Question - will ClickHouse use some (here: incorrect) default credentials if not supplied, even when the credentials are already in the connection string?

Not sure I understand the question. I've confirmed this function works as needed locally, on staging and on prod. I'm not confident about all self-hosted setups, hence the `dictionary_connection_string` option escape hatch provided for them.

PR #3845, opened 2024-02-28
CSV imports (UI)

ruslandoga:
### Changes

This PR adds UI for CSV imports via S3.

https://github.com/plausible/analytics/assets/67764432/7ee9daf4-411c-4fa0-9f01-b174e8233862

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/csv_import.ex
+++ b/lib/plausible_web/live/csv_import.ex
@@ -153,10 +153,10 @@ defmodule PlausibleWeb.Live.CSVImport do
         uploads: uploads
       )
 
-    {:noreply,
-     redirect(socket,
-       to: Routes.site_path(socket, :settings_imports_exports, URI.encode_www_form(site.domain))
-     )}
+    redirect_to =
+      Routes.site_path(socket, :settings_imports_exports, URI.encode_www_form(site.domain))
+
+    {:noreply, redirect(socket, external: redirect_to)}
   end
 
   defp error_to_string(:too_large), do: "is too large (max size is 1 gigabyte)"
--- a/lib/plausible_web/live/csv_import.ex
+++ b/lib/plausible_web/live/csv_import.ex
@@ -46,26 +46,26 @@ defmodule PlausibleWeb.Live.CSVImport do
     ~H"""
     <label
       phx-drop-target={@upload.ref}
-      class="block border-2 dark:border-gray-600 rounded p-4 hover:border-indigo-300 dark:hover:border-indigo-600 transition cursor-pointer"
+      class="block border-2 dark:border-gray-600 rounded p-4 group hover:border-indigo-500 dark:hover:border-indigo-600 transition cursor-pointer"
     >
       <div class="flex items-center">
-        <div class="bg-indigo-200 dark:bg-indigo-700 rounded p-1 hover:bg-indigo-300 dark:hover:bg-indigo-600 transition cursor-pointer">
-          <Heroicons.plus class="w-4 h-4" />
+        <div class="bg-gray-200 dark:bg-gray-600 rounded p-1 group-hover:bg-indigo-500 dark:group-hover:bg-indigo-600 transition">
+          <Heroicons.folder_plus class="w-5 h-5 group-hover:text-white transition" />
         </div>
         <span class="ml-2 text-sm text-gray-600 dark:text-gray-500">
-          (or drag-and-drop here)
+          (or drag-and-drop your folder here)
         </span>
         <.live_file_input upload={@upload} class="hidden" webkitdirectory />
       </div>
 
-      <div id="imported-tables" class="mt-5 mb-1 space-y-1">
+      <ul id="imported-tables" class="mt-3.5 mb-0.5 space-y-1.5">
         <.imported_table
           :for={{table, upload} <- @imported_tables}
           table={table}
           upload={upload}
           errors={if(upload, do: upload_errors(@upload, upload), else: [])}
         />
-      </div>
+      </ul>
     </label>
     """
   end
@@ -101,7 +101,7 @@ defmodule PlausibleWeb.Live.CSVImport do
     assigns = assign(assigns, status: status)
 
     ~H"""
-    <div id={@table} class="ml-1">
+    <li id={@table} class="ml-1.5">
       <div class="flex items-center space-x-2">
         <Heroicons.document_check :if={@status == :success} class="w-4 h-4 text-indigo-600" />
         <PlausibleWeb.Components.Generic.spinner
@@ -127,7 +127,7 @@ defmodule PlausibleWeb.Live.CSVImport do
       <p :for={error <- @errors} class="ml-6 text-sm text-red-600 dark:text-red-700">
         <%= error_to_string(error) %>
       </p>
-    </div>
+    </li>
     """
   end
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ postgres-stop: ## Stop and remove the postgres container
 	docker stop plausible_db && docker rm plausible_db
 
 minio: ## Start a transient container with a recent version of minio (s3)
-	docker run -d --rm -p 6000:6000 -p 6001:6001 --name plausible_minio minio/minio server /data --address ":6000" --console-address ":6001"
+	docker run -d --rm -p 10000:10000 -p 10001:10001 --name plausible_minio minio/minio server /data --address ":10000" --console-address ":10001"
```
ruslandoga, line 41: I'm having trouble making uploads to port localhost:6000 from the browser on Mac since that port is sometimes used by X111 and therefore is "protected"

zoldar: It's possible to submit "implicit" form with `PlausibleWeb.Components.Generic.button` too. Example: https://github.com/plausible/analytics/blob/4d7d88cfeccbe35224bad5ffd3ed161784fa67d7/lib/plausible_web/live/imports_exports_settings.ex#L94-L101

lib/plausible_web/live/csv_import.ex
```diff
@@ -0,0 +1,228 @@
...
+
+  @impl true
+  def handle_event("validate-upload-form", _params, socket) do
+    {:noreply, process_imported_tables(socket)}
+  end
+
+  def handle_event("submit-upload-form", _params, socket) do
+    %{site_id: site_id, user_id: user_id, date_range: date_range} = socket.assigns
+    site = Plausible.Repo.get!(Plausible.Site, site_id)
+    user = Plausible.Repo.get!(Plausible.Auth.User, user_id)
```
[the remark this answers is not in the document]
ruslandoga: I think it's equivalent. Right now there are two requests to DB on submit. with `assign_new` there would be two request on "connected" render, only the "disconnected" render would pick these values up from the conn.

lib/plausible_web/live/csv_import.ex
```diff
@@ -0,0 +1,228 @@
...
+    {:ok, _job} =
+      CSVImporter.new_import(site, user,
+        start_date: date_range.first,
+        end_date: date_range.last,
+        uploads: uploads
+      )
+
+    {:noreply,
+     redirect(socket,
+       to: Routes.site_path(socket, :settings_imports_exports, URI.encode_www_form(site.domain))
```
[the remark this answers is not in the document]
ruslandoga: Done in https://github.com/plausible/analytics/pull/3845/commits/c817a13f3eacc857a5e89d3b808e03d9234228a4

lib/plausible_web/live/csv_import.ex
```diff
@@ -46,26 +46,26 @@ defmodule PlausibleWeb.Live.CSVImport do
...
     <label
       phx-drop-target={@upload.ref}
-      class="block border-2 dark:border-gray-600 rounded p-4 hover:border-indigo-300 dark:hover:border-indigo-600 transition cursor-pointer"
+      class="block border-2 dark:border-gray-600 rounded p-4 group hover:border-indigo-500 dark:hover:border-indigo-600 transition cursor-pointer"
     >
       <div class="flex items-center">
-        <div class="bg-indigo-200 dark:bg-indigo-700 rounded p-1 hover:bg-indigo-300 dark:hover:bg-indigo-600 transition cursor-pointer">
-          <Heroicons.plus class="w-4 h-4" />
+        <div class="bg-gray-200 dark:bg-gray-600 rounded p-1 group-hover:bg-indigo-500 dark:group-hover:bg-indigo-600 transition">
+          <Heroicons.folder_plus class="w-5 h-5 group-hover:text-white transition" />
```
ruslandoga, line 53: Screenshot 2024-03-21 at 20 32 39

zoldar:
1. The upload dialog does not let me select particular files, I'm not sure why 🤔. The only way I can select the files is entering the (extracted) export directory and clicking "Upload". FF gives a warning that I'm trying to upload all the files from a given folder but they appear in the upload list once I confirm it. 
    image

2. I have first made an export from freshly loaded seeds, made sure minio container is running, extracted the export, uploaded it and initiated import. The worker seems to start and it reaches https://github.com/ruslandoga/analytics/blob/50f6a3d7bab940a2f3bc11cd4e26590e0533019f/lib/plausible/imported/csv_importer.ex#L41 for the first upload but remains stuck there for several minutes without any activity (the task is hanging in "executing" state for that time as well). Have you experienced anything like that?

3. I'm not sure if we have discussed this, but are we planning to support uploading zipped archive as well? That would make the export->import process more streamlined for the common scenario of moving between hosted and cloud, one way or another.

ruslandoga:
1. I'll try using Firefox tomorrow.
2. No
3. Yeah, but the importer would need to be reworked to support Zip archives, maybe https://clickhouse.com/blog/clickhouse-release-23-08#direct-import-from-archives-nikita-keba-antonio-andelic-pavel-kruglov could be used (https://github.com/ClickHouse/ClickHouse/issues/54726)

ruslandoga:
Removing `webkitdirectory` added in https://github.com/plausible/analytics/pull/3845/commits/e8039097473e6528e0df32c08517d211f42b9b03 makes it work in FF Thank you very much for spotting it! I should've tested other browsers before adding it.

ruslandoga:
Re 2, are there any errors on the Oban job? Maybe ClickHouse S3 timeout or similar. In that case it probably means you need to set `S3_CLICKHOUSE_HOST` to make MinIO visible to ClickHouse.

zoldar:
Oban Job kept executing without any errors.

I have resolved the issue by swapping "host.docker.internal" for "localhost" in S3 URLs sent to import worker:

```diff
diff --git a/lib/plausible/imported/csv_importer.ex b/lib/plausible/imported/csv_importer.ex
index 6376d851d..6f6c422ee 100644
--- a/lib/plausible/imported/csv_importer.ex
+++ b/lib/plausible/imported/csv_importer.ex
@@ -65,7 +65,9 @@ defmodule Plausible.Imported.CSVImporter do
           "end_date" => end_date
         }
 
+      IO.inspect upload, label: :ABOUT_TO_QUERY
       Ch.query!(ch, statement, params, timeout: :infinity)
+      IO.inspect upload, label: :AFTER_QUERY # THIS WAS NEVER SHOWN UNTIL I CHANGED THE HOST
     end)
   rescue
     # we are cancelling on any argument or ClickHouse errors
diff --git a/lib/plausible_web/live/csv_import.ex b/lib/plausible_web/live/csv_import.ex
index 38c97ca17..f538b3e5a 100644
--- a/lib/plausible_web/live/csv_import.ex
+++ b/lib/plausible_web/live/csv_import.ex
@@ -168,6 +168,8 @@ defmodule PlausibleWeb.Live.CSVImport do
     %{s3_url: s3_url, presigned_url: upload_url} =
       Plausible.S3.import_presign_upload(socket.assigns.site_id, entry.client_name)
 
+    s3_url = String.replace(s3_url, "localhost", "host.docker.internal")
+
     {:ok, %{uploader: "S3", s3_url: s3_url, url: upload_url}, socket}
   end
```

After the change, the import proceeds basically immediately.

In CH logs I can see connection failing with "connection refused" when running the insert query but it surprisingly seems to not propagate back to the client somehow? I'm not sure, I haven't digged too deep into this.

zoldar:
BTW something for a follow-up perhaps but IMHO it would make sense to hide CSV export and import UI when S3 is not configured, at least until local storage version is implemented. WDYT?

zoldar (approved):

PR #3851, opened 2024-02-29
Implement basics of GA4 import

zoldar:
### Changes

This is a very first pass of GA4 import implementation. There are already subsequent follow-up changes planned which will be done on top of that but we want to get these basics merged first so that we avoid working off a long-living branch.

- GoogleAnalytics4 importer is introduced
- Google.API got extended with GA4 specific calls on API and HTTP level and existing API for UA got refactored so that 3 are clearly separated (UA logic is going to be removed in a matter of months, most likely)
- Import specific controller actions are extracted into separate controllers for UA and GA4 (eventually, they are going to be merged into one, as there will be a single unified flow where UA views and GA4 properties will be selectable from a single list)

Everything is still behind a feature flag.


pau1phi11ips:
Very keen to start using this.

lib/plausible/google/ga4/api.ex
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  def get_property(access_token, lookup_property) do
+    case list_properties(access_token) do
+      {:ok, properties} ->
+        property =
+          properties
+          |> Enum.map(&elem(&1, 1))
+          |> List.flatten()
+          |> Enum.find(fn {_name, property} -> property == lookup_property end)
```
aerosol, line 48: any chance we won't find it here?

lib/plausible/google/ga4/http.ex
```diff
@@ -0,0 +1,199 @@
...
+        url,
+        [{"Authorization", "Bearer #{report_request.access_token}"}],
+        params,
+        receive_timeout: 60_000
+      )
+
+    with {:ok, %{body: body}} <- response,
+         # File.write!("fixture/ga4_report_#{report_request.dataset}.json", Jason.encode!(body)),
+         {:ok, report} <- parse_report_from_response(body),
+         row_count <- Map.get(report, "rowCount"),
```
aerosol, line 54: fallback when not there or `fetch!`?

lib/plausible/google/ga4/http.ex
```diff
@@ -0,0 +1,199 @@
...
+      {:error, %HTTPClient.Non200Error{} = error} when error.reason.status in [401, 403] ->
+        {:error, :authentication_failed}
+
+      {:error, %HTTPClient.Non200Error{} = error} ->
+        Sentry.capture_message("Error listing Google accounts for user", extra: %{error: error})
+        {:error, :unknown}
+    end
+  end
+
+  @earliest_valid_date "2015-08-14"
```
aerosol, line 150: where does that come from?

lib/plausible/imported/google_analytics4.ex
```diff
@@ -0,0 +1,253 @@
...
+  #     exit_page: Map.fetch!(row.dimensions, "exitPage"),
+  #     visitors: row.metrics |> Map.fetch!("totalUsers") |> parse_number(),
+  #     exits: row.metrics |> Map.fetch!("sessions") |> parse_number()
+  #   }
+  # end
+
+  defp new_from_report(site_id, import_id, "imported_locations", row) do
+    country_code = row.dimensions |> Map.fetch!("countryId") |> default_if_missing("")
+    city_name = row.dimensions |> Map.fetch!("city") |> default_if_missing("")
+    city_data = Location.get_city(city_name, country_code)
```
aerosol, line 158: so not a good chance we'll have a match here?

aerosol (approved):
I think we should push this forward. Skipping JSON fixtures and code moved around, makes it less scary to read, but since everything is behind a FF, better to land this early on and iterate.

lib/plausible_web/controllers/google_analytics4_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+        "access_token" => access_token,
+        "refresh_token" => refresh_token,
+        "expires_at" => expires_at
+      }) do
+    site = conn.assigns.site
+    start_date = Plausible.Google.GA4.API.get_analytics_start_date(access_token, property)
+
+    case start_date do
+      {:ok, nil} ->
+        {:ok, properties} = Plausible.Google.GA4.API.list_properties(access_token)
```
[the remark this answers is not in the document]
zoldar: I'll add better timeout handling eventually.

lib/plausible/google/ga4/http.ex
zoldar: Google Data API returns an error if you query for anything before that date.

lib/plausible/google/ga4/http.ex
zoldar: It should be `fetch!` 👍

lib/plausible/google/ga4/api.ex
zoldar: That's crude copypasta. There's actually a dedicated endpoint for querying a single property which we should use instead. I'll rework this.

zoldar:
I'll address the remaining feedback in https://github.com/plausible/analytics/pull/3888 which is already drafted on top of this branch.

PR #3854, opened 2024-03-01
Move `add_percentage` logic into clickhouse

macobo:
### Changes

This PR:
1. Moves logic for `add_percentage` into clickhouse query building
2. Removes special-case (none) handling for custom property breakdowns

Why? I got stuck shaving yaks when planning on how to query events/sessions once we have session properties available on events table. Both add_percentage and conversion rate calculations add a lot of in-app complexity which we might be able to avoid.


aerosol (approved):

aerosol:
Will that help with https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/6970466136 ?

RobertJoonas (approved):
Nicely done! 🙌 Looks good to me!

lib/plausible/stats/imported.ex
```diff
@@ -524,6 +528,29 @@ defmodule Plausible.Stats.Imported do
     |> select_joined_metrics(rest)
   end
 
+  defp select_joined_metrics(q, [:percentage | rest]) do
+    q
+    |> select_merge([s, i], %{
+      __total_visitors:
+        selected_as(
+          fragment(
+            "coalesce(sum(?) OVER (), 0) + coalesce(sum(?) OVER (), 0)",
```
macobo, line 537: This is incorrect (but not caught by tests) - number of visitors isn't neccessarily the sum of visitor values.

lib/plausible/stats/base.ex
```diff
@@ -566,4 +586,20 @@ defmodule Plausible.Stats.Base do
           fragment("arrayExists(k -> match(?, k), ?)", get_by_key(e, :meta, ^prop_name), ^regexes)
     )
   end
+
+  defp add_visitors_percentage(q) do
+    from(e in q,
+      select_merge: %{
+        __total_visitors:
+          selected_as(fragment("sum(?) OVER ()", selected_as(:visitors)), :__total_visitors),
```
macobo, line 594: This is incorrect (but not caught by tests) - number of visitors isn't neccessarily the sum of visitor values.

macobo:
Marking it as a draft - discovered a bug in how the new logic handles users making multiple sessions. Added a test for the case for now, will fix later.

lib/plausible/stats/breakdown.ex
```diff
@@ -445,10 +422,19 @@ defmodule Plausible.Stats.Breakdown do
...
+            fragment(
+              "if(not empty(?), ?, '(none)')",
+              get_by_key(e, :meta, ^prop),
+              get_by_key(e, :meta, ^prop)
+            ),
+            :breakdown_prop_value
+          )
+      },
+      group_by: selected_as(:breakdown_prop_value),
+      order_by: {:asc, selected_as(:breakdown_prop_value)}
```
zoldar, line 437: It seems that "(none)" => "Direct / None" entry is rendered according to the sorting order (closer to the end in practice) of all the breakdown lists on stats page while currently it's at the top. Are we sure we want it like this?

macobo: Will take a look, great catch. I was hoping to go for a no (significant) behavioral changes here, so this might be something which tests don't cover.

zoldar: Hmm, sorry, I might have got it wrong - or at least, when checking against current `master`, the ordering for top sources is the same ("Direct / None" is last). Staging and production have it first, however. Now I'm not sure if that's because of flawed seeds or is there something else at play.

zoldar (approved):
Ordering of "(none)" seems to be fine in the end - seems I got it wrong at first 😅

PR #3870, opened 2024-03-07
Set exit_page only on pageviews

macobo:
### Changes

Previously we set and updated `exit_page` on all events, which can cause issues when:
1. Sessions without any pageviews are made (e.g. user leaves a tab open and returns to it)
2. It results in exit_page exit_rate calculations returning numbers above 100

Additional context under https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/7162713530

### Tests

### Changelog


aerosol (changes requested):
Would be good to look into those test failures. Can't quickly tell if it's a fixture error, a regression or if there's something easily abused with the way we populate test data.

macobo:
Test failures were real - we were creating Signup events in the relevant tests.

Given that this also affects bounce rates, etc as per these example tests, it would be great to have @ukutaht's input before merging this.

> but it seems to also require updating how we query for stats so that we filter out custom events-only sessions from breakdowns/aggregates in which they do not apply

I'm considering that a separate task that's slightly trickier than this given the wide-reaching ramifications. I was planning to make sure this gets a satisfying solution, but is less urgent than this fix. If you have input/thoughts or want to lead that feel free! :bow:

test/plausible/session/cache_store_test.exs
```diff
@@ -89,6 +89,61 @@ defmodule Plausible.Session.CacheStoreTest do
...
+        site_id: site_id,
+        pathname: "/path/3",
+        user_id: 1
+      )
+
+    flush([event1, event2, event3])
+
+    session = get_session(site_id)
+
+    assert session.exit_page == "/path/2"
```
ukutaht, line 142: For extra clarity I think this test could also assert that `session.entry_page == ""`

ukutaht (approved):

ukutaht:
Wait, I approved too early. Why does this affect bounce rates? 👀

ukutaht:
As per [https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/7162713530#__recording_7186457996](discussion), this solution is good. Let's do it!

test/plausible/session/cache_store_test.exs
macobo: This brings up a question: Should we also update entry_page on a subsequent event if a) it's a pageview and b) session doesn't have entry_page set?

ukutaht: > After re-visiting I think we should set exit_page to /path/2 here as well, thoughts?

you mean entry_page rather than exit_page, right?

Yeah I think that does make sense. Then entry page and exit page will always correspond to the first and last pageview in a session which feels intuitive to me.

PR #3878, opened 2024-03-11
Replace caching engine

aerosol:
### Changes

This PR swaps out [Cachex](https://github.com/whitfin/cachex) for [ConCache](https://github.com/sasa1977/con_cache). The former feels too advanced for our needs, plus with https://github.com/plausible/analytics/commit/83a46fb47e96f51f058ad7d3315b0d218c331c6c we're not evicting user agents properly (ref: https://github.com/whitfin/cachex/issues/329).

All application caches are now accessed through a single interface (`Plausible.Cache.Adapter`), even if they don't implement the periodic refresh semantics. This lets us test drive ConCache uniformly and offers future-proof flexibility.

Notable mentions:

- ConCache alone does not offer hit/miss ratio tracking, so this was added so we can still track utilization in grafana. Metric names will change (`s/cachex/cache`), we could stick with keeping the `cachex` string, but I was annoyed by that.
- ConCache does not offer cache eviction by specifying the number of items, but this isn't a hard requirement - from now on, the User Agents cache uses (roughly) 1 hour global TTL per item
- merging periodic cache refreshes is now much faster, since we're levering dirty ets writes 
- all caches are now configured at the supervision tree definition level, arguably resulting in a better overview

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/telemetry/plausible_metrics.ex
```diff
@@ -88,29 +88,23 @@ defmodule Plausible.PromEx.Plugins.PlausibleMetrics do
...
-      hit_rate: user_agents_hit_rate
+    :telemetry.execute([:prom_ex, :plugin, :cache, :user_agents], %{
+      count: user_agents_stats.count,
+      hit_rate: user_agents_stats.hit_rate
     })
 
-    :telemetry.execute([:prom_ex, :plugin, :cachex, :sessions], %{
-      count: sessions_count,
-      hit_rate: sessions_hit_rate
+    :telemetry.execute([:prom_ex, :plugin, :cache, :sessions], %{
```
zoldar, line 102: Just a note that we'll need to update respective graphs/dashboards in Grafana, right?

zoldar (approved):

lib/plausible/telemetry/plausible_metrics.ex
aerosol: yup

lib/plausible/cache/adapter.ex
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  @spec size(atom()) :: non_neg_integer() | nil
+  def size(cache_name) do
+    try do
+      ConCache.size(cache_name)
+    catch
+      :exit, _ -> nil
+    end
+  end
```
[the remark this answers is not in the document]
aerosol: Ah excellent. bac7fde2f

lib/plausible/cache.ex
```diff
@@ -165,21 +165,8 @@ defmodule Plausible.Cache do
...
-      end
-
-      @spec hit_rate() :: number()
-      def hit_rate(cache_name \\ name()) do
-        case Cachex.stats(cache_name) do
-          {:ok, stats} -> Map.get(stats, :hit_rate, 0)
-          _ -> 0
-        end
-      end
+      defdelegate size(cache_name \\ name()), to: Plausible.Cache.Stats
```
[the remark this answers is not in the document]
aerosol: d6717b779

cnkk:
> ConCache does not offer cache eviction by specifying the number of items, but this isn't a hard requirement - from now on, the User Agents cache uses (roughly) 1 hour global TTL per item

If increasing this helps with performance, let's do it?

aerosol:
> > ConCache does not offer cache eviction by specifying the number of items, but this isn't a hard requirement - from now on, the User Agents cache uses (roughly) 1 hour global TTL per item
> 
> If increasing this helps with performance, let's do it?

No proof it does - first cold UA takes a bit longer (even up to 10ms, YMMV), but how many of them come back frequently enough for this to be effective is unknown. I think we must be wary of modest setups, memory-wise, running this too. Perhaps I should make this configurable via env?

aerosol:
> even up to 10ms

up to 100ms

ruslandoga (approved):

lib/plausible/telemetry/plausible_metrics.ex
aerosol: in the end - no - it's only the event shape that changes, metric names remain the same 😅

PR #3879, opened 2024-03-11
changes to the in-app footer

metmarkosaric:
changes to the in-app footer to focus on product usage links rather than marketing pages


aerosol:
fix up in https://github.com/plausible/analytics/pull/3880

PR #3880, opened 2024-03-11
changes to the in-app footer (clean)

aerosol:
ref https://github.com/plausible/analytics/pull/3879


aerosol:
@metmarkosaric just double checking if the margins are intentional here:

image

aerosol:
I think we could also fix the capitalization - e.g. "Search Console" vs "WordPress plugin".

metmarkosaric:
@aerosol ah no that's not intentional. I was just coding by copy pasting things and didn't have a way to test the end result 😀 thanks!

PR #3888, opened 2024-03-12
Unify UA and GA4 import flow into one

zoldar:
Follow-up to https://github.com/plausible/analytics/pull/3851, unifying GA4 and UA import flows into a single one where user selected view _or_ property to import from a single list.

- [x] apply feedback from https://github.com/plausible/analytics/pull/3851


lib/plausible_web/controllers/google_analytics_controller.ex
```diff
@@ -178,21 +189,36 @@ defmodule PlausibleWeb.UniversalAnalyticsController do
...
         legacy: legacy == "true"
       )
+    end
 
     conn
     |> put_flash(:success, "Import scheduled. An email will be sent when it completes.")
     |> redirect(external: redirect_route)
   end
+
+  def property?(value) when is_binary(value), do: String.starts_with?(value, "properties/")
```
RobertJoonas, line 223: This function appears to be a duplicate for the one in `Plausible.Google.API`. WDYT about making the API one public and using it in this module too?

RobertJoonas (approved):
Looks good to me! I didn’t manage to test the UI yet, but it's behind a FF anyway, so I can do that on prod when this is live.

Good job on the test coverage! 🙌

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -750,43 +750,29 @@ defmodule PlausibleWeb.AuthController do
...
-              Routes.universal_analytics_path(conn, :view_id_form, site.domain,
-                access_token: res["access_token"],
-                refresh_token: res["refresh_token"],
-                expires_at: NaiveDateTime.to_iso8601(expires_at),
-                legacy: legacy
-              )
-          )
-        end
+        redirect(conn,
+          external:
```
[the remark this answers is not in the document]
aerosol: Because of domains with Unicode characters: https://github.com/plausible/analytics/pull/3560

lib/plausible_web/controllers/google_analytics_controller.ex
zoldar: @RobertJoonas ah, thanks for bringing it up, totally slipped me 😅

PR #3858, opened 2024-03-04
Stats api time on page

RobertJoonas:
### Changes

This PR adds support for the time_on_page metric in Stats API.

### Tests
- [x] Automated tests have been added

### Changelog
- [x] Entry has been added to changelog

### Documentation
- [x] PR pending: https://github.com/plausible/docs/pull/476

### Dark mode
- [x] This PR does not change the UI

PR #3890, opened 2024-03-12
Fix division by zero in imported queries

RobertJoonas:
### Changes

This PR tackles the problem that surfaced with https://github.com/plausible/analytics/pull/3858. The PR stopped casting `nil` values into zeros in the aggregate API **after** having fetched those results from ClickHouse:

image

Consequently some Top Stats queries crashed due to the `calculate_change` function blowing up when getting `nil` as input when expecting a number. This was mitigated with https://github.com/plausible/analytics/pull/3886, but it did not tackle the root cause of the issue. Hence, this PR rolls that change back - we can still keep relying on `bounce_rate` being a number in the `calculate_change` function.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -488,7 +488,15 @@ defmodule Plausible.Stats.Imported do
     |> select_merge([s, i], %{
       views_per_visit:
         fragment(
-          "round((? + ? * coalesce(?, 0)) / (coalesce(?, 0) + coalesce(?, 0)), 2)",
+          """
+          if(
+            coalesce(?, 0) + coalesce(?, 0) > 0,
+            round((? + ? * coalesce(?, 0)) / (coalesce(?, 0) + coalesce(?, 0)), 2),
+            0
+          )
+          """,
+          s.__internal_visits,
+          i.__internal_visits,
           i.pageviews,
           s.views_per_visit,
           s.__internal_visits,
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/imported.ex
```diff
@@ -504,7 +512,15 @@ defmodule Plausible.Stats.Imported do
     |> select_merge([s, i], %{
       bounce_rate:
         fragment(
-          "round(100 * (coalesce(?, 0) + coalesce((? * ? / 100), 0)) / (coalesce(?, 0) + coalesce(?, 0)))",
+          """
+          if(
+            coalesce(?, 0) + coalesce(?, 0) > 0,
+            round(100 * (coalesce(?, 0) + coalesce((? * ? / 100), 0)) / (coalesce(?, 0) + coalesce(?, 0))),
```
aerosol, line 518: shouldn't that be rounded to 2 decimal points for consistency?

aerosol (approved):
looks good, one comment re rounding

lib/plausible/stats/imported.ex
RobertJoonas: `bounce_rate` has always been reported as integer, it's also rounded to an integer in the native queries. I think that works pretty well right now and people don't seem to be requesting more precision for this metric.

PR #3891, opened 2024-03-12
Reapply #3878 + bugfix hit rate tracking

aerosol:
### Changes

This PR re-applies [#3878](https://github.com/plausible/analytics/pull/3878) with one extra commit on top of it: https://github.com/plausible/analytics/pull/3891/commits/6019ebe14681497d2f64f693df423cf5f2201d4f

Previous approach wrongly assumed we can detect if an item was warm or cold on `get_or_store`.
Ideally the telemetry (or any functionally equivalent approach) would have been implemented in con_cache - [I guess we'll find out](https://github.com/sasa1977/con_cache/issues/75). Luckily for us, `con_cache` exposes low-level `Operations` interface for which I'm very grateful :)

### Tests

### Changelog

### Documentation

### Dark mode


mix.lock
```diff
@@ -8,7 +8,6 @@
...
   "bunt": {:hex, :bunt, "1.0.0", "081c2c665f086849e6d57900292b3a161727ab40431219529f13c4ddcf3e7a44", [:mix], [], "hexpm", "dc5f86aa08a5f6fa6b8096f0735c4e76d54ae5c9fa2c143e5a1fc7c1cd9bb6b5"},
   "bypass": {:hex, :bypass, "2.1.0", "909782781bf8e20ee86a9cabde36b259d44af8b9f38756173e8f5e2e1fabb9b1", [:mix], [{:plug, "~> 1.7", [hex: :plug, repo: "hexpm", optional: false]}, {:plug_cowboy, "~> 2.0", [hex: :plug_cowboy, repo: "hexpm", optional: false]}, {:ranch, "~> 1.3", [hex: :ranch, repo: "hexpm", optional: false]}], "hexpm", "d9b5df8fa5b7a6efa08384e9bbecfe4ce61c77d28a4282f79e02f1ef78d96b80"},
-  "cachex": {:hex, :cachex, "3.6.0", "14a1bfbeee060dd9bec25a5b6f4e4691e3670ebda28c8ba2884b12fe30b36bf8", [:mix], [{:eternal, "~> 1.2", [hex: :eternal, repo: "hexpm", optional: false]}, {:jumper, "~> 1.0", [hex: :jumper, repo: "hexpm", optional: false]}, {:sleeplocks, "~> 1.1", [hex: :sleeplocks, repo: "hexpm", optional: false]}, {:unsafe, "~> 1.0", [hex: :unsafe, repo: "hexpm", optional: false]}], "hexpm", "ebf24e373883bc8e0c8d894a63bbe102ae13d918f790121f5cfe6e485cc8e2e2"},
```
ruslandoga, line 11: `mix deps.clean --unused --unlock` also removes and unlocks the following deps for me:

```
* eternal
* jumper
* sleeplocks
* unsafe
```

lib/plausible/cache/stats.ex
```diff
@@ -0,0 +1,84 @@
...
+  def record_miss(cache_name) do
+    :telemetry.execute(@telemetry_miss, %{}, %{cache_name: cache_name})
+  end
+
+  def init(nil) do
+    __MODULE__ =
+      :ets.new(__MODULE__, [
+        :public,
+        :named_table,
+        :ordered_set,
```
ruslandoga, line 31: I wonder why not just `:set`? It doesn't seem to do range queries.

ruslandoga: I ran a benchmark and there doesn't seem to be any difference in `ips`

I thought `:set` would be faster...

ruslandoga: Seems like the telemetry.execute "dominates" over the ets ops hence the little difference. Either way it's super fast.

mix.exs
```diff
@@ -40,7 +40,9 @@ defmodule Plausible.MixProject do
         :logger,
         :runtime_tools,
         :tls_certificate_check,
-        :opentelemetry_exporter
+        :opentelemetry_exporter,
+        :wx,
+        :observer
```
ruslandoga, line 45: I don't know if it affects anything (like increasing the app size), but would it make sense to not include wx and observer in prod builds?

aerosol: Ah, completely unrelated change, thanks for spotting

aerosol: 56d895114

lib/plausible/cache/stats.ex
aerosol: Oh cool, thanks for running these.

Re:

> I thought :set would be faster...

From your benchmarks it looks like it is, still? Or do you mean you expected a bigger difference? I guess it doesn't cost anything to change it to `:set`, do you agree?

@zoldar even suggested [atomic counters](https://www.erlang.org/doc/man/counters), but I thought with the necessity of keeping refs, it wasn't worth it. WYT?

mix.lock
aerosol: Great, thanks, f38b4067a

lib/plausible/cache/stats.ex
aerosol: 9fb8f2bba

ruslandoga: > Or do you mean you expected a bigger difference?

Yes, I expected the difference to be more noticeable.

> I guess it doesn't cost anything to change it to :set, do you agree?

I agree :)

> [atomic counters](https://www.erlang.org/doc/man/counters) WYT?

In my tests for the rate limiter they were somehow slower than ets.update_counter but the use case there was slightly different.

- https://github.com/ruslandoga/rate_limit, atomics were used in rate_limiter and atomics3

I think Stats are already fast enough :)

ruslandoga (approved):

Issue #3762, opened 2024-02-07
cookie named "_plausible_key" exceeds maximum size of 4096 bytes


gillbates:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

version: plausible/analytics:v2.0

it always shows following error when I try to add a new site.

```
Server: 127.0.0.1:8000 (http)
Request: POST /sites
** (exit) an exception was raised:
    ** (Plug.Conn.CookieOverflowError) cookie named "_plausible_key" exceeds maximum size of 4096 bytes
        (plug 1.14.2) lib/plug/conn.ex:1857: Plug.Conn.verify_cookie!/2
        (plug 1.14.2) lib/plug/conn.ex:1850: anonymous fn/2 in Plug.Conn.merge_headers/2
        (stdlib 4.2) maps.erl:411: :maps.fold_1/3
        (plug 1.14.2) lib/plug/conn.ex:1842: Plug.Conn.run_before_send/2
        (plug 1.14.2) lib/plug/conn.ex:443: Plug.Conn.send_resp/1
        (plausible 0.0.1) lib/plausible_web/controllers/site_controller.ex:1: PlausibleWeb.SiteController.action/2
        (plausible 0.0.1) lib/plausible_web/controllers/site_controller.ex:1: PlausibleWeb.SiteController.phoenix_controller_pipeline/2
        (phoenix 1.7.2) lib/phoenix/router.ex:430: Phoenix.Router.__call__/5
03:33:57.914 [error] #PID<0.29009.1316> running PlausibleWeb.Endpoint (connection #PID<0.27803.1316>, stream id 20) terminated
```

but the site is listed after I refresh the page, any ideas why?

### Expected behavior

we could add site ...

### Screenshots

![CleanShot 2024-02-07 at 11 38 18@2x](https://github.com/plausible/analytics/assets/783547/9269fcf8-4be1-421e-a92f-21740a13e10c)

### Environment

```markdown
- OS:
- Browser:
- Browser Version:
```

ruslandoga:
👋 @gillbates 

Can you please show the contents of the cookie? You can decode it using your local Elixir's `iex` or `bin/plausible remote`

```console
$ cd hosting
$ docker compose exec plausible bin/plausible remote
```
[...]

PR #3892, opened 2024-03-13
Use sessionStorage for "dashboard first launch" banner tracking

aerosol:
Keeping it within the cookie is problematic, as the banners don't expire and overflow the cookie with data when enough new sites are added.

Ref https://github.com/plausible/analytics/issues/3762

Data in sessionStorage is cleared when the page session ends. 

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/templates/site/snippet.html.heex
```diff
@@ -1,3 +1,6 @@
+<script :if={@conn.params["site_created"] == "true"}>
```
zoldar, line 1: I know this won't be reused but, for the sake of keeping related stuff in close proximity, perhaps we could move this script and related banner to a dedicated component module, like `PlausibleWeb.Components.EmailReportBanner` ? The name of the storage item could also be put together in one place too, this way.

aerosol: Ah, nice. Something like 7da6ea26e ?

zoldar (approved):

PR #3895, opened 2024-03-14
CSV imports (no UI)

ruslandoga:
### Changes

This PR extracts the "date range in filename" change from https://github.com/plausible/analytics/pull/3845

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/workers/export_csv.ex
+++ b/lib/workers/export_csv.ex
@@ -25,13 +25,14 @@ defmodule Plausible.Workers.ExportCSV do
       |> Keyword.replace!(:pool_size, 1)
       |> Ch.start_link()
 
+    # NOTE: what if 1970-01-01?
     # NOTE: should we use site.timezone?
-    # %Ch.Result{rows: [[min_date, max_date]]} =
-    #   Ch.query!(
-    #     ch,
-    #     "SELECT toDate(min(timestamp)), toDate(max(timestamp)) FROM events_v2 WHERE site_id={site_id:UInt64}",
-    #     %{"site_id" => site_id}
-    #   )
+    %Ch.Result{rows: [[min_date, max_date]]} =
+      Ch.query!(
+        ch,
+        "SELECT toDate(min(timestamp)), toDate(max(timestamp)) FROM events_v2 WHERE site_id={site_id:UInt64}",
+        %{"site_id" => site_id}
+      )
 
     download_url =
       DBConnection.run(
@@ -39,8 +40,10 @@ defmodule Plausible.Workers.ExportCSV do
         fn conn ->
           conn
           |> Plausible.Exports.stream_archive(
-            # date_range: Date.range(min_date, max_date)
-            Plausible.Exports.export_queries(site_id, extname: ".csv"),
+            Plausible.Exports.export_queries(site_id,
+              date_range: Date.range(min_date, max_date),
+              extname: ".csv"
+            ),
             format: "CSVWithNames"
           )
           |> Plausible.S3.export_upload_multipart(s3_bucket, s3_path, s3_config_overrides(args))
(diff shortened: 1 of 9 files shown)


lib/workers/export_csv.ex
```diff
@@ -25,22 +25,25 @@ defmodule Plausible.Workers.ExportCSV do
       |> Keyword.replace!(:pool_size, 1)
       |> Ch.start_link()
 
+    # NOTE: what if 1970-01-01?
     # NOTE: should we use site.timezone?
```
zoldar, line 29: If we were to use timezone bound date, we'd have to alter all the imports to output timezone adjusted dates as well. Otherwise we risk the explicit date range being shifted relative to the contents of the export.

Existing importers fetch dates via API and these are always UTC based AFAIK. IMHO we could stick to UTC to avoid extra complexity and potential loss of data along the way. It also makes sense in the light of future option to do CSV imports from sources other than Plausible.

lib/workers/export_csv.ex
```diff
@@ -25,22 +25,25 @@ defmodule Plausible.Workers.ExportCSV do
       |> Keyword.replace!(:pool_size, 1)
       |> Ch.start_link()
 
+    # NOTE: what if 1970-01-01?
```
zoldar, line 28: In what case can it happen, other than when the dataset is empty?

lib/workers/export_csv.ex
ruslandoga: Hm, I wonder if https://github.com/plausible/analytics/pull/3899#discussion_r1527087786 should be updated then.

ruslandoga: Right now a local start date is used for imported tables as well.

lib/workers/export_csv.ex
ruslandoga: I think that's the only possibility.

ruslandoga: Should we abort the export in that case?

ruslandoga: No, we'd export an archive with empty CSV files.

zoldar: Then perhaps we should error out with "There's nothing to export" instead?

lib/workers/export_csv.ex
zoldar: Could you point me to a relevant spot in the code? 🙏

ruslandoga: https://github.com/plausible/analytics/blob/02d2256483e74f8689125e05b275ba2238f8d2c5/lib/plausible/sites.ex#L222

zoldar: Sorry, I meant a spot where we use it for imported tables.

ruslandoga: Seems like the imported data is somewhat misaligned when merged with native data in general since native data is using localized dates and imported -- UTC.

zoldar: Yeah, it's a bit of a mess. We'll need to take a closer look and revise it. For now though let's perhaps stick with UTC for imports.

ruslandoga: OK! I've removed the note about timezones in https://github.com/plausible/analytics/pull/3895/commits/91a9505594449c486fa3aecc8415b34c7790013b

lib/workers/export_csv.ex
ruslandoga: I made the job send an "error email" in https://github.com/plausible/analytics/pull/3895/commits/91a9505594449c486fa3aecc8415b34c7790013b

ruslandoga: Or should the job be `:cancel`led and the error handled in ObanErrorReporter?

zoldar (approved):

PR #3896, opened 2024-03-15
don't log migrations in tests

ruslandoga:
### Changes

When tests fail in GitHub actions, I need to scroll through the migrations output to see the failure reason. This PR tries to make seeing the failure reason easier.

I expect the data migration output to still be logged, so there is another approach with running migrations in a separate step: https://github.com/plausible/analytics/pull/3897

### Tests

### Changelog

### Documentation

### Dark mode


ruslandoga:
Closing in favor of https://github.com/plausible/analytics/pull/3897 since non-Ecto migrations still get logged.

PR #3900, opened 2024-03-15
Allow using verified routes

ruslandoga:
### Changes

This PR allows using [verified routes](https://hexdocs.pm/phoenix/routing.html#verified-routes) in liveviews, controllers, and views.

The reason for this PR is that I'd like to use verified routes in https://github.com/plausible/analytics/pull/3845 but it feels wrong to include these changes there.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/endpoint.ex
+++ b/lib/plausible_web/endpoint.ex
@@ -37,7 +37,7 @@ defmodule PlausibleWeb.Endpoint do
     at: "/",
     from: :plausible,
     gzip: false,
-    only: ~w(css js images favicon.ico robots.txt)
+    only: PlausibleWeb.static_paths()
   )
 
   on_full_build do
--- a/lib/plausible_web/live/sites.ex
+++ b/lib/plausible_web/live/sites.ex
@@ -73,7 +73,7 @@ defmodule PlausibleWeb.Live.Sites do
           You don't have any sites yet.
         </p>
         <div class="mt-4 flex sm:ml-4 sm:mt-0">
-          <a href="/sites/new" class="button">
+          <a href={~p"/sites/new"} class="button">
             + Add Website
           </a>
         </div>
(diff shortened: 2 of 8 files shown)


lib/plausible_web/live/sites.ex
```diff
@@ -73,7 +73,7 @@ defmodule PlausibleWeb.Live.Sites do
           You don't have any sites yet.
         </p>
         <div class="mt-4 flex sm:ml-4 sm:mt-0">
-          <a href="/sites/new" class="button">
+          <a href={~p"/sites/new"} class="button">
```
ruslandoga, line 76: I also tried to replace some "raw" routes I found.

lib/plausible_web/templates/layout/app.html.heex
```diff
@@ -20,7 +20,7 @@
       <%= assigns[:title] ||
         "Plausible · Simple, privacy-friendly alternative to Google Analytics" %>
     </title>
-    <link rel="stylesheet" href={Routes.static_path(@conn, "/css/app.css")} />
+    <link rel="stylesheet" href={~p"/css/app.css"} />
```
ruslandoga, line 23: And static routes.

lib/plausible_web/endpoint.ex
```diff
@@ -37,7 +37,7 @@ defmodule PlausibleWeb.Endpoint do
     at: "/",
     from: :plausible,
     gzip: false,
```
ruslandoga, line 39: Off-topic, but I wonder if we can allow serving gzipped assets?

aerosol:
Just linking https://github.com/plausible/analytics/pull/3290#issuecomment-1717391544 here for some additional context

lib/plausible_web/endpoint.ex
aerosol: hmm aren't we using gzip successor already?

image

ruslandoga: Via BunnyCDN, probably :) In self-hosted setups assets are served uncompressed.

aerosol: Ah, right. Makes sense to add something for self-hosted, we can also test on staging if CDN handles the change ok.

PR #3905, opened 2024-03-17
make robots.txt restrictive by default

ruslandoga:
### Changes

This PR makes `robots.txt` restrictive by default in self-hosted setups.

Related:
- https://github.com/plausible/analytics/discussions/1798#discussioncomment-8209550
- https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/6961830456

### Tests
- [ ] Automated tests have been added (I'll add some tests to make sure it works as expected in full build vs small build)

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file.
 - Add `CLICKHOUSE_MAX_BUFFER_SIZE_BYTES` env var which defaults to `100000` (100KB)
 - Add alternative SMTP adapter plausible/analytics#3654
 - Add `EXTRA_CONFIG_PATH` env var to specify extra Elixir config plausible/analytics#3906
+- Add restrictive `robots.txt` for self-hosted plausible/analytics#3905
 
 ### Removed
 - Removed the nested custom event property breakdown UI when filtering by a goal in Goal Conversions
--- a/lib/plausible_web/endpoint.ex
+++ b/lib/plausible_web/endpoint.ex
@@ -37,6 +37,7 @@ defmodule PlausibleWeb.Endpoint do
 
   static_paths =
     on_full_build do
+      # NOTE: The Cloud uses custom robots.txt from https://github.com/plausible/website: https://plausible.io/robots.txt
       static_paths
     else
       static_paths ++ ["robots.txt"]
--- a/lib/plausible_web/endpoint.ex
+++ b/lib/plausible_web/endpoint.ex
@@ -33,11 +33,20 @@ defmodule PlausibleWeb.Endpoint do
   plug(PlausibleWeb.Tracker)
   plug(PlausibleWeb.Favicon)
 
+  static_paths = ~w(css js images favicon.ico)
+
+  static_paths =
+    on_full_build do
+      static_paths
+    else
+      static_paths ++ ["robots.txt"]
+    end
+
   plug(Plug.Static,
     at: "/",
     from: :plausible,
     gzip: false,
-    only: ~w(css js images favicon.ico robots.txt)
+    only: static_paths
   )
 
   on_full_build do
--- a/priv/static/robots.txt
+++ b/priv/static/robots.txt
@@ -1,5 +1,2 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
-#
-# To ban all spiders from the entire site uncomment the next two lines:
-# User-agent: *
-# Disallow: /
+User-agent: *
+Disallow: /


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file.
 - Add 'browser_versions.csv' to CSV export
 - Add `CLICKHOUSE_MAX_BUFFER_SIZE_BYTES` env var which defaults to `100000` (100KB)
 - Add alternative SMTP adapter plausible/analytics#3654
+- Add restrictive `robots.txt` for in self-hosted#3905
```
ruslandoga, line 34:
```suggestion
- Add restrictive `robots.txt` for in self-hosted#3905
+ Add restrictive `robots.txt` for self-hosted plausible/analytics#3905
```

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ All notable changes to this project will be documented in this file.
...
++ Add restrictive `robots.txt` for self-hosted plausible/analytics#3905
```
ruslandoga, line 35:
```suggestion
- Add restrictive `robots.txt` for self-hosted plausible/analytics#3905
```

lib/plausible_web/endpoint.ex
```diff
@@ -33,11 +33,20 @@ defmodule PlausibleWeb.Endpoint do
   plug(PlausibleWeb.Tracker)
   plug(PlausibleWeb.Favicon)
 
+  static_paths = ~w(css js images favicon.ico)
+
+  static_paths =
+    on_full_build do
+      static_paths
```
ruslandoga, line 40: https://chat.openai.com/share/aa80093e-3f71-43af-9f48-87d7c622311d

I wonder what the solution should be for the Cloud version. I omitting robots.txt OK or should we maybe just COPY different files when building the container.

Cloud (to keep the previous behavior)
```
# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
#
# To ban all spiders from the entire site uncomment the next two lines:
# User-agent: *
# Disallow: /
 ```
 
 Self-hosted
 
```
User-agent: *
Disallow: /
```

Comment (ukutaht) [reply lib/plausible_web/endpoint.ex:40]: The current robots.txt isn't event used in prod because plausible/website serves its own one which the reverse proxy prefers: https://plausible.io/robots.txt

It should be fine to omit. But we should add a comment/doc line for some clarity. We never want CE to be indexed but in Cloud we do want pages like public dashboards and live demo to be indexed. This is controlled via X-Robots-Tag header rather than robots.txt file.

Comment (ruslandoga) [reply lib/plausible_web/endpoint.ex:40]: Then instead of omitting, would it be OK to simply use

```
User-agent: *
Disallow: /
```

for both? With a comment, of course.

Comment (ruslandoga) [reply lib/plausible_web/endpoint.ex]: Like this https://github.com/plausible/analytics/pull/3905/commits/c71c1146104e66b07e6873ec7992386c493089b2

Comment (ukutaht) [reply lib/plausible_web/endpoint.ex]: I prefer the original solution of omitting it. It's quite likely we will move off the Jekyll-based static site in the future. And in that case, if we forget about robots.txt, we might end up in a place where we disallow all indexing for plausible.io which would be bad.

I know omitting adds more code, but it feels safer to me.

Comment (ruslandoga) [reply lib/plausible_web/endpoint.ex:40]: What about copying different robots.txt into the container image based on some ARG? So that EE would get

```
# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
#
# To ban all spiders from the entire site uncomment the next two lines:
# User-agent: *
# Disallow: /
```
and CE would get
```
User-agent: *
Disallow: /
```

Comment (ruslandoga) [reply lib/plausible_web/endpoint.ex:40]: Or having it configured with app env and served by a custom plug.

```elixir
# in endpoint.ex
plug :serve_robots_txt

on_full_build do
# ...
else
def robots_txt do
  """
  User-agent: *
  Disallow: /
  """
end
end

def serve_robots_txt(conn, _opts) do
  case conn.path_info do
    ["robotx.txt"] -> conn |> send_resp(200, robots_txt()) |> halt()
    _other -> conn
  end
end
```

ukutaht: I think omitting, the way you've done, is good enough for now.

ruslandoga: Ok! I've added a note about the Cloud's robots.txt in https://github.com/plausible/analytics/pull/3905/commits/11409f3f0f902ac7df996a25425764587e80c311

zoldar (approved):

PR #3908, opened 2024-03-17
Redact search params and fragment from referrer

Zwyx:
#### Redact search params and fragment from referrer

In the referrer URL, the search parameters (following the `?`) and the fragment (following the `#`) might contain personal/sensitive information.

They are already dropped by Plausible server, but I think we could even make them never leave the user's device.

It could be quite frightening for a user observing network requests in their browser's dev tools, to see that sensitive information is sent to Plausible, a third party service they might have never heard of.

Observing that the search params and fragment are not being sent would be reassuring.

> Note: apologises for dropping a PR here before asking in an issue — the change is so trivial that I preferred to make it directly. Feel free to simply reject it.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -74,7 +74,7 @@
     payload.u = location.href
     {{/if}}
     payload.d = scriptEl.getAttribute('data-domain')
-    payload.r = document.referrer || null
+    payload.r = document.referrer?.split("?")[0].split("#")[0] || null
     if (options && options.meta) {
       payload.m = JSON.stringify(options.meta)
     }


Zwyx:
Hi @ukutaht, any chance I could get an explanation about the closing of this issue?

ukutaht:
Hi, was just closing old PRs with no activity. Sorry for no explanation.

I'm not against the proposal in principle. However, it didn't look attractive to me from our 'trying to stay as lightweight as possible' perspective. Referrers are already redacted to just the hostname in the vast majority of requests by `Referrer-Policy`. So my intuition is that this code would only take effect in very rare instances. Sending the scrubbing code to every client will increase the amount of data being sent over the wire instead of scrubbing on the server in the rare cases that query param or hash is sent.

But as I said the principle is sound, it's not necessary to send this information to our server.

> (Even with strict-origin-when-cross-origin, search parameters and fragment will be present in the referrer if, for instance, a user middle-clicks a link in the page.)

In this case do you mean internal links and with Plausible running on the same origin (i.e. `analytics.yourhostname.com`)?

Zwyx:
Thanks for replying.

> In this case do you mean internal links and with Plausible running on the same origin (i.e. `analytics.yourhostname.com`)?

Yes for the same origin, but not necessarily internal links I think. I was working on the app [github.com/Zwyx/library-of-babel](https://github.com/Zwyx/library-of-babel), which provides end-to-end encryption, and so I [modified Plausible's script](https://github.com/Zwyx/library-of-babel/blob/main/src/lib/plausible.ts) to ensure IDs and encryption keys never leave the device.

PR #3910, opened 2024-03-17
fix manual embed width

ruslandoga:
### Changes

This PR fixes https://github.com/plausible/docs/issues/470

Here's how it looks in `bg-red-200 mx-auto max-w-md` container
Screenshot 2024-03-18 at 07 50 30

And here's how it looks in `bg-red-200` container

Screenshot 2024-03-18 at 07 52 35

And here's how it looks on master (where `max-w-screen-xl` was applied regardless of `width=manual`)

Screenshot 2024-03-18 at 07 53 49

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/views/stats_view.ex
+++ b/lib/plausible_web/views/stats_view.ex
@@ -42,7 +42,7 @@ defmodule PlausibleWeb.StatsView do
 
   def stats_container_class(conn) do
     cond do
-      conn.assigns[:embedded] && conn.assigns[:width] == "manual" -> ""
+      conn.assigns[:embedded] && conn.params["width"] == "manual" -> "px-6"
       conn.assigns[:embedded] -> "max-w-screen-xl mx-auto px-6"
       !conn.assigns[:embedded] -> "container print:max-w-full"
     end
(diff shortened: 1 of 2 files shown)


lib/plausible_web/views/stats_view.ex
```diff
@@ -42,7 +42,7 @@ defmodule PlausibleWeb.StatsView do
 
   def stats_container_class(conn) do
     cond do
-      conn.assigns[:embedded] && conn.assigns[:width] == "manual" -> ""
```
ruslandoga, line 45: There was no such assign. I thought about adding it but was afraid that it's used someplace else in the template. So instead I'm just checking the params.

lib/plausible_web/views/stats_view.ex
```diff
@@ -42,7 +42,7 @@ defmodule PlausibleWeb.StatsView do
...
+      conn.assigns[:embedded] && conn.params["width"] == "manual" -> "px-6"
```
ruslandoga, line 45: I'm still adding padding since it looks much nicer that way.

zoldar (approved):

PR #3918, opened 2024-03-20
Pages shield

aerosol:
### Changes

This PR enables "Pages" Shield (in case `shield_pages` feature flag is enabled), that allows filtering incoming traffic by matching the URL path names. Wildcards are allowed, that are turned into regexes. Reviewing commit-by-commit should make this a bit more digestible after familiarizing yourself with the following overview.

**Migration:** new postgres table, will be extracted to a separate PR. We keep the actual pattern separate from what we display to the users.

**Cache:** The caching layer called on ingestion responds with already compiled regular expressions to spare the needless complexity. The cache ets table had to be made a `:bag` so that all rules per domain are looked up in a single shot. As the caching variety increases, I do see the need of better (i.e. more declarative/readable) supervision tree child specs syntax, but wanted to keep it out of scope in this PR.

**UI:** The site settings area is very similar to other shield types. The UI is built with a ComboBox autosuggesting existing URLs, just like it's done e.g. in Pageview Goals. The `creatable` option is turned on, so that custom paths can be entered, including wildcards. Additionally, potentially redundant rules (when one or more rule may override another) are highlighted for awareness. Details are presented in a standard tooltip.

https://github.com/plausible/analytics/assets/173738/6f58aace-ed8f-409a-ba92-89f169eb2bb3

**Regular expressions:** there's potentially room for some ingest-phase optimizations, i.e. we can detect whether a match should be exact, or a string prefix, only falling back to regex engine when wildcard(s) are used in the middle. Agreed with @ukutaht not to prematurely optimize it for now, and keep regex matching applied uniformly.

**Restrictions:** user input must start with a slash and must be no longer than 250 characters - it's a bit arbitrary at this point, but wildcards should compensate, in case it's not enough.

**Other changes:** The `Plausible.Shields` context module now encapsulates the logic determining whether the request is dropped. I've split up the tests, so that each shield type runs its own suite.

**Request for comments:** areas that I'm still looking at for potential flaws: a) performance/exploitation/efficiency, b) cache refreshes, c) improvements to `CompiledRegex` ecto type. Would appreciate any thoughts in addition to your review comments.
 
### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated - https://github.com/plausible/docs/pull/477

### Dark mode


lib/plausible/application.ex
```diff
@@ -77,6 +77,21 @@ defmodule Plausible.Application do
          interval: :timer.seconds(35),
          warmer_fn: :refresh_updated_recently
        ]},
+      {Plausible.Shield.PageRuleCache, ttl_check_interval: false, ets_options: [:bag]},
```
aerosol, line 80: ConCache allows us to pass raw ets options

lib/plausible/shield/page_rule.ex
```diff
@@ -0,0 +1,85 @@
...
+  Ensures that the regex is compiled on load
+  """
+  use Ecto.Type
+
+  def type, do: :string
+
+  def cast(val) when is_binary(val), do: {:ok, val}
+  def cast(_), do: :error
+
+  def load(val), do: {:ok, Regex.compile!(val)}
```
aerosol, line 12: Upon cache querying the database, the regexes are compiled. They are not however when `Shields.add_page_rule` is called - I wonder if this can/should be changed?

lib/plausible/shield/page_rule.ex
```diff
@@ -0,0 +1,85 @@
...
+      "/" <> _ = page_path ->
+        regex =
+          page_path
+          |> Regex.escape()
+          |> String.replace("\\*\\*", ".*")
+          |> String.replace("\\*", ".*")
+
+        regex = "^#{regex}$"
+
+        verify_valid_regex(changeset, regex)
```
aerosol, line 69: Just making sure we never store anything that throws a compilation error

lib/plausible/shield/page_rule.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  defp store_regex(changeset) do
+    case get_field(changeset, :page_path) do
+      "/" <> _ = page_path ->
+        regex =
+          page_path
+          |> Regex.escape()
+          |> String.replace("\\*\\*", ".*")
+          |> String.replace("\\*", ".*")
```
aerosol, line 65: This is slightly different than `page_regex()` used elsewhere - no pipes are allowed.

lib/plausible/shield/page_rule.ex
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  defp verify_valid_regex(changeset, regex) do
+    case Regex.compile(regex) do
+      {:ok, _} ->
+        put_change(changeset, :page_path_pattern, regex)
+
+      {:error, _} ->
+        add_error(changeset, :page_path, "could not compile regular expression")
```
aerosol, line 82: Defensive. I can't reproduce it to cover with test.

lib/plausible/shields/country_test.exs
```diff
@@ -0,0 +1,136 @@
+defmodule Plausible.Shields.CountryTest do
```
aerosol, line 1: extracted from `ShieldsTest`

lib/plausible/shields/ip_test.exs
```diff
@@ -0,0 +1,147 @@
+defmodule Plausible.Shields.IPTest do
```
aerosol, line 1: extracted from `ShieldsTest`

lib/plausible/shields.ex
```diff
@@ -13,11 +13,66 @@ defmodule Plausible.Shields do
...
+        false
+    end
+  end
+
+  @spec page_blocked?(Site.t() | String.t(), String.t()) :: boolean()
+  def page_blocked?(%Site{domain: domain}, address) do
+    page_blocked?(domain, address)
+  end
+
+  def page_blocked?(domain, pathname) when is_binary(domain) and is_binary(pathname) do
```
aerosol, line 44: This bothers me because we'll always try all the rules against each event. Whether it's bad/should be avoided I have not proven yet.

lib/plausible/shields.ex
```diff
@@ -13,11 +13,66 @@ defmodule Plausible.Shields do
...
+
+    if page_rules do
+      found? =
+        page_rules
+        |> List.wrap()
+        |> Enum.find(fn rule ->
+          rule.action == :deny and Regex.match?(rule.page_path_pattern, pathname)
+        end)
+
+      !!found?
```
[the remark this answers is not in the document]
aerosol: thanks!

test/plausible/shield/page_rule_cache_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      site1 = insert(:site)
+      site2 = insert(:site)
+
+      {:ok, %{id: rid1}} = Shields.add_page_rule(site1, %{"page_path" => "/test/1"})
+      {:ok, %{id: rid2}} = Shields.add_page_rule(site1, %{"page_path" => "/test/2"})
+      {:ok, %{id: rid3}} = Shields.add_page_rule(site2, %{"page_path" => "/test/2"})
+
+      :ok = PageRuleCache.refresh_all(cache_name: test)
+
+      :ok = Shields.remove_page_rule(site1, rid1)
```
ruslandoga, line 31: Is this to show that cache is out-of-sync with the DB? A comment might help understand what it's doing in this test for clueless people like me :)

lib/plausible_web/live/shields/page_rules.ex
```diff
@@ -0,0 +1,304 @@
...
+
+        # Make sure to clear the combobox input after adding a page rule, on subsequent modal reopening
+        send_update(PlausibleWeb.Live.Components.ComboBox,
+          id: "page_rule_page_code",
+          display_value: ""
+        )
+
+        send_flash(
+          :success,
+          "page rule added successfully. Traffic will be rejected within a few minutes."
```
ruslandoga, line 222:
```suggestion
          "Page rule added successfully. Traffic will be rejected within a few minutes."
```

lib/plausible_web/live/shields/page_rules.ex
```diff
@@ -0,0 +1,304 @@
...
+        {:noreply, assign(socket, form: to_form(changeset))}
+    end
+  end
+
+  def handle_event("remove-page-rule", %{"rule-id" => rule_id}, socket) do
+    Shields.remove_page_rule(socket.assigns.site.id, rule_id)
+
+    send_flash(
+      :success,
+      "page rule removed successfully. Traffic will be resumed within a few minutes."
```
ruslandoga, line 237:
```suggestion
      "Page rule removed successfully. Traffic will be resumed within a few minutes."
```

ruslandoga (approved):

ruslandoga:
I can run some benchmarks with [k6](https://github.com/plausible/analytics/tree/master/test/load) and flame graphs tomorrow if needed :)

test/plausible/shield/page_rule_cache_test.exs
aerosol: Yeah exactly, to make sure the cache is actually running and not falling back to the raw db lookups as it does with all the other tests. I'll add a comment

aerosol:
> I can run some benchmarks with [k6](https://github.com/plausible/analytics/tree/master/test/load) and flame graphs tomorrow if needed :) I guess the scenarios would be: no page rules, one page rule matching, one page rule not matching, many page rules with early match, many page rules with "late" match, many page rules with no matches.

Aw TIL, that would be lovely, appreciate the help. The scenarios make sense. Normally it takes a while for the cache to refresh after adding new rules -- best to call `Plausible.Shield.PageRuleCache.refresh_all` before each run. I'll try similar with wrk. I'm going to also add [metrics #add](https://github.com/plausible/analytics/pull/3927)

ukutaht (approved):
LGTM 👍

lib/plausible/shield/page_rule.ex
aerosol: Not quite at this level, because the input is already sanitized so "*foo" will become ".*foo"

lib/plausible/shield/page_rule.ex
```diff
@@ -0,0 +1,85 @@
...
+    )
+  end
+
+  defp store_regex(changeset) do
+    case get_field(changeset, :page_path) do
+      "/" <> _ = page_path ->
+        regex =
+          page_path
+          |> Regex.escape()
+          |> String.replace("\\*\\*", ".*")
```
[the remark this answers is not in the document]
aerosol: I took it from [`page_regex`](https://github.com/plausible/analytics/blob/fd879eeb165698ad2ca429f3992facc77e5969f5/lib/plausible/stats/base.ex#L481) - I thought we keep it because people tend to use `**` for some reason. I suppose `.*.*` is functionally equivalent, even if silly.

lib/plausible/shields.ex
```diff
@@ -13,11 +13,63 @@ defmodule Plausible.Shields do
...
+  end
+
+  def page_blocked?(domain, pathname) when is_binary(domain) and is_binary(pathname) do
+    page_rules = Shield.PageRuleCache.get(domain)
+
+    if page_rules do
+      page_rules
+      |> List.wrap()
+      |> Enum.find_value(false, fn rule ->
+        rule.action == :deny and Regex.match?(rule.page_path_pattern, pathname)
```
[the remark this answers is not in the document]
aerosol: Really none at the moment. This was compulsively modeled after the other rules (mindless copy pasta I suppose). I thought that at some point we might want to have "allow-lists" rather than "block-lists" (for hostnames and maybe countries according to Marko's feedback), or even let the user decide what the default list behavior is. Not sure if that'll ever be true for page paths. Do you think I should remove it and assume it's always a block-list?

ruslandoga:
I decided to start with Benchee

<details><summary>bench/event.exs</summary>

```elixir
# params = %{}

conn = fn params ->
  %Plug.Conn{
    assigns: %{},
    body_params: params,
    cookies: %{},
    halted: false,
    host: "localhost",
    method: "POST",
    params: %{},
    path_info: ["api", "event"],
    path_params: %{},
    port: 8000,
    private: %{
      :phoenix_view => %{_: PlausibleWeb.Api.ExternalView},
      :plug_session_fetch => :done,
      :plug_session => %{},
      :before_send => [],
      :phoenix_controller => PlausibleWeb.Api.ExternalController,
      :phoenix_action => :event,
      :phoenix_format => "json",
      :phoenix_endpoint => PlausibleWeb.Endpoint,
      :phoenix_router => PlausibleWeb.Router,
      :phoenix_layout => %{_: {PlausibleWeb.LayoutView, :app}},
      PlausibleWeb.Router => []
    },
    query_params: %{},
    query_string: "",
    remote_ip: {1, 1, 1, 1},
    req_cookies: %{},
    req_headers: [
      {"content-length", "128"},
      {"content-type", "application/json, text/plain"},
      {"host", "localhost:8000"},
      {"user-agent",
       "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.121 Safari/537.36 OPR/71.0.3770.284"},
      {"x-forwarded-for", "1.1.1.1"}
    ],
    request_path: "/api/event",
    resp_body: nil,
    resp_cookies: %{},
    resp_headers: [
      {"cache-control", "max-age=0, private, must-revalidate"},
      {"x-request-id", "F5jzh_7V103l2LkAAQJF"},
      {"access-control-allow-credentials", "true"},
      {"access-control-allow-origin", "*"},
      {"access-control-expose-headers", ""}
    ],
    scheme: :http,
    script_name: [],
    state: :unset,
    status: nil
  }
end

user =
  Plausible.Repo.get_by(Plausible.Auth.User, email: "user@plausible.bench") ||
    Plausible.Factory.insert(:user, email: "user@plausible.bench")

site = fn domain ->
  Plausible.Repo.get_by(Plausible.Site, domain: domain) ||
    Plausible.Factory.insert(:site,
      domain: domain,
      memberships: [
        Plausible.Factory.build(:site_membership, user: user, role: :owner)
      ]
    )
end

no_rules_site = site.("no-rules.site")
one_rule_site = site.("one-rule.site")
five_rules_site = site.("five-rules.site")
thirty_rules_site = site.("30-rules.site")

page_rules = fn site ->
  import Ecto.Query

  Plausible.Shield.PageRule
  |> where(site_id: ^site.id)
  |> where(action: :deny)
  |> select([s], s.page_path)
  |> Plausible.Repo.all()
end

if page_rules.(one_rule_site) == [] do
  {:ok, _} = Plausible.Shields.add_page_rule(one_rule_site, %{"page_path" => "/blog/*"})
end

if page_rules.(five_rules_site) == [] do
  for i <- 1..5 do
    {:ok, _} = Plausible.Shields.add_page_rule(five_rules_site, %{"page_path" => "/blog/#{i}/*"})
  end
end

if page_rules.(thirty_rules_site) == [] do
  for i <- 1..30 do
    {:ok, _} =
      Plausible.Shields.add_page_rule(thirty_rules_site, %{"page_path" => "/blog/#{i}/*"})
  end
end

Plausible.Site.Cache.refresh_all()
Plausible.Shield.CountryRuleCache.refresh_all()

alias Plausible.Ingestion

Benchee.run(
  %{
    "build and buffer" => fn params ->
      {:ok, request} = Ingestion.Request.build(conn.(params))

      case Ingestion.Event.build_and_buffer(request) do
        {:ok, %{buffered: [_], dropped: []}} -> :ok
        {:ok, %{buffered: [], dropped: [%{drop_reason: :site_page_blocklist}]}} -> :ok
      end
    end
  },
  inputs: %{
    "no rules" => %{
      "domain" => no_rules_site.domain,
      "name" => "pageview",
      "props" => "{}",
      "referrer" => nil,
      "revenue" => nil,
      "url" => "http://#{no_rules_site.domain}/blog/post"
    },
    "one rule (match)" => %{
      "domain" => one_rule_site.domain,
      "name" => "pageview",
      "props" => "{}",
      "referrer" => nil,
      "revenue" => nil,
      "url" => "http://#{one_rule_site.domain}/blog/post"
    },
    "one rule (no match)" => %{
      "domain" => one_rule_site.domain,
      "name" => "pageview",
      "props" => "{}",
      "referrer" => nil,
      "revenue" => nil,
      "url" => "http://#{one_rule_site.domain}/blogging/now"
    },
    "five rules (no match)" => %{
      "domain" => five_rules_site.domain,
      "name" => "pageview",
      "props" => "{}",
      "referrer" => nil,
      "revenue" => nil,
      "url" => "http://#{five_rules_site.domain}/blogging/now"
    },
    "30 rules (no match)" => %{
      "domain" => thirty_rules_site.domain,
      "name" => "pageview",
      "props" => "{}",
      "referrer" => nil,
      "revenue" => nil,
      "url" => "http://#{thirty_rules_site.domain}/blogging/now"
    }
  }
)
```

</details>

<details><summary>results</summary>

```
Operating System: macOS
CPU Information: Apple M2
Number of Available Cores: 8
Available memory: 8 GB
Elixir 1.16.0
Erlang 26.2.1
JIT enabled: true

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: 30 rules (no match), five rules (no match), no rules, one rule (match), one rule (no match)
Estimated total run time: 35 s

Benchmarking build and buffer with input 30 rules (no match) ...
Benchmarking build and buffer with input five rules (no match) ...
Benchmarking build and buffer with input no rules ...
Benchmarking build and buffer with input one rule (match) ...
Benchmarking build and buffer with input one rule (no match) ...
Calculating statistics...
Formatting results...

##### With input 30 rules (no match) #####
Name                       ips        average  deviation         median         99th %
build and buffer       13.89 K       72.00 μs    ±25.19%       70.75 μs       96.21 μs

##### With input five rules (no match) #####
Name                       ips        average  deviation         median         99th %
build and buffer       17.95 K       55.71 μs    ±13.03%       54.58 μs       80.64 μs

##### With input no rules #####
Name                       ips        average  deviation         median         99th %
build and buffer       19.40 K       51.55 μs    ±13.34%       50.42 μs       73.75 μs

##### With input one rule (match) #####
Name                       ips        average  deviation         median         99th %
build and buffer       77.20 K       12.95 μs    ±32.15%       12.38 μs       25.83 μs

##### With input one rule (no match) #####
Name                       ips        average  deviation         median         99th %
build and buffer       18.80 K       53.19 μs    ±14.44%          52 μs       79.71 μs
```

</details>

lib/plausible/shields.ex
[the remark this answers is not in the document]
ukutaht: > I thought that at some point we might want to have "allow-lists" rather than "block-lists"

In the future I can see allowing the user to flip between allowlist vs blocklist for countries and hostnames, yes. For now it's not a concern, each list is fixed in terms of default behaviour (block for everything except hostnames which is allowlist)

> Do you think I should remove it and assume it's always a block-list?

Let's keep it as is

ruslandoga:
Benchmarking with k6 and profiling random requests seems to show that the slowest action now is ConCache's locked update.

- https://copycat.fun/send_pageview_no_rules_site is flame graph for a `mix send_pageview` request (i.e. not under load) for a site with no page rules set, the slowest part is parsing the referrer (RefInspector)
- https://copycat.fun/k6_no_rules_site same as above but while running k6 (i.e. under some load), the slowest part now is `Plausible.Session.CacheStore.persist_session/1`
- https://copycat.fun/k6_30_rules_site_unmatched same as above but uses a site with 30 page rules (none of which match), the page rule check takes less than 1% of the whole request duration

Screenshot 2024-03-23 at 16 40 39

aerosol:
Thanks for running the benchmarks @ruslandoga. 

> Plausible.Session.CacheStore.persist_session/1 (maybe a "dirty" put could be used instead?)

I think this can be done, but

> 35ms? seems to be an outlier >p(95) according to the final k6 stats

Is it 35ms or 35μs? I think the latter, in which case, not only unrelated to this PR, but also not worth any extra optimization at this point? We'll definitely have some real world data with https://github.com/plausible/analytics/pull/3927. WYT?

ruslandoga:
It's 35000 us which is 35ms.

aerosol:
> The real objective of the benchmarks was to check performance implications of PageRules, and the conclusion is PageRules are fast :) Everything else is just off-topic and something I noticed / thought was interesting. "Further research is needed" regarding salts and dirty_put.

Thank you for clarification. This makes sense, and indeed very interesting to zoom in. Also currently very easy to switch to dirty_put, in case we decide to do it. I'm going to push this live for now.

PR #3919, opened 2024-03-20
Timeseries for conversion rate

RobertJoonas:
### Changes

This PR adds the `conversion_rate` metric into the Stats API timeseries, and also makes it possible to display it on the main graph.

Keeping this as a draft for now, since it builds upon https://github.com/plausible/analytics/pull/3887

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -436,7 +436,7 @@ export default class VisitorGraph extends React.Component {
     const selectableMetrics = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
     const canSelectSavedMetric = selectableMetrics && selectableMetrics.includes(savedMetric)
 
-    if (query.filters.goal) {
+    if (query.filters.goal && savedMetric === 'visitors') {
       this.setState({ metric: 'conversions' })
     } else if (canSelectSavedMetric) {
       this.setState({ metric: savedMetric })
--- a/assets/js/dashboard/stats/graph/visitor-graph.js
+++ b/assets/js/dashboard/stats/graph/visitor-graph.js
@@ -436,7 +436,7 @@ export default class VisitorGraph extends React.Component {
     const selectableMetrics = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
     const canSelectSavedMetric = selectableMetrics && selectableMetrics.includes(savedMetric)
 
-    if (query.filters.goal && savedMetric === 'visitors') {
+    if (query.filters.goal && savedMetric !== 'conversion_rate') {
       this.setState({ metric: 'conversions' })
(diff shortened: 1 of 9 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1321,6 +1305,21 @@ defmodule PlausibleWeb.Api.StatsController do
     end
   end
 
+  def parse_and_validate_graph_metric(params, query) do
```
macobo, line 1308: defp?

assets/js/dashboard/stats/graph/visitor-graph.js
```diff
@@ -436,7 +436,7 @@ export default class VisitorGraph extends React.Component {
     const selectableMetrics = topStatData && topStatData.top_stats.map(({ name }) => METRIC_MAPPING[name]).filter(name => name)
     const canSelectSavedMetric = selectableMetrics && selectableMetrics.includes(savedMetric)
 
-    if (query.filters.goal) {
+    if (query.filters.goal && savedMetric === 'visitors') {
```
macobo, line 439: Can you clarify what this achieves?

RobertJoonas: Sure, I totally get it that this change looks pretty abstract. Should have explained it in a comment straight away...

This is how we're currently mapping the Top Stats from their display name into the actual name of the metric queried from the Timeseries API:
 
image

"Unique visitors" here maps to visitors. That name is also displayed in the goal filtered view. However, the metric really corresponds to `:total_visitors`. The metric that really represents `:visitors` is Unique conversions. But since the visitors name is taken, we are using `conversions` instead.

image

After my change, whenever there was a metric reset (after fetching new Top Stats), the metric changed to `conversions` when a goal filter was applied. Before, that made sense, since "Unique conversions" was the only graphable metric in a goal-filtered view. But not anymore. Now we should only set the metric to `conversions` when the saved metric is "visitors".

To put it in other words, now the `saved_metric` can also be `conversion_rate` in a goal filtered view. And if it is, we don't want to override it to `visitors`.

That all said, the code of the metric selection part (and the top stats <> visitor graph combination in general) is unnecessarily complex. I have some ideas on refactoring it, but I'm gonna let it slide for now, since we might be rebuilding all of it when building the ability to graph multiple metrics.

Hope this essay answers your question 🙂

zoldar (approved):

assets/js/dashboard/stats/graph/visitor-graph.js
RobertJoonas: Ha! Just realised that I made a mistake there. If no metric is saved or it's something other than `visitors` that cannot be shown in the goal view, we would currently still be setting the metric to `visitors`. Actually we want to do that in all the cases where metric is not `conversion_rate`. My bad. Fixed now in https://github.com/plausible/analytics/pull/3919/commits/4f4c30f561895a85d1e7edb4bf56644978e3465e

PR #3921, opened 2024-03-21
Include domain and dates in zip archive filename

ruslandoga:
### Changes

This PR changes how the exported Zip archive is named, from `Plausible.zip` to `#{domain}_#{start_date}_#{end_date}.zip`

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/exports.ex
+++ b/lib/plausible/exports.ex
@@ -7,6 +7,22 @@ defmodule Plausible.Exports do
   import Ecto.Query
 
   @doc """
+  Renders filename for the Zip archive containing the exported CSV files.
+
+  Examples:
+
+      iex> archive_filename("plausible.io", ~D[2021-01-01], ~D[2024-12-31])
+      "plausible.io_20210101_20241231.zip"
+
+      iex> archive_filename("Bücher.example", ~D[2021-01-01], ~D[2024-12-31])
+      "B%C3%BCcher.example_20210101_20241231.zip"
+
+  """
+  def archive_filename(domain, min_date, max_date) do
+    "#{URI.encode_www_form(domain)}_#{Calendar.strftime(min_date, "%Y%m%d")}_#{Calendar.strftime(max_date, "%Y%m%d")}.zip"
+  end
+
+  @doc """
   Builds Ecto queries to export data from `events_v2` and `sessions_v2`
   tables  into the format of `imported_*` tables for a website.
   """
--- a/lib/plausible/exports.ex
+++ b/lib/plausible/exports.ex
@@ -15,11 +15,11 @@ defmodule Plausible.Exports do
       "plausible.io_20210101_20241231.zip"
 
       iex> archive_filename("Bücher.example", ~D[2021-01-01], ~D[2024-12-31])
-      "B%C3%BCcher.example_20210101_20241231.zip"
+      "Bücher.example_20210101_20241231.zip"
 
   """
   def archive_filename(domain, min_date, max_date) do
-    "#{URI.encode_www_form(domain)}_#{Calendar.strftime(min_date, "%Y%m%d")}_#{Calendar.strftime(max_date, "%Y%m%d")}.zip"
+    "#{domain}_#{Calendar.strftime(min_date, "%Y%m%d")}_#{Calendar.strftime(max_date, "%Y%m%d")}.zip"
   end
 
   @doc """
--- a/lib/plausible/s3.ex
+++ b/lib/plausible/s3.ex
@@ -88,12 +88,21 @@ defmodule Plausible.S3 do
           :uri_string.uri_string()
   def export_upload_multipart(stream, s3_bucket, s3_path, filename, config_overrides \\ []) do
     config = ExAws.Config.new(:s3)
-    filename = String.replace(filename, "\"", "\\\"")
+
+    encoded_filename = URI.encode(filename)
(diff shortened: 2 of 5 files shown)


lib/plausible/exports.ex
```diff
@@ -6,6 +6,22 @@ defmodule Plausible.Exports do
   require Plausible
   import Ecto.Query
 
+  @doc """
+  Renders filename for the Zip archive containing the exported CSV files.
+
+  Examples:
+
+      iex> archive_filename("plausible.io", ~D[2021-01-01], ~D[2024-12-31])
+      "plausible.io_20210101_20241231.zip"
```
ruslandoga, line 15: or maybe "plausible_export_plausible.io_20210101_20241231.zip"

lib/plausible/exports.ex
```diff
@@ -6,6 +6,22 @@ defmodule Plausible.Exports do
...
+
+      iex> archive_filename("Bücher.example", ~D[2021-01-01], ~D[2024-12-31])
+      "B%C3%BCcher.example_20210101_20241231.zip"
```
ruslandoga, line 18: I guess we can skip escaping since most filesystems probably support unicode.

ruslandoga: https://chat.openai.com/share/26fdb3bc-1301-42f9-bf86-65b90e105e04

zoldar: Yeah. At first I was a bit concerned about allowing arbitrary UTF in here (including stuff like [LTR mark](https://unicode-explorer.com/c/200E) which is often used for deception) but this is domain name which is pretty much only under owner's control, so there's really not much in the way of vector of a potential attack 😅 Do you want to drop the encoding then?

lib/workers/export_csv.ex
```diff
@@ -79,6 +88,16 @@ defmodule Plausible.Workers.ExportCSV do
...
+  defp fetch_site_domain(site_id) do
+    import Ecto.Query, only: [from: 2]
+
+    Plausible.Repo.one!(
+      from s in Plausible.Site,
+        where: [id: ^site_id],
+        select: s.domain
+    )
+  end
+
```
zoldar, line 100: Perhaps let's move it to `Plausible.Sites.get_domain_by_id!`?

lib/plausible/s3.ex
```diff
@@ -84,15 +84,16 @@ defmodule Plausible.S3 do
 
   In the current implementation the bucket always goes into the path component.
   """
-  @spec export_upload_multipart(Enumerable.t(), String.t(), Path.t(), keyword) ::
+  @spec export_upload_multipart(Enumerable.t(), String.t(), Path.t(), String.t(), keyword) ::
           :uri_string.uri_string()
-  def export_upload_multipart(stream, s3_bucket, s3_path, config_overrides \\ []) do
+  def export_upload_multipart(stream, s3_bucket, s3_path, filename, config_overrides \\ []) do
     config = ExAws.Config.new(:s3)
+    filename = String.replace(filename, "\"", "\\\"")
```
zoldar, line 91: We don't accept quotes in domain names from what I can tell. Is there any other way it might appear in the name?

ruslandoga: I'm just being defensive here. This function could potentially be used someplace else where the filename is not under our full control.

zoldar: Alright, but then perhaps we should follow the suite of what [`Phoenix.Controller.send_download`](https://hexdocs.pm/phoenix/Phoenix.Controller.html#send_download/3) proposes? Stripping also newlines, path separators etc.?

See also:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename_2
- https://elixirforum.com/t/file-download-from-browser-with-given-filename/20855

ruslandoga: > Stripping also newlines, path separators etc.?
Do you mean `URI.encode`? 

https://github.com/phoenixframework/phoenix/blob/222f684fcff76c2835823c61c68acc5d2aa25bc2/lib/phoenix/controller.ex#L1277

zoldar: That would be probably the safer choice, yeah. AFAIK the downloaded file name should still contain UTF chars then, right?

lib/plausible/sites.ex
```diff
@@ -23,6 +23,14 @@ defmodule Plausible.Sites do
     Repo.get_by!(Site, domain: domain)
   end
 
+  def get_domain!(site_id) do
```
ruslandoga, line 26: I don't think it makes much sense however since it's not that generic. Only ExportCSV needs it right now.

lib/plausible/s3.ex
```diff
@@ -88,12 +88,21 @@ defmodule Plausible.S3 do
...
+        disposition <> "; filename*=utf-8''#{encoded_filename}"
+      else
+        disposition
+      end
 
     # 5 MiB is the smallest chunk size AWS S3 supports
     chunk_into_parts(stream, 5 * 1024 * 1024)
     |> ExAws.S3.upload(s3_bucket, s3_path,
-      content_disposition: ~s|attachment; filename="#{filename}"|,
+      content_disposition: disposition,
```
ruslandoga, line 105: I copied the content disposition construction logic from Phoenix.

zoldar (approved):

PR #3924, opened 2024-03-21
Re-apply Move conversion_rate logic from elixir to clickhouse

macobo:
Re-applies https://github.com/plausible/analytics/pull/3887

There was a bug when filtering by session attributes due to changed column ordering. See last commit for the reproducing test + fix.


ukutaht (approved):

zoldar (approved):
LGTM. Something for later, if at all, but perhaps we could consider converting queries to always use named joins to avoid similar issues in the future? 🤔

PR #3925, opened 2024-03-21
Fix issue with query name clash

macobo:
Unexpectedly, table.name caused a name clash after CR refactor, so using a unique name for the output column

Sentry issue: https://sentry.plausible.io/organizations/sentry/issues/5612


zoldar (approved):

RobertJoonas (approved):
Yeah, interesting why this became evident after your changes. But the fix looks solid, I actually prefer the `breakdown_prop_value` for the select name there.

PR #3926, opened 2024-03-21
Add `label` column to `site_imports` schema

zoldar:
### Changes

Migration that has to be run before releasing https://github.com/plausible/analytics/pull/3914

### DIFF
--- a/priv/repo/migrations/20240319094940_add_label_to_site_imports.exs
+++ b/priv/repo/migrations/20240319094940_add_label_to_site_imports.exs
@@ -1,0 +1,9 @@
+defmodule Plausible.Repo.Migrations.AddLabelToSiteImports do
+  use Ecto.Migration
+
+  def change do
+    alter table(:site_imports) do
+      add :label, :string
+    end
+  end
+end


priv/repo/migrations/20240319094940_add_label_to_site_imports.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.Repo.Migrations.AddLabelToSiteImports do
+  use Ecto.Migration
+
+  def change do
+    alter table(:site_imports) do
+      add :label, :string
```
macobo, line 6: Can this be null, should this default to empty string?

zoldar: Yeah, it's nullable on purpose and accounted for when building the final import label https://github.com/plausible/analytics/pull/3914/files#diff-d1158a106c4d2d096d9b8ed4b852ba9c17d293f4eb5aabb34b2d91a41f76bd7eR93-R94

macobo (approved):

PR #3933, opened 2024-03-22
Ignore sessions without entry/exit pages when breaking down entry/exit pages

macobo:
### Changes

Soft follow-up to https://github.com/plausible/analytics/pull/3870

We are currently displaying blank entry/exit pages in the dashboard. After this PR we won't. 

Basecamp: https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/7162713530


test/plausible_web/controllers/CSVs/30d-filter-goal/entry_pages.csv
```diff
@@ -1,2 +1 @@
 name,conversions,conversion_rate
-,1,100.0
```
ukutaht, line 2: I think we still want to test entry page behaviour with goal filters here. I haven't looked but guessing the test setup is just creating a custom event and therefore entry page does not get recorded. Would be more interesting to make this a more realistic example like 
1. Pageview on /
2. Custom event

And make sure the `/` entry page is recorded.

Same goes for the other csv tests.

ukutaht (approved):
LGMT, just one comment about testing CSV exports.

I know it's a bit weird to request changes but approve the PR. I just want to avoid another review cycle cause it's such a small comment/change.

test/plausible_web/controllers/CSVs/30d-filter-goal/entry_pages.csv
macobo: Agreed - will update the test rig!

PR #3936, opened 2024-03-24
Refactor VisitorGraph

RobertJoonas:
## Changes

This PR was planned to be a refactor preparing for displaying multiple metrics on the graph, which we've decided to push back in the priority queue. However, I had already made some valuable refactoring progress, so I decided to finish that.

This should make the VisitorGraph more readable, and easier to work with in the future.

I wasn't planning on changing any behaviour, but...

### Bugfix

There was one bug which the refactor fixed naturally. The interval picker currently relies on the graph data to exist in order to display the "currently selected interval", causing flickers whenever new graph data is fetched.

**Before** (notice how interval picker displays "Month" for a moment):

https://github.com/plausible/analytics/assets/56999674/cafe372e-ea9c-4af2-985f-f0c7d40fa32b

Same behaviour can be seen when selecting a different metric for the graph.

**After**:

https://github.com/plausible/analytics/assets/56999674/366e16d7-8f64-4f1f-8707-f2e89e4f5199

### Stored graph metric

Another slight behavioural change is that whenever the stored metric is unavailable (e.g. `pageviews` is stored but we're filtering by a goal), we're going to overwrite the stored metric with the first available metric (in this case `visitors`(=conversions)).

Before, `pageviews` would have only been updated to `visitors` in the internal state, and for example, after returning to the regular view, pageviews would still be displayed on the graph. After this change, the default graph metric will always change to `visitors` after having visited a goal-filtered view.

IMO, it should be really minor - I  don't think too many users will notice any difference. Getting rid of the `metric` in the React state simplifies things in the code by a great deal though. However, should it turn out to be a problem for the users, I think the best UX would be to start saving the metric per goal-filtered view separately - e.g:

* `metric__dummy.site = pageviews`
* `metric__goalFilter__dummy.site = conversion_rate`

And maybe even create a separate entry for realtime / page filter mode too.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (changes requested):
Overall the code is much cleaner and I like the refactorings. Looking really neat! 👌

There is one problem with this that prevents me from approving it. On current master, the requests for top stats and main graph are fired in parallel as you can see from this timeline below:

Screenshot 2024-03-28 at 12 49 46

With your changes, during the initial dashboard load, the main graph request can only be initiated once top stats has finished so they are run in sequence:

Screenshot 2024-03-28 at 12 50 18


In the real world, this roughly doubles the amount of time it takes for the graph to load. Given that most of our customers are in the US and our servers in Germany, I believe this needs to be avoided.

One way I can think to avoid it is for the front-end to know which metrics can be graphed or not, and use that to replace the current `graph_metric?` machinery.

RobertJoonas:
That's a good catch @ukutaht, you're absolutely right!

I was so hung up on the idea of synchronizing the requests because I was sure that the backend alone should decide whether revenue metrics are returned or not...

Our current frontend is totally unaware of whether we're in a state where revenue metrics can be graphed - it solely depends on whether they are returned by Top Stats or not. For that reason, the current behaviour on master is this:

https://github.com/plausible/analytics/assets/56999674/b8cdc875-a6fe-47bf-bd7b-d964248377ef

We can select the revenue metric to be displayed on the graph, but as soon as we refresh the page or modify the query, the graph metric will return to default.

I've now made the graph and top stats requests async again, and I've also made the frontend aware of when revenue metrics can be graphed. For that, it was necessary to pass all revenue metrics and their currencies into the `site` variable in the React dashboard.

With these changes, a revenue metric can be persisted on the graph throughout query change - as long as the following query state supports revenue metric graphing as well:

https://github.com/plausible/analytics/assets/56999674/6246473d-5dea-4abb-80cf-20e56088d986

Also, I've still decided to keep the `graph_metric` fields returned by Top Stats - they allow us to avoid mapping keys like "Unique visitors (last 30min)" into actual metrics. We can use the new frontend logic for getting the graphable metrics though (to avoid relying on two sources of truth about which metrics can be graphed).

ruslandoga:
CI seems to fail after this PR.

aerosol:
Fixing via #3973 @ruslandoga

PR #3937, opened 2024-03-25
"on-the-fly" CSV export

ruslandoga:
### Changes

This PR adds support for "on-the-fly" (streaming Zip archive chunks to Plug.Conn) CSV exports for community-edition builds. The benefit of this approach is that it doesn't rely on file system and provides a "natural" and immediate progress indicator. The downside is that errors cannot be shown since due to chunked response, the connection is simply closed in case of an error and the user is left with an invalid (incomplete) Zip archive. And https://github.com/ninenines/cowboy/issues/1278 (the connection might be closed while we are still streaming, unlikely but something to keep in mind)

Despite the downsides, I'd first try this approach and change to something else only after/if someone complains :)

Alternative method: https://github.com/plausible/analytics/pull/3938 (writing the chunks to tmp and then send_file)
Alternative method: https://github.com/plausible/analytics/pull/3944 (writing the chunks to tmp in Oban and then email a download link)

Depends on #3845

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/site/settings_imports_exports.html.heex
+++ b/lib/plausible_web/templates/site/settings_imports_exports.html.heex
@@ -20,9 +20,9 @@
     </PlausibleWeb.Components.Generic.button_link>
 
     <PlausibleWeb.Components.Generic.button_link
-      class="w-36 h-20 opacity-40 cursor-not-allowed"
+      class="w-36 h-20"
       theme="bright"
-      href=""
+      href={"/#{URI.encode_www_form(@site.domain)}/settings/import"}
     >
       <img class="h-16" src="/images/icon/csv_logo.svg" alt="New CSV import" />
     </PlausibleWeb.Components.Generic.button_link>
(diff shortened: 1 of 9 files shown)


lib/plausible_web/templates/site/settings_imports_exports.html.heex
```diff
@@ -42,12 +42,18 @@
...
 
   <div class="mt-4">
-    <PlausibleWeb.Components.Generic.button
-      data-method="post"
-      data-to={"/#{URI.encode_www_form(@site.domain)}/settings/export"}
-      data-csrf={Plug.CSRFProtection.get_csrf_token()}
-    >
-      Export to CSV
-    </PlausibleWeb.Components.Generic.button>
+    <%= on_full_build do %>
```
ruslandoga, line 45: - [ ] TODO check that this macro works in templates

ruslandoga:
Closing in favor of https://github.com/plausible/analytics/pull/3944

PR #3938, opened 2024-03-25
"temp file" CSV export

ruslandoga:
### Changes

This PR adds support for "temp file" (streaming Zip archive chunks to a temp file which is then `sendfile`d) CSV exports for community-edition builds. The benefit of this approach is that the export archive is sent only after it's completely and successfully written to disk. This allows us to show errors in case of failures. The downside is that we need to add some progress indicator that the work is being performed.

Alternative method: https://github.com/plausible/analytics/pull/3937 (sending Zip chunks to conn "on the fly")
Alternative method: https://github.com/plausible/analytics/pull/3944 (writing the chunks to tmp in Oban and then email a download link)

Depends on https://github.com/plausible/analytics/pull/3845

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/site/settings_imports_exports.html.heex
+++ b/lib/plausible_web/templates/site/settings_imports_exports.html.heex
@@ -20,9 +20,9 @@
     </PlausibleWeb.Components.Generic.button_link>
 
     <PlausibleWeb.Components.Generic.button_link
-      class="w-36 h-20 opacity-40 cursor-not-allowed"
+      class="w-36 h-20"
       theme="bright"
-      href=""
+      href={"/#{URI.encode_www_form(@site.domain)}/settings/import"}
     >
       <img class="h-16" src="/images/icon/csv_logo.svg" alt="New CSV import" />
     </PlausibleWeb.Components.Generic.button_link>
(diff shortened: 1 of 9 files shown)


lib/plausible_web/templates/site/settings_imports_exports.html.heex
```diff
@@ -42,12 +42,18 @@
...
+        data-method="post"
+        data-to={"/#{URI.encode_www_form(@site.domain)}/settings/export"}
+        data-csrf={Plug.CSRFProtection.get_csrf_token()}
+      >
+        Export to CSV
+      </PlausibleWeb.Components.Generic.button>
+    <% else %>
+      <PlausibleWeb.Components.Generic.button_link href={"/#{URI.encode_www_form(@site.domain)}/settings/export"}>
+        Export to CSV
+      </PlausibleWeb.Components.Generic.button_link>
```
ruslandoga, line 56: Not sure how to add a spinner while the file is loading. The cookie approach on the main dashboard doesn't seem to work reliably for me (Mac, Safari) and keeps spinning even after the file has been downloaded. I'll leave this for later since maybe the other, https://github.com/plausible/analytics/pull/3937, PR would be preferred.

ruslandoga:
Closing in favor of https://github.com/plausible/analytics/pull/3944

PR #3942, opened 2024-03-25
Rename internal build symbols

aerosol:
This drops the small/full build naming in favour of ce/ee abbreviations for Community and Enterprise Edition respectively.

### Tests

### Changelog

### Documentation

### Dark mode


.github/workflows/elixir.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        mix_env: ['test', 'small_test']
```
ruslandoga, line 22: Sorry for unsolicited review, but seems like config/small_test.exs and other config filenames need to be updated as well.

aerosol: oh right, not unsolicited - highly appreciated

aerosol: still, something is choking on the CI end

ruslandoga: I'm not seeing `config/*.exs` changes other than runtime.exs being modified. CI fails with this error:

```
** (File.Error) could not read file "/home/runner/work/analytics/analytics/config/ce_test.exs": no such file or directory
      (elixir 1.16.0) lib/file.ex:384: File.read!/1
      (elixir 1.16.0) lib/config.ex:275: Config.__import__!/1
      /home/runner/work/analytics/analytics/config/config.exs:80: (file)
      (stdlib 5.2) erl_eval.erl:750: :erl_eval.do_apply/7
      (stdlib 5.2) erl_eval.erl:136: :erl_eval.exprs/6
```

lib/plausible.ex
```diff
@@ -12,16 +12,16 @@ defmodule Plausible do
     end
   end
 
-  defmacro on_full_build(clauses) do
-    do_on_full_build(clauses)
+  defmacro on_ee(clauses) do
```
macobo, line 15: How would you feel like if we named this `on_enterprise_edition` and other functions followed the same form?

I personally find that abbreviations can confuse more than help reading code - my first thought for ee is always "electrical engineering".

ruslandoga: FWIW I agree, and also a macro like `on_community_edition` or `in_community_edition` would be useful for me in https://github.com/plausible/analytics/pull/3944 and some other self-hosting-oriented features.

aerosol: Thanks for the suggestions! I haven't put much thought into it, was following the [original request](https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7077291662). 

I like 'in_' prefix more than 'on_' since we're now referring to edition names.

Do you also mean to expand the MIX_ENV names/CI vars?

ruslandoga: I'm not sure... `MIX_ENV=ce_test mix test` is faster to type than `MIX_ENV=comminuty_edition_test mix test` but is not consistent with the macro.

ukutaht (approved):
Sorry for the delay on the review, got a bunch of conflicts now 😬 Looks good though

lib/plausible.ex
aerosol: > Sorry for the delay on the review, got a bunch of conflicts now 😬 Looks good though

@ukutaht no worries, I'll rebase. Opinion on this thread please? ^^

ukutaht: I think the short version is fine. I get that the shorthand can be confusing if not familiar. But the long version is very long to type out and to use everywhere. Even in public comms we will mostly use EE and CE to differentiate the two. I think the abbreviations are quite common in open-core systems when dealing with these concepts regularly. Example from Gitlab:

Screenshot 2024-04-23 at 11 59 05

PR #3944, opened 2024-03-25
Local CSV exports/imports and S3/UI updates

ruslandoga:
### Changes

This PR adds support for Oban CSV exports/imports for community-edition builds. The archives are written to a temp location first and then moved to a more permanent location defined by `PERSISTENT_CACHE_DIR` env var. An email is sent with a link to download the export when it's ready. The export can also be downloaded and deleted on the settings page.

Imports are written to a temp location on upload and moved to a more permanent location defined by `PERSISTENT_CACHE_DIR` env var on "Confirm import". Everything else is the same as in the S3 edition.

Trying it out:
```console
-- cloud, s3
$ make minio
$ mix phx.server
-- ce, local
$ make minio-stop
$ MIX_ENV=small_dev mix phx.server
```

~~Demo: https://pub-231239eb04964f3e89c831feb238275a.r2.dev/Screen%20Recording%202024-03-27%20at%2011.52.02.mp4 (couldn't upload it to GitHub due to file size limitation)~~

~~Alternative method: https://github.com/plausible/analytics/pull/3937 (sending Zip chunks to browser "on the fly")~~
~~Alternative method: https://github.com/plausible/analytics/pull/3938 (writing Zip chunks to tmp and then sendfile to browser)~~

This PR also updates the existing S3 exports according to the suggestions in https://3.basecamp.com/5308029/buckets/35871979/card_tables/cards/6945042601

**TODOs:**
- [x] local dates (in site.timezone) in exported CSVs
- [x] clean (local) imported files
- [x] there can be zero or one exports, user needs to delete any existing export before creating a new one
- [x] use Robert's UI
- [x] return emails

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/site_controller.ex
```diff
@@ -708,17 +710,57 @@ defmodule PlausibleWeb.SiteController do
...
+
+      site = conn.assigns.site
+      exports_dir = Exports.local_dir(site.id)
+      export_path = Path.join(exports_dir, file)
+
+      if File.exists?(export_path) do
+        export_content_disposition =
+          Exports.content_disposition(
+            String.replace(site.domain, ".", "_") <> "-" <> file <> ".zip"
+          )
```
ruslandoga, line 746: The downloaded filename is a bit of WIP. Not sure what it should be yet.

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -708,17 +710,57 @@ defmodule PlausibleWeb.SiteController do
...
+        })
+      )
+    end
+  else
+    defp schedule_csv_export(site, user) do
+      Oban.insert!(
+        Plausible.Workers.ExportCSVLocal.new(%{
+          "site_id" => site.id,
+          "email_to" => user.email,
+          "local_path" => Plausible.Exports.random_file(site.id)
```
ruslandoga, line 730: I'm thinking about running the exportable date range query before scheduling the job to know the filename (with date range) before hand.

lib/plausible_web/live/local_exports.ex
```diff
@@ -0,0 +1,180 @@
...
+
+    {:ok, socket}
+  end
+
+  defp list_local_exports(socket) do
+    assign(socket, local_exports: Exports.list_local_exports(socket.assigns.site_id))
+  end
+
+  @impl true
+  def render(assigns) do
```
ruslandoga, line 30: I'm wondering what should be shown in the UI.

- datetime the job was started / will be started?
- date range for the export?

ruslandoga: (there is not much styling right now, to be discussed!)

mix.lock
```diff
@@ -64,7 +64,7 @@
...
-  "heroicons": {:hex, :heroicons, "0.5.3", "ee8ae8335303df3b18f2cc07f46e1cb6e761ba4cf2c901623fbe9a28c0bc51dd", [:mix], [{:castore, ">= 0.0.0", [hex: :castore, repo: "hexpm", optional: false]}, {:phoenix_live_view, ">= 0.18.2", [hex: :phoenix_live_view, repo: "hexpm", optional: false]}], "hexpm", "a210037e8a09ac17e2a0a0779d729e89c821c944434c3baa7edfc1f5b32f3502"},
+  "heroicons": {:hex, :heroicons, "0.5.5", "c2bcb05a90f010df246a5a2a2b54cac15483b5de137b2ef0bead77fcdf06e21a", [:mix], [{:castore, ">= 0.0.0", [hex: :castore, repo: "hexpm", optional: false]}, {:phoenix_live_view, ">= 0.18.2", [hex: :phoenix_live_view, repo: "hexpm", optional: false]}], "hexpm", "2f4bf929440fecd5191ba9f40e5009b0f75dc993d765c0e4d068fcb7026d6da1"},
```
ruslandoga, line 67: it adds `micro` support https://heroicons.com/micro

lib/workers/export_csv_local.ex
```diff
@@ -0,0 +1,120 @@
+defmodule Plausible.Workers.ExportCSVLocal do
+  @moduledoc """
+  Worker for running local CSV export jobs.
+  """
```
ruslandoga, line 4: I'm going to move this logic into `ExportCSV` and the storage would be identified by `args["storage"] :: "s3" |  "local"`

lib/plausible_web/live/local_exports.ex
```diff
@@ -0,0 +1,180 @@
+defmodule PlausibleWeb.Live.LocalExports do
+  @moduledoc """
+  LiveView allowing listing and deleting local (i.e. on file system) exports.
+  """
```
ruslandoga, line 4: I'm going to add a similar liveview (or extend this one) for Cloud/S3 in a separate PR.

config/runtime.exs
```diff
@@ -543,14 +544,24 @@ base_queues = [
   s3_csv_export: 1
 ]
 
+selfhost_queues = [
+  local_csv_export: 1
```
ruslandoga, line 548: I'm thinking about using the same queue for cloud and local exports (:csv_export) but differentiating the "logic" based on args["storage"]

lib/plausible_web/controllers/site_controller.ex
ruslandoga: https://github.com/plausible/analytics/pull/3958

lib/plausible_web/controllers/site_controller.ex
ruslandoga: Done in https://github.com/plausible/analytics/pull/3958

lib/workers/export_csv_local.ex
ruslandoga: Done in https://github.com/plausible/analytics/pull/3958

lib/plausible/s3.ex
```diff
@@ -79,34 +79,32 @@ defmodule Plausible.S3 do
...
-        disposition <> "; filename*=utf-8''#{encoded_filename}"
-      else
-        disposition
-      end
-
     # 5 MiB is the smallest chunk size AWS S3 supports
     chunk_into_parts(stream, 5 * 1024 * 1024)
     |> ExAws.S3.upload(s3_bucket, s3_path,
-      content_disposition: disposition,
+      content_disposition: Plausible.Exports.content_disposition(filename),
```
ruslandoga, line 90: - [ ] if current provider doesn't support it, no need to supply it either

config/runtime.exs
```diff
@@ -710,6 +710,9 @@ if not is_selfhost do
   config :plausible, Plausible.Site, default_ingest_threshold: site_default_ingest_threshold
 end
 
+exports_storage = if is_selfhost, do: "local", else: "s3"
+config :plausible, Plausible.Exports, storage: exports_storage
```
ruslandoga, line 714: - [ ] replace with a bunch of `on_full_build` macros everywhere

config/runtime.exs
```diff
@@ -537,10 +538,9 @@ base_queues = [
   site_setup_emails: 1,
   clean_invitations: 1,
   analytics_imports: 1,
+  analytics_exports: 1,
```
ruslandoga, line 541: I renamed s3_csv_export -> analytics_exports

config/runtime.exs
```diff
@@ -296,7 +296,8 @@ config :plausible,
   is_selfhost: is_selfhost,
   custom_script_name: custom_script_name,
   log_failed_login_attempts: log_failed_login_attempts,
-  license_key: license_key
+  license_key: license_key,
+  persistent_cache_dir: persistent_cache_dir
```
ruslandoga, line 300: Saving into `:plausible` config to namespace plausible-exports and plausible-imports there, used only in local storage exports and imports (i.e. community-edition)

ruslandoga: maybe a better named env variable could be used? instead of `PERSISTENT_CACHE_DIR`, could it be `EXPORTS_DIR` or `IMPORT_DIR` or `STORAGE_DIR`

lib/plausible_web/email.ex
```diff
@@ -346,6 +347,48 @@ defmodule PlausibleWeb.Email do
...
+
+    expires_in =
+      if expires_at do
+        Timex.Format.DateTime.Formatters.Relative.format!(
+          expires_at,
+          "{relative}"
+        )
+      end
+
+    priority_email()
```
ruslandoga, line 366: Both base email templates, priority and not, include "Plausible Team" in the footer.

test/test_helper.exs
```diff
@@ -13,6 +13,11 @@ end
 
 Ecto.Adapters.SQL.Sandbox.mode(Plausible.Repo, :manual)
 
+# warn about minio if it's included in tests but not running
+if :minio in Keyword.fetch!(ExUnit.configuration(), :include) do
+  Plausible.TestUtils.ensure_minio()
+end
```
ruslandoga, line 19: Screenshot 2024-04-04 at 21 05 35

test/plausible_web/controllers/site_controller_test.exs
```diff
@@ -639,7 +639,7 @@ defmodule PlausibleWeb.SiteControllerTest do
   end
 
   describe "GET /:website/settings/imports-exports" do
-    setup [:create_user, :log_in, :create_site]
+    setup [:create_user, :log_in, :create_site, :maybe_fake_minio]
```
ruslandoga, line 642: Instead of mocking functions, I decided to mock minio :) It's less intrusive and easier.

lib/plausible_web/live/csv_export.ex
```diff
@@ -0,0 +1,270 @@
...
+      |> fetch_export()
+
+    if connected?(socket) do
+      Exports.oban_listen()
+    end
+
+    {:ok, socket}
+  end
+
+  defp fetch_export(socket) do
```
ruslandoga, line 35: I know this is ugly, but it covers (?) all possibilities and seems to work for now.

lib/plausible/sites.ex
```diff
@@ -23,14 +23,6 @@ defmodule Plausible.Sites do
     Repo.get_by!(Site, domain: domain)
   end
 
-  def get_domain!(site_id) do
```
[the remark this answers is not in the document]
ruslandoga: When a function like that is defined in the same module as where it's used, `defp` and compiler ensure it'd be removed once no longer necessary.

lib/plausible/s3.ex
```diff
@@ -43,31 +43,29 @@ defmodule Plausible.S3 do
...
   if Mix.env() in [:dev, :test, :small_dev, :small_test] do
     defp extract_s3_url(presigned_url) do
       [s3_url, _] = String.split(presigned_url, "?")
-
-      if ch_host = System.get_env("S3_CLICKHOUSE_HOST") do
-        URI.to_string(%URI{URI.parse(s3_url) | host: ch_host})
-      else
-        s3_url
-      end
+      ch_host = System.get_env("MINIO_HOST_FOR_CLICKHOUSE", "host.docker.internal")
```
ruslandoga, line 67: I'm defaulting to `host.docker.internal` in dev and tests envs now. Makes everything work with no setup other than `make minio`

lib/plausible_web/live/csv_export.ex
```diff
@@ -0,0 +1,271 @@
...
+        <.download storage={@storage} export={@export} />
+    <% end %>
+    """
+  end
+
+  defp prepare_download(assigns) do
+    ~H"""
+    <Generic.button phx-click="export">Prepare download</Generic.button>
+    <p class="text-sm mt-4 text-gray-500">
+      Prepare your data for download by clicking the button above. When that's done, a Zip file that you can download would appear.
```
zoldar, line 113:
```suggestion
      Prepare your data for download by clicking the button above. When that's done, a Zip file that you can download will appear.
```

lib/plausible_web/live/csv_export.ex
```diff
@@ -0,0 +1,271 @@
...
+      <Heroicons.exclamation_circle class="w-4 h-4 text-red-500" />
+      <p class="ml-2 text-sm text-gray-500">
+        Something went wrong when preparing your download. Please
+        <button phx-click="export" class="text-indigo-500">try again.</button>
+      </p>
+    </div>
+    """
+  end
+
+  defp download(assigns) do
```
zoldar, line 145: Something for a follow-up but we could try to slightly improve consistency of the layout - for instance by using fully labelled buttons for cancelling/deletion instead of a trash icon. But again, we can smooth that out later on. \cc @RobertJoonas

lib/workers/export_analytics.ex
```diff
@@ -0,0 +1,123 @@
...
+
+    try do
+      case storage do
+        "s3" -> perform_s3_export(ch, site, queries, args)
+        "local" -> perform_local_export(ch, queries, args)
+      end
+    catch
+      class, reason ->
+        last_attempt_failed? = job.attempt == job.max_attempts
+        if last_attempt_failed?, do: email_failure(args)
```
zoldar, line 58: Are you doing it here in order to avoid going via `ObanErrorReporter`?

lib/workers/local_import_analytics_cleaner.ex
```diff
@@ -0,0 +1,30 @@
...
+      end)
+    end
+  end
+
+  defp import_in_progress?(import_id) do
+    import Ecto.Query
+
+    Plausible.Imported.SiteImport
+    |> where(id: ^import_id)
+    |> where([i], i.status in [:pending, :importing])
```
zoldar, line 27: It would be good to use status macros (`SiteImport.pending()` and `SiteImport.importing()`).

lib/workers/export_analytics.ex
ruslandoga: Yes, and I wanted to keep the logic in one place. Ideally ObanErrorReporter would only concern itself with Sentry reports, and other handlers would be used for app logic, including one that reports on completed jobs (it's hard right now to notify about a job completion).

lib/plausible/exports.ex
```diff
@@ -3,85 +3,285 @@ defmodule Plausible.Exports do
...
+        from e in "events_v2",
+          where: [site_id: ^site_id],
+          select: [
+            fragment("toDate(toTimeZone(min(?),?))", e.timestamp, ^timezone),
+            fragment("toDate(toTimeZone(max(?),?))", e.timestamp, ^timezone)
+          ]
       )
 
-    name <> ".zip"
+    # NOTE: is it still 0 if timezone conversion is applied?
```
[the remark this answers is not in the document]
ruslandoga: More descriptive

```sql
7bd1e16b6ecd :) select toDate(toTimeZone(max(timestamp),'Etc/GMT-14')) from events_v2 where site_id = 3;

SELECT toDate(toTimeZone(max(timestamp), 'Etc/GMT-14'))
FROM events_v2
WHERE site_id = 3

Query id: 5e12a5c6-983d-4fc9-ac78-84a3a9353d02

┌─toDate(toTimeZone(max(timestamp), 'Etc/GMT-14'))─┐
│                                       1970-01-01 │
└──────────────────────────────────────────────────┘

1 row in set. Elapsed: 0.029 sec.

7bd1e16b6ecd :) select toDate(toTimeZone(max(timestamp),'Etc/GMT+12')) from events_v2 where site_id = 3;

SELECT toDate(toTimeZone(max(timestamp), 'Etc/GMT+12'))
FROM events_v2
WHERE site_id = 3

Query id: cdf7a212-a8ce-4265-8a5e-41baf10cc0c4

┌─toDate(toTimeZone(max(timestamp), 'Etc/GMT+12'))─┐
│                                       1970-01-01 │
└──────────────────────────────────────────────────┘

1 row in set. Elapsed: 0.007 sec.
```

lib/plausible/exports.ex
```diff
@@ -195,11 +392,18 @@ defmodule Plausible.Exports do
           _sample_factor: fragment("_sample_factor")
         }
 
+    window_q =
+      if date_range do
+        where(window_q, [e], e.timestamp >= ^date_range.first and e.timestamp <= ^date_range.last)
```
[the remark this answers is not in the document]
ruslandoga: It's already done on line 380.

lib/plausible/exports.ex
```diff
@@ -174,44 +371,48 @@ defmodule Plausible.Exports do
...
   end
 
-  @spec export_pages_q(pos_integer) :: Ecto.Query.t()
-  def export_pages_q(site_id) do
+  defp export_pages_q(site_id, timezone, date_range) do
     window_q =
-      from e in sampled("events_v2"),
+      from e in sampled("events_v2", nil),
         where: e.site_id == ^site_id,
+        where: [name: "pageview"],
```
ruslandoga, line 378: I think we discussed it on the call, but just wanted to double check. Is it OK to only filter pageviews for `imported_pages` export?

lib/plausible/exports.ex
```diff
@@ -174,44 +371,48 @@ defmodule Plausible.Exports do
...
           fragment("toUInt64(round(uniq(?)*any(_sample_factor)))", e.session_id),
           :visits
         ),
         visitors(e),
-        selected_as(
-          fragment("toUInt64(round(countIf(?='pageview')*any(_sample_factor)))", e.name),
-          :pageviews
-        ),
-        # NOTE: are exits pageviews or any events?
+        selected_as(fragment("toUInt64(round(count()*any(_sample_factor)))"), :pageviews),
```
ruslandoga, line 415: Then `pageviews` column is just a `count()`

lib/plausible/exports.ex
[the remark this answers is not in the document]
ruslandoga: https://github.com/plausible/analytics/pull/3944/commits/abbc0207c6c2a3a6c672b02ef43873b08e25b745

lib/plausible/exports.ex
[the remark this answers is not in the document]
ruslandoga: https://github.com/plausible/analytics/pull/3944/commits/faa35d419e9e9273c652a2645631166d3af4bb3e

Thank you again! I really missed that one ...

lib/workers/local_import_analytics_cleaner.ex
ruslandoga: https://github.com/plausible/analytics/pull/3944/commits/bb8f90402b313c1e71e0bbf0d24c97c3495ee42c
https://github.com/plausible/analytics/pull/3944/commits/1644e38127a1885e0de67de6f6a8f718c89d4415

lib/plausible_web/email.ex
ruslandoga: https://github.com/plausible/analytics/pull/3980

lib/plausible_web/live/csv_import.ex
```diff
@@ -272,4 +301,55 @@ defmodule PlausibleWeb.Live.CSVImport do
...
+  defp find_free_ranges(start_date, end_date, occupied_ranges) do
+    free_ranges(Date.range(start_date, end_date), start_date, occupied_ranges, [])
+  end
+
+  # This function recursively finds open ranges that are not yet occupied
+  # by existing imported data. The idea is that we keep moving a dynamic
+  # date index `d` from start until the end of `imported_range`, hopping
+  # over each occupied range, and capturing the open ranges step-by-step
+  # in the `result` array.
+  defp free_ranges(import_range, d, [occupied_range | rest_of_occupied_ranges], result) do
```
ruslandoga, line 327: This is a temp solution to ask for advice / changes to the existing function.

1. `check_dates` function has a typespec that includes `no_data`
2. `check_dates` function queries DB for site_imports on each call, seems a bit expensive in a form validation function so I'm preloading the site imports for occupied ranges beforehand
3. `find_free_ranges`, and `free_ranges` are private -> I copied them here

The quick resolution would be to make some of these functions public and the free ranges check function pure. E.g. `list_occupied_ranges` which I would call on `mount`, `find_free_ranges` which  would accept `occupied_ranges` list which I would call on form validation, `cutoff_date` and co. would also be useful.

zoldar: How about adding `site_imports` as an optional argument to `check_dates`? The list can then be passed down to `find_free_ranges`, instead to of `site`. `Imported` can expose extra function, `get_occupied_ranges/1` accepting `site` which can be used to compute the default value of that argument.

`native_stats_start_date` is computed at most once anyway, so IMHO there's no point in precomputing it outside the function.

Sorry, maybe I'm missing something but it seems like we don't handle a case where original range is completely covered by existing imports (and native stats)? (it's effectively `no_time_window` error). 

As for `no_data`, yeah, this one's a bit of a pain. One way around this I see:

- remove `:no_data` clause from `check_dates`
- wrap `check_dates` in the context of GA imports and put that clause there specifically

This error is pretty specific to contexts where import data are queried for via 3rd party APIs anyway.

WDYT?

ruslandoga: > Sorry, maybe I'm missing something but it seems like we don't handle a case where original range is completely covered by existing imports (and native stats)? (it's effectively no_time_window error).

Is that the case where there are no free ranges? It's handled.

ruslandoga: A nil is returned which means no clamped data range which prevents form submission and shows a warning

ruslandoga: Re native_stats_start_date, it's queried each time for a new site with no data, which in my mind is a common scenario for imports.

zoldar: > A nil is returned which means no clamped data range which prevents form submission and shows a warning

So just to be clear, this means the warning will serve 2 different purposes, right? One where clamped range differs from original range (a warning is show but the user can still carry on), and another, where there's no free range to return (an error, preventing the user from continuing.

> Re native_stats_start_date, it's queried each time for a new site with no data, which in my mind is a common scenario for imports.

Ah, ok, I got it wrong anyway, as it's `Sites.stats_start_date` that does the caching. I wonder though, given the query is relatively light (search on primary index alone will turn up an empty result) and the validation isn't run with a frequency of a machine gun, maybe it's acceptable? If not, perhaps an alternative clause of `check_dates` accepting imports list and native stats start date _instead_ of site would be an option?

ruslandoga: The warning and confirm button serves these two purposes together. Warning is shown when clamped date range is not the same as the original one, i.e. when the original date range can be used. Confirm button is enabled when everything is ok and ready and disabled otherwise.

ruslandoga: I don't want to do any db queries in the form validation function.

lib/plausible_web/live/csv_import.ex
```diff
@@ -122,6 +141,12 @@ defmodule PlausibleWeb.Live.CSVImport do
     """
   end
 
+  defp date_range_warning(assigns) do
+    ~H"""
+    TODO
```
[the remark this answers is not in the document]
zoldar: We need 2 different messages here, right?

1. When the final date range differs from original, meaning the import's stats will be cropped to fit the gap between existing stats.
2. When there's no free gap for imported stats.

lib/plausible/exports.ex
zoldar: AFAIR, yeah. @RobertJoonas ?

ruslandoga: Or can the exits be dropped entirely? Do pageviews count exits in?

lib/plausible_web/live/csv_import.ex
ruslandoga: So how do we proceed?

ruslandoga: What would check_dates do? Wouldn't it become find_free_ranges with these changes?

zoldar: It would still transform imports into a list of ranges and determine effective cutoff_date (either native stats start date or today). It would also handle case of too short time range (less than 2 days) and final extraction of maximum gap. This logic is still shared between GA and CSV imports AFAICT.

Just to be clear, I meant something like maintaining 2 clauses:

```elixir
def check_dates(site, start_date, end_date)
```

and

```elixir
def check_dates(site_imports, native_stats_start_date, start_date, end_date)
```

Where the former one can rely on the latter one, providing site imports and native_stats_start_date where caching is not necessary.

ruslandoga: Would you be able to open a PR with these changes or should I do it?

zoldar: @ruslandoga will https://github.com/plausible/analytics/pull/3983 work for you?

ruslandoga: Yes, thank you!

test/plausible/exports_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule Plausible.ExportsTest do
-  use Plausible.DataCase, async: true
+  use Plausible.DataCase
```
ruslandoga, line 2: I couldn't find a reason for why, but the "stops on error" was raising a Mint.TransportError (socket closed) instead of Ch.Error when tests were async with seed 643472.

test/workers/import_analytics_test.exs
```diff
@@ -145,6 +145,7 @@ defmodule Plausible.Workers.ImportAnalyticsTest do
         Ecto.Adapters.SQL.Sandbox.unboxed_run(Plausible.Repo, fn ->
           Repo.delete_all(Plausible.Site)
           Repo.delete_all(Plausible.Auth.User)
+          Repo.delete_all(Oban.Job)
```
ruslandoga, line 148: [Oban.drain_queue](https://hexdocs.pm/oban/testing_queues.html#integration-testing-queues) which I use in export/import tests picks up these old jobs from unboxed runs. I think it's OK to delete them.

lib/plausible_web/live/csv_import.ex
```diff
@@ -231,10 +231,18 @@ defmodule PlausibleWeb.Live.CSVImport do
...
-  def handle_info({:notification, :analytics_imports_jobs, _details}, socket) do
-    # TODO limit to current site
-    occupied_ranges = Imported.get_occupied_date_ranges(socket.assigns.site)
-    socket = socket |> assign(occupied_ranges: occupied_ranges) |> process_imported_tables()
+  def handle_info({:notification, :analytics_imports_jobs, details}, socket) do
+    %{"site_id" => notification_site_id} = details
+    site = socket.assigns.site
+
+    socket =
+      if notification_site_id == site.id and match?(%{"complete" => _import_id}, details) do
```
zoldar, line 246: Sorry, haven't had my coffee yet, but wouldn't `... and details["complete"]` work the same? 🤔

ruslandoga: It would, and it would be better. I'll change it.

lib/plausible/exports.ex
```diff
@@ -3,85 +3,285 @@ defmodule Plausible.Exports do
...
+    if export = get_s3_export(site_id) do
+      exports_bucket = Plausible.S3.exports_bucket()
+      delete_op = ExAws.S3.delete_object(exports_bucket, export.path)
+      ExAws.request!(delete_op)
+    end
+
+    :ok
+  end
+
+  defp s3_export_key(site_id), do: Integer.to_string(site_id)
```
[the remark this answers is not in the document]
zoldar: Discussed this with @ukutaht on a call shortly. We'll eventually want to apply some form of hashing here and in presigned upload URLs (CSV imports). This though can be addressed in a follow-up.

PR #3948, opened 2024-03-26
Ensure all the react container attributes are strings

aerosol:
Ref #3762 

#3892 introduced a regression - during template conversion from `eex` to `heex` the booleans were rendered verbatim, whilst the react code is expecting string literals. This caused, for example, "set up your goals" prompts appear on customer dashboards, despite them having goals configured.


zoldar (approved):
Sneaky 😬

PR #3949, opened 2024-03-26
Migration only - add new metric columns to imported tables

RobertJoonas:
### Changes

Since we want to be able to support very basic filtering with imported data at some point, we need to start fetching some additional metrics. This PR contains only the migration adding those fields.

### DIFF
--- a/priv/ingest_repo/migrations/20240326134840_add_metrics_to_imported_tables.exs
+++ b/priv/ingest_repo/migrations/20240326134840_add_metrics_to_imported_tables.exs
@@ -1,0 +1,30 @@
+defmodule Plausible.IngestRepo.Migrations.AddMetricsToImportedTables do
+  use Ecto.Migration
+
+  @add_pageviews_for [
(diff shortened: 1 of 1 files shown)


lib/plausible/imported/csv_importer.ex
```diff
@@ -43,9 +43,17 @@ defmodule Plausible.Imported.CSVImporter do
...
+        |> String.split(",", trim: true)
+        |> Enum.map_join(", ", fn kv ->
+          [col, _type] = String.split(kv)
+          col
+        end)
+
       statement =
         """
-        INSERT INTO {table:Identifier} \
+        INSERT INTO {table:Identifier}(site_id, #{s3_structure_cols_expr}, import_id) \
```
macobo, line 56: Note: You might need to split this PR into 3 to do this safely:
1. for adding strict column names to insert
2. for adding new columns
3. for writing to new columns

:sweat_smile:

zoldar (approved):

PR #3952, opened 2024-03-27
Oban (local) CSV import

ruslandoga:
### Changes

This PR extends the current CSVImporter and it's LiveView to allow uploading imports both to S3 and to local storage.

Depends on #3944 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/imported/csv_importer.ex
+++ b/lib/plausible/imported/csv_importer.ex
@@ -123,7 +123,9 @@ defmodule Plausible.Imported.CSVImporter do
         "end_date" => end_date
       }
 
-      File.stream!(local_path, _512KB = 512_000)
+      # we are reading in 512KB chunks for better performance
+      # the default would've been line by line (not great for a CSV)
+      File.stream!(local_path, 512_000)
       |> Stream.into(Ch.stream(ch, statement, params))
       |> Stream.run()
     end)
--- a/lib/plausible/imported/csv_importer.ex
+++ b/lib/plausible/imported/csv_importer.ex
@@ -100,36 +100,42 @@ defmodule Plausible.Imported.CSVImporter do
       |> Keyword.replace!(:pool_size, 1)
       |> Ch.start_link()
 
-    Enum.each(uploads, fn upload ->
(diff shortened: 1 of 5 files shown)


lib/plausible_web/live/csv_import.ex
```diff
@@ -159,6 +165,31 @@ defmodule PlausibleWeb.Live.CSVImport do
...
+    persistent_cache_dir = Application.get_env(:plausible, :persistent_cache_dir)
+
+    local_dir =
+      Path.join([
+        persistent_cache_dir || System.tmp_dir!(),
+        "plausible-imports",
+        site_id
+      ])
+
+    File.mkdir_p!(local_dir)
```
ruslandoga, line 184: I'll extract this to some other module, just need to decide which one. I guess CSVImporter is good enough choice.

lib/plausible/imported/csv_importer.ex
```diff
@@ -67,10 +83,57 @@ defmodule Plausible.Imported.CSVImporter do
...
+      # the default would've been line by line (not great for a CSV)
+      File.stream!(local_path, 512_000)
+      |> Stream.into(Ch.stream(ch, statement, params))
+      |> Stream.run()
+    end)
+
+    Enum.each(uploads, fn upload ->
+      %{"local_path" => local_path} = upload
+      File.rm!(local_path)
+    end)
```
ruslandoga, line 136: I'm cleaning the old files here. Something feels wrong about this.

lib/plausible/imported/csv_importer.ex
```diff
@@ -67,10 +83,64 @@ defmodule Plausible.Imported.CSVImporter do
 
       Ch.query!(ch, statement, params, timeout: :infinity)
     end)
-  rescue
-    # we are cancelling on any argument or ClickHouse errors
-    e in [ArgumentError, Ch.Error] ->
-      {:error, Exception.message(e)}
+  end
+
+  defp import_local(site_import, opts) do
```
ruslandoga, line 88: - [ ] needs tests

PR #3958, opened 2024-03-28
Oban (local) CSV export improvements

ruslandoga:
This PR improves https://github.com/plausible/analytics/pull/3944, but to avoid breaking that other PR during review, I'm keeping it separate until it 100% works.


lib/plausible/exports.ex
```diff
@@ -192,6 +281,8 @@ defmodule Plausible.Exports do
       name <> extname
     end
 
+    # TODO limit export queries to the provided date range (if provided)
```
ruslandoga, line 284: - [ ] todo

PR #3959, opened 2024-03-28
Worker to clean site data from ClickHouse

macobo:
### Changes

Add a new worker to clean deleted sites data from clickhouse.

This will eventually become a cronjob, for now running it manually on cloud.

### DIFF
--- a/lib/workers/clickhouse_clean_sites.ex
+++ b/lib/workers/clickhouse_clean_sites.ex
@@ -1,0 +1,63 @@
+defmodule Plausible.Workers.ClickhouseCleanSites do
+  use Plausible.Repo
+  use Plausible.ClickhouseRepo
+  use Plausible.IngestRepo
+  use Oban.Worker, queue: :clickhouse_clean_sites
+
+  import Ecto.Query
+
+  require Logger
+
+  @tables_to_clear [
+    "events_v2",
+    "sessions_v2",
+    "ingest_counters",
+    "imported_browsers",
+    "imported_devices",
+    "imported_entry_pages",
+    "imported_exit_pages",
+    "imported_locations",
+    "imported_operating_systems",
+    "imported_pages",
+    "imported_sources",
+    "imported_visitors"
+  ]
+
+  def perform(_job) do
+    deleted_sites = get_deleted_sites_with_clickhouse_data()
+
+    if length(deleted_sites) > 0 do
+      Logger.info(
+        "Clearing ClickHouse data for the following #{length(deleted_sites)} sites which have been deleted: #{inspect(deleted_sites)}"
+      )
+
+      site_ids_expr = deleted_sites |> Enum.map(&to_string/1) |> Enum.join(", ")
+
+      for table <- @tables_to_clear do
+        IngestRepo.query(
+          "ALTER TABLE #{table} DELETE WHERE site_id IN (#{site_ids_expr})",
+          [],
+          if(Mix.env() == :test, do: [mutations_sync: 1], else: [])
+        )
+      end
+    end
+
+    :ok
+  end
+
+  def get_deleted_sites_with_clickhouse_data() do
+    pg_sites =
+      from(s in Plausible.Site, select: %{id: s.id})
+      |> Plausible.Repo.all()
+      |> Enum.map(&(&1 |> Map.fetch!(:id)))
+      |> MapSet.new()
+
+    ch_sites =
+      from(e in "events_v2", group_by: e.site_id, select: %{site_id: e.site_id})
+      |> Plausible.ClickhouseRepo.all(timeout: :infinity)
+      |> Enum.map(& &1.site_id)
+      |> MapSet.new()
+
+    MapSet.difference(ch_sites, pg_sites) |> MapSet.to_list()
+  end
+end
--- a/lib/workers/clickhouse_clean_sites.ex
+++ b/lib/workers/clickhouse_clean_sites.ex
@@ -30,6 +30,8 @@ defmodule Plausible.Workers.ClickhouseCleanSites do
     "imported_visitors"
   ]
 
+  @settings if Mix.env() in [:test, :small_test], do: [mutations_sync: 2], else: []
+
   def perform(_job) do
     deleted_sites = get_deleted_sites_with_clickhouse_data()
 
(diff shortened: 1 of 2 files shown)


lib/workers/clickhouse_clean_sites.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  def perform(_job) do
+    deleted_sites = get_deleted_sites_with_clickhouse_data()
+
+    if length(deleted_sites) > 0 do
+      Logger.info(
+        "Clearing ClickHouse data for the following #{length(deleted_sites)} sites which have been deleted: #{inspect(deleted_sites)}"
+      )
+
+      site_ids_expr = deleted_sites |> Enum.map_join(", ", &to_string/1)
```
zoldar, line 41: Just double-checking - is there some practical limit on the length of an array in such queries? Or maybe is there a point at which it's better to check such array in chunks for perf reasons?

macobo: Good question. This is determined by [max_query_size](https://clickhouse.com/docs/en/operations/settings/settings#max_query_size) setting in clickhouse, defaults to 256 KiB.

Chunking isn't likely a good idea - less mutations is better after all. But I might need to set a limit on the number of sites deleted per run, will see when I run it on cloud.

ukutaht (approved):

zoldar (approved):

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -0,0 +1,70 @@
...
+
+      for table <- @tables_to_clear do
+        IngestRepo.query(
+          "ALTER TABLE #{table} DELETE WHERE site_id IN (#{site_ids_expr})",
+          [],
+          if(Mix.env() == :test, do: [mutations_sync: 2], else: [])
```
ruslandoga, line 47: `Mix.env()` is not available in releases so it would crash.

ruslandoga: This expression can be moved outside of `perform/1` and the result can be copied into the function.


```elixir
@settings if Mix.env() == :test, do: [mutations_sync: 2], else: []

def perform(...) do
  # ...
  IngestRepo.query(
           "ALTER TABLE #{table} DELETE WHERE site_id IN (#{site_ids_expr})",
           [],
           settings: @settings
  # ...
  end
```

ruslandoga: `site_ids_expr` can be passed as params but I guess it's not that importent.

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -0,0 +1,70 @@
...
+        )
+      end
+    end
+
+    :ok
+  end
+
+  def get_deleted_sites_with_clickhouse_data() do
+    pg_sites =
+      from(s in Plausible.Site, select: %{id: s.id})
```
ruslandoga, line 57: from(s in Plausible.Site, select: s.id)` is also possible for both queries.

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -0,0 +1,70 @@
...
+    "imported_operating_systems",
+    "imported_pages",
+    "imported_sources",
+    "imported_visitors"
+  ]
+
+  def perform(_job) do
+    deleted_sites = get_deleted_sites_with_clickhouse_data()
+
+    if length(deleted_sites) > 0 do
```
ruslandoga, line 36:
```suggestion
    if deleted_sites != [] do
```

ruslandoga: Otherwise `length(list)` performs unnecessary work.

lib/workers/clickhouse_clean_sites.ex
ruslandoga: `IngestRepo.query!` might be a better fit here to make sure the job fails if any of the queries fail.

macobo: What's the way to check if something works at release time?

ruslandoga: You can build the release locally and try running the relevant functions. Here it would've been

```console
$ MIX_ENV=prod mix release
-- set required env vars, start ClickHouse and PostgreSQL dbs
$ _build/...../bin start_iex
```
```elixir
iex> Oban.insert!(Plausible.Workers.ClickhouseCleanSites.new())
# optionally Oban.rain_queue
```

But in reality this is unlikely to be necessary, as there is only one such library that's available during build time but not at runtime, and it's Mix.

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -38,13 +40,11 @@ defmodule Plausible.Workers.ClickhouseCleanSites do
...
-
       for table <- @tables_to_clear do
-        IngestRepo.query(
-          "ALTER TABLE #{table} DELETE WHERE site_id IN (#{site_ids_expr})",
-          [],
-          if(Mix.env() == :test, do: [mutations_sync: 2], else: [])
+        IngestRepo.query!(
+          "ALTER TABLE #{table} DELETE WHERE site_id IN {site_ids:Array(UInt64)}",
+          %{site_ids: deleted_sites},
+          @settings
```
ruslandoga, line 47: Isn't it `settings: @settings`?

lib/workers/clickhouse_clean_sites.ex
macobo: Thanks! Note the idea on using maps as params is perfect, yet fails lint:

https://github.com/plausible/analytics/actions/runs/8518356635/job/23330371745?pr=3959

ruslandoga: Ah, then it can be a list :)

```elixir
IngestRepo.query(
  "ALTER TABLE {$0:Identifier} DELETE WHERE site_id IN {$1:Array(UInt64)}",
  [table, site_ids],
  settings: @settings
)
```

test/workers/clickhouse_clean_sites_test.exs
```diff
@@ -0,0 +1,69 @@
...
+             ClickhouseCleanSites.get_deleted_sites_with_clickhouse_data(),
+             deleted_site.id
+           )
+
+    assert not Enum.member?(
+             ClickhouseCleanSites.get_deleted_sites_with_clickhouse_data(),
+             site.id
+           )
+
+    ClickhouseCleanSites.perform(nil)
```
ruslandoga, line 43: It can also be tested through the whole Oban pipeline, this would make sure that the `:clickhouse_clean_sites` queue is actually defined and active.

https://hexdocs.pm/oban/testing_queues.html#integration-testing-queues

```elixir
Oban.insert!(ClickhouseCleanSites.new())
assert %{success: 1} = Oban.drain_queue(queue: :clickhouse_clean_sites, with_safety: false)
```

PR #3953, opened 2024-03-27
CH Migration: exit page hostname on sessions_v2

aerosol:
### Changes

This PR adds `exit_page_hostname` to `sessions_v2` table, so that we can filter by hostnames avoiding joining `events_v2` where possible.

https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/6925576254

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #3963, opened 2024-04-01
Filter by hostnames

aerosol:
https://github.com/plausible/analytics/pull/3953 already deployed and merged.


test/plausible_web/controllers/api/stats_controller/pages_test.exs
```diff
@@ -1227,6 +1333,68 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
...
+        )
+
+      assert json_response(conn, 200) == [
+               %{
+                 "visitors" => 1,
+                 "visits" => 1,
+                 "name" => "/page1",
+                 "visit_duration" => 0
+               },
+               %{
```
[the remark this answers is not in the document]
aerosol: I suppose this demonstrates how unreadable the populate_events fixture is. The very last event arrives with a timestamp that's way over of the session window, that's why "/page2" is a valid entry page.

test/plausible/session/cache_store_test.exs
```diff
@@ -89,6 +89,124 @@ defmodule Plausible.Session.CacheStoreTest do
...
+        )
+      ]
+
+      flush(events)
+      session = get_session(site_id)
+      assert session.hostname == "example.com"
+      assert session.exit_page_hostname == "example.com"
+    end
+
+    test "hostname change" do
```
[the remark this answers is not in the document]
aerosol: Welp, got it. That breaks pretty much every test written here, so I'll start there.

test/plausible_web/controllers/api/stats_controller/top_stats_test.exs
```diff
@@ -642,6 +642,32 @@ defmodule PlausibleWeb.Api.StatsController.TopStatsTest do
...
 
+    test "returns only visitors with specific screen size for a given hostname", %{
+      conn: conn,
+      site: site
+    } do
+      populate_stats(site, [
+        build(:pageview, screen_size: "Desktop", hostname: "blog.example.com"),
+        build(:pageview, screen_size: "Desktop", hostname: "example.com", user_id: @user_id),
+        build(:pageview, screen_size: "Desktop", hostname: "blog.example.com", user_id: @user_id),
+        build(:pageview, screen_size: "Mobile")
```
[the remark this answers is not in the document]
aerosol: So here, filtering by hostname should exclude `@user_id` session effectively, right? Since they're entering at example.com. That's what's happening if I apply the suggestion from https://github.com/plausible/analytics/pull/3963#discussion_r1546394817

test/plausible_web/controllers/api/stats_controller/pages_test.exs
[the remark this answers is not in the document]
ukutaht: Ah you're right. 

Agree that it should show up but 2 visits and 480 visit duration seem wrong still, no? AFAICT there's a single visit that entered through `es.example.com/page2` and it lasted 0 seconds because it's the only pageview in that session.

test/plausible/session/cache_store_test.exs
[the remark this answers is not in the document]
aerosol: Also this test was missing `user_id`s by mistake, so that subsequent events weren't forming a session exactly :facepalm:

test/plausible_web/controllers/api/stats_controller/pages_test.exs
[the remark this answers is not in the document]
aerosol: Both variants addressed here: https://github.com/plausible/analytics/pull/3963/files#diff-bb15a81aed739a750c28dd95710b97eb872ece8db6f8513423527be89379b220R1387-R1489

test/plausible_web/controllers/api/stats_controller/pages_test.exs
```diff
@@ -888,6 +919,81 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
...
+
+      conn =
+        get(
+          conn,
+          "/api/stats/#{site.domain}/pages?period=day&date=2021-01-01&detailed=true&filters=#{filters}"
+        )
+
+      assert json_response(conn, 200) == [
+               %{
+                 "bounce_rate" => 50,
```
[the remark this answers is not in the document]
aerosol: The extra case added here: https://github.com/plausible/analytics/pull/3963/files#diff-bb15a81aed739a750c28dd95710b97eb872ece8db6f8513423527be89379b220R922-R960

test/plausible_web/controllers/api/stats_controller/top_stats_test.exs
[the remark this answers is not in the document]
aerosol: Updated: https://github.com/plausible/analytics/pull/3963/files#diff-7d7a21813635f36c53370eb9e8fb6cb97daa9aa01b52811a3d5382dc6f8a2095R645-R676

test/plausible/session/cache_store_test.exs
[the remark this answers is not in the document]
aerosol: Done via https://github.com/plausible/analytics/pull/3963/commits/9eeaba42f09c34ac2c764d3a40fa018638d4b4b8

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -428,6 +428,14 @@ defmodule PlausibleWeb.Api.StatsController do
     query = Query.from(site, params)
     pagination = parse_pagination(params)
 
+    # untested yet
```
aerosol, line 431: cc @ukutaht

assets/js/dashboard/util/filters.js
```diff
@@ -1,3 +1,5 @@
+const flags = JSON.parse(document.getElementById('stats-react-container').dataset.flags)
```
aerosol, line 1: cc @ukutaht - is this enough?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -423,7 +423,6 @@ defmodule PlausibleWeb.Api.StatsController do
     query = Query.from(site, params)
     pagination = parse_pagination(params)
 
-    # untested yet
```
aerosol: cc @ukutaht - added tests for the referrer/goal case you brought up on basecamp

lib/plausible/stats/query.ex
```diff
@@ -36,12 +37,14 @@ defmodule Plausible.Stats.Query do
     query
   end
 
-  defp put_experimental_session_count(query, params) do
-    if Map.get(params, "experimental_session_count") == "true" do
-      struct!(query, experimental_session_count?: true)
-    else
-      query
-    end
+  defp put_experimental_flags(query, params) do
```
macobo, line 40: Note: https://github.com/plausible/analytics/pull/3966 goes a different route with the flags. I suggest doing the same here to avoid needing to double-check flag + query flag each time.

lib/plausible/stats/base.ex
```diff
@@ -130,7 +133,8 @@ defmodule Plausible.Stats.Base do
     "os_version" => "operating_system_version",
     "country" => "country_code",
     "region" => "subdivision1_code",
-    "city" => "city_geoname_id"
+    "city" => "city_geoname_id",
+    "entry_page_hostname" => "hostname"
```
macobo, line 137: Why not name the api prop name the same as the db column?

Note we'll likely want to get rid of all this back-and-forth mapping eventually to simplify fe -> query -> response mapping.

assets/js/dashboard/util/filters.js
```diff
@@ -7,10 +9,12 @@ export const FILTER_GROUPS = {
...
   'utm': ['utm_medium', 'utm_source', 'utm_campaign', 'utm_term', 'utm_content'],
   'goal': ['goal'],
-  'props': ['prop_key', 'prop_value']
+  'props': ['prop_key', 'prop_value'],
+  ...(flags.hostname_filter ? { 'hostname': ['hostname', 'experimental_hostname_filter'] } : {})
+
 }
 
-export const NO_CONTAINS_OPERATOR = new Set(['goal', 'screen'].concat(FILTER_GROUPS['location']))
+export const NO_CONTAINS_OPERATOR = new Set(['experimental_hostname_filter', 'goal', 'screen'].concat(FILTER_GROUPS['location']))
```
macobo, line 17: Q: Why not allow contains for hostnames?

lib/plausible/stats/base.ex
aerosol: Because `hostname` is already taken by `event:hostname` and we'd clash here: https://github.com/plausible/analytics/blob/0e70fb37ffc07b5b6f3823f9d97203fd68159705/lib/plausible/stats/filters/filters.ex#L27-L32 

and eventually here:

https://github.com/plausible/analytics/blob/0e70fb37ffc07b5b6f3823f9d97203fd68159705/lib/plausible/stats/base.ex#L18-L31

or here:

https://github.com/plausible/analytics/blob/0e70fb37ffc07b5b6f3823f9d97203fd68159705/lib/plausible/stats/filters/dashboard_filter_parser.ex#L14-L17

assets/js/dashboard/util/filters.js
aerosol: It's allowed. 

image

lib/plausible/stats/query.ex
aerosol: Could you elaborate?

aerosol: This PR introduces only one feature flag - `:hostname_filter` that is being used to either display or hide the filter dropdown item. The `experimental_hostname_filter` is a query string parameter, that's temporarily exposed in the filter modal, so we can quickly test two competing behaviours, before deciding on the final behaviour. Not sure what's there to avoid.

macobo: Ah. In https://github.com/plausible/analytics/pull/3966 I unified `experimental_session_count` param flag check and feature flag check into query.ex (and did the same for the new flag).

lib/plausible/stats/base.ex
macobo: Thanks. This will falls into the bucket of "needs cleanup when we try unify filtering further"

aerosol: That's why I asked about column renaming before - but wasn't worth it in the end.

assets/js/dashboard/util/filters.js
ukutaht: not sure what you mean. If it works it works :)

aerosol: just meant that if it's enough to just hide the filter dropdown on flag disabled

lib/plausible_web/controllers/api/stats_controller.ex
ukutaht: LGTM

ukutaht (approved):

PR #3966, opened 2024-04-03
Reducing JOINs in queries

macobo:
### Changes

We're currently denormalizing session properties such as browser, utm tag and others to be written onto events_v2 table. The ultimate goal is to speed up queries, which this PR is a step towards.

To this effect, this PR:
- introduces a new Feature Flag, `experimental_reduced_joins` 
- adds a new module, `Plausible.Stats.TableDecider` that helps decide what table(s) need to be queried given target metrics and query.
- makes fewer joins and filters on event table if `experimental_reduced_joins` flag is enabled
- runs all tests with both `experimental_reduced_joins` flag on and off

Plan after this PR:
- [Finish migrating cloud](https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/7231642925)
- [Verify the new code on cloud using the old code](https://3.basecamp.com/5308029/buckets/35611491/card_tables/cards/7253886127)

### Tests

### Changelog


lib/plausible/stats/table_decider.ex
```diff
@@ -0,0 +1,112 @@
...
+      true ->
+        {event_metrics ++ sample_percent, session_metrics ++ either_metrics ++ sample_percent,
+         other_metrics}
+    end
+  end
+
+  defp metric_partitioner(_, :conversion_rate), do: :event
+  defp metric_partitioner(_, :average_revenue), do: :event
+  defp metric_partitioner(_, :total_revenue), do: :event
+  defp metric_partitioner(_, :pageviews), do: :event
```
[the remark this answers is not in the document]
macobo: I thought so too, but tests caught an important edge case: If session starts before time window, that would count events/pageviews from before the time window.

lib/plausible/stats/table_decider.ex
```diff
@@ -0,0 +1,109 @@
...
+      if breakdown_property do
+        Map.put(query.filters, breakdown_property, nil)
+      else
+        query.filters
+      end
+
+    %{event: event_only_filters, session: session_only_filters} =
+      partition(filters, query, &filters_partitioner/2)
+
+    cond do
```
ukutaht, line 36: Aren't the first two clauses superfluous here?

lib/plausible/stats/breakdown.ex
```diff
@@ -200,9 +199,30 @@ defmodule Plausible.Stats.Breakdown do
...
+    end
+  end
+
+  defp breakdown_table(_query, _metrics, "visit:entry_page"), do: :session
+  defp breakdown_table(_query, _metrics, "visit:exit_page"), do: :session
+
+  defp breakdown_table(query, metrics, property) do
+    {_, session_metrics, _} = TableDecider.partition_metrics(metrics, query, property)
+
+    if not Enum.empty?(session_metrics) do
```
ukutaht, line 221:
```suggestion
    if Enum.any?(session_metrics) do
```

ukutaht (commented):
Should `TableDecider` also be used for [timeseries](https://github.com/plausible/analytics/blob/2ba4988c95cdb311e4b90fb6e39f1ba147741cc5/lib/plausible/stats/timeseries.ex#L28-L29)?

I'm not fully convinced that with the flag disabled we keep exactly the same queries/logic as before. Take, for example, a breakdown of sources with a page filter. It used to go through the `breakdown_sessions` code path but now it will do `breakdown_events` regardless of whether the flag is enabled or disabled.

Aggregate seems to be equivalent with the flag off to me.

I'm thinking for safety and fair comparison we should branch in the breakdown code based on the `experimental_reduced_joins` flag. If false, keep the old `breakdown_sessions` code exactly as it was. WDYT?

macobo:
> Should TableDecider also be used for [timeseries](https://github.com/plausible/analytics/blob/2ba4988c95cdb311e4b90fb6e39f1ba147741cc5/lib/plausible/stats/timeseries.ex#L28-L29)?

It is! https://github.com/plausible/analytics/pull/3966/files#diff-10a2a2c8ee92f94617f55392ff859fc46f40c933798cadfe644fdc403d6fe2e0R24-R25

> I'm not convinced that with the flag disabled we keep exactly the same queries/logic as before. Take, for example, a breakdown of sources with a page filter. It used to go through the breakdown_sessions code path but now it will do breakdown_events regardless of whether the flag is enabled or disabled.

Hmm. Good catch. The issue seems to be that unlike timeline/aggregates, visitors is supported as a session metric in breakdowns implicitly. We could mark it as _either_ but that would be a break in behavior for timeline/aggregates so let's go with your suggestion and default to `session` for now if flag is off!

lib/plausible/stats/breakdown.ex
macobo: This works, but not a fan. Consider the documentation:

image

When reading code I'd be wondering if there could be any falsy values in the list.

ukutaht: Fair point. I suggested this because generally I find negated conditions harder to parse in my head than positive 'assertions'.  Not a big deal, I don't like arguing about style 😀

lib/plausible/stats/table_decider.ex
ukutaht: Ah I see now. Makes sense!

ukutaht (approved):

test/plausible/stats/table_decider_test.exs
```diff
@@ -0,0 +1,164 @@
...
+  end
+
+  describe "partition_metrics" do
+    test "with no metrics or filters" do
+      query = make_query(false, %{})
+
+      assert partition_metrics([], query) == {[], [], []}
+    end
+
+    test "events-only metrics accordingly" do
```
RobertJoonas, line 30:
```suggestion
    test "session-only metrics accordingly" do
```

test/plausible/stats/table_decider_test.exs
```diff
@@ -0,0 +1,164 @@
...
+      query = make_query(false, %{})
+
+      assert partition_metrics([:bounce_rate, :views_per_visit], query) ==
+               {[], [:bounce_rate, :views_per_visit], []}
+    end
+
+    test "session-only metrics accordingly" do
```
RobertJoonas, line 37:
```suggestion
    test "event-only metrics accordingly" do
```

RobertJoonas (approved):
Looks good to me! 👍

test/plausible/stats/table_decider_test.exs
macobo: :see_no_evil:

PR #3969, opened 2024-04-03
Populate new columns in imports and exports

zoldar:
### Changes

Follow-up to #3949 doing what follows:

* Extend all `Imported.*` schemas with columns added by the migration
* Extend GA4 import procedure to populate the newly added fields (or more precisely, a subset of them which can be currently extracted from GA4 API)
* Extend CSV export queries and add new fields to the exported files
* Extend CSV imports to ingest newly additional exported data back
* Generate new fixtures for GA4 import test with larger and more varied payloads

### Tests
- [x] Automated tests have been updated (but will require further extension going forward)


RobertJoonas (approved):
Looks good! Could use more tests, but we can work on that in a follow-up.

PR #3970, opened 2024-04-04
Improve import procedure to ensure no time range overlaps

zoldar:
### Changes

image

This PR ensures that a new import can be started only if all the current ones are in completed state for a given site.

When check import's time range and altering it so that we avoid overlap, only complete imports are considered. If more than a single import is in progress at a time, there's a risk of ending up in inconsistent state where imports do overlap one another.

The check is enforced on UI level for Google imports as well as on Importer level to avoid a situation where multiple jobs get created for the same import due to, for instance, double form submission.

### Tests

### Dark mode

### DIFF
--- a/lib/plausible/imported.ex
+++ b/lib/plausible/imported.ex
@@ -62,9 +62,9 @@ defmodule Plausible.Imported do
     }
   end
 
-  @spec get_import(non_neg_integer()) :: SiteImport.t() | nil
-  def get_import(import_id) do
-    Repo.get(SiteImport, import_id)
+  @spec get_import(Site.t(), non_neg_integer()) :: SiteImport.t() | nil
+  def get_import(site, import_id) do
+    Repo.get_by(SiteImport, id: import_id, site_id: site.id)
   end
 
   defdelegate listen(), to: Imported.Importer
--- a/lib/plausible/imported.ex
+++ b/lib/plausible/imported.ex
@@ -83,6 +83,16 @@ defmodule Plausible.Imported do
     end
   end
 
+  @spec other_imports_in_progress?(SiteImport.t()) :: boolean()
+  def other_imports_in_progress?(site_import) do
+    Repo.exists?(
+      from(i in SiteImport,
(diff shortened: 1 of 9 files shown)


lib/plausible/imported.ex
```diff
@@ -83,6 +83,16 @@ defmodule Plausible.Imported do
     end
   end
 
+  @spec other_imports_in_progress?(SiteImport.t()) :: boolean()
+  def other_imports_in_progress?(site_import) do
+    Repo.exists?(
+      from(i in SiteImport,
+        where: i.site_id == ^site_import.site_id and i.id != ^site_import.id,
```
RobertJoonas, line 90: Is the `i.id != ^site_import.id` condition relevant? Why not simply query for any in-progress `site_imports` for a given site? I mean something like:

`def any_imports_in_progress?(site)`

So we'd prevent from scheduling a new job if there is an existing SiteImport in progress.

RobertJoonas (approved):
Left a couple of suggestions but overall looks good to me 👍

lib/plausible/imported.ex
zoldar: The check is done right before scheduling the job but after creating the site import entry, inside a transaction at https://github.com/plausible/analytics/pull/3970/files#diff-5fd606e6c7bee16b593b9d46e27286e75067242bfa30fb1546d59923c816a4bdL162-L176. That's why the check has to explicitly exclude the current site import.

lib/plausible_web/controllers/google_analytics_controller.ex
```diff
@@ -265,48 +265,58 @@ defmodule PlausibleWeb.GoogleAnalyticsController do
...
+
+    opts =
+      if property? do
+        Keyword.put(opts, :property, property_or_view)
+      else
+        Keyword.put(opts, :view_id, property_or_view)
+      end
+
+    importer_fun.(site, current_user, opts)
+  end
```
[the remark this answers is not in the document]
zoldar: https://github.com/plausible/analytics/pull/3970/commits/56e392782ee929ea82f44f457399cedb743720c6

PR #3971, opened 2024-04-04
Include query string when logging the request

macobo:
Ultimate goal is to be able to compare results with and without a flag against each other. To do this we need logging which displays the full request url with parameters.

Example logs:
```
14:46:09.042 request_id=F8MRLSsaKB7BeIkAAAHk [info] (200) GET /api/sites took 17ms
14:46:09.175 request_id=F8MRLTKYV3G-GqEAAAZB [info] (200) GET /api/stats/dummy.site/current-visitors took 24ms
14:46:09.396 request_id=F8MRLUDfav28LIkAAAIE [info] (202) POST /api/event took 5ms
14:46:09.501 request_id=F8MRLUDS_YhftUkAAAAD [info] (200) GET /api/stats/dummy.site/sources?period=30d&date=2024-04-04&filters=%7B%7D&with_imported=true&limit=9 took 111ms
14:46:09.508 request_id=F8MRLUDhHbK8WKUAAAah [info] (200) GET /api/stats/dummy.site/main-graph?period=30d&date=2024-04-04&filters=%7B%7D&with_imported=true&metric=visitors took 117ms
14:46:09.511 request_id=F8MRLUDS1CYntK4AAAaB [info] (200) GET /api/stats/dummy.site/entry-pages?period=30d&date=2024-04-04&filters=%7B%7D&with_imported=true&limit=9 took 121ms
14:46:09.541 request_id=F8MRLTk5sIPYSn4AABoC [info] (200) GET /api/stats/dummy.site/top-stats?period=30d&date=2024-04-04&filters=%7B%7D&with_imported=true&comparison=previous_period&compare_from=undefined&compare_to=undefined&match_day_of_week=true took 278ms
```

### DIFF
--- a/lib/plausible_web/endpoint.ex
+++ b/lib/plausible_web/endpoint.ex
@@ -69,7 +69,6 @@ defmodule PlausibleWeb.Endpoint do
 
   plug(Plug.RequestId)
   plug(PromEx.Plug, prom_ex_module: Plausible.PromEx)
-  plug(Plug.Telemetry, event_prefix: [:phoenix, :endpoint])
 
   plug(Plug.Parsers,
     parsers: [:urlencoded, :multipart, :json],
--- a/lib/plausible_web/endpoint.ex
+++ b/lib/plausible_web/endpoint.ex
@@ -69,6 +69,7 @@ defmodule PlausibleWeb.Endpoint do
 
   plug(Plug.RequestId)
   plug(PromEx.Plug, prom_ex_module: Plausible.PromEx)
+  plug(Plug.Telemetry, event_prefix: [:phoenix, :endpoint], log: false)
 
   plug(Plug.Parsers,
     parsers: [:urlencoded, :multipart, :json],
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -100,6 +100,7 @@ defmodule Plausible.Application do
 
     opts = [strategy: :one_for_one, name: Plausible.Supervisor]
 
+    setup_request_logging()
     setup_sentry()
     setup_opentelemetry()
 
@@ -173,6 +174,15 @@ defmodule Plausible.Application do
       true ->
         pool_config
     end
+  end
+
+  def setup_request_logging() do
+    :telemetry.attach(
+      "plausible-request-logging",
+      [:phoenix, :router_dispatch, :stop],
+      &Plausible.RequestLogger.log_request/4,
+      %{}
+    )
   end
 
   def setup_sentry() do
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -179,7 +179,7 @@ defmodule Plausible.Application do
   def setup_request_logging() do
     :telemetry.attach(
       "plausible-request-logging",
-      [:phoenix, :router_dispatch, :stop],
+      [:phoenix, :endpoint, :stop],
       &Plausible.RequestLogger.log_request/4,
       %{}
     )
--- a/lib/plausible/request_logger.ex
+++ b/lib/plausible/request_logger.ex
@@ -1,0 +1,19 @@
+defmodule Plausible.RequestLogger do
+  @moduledoc """
+  Custom request logger which:
+  - Includes query parameters on the same line
+  - Includes request duration on the same line
+  """
+
+  require Logger
+
+  def log_request(_, %{duration: duration}, %{conn: conn}, _) do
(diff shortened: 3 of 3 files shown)


lib/plausible_web/endpoint.ex
```diff
@@ -69,7 +69,6 @@ defmodule PlausibleWeb.Endpoint do
 
   plug(Plug.RequestId)
   plug(PromEx.Plug, prom_ex_module: Plausible.PromEx)
-  plug(Plug.Telemetry, event_prefix: [:phoenix, :endpoint])
```
macobo, line 72: @aerosol any way this can have unintended consequences on other metrics we emit? It only seems to affect logging locally.

zoldar: `opentelemetry_phoenix` (which is in turn used by `prom_ex`) does rely on endpoint events. This in turn _might_ affect Honeycomb integration for instance, though I can't tell for sure.

zoldar: Oh and we also do have a dedicated phoenix dashboard which relies on endpoint metrics quite heavily, so that would break as well.

zoldar: According to https://github.com/phoenixframework/phoenix/blob/2c5da88a2642d2249dbba4e6845222a294ba90d3/lib/phoenix/logger.ex#L10, another possible option is to disable logging specifically but still emit the event, like this:

```elixir
  plug(Plug.Telemetry, event_prefix: [:phoenix, :endpoint], log: false)
```

aerosol (commented):
> Ultimate goal is to be able to compare results with and without a flag against each other.

Could you explain what results and how are you going to compare them?

Do you mean compare request times? Is this even doable in our log browser?
 
Knowing no context, I'm thinking you could tag grafana metrics instead, with specific flag(s) setup identifier(s).

lib/plausible_web/endpoint.ex
aerosol: > any way this can have unintended consequences on other metrics we emit? It only seems to affect logging locally.

check `localhost:8000/metrics` - I think PromEx subscribes to those

macobo: Suggestion from @zoldar above works perfectly, thank you!

macobo:
> Could you explain what results and how are you going to compare them?
> Do you mean compare request times? Is this even doable in our log browser?

Not at all. Let me try explain.

So changing how our queries work on a fundamental way as https://github.com/plausible/analytics/pull/3966 does is risky - we might run into bugs, untested cases or data issues. So instead of yoloing it, I plan to:

1. Dump a bunch of queries made by users via grafana logs (which requires this PR)
2. Run them with and without the new experiment
3. Compare results, gaining confidence whether new results are identical to previous ones.

aerosol:
> Not at all. Let me try explain.

Ah got it, thanks. Wondering if annotating queries with flag label(s) would make it easier to retrieve them later on. 

Something like you attempted at https://github.com/plausible/analytics/pull/3699. I like how clever this was, modulo the OTEL dependency that gave us no lookup API.

I'm thinking that maybe we could utilize the Sentry context for this. For example currently we `include_sentry_replay_info()` in `Plausible.Stats`. Possibly similar context extension could be done for flagging the new vs old queries. I think it's retrievable at the query preparation level (via `Sentry.Context.get_all()` perhaps?). 

Another option would be to pass a flag-option to `ClickhouseRepo.all()` via main entry points like `Plausible.Stats.{breakdown,aggregate,timeseries,current_visitors}` etc. - I think those would be then retrievable at `prepare_query` as well. I haven't checked but what I'm thinking is:

```elixir
  def breakdown(site, query, prop, metrics, pagination, opts \\ []) do # opts is new
    include_sentry_replay_info()
    Breakdown.breakdown(site, query, prop, metrics, pagination, opts)
  end
``` 

then for example

```elixir
# log_flags
Stats.breakdown(site, query, "visit:source", metrics, pagination, log_flags: [...]) 
```

and

```
def prepare_query(_operation, query, opts) do
  log_flags = sanitize(Keyword.get(opts, :log_flags))
  if log_flags do
    setting = {:log_comment, log_flags}
  end
end
```

A bit more work for sure, but perhaps makes the intent clearer, and comes with a smaller risk of breaking the weaker dependency links? 

If you'll dismiss those ideas, feel free to carry on with logging query params.

zoldar (approved):

PR #3977, opened 2024-04-05
Clear `sites.imported_data` after backfilling site imports in data migration script

zoldar:
### Changes

This is one final change that should ensure import processing remains consistent after running the data migration. Otherwise, the outdated `sites.imported_data` will interfere with date range calculation for further imports until they are finally cleared. Doing it right away minimizes chance of further inconsistencies popping up.

### Tests


zoldar:
On a second thought, it's going to break existing legacy import flow which must stay operational. I'm going to close this PR and open another one where data migration adjusts imported_data along with backfilled site_import and only clears it when stats are empty.

PR #3979, opened 2024-04-05
Keep `sites.imported_data` in sync with backfilled `SiteImport` when migrating

zoldar:
### Changes

Alternative, more correct solution in place of #3977.

The legacy `imported_data` are kept in sync as the struct is still used by logic rendering import state in "Integrations" view.

Once we fully switch to the new "Imports & Exports" UI, `imported_data` will be cleared from the logic and removed completely.

### Tests

### DIFF
--- a/lib/plausible/data_migration/site_imports.ex
+++ b/lib/plausible/data_migration/site_imports.ex
@@ -11,6 +11,8 @@ defmodule Plausible.DataMigration.SiteImports do
   alias Plausible.Repo
   alias Plausible.Site
 
+  require Plausible.Imported.SiteImport
+
   def run(opts \\ []) do
     dry_run? = Keyword.get(opts, :dry_run?, true)
 
@@ -77,7 +79,8 @@ defmodule Plausible.DataMigration.SiteImports do
             "Site import #{site_import.id} (site ID #{site.id}) does not have any recorded stats. Removing it."
           )
 
-          if site_import.legacy do
+          if site.imported_data && site_import.legacy &&
+               site_import.status == Imported.SiteImport.completed() do
             clear_imported_data(site, dry_run?)
           end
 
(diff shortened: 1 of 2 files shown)


lib/plausible/data_migration/site_imports.ex
```diff
@@ -98,6 +101,13 @@ defmodule Plausible.DataMigration.SiteImports do
           |> Ecto.Changeset.change(end_date: end_date)
           |> update!(dry_run?)
 
+          if site.imported_data && site_import.legacy &&
+               site_import.status == Imported.SiteImport.completed() do
```
zoldar, line 105: Ditto about handling a case of multiple failed import attempts before successful one.

lib/plausible/data_migration/site_imports.ex
```diff
@@ -77,7 +79,8 @@ defmodule Plausible.DataMigration.SiteImports do
             "Site import #{site_import.id} (site ID #{site.id}) does not have any recorded stats. Removing it."
           )
 
-          if site_import.legacy do
+          if site.imported_data && site_import.legacy &&
+               site_import.status == Imported.SiteImport.completed() do
```
RobertJoonas, line 83: If a `SiteImport` is currently in progress, we will leave its corresponding `site.imported_data` entry untouched. Yet, we will delete the SiteImport itself.

So if I'm reading it right, this situation could leave us with a `site.imported_data` entry without a corresponding SiteImport.

Perhaps instead of checking the import status, we should make sure that no imports can be ran at the this script is being executed? 

I.e. we'll make this script wait for any in progress imports to finish, and then it can start its work. Also ensuring that while the script is being executed, all further import jobs will be scheduled with the pending status?

zoldar: Under the current legacy import flow initiated from integrations, there should be no in-progress import when there's a `completed` one, as restarting the import requires clearing all existing imports first. Additionally we have this extra lock on preventing starting another import when there's any other pending/in-progress import for a given site. So _I think_ we should be covered here 😅

RobertJoonas (approved):

PR #3986, opened 2024-04-09
Return 400 when using invalid filters for stats api

macobo:
Currently a 500 is returned instead and logged to sentry.

Long-term it would be great to unify how filters work and unify parsing and validation.


RobertJoonas (approved):
Looks good to me! 👍

Perhaps we should also document the `event:hostname` filtering ability in the [Stats API Properties section](https://plausible.io/docs/stats-api#properties)? (cc @aerosol)

macobo:
Probably - flagged under the issue: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/6925576254#__recording_7274008276

PR #3987, opened 2024-04-09
Ensure only complete imports are considered in site imports data migration

zoldar:
### Changes

This PR makes further adjustments to ensure only complete site imports and "ok" imported data are considered for migration.

### Tests

### DIFF
--- a/lib/plausible/data_migration/site_imports.ex
+++ b/lib/plausible/data_migration/site_imports.ex
@@ -20,10 +20,18 @@ defmodule Plausible.DataMigration.SiteImports do
     dry_run? = Keyword.get(opts, :dry_run?, true)
 
     site_import_query =
-      from(i in Imported.SiteImport, where: i.site_id == parent_as(:site).id, select: 1)
+      from(i in Imported.SiteImport,
+        where: i.site_id == parent_as(:site).id and i.status == ^SiteImport.completed(),
+        select: 1
+      )
 
     sites_with_imports =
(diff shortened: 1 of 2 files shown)


lib/plausible/data_migration/site_imports.ex
```diff
@@ -42,26 +50,42 @@ defmodule Plausible.DataMigration.SiteImports do
...
-
-          legacy_site_import =
-            %Imported.SiteImport{}
-            |> Ecto.Changeset.change(params)
-            |> insert!(dry_run?)
-
-          [legacy_site_import | site_imports]
-        else
-          IO.puts("Legacy site import entry for site ID #{site.id} already exists")
+        cond do
```
RobertJoonas, line 53: I think this cond clause could be avoided if we'd do things a bit differently from the start...

How about, instead of fetching `sites_with_imports` (i.e. all sites that have **either** more than 0 site_imports, **or** site.imported_data), we fetch:

1. `sites_with_site_imports` (including the ones that have `site.imported_data` too, but **at least one SiteImport**)
2. `sites_with_only_legacy_imports` (here we'd exclude those sites that have any SiteImports)

Now, in order to backfill, we can just iterate over `sites_with_only_legacy_imports` and create a `%SiteImport{legacy: true}` for all of them - no other conditions needed here.

We don't need to account for the status - they're all completed or failed. Those in progress have their SiteImports populated as we speak.

RobertJoonas (approved):

PR #3990, opened 2024-04-09
Hostnames shield

aerosol:
### Changes

This PR introduces hostnames shield - behind a `:shield_hostnames` feature flag.
Migration will be extracted to a separate PR - #3992 

The Hostname Shield is very similar to #3918 - notable differences:
  - it's an allow-list rather than block-list, no rules -> everything is allowed, some rules -> everything else is blocked
  - affects hostname filter (#3963) - non-matching hostnames won't appear in the hostname filter suggestions, once the cache propagates

Currently only /api/event payload is guarded against hostname checks, i.e. no Host header is checked. That'll maybe require a follow up. TBD. 

image

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/shield/hostname_rule.ex
```diff
@@ -0,0 +1,63 @@
...
+    # If `from_cache?` is set, the struct might be incomplete - see `Plausible.Site.Shield.Rules.IP.Cache`
+    field :from_cache?, :boolean, virtual: true, default: false
+    timestamps()
+  end
+
+  def changeset(rule \\ %__MODULE__{}, attrs) do
+    rule
+    |> cast(attrs, [:site_id, :hostname])
+    |> validate_required([:site_id, :hostname])
+    |> validate_length(:hostname, max: 250)
```
zoldar, line 27: 👍

lib/plausible/shield/hostname_rule.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Plausible.Shield.HostnameRule do
+  @moduledoc """
+  Schema for Hostnames allow list
+  """
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @type t() :: %__MODULE__{}
+
+  @primary_key {:id, :binary_id, autogenerate: true}
```
zoldar, line 10: Are binary PKs to be preferred for future schemas going forward?

lib/plausible/shield/hostname_rule.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  defp verify_valid_regex(changeset, regex) do
+    case Regex.compile(regex) do
+      {:ok, _} ->
+        put_change(changeset, :hostname_pattern, regex)
+
+      {:error, _} ->
+        add_error(changeset, :hostname, "could not compile regular expression")
+    end
+  end
```
zoldar, line 62:
```suggestion
  defp store_regex(changeset) do
    case fetch_change(changeset, :hostname) do
      {:ok, hostname} ->
        hostname
        |> build_regex()
        |> verify_and_put_regex(changeset)

      :error ->
        changeset
    end
  end

  defp build_regex(hostname) do
    regex =
      hostname
      |> Regex.escape()
      |> String.replace("\\*\\*", ".*")
      |> String.replace("\\*", ".*")

    "^#{regex}$"
  end

  defp verify_and_put_regex(regex, changeset) do
    case Regex.compile(regex) do
      {:ok, _} ->
        put_change(changeset, :hostname_pattern, regex)

      {:error, _} ->
        add_error(changeset, :hostname, "could not compile regular expression")
    end
  end
```

WDYT? `fetch_change` should be enough - `hostname` is non-null and required already.

lib/plausible/shields/hostname_test.exs
```diff
@@ -0,0 +1,188 @@
+defmodule Plausible.Shields.HostnameTest do
```
zoldar, line 1: Wrong path? It's under `lib/`

lib/plausible_web/live/shields/hostname_rules.ex
```diff
@@ -0,0 +1,308 @@
...
+  alias PlausibleWeb.Live.Components.Modal
+  alias Plausible.Shields
+  alias Plausible.Shield
+
+  import PlausibleWeb.ErrorHelpers
+
+  def update(assigns, socket) do
+    socket =
+      socket
+      |> assign(
```
zoldar, line 18: Perhaps `assign_new` would be more appropriate for those functions that short-circuit assigns and socket.assigns?

lib/plausible/shield/hostname_rule.ex
aerosol: Not exactly, IIRC I asked people for their preferences and there was no strong opinions either way, so I went ahead with UUIDs.

lib/plausible/shields/hostname_test.exs
aerosol: Oh :hankey:

lib/plausible_web/live/shields/hostname_rules.ex
aerosol: I don't think this the short-circuit is even necessary, must've been some brain fart, checking

lib/plausible/shields/hostname_test.exs
aerosol: f5774f4b2 for future me

lib/plausible_web/live/shields/hostname_rules.ex
aerosol: https://github.com/plausible/analytics/pull/3990/commits/51037a5e491f92ef163e4db2c317eb5e9ce12d20

zoldar (approved):

ukutaht (commented):
Note about the UI, specifically the form for adding a hostname to the list:

Screenshot 2024-04-16 at 15 13 58

I think that the text at the bottom should depend on whether there are any hostnames configured or not. It would be misleading to tell the user that we _will start blocking_ hostnames when they expand an existing allowlist.

When the user has no hostnames configured: _"NB: Once added, we will start rejecting traffic from non-matching hostnames within a few minutes."_

When the user has N >= 1 hostnames configured: _"Once added, we will start accepting traffic from this hostname within a few minutes."_

Also I think it's best to use `*.example.com` in the help text. It's a more universal and realistic placeholder than `*.dummy.site`

ukutaht:
Screenshot 2024-04-16 at 15 29 07
 
I don't see the redundancy help text anywhere, even on hover..

ukutaht (approved):
Code looks good 👍

aerosol:
> Also I think it's best to use *.example.com in the help text. It's a more universal and realistic placeholder than *.dummy.site

It shows `@site.domain` in fact, so the most realistic it can get.

PR #3998, opened 2024-04-10
Move hostname filter appending to Breakdown module

aerosol:
We currently update the query filtered by hostname to its respective visit props in some cases.
This patch moves it down, from controllers, to the Breakdown module, so any changes in logic will be also reflected in the Stats API.

### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -195,6 +195,7 @@ defmodule Plausible.Stats.Breakdown do
   end
 
   def breakdown(site, query, property, metrics, pagination, opts) do
+    query = maybe_update_breakdown_filters(property, query)
     if !Keyword.get(opts, :skip_tracing), do: trace(query, property, metrics)
 
     metrics_to_select = Util.maybe_add_visitors_metric(metrics) -- @computed_metrics
@@ -209,6 +210,32 @@ defmodule Plausible.Stats.Breakdown do
         breakdown_events(site, query, property, metrics_to_select)
         |> maybe_add_group_conversion_rate(&breakdown_events/4, site, query, property, metrics)
         |> paginate_and_execute(metrics, pagination)
+    end
+  end
+
+  defp maybe_update_breakdown_filters("visit:source", query) do
+    update_hostname(query, "visit:entry_page_hostname")
+  end
+
+  defp maybe_update_breakdown_filters("visit:entry_page", query) do
+    update_hostname(query, "visit:entry_page_hostname")
+  end
+
+  defp maybe_update_breakdown_filters("visit:exit_page", query) do
+    update_hostname(query, "visit:exit_page_hostname")
+  end
+
+  defp maybe_update_breakdown_filters(_, query) do
+    query
+  end
+
+  defp update_hostname(query, visit_prop) do
+    case query.filters["event:hostname"] do
+      nil ->
+        query
+
+      some ->
+        Plausible.Stats.Query.put_filter(query, visit_prop, some)
     end
   end
 
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -458,13 +458,6 @@ defmodule PlausibleWeb.Api.StatsController do
     query = Query.from(site, params)
     pagination = parse_pagination(params)
 
-    query =
-      if query.filters["event:hostname"] do
-        Query.put_filter(query, "visit:entry_page_hostname", query.filters["event:hostname"])
-      else
-        query
-      end
-
     extra_metrics =
       if params["detailed"], do: [:bounce_rate, :visit_duration], else: []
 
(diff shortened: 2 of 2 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -212,6 +213,32 @@ defmodule Plausible.Stats.Breakdown do
     end
   end
 
+  defp maybe_update_breakdown_filters("visit:source", query) do
```
macobo, line 216: Probably will need to be removed long-term but that's a job for another project

aerosol:
@macobo this is transparent and already tested via controller tests submitted with #3963 - no failures in existing tests means there's no regression.

I think going forward we should extensively test the `Plausible.Stats` delegations in isolation, regardless of where the input is coming from - since it's the most critical point of contact but is never tested directly. Out of scope here however.

macobo (approved):

ukutaht:
Just noticed we only massage the filters for `entry_page`, `exit_page` and `source` breakdowns. I think the same treatment should be applied to `utm_*` params which are set when the session starts. Otherwise we'll be showing these properties 'accross hostnames' so to speak.

aerosol:
@ukutaht thanks! on it

ukutaht:
Also `visit:referrer` :)

aerosol:
@ukutaht I can't seem to write a failing test, I think in those cases the `events_v2.hostname` filtering is enough?


```elixir
  describe "UTM parameters with hostname filter" do
    setup [:create_user, :log_in, :create_new_site]

    for {resource, attr} <- [
          utm_campaigns: :utm_campaign,
          utm_sources: :utm_source,
          utm_terms: :utm_term,
          utm_contents: :utm_content
        ] do
      test "returns #{resource} when filtered by hostname", %{conn: conn, site: site} do
        populate_stats(site, [
          build( # ad-driven user at another host
            :pageview,
            [
              {unquote(attr), "ad"},
              {:user_id, @user_id + 1},
              {:hostname, "two.example.com"},
              {:timestamp, ~N[2021-01-01 00:00:00]}
            ]
          ),
          build( # ad-driven session at target host
            :pageview,
            [
              {unquote(attr), "ad"},
              {:user_id, @user_id},
              {:hostname, "one.example.com"},
              {:timestamp, ~N[2021-01-01 00:00:00]}
            ]
          ),
          build( # continuation
            :pageview,
            [
              {unquote(attr), "ad"},
              {:user_id, @user_id},
              {:hostname, "one.example.com"},
              {:timestamp, ~N[2021-01-01 00:15:00]}
            ]
          ),
          build( # another ad-driven user at another host
            :pageview,
            [
              {unquote(attr), "ad2"},
              {:user_id, @user_id},
              {:hostname, "two.example.com"},
              {:timestamp, ~N[2021-01-01 10:16:00]}
            ]
          ),
          build(:pageview, hostname: "one.example.com", timestamp: ~N[2021-01-01 00:00:00]), # no utm
          build(:pageview, hostname: "two.example.com", timestamp: ~N[2021-01-01 00:00:00]) # no utm
        ])

        filters = Jason.encode!(%{hostname: "one.example.com"})

        conn =
          get(
            conn,
            "/api/stats/#{site.domain}/#{unquote(resource)}?period=day&date=2021-01-01&filters=#{filters}"
          )

        assert json_response(conn, 200) == [
                 %{
                   "name" => "ad",
                   "visitors" => 1,
                   "bounce_rate" => 0,
                   "visit_duration" => 900.0
                 }
               ]
      end
    end
  end
```

ukutaht:
Hmm, I don't think the test setup is describing what I meant. Need a session that starts on `two.example.com` with an utm_param and continues  on `one.example.com`. In that case `ad` will be included in breakdown when it should not. Something like this:

```elixir
  describe "UTM parameters with hostname filter" do
    setup [:create_user, :log_in, :create_new_site]

    for {resource, attr} <- [
      utm_campaigns: :utm_campaign,
      utm_sources: :utm_source,
      utm_terms: :utm_term,
      utm_contents: :utm_content
    ] do
      test "returns #{resource} when filtered by hostname", %{conn: conn, site: site} do
        populate_stats(site, [
          build( # session starts at two.example.com with utm_param=ad
            :pageview,
            [
              {unquote(attr), "ad"},
              {:user_id, @user_id},
              {:hostname, "two.example.com"},
              {:timestamp, ~N[2021-01-01 00:00:00]}
            ]
          ),
          build( # session continues on one.example.com without any utm_params
            :pageview,
            [
              {:user_id, @user_id},
              {:hostname, "one.example.com"},
              {:timestamp, ~N[2021-01-01 00:15:00]}
            ]
          )
        ])

        filters = Jason.encode!(%{hostname: "one.example.com"})

        conn =
          get(
            conn,
            "/api/stats/#{site.domain}/#{unquote(resource)}?period=day&date=2021-01-01&filters=#{filters}"
          )

        assert json_response(conn, 200) == []  # nobody landed on one.example.com from utm_param=ad
      end
    end
  end

```

aerosol:
Ah yes, that makes sense, thanks!

PR #4002, opened 2024-04-11
redirect to s3 url when downloading exports

ruslandoga:
### Changes

https://3.basecamp.com/5308029/buckets/35871979/card_tables/cards/6945042601#__recording_7282802614

### Tests
- [x] try it out, including "save as..."
- [x] add some tests (email template)

### DIFF
--- a/lib/plausible/s3.ex
+++ b/lib/plausible/s3.ex
@@ -93,7 +93,7 @@ defmodule Plausible.S3 do
 
   @doc """
   Returns a presigned URL to download the exported Zip archive from S3.
-  The URL expires in 24 hours.
+  The URL expires in 300 seconds, which should be enough for a redirect.
 
   In the current implementation the bucket always goes into the path component.
   """
@@ -101,11 +101,8 @@ defmodule Plausible.S3 do
   def download_url(s3_bucket, s3_path) do
     config = ExAws.Config.new(:s3)
 
-    # ex_aws_s3 doesn't allow expires_in longer than one week
-    one_week = 60 * 60 * 24 * 7
-
     {:ok, download_url} =
-      ExAws.S3.presigned_url(config, :get, s3_bucket, s3_path, expires_in: one_week)
+      ExAws.S3.presigned_url(config, :get, s3_bucket, s3_path, expires_in: 300)
 
     download_url
   end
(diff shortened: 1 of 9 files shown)


lib/plausible/s3.ex
```diff
@@ -93,19 +93,16 @@ defmodule Plausible.S3 do
 
   @doc """
   Returns a presigned URL to download the exported Zip archive from S3.
-  The URL expires in 24 hours.
+  The URL expires in 300 seconds.
```
ruslandoga, line 96: GitHub uses 300 seconds, should be enough for a redirect.

test/plausible/imported/csv_importer_test.exs
```diff
@@ -488,6 +488,16 @@ defmodule Plausible.Imported.CSVImporterTest do
...
+      # check mailbox
+      assert_receive {:delivered_email, email}, 5000
+      assert email.to == [{user.name, user.email}]
+
+      assert email.html_body =~
+               ~s[Please click <a href="http://localhost:8000/#{URI.encode_www_form(site.domain)}/download/export">here</a> to start the download process.]
+
       # download archive
       on_full_build do
         ExAws.request!(
```
ruslandoga, line 503: I think I'll be moving this e2e test to somewhere where we have access to `conn` (probably site_controller_test) to make use of the actual download link. I'll do that in a separate PR which would also start using staging CSV fixtures.

zoldar (approved):

test/plausible/imported/csv_importer_test.exs
```diff
@@ -488,6 +490,16 @@ defmodule Plausible.Imported.CSVImporterTest do
 
       assert %{success: 1} = Oban.drain_queue(queue: :analytics_exports, with_safety: false)
 
+      assert %{success: 1} =
+               Oban.drain_queue(queue: :notify_exported_analytics, with_safety: false)
+
+      # check mailbox
+      assert_receive {:delivered_email, email}, _within = :timer.seconds(55)
```
zoldar, line 497: I have tried reproducing this problem locally a number of times but I can't 🤔 

image

ruslandoga: Great! I set the wait time to five seconds and I can't reproduce the problem anymore either, including using the seeds for which the tests were failing on the CI

- https://github.com/plausible/analytics/actions/runs/8801619037/job/24155414311 - `mix test --include slow --include minio --seed 102492` all green
- https://github.com/plausible/analytics/actions/runs/8801503492/job/24155010928 - `mix test --include slow --include minio --seed 567429` all green
- https://github.com/plausible/analytics/actions/runs/8801439695/job/24154779192 this one was using the default 100ms, but still - `mix test --include slow --include minio --seed 292237` all green

Computers...

PR #4006, opened 2024-04-12
Remove references to `site.imported_data`

zoldar:
### Changes

Since all legacy site imports have by now respective `SiteImport` entry created, we can get rid of remaining references.

What's left is basically only embedded schema definition needed for the migration script to operate.

One additional change is considering pre-existing ID 0 import data when displaying number of pageview counts for legacy imports in "Imports & Exports" view.

### Tests

### DIFF
--- a/lib/plausible_web/controllers/site_controller.ex
+++ b/lib/plausible_web/controllers/site_controller.ex
@@ -237,8 +237,10 @@ defmodule PlausibleWeb.SiteController do
         Plausible.Google.API.fetch_verified_properties(site.google_auth)
       end
 
+    legacy_import = Plausible.Imported.get_legacy_import(site)
+
     imported_pageviews =
-      if site.imported_data do
+      if legacy_import do
         Plausible.Stats.Clickhouse.imported_pageview_count(site)
       else
         0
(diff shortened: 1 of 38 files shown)


lib/plausible_web/controllers/site_controller.ex
```diff
@@ -646,34 +649,22 @@ defmodule PlausibleWeb.SiteController do
...
-        Plausible.Purge.delete_imported_stats!(site_import)
-
-        Plausible.Repo.delete!(site_import)
+      Plausible.Purge.delete_imported_stats!(site_import)
 
-        if site_import.legacy do
-          Plausible.Purge.delete_imported_stats!(site, 0)
+      if site_import.legacy do
+        Plausible.Purge.delete_imported_stats!(site, 0)
+      end
```
RobertJoonas, line 665: Perhaps let's move this logic into `Plausible.Purge.delete_imported_stats!/1`, as it's always supposed to happen together in a single action anyway?

RobertJoonas (approved):
Looks good to me! 👍

I also wanted to note that we should probably hide the "Google Analytics Data Import" section from Site Settings > Integrations view when the feature flag is enabled. Once we release in a private preview, I think we wouldn't want to expose two different places for GA imports.

RobertJoonas:
Hmm, another thing - I understand removing the `site.imported_data` schema is impossible because we have the migration and the tests for that migration. But how about adding some deprecation comments in the `site` schema and also in the `imported_data` embedded schema module as well?

lib/plausible_web/controllers/site_controller.ex
zoldar: https://github.com/plausible/analytics/pull/4006/commits/c7e97a2760cda6502530a281d3fa926a3c7f7788

PR #4010, opened 2024-04-15
Debug queries for super-admins

aerosol:
### Changes

- For superadmins, clikchouse queries are visible in the JS console
- All queries are logged with sentry metadata
- Queries included in sentry reports

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/clickhouse_repo.ex
```diff
@@ -12,17 +12,45 @@ defmodule Plausible.ClickhouseRepo do
     end
   end
 
+  def prepare_query(operation, query, opts) do
```
ruslandoga, line 15: It might be not immediately obvious where this comes from. Maybe we can add something like `@impl true` to indicate that it overrides the default `prepare_query` callback?

ruslandoga, line 15:
```suggestion
  @impl Ecto.Repo
  def prepare_query(operation, query, opts) do
```

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -13,6 +13,26 @@ defmodule PlausibleWeb.Api.StatsController do
   @revenue_metrics on_full_build(do: Plausible.Stats.Goal.Revenue.revenue_metrics(), else: [])
 
   plug(:date_validation_plug)
+  plug(:test)
```
aerosol, line 16: will be renamed of course

aerosol:
will add relevant tests once the PoC is approved

test/plausible/clickhouse_repo_test.exs
```diff
@@ -0,0 +1,89 @@
...
+
+    Plausible.ClickhouseRepo.one(from(u in "sessions_v2", select: true, limit: 0),
+      debug_label: "log_one"
+    )
+
+    assert [
+             [q1, c1],
+             [q2, c2]
+           ] =
+             eventually(
```
macobo, line 48: Note: you can do [`SYSTEM FLUSH LOGS`](https://clickhouse.com/docs/en/sql-reference/statements/system#flush-logs) which will speed up this test

lib/plausible/clickhouse_repo.ex
```diff
@@ -12,17 +12,59 @@ defmodule Plausible.ClickhouseRepo do
...
+        opts[:debug_label] || "unlabelled"
+      )
+    end
+
+    sentry_context = Sentry.Context.get_all()
+
+    setting =
+      {:log_comment,
+       Jason.encode!(%{
+         user_id: sentry_context[:user][:id],
```
macobo, line 29: Please also include site_id.

lib/plausible/clickhouse_repo.ex
```diff
@@ -12,17 +12,59 @@ defmodule Plausible.ClickhouseRepo do
...
+        to_inline_sql(operation, query),
+        opts[:debug_label] || "unlabelled"
+      )
+    end
+
+    sentry_context = Sentry.Context.get_all()
+
+    setting =
+      {:log_comment,
+       Jason.encode!(%{
```
macobo, line 28: I wonder what's the best way to extend this going forward.

`query` object has a bunch of useful information about the query that's really useful for debugging: flags, time ranges, periods, whether imports are included, etc. Main thing it doesn't include is metrics and I'm hoping to change that soon.

Do you think as a follow-up we could evolve this to allow `ClickhouseRepo.all(metadata: query.metadata, label: label)`?

lib/plausible/clickhouse_repo.ex
```diff
@@ -12,17 +12,59 @@ defmodule Plausible.ClickhouseRepo do
     end
   end
 
+  @impl Ecto.Repo
+  def prepare_query(operation, query, opts) do
+    if Plausible.DebugReplayInfo.super_admin?() do
+      Plausible.DebugReplayInfo.track_query(
+        to_inline_sql(operation, query),
+        opts[:debug_label] || "unlabelled"
```
macobo, line 20: `debug_label` is a mouthful. Suggestion: `label` conveys the same meaning, let's use that.

lib/plausible/clickhouse_repo.ex
aerosol: > Do you think as a follow-up we could evolve this to allow ClickhouseRepo.all(metadata: query.metadata, label: label)?

Absolutely, here we can merge whatever extra is provided.

aerosol: Would you like the `metadata` to be a root-level key for these or shall we merge it flat? I'll make room for it

lib/plausible/clickhouse_repo.ex
aerosol: https://github.com/plausible/analytics/pull/4010/commits/44fd153d10597efe66b0c920a8680ca69c00cd3c

lib/plausible/clickhouse_repo.ex
aerosol: Went for a root-level `metadata` key via https://github.com/plausible/analytics/pull/4010/commits/faab04e3b8195703258f71e77480755311c58d4a

test/plausible/clickhouse_repo_test.exs
aerosol: https://github.com/plausible/analytics/pull/4010/commits/faab04e3b8195703258f71e77480755311c58d4a

macobo (approved):

mix.exs
```diff
@@ -69,7 +69,7 @@ defmodule Plausible.MixProject do
       {:bamboo_mua, "~> 0.1.4"},
       {:bcrypt_elixir, "~> 3.0"},
       {:bypass, "~> 2.1", only: [:dev, :test, :ce_test]},
-      {:ecto_ch, "~> 0.3"},
+      {:ecto_ch, github: "plausible/ecto_ch", branch: "pretty-sql"},
```
ruslandoga, line 72: Should we merge that branch and cut a new release?

aerosol: That would be lovely

ruslandoga: https://github.com/plausible/analytics/pull/4054

PR #4011, opened 2024-04-15
use custom email template for csv imports

ruslandoga:
### Changes

This PR adds a new email template for CSV imports and starts using it in CSVImporter instead of the GA one.

### DIFF
--- a/lib/plausible/imported/csv_importer.ex
+++ b/lib/plausible/imported/csv_importer.ex
@@ -12,9 +12,8 @@ defmodule Plausible.Imported.CSVImporter do
   @impl true
   def label(), do: "CSV"
 
-  # NOTE: change it once CSV import is implemented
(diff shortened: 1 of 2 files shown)


lib/plausible/imported/csv_importer.ex
```diff
@@ -12,9 +12,8 @@ defmodule Plausible.Imported.CSVImporter do
   @impl true
   def label(), do: "CSV"
 
-  # NOTE: change it once CSV import is implemented
   @impl true
-  def email_template(), do: "google_analytics_import.html"
+  def email_template(), do: "csv_import.html"
```
ruslandoga, line 16: I'm adding tests for everything I can think of in a separate branch / PR.

zoldar (approved):

PR #4012, opened 2024-04-15
Stop importing page referrer from GA4

zoldar:
### Changes

This PR removes `pageReferrer` from GA4 import. As this is en event-scoped dimension, combining it with session-scoped dimensions results in overcounting due to explosion of combinations (which also means that the results contain significant overlap between sessions and can't be added for calculating aggregates). See also: https://github.com/plausible/analytics/pull/4014/commits/4cbd29789cfaef1f80a7ac42680b1c5eb535790d

### Tests

### DIFF
--- a/fixture/ga4_report_imported_sources.json
+++ b/fixture/ga4_report_imported_sources.json
@@ -20,9 +20,6 @@
         },
         {
           "name": "sessionGoogleAdsKeyword"
-        },
-        {
-          "name": "pageReferrer"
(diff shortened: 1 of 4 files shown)


ukutaht (approved):
> As this is en event-scoped dimension, combining it with session-scoped dimensions results 

Ah, makes sense 👍

RobertJoonas (approved):
👍

PR #4019, opened 2024-04-17
Rely on con_cache telemetry

aerosol:
Now that https://github.com/sasa1977/con_cache/pull/76 is released, we don't have to use low-level operations to emit hit/miss events.

This PR also wraps cache processes with
a function returning appropriate child specs lists.

Ideally each cache will have its own supervisor/child specs going forward. This is an intermediate step in that direction.


lib/plausible/application.ex
```diff
@@ -24,89 +24,62 @@ defmodule Plausible.Application do
...
+        adapter_opts: [ttl_check_interval: :timer.seconds(1), global_ttl: :timer.minutes(30)]
+      ),
+      cache(Plausible.Site.Cache,
+        adapter_opts: [ttl_check_interval: false],
+        warmers: [
+          refresh_all:
+            {Plausible.Site.Cache.All,
+             interval: :timer.minutes(15) + Enum.random(1..:timer.seconds(10))},
+          refresh_updated_recently:
+            {Plausible.Site.Cache.ReventlyUpdated, interval: :timer.seconds(30)}
```
zoldar, line 40:
```suggestion
            {Plausible.Site.Cache.RecentlyUpdated, interval: :timer.seconds(30)}
```

🙈

lib/plausible/application.ex
```diff
@@ -222,4 +195,31 @@ defmodule Plausible.Application do
...
+
+  defp cache(id_or_module, opts) do
+    warmers = Keyword.get(opts, :warmers)
+
+    if is_nil(warmers) do
+      Plausible.Cache.Adapter.child_spec(
+        id_or_module,
+        String.to_atom("cache_#{id_or_module}"),
+        Keyword.fetch!(opts, :adapter_opts)
+      )
```
zoldar, line 207: Are these `id`-based cases going to be eventually turned into module with `use Plausible.Cache`? If not, maybe the original definitions should be kept intact?

zoldar (changes requested):

lib/plausible/application.ex
zoldar: I'm not sure, maybe that's adding too much complexity for little benefit? IMHO we could keep id-based definitions as they were and rename the wrapper function from `cache` to `warmed_cache` (?) and require `warmers` as an argument.

Currently the code is giving impression that the warmer-based cache supports "id" caches, while it won't as `cache_impl` expects a module.

aerosol: Gotcha, on it

aerosol: f86e5ddfa

zoldar (approved):
✨

PR #4020, opened 2024-04-18
Add link to settings on the "Waiting for first pageview" screen

metmarkosaric:
we've had people ask how to start by importing their data first before integrating Plausible into their site (and some also want just to import and store their GA stats in Plausible rather than use Plausible to collect new stats) so the idea here is to give people access to the site settings area directly from the "Waiting for first pageview" screen

### DIFF
--- a/lib/plausible_web/templates/stats/waiting_first_pageview.html.heex
+++ b/lib/plausible_web/templates/stats/waiting_first_pageview.html.heex
@@ -44,10 +44,10 @@
         <br />
         <span :if={full_build?()}>
           Check the
-              <.styled_link href={Routes.site_path(@conn, :settings_general, @site.domain)}>
-                site settings
-              </.styled_link>
-              to invite team members, import historical stats and more.
+          <.styled_link href={Routes.site_path(@conn, :settings_general, @site.domain)}>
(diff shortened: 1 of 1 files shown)


lib/plausible_web/templates/stats/waiting_first_pageview.html.heex
```diff
@@ -43,9 +43,11 @@
 
         <br />
         <span :if={full_build?()}>
-          Still not working?
-          <.styled_link new_tab href="https://plausible.io/contact">Contact us</.styled_link>
-          and we will help you with your setup
+          Check the
```
ukutaht, line 46: Checking just in case it wasn't - is removing the contact link intentional?

metmarkosaric: @ukutaht yeah I thought it's better in order to not make that page too noisy. for people to whom our troubleshooting docs doesn't help, they will find a way to contact us

zoldar:
image

Added a linebreak before "import historical stats and more." because otherwise the formatting looked weird with first line much shorter than the second.

zoldar (approved):

PR #4022, opened 2024-04-18
add some more time on page tests

ruslandoga:
From discussions.


ruslandoga:
Closing as this was used as an example only.

PR #4024, opened 2024-04-18
Fix main graph intervals

RobertJoonas:
### Changes

This PR fixes a regression caused by https://github.com/plausible/analytics/pull/3936 where invalid intervals were returned for the custom period. It also makes the default intervals more sensible.

### DIFF
--- a/assets/js/dashboard/stats/graph/interval-picker.js
+++ b/assets/js/dashboard/stats/graph/interval-picker.js
@@ -4,7 +4,6 @@ import React, { Fragment, useCallback, useEffect } from 'react';
 import classNames from 'classnames'
 import * as storage from '../../util/storage'
 import { isKeyPressed } from '../../keybinding.js'
-import { monthsBetweenDates } from '../../util/date.js'
 
 const INTERVAL_LABELS = {
   'minute': 'Minutes',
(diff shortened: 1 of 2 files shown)


assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -15,13 +14,33 @@ const INTERVAL_LABELS = {
 }
 
 function validIntervals(site, query) {
-  if (query.period === "custom" && monthsBetweenDates(query.from, query.to) > 12) {
-    return ["week", "month"]
+  if (query.period === 'custom') {
+    if (query.to.diff(query.from, 'days') < 7) {
```
aerosol, line 18: The fix looks good. @metmarkosaric [mentioned](https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7298629708#__recording_7304694988) we could go by 

<30 days => date
1-6 mo => week

aerosol (approved):

assets/js/dashboard/stats/graph/interval-picker.js
RobertJoonas: Makes sense! https://github.com/plausible/analytics/pull/4024/commits/af8f51c5db93c85235399a0f58b507a1dde8960e

PR #4025, opened 2024-04-18
Make final touches to Imports & Exports

zoldar:
### Changes

* Remove legacy import flow entirely
* Change meaning of `imports_exports` flag:
  * when enabled, GA4 properties are shown in property/view selection step and max number of imports is at current 5
  * when disabled, only UA views are shown in property/view selection step and max number of allowed imports is at 1

### Tests
- [x] Automated tests have been revised


RobertJoonas (approved):
Looking good! 👍

PR #4033, opened 2024-04-23
Add support for imported custom events

zoldar:
### Changes

### Tests

### Changelog

### Documentation


priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
```diff
@@ -0,0 +1,40 @@
+defmodule Plausible.IngestRepo.Migrations.AddImportedCustomEvents do
```
zoldar, line 1: @macobo Your  👀  on this migration would be very appreciated 🙏

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
```diff
@@ -0,0 +1,40 @@
...
+        SETTINGS index_granularity = 8192, replicated_deduplication_window = 0
+        """
+      end
+
+    execute """
+            CREATE TABLE IF NOT EXISTS imported_custom_events #{on_cluster}
+                (
+                `site_id` UInt64,
+                `import_id` UInt64,
+                `date` Date,
```
macobo, line 28: Thoughts on using `DateTime CODEC(Delta(4), LZ4)`?

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
```diff
@@ -0,0 +1,40 @@
...
+  def change do
+    # NOTE: Using another table for determining cluster presence
+    on_cluster = Plausible.MigrationUtils.on_cluster_statement("imported_pages")
+
+    settings =
+      if Plausible.MigrationUtils.clustered_table?("imported_pages") do
+        """
+        ENGINE = ReplicatedMergeTree('/clickhouse/{cluster}/tables/{shard}/{database}/imported_custom_events', '{replica}')
+        ORDER BY (site_id, import_id, date, name)
+        SETTINGS index_granularity = 8192, replicated_deduplication_window = 0, storage_policy = 'tiered'
```
macobo, line 13: This storage policy is cloud-specific. Let's make extra merge tree table settings e.g. an env var?

zoldar: WDYM? We do already use different settings based on whether the target database has clustered imported tables or not (https://github.com/plausible/analytics/pull/4033/files#diff-e2d070d31359de2724eeb4871d7d3a9659c2e11f78471d004b4a589b1fbc0d3bR9)

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
zoldar: Given it's a `Date` field, is using `DateTime` instead intentional?

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
macobo: I meant `storage_policy = 'tiered'` which is a cloud-specific setting. Other clickhouse clusters (even clustered) won't have that storage policy configured.

I suggest replacing this line with

```
        SETTINGS replicated_deduplication_window = 0, #{extra_settings}
```

Note index_granularity = 8192 is the default and doesn't need to get specified.

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
macobo: Oh, just saw that every other table already uses `Date`. Ignore this comment.

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
RobertJoonas: Answered in #4076

lib/plausible/stats/goal_suggestions.ex
```diff
@@ -0,0 +1,70 @@
...
+        where: i.date >= ^query.date_range.first and i.date <= ^query.date_range.last,
+        where: i.visitors > 0,
+        where: fragment("? ilike ?", i.name, ^matches),
+        where: i.name not in ^excluded,
+        select: %{
+          name: i.name,
+          visitors: selected_as(sum(i.visitors), :visitors)
+        },
+        group_by: i.name,
+        limit: 25
```
macobo, line 57: You likely want an explicit order by as well

lib/plausible/stats/goal_suggestions.ex
```diff
@@ -0,0 +1,70 @@
...
+      from(i in "imported_custom_events",
+        where: i.site_id == ^site.id,
+        where: i.import_id in ^site.complete_import_ids,
+        where: i.date >= ^query.date_range.first and i.date <= ^query.date_range.last,
+        where: i.visitors > 0,
+        where: fragment("? ilike ?", i.name, ^matches),
+        where: i.name not in ^excluded,
+        select: %{
+          name: i.name,
+          visitors: selected_as(sum(i.visitors), :visitors)
```
macobo, line 54: Does this lead to double-counting visitors who do multiple custom events?

lib/plausible/stats/goal_suggestions.ex
zoldar: Do you mean that as an optimization? Because the top level query already does the ordering at https://github.com/plausible/analytics/pull/4033/files#diff-6c996b6c3d6a7231d4844a073854796d35b17432a527cf8838201586870bbae9R64

lib/plausible/stats/goal_suggestions.ex
zoldar: It might but the imported data is already aggregated so it's impossible to determine those cases of double-counting.

lib/plausible/stats/goal_suggestions.ex
macobo: Yes, but you are limiting number of rows here. It will be a random 25 rows that's returned from this subquery.

lib/plausible/stats/goal_suggestions.ex
macobo: I think it might be wiser not to export visitor numbers at all rather than wildly inflated numbers. But ymmv, just a drive-by comment.

RobertJoonas: > I think it might be wiser not to export visitor numbers at all rather than wildly inflated numbers. But ymmv, just a drive-by comment.

What do you mean? We're only talking about querying the imported data here. Not exporting it.

RobertJoonas (approved):

lib/plausible/stats/goal_suggestions.ex
RobertJoonas: @macobo we actually extracted the suggestions changes into [another PR](https://github.com/plausible/analytics/pull/4092), so marking this conversation as resolved now

PR #4038, opened 2024-04-23
Fix a test

RobertJoonas:
### Changes

Fix a test failing due to inconsistent ordering of results.


aerosol (approved):
oooor we should add ordering to the suggestions query

RobertJoonas:
> oooor we should add ordering to the suggestions query

It's already ordered by the number of `visitors` but in this test that number is 1 for all entries. I don't think we need to order by anything else. An alternative would be to populate more visitors for some hostnames in the test and still assert on a list directly, but I think the ordering doesn't really matter in this test.

PR #4049, opened 2024-04-29
Implement resumable GA4 imports to work around rate limiting

zoldar:
### Changes

image

This PR implements graceful handling of rate limiting error (HTTP code 429) returned by Google Data API (GA4). When it occurs, instead of retrying the request immediately or discarding the import entirely, another import job is scheduled for the same exact site import, resuming the import process at the exact dataset and offset position where it failed. The job is scheduled 65 minutes in the future - the hourly rate limit on the property should be reset by then. All the existing data fetched for that import so far are left intact.

In addition, all other HTTP calls to the GA Data API are returning explicit error message on hitting rate limit error so that the user is aware that it should work again when they try later.

The imports list displays a notice tooltip by in-progress import when it's taking longer than 5 minutes hinting that the import might take longer due to the amount of the data and rate limiting enforced by the Google API.

### Tests

### Dark mode

### DIFF
--- a/lib/oban_error_reporter.ex
+++ b/lib/oban_error_reporter.ex
@@ -48,7 +48,7 @@ defmodule ObanErrorReporter do
     site_import = Plausible.Repo.get(Plausible.Imported.SiteImport, import_id)
 
     if site_import do
-      Plausible.Workers.ImportAnalytics.import_fail(site_import)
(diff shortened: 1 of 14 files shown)


RobertJoonas (approved):
Looks good to me! 👍

PR #4052, opened 2024-04-30
Return session in each time bucket its active in for hourly/minute timeseries

macobo:
### Changes

After this PR:
- For timeseries, sessions are counted each time bucket its active in for hourly/minute timeseries
- `experimental_session_count` flag is tested together with `experimental_reduced_joins`

Motivation: https://3.basecamp.com/5308029/buckets/35611491/messages/7085680123

If graphing timeseries by session.timestamp (end time) we don't show sufficient data. Instead, for hourly and minute graphs we show sessions in each bucket they're active in. We're not doing this for other graphs since the error margin is not worth the extra overhead.

### Tests

### Changelog
- [x] Changelog entry will be added as the flag is enabled

### Documentation
- [x] Changelog entry will be added as the flag is enabled

### Dark mode

### DIFF
--- a/lib/plausible/stats/timeseries.ex
+++ b/lib/plausible/stats/timeseries.ex
@@ -55,7 +55,7 @@ defmodule Plausible.Stats.Timeseries do
     metrics = Util.maybe_add_visitors_metric(metrics)
 
     from(e in base_event_query(site, query), select: ^select_event_metrics(metrics))
-    |> select_bucket(site, query)
+    |> select_bucket(:events, site, query)
     |> maybe_add_timeseries_conversion_rate(site, query, metrics)
     |> Plausible.Stats.Imported.merge_imported_timeseries(site, query, metrics)
     |> ClickhouseRepo.all()
@@ -66,7 +66,7 @@ defmodule Plausible.Stats.Timeseries do
   defp sessions_timeseries(site, query, metrics) do
     from(e in query_sessions(site, query), select: ^select_session_metrics(metrics, query))
     |> filter_converted_sessions(site, query)
-    |> select_bucket(site, query)
+    |> select_bucket(:sessions, site, query)
     |> Plausible.Stats.Imported.merge_imported_timeseries(site, query, metrics)
     |> ClickhouseRepo.all()
     |> Util.keep_requested_metrics(metrics)
@@ -152,7 +152,7 @@ defmodule Plausible.Stats.Timeseries do
     date
   end
 
-  defp select_bucket(q, site, %Query{interval: "month"}) do
+  defp select_bucket(q, _table, site, %Query{interval: "month"}) do
     from(
       e in q,
       group_by: fragment("toStartOfMonth(toTimeZone(?, ?))", e.timestamp, ^site.timezone),
@@ -163,7 +163,7 @@ defmodule Plausible.Stats.Timeseries do
     )
   end
 
-  defp select_bucket(q, site, %Query{interval: "week"} = query) do
+  defp select_bucket(q, _table, site, %Query{interval: "week"} = query) do
     {first_datetime, _} = utc_boundaries(query, site)
 
     from(
@@ -174,7 +174,7 @@ defmodule Plausible.Stats.Timeseries do
     )
   end
 
-  defp select_bucket(q, site, %Query{interval: "date"}) do
+  defp select_bucket(q, _table, site, %Query{interval: "date"}) do
     from(
       e in q,
       group_by: fragment("toDate(toTimeZone(?, ?))", e.timestamp, ^site.timezone),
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/timeseries.ex
```diff
@@ -218,6 +248,35 @@ defmodule Plausible.Stats.Timeseries do
...
+          s.start,
+          ^site.timezone,
+          s.start,
+          s.timestamp,
+          ^period_in_seconds
+        ),
+      group_by: bucket,
+      order_by: bucket,
+      select_merge: %{
+        date: bucket
```
macobo, line 275: @ruslandoga there's a bug in array_join implementation causing this to error in a test when including imports:

```
Error:      test/plausible_web/controllers/api/stats_controller/main_graph_test.exs:60
         ** (Ecto.QueryError) atoms, structs, maps, lists, tuples and sources are not allowed as map values in subquery, got: `%{
           visitors:
             selected_as(
               fragment(
                 {:raw, "toUInt64(round(uniq("},
                 {:expr, &0.user_id()},
                 {:raw, ") * any(_sample_factor)))"}
               ),
               :visitors
             ),
           date: &1
         }` in query:
         
         from s0 in "sessions_v2",
           array_join: f1 in fragment(
           "timeSlots(toTimeZone(?, ?), toUInt32(timeDiff(?, ?)), toUInt32(?))",
           s0.start,
           ^"UTC",
           s0.start,
           s0.timestamp,
           ^3600
         ),
           on: true,
           where: s0.site_id == ^1747,
           where: s0.timestamp >= ^~U[2021-01-01 00:00:00Z] and s0.start < ^~U[2021-01-02 00:00:00Z],
           group_by: [f1],
           order_by: [asc: f1],
           select: %{
           visitors:
             selected_as(fragment("toUInt64(round(uniq(?) * any(_sample_factor)))", s0.user_id), :visitors),
           date: f1
         }
```

Code works well in all other cases.

It seems `bucket` in this case is a binding (?) which starts blowing up if used in subqueries as we do with imports.

Is there a simple workaround or thoughts on how to fix?

ruslandoga: Ah, it seems like you need to fragment the date. Fragment("?", f1) or similar.

ruslandoga: Note that this type erasure is required everythere array join result is selected, not just in sub queries. It's not required when a Nested type is used since it returns a table-like object on array join.

ruslandoga: I'm going to raise when table binding is selected in ecto_ch. That would make this error clearer. I also think I'll move array join to hints hack like with asof anti semi etc. and remove array_join keyword from Ecto since it's not a good fit there, too adapter specific.

ruslandoga: To be clearer to Ecto wants in that query is f1.column, not just f1.

ruslandoga: BTW, would https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier help here?

macobo: `fragment` does the trick, thanks.

ORDER BY WITH FILL does not help here - it only helps when trying to fill gaps in results with zeroes.

ruslandoga: I meant `arrayJoin([start, timestamp]) as bucket` with `order by bucket fill in with step 3600`, it produces the same output for me.

lib/plausible/stats/timeseries.ex
```diff
@@ -218,6 +248,35 @@ defmodule Plausible.Stats.Timeseries do
...
+          ^site.timezone,
+          s.start,
+          s.timestamp,
+          ^period_in_seconds
+        ),
+      group_by: bucket,
+      order_by: bucket,
+      select_merge: %{
+        date: fragment("?", bucket)
+      }
```
RobertJoonas, line 276:
```suggestion
      select_merge: %{date: bucket}
```

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -56,6 +56,7 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
       assert %{"plot" => ^zeroes, "with_imported" => false} = json_response(conn, 200)
     end
 
+    # Blows up with time slots!
```
RobertJoonas, line 59: Was this comment unintentionally commited? I don't see anything blowing up 😀

test/plausible_web/controllers/api/stats_controller/main_graph_test.exs
```diff
@@ -433,6 +434,205 @@ defmodule PlausibleWeb.Api.StatsController.MainGraphTest do
...
+                 0,
+                 0,
+                 0,
+                 0,
+                 1,
+                 0,
+                 1,
+                 0,
+                 0
+               ]
```
RobertJoonas, line 549:
```suggestion
      expected_plot =
        if FunWithFlags.enabled?(:experimental_session_count) do
          ~w[1 1 1 1 1 2 2 2 2 2 2 1 1 1 1 1 0 0 0 0 0 0 0 0 0 1 1 1 0 0]
        else
          ~w[0 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 0 0 0 0 0 1 0 1 0 0]
        end
      
      assert plot == Enum.map(expected_plot, &String.to_integer/1) 
```

RobertJoonas (approved):
Some comments inline but all in all looks good to me 👍

lib/plausible/stats/timeseries.ex
macobo: Sadly needed this - otherwise ecto would result in a compile error (see collapsed comment thread above)

PR #4053, opened 2024-04-30
Include breakdown property in the Query struct

RobertJoonas:
### Changes

This PR adds a `property` field into the `Stats.Query` struct, and uses it around the codebase instead of passing it as a separate argument to `Stats.Breakdown`.

There are no behavioural changes introduced here, except a tiny modification to tracing the `event:page` property in the `event:page` breakdown query (which I think is the desired behaviour). Instead of `visit:entry_page` which is the actual requested breakdown prop.

**Motivation for this change**

With the upcoming changes related to imported custom events and filtering by imported data, evaluating the `include_imported` field in the Query struct depends on the breakdown property.

Simply passing the `property` into the `Query.from` function and **not** persisting it in the struct is not enough. That's because we need to be able to construct a comparison Query struct based on the given source Query struct.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -5,7 +5,6 @@ defmodule Plausible.Stats.Breakdown do
 
   import Plausible.Stats.{Base, Imported}
   import Ecto.Query
-  require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{Query, Util, TableDecider}
 
   @no_ref "Direct / None"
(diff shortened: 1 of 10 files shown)


lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -13,11 +13,13 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
   def aggregate(conn, params) do
     site = Repo.preload(conn.assigns.site, :owner)
 
+    params = Map.put(params, "property", nil)
```
RobertJoonas, line 16: This explicit check is here to keep the behaviour same as it was before.

I also wrote a test for this, where I'm querying the `visit_duration` metric with the `event:props:*` property. Without explicitly setting `property` to `nil` here, the test would return a 400 with a message saying "Session metric `visit_duration` cannot be queried in a breakdown by ...".

This is not the desired error message in an endpoint that should completely ignore this parameter instead.

Everything I just said applies to Timeseries as well, which is why there are two tests.

lib/plausible/stats/breakdown.ex
```diff
@@ -130,48 +146,50 @@ defmodule Plausible.Stats.Breakdown do
...
   end
 
-  def breakdown(site, query, "event:page" = property, metrics, pagination, opts) do
+  def breakdown(site, %Query{property: "event:page"} = query, metrics, pagination, opts) do
     event_metrics =
       metrics
       |> Util.maybe_add_visitors_metric()
       |> Enum.filter(&(&1 in @event_metrics))
 
+    if !Keyword.get(opts, :skip_tracing), do: Query.trace(query, metrics)
```
RobertJoonas, line 164: This is the place mentioned in the PR description. We'll now be tracing `event:page` instead of `visit:entry_page`

zoldar (approved):
LGTM with one optional suggestion - maybe it would be worth it to create an additional clause from `Query.from` which accepts 3 arguments - `property` being the third. That would reduce the cases of mangling params coming from user input.

RobertJoonas:
>  maybe it would be worth it to create an additional clause from Query.from which accepts 3 arguments - property being the third. That would reduce the cases of mangling params coming from user input.

I opted for a `Query.from` function with 2 arguments because there are possibly more things we'd like to include in the Query struct such as `metrics` and `pagination`. Similarly, we'd have to pass those into the `Query.from` function too and the argument list would grow bigger, so I think it's better to stick to a single map of `params`.

I'm leaving it like that for now. I guess we might revisit this with the upcoming Stats API v2 work.

PR #4058, opened 2024-05-02
Add migration to add ALIAS columns to common session/visit properties

macobo:
This allows for simplifications in the API code.

Companion (App code) PR: https://github.com/plausible/analytics/pull/4059

PR #4059, opened 2024-05-02
Use ALIAS columns to simplify querying API

macobo:
### Changes

Uses https://github.com/plausible/analytics/pull/4058 to simplify some API code


RobertJoonas (approved):
👍

PR #4068, opened 2024-05-05
Remove dead code under Plausible.Google.HTTP#list_stats

macobo:
I'm lacking proper context to evaluate this code, discovered when refactoring filters.

`filters["page"]` is a typo for `filters["event:page"]`. It does not exist currently, meaning all code doesn't get referrer stats for individual pages if filtered so.

The fix seems non-trivial since the `event:page` filter might be e.g. contains or match one of many, so thought it might be best to remove this. Please do share your thoughts if you disagree!


ukutaht:
Ah, this code was an early attempt to satisfy [this feature request](https://feedback.plausible.io/202) which probably worked at some point but got broken as filters format changed.

Since this is a fairly highly requested feature I'll play around with the code a bit today to see if I can get it to work again.

macobo:
Thanks for the context. Let me know how it goes, will be keeping an eye on this as it blocks small part of the filters refactor!

ukutaht:
Very much WIP but it's functional and I wanted to share as soon as possible: https://github.com/plausible/analytics/pull/4077

Need to finish up:
* Tests which are currently very bad/incomplete
* UI messaging for filters that cannot be applied to Search Console data

PR #4070, opened 2024-05-06
Add override on CSV exports for super admins

zoldar:
### Changes

Adds a feature flag override on CSV export UI.

### DIFF
--- a/lib/plausible_web/templates/site/settings_imports_exports.html.heex
+++ b/lib/plausible_web/templates/site/settings_imports_exports.html.heex
@@ -20,7 +20,7 @@
   ) %>
 </div>
 
-<%= if FunWithFlags.enabled?(:csv_imports_exports, for: @site) do %>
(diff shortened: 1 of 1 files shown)


macobo (approved):

lib/plausible_web/templates/site/settings_imports_exports.html.heex
```diff
@@ -20,7 +20,7 @@
   ) %>
 </div>
 
-<%= if FunWithFlags.enabled?(:csv_imports_exports, for: @site) do %>
+<%= if FunWithFlags.enabled?(:csv_imports_exports, for: @site) or Plausible.Auth.is_super_admin?(@current_user) do %>
```
macobo, line 23: Q: Why not enable feature flag for all admins?

RobertJoonas: That could work as well, no difference really.

PR #3692, opened 2024-01-16
Add CSV Importer

ruslandoga:
### Changes

This PR  implements exports from `events_v2` and `sessions_v2` and imports into `imported_*` tables.
 
- [ ] check export queries, are they correct, compare with CSV report queries (name=pageview filtering, etc.), ensure that after exporting website.com and then importing it as imported.website.com, the dashboards for website.com and imported.website.com show the same statistics
- [ ] tests tests tests
- [ ] adapt to #3724 
- [ ] `import_id` and [staging tables](https://github.com/plausible/analytics/pull/3692#discussion_r1458239447)
- [ ] ~~[zstd? Native?](https://github.com/plausible/analytics/pull/3692#discussion_r1464180510)~~
- [ ] ~~flash message with success or error on import~~

### Tests
- [ ] Automated tests have been added

### Changelog
- [ ] Entry has been added to changelog

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated

### Dark mode
- [ ] The UI has been tested both in dark and light mode

PR #4074, opened 2024-05-06
Apply sampling factor multiplication consistently across CSV export queries

zoldar:
### Changes

Addresses a problem of undercounting visits when number of entries processed in the export query exceeds sampling threshold.

### DIFF
--- a/lib/plausible/exports.ex
+++ b/lib/plausible/exports.ex
@@ -277,7 +277,11 @@ defmodule Plausible.Exports do
   defmacrop visit_duration(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).duration),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
+          unquote(t).sign,
+          unquote(t).duration
+        ),
         :visit_duration
       )
     end
@@ -294,14 +298,21 @@ defmodule Plausible.Exports do
 
   defmacrop visits(t) do
     quote do
-      selected_as(sum(unquote(t).sign), :visits)
+      selected_as(
+        fragment("toUInt64(round(greatest(sum(?), 0)*any(_sample_factor)))", unquote(t).sign),
+        :visits
+      )
     end
   end
 
   defmacrop bounces(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).is_bounce),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
+          unquote(t).sign,
+          unquote(t).is_bounce
+        ),
         :bounces
       )
     end
@@ -310,7 +321,11 @@ defmodule Plausible.Exports do
   defmacrop pageviews(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).pageviews),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
+          unquote(t).sign,
+          unquote(t).pageviews
+        ),
         :pageviews
       )
     end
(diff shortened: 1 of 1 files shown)


ruslandoga (approved):

lib/plausible/exports.ex
```diff
@@ -277,7 +277,11 @@ defmodule Plausible.Exports do
   defmacrop visit_duration(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).duration),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
+          unquote(t).sign,
+          unquote(t).duration
+        ),
         :visit_duration
```
ruslandoga, line 285: I don't think visit duration needs to be multiplied.

lib/plausible/exports.ex
```diff
@@ -294,14 +298,21 @@ defmodule Plausible.Exports do
 
   defmacrop visits(t) do
     quote do
-      selected_as(sum(unquote(t).sign), :visits)
+      selected_as(
+        fragment("toUInt64(round(greatest(sum(?), 0)*any(_sample_factor)))", unquote(t).sign),
```
ruslandoga, line 302:
```suggestion
        fragment("toUInt64(round(greatest(sum(?),0)*any(_sample_factor)))", unquote(t).sign),
```

lib/plausible/exports.ex
```diff
@@ -294,14 +298,21 @@ defmodule Plausible.Exports do
...
+      )
     end
   end
 
   defmacrop bounces(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).is_bounce),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
```
ruslandoga, line 312:
```suggestion
          "toUInt32(round(greatest(sum(?*?),0)*any(_sample_factor)))",
```

ruslandoga: I was going through my old commits, and think that it should be `sum(greatest(?*?,0))`

ruslandoga, line 312:
```suggestion
          "toUInt32(round(sum(greatest(?*?,0))*any(_sample_factor)))",
```

ruslandoga: https://github.com/plausible/analytics/pull/3692#discussion_r1464206792

lib/plausible/exports.ex
```diff
@@ -310,7 +321,11 @@ defmodule Plausible.Exports do
   defmacrop pageviews(t) do
     quote do
       selected_as(
-        fragment("greatest(sum(?*?),0)", unquote(t).sign, unquote(t).pageviews),
+        fragment(
+          "toUInt64(round(greatest(sum(?*?),0)*any(_sample_factor)))",
```
ruslandoga, line 325:
```suggestion
          "toUInt64(round(sum(greatest(?*?,0))*any(_sample_factor)))",
```

lib/plausible/exports.ex
ruslandoga: Ah, no. It wouldn't work.

ruslandoga:
I think our usage of `toUInt64` is wrong. See https://play.clickhouse.com/play?user=play#c2VsZWN0IHRvVUludDgoMjU2KTs= where `select toUInt8(256);` doesn't truncate to the largest UInt8 value but instead performs something like `div/2` and returns 0 instead of 255.

We should probably replace `toUInt64` and `toUInt32` with `least` and something like `trunc`

lib/plausible/exports.ex
zoldar: Ahh, thanks, `imported_visitors` has `bounces` as `UInt64` but indeed, all other tables have it as `UInt32`.

zoldar:
@ruslandoga Yeah, what you suggest makes more sense as a failure mode 👍  - maybe we indeed should do away with casting entirely given CH (currently?) does truncation on its own.

Let's tackle that in a follow-up though. Given those aggregates are over a single day, overflow is very unlikely. Most cases of this blowing up in the past were due to forgetting to ensure the numerical value is positive (not accounting for negative outcome of `sum(sign...)`, which ended up with toUInt.. casting it to MAX - number).

PR #4076, opened 2024-05-07
Migration: add imported custom events

RobertJoonas:
### Changes

Extracted migration from #4033. Also added `path` column which will be needed later on when exporting 404 goals into CSV.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
+++ b/priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
@@ -28,6 +28,7 @@ defmodule Plausible.IngestRepo.Migrations.AddImportedCustomEvents do
                 `date` Date,
                 `name` String CODEC(ZSTD(3)),
                 `link_url` String CODEC(ZSTD(3)),
+                `path` String CODEC(ZSTD(3)),
                 `visitors` UInt64,
                 `events` UInt64
             )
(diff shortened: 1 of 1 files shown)


priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
```diff
@@ -0,0 +1,41 @@
...
+        ORDER BY (site_id, import_id, date, name)
+        SETTINGS replicated_deduplication_window = 0, storage_policy = 'tiered'
+        """
+      else
+        """
+        ENGINE = MergeTree()
+        ORDER BY (site_id, import_id, date, name)
+        SETTINGS replicated_deduplication_window = 0
+        """
+      end
```
RobertJoonas, line 21: @macobo I think there was some misunderstanding in our earlier conversation - there are two different settings that we're giving from this if block:

* ENGINE
* SETTINGS storage_policy = 'tiered'

What I was reading from your suggestion in #4033 was to make it more DRY. While the rest is the same, I think we can still keep this if block as it is. Less noisy that way. WDYT?

zoldar (approved):

priv/ingest_repo/migrations/20240423094014_add_imported_custom_events.exs
RobertJoonas: > I meant storage_policy = 'tiered' which is a cloud-specific setting

Yes - that setting is only applied for the cloud

PR #4077, opened 2024-05-07
Filtering Search Console keywords

ukutaht:
### Changes

Fixes and expands the capability to filter Search Console keywords with our integration.

I saw @macobo removing some code that was not functional in https://github.com/plausible/analytics/pull/4068. However, this code is worth fixing because it's not that complicated while being a [highly requested feature](https://feedback.plausible.io/202).

I was able to fix the feature and improve the following:
* All page filter types are supported (`is`, `member`, `matches`, `matches_member`)
* Alias `event:page` and `visit:entry_page` filters for the purpose of keyword data
* Filter Search Console data by device
* Filter Search Console data by country
* Show a message to the user if the set of filters cannot be requested from Search Console (currently we just show unfiltered dataset which can be misleading)

### Demo (In which I find a typo that has since been fixed)

EDIT: Since recording this @aerosol and I have also added impression, CTR and position data to the keyword modal

https://github.com/plausible/analytics/assets/3731516/55e82190-a18f-4e83-ae96-3e01536f1b8a

### Tests

### Changelog

### Documentation
- [x] Docs do not need to be changed

### Dark mode


lib/plausible/google/api.ex
```diff
@@ -74,21 +75,27 @@ defmodule Plausible.Google.API do
...
-             filters["page"]
+             search_console_filters
            ) do
       stats
       |> Map.get("rows", [])
       |> Enum.filter(fn row -> row["clicks"] > 0 end)
       |> Enum.map(fn row -> %{name: row["keys"], visitors: round(row["clicks"])} end)
       |> then(&{:ok, &1})
+    else
+      :google_property_not_configured -> {:err, :google_property_not_configured}
```
aerosol, line 96: err?

lib/plausible/google/api.ex
```diff
@@ -74,21 +75,26 @@ defmodule Plausible.Google.API do
   end
 
   def fetch_stats(site, %{filters: %{} = filters, date_range: date_range}, limit) do
-    with site <- Plausible.Repo.preload(site, :google_auth),
+    with {:ok, site} <- ensure_search_console_property(site),
          {:ok, access_token} <- maybe_refresh_token(site.google_auth),
+         {:ok, search_console_filters} <-
+           SearchConsole.Filters.transform(site.google_auth.property, filters),
```
macobo, line 81: This is going to cause an annoying conflict with https://github.com/plausible/analytics/pull/4082. I propose we merge this before that PR since I have an idea on how to fix it.

lib/plausible/google/http.ex
```diff
@@ -39,26 +39,18 @@ defmodule Plausible.Google.HTTP do
...
-
-    filter_groups =
-      if page do
-        url = property_base_url(property)
-        [%{filters: [%{dimension: "page", expression: "https://#{url}#{page}"}]}]
-      else
-        %{}
-      end
-
+  def list_stats(access_token, property, date_range, limit, filters) do
```
macobo, line 42: Since filters here isn't the normal plausible filters, should we name this `search_console_filters`?

macobo (approved):

PR #4071, opened 2024-05-06
Refactor: Use common current_visitors code

macobo:
### Changes

This PR:
- Unifies ways we calculate current_visitors throughout the app
- Removes a bunch of resulting dead code

Motivation: I was refactoring for API v2 Query object usage, this file stuck out as in need of modernizing. Extracting change from a larger PR for easier reviewability.

PR #4082, opened 2024-05-08
Refactor internal Query schema and introduce WhereBuilder

macobo:
### Changes

This PR changes:
1. how we store filters internally in Query object
2. how we build the WHERE clause of a given query

Motivation is we want to build a saved segments feature and a new version of the API with a more [flexible filters system](https://github.com/plausible/knowledge_base/pull/49). This work is a prerequisite for both of these.

WhereBuilder specifically makes how we "compile" a query from filters more explicit and makes supporting and/or/not blocks easier in the future. 

Note this is a pure refactor - no user-facing changes are occurring here.

Related PRs: https://github.com/plausible/analytics/pull/4066, https://github.com/plausible/analytics/pull/4071, https://github.com/plausible/analytics/pull/4068

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/google/api.ex
```diff
@@ -73,16 +73,17 @@ defmodule Plausible.Google.API do
...
          {:ok, stats} <-
            HTTP.list_stats(
              access_token,
              site.google_auth.property,
-             date_range,
+             query.date_range,
              limit,
-             filters["page"]
+             # :BUG: This type of filter does not exist.
+             Plausible.Stats.Query.get_filter(query, "page")
```
macobo, line 86: Note: This is already broken, @ukutaht is working on a proper replacement under https://github.com/plausible/analytics/pull/4077

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -718,9 +734,11 @@ defmodule PlausibleWeb.Api.StatsController do
     site = conn.assigns[:site]
     params = Map.put(params, "property", "visit:referrer")
 
+    referrer_filter = DashboardFilterParser.filter_value("visit:source", referrer)
```
ukutaht, line 737: This is [guaranteed](https://github.com/plausible/analytics/blob/c41e5980fb3c8b13403290d58e5ef5807b57e6bd/assets/js/dashboard/stats/sources/index.js#L11) by the frontend to be a standard `is` filter so I think the DashboardFilterParser can be bypassed here. We can use `["is", "visit:source", referrer]` instead.

ukutaht (approved):
Good stuff 👍

RobertJoonas (approved):
LGTM 👍

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -684,10 +699,11 @@ defmodule PlausibleWeb.Api.StatsController do
 
     query =
       Query.from(site, params)
-      |> Query.put_filter("visit:source", "Google")
+      |> Query.put_filter([:is, "visit:source", "Google"])
 
+    # :TODO: Goal exclusion doesn't work here!
```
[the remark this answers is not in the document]
macobo: This will be fixed w/ https://github.com/plausible/analytics/pull/4077

lib/plausible_web/controllers/api/stats_controller.ex
macobo: Yes, but the "value" is still serialized. The previous put_filter method had special-case parsing logic for just this route - I decided to move it here instead of having it in the general Query class as it's not needed in any other case currently.

PR #4093, opened 2024-05-09
Autoconfigure event goals

RobertJoonas:
### Changes

Instead of providing event name suggestions when adding a goal, we've decided for the time being to add an option to "autoconfigure" all event names that we've received as custom event goals. The UX is very similar to how custom properties are added, based on the data from the last 6 months.

image

### Tests

### Changelog

### Documentation
- [x] https://github.com/plausible/docs/pull/496 is waiting to be merged

### Dark mode


zoldar (approved):

Wuemeli:
This would be nice

PR #4094, opened 2024-05-09
Avoid crashing on empty GA4 HTTP report response

zoldar:
### Changes

It turns out that google Data API (GA4 import) payload for dataset without any results is missing `rows` and `rowCount` fields. This change makes import procedure handle such cases gracefully.

### Tests


RobertJoonas (approved):
LGTM 👍

PR #4105, opened 2024-05-15
Fix StatsAPI access for new accounts on business tier

macobo:
Currently, business tier users created after business tier launch who never used the trial can't access Stats API due to faulty grandfathering logic. This change should fix that.


RobertJoonas (approved):
Looks good to me! 👍

PR #4106, opened 2024-05-15
Snippet integration verification

aerosol:
### Changes

This PR introduces a new feature - snippet integration verification.
The feature is exposed in two places: 
  - during standard onboarding (site creation)
  - in the General Settings tab, for manual execution

https://github.com/plausible/analytics/assets/173738/03e71769-9e1f-42fb-8878-d83aa52b75d6

#### Note for reviewers

This PR is rather sizeable - reviewing commit by commit should make it bearable. 
To play with it locally end to end, one has to `make browserless` - this will instantiate a self-hosted browserless.io docker image.
Please look at the tests first when in doubt.

#### How it works

Several modules implementing `Plausible.Verification.Check` run in a pipeline, modifying `Plausible.Verification.State`.
The state is later on interpreted into `Plausible.Verification.Diagnostics.Rating`, that carries the final result with, in case of failure, associated user-facing error and troubleshooting recommendation.

At a higher level, this is coordinated by `Plausible.Live.Verification` LiveView, that can be either embedded standalone (onboarding) or as in a modal (settings). There are slight differences for each mode. For instance, onboarding verification still keeps waiting for the first organic page-view (but lets the user skip to the dashboard however).

In general the following checks are made:
  - retrieving, parsing and inspecting the site's HTTP response server-side
  - running a headless browser script, verifying if `plausible()` function is defined at the target site, sending a test verification event to the server, and checking ingestion response status

A bit more detailed descriptions can be found in each check's `moduledoc`. 
The headless browser integration is done via [browserless.io function REST API](https://docs.browserless.io/HTTP-APIs/function).

The most critical piece is `Diagnostics.Rating` - we try to make it as accurate as possible, in order to help people troubleshoot their integrations without having to go through customer support. Tests should capture most of the scenarios we initially want to support.

#### Feature flag

We'll deploy it behind `:verification` flag initially. Alternatively, verification can be enabled permanently by setting `VERIFICATION_ENABLED=true` environment variable. That variable is on by default for `dev` environment, off otherwise.

#### Documentation updates

Documentation/troubleshooting pages will require an update - pending.

#### Self-hosting

cc @ruslandoga - by default the feature is off and should not interfere with self-hosted deployments in any way. 
We can later on add documentation on how to deploy it along with self-hosted browserless.io, if need be.

#### Gotchas

:warning: This PR modifies the plausible tracker. For that reason, in order to run it end to end, we must make sure that the snippet refers to the updated tracker. Otherwise, the error messages from diagnostic rating can get rather confusing.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (commented):
Looks good in general. I would like to play around with the UI as well but I don't think I can get to it today and I'm off for the rest of the week. Don't let it block this from being released though.

lib/plausible/verification/diagnostics.ex
```diff
@@ -0,0 +1,333 @@
...
+        %__MODULE__{
+          plausible_installed?: true,
+          service_error: nil,
+          body_fetched?: false
+        },
+        url
+      ) do
+    %Rating{
+      ok?: false,
+      errors: ["We couldn't reach #{url}. Is your site up?"],
```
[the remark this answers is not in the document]
aerosol: network issues? e.g. our DNS doesn't resolve the domain, but browserless does

I had this while using wi-fi on train, doesn't hurt to keep I guess

lib/plausible/verification/checks.ex
```diff
@@ -0,0 +1,75 @@
...
+
+    if async? do
+      Task.start_link(fn -> do_run(url, data_domain, checks, report_to, slowdown) end)
+    else
+      do_run(url, data_domain, checks, report_to, slowdown)
+    end
+  end
+
+  def interpret_diagnostics(%State{} = state) do
+    Plausible.Verification.Diagnostics.rate(state.diagnostics, state.url)
```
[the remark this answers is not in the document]
aerosol: sure thing, will change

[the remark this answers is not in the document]
aerosol: 4cbcbd26d

lib/plausible/verification/checks/csp.ex
```diff
@@ -0,0 +1,34 @@
...
+
+  @impl true
+  def perform(%State{assigns: %{headers: headers}} = state) do
+    case headers["content-security-policy"] do
+      [policy] ->
+        directives = String.split(policy, ";")
+
+        allowed? =
+          Enum.any?(directives, fn directive ->
+            String.contains?(directive, PlausibleWeb.Endpoint.host())
```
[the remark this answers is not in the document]
aerosol: Good call. I'll instrument it with no proxy match. a2181a449

lib/plausible_web/live/verification.ex
```diff
@@ -0,0 +1,162 @@
...
+        :not_mounted_at_router,
+        %{"domain" => domain} = session,
+        socket
+      ) do
+    socket =
+      assign(socket,
+        domain: domain,
+        modal?: !!session["modal?"],
+        component: @component,
+        report_to: session["report_to"] || self(),
```
zoldar, line 26: Is that something that can be even passed via session? (outside tests, maybe)

lib/plausible_web/live/verification.ex
```diff
@@ -0,0 +1,162 @@
...
+          if(socket.assigns.modal?,
+            do: "verification-within-modal",
+            else: "verification-standalone"
+          )
+      )
+    )
+  end
+
+  defp launch_delayed(socket) do
+    Process.send_after(self(), {:start, socket.assigns.report_to}, socket.assigns.delay)
```
zoldar, line 154: As `report_to` is set in socket assigns already - isn't sending it in the message redundant?

lib/plausible_web/live/verification.ex
```diff
@@ -0,0 +1,162 @@
...
+  def mount(
+        :not_mounted_at_router,
+        %{"domain" => domain} = session,
+        socket
+      ) do
+    socket =
+      assign(socket,
+        domain: domain,
+        modal?: !!session["modal?"],
+        component: @component,
```
zoldar, line 25: Under what circumstances might that be swapped?

lib/plausible/verification.ex
```diff
@@ -4,6 +4,16 @@ defmodule Plausible.Verification do
   """
   use Plausible
 
+  @feature_flag :verification
+
+  def enabled?(user) do
+    enabled_via_config? =
+      :plausible |> Application.get_env(__MODULE__) |> Keyword.fetch!(:enabled?)
```
zoldar, line 11:
```suggestion
      :plausible |> Application.fetch_env!(__MODULE__) |> Keyword.fetch!(:enabled?)
```

test/test_helper.exs
```diff
@@ -1,3 +1,5 @@
+FunWithFlags.enable(:verification)
```
zoldar, line 1: Let's maybe enable the flag in `seeds.exs` too?

lib/plausible/verification/checks/snippet_cache_bust.ex
```diff
@@ -27,6 +27,7 @@ defmodule Plausible.Verification.Checks.SnippetCacheBust do
 
     state2 =
       %{state | url: busted_url}
+      |> Plausible.Verification.Checks.ScanBody.perform()
```
zoldar: Shouldn't that be executed after `FetchBody`? I'm probably missing something obvious.

lib/plausible/verification/checks/installation.ex
```diff
@@ -27,7 +27,7 @@ defmodule Plausible.Verification.Checks.Installation do
         Jason.encode!(%{
           code: @code,
           context: %{
-            url: url,
+            url: Plausible.Verification.URL.bust_url(url),
```
zoldar, line 30: Doesn't pre-emptive busting defeat the purpose of `SnippetCacheBust` cache?

test/plausible/site/verification/checks_test.exs
```diff
@@ -301,8 +301,58 @@ defmodule Plausible.Verification.ChecksTest do
...
+    <head>
+    <meta name="foo" content="/wp-content/plugins/bar"/>
+    <meta name='plausible-analytics-version' content='2.0.9' />
+    <script defer data-domain="example.com" src="http://localhost:8000/js/script.js"></script>
+    </head>
+    <body>Hello</body>
+    </html>
+    """
+
+    test "detecting snippet after busting WordPress cache -  official plugin" do
```
zoldar, line 322:
```suggestion
    test "detecting snippet after busting WordPress cache - official plugin" do
```

zoldar (approved):
Part of the review comments I've made are actually addressed in further commits from the ones I was reviewing at the time - please ignore those. That aside, there are only really some tiny or almost cosmetic nits.

All that aside, awesome work ✨

zoldar:
Oh one more thing that just came to my mind - running browserless in detached mode per default, explicitly named with an easy way to stop it (following example of minio setup).

```diff
diff --git a/Makefile b/Makefile
index a85ee3b8f..527711a8d 100644
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,10 @@ postgres-stop: ## Stop and remove the postgres container
        docker stop plausible_db && docker rm plausible_db
 
 browserless:
-       docker run -e "TOKEN=dummy_token" -p 3000:3000 --network host ghcr.io/browserless/chromium
+       docker run -d -e "TOKEN=dummy_token" -p 3000:3000 --name plausible_browserless --network host ghcr.io/browserless/chromium
+
+browserless-stop:
+       docker stop plausible_browserless
```

WDYT?

test/test_helper.exs
aerosol: No need, since it's enabled via .dev.env

lib/plausible/verification/checks/installation.ex
aerosol: No, busting cache is only relevant if we found no snippet at first fetch. Busting cache at all times via the headless browser is addressing a case when wp proxy was re-enabled without clearing cache.

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -56,9 +56,10 @@ defmodule PlausibleWeb.StatsController do
     can_see_stats? = not Sites.locked?(site) or conn.assigns[:current_user_role] == :super_admin
     demo = site.domain == PlausibleWeb.Endpoint.host()
     dogfood_page_path = if !demo, do: "/:dashboard"
+    skip_to_dashboard? = conn.params["skip_to_dashboard"] == "true"
 
     cond do
-      stats_start_date && can_see_stats? ->
+      (stats_start_date && can_see_stats?) || (can_see_stats? && skip_to_dashboard?) ->
```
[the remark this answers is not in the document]
aerosol: Doesn't feel like `stats_start_date != nil` is a reliable way of saying "has stats". I'll let it be for now, not really related here.

lib/plausible/verification/check.ex
```diff
@@ -0,0 +1,46 @@
...
+
+      alias Plausible.Verification.Checks
+      alias Plausible.Verification.State
+      alias Plausible.Verification.Diagnostics
+
+      require Logger
+
+      @behaviour Plausible.Verification.Check
+
+      def perform_wrapped(state) do
```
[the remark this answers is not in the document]
aerosol: 25c7bd4b3

aerosol:
> Oh one more thing that just came to my mind - running browserless in detached mode per default, explicitly named with an easy way to stop it (following example of minio setup).

Will do that in a follow-up maybe. For now it's quite convienent to be able to see the logs attached and being able to stop it with C-c.

PR #4109, opened 2024-05-16
Fix unselecting filter from filter modal

ukutaht:
Quick fix for a frontend bug. Explanation: https://cap.link/4sh44v2japssn2w


metmarkosaric:
@ukutaht just an fyi that the video link goes to https://cap.so front page

zoldar (approved):
LGTM, reproduced the problem locally and confirmed the fix addresses it.

For reference, here's the recording of how the issue manifests:

[filter_issue.webm](https://github.com/plausible/analytics/assets/588351/1ca2a302-72cf-40e6-82b4-aead15fc3ea8)

PR #4110, opened 2024-05-17
Improve test coverage and fix stats API and dashboard CSV export

zoldar:
### Changes

This PR improves graceful fallback for unsupported metrics and properties combinations when querying imported stats and also fixes breaking down by `event:hostname` property which turned out to be defunct.

The assisting documentation update is available at https://github.com/plausible/docs/pull/499

### Tests

### DIFF
--- a/lib/plausible/stats/imported.ex
+++ b/lib/plausible/stats/imported.ex
@@ -29,7 +29,6 @@ defmodule Plausible.Stats.Imported do
     "visit:os_version" => "imported_operating_systems",
     "event:page" => "imported_pages",
     "event:name" => "imported_custom_events",
-    "event:hostname" => "imported_pages",
     "event:props:url" => "imported_custom_events",
     "event:props:path" => "imported_custom_events"
   }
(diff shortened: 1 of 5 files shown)


lib/plausible/stats/imported.ex
```diff
@@ -320,6 +321,18 @@ defmodule Plausible.Stats.Imported do
...
+         %Ecto.Query{from: %Ecto.Query.FromExpr{source: {"imported_pages", _}}} = q,
+         [:bounce_rate | rest]
+       ) do
+    q
+    |> select_merge([i], %{
+      bounces: 0,
+      __internal_visits: 0
+    })
+    |> select_imported_metrics(rest)
+  end
```
RobertJoonas, line 334: Curious to know where it breaks currently? We should always be getting bounce rate from `imported_entry_pages`.

This is actually related to filtering that I've been working on too - when we're filtering by page, we will not show bounce rate at all in the Top Stats. (That is only when imported data is included).

We discussed returning warnings under a new key in the API, so we might want to add an explicit warning like this:

```
{
  warnings: [
    "Even though imported data is included in the response, `bounce_rate`, and `visit_duration` are only based on native stats since they cannot be queried for imported pages."
  ],
  results: {
    bounce_rate: 23,
    visit_duration: 55,
    pageviews: 100,
    visitors: 50
  }
}
```

zoldar: Currently, the mapping for imported data only happens based on property - so when `event:page` is passed, the table in use will always be `imported_pages`. Do you think there should be special treatment to specific cases like this, so that particular property/metric combinations can be obtained from different tables?

RobertJoonas: About the warning suggested - this would be irrelevant since these metrics are actually queryable (simply coming from `imported_entry_pages`)

zoldar: Dropped support for breaking down by `event:hostname` entirely for now, with 400 validation error: https://github.com/plausible/analytics/pull/4110/commits/54f873cd2fc23dbc8810e5744acb1b58975551f4

RobertJoonas (approved):

PR #4096, opened 2024-05-10
Export and import custom events via CSV

zoldar:
### Changes

Please describe the changes made in the pull request here.

Below you'll find a checklist. For each item on the list, check one option and delete the other.

### Tests
- [ ] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [ ] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [ ] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [ ] This PR does not change the UI

PR #4114, opened 2024-05-20
Improve CSV importer tests and visit duration query for imported data

zoldar:
### Changes

This PR applies post-merge feedback from https://github.com/plausible/analytics/pull/4096#pullrequestreview-2063866535 (h/t @ruslandoga) and changes visit duration computation in imported data query to be in line with native one, always rounding to full integers instead of first decimal place.

### Tests


RobertJoonas (approved):
LGTM 👍

PR #4115, opened 2024-05-20
Introduce :skip_imported_reason field into Query struct

RobertJoonas:
### Changes

Just some groundwork that is relevant for me and @zoldar to work with. No behavioural changes, just introducing a new field.


zoldar (approved):
✨

PR #4116, opened 2024-05-20
Return warning when imported stats are skipped in Stats API due to unsupported query

zoldar:
### Changes

This PR adds a hint in a form of warning to Stats API responses when imported data are requested but can't be processed because query parameters are not supported for it.

Goes together with changes in docs at https://github.com/plausible/docs/pull/499/files

### Tests

### DIFF
--- a/lib/plausible_web/controllers/api/external_stats_controller.ex
+++ b/lib/plausible_web/controllers/api/external_stats_controller.ex
@@ -41,7 +41,9 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
           Plausible.Stats.aggregate(site, query, metrics)
         end
 
-      json(conn, %{results: results})
+      payload = maybe_add_warning(%{results: results}, query)
+
+      json(conn, payload)
     else
       err_tuple -> send_json_error_response(conn, err_tuple)
     end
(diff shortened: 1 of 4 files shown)


lib/plausible_web/controllers/api/external_stats_controller.ex
```diff
@@ -375,6 +379,20 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
...
+  end
+
+  defp maybe_add_warning(payload, %{skip_imported_reason: :unsupported_query}) do
+    Map.put(
+      payload,
+      :warning,
+      "Imported stats are not included in the results because query parameters are not supported. " <>
+        "For more information, see: https://plausible.io/docs/stats-api#filtering-imported-stats"
+    )
+  end
```
RobertJoonas, line 394: How about matching with `%{skip_imported_reason: :unsupported_query}` first, and in the second function definition clause we can just ignore the query argument?

RobertJoonas (approved):
one suggestion in line, otherwise looks good to me! 👍

lib/plausible_web/controllers/api/external_stats_controller.ex
zoldar: @RobertJoonas my thinking here was to explicitly match on all supported outcomes so that in the future it will surface if we ever change or expand it.

RobertJoonas: Good thinking. Agreed 👍

PR #4118, opened 2024-05-21
Implement filtering for imported data

RobertJoonas:
### Changes

This PR adds support for filtering by imported data. What we can or cannot filter by is limited by the imported schemas. Usually that means that the relation between different properties does not exist - e.g. we cannot filter by page and break down by source. That information doesn't exist. However, in some tables, there are several properties - e.g. `imported_locations` contains `country`, `region` and `city` propertis. That means that filtering by `country` and breaking down by `city` is possible.

Whenever filters are applied and imported data is included in the currently viewed period, we will display warning bubbles if a certain report is not able to include imported data:

image

Note that this does not mean that native data is excluded - the dashboard in the screenshot simply doesn't have any.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/query.ex
```diff
@@ -201,19 +203,39 @@ defmodule Plausible.Stats.Query do
...
+  end
+
+  def exclude_imported(query) do
+    struct!(query,
+      include_imported: false,
+      skip_imported_reason: :manual_exclusion
+    )
+  end
+
+  defp refresh(query) do
```
macobo, line 237: What is going on here? The consumers of `query` now seem to think about refreshing a lot, which seems like a leaky abstraction.

lib/plausible/stats/imported/base.ex
```diff
@@ -0,0 +1,192 @@
...
+    |> apply_filter(query)
+  end
+
+  def decide_table(query) do
+    query
+    |> transform_filters()
+    |> do_decide_table()
+  end
+
+  defp transform_filters(query) do
```
macobo, line 86: Q: Can you explain the business logic here?

I'm planning on moving business logic soonish so would need to understand this better.

lib/plausible/stats/imported/base.ex
```diff
@@ -0,0 +1,192 @@
...
+            |> Enum.map(fn
+              {:event, names} -> [op, "event:name", names]
+              {:page, pages} -> [op, "event:page", pages]
+            end)
+
+          filter ->
+            [filter]
+        end
+      end)
+      |> Enum.concat()
```
macobo, line 113: Code: Use `flat_map` above.

lib/plausible/stats/imported/base.ex
```diff
@@ -0,0 +1,192 @@
...
+    query
+    |> transform_filters()
+    |> do_decide_table()
+  end
+
+  defp transform_filters(query) do
+    new_filters =
+      query.filters
+      |> Enum.reject(fn
+        [:is, "event:name", "pageview"] -> true
```
macobo, line 90: Is this correct? `[:member, "event:name", ["pageview"]]` is functionally equivelent to this code

macobo:
Request: If we can merge this faster by splitting this into separate FE and BE PRs I'd appreciate that. Blocked on a large chunk of query/APIv2 work on these changes.

lib/plausible/stats/imported/base.ex
RobertJoonas: The `transform_filters` function is used for:

1) dropping the redundant "is name pageview" filter which doesn't make sense  it imported data
2) mapping an `event:goal` filter into either `event:name` or `event:page`

In the imported querying logic we need to map goals to name/page in two places:

1) deciding whether imported data can be included or not (i.e. schema_supports_query)
2) actually applying the filter

The general business logic with imported data is that we can query different properties in relation to each other (breakdown prop / filters) as long as they belong to the same `imported_*` table.

For example, when filtering by `event:page==/blog` and `event:goal==Visit /blog`, the query is supported and the data should be fetched from the `imported_pages` table.

Happy to clarify it more in depth if you have any more specific questions :)

lib/plausible/stats/imported/base.ex
RobertJoonas: True, we'd leave the `[:member, "event:name", ["pageview"]` in the query without rejecting it. But if a filter like that is met, it means it came from the client, not from our own internal logic of querying pageview goals, in which case it makes sense to keep it, I think. I wouldn't add it as an explicit clause here.

RobertJoonas:
> Request: If we can merge this faster by splitting this into separate FE and BE PRs I'd appreciate that. Blocked on a large chunk of query/APIv2 work on these changes.

I think we don't need to split it and can go ahead and merge it on Monday.

lib/plausible/stats/imported/base.ex
RobertJoonas: Good catch! https://github.com/plausible/analytics/pull/4118/commits/e71466d701607019902f3a5a0ee9ce117e7b9cfc

lib/plausible/stats/query.ex
RobertJoonas: The `refresh/1` function currently only calls `refresh_imported_opts/1` which is meant to re-evaluate `include_imported` and `skip_imported_reason`. After we modify the filters or breakdown property, whether we support imported data or not might change.

However, sometimes, for example when calling `remove_filters` to calculate conversion rate / percentage totals, even though it would make imported data queryable, we still don't want it. Hence the `:skip_refresh` option.

> The consumers of query now seem to think about refreshing a lot, which seems like a leaky abstraction.

Agreed. We can skip the `refresh` function and call `refresh_imported_opts`directly to make intent a bit clearer. The option can also be renamed to `:skip_refresh_imported_opts`. Done in https://github.com/plausible/analytics/pull/4118/commits/9befb9f8f97bf1e6f95a5155fb5ed323933c747a

zoldar (approved):

PR #4132, opened 2024-05-24
Verification improvements

aerosol:
Various integration verification improvements found out during live testing.

### DIFF
--- a/lib/plausible/verification/diagnostics.ex
+++ b/lib/plausible/verification/diagnostics.ex
@@ -14,7 +14,7 @@ defmodule Plausible.Verification.Diagnostics do
             body_fetched?: false,
             wordpress_likely?: false,
             gtm_likely?: false,
-            callback_status: -1,
+            callback_status: 0,
             proxy_likely?: false,
             data_domain_mismatch?: false,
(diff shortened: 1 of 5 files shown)


lib/plausible/verification/diagnostics.ex
```diff
@@ -150,13 +151,77 @@ defmodule Plausible.Verification.Diagnostics do
...
+
+  def interpret(
+        %__MODULE__{
+          plausible_installed?: true,
+          callback_status: 0,
           proxy_likely?: true
         },
         _url
-      )
-      when callback_status != 202 do
```
zoldar, line 159: So the case of callback failing due to other HTTP code will be handled by the final, catch-all clause, right? (like 500/503s, however unlikely).

zoldar (approved):

lib/plausible/verification/diagnostics.ex
aerosol: Yes, I'll add a test.

aerosol: d88366ca7

PR #4134, opened 2024-05-24
always enable ipv6 connections in CE

ruslandoga:
### Changes

This PR removes the need for custom IPv6 configuration (ECTO_IPV6, ECTO_CH_IPV6, FINCH_IPV6) in CE and makes Mint and Mua try IPv6 connection first with fallback to IPv4 and Postgrex try default (URL) and then interspersed IPv4 and  IPv6 addresses.

This PR is the first step to full IPv6 support. tzdata (hackney), locus (httpc without inet6fb4), ua_inspector (hackney) and some other libraries would still fail in IPv6-only setups.

### Tests

### Changelog

### Documentation

### Dark mode


ruslandoga:
closing in favor of #4245

PR #4136, opened 2024-05-24
More fixes to the verification process

aerosol:
- extend the list of known attributes
 - follow redirect chain host to verify data-domain difference
 - display the actual data-domain in error message, not URL

### DIFF
--- a/lib/plausible/verification/checks/snippet.ex
+++ b/lib/plausible/verification/checks/snippet.ex
@@ -40,7 +40,7 @@ defmodule Plausible.Verification.Checks.Snippet do
     "defer",
     "data-api",
     "data-exclude",
-    "data-include",
+    "data-include"
   ]
 
   defp unknown_attributes?(nodes) do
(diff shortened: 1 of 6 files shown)


lib/plausible/verification/checks/snippet.ex
```diff
@@ -33,20 +34,33 @@ defmodule Plausible.Verification.Checks.Snippet do
...
-  defp data_domain_mismatch?(nodes, data_domain) do
+  defp data_domain_mismatch?(nodes, data_domain, final_data_domain) do
     nodes
     |> Floki.attribute("data-domain")
-    |> Enum.any?(&(&1 != data_domain and data_domain not in String.split(&1, ",")))
+    |> Enum.any?(fn script_data_domain ->
+      multiple = String.split(script_data_domain, ",")
+
+      script_data_domain not in [data_domain, final_data_domain] and data_domain not in multiple and
+        final_data_domain not in multiple
```
zoldar, line 63:
```suggestion
      data_domain not in multiple and final_data_domain not in multiple
```

I think this is enough to cover the whole condition as `String.split(..., ",")` is going to return a single element list in absence of `,`?

zoldar (approved):

lib/plausible/verification/checks/snippet.ex
aerosol: nice

PR #4138, opened 2024-05-24
First pass: extract verification errors

aerosol:
We like to make them easier to maintain.
An update fixing documentation URLs will follow.


aerosol:
cc @metmarkosaric - list of errors will be kept here: https://github.com/plausible/analytics/blob/eeaaabc89daf5cdf58e8c1dcaf5a95c6c6d03287/lib/plausible/verification/errors.ex

zoldar (approved):
🧹 ✨

PR #4148, opened 2024-05-27
Fix setting with_imported to false explicitly in the URL

RobertJoonas:
### Changes

Currently, when clicking on the imported exclude/include icon on the dashboard (see `with-imported-switch.js`), it doesn't set the `with_imported` to false due to PlausibleSearchParams class dropping any keys with `false` boolean value. The default for the `with_imported` field is `true` so we need to explicitly set it to false sometimes. Otherwise we'd end up with a URL identical to the current location, and the button simply doesn't work.

What I've done here is set the value as a string instead. Not aware of any other cases of this causing problems but maybe it would be worth to double-check (cc @macobo ). Is it necessary to drop the keys with false values or is it only an optimization?


macobo:
Right, this broke during refactoring. The dropping of flags behavior is intentional as it simplifies components reading/setting values  significantly. I guess this flag behaves unconventionally as it doesn't _only_ read from the URL.

As I don't have a great fix in mind for this of top of my head, let's go with this solution!

PR #4159, opened 2024-05-29
Make React frontend capable of reading API results under a results key

RobertJoonas:
### Changes

With the new imported data filtering feature, we'll start returning warnings in the dashboard API results, which means we'll need to move the actual list of results under a separate `results` key.

This PR is a preparation in order to not break the frontend when suddenly receiving API results under a `results` key. For some time, we'll need to be able to handle both a straight-up list response, or a map, containing the `results` key.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/locations/index.js
+++ b/assets/js/dashboard/stats/locations/index.js
@@ -11,9 +11,7 @@ import { getFiltersByKeyPrefix } from '../../util/filters';
 
 function Countries({query, site, onClick}) {
   function fetchData() {
-    return api.get(apiPath(site, '/countries'), query, {limit: 9}).then((res) => {
-      return res.map(row => Object.assign({}, row, {percentage: undefined}))
-    })
+    return api.get(apiPath(site, '/countries'), query, { limit: 9 })
   }
 
   function renderIcon(country) {
--- a/assets/js/dashboard/stats/modals/props.js
+++ b/assets/js/dashboard/stats/modals/props.js
@@ -38,11 +38,13 @@ function PropsModal(props) {
 
   function fetchData() {
     api.get(url.apiPath(site, `/custom-prop-values/${propKey}`), query, { limit: 100, page })
-      .then((res) => {
+      .then((response) => {
+        const results = !!response.results ? response.results : response
+
         setLoading(false)
-        setList(list.concat(res))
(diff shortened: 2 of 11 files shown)


assets/js/dashboard/stats/locations/index.js
```diff
@@ -11,9 +11,7 @@ import { getFiltersByKeyPrefix } from '../../util/filters';
 
 function Countries({query, site, onClick}) {
   function fetchData() {
-    return api.get(apiPath(site, '/countries'), query, {limit: 9}).then((res) => {
-      return res.map(row => Object.assign({}, row, {percentage: undefined}))
-    })
+    return api.get(apiPath(site, '/countries'), query, { limit: 9 })
   }
```
RobertJoonas, line 15: I believe this manual mapping to percentage undefined was used in order to prevent displaying percentages in this view:

image

However, the `ListReport` component does not render it anyway unless it's specified in the `metrics` prop, so the outcome is essentially the same without this manual mapping.

assets/js/dashboard/stats/modals/props.js
```diff
@@ -38,11 +38,13 @@ function PropsModal(props) {
...
-      .then((res) => {
+      .then((response) => {
+        const results = response.results ? response.results : response
+
         setLoading(false)
-        setList(list.concat(res))
+        setList(list.concat(results))
         setPage(page + 1)
-        setMoreResultsAvailable(res.length >= 100)
+        setMoreResultsAvailable(results >= 100)
```
zoldar, line 47:
```suggestion
        setMoreResultsAvailable(results.length >= 100)
```

zoldar (approved):

ukutaht (approved):
Is there a plan to clean up the fallbacks once backend changes are done as well? I'm asking because IMO it doesn't make sense to keep accepting multiple different response formats when we know for sure what is being returned from internal APIs.

RobertJoonas:
> Is there a plan to clean up the fallbacks once backend changes are done as well? I'm asking because IMO it doesn't make sense to keep accepting multiple different response formats when we know for sure what is being returned from internal APIs.

Yes, @ukutaht. The PR that implements filtering will remove it so we'll start relying only on the results key.

PR #4165, opened 2024-05-30
fix registration in ce

ruslandoga:
### Changes

This PR fixes the registration issue reported in https://github.com/plausible/analytics/discussions/4164 (submitting the form doesn't redirect to the next step)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/register_form.ex
+++ b/lib/plausible_web/live/register_form.ex
@@ -317,7 +317,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
 
           {:noreply, push_event(socket, "send-metrics", metrics_params)}
         else
-          {:noreply, socket}
+          {:noreply, assign(socket, trigger_submit: true)}
         end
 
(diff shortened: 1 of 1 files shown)


lib/plausible_web/live/register_form.ex
```diff
@@ -317,7 +317,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
 
           {:noreply, push_event(socket, "send-metrics", metrics_params)}
         else
-          {:noreply, socket}
+          {:noreply, assign(socket, trigger_submit: true)}
```
ruslandoga, line 320: On CE `Metrics` hook isn't used so `"send-metrics-after"` event never arrives.

ruslandoga: The fix: trigger submit on successful insert instead of in `handle_event("send-metrics-after", ...)`

zoldar (approved):

PR #4183, opened 2024-06-04
use nimble_csv everywhere

ruslandoga:
### Changes

This PR replaces https://github.com/beatrichartz/csv with https://github.com/dashbitco/nimble_csv as a more performant option (fewer binaries are created this way)

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -154,6 +154,7 @@ defmodule PlausibleWeb.StatsController do
         Map.keys(csvs)
         |> Enum.zip(csv_values)
         |> Enum.reject(fn {_k, v} -> is_nil(v) end)
+        |> Enum.map(fn {k, v} -> {k, IO.iodata_to_binary(v)} end)
 
       {:ok, {_, zip_content}} = :zip.create(filename, csvs, [:memory])
```
ruslandoga, line 159: `:zip.create` doesn't support iodata, maybe it makes sense to use zstream here and send the export in chunks.

ukutaht:
Hey @ruslandoga I'd like to merge this if possible. Why is it in draft status?

ruslandoga:
The main reason is that I thought it would improve performance by using iolists but it turned out to be impossible since `:zip.create` only supports binaries. So I then marked it as Draft to make further changes to get rid of `:zip.create` and forgot. We can still merge it to remove the :csv dependency.

ruslandoga:
I'll reabse the PR and remove the Draft tag.

ukutaht (approved):

ruslandoga:
CSV export worked for https://pr-4183.review.plausible.io/dummy.site

PR #4185, opened 2024-06-04
Fix parsing = in jsonurl

macobo:
This works around a bug in the jsonurl library where the literal causes a syntaxerror. The urlencoded = however is handled without errors


aerosol (approved):

aerosol:
Ref https://github.com/jsonurl/jsonurl-js/issues/673

PR #4187, opened 2024-06-05
Fix with imported switch + warning bubble renders

RobertJoonas:
### Changes

The main goal of this PR was to fix some behaviour with the dashboard "with-imported" switch and move its logic into the backend, making its intent more clear and easily testable. I've also included some other changes that I saw necessary. Here's a summary:

* Change the tooltip message when the icon is hovered:
  * "Stats do not include imported data" -> "Click to include imported data" 
  * "Stats include imported data" -> "Click to exclude imported data"
* Slight refactor: instead of keeping a boolean state of whether the imported query is unsupported, keep the `skipImportedReason` in the state instead and let the `ImportedQueryUnsupportedWarning` component decide how to interpret that.
* When the query does not support imported data and imported data is excluded, disable clicking the imported-switch and render a tooltip saying "Imported data cannot be included"

Before (with imported switch still clickable when filters are unsupported, when clicked, also triggers an update):

https://github.com/plausible/analytics/assets/56999674/1e45e336-0583-4a21-8d12-6c89e57c022e

After (the icon is disabled when filters are unsupported):

https://github.com/plausible/analytics/assets/56999674/9d21f127-8f3b-440f-b99a-2092d0806c72

* Hide the `ImportedQueryUnsupportedWarning` bubble whenever the query or tab changes

Before (notice how the bubble keeps hanging, and then disappears suddenly):

https://github.com/plausible/analytics/assets/56999674/9e981974-05fc-41be-a4c8-64fad13ed8d1

After: (bubble not displayed by default, fades in when the request finishes)

https://github.com/plausible/analytics/assets/56999674/7a31f3e4-624f-4540-a276-6d788ff1e009

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/behaviours/index.js
+++ b/assets/js/dashboard/stats/behaviours/index.js
@@ -333,9 +333,9 @@ export default function Behaviours(props) {
 
   function renderImportedQueryUnsupportedWarning() {
     if (mode === CONVERSIONS) {
-      return <ImportedQueryUnsupportedWarning skipImportedReason={skipImportedReason}/>
+      return <ImportedQueryUnsupportedWarning query={query} skipImportedReason={skipImportedReason}/>
     } else if (mode === PROPS) {
-      return <ImportedQueryUnsupportedWarning skipImportedReason={skipImportedReason} message="Imported data is unavailable in this view"/>
+      return <ImportedQueryUnsupportedWarning query={query} skipImportedReason={skipImportedReason} message="Imported data is unavailable in this view"/>
     } else {
       return <ImportedQueryUnsupportedWarning alt_condition={props.importedDataInView} message="Imported data is unavailable in this view"/>
     }
--- a/assets/js/dashboard/stats/behaviours/index.js
+++ b/assets/js/dashboard/stats/behaviours/index.js
@@ -48,7 +48,7 @@ export default function Behaviours(props) {
 
   const [showingPropsForGoalFilter, setShowingPropsForGoalFilter] = useState(false)
 
-  const [importedQueryUnsupported, setImportedQueryUnsupported] = useState(false)
+  const [skipImportedReason, setSkipImportedReason] = useState(null)
 
   const onGoalFilterClick = useCallback((e) => {
     const goalName = e.target.innerHTML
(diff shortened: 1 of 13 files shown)


assets/js/dashboard/stats/behaviours/index.js
```diff
@@ -332,6 +335,16 @@ export default function Behaviours(props) {
     }
   }
 
+  function renderImportedQueryUnsupportedWarning() {
+    if (mode === CONVERSIONS) {
+      return <ImportedQueryUnsupportedWarning loading={loading} query={query} skipImportedReason={skipImportedReason}/>
+    } else if (mode === PROPS) {
+      return <ImportedQueryUnsupportedWarning loading={loading} query={query} skipImportedReason={skipImportedReason} message="Imported data is unavailable in this view"/>
+    } else {
+      return <ImportedQueryUnsupportedWarning alt_condition={props.importedDataInView} message="Imported data is unavailable in this view"/>
```
zoldar, line 344: Wouldn't `altCondition` for property name be more in line with naming conventions?

assets/js/dashboard/stats/locations/index.js
```diff
@@ -142,6 +143,10 @@ export default class Locations extends React.Component {
     if (this.state.mode === 'regions' && isRemovingFilter('country')) {
       this.setMode(this.countriesRestoreMode || 'countries')()
     }
+
+    if (this.props.query !== prevProps.query || this.state.mode !== prevState.mode) {
```
zoldar, line 147: Not related to this PR specifically - would it make sense to make components like this one use `useState` and `useEffect` for managing state for consistency?

zoldar (approved):
Nice work ✨

assets/js/dashboard/stats/behaviours/index.js
RobertJoonas: Oh, totally! https://github.com/plausible/analytics/pull/4187/commits/5c99c2b6fc8fbbb13267e14cc2a4947b815a906c

assets/js/dashboard/stats/locations/index.js
RobertJoonas: Yes, it would. I think we should refactor all React class components into function components in the codebase. This one looked a bit more tricky to do at first glance so I didn't get into it at the moment. But yeah, would be a nice refactoring.

PR #4198, opened 2024-06-07
Refactor: Move from class to function components in React

RobertJoonas:
### Changes

This PR is only a refactor, and doesn't change any behaviour. The goal is to modernise the React app and move from class components to function components. Also an attempt to make the code more readable and remove unnecessary/dead functionality.

Note that this PR doesn't get rid of all class components - several still exist around the codebase.

Reviewing commit-by-commit recommended since I believe there's hardly any overlap between commits - each is doing its own thing.

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/filters.js
```diff
@@ -119,94 +119,66 @@ function DropdownContent({ history, site, query, wrapped }) {
...
-  componentDidUpdate(prevProps, prevState) {
-    const { query } = this.props;
-    const { viewport, wrapped } = this.state;
-
-    if (JSON.stringify(query) !== JSON.stringify(prevProps.query) || viewport !== prevState.viewport) {
-      // eslint-disable-next-line react/no-did-update-set-state
-      this.setState({ wrapped: 1 });
+function Filters(props) {
+  const { history, query, site } = props
+  const [wrapped, setWrapped] = useState(1) // 0=unwrapped, 1=waiting to check, 2=wrapped
```
macobo, line 124: Suggestion: create an object above and use `useState(WRAPSTATE.waiting)`

assets/js/dashboard/filters.js
```diff
@@ -119,94 +121,66 @@ function DropdownContent({ history, site, query, wrapped }) {
...
 
-    this.setState({ wrapped: 0 });
-
-    // Don't rewrap if we're already properly wrapped, there are no DOM children, or there is only filter
-    if (wrapped !== 1 || !items || this.props.query.filters.length === 1) {
-      return;
-    };
-
-    let prevItem = null;
+    setWrapped(WRAPSTATE.unwrapped)
```
RobertJoonas, line 171: I'm not a fan of this here - we're setting the `wrapped` state to `unwrapped`, and then a few lines below we're doing another state update to set it to `unwrapped`. It's not obvious to me how it works, and I've tried to refactor this but without any luck :/

ukutaht: Yeah I don't really get this either but I guess we can just leave it

ukutaht (approved):
Good stuff, thanks for this!

PR #4211, opened 2024-06-11
Add "Show more" link to browse more than 12 invoices

aerosol:
### Changes

We used to limit the amount of show paddle invoices to 12 - not anymore.
The limit is removed completely. In case there are more than 12 invoices, we'll display a "Show more" link unfolding the remaining ones.

https://github.com/plausible/analytics/assets/173738/f1c7765e-4536-451b-9046-f731efbe3111

Not sure if there's a good way to unit test this, open to suggestions.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (approved):
> Not sure if there's a good way to unit test this, open to suggestions.

Only suggestion I have is to turn it into a LiveView but that feels like a big yak shave for such a small feature.

PR #4212, opened 2024-06-11
Reset grace period if usage drops within the last cycle

aerosol:
So that `LockSites` job unlocks the site, encountering no grace period and no upgrade needed.

A bigger refactoring is needed IMHO so that CheckUsage and LockSites aren't separate. There's implicit dependency between the two - one (re)sets the grace period, and the other (un)locks dashboards depending on grace period state.
Thus, there is no trace of the reason for which a site is being unlocked.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (approved):
Looks good to me 👍

> A bigger refactoring is needed IMHO so that CheckUsage and LockSites aren't separate

Yeah, not sure why these are separate jobs at the moment.

PR #4216, opened 2024-06-11
APIv2 - initial PR

macobo:
### Changes

This PR:
- Introduces /api/v2/query - single endpoint to query stats
- Introduces new query parsing and result serialization methods
- Ports a bunch of breakdown tests over
- Updates aggregates.ex to calculate results via new code
- Updating imports querying logic to handle single-breakdown cases 

Note this is a snapshot of progress. There's a bunch of things this doesn't handle
- breakdowns by event:page and event:goal
- revenue metrics
- timeseries
- time range parsing and testing
- event:hostname special business logic
- conversion_rate special business logic
- missing validations
- time_on_page metric
... and probably more

These will be solved in follow-up PRs

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/table_decider.ex
```diff
@@ -9,10 +9,18 @@ defmodule Plausible.Stats.TableDecider do
...
 
   def events_join_sessions?(query) do
-    Enum.any?(query.filters, &(filters_partitioner(query, &1) == :session))
+    query
+    |> filter_keys()
+    |> Enum.any?(&(filters_partitioner(query, &1) == :session))
   end
 
-  def partition_metrics(metrics, query, breakdown_property \\ nil) do
+  def event_filters?(query) do
```
ukutaht, line 17: I think this is a duplicate of `Query.has_event_filters?`

I prefer having it here, and replacing all uses of `Query.has_event_filters?` with `TableDecider.event_filters?`

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -192,6 +241,89 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+         not is_nil(Query.get_filter(query, "event:goal")) do
+      :ok
+    else
+      {:error, "Metric `#{metric}` can only be queried with event:goal filters or dimensions"}
+    end
+  end
+
+  defp validate_metric(_, _), do: :ok
+
+  defp validate_no_metrics_filters_conflict(query) do
```
ukutaht, line 314: This is a point of contention at the moment. We do allow session metrics with custom prop filters in the UI for example and some customers are unhappy that it's not allowed in the API.

Not a concern for this PR but just flagging that this validation needs to be adjusted to make it pass for the UI. This will also make the API more conistent with the UI which is a win.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -1,22 +1,30 @@
...
            filters: filters,
            date_range: date_range,
            dimensions: dimensions,
-           order_by: order_by
+           order_by: order_by,
+           timezone: site.timezone,
+           imported_data_requested: Map.get(include, :imports, false)
          },
-         :ok <- validate_order_by(query) do
+         :ok <- validate_order_by(query),
```
ukutaht, line 24: Maybe it makes sense to separate `QueryParser` and `QueryValidator`...

lib/plausible/stats/imported/imported.ex
```diff
@@ -274,18 +274,18 @@ defmodule Plausible.Stats.Imported do
       site
```
ukutaht, line 274: I don't really understand the changes to this file.. can you explain what the difference between v2 and v1 is here?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -86,16 +94,26 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+  defp parse_date_range(site, "day") do
+    now = Timex.now(site.timezone) |> Timex.to_date()
+    {:ok, Date.range(now, now)}
+  end
+
+  defp parse_date_range(_site, "7d"), do: {:ok, "7d"}
+  defp parse_date_range(_site, "30d"), do: {:ok, "30d"}
+  defp parse_date_range(_site, "month"), do: {:ok, "month"}
+  defp parse_date_range(_site, "6mo"), do: {:ok, "6mo"}
+  defp parse_date_range(_site, "12mo"), do: {:ok, "6mo"}
```
ukutaht, line 106: Typo

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -86,16 +94,26 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
-  defp parse_date_range("6mo"), do: {:ok, "6mo"}
-  defp parse_date_range("12mo"), do: {:ok, "6mo"}
-  defp parse_date_range("year"), do: {:ok, "year"}
-  defp parse_date_range("all"), do: {:ok, "all"}
+  defp parse_date_range(site, "day") do
+    now = Timex.now(site.timezone) |> Timex.to_date()
+    {:ok, Date.range(now, now)}
+  end
+
+  defp parse_date_range(_site, "7d"), do: {:ok, "7d"}
```
ukutaht, line 102: It bothers me that sometimes query.date_range is a DateRange and sometimes a String. Can we always make it one or the other?

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -0,0 +1,3106 @@
+defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
```
ukutaht, line 1: Are these ported over from the old ExternalStatsController tests?

I'm thinking a couple steps ahead but eventually I suppose we can completely remove the old `Aggregate`,  `TimeSeries` and `Breakdown` modules and porting those tests over to test the new `Stats.query` function instead.

lib/plausible/stats/ecto/expression.ex
```diff
@@ -0,0 +1,95 @@
+defmodule Plausible.Stats.Ecto.Expression do
```
ukutaht, line 1: I don't know how I feel about having a namespace like this that can alias a library. For example in `Plausible.Stats` this module is aliased so it can be confusing whether `Ecto` refers to the library or our module.

The best suggestion I have at the moment would be to rename this to `SQL` instead - `Plausible.Stats.SQL.Expression`. IDK, it's more generic but at least doesn't clash with anything else.

ukutaht (commented):
Overall I really like how logical & compact the new stuff is. Had some comments with suggestions. I also made some notes for myself to poke around with, I'll do that now.

lib/plausible/stats/ecto/expression.ex
```diff
@@ -0,0 +1,95 @@
...
+  def dimension("event:name", _query, _mode), do: dynamic([t], t.name)
+  def dimension("event:page", _query, _mode), do: dynamic([t], t.pathname)
+  def dimension("event:hostname", _query, _mode), do: dynamic([t], t.hostname)
+
+  def dimension("event:props:" <> property_name, _query, _mode) do
+    dynamic(
+      [t],
+      fragment(
+        "if(not empty(?), ?, '(none)')",
+        get_by_key(t, :meta, ^property_name),
```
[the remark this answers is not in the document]
ukutaht: not in scope for this PR, just a generic thought/question

lib/plausible/stats/filters/query_parser.ex
macobo: Yup! It will eventually all be date_range and be properly tested. Was planning to do this once I get to refactoring TimeSeries!

lib/plausible/stats/ecto/expression.ex
macobo: I was going back-and-forth between Ecto, Query and SQL. Query being already overloaded, I think let's go with SQL.

lib/plausible/stats/filters/filters.ex
```diff
@@ -87,6 +87,6 @@ defmodule Plausible.Stats.Filters do
     property
     |> String.split(":")
     |> List.last()
-    |> String.to_existing_atom()
+    |> String.to_atom()
```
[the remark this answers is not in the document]
macobo: Hmm yeah. Ecto needs to_atom for selected_as and maybe select_merge field names. The problematic bit here are custom props.

This is slightly tricky to fix (since it also involves imports querying changes) so I created this ticket: https://3.basecamp.com/5308029/buckets/37310803/card_tables/cards/7503023629

lib/plausible/stats/filters/query_parser.ex
macobo: I dug more into this in a follow-up PR where the condition becomes `TableDecider.event_dimensions?`.

Turns out that event filters with session metrics work just fine (with the semantics being _has any event with matching_) and is already used that way in aggregates.

lib/plausible/stats/table_decider.ex
macobo: See https://github.com/plausible/analytics/pull/4216#discussion_r1637680750, this will change in a follow-up soon!

lib/plausible/stats/filters/query_parser.ex
macobo: I prefer having them together with the logic being that everything coming out of this module is already valid. That some constraints require a broader context is a bit unfortunate but eh.

lib/plausible/stats/imported/imported.ex
macobo: Imports logic tries to re-encode a bunch of special case logic from lib/plausible/stats/breakdown.ex (e.g. some field naming, os_version/browser_version doing a multiple breakdown). This is hackery that is required to make both breakdown.ex and query_builder.ex work with this code.

I'm hoping to get rid of the breakdown.ex code as part of this project, which will naturally lead to these hacks going away again (and us perhaps needing to change imports logic again to support multiple breakdowns).

lib/plausible/stats/ecto/query_builder.ex
```diff
@@ -0,0 +1,198 @@
...
+    else
+      q
+    end
+  end
+
+  defp build_group_by(q, query) do
+    Enum.reduce(query.dimensions, q, fn dimension, q ->
+      q
+      |> select_merge(^%{shortname(dimension) => Expression.dimension(dimension, query, :label)})
+      |> group_by(^Expression.dimension(dimension, query, :group_by))
```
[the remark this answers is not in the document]
macobo: Clickhouse-wise it can, I couldn't make it work with ecto though! The problem is the select field aliases we set in select_merge don't actually end up in the query so we can't really use `fragment`. If you can make this work, let me know.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
macobo: > Are these ported over from the old ExternalStatsController tests?

Pretty much. This module currently contains just about all the breakdown.ex tests, minus pagination-related ones. Ones which don't pass yet due to long-tail issues (e.g. conversion_rate) are commented out.

I was planning on adding aggregation and timeseries tests under this as well.

> I'm thinking a couple steps ahead but eventually I suppose we can completely remove the old Aggregate, TimeSeries and Breakdown modules and porting those tests over to test the new Stats.query function instead.

My plan is to have this file contain all tests for the new /query endpoint and perhaps clean up aggregate_test/timeseries_test/breakdown_test to be a bit more bare-bones once they all point to the new QueryBuilder. We still need tests for the old controller though due to the different response format and validations.

lib/plausible/stats/ecto/expression.ex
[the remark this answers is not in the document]
macobo: We can. IMO not worth it though since it's a two-array structure and not a json object - there's a learning curve either way.

lib/plausible/stats/ecto/query_builder.ex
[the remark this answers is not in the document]
ukutaht: Ugh I'm not a fan of how this stuff works in Ecto. Aliasing fields and referring to them is so hard!

I'm sure it's possible with some custom code/macros but that would probably make the whole thing more complicated than it is now.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
ukutaht: Yeah makes sense. I wonder if the test file can be broken down by some logical chunks. It's already 3k lines long :D

lib/plausible/stats/sql/expression.ex
```diff
@@ -0,0 +1,95 @@
...
+  import Ecto.Query
+
+  use Plausible.Stats.Fragments
+
+  @no_ref "Direct / None"
+  @not_set "(not set)"
+
+  defmacrop field_or_blank_value(expr, mode, empty_value) do
+    quote do
+      case unquote(mode) do
```
ukutaht, line 13: When I remove this conditional and only use the `:label` variant, all the tests still pass. What's the purpose of this case statement?

macobo: Hmm. My original thinking was to emulate how current group_by is doing it and effectiveness. But I haven't tested out whether it's actually better this way and it would cause problems when `dimension=["visit:source"]` and ordering by source name asc. Will simplify!

ukutaht (approved):

PR #4217, opened 2024-06-11
Do not parse query params from dogfooded events

zoldar:
### Changes

This PR introduces a mechanism for ignoring URL parameters when parsing UTM tags and referrer source for dogfooded events coming from Plausible itself. Without it, any session initiating request with URL containing one of the accepted filter params, like `source` or `utm_source` will accidentally record this filter parameter as respective session property as well.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible.ex
+++ b/lib/plausible.ex
@@ -58,8 +58,4 @@ defmodule Plausible do
       "Plausible Analytics"
     end
   end
-
-  def dogfood_prop_key do
-    "__plausible_dogfood"
-  end
 end
--- a/lib/plausible.ex
+++ b/lib/plausible.ex
@@ -58,4 +58,8 @@ defmodule Plausible do
       "Plausible Analytics"
     end
   end
+
+  def dogfood_prop_key do
+    "__plausible_dogfood"
+  end
 end
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -258,19 +258,15 @@ defmodule Plausible.Ingestion.Event do
   end
 
   defp put_utm_tags(%__MODULE__{} = event) do
-    if !event.request.props[Plausible.dogfood_prop_key()] do
-      query_params = event.request.query_params
+    query_params = event.request.query_params
 
-      update_session_attrs(event, %{
-        utm_medium: query_params["utm_medium"],
-        utm_source: query_params["utm_source"],
-        utm_campaign: query_params["utm_campaign"],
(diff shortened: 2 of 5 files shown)


lib/plausible.ex
```diff
@@ -58,4 +58,8 @@ defmodule Plausible do
       "Plausible Analytics"
     end
   end
+
+  def dogfood_prop_key do
```
aerosol, line 62: There's `Plausible.Dogfood` module though.
And `_tracking.html.heex` which could be a better place to massage the dogfooding events? I don't like we're polluting ingestion with special dogfooding cases tbh

zoldar: Makes sense, URL cleanup can be done client-side in `_tracking.html.heex` entirely 👍

aerosol: Perhaps setting those params in there (to null values) will work?

ukutaht: Can also provide a non-nil `dogfood_path` here: https://github.com/plausible/analytics/blob/4603aa7a51e825ca74faf7d2cccdd40cf997f2bc/lib/plausible_web/controllers/stats_controller.ex#L58

This will just [send the pathname](https://github.com/plausible/analytics/blob/4603aa7a51e825ca74faf7d2cccdd40cf997f2bc/lib/plausible_web/templates/layout/_tracking.html.heex#L13) instead of the full URL from the browser.

IDK, maybe simpler than manually cleaning the URL. I've got nothing against stripping bad URL parameters using JS [here](https://github.com/plausible/analytics/blob/4603aa7a51e825ca74faf7d2cccdd40cf997f2bc/lib/plausible_web/templates/layout/_tracking.html.heex#L16) either.

zoldar: Shouldn't we then always set `dogfood_page_path` to `/:dashboard` here then, regardless if the site is plausible.io or not?

ukutaht: I don't think so. The live demo should be tracked as `/plausible.io` and everything else as `/:dashboard`.  If we use `/:dashboard` unconditionally we'll lose the ability to see how many people visit the live demo:

Screenshot 2024-06-12 at 11 36 58

zoldar: Right. Then I guess ignoring query params entirely and passing "/plausible.io" for `dogfood_page_path` for demo dashboard is the way to go? https://github.com/plausible/analytics/pull/4217/commits/58a09b4357f940d8ae33935a1cfb5d06fec79a3e

PR #4174, opened 2024-06-03
Multiple filters on the frontend

macobo:
### Changes

This PR does a few things:
- It adds a new `multiple_filters` feature flag which allows adding multiple filters on the frontend (except for goals).
- Updates filter serialization on the frontend
- Adds new filter (and query) parsing methods (for a follow-up PR)
- Reduces number of internal filter operators we support, collapsing list and single-item operators

image

Blocked on https://github.com/plausible/analytics/pull/4118 - will update filtering within after that PR lands.

### Tests
- [x] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [x] The UI has been tested both in dark and light mode
- [ ] This PR does not change the UI

PR #4218, opened 2024-06-11
Revert "Multiple filters on the frontend (#4174)"

aerosol:
This reverts commit e1f0002d2ebcfae2145001b1c3957158f39e4792.

cc @macobo - see sentry/city filters


macobo (approved):
Approved. For reference the issue seems to be in quoting/not quoting city/region ids.

PR #4219, opened 2024-06-11
Re-apply multiple filters in FE

macobo:
This re-applies https://github.com/plausible/analytics/pull/4218. Fix in last commit: https://github.com/plausible/analytics/pull/4219/commits/81b451733ca7ebdb2b7565351f7cb2e85d3742b0


aerosol (approved):

ukutaht:
What do you think about changing the backend instead to consider an integer value for city ID as valid? And fail validation if String cannot be cast to Integer?

If we keep only accepting String for city ID then I think we should also make sure to cast it to String in API responses, otherwise we won't be accepting the exact thing we return. But forcing it into an Integer feels better to me since that's what it actually is :)

macobo:
@ukutaht Agreed. Will create a follow-up PR for this to keep PR history a bit cleaner though and since it makes most sense to test this with the apiv2 tests.

PR #4221, opened 2024-06-12
Unify default pricing currency selection

aerosol:
### Changes

This PR queries Paddle prices with customer IP, in order to utilise their geolocation-based currency selection.
Additionally, `/api/paddle/currency` endpoint is exposed, so that [currency displayed on the website](https://github.com/plausible/website/pull/579) is the same.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/paddle_controller.ex
+++ b/lib/plausible_web/controllers/api/paddle_controller.ex
@@ -54,7 +54,6 @@ defmodule PlausibleWeb.Api.PaddleController do
       {:ok, currency} ->
         conn
         |> put_status(200)
-        |> Plug.Conn.put_resp_header("cache-control", "max-age=86400, must-revalidate)")
         |> json(%{currency: Cldr.Currency.currency_for_code!(currency).narrow_symbol})
 
       {:error, :fetch_prices_failed} ->
--- a/lib/plausible_web/controllers/api/paddle_controller.ex
+++ b/lib/plausible_web/controllers/api/paddle_controller.ex
@@ -3,7 +3,7 @@ defmodule PlausibleWeb.Api.PaddleController do
   use Plausible.Repo
   require Logger
 
-  plug :verify_signature
+  plug :verify_signature when action in [:webhook]
 
   def webhook(conn, %{"alert_name" => "subscription_created"} = params) do
     Plausible.Billing.subscription_created(params)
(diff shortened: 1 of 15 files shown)


lib/plausible_web/controllers/api/paddle_controller.ex
```diff
@@ -29,6 +29,43 @@ defmodule PlausibleWeb.Api.PaddleController do
...
+
+            {:error, :fetch_prices_failed}
+        end
+      end)
+
+    case result do
+      {:ok, currency} ->
+        conn
+        |> put_status(200)
+        |> Plug.Conn.put_resp_header("cache-control", "max-age=86400, must-revalidate)")
```
aerosol, line 57: @ukutaht @cnkk - I'm not sure if adding cache-control here is necessary or done right even. We'll cache ip=>currency locally, but I was wondering if the CDN could be utilized here as well (or, instead). Since the CDN caches on per region basis, perhaps it's safe to do? 🤔

lib/plausible_web/templates/billing/choose_plan.html.heex
```diff
@@ -1,4 +1,5 @@
+<% remote_ip = PlausibleWeb.RemoteIP.get(@conn) %>
 <%= live_render(@conn, PlausibleWeb.Live.ChoosePlan,
   id: "choose-plan",
-  session: %{"current_user_id" => @user.id}
+  session: %{"current_user_id" => @user.id, "remote_ip" => remote_ip}
```
zoldar, line 4: Why not put the remote IP assigment directly in session map literal?

lib/plausible_web/remote_ip.ex
```diff
@@ -4,26 +4,26 @@ defmodule PlausibleWeb.RemoteIP do
...
-    x_plausible_ip = List.first(Plug.Conn.get_req_header(conn, "x-plausible-ip"))
-    cf_connecting_ip = List.first(Plug.Conn.get_req_header(conn, "cf-connecting-ip"))
-    x_forwarded_for = List.first(Plug.Conn.get_req_header(conn, "x-forwarded-for"))
-    b_forwarded_for = List.first(Plug.Conn.get_req_header(conn, "b-forwarded-for"))
-    forwarded = List.first(Plug.Conn.get_req_header(conn, "forwarded"))
+    x_plausible_ip = List.first(Plug.Conn.get_req_header(conn, "x-plausible-ip")) || ""
+    cf_connecting_ip = List.first(Plug.Conn.get_req_header(conn, "cf-connecting-ip")) || ""
+    x_forwarded_for = List.first(Plug.Conn.get_req_header(conn, "x-forwarded-for")) || ""
+    b_forwarded_for = List.first(Plug.Conn.get_req_header(conn, "b-forwarded-for")) || ""
+    forwarded = List.first(Plug.Conn.get_req_header(conn, "forwarded")) || ""
```
zoldar, line 11: Are there legit cases of empty string headers out in the wild? 🤔  I'm wondering if we shouldn't double-check our haproxy agent too then.

lib/plausible_web/controllers/api/paddle_controller.ex
zoldar: I'm not sure about CDN caching - the region division is not that granular (for instance I remember hitting UA node from PL when investigating remote IP parsing). Also, remote IP passed via various headers might not be even considered by CDN (and probably isn't). I'm wondering if this endpoint shouldn't be queried directly, without CDN to get reliable results?

lib/plausible_web/remote_ip.ex
cnkk: @zoldar Just happened because the proxy config was different between prod and staging 🙈
But in normal cases this should not happen

lib/plausible_web/templates/billing/choose_plan.html.heex
aerosol: https://github.com/plausible/analytics/pull/4221/commits/4c19ffb5dc5671203ccbdee87a569e52730b9f33

lib/plausible_web/controllers/api/paddle_controller.ex
aerosol: Cool, thanks for looking into it. Removed via https://github.com/plausible/analytics/pull/4221/commits/4aabd495fe6d84a24cffbb396d144f97ae7ab96e

zoldar (approved):

PR #4222, opened 2024-06-13
Always recommend a suitable plan on the choose-plan page

RobertJoonas:
### Changes

* Make sure that `last_bill_date` is saved when pulling the sandbox subscription via `mix pull_sandbox_subscription <id>` 
* Refactor: move the `exceeds_last_two_billing_cycles` from `CheckUsage` into `Quota`
* Refactor: split up the quota module and deal with usage and limits in their separate modules.
* If a user has billing cycles defined by their current subscription (i.e. the subscription is `active` or `past_due`), show their last cycle usage on the choose-plan page, rather than last 30 days usage.
* Always set the pageview slider to the "recommended" position on the choose-plan page, rather than setting it to the volume of the current subscription and leaving the buttons unclickable by default.

### Tests
- [x] Automated tests have been revised

### Changelog

### Documentation

### Dark mode


zoldar (approved):
Good call on splitting billing into separate subcontexts 👍

PR #4229, opened 2024-06-14
Add CRM QoL improvements around usage info and linking Paddle

zoldar:
### Changes

A bunch of small improvements to the CRM.

- display usage info directly above user form instead of a separate page opened from list view
- link to respective Paddle subscription profile from `Status` column as well any time paddle ID is available, regardless of the actual status

### DIFF
--- a/lib/plausible/crm_extensions.ex
+++ b/lib/plausible/crm_extensions.ex
@@ -7,25 +7,22 @@ defmodule Plausible.CrmExtensions do
 
   on_ee do
     def javascripts(%{assigns: %{context: "auth", resource: "user", entry: %{} = user}}) do
-      Enum.map(
-        [
-          """
-          <script type="text/javascript">
-            (async () => {
-              const response = await fetch("/crm/auth/user/#{user.id}/usage?embed=true")
-              const usageHTML = await response.text()
-              const cardBody = document.querySelector(".card-body")
-              if (cardBody) {
-                const usageDOM = document.createElement("div")
-                usageDOM.innerHTML = usageHTML
-                cardBody.prepend(usageDOM)
-              }
-            })()
-          </script>
-          """
-        ],
-        &Phoenix.HTML.raw/1
-      )
+      [
+        Phoenix.HTML.raw("""
+        <script type="text/javascript">
+          (async () => {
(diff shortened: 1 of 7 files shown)


lib/plausible_web/views/auth_view.ex
```diff
@@ -9,14 +9,20 @@ defmodule PlausibleWeb.AuthView do
...
+        "unlimited"
+      else
+        PlausibleWeb.StatsView.large_number_format(pageview_limit)
+      end
+
+    if Keyword.get(options, :format) == :long do
+      "#{quota} pageviews"
+    else
+      quota
+    end
```
zoldar, line 25: This change is not strictly related. It has surfaced due to environment mismatch which caused Paddle requests to fail. The `Quota.monthly_pageview_limit()` returns `:unlimited` in such case while recording an entry in Sentry at the same time. As the Quota function degrades gracefully and other spots where the function is used seem to handle the fallback gracefully as well - perhaps it can be altered here accordingly as well?

lib/plausible_web/views/stats_view.ex
```diff
@@ -35,7 +35,7 @@ defmodule PlausibleWeb.StatsView do
           "#{billions}B"
         end
 
-      true ->
+      is_integer(n) ->
```
zoldar, line 38: This will trigger a crash a bit earlier and make the reported error a bit more clear.

lib/plausible/crm_extensions.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Plausible.CrmExtensions do
+  @moduledoc """
+  Extensions for Kaffy CRM
+  """
+
+  use Plausible
+
+  on_ee do
+    def javascripts(%{assigns: %{context: "auth", resource: "user", entry: %{} = user}}) do
+      Enum.map(
```
aerosol, line 10: clever, too bad it's not built statically so we can spot any errors early on.
Why map over a single element though?

aerosol (approved):

lib/plausible/crm_extensions.ex
zoldar: No reason really, I'll change it so that only single element is processed. If there ever will be more conditional JS snippets included, it can always be adjusted.

PR #4236, opened 2024-06-17
Migration: Increase the precision of subscription timestamp fields

RobertJoonas:
### Changes

Motivation: Paddle sometimes sends us two webhooks at the same time when a subscription is updated. We only need to count the latest one of those, for which we need more precision than 1s.

The migration will set both `inserted_at` and `updated_at` columns in the `subscriptions` table to use the `:naive_datetime_usec` type instead of just `:naive_datetime`. In practice, we currently need more precision only on the `updated_at` field. But I suppose it's better to keep the type between both timestamp fields consistent.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20240617102748_increase_subscription_timestamps_precision.exs
+++ b/priv/repo/migrations/20240617102748_increase_subscription_timestamps_precision.exs
@@ -1,5 +1,8 @@
 defmodule Plausible.Repo.Migrations.IncreaseSubscriptionTimestampsPrecision do
   use Ecto.Migration
+  
+  @disable_ddl_transaction true
+  @disable_migration_lock true
 
   def up do
     alter table(:subscriptions) do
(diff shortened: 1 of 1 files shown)


zoldar (approved):
Migrated on staging, everything seems to work fine afterwards.

ukutaht:
Can this be closed @RobertJoonas ? If I remember right this was solved some other way?

aerosol:
>  If I remember right this was solved some other way?

What way? I think this might be a lead for https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7231303811

zoldar:
Yup, finally found it while looking into something else - this particular issue was solved by ignoring particular state update for subscription which always precedes resuming the subscription https://github.com/plausible/analytics/pull/4240

https://github.com/plausible/analytics/blob/master/lib/plausible/billing/billing.ex#L140-L155

We can close this PR for now

PR #4253, opened 2024-06-20
Autoselect prop key

zoldar:
### Changes

Adjustments to the `prop_key` selection in the Properties breakdown tab.

* If any prop keys are available for the current query, and nothing is yet stored, set the default to the first prop key found
* If nothing is available, set the placeholder value to "No custom properties found", and disable the combobox
* Validate stored properties against what prop keys are actually available for selection. If the stored key is not available, pick the first available as the default.
* Never let an applied custom property filter decide what prop key is selected by default (we now support breakdown by other properties that are not filtered by as well, so this behaviour is irrelevant, even disturbing)

### Tests

### Changelog

### Documentation

### Dark mode


RobertJoonas:
Just pushed another fix in https://github.com/plausible/analytics/pull/4253/commits/b559d98800a8c0769d6a87eb3aad96ff4773fa08

### Before

The combobox shows up empty when no prop keys are found, yet the user is still able to interact with it and make another request, which is pointless at that point.

https://github.com/plausible/analytics/assets/56999674/905e00aa-3433-4052-9a4d-494086979fe9

### After

After having returned no results on the initial render of the combobox, we're disabling clicks on it and setting it's placeholder value to "No custom properties found"

https://github.com/plausible/analytics/assets/56999674/0226f8aa-25ef-476d-ac98-451be22d1719

RobertJoonas:
Another fix in https://github.com/plausible/analytics/pull/4253/commits/574eda1bf41680fc8cef6fedf02f7a50ebf82408 

### Before

The previous prop key keeps hanging when the query changes

https://github.com/plausible/analytics/assets/56999674/1cb130e6-b339-4143-96d4-b2ae2d5c0e1d

### After

Set the `propKey` to null, and `propKeyLoading` to true, whenever query changes.

https://github.com/plausible/analytics/assets/56999674/0bd075fd-d68f-44b0-84e4-ea160d90f3ca

PR #4260, opened 2024-06-21
Use max `pageSize` when fetching data for import from UA

zoldar:
### Changes

This PR bumps `pageSize` to the maximum of 100k entries when fetching data for import from Universal Analytics (per https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#ReportRequest). This will make processing larger imports faster and reduce risk of running over hourly requests limit of google API.


aerosol (commented):
Were you aware of https://github.com/plausible/analytics/commit/fb05e447f9636564c03cd06d812099ef316ecb6d?

ukutaht (approved):

RobertJoonas (approved):

PR #4255, opened 2024-06-20
update runtime to elixir-1.17.1 and erlang-27.0

ruslandoga:
### Changes

This PR updates Elixir to [1.17](https://elixir-lang.org/blog/2024/06/12/elixir-v1-17-0-released/) and Erlang to [27](https://www.erlang.org/downloads/27) mostly to bring in [Duration.](https://hexdocs.pm/elixir/1.17.1/Duration.html) I picked the base image with the smallest alpine version difference (3.18.6 vs 3.18.4) but we can update it as well (to 3.20).

`mjml_eex`  (https://github.com/akoutmos/mjml_eex/pull/19) and `credo` had to be updated to make them work again.

Warnings from the new Elixir type checker for phoenix_live_view failed CI tests so I backported https://github.com/phoenixframework/phoenix_live_view/commit/489e8de024e03976e9ae38138eec517fbd456d27 in https://github.com/plausible/analytics/pull/4255/commits/0d07f4df226bd3895d26652656725ca043023101. This workaround can be removed once Plausible starts using `phoenix_live_view` [1.0.0 (current in rc).](https://github.com/phoenixframework/phoenix_live_view/tags)

There are warnings from `ex_money` and `phoenix_view` too but they don't seem to fail CI so I left them as is.
Later `ex_money` can be updated to [5.17.0](https://hexdocs.pm/ex_money/changelog.html#money-v5-17-0) and `phoenix_view` -- to [2.0.4.](https://hexdocs.pm/phoenix_view/changelog.html#2-0-4-2024-06-13)

### Tests
- [x] Tests have been adapted

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #4267, opened 2024-06-24
Update deps - follow up to #4255

aerosol:
Addressing Ruslan's comment from the original PR - getting rid of new warnings during the test run


ruslandoga:
Seems like a `mix format` is needed. Other than that: 👍

ruslandoga (approved):

PR #4275, opened 2024-06-25
Add /var/lib/plausible for easier mounts

ruslandoga:
### Changes

This PR aims to simplify permissions in the "data" volume mounted into Plausible container. After adding [automatic TLS](https://github.com/ruslandoga/plausible/pull/302) we would be mounting a volume in `docker-compose.yml` by default to persist the certificates across restarts, so I thought we might as well allow writes for it by default too.

### What users need to do now

User needs to create a directory on the host and set the correct permissions manually.
After that it can be mounted into the container.

<sub>`console`</sub>
```console
$ cd hosting
$ mkdir ./data
$ chown -R 999:nogroup ./data
```
<sub>`docker-compose.yml`</sub>
```yml
plausible:
  volumes:
    - ./data:/data
  environment:
    - DATA_DIR=/data
```

### How it would be used after this PR

A named volume can be mounted into the container without any extra steps.

<sub>`docker-compose.yml`</sub>
```yml
plausible:
  volumes:
    - plausible-data:/var/lib/plausible
    
volumes:
  plausible-data:
```

### Tests

### Changelog

### Documentation

### Dark mode


Dockerfile
```diff
@@ -71,9 +71,15 @@ RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates
...
+# we need to allow "others" access to app folder, because
+# docker container can be started with arbitrary uid
+RUN mkdir -p /var/lib/plausible && chmod ugo+rw -R /var/lib/plausible
+
 USER 999
 WORKDIR /app
 ENV LISTEN_IP=0.0.0.0
 ENTRYPOINT ["/entrypoint.sh"]
 EXPOSE 8000
+ENV DEFAULT_DATA_DIR=/var/lib/plausible
```
ruslandoga, line 83: This indirection allows to keep everything working as before in EE.

config/runtime.exs
```diff
@@ -232,7 +232,8 @@ maxmind_edition = get_var_from_path_or_env(config_dir, "MAXMIND_EDITION", "GeoLi
 data_dir = get_var_from_path_or_env(config_dir, "DATA_DIR")
 persistent_cache_dir = get_var_from_path_or_env(config_dir, "PERSISTENT_CACHE_DIR")
 
-data_dir = data_dir || persistent_cache_dir
+# DEFAULT_DATA_DIR comes from the container image, please see our Dockerfile
+data_dir = data_dir || persistent_cache_dir || System.get_env("DEFAULT_DATA_DIR")
```
ruslandoga, line 236: Meaning, `PERSISTENT_CACHE_DIR` still takes precedence over the new `/var/lib/plausible`

ruslandoga: I'm using `System.get_env("DEFAULT_DATA_DIR")` since it wouldn't be coming from secrets or mounted files, it's set during container build.

test/plausible/config_test.exs
```diff
@@ -312,14 +329,8 @@ defmodule Plausible.ConfigTest do
...
 
-    test "with only PERSISTENT_CACHE_DIR set" do
-      env = [
-        {"MAXMIND_LICENSE_KEY", "abc"},
-        {"PERSISTENT_CACHE_DIR", "/cache"},
-        {"DATA_DIR", nil}
-      ]
-
-      config = runtime_config(env)
+    test "with only PERSISTENT_CACHE_DIR set", %{env: env} do
```
ruslandoga, line 332: I assume this is what EE does.

cnkk (approved):

PR #4276, opened 2024-06-25
Make new tab dashboard links work

aerosol:
I have no idea what I'm doing wrt react router, but this seems to work?

### DIFF
--- a/assets/js/dashboard.js
+++ b/assets/js/dashboard.js
@@ -31,7 +31,8 @@ if (container) {
     background: container.dataset.background,
     isDbip: container.dataset.isDbip === 'true',
     flags: JSON.parse(container.dataset.flags),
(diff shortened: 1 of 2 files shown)


ukutaht:
Ah, nice find with the `basename` prop. Seems to fix the linking issues indeed 👍

For shared links this should work: https://github.com/plausible/analytics/commit/3ba5a3c28547bbd5f7afbcdd6d5c9cc8ebb06006

aerosol:
There's still some discrepancy between staging/preview and localhost. On localhost I'm getting duplicated domain fragment on left-clicks:

image

aerosol:
@ukutaht how about 7190e7e?

ukutaht:
> @ukutaht how about [7190e7e](https://github.com/plausible/analytics/commit/7190e7e10e9063af8dec9e3754e5a40492c74354)?

Much better 👌

PR #4277, opened 2024-06-25
Count pending ownership sites towards account usage on the upgrade page

RobertJoonas:
### Changes

When the user goes to upgrade to a plan, currently we let them upgrade to a suitable plan based on the usage across their owned sites. When the user has pending ownerships however, they might not be able to accept them, even after upgrading to a seemingly suitable plan, which can be annoying/confusing.

Whenever there are pending site ownerships, a message will be shown to the user:

image

Counting the usage from pending ownerships means that we are not letting them upgrade to a plan that would not accommodate all the pending sites. Same goes for warning about losing access to a feature used be a pending ownership site.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/qouta/usage.ex
+++ b/lib/plausible/billing/qouta/usage.ex
@@ -283,8 +283,8 @@ defmodule Plausible.Billing.Quota.Usage do
         ]
       else
         [
-          {Props, props_usage_query},
-          {RevenueGoals, revenue_goals_usage}
+          {Props, props_usage_q},
+          {RevenueGoals, revenue_goals_usage_q}
         ]
       end
 
--- a/lib/plausible/billing/qouta/usage.ex
+++ b/lib/plausible/billing/qouta/usage.ex
@@ -22,16 +22,36 @@ defmodule Plausible.Billing.Quota.Usage do
            total: non_neg_integer()
          }
 
+  @doc """
+  Returns a full usage report for the user.
+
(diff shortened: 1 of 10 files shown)


lib/plausible/billing/qouta/usage.ex
```diff
@@ -151,151 +171,125 @@ defmodule Plausible.Billing.Quota.Usage do
...
+    additional team member usage. Without this option, usage is queried only
+    across sites owned by the given user.
+  """
+  def team_member_usage(user, opts \\ [])
 
-    excluded_emails =
-      opts
-      |> Keyword.get(:exclude_emails, [])
-      |> List.wrap()
+  def team_member_usage(%User{} = user, opts) do
```
RobertJoonas, line 198: This went through a bit more extensive refactor. I imagine it's pretty difficult to grasp at first sight so let me try to break it down a bit:

* We don't actually need the site option - this was only used to calculate the usage of the "to be transferred" site, which was added to the user's current team member usage. The same can be achieved through the `pending_ownership_site_ids` option.
* When querying memberships, we will count the owner too, and later filter that out by `exclude_emails` which always excludes the given user anyway. That's because for pending ownerships we do want to count owners (after ownership is accepted by the user, the previous owner will remain an admin)
* When querying invitations, we never need to count the owners

ukutaht (approved):

PR #4251, opened 2024-06-20
APIv2: Aggregates, timeseries, conversion_rate, hostname

macobo:
### Changes

- Aggregates now use new QueryBuilder
- Limited timeseries support in APIv2 - everything works, but results are only returned in bucket they're active in
- `conversion_rate` and `group_conversion_rate` metrics support
- Imports-related changes to support apiv2 - next PR will re-simplify
- Goals are preloaded into `Query` object for validation purposes
- event:goal breakdowns support. Now a single query and system is more flexible for renaming goals.
- event:page breakdows support. QueryOptimizer.split now splits query to two - sessions query uses visit:entry page
- fixed the memory leak around event:props:x creating symbols

PR #4283, opened 2024-06-27
APIv2: Replace breakdown module with QueryBuilder

macobo:
### Changes

This PR starts using the new QueryBuilder in breakdown module with a compatibility layer for the old response schema.

The following changes also happen as a side-effect:
- bounce_rate is reported as 0 not nil when site is not entry page
- revenue metrics are handled in breakdown module. When added to apiv2 it would be wise to include currency in response and add validation errors when site does not have access to revenue
- `time_on_page` is a special case within the breakdown module.
- imports.ex gets significantly changed (to remove old compatibility hacks)
- QueryOptimizer also changes event:page to visit:entry_page filter when doing a breakdown by event:page and querying session metrics. Required for imports.
- Garbage filters (which pass validation) cause query to return false
- Breakdown module default ordering subtly changes, causing changes in some tests.
- Simplify query building around order_by

Depends on https://github.com/plausible/analytics/pull/4251

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -5,7 +5,7 @@ defmodule Plausible.Stats.Breakdown do
 
   import Plausible.Stats.{Base, Imported}
   import Ecto.Query
-  alias Plausible.Stats.{Query, Util, TableDecider}
+  alias Plausible.Stats.{Query, Util, TableDecider, QueryResult, SQL}
 
   @no_ref "Direct / None"
   @not_set "(not set)"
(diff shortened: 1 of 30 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -3,344 +3,61 @@ defmodule Plausible.Stats.Breakdown do
...
-  defp paginate_and_execute(_, [], _), do: []
-
-  defp paginate_and_execute(q, metrics, pagination) do
-    q
-    |> apply_pagination(pagination)
-    |> ClickhouseRepo.all()
-    |> Util.keep_requested_metrics(metrics)
-  end
+  defp result_key("event:props:" <> custom_property), do: custom_property
+  defp result_key("event:" <> key), do: key |> String.to_atom()
```
ukutaht, line 50: Does `String.to_existing_atom` work here?

lib/plausible/stats/breakdown.ex
```diff
@@ -3,344 +3,61 @@ defmodule Plausible.Stats.Breakdown do
...
+  defp result_key("visit:" <> key), do: key |> String.to_atom()
+  defp result_key(dimension), do: dimension
 
   defp maybe_add_time_on_page(event_results, site, query, metrics) do
```
ukutaht, line 54: I guess this means we won't be able to order by time_on_page?

lib/plausible/stats/filters/where_builder.ex
```diff
@@ -150,6 +152,15 @@ defmodule Plausible.Stats.Filters.WhereBuilder do
     true
   end
 
+  defp add_filter(table, _query, filter) do
+    Logger.info("Unable to process garbage filter. No results are returned",
```
ukutaht, line 156: At first glance I'm not a fan of silently accepting/ignoring garbage. Would it make sense to error out the whole query here due to invalid filter?

lib/plausible/stats/query_result.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Plausible.Stats.QueryResult do
   @moduledoc false
 
+  use Plausible
```
ukutaht, line 4: What is this for?

lib/plausible/stats/sql/query_builder.ex
```diff
@@ -123,36 +127,23 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
...
-    |> select_merge(^%{shortname(query, dimension) => Expression.dimension(dimension, query)})
-    |> group_by(^Expression.dimension(dimension, query))
-  end
+    key = shortname(query, dimension)
 
-  defp build_order_by(q, query, mode) do
-    Enum.reduce(query.order_by || [], q, &build_order_by(&2, query, &1, mode))
+    q
+    |> select_merge(^%{key => Expression.dimension(dimension, query, key)})
+    |> group_by([], selected_as(^key))
```
ukutaht, line 139: nice 👍

lib/plausible/stats/query_result.ex
macobo: Oops! This can be removed - I had an on_ee flag here previously.

lib/plausible/stats/filters/where_builder.ex
macobo: Not a fan either but there's an existing test that tested against sentry spam when security scanners are run against our API: test/plausible_web/controllers/api/stats_controller/conversions_test.exs:211. 

Not sure if this is indeed the best solution - do you recommend keeping this, removing the test or something else.

lib/plausible/stats/breakdown.ex
macobo: Yes. If needed, I _think_ we can restore this but given we have a `time_on_page` project in plans I can live with this limitation.

lib/plausible/stats/breakdown.ex
macobo: Discussed in basecamp. It does!

lib/plausible/stats/filters/where_builder.ex
ukutaht: >  do you recommend keeping this, removing the test or something else.

Can we make it a 400 Bad Request rather than 200 OK?

ukutaht (approved):

PR #4286, opened 2024-06-28
APIv2: macros, SQL cleanup

macobo:
Cleanup PR. 

QueryBuilder requires selected fields to be aliased via both field names passed to `select_merge` as well as setting `selected_as`. This PR introduces utility macros to reduce this sort of boilerplate.

It also moves more code under Plausible.Stats.SQL

### DIFF
--- a/lib/plausible/stats/sql/fragments.ex
+++ b/lib/plausible/stats/sql/fragments.ex
@@ -156,35 +156,26 @@ defmodule Plausible.Stats.SQL.Fragments do
   def meta_value_column(:meta), do: :"meta.value"
   def meta_value_column(:entry_meta), do: :"entry_meta.value"
 
-  @doc """
-  Convenience Ecto macro for wrapping select_merge where each value gets in turn passed to selected_as.
-
-  ### Examples
-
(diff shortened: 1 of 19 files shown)


lib/plausible/stats/sql/fragments.ex
```diff
@@ -143,9 +156,51 @@ defmodule Plausible.Stats.Fragments do
...
+  @doc """
+  Convenience Ecto macro for wrapping a map passed to select_merge_as such that each
+  expression gets wrapped in dynamic and set as selected_as.
+
+  ### Examples
+
+    iex> wrap_expression([t], %{ foo: t.column }) |> expand_macro_once
+    "%{foo: dynamic([t], selected_as(t.column, :foo))}"
+  """
+  defmacro wrap_expression(binding, map_literal) do
```
ukutaht, line 168: I like this macro but the name feels quite generic. Can we come up with something better? My proposals would be `with_alias`, `with_selected_as`, `wrap_alias` or something along these lines.

ukutaht (approved):

lib/plausible/stats/sql/fragments.ex
macobo: Great suggestion, thanks!

PR #4287, opened 2024-06-28
Fix recommending a tier on the upgrade page

RobertJoonas:
### Changes

A follow-up to https://github.com/plausible/analytics/pull/4277, which started counting pending ownership sites towards usage on the upgrade page. However, it didn't change the logic of how a plan box is highlighted as the "recommended tier". 

This PR fixes that by implementing the following logic behind the recommended tier decision:

* Recommend a plan based on the usage and highest available Growth and Business plans:
  * take the highest growth plan available for the user and see if its usage is exceeded
  * if exceeded, check if the highest Business plan accommodates this usage
  * if even the business plan does not accommodate the usage, do not recommend a tier at all (means that the user should get on a custom plan instead)
* Prioritize a "Recommended" highlight pill for the plan boxes, over "Current".
  * The "Current" highlight will only be displayed if the currently owned tier is also recommended

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/qouta/quota.ex
+++ b/lib/plausible/billing/qouta/quota.ex
@@ -58,6 +58,34 @@ defmodule Plausible.Billing.Quota do
 
   def eligible_for_upgrade?(usage), do: usage.sites > 0
 
+  def ensure_feature_access(usage, plan) do
+    case usage.features -- plan.features do
+      [] -> :ok
+      features -> {:error, {:unavailable_features, features}}
+    end
+  end
+
+  @doc """
+  Suggests a suitable tier (Growth or Business) for the given usage map.
+
+  If even the highest Business plan does not accommodate the usage, then
+  `nil` is returned. This means that this kind of usage should get on a
+  custom plan.
+  """
+  def suggest_tier(usage, highest_growth_plan, highest_business_plan) do
+    growth_check =
+      with :ok <- ensure_within_plan_limits(usage, highest_growth_plan),
+           :ok <- ensure_feature_access(usage, highest_growth_plan) do
+        :ok
+      end
+
+    cond do
+      growth_check == :ok -> :growth
+      ensure_within_plan_limits(usage, highest_business_plan) == :ok -> :business
+      true -> nil
+    end
+  end
+
   defp exceeded_limits(usage, plan, opts) do
     for {limit, exceeded?} <- [
           {:team_member_limit, not within_limit?(usage.team_members, plan.team_member_limit)},
--- a/lib/plausible_web/components/billing/plan_box.ex
+++ b/lib/plausible_web/components/billing/plan_box.ex
@@ -10,7 +10,7 @@ defmodule PlausibleWeb.Components.Billing.PlanBox do
   def standard(assigns) do
     highlight =
       cond do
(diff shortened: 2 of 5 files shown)


lib/plausible/billing/qouta/quota.ex
```diff
@@ -58,6 +58,33 @@ defmodule Plausible.Billing.Quota do
...
+  end
+
+  @doc """
+  Suggests a suitable tier (Growth or Business) for the given usage map.
+
+  If even the highest Business plan does not accommodate the usage, then
+  `nil` is returned. This means that this kind of usage should get on a
+  custom plan.
+  """
+  def suggest_tier(usage, highest_growth_plan, highest_business_plan) do
```
aerosol, line 75: flagging this function for later

aerosol: I had a really long pause reading it. I've noticed we're checking for upgrade eligibility at the LV level. Perhaps encapsulating it here and making a `:custom` recommendation explicit reads better?
This version also checks feature access for business tier, which was omitted previously, not sure if intentionally.


```elixir
  @doc """
  Suggests a suitable tier (Growth or Business) for the given usage map.

  If even the highest Business plan does not accommodate the usage, then
  `:custom` is returned. This means that this kind of usage should get on a
  custom plan.

  `nil` is returned if the usage is not eligible for upgrade.
  """
  def suggest_tier(usage, highest_growth_plan, highest_business_plan) do
    if eligible_for_upgrade?(usage) do
      cond do
        usage_fits_plan?(usage, highest_growth_plan) -> :growth
        usage_fits_plan?(usage, highest_business_plan) -> :business
        true -> :custom
      end
    end
  end

  defp usage_fits_plan?(usage, plan) do
    with :ok <- ensure_within_plan_limits(usage, plan),
         :ok <- ensure_feature_access(usage, plan) do
      true
    else _ -> 
        false
    end
  end
```

RobertJoonas: By "encapsulating it here" do you mean moving this function into the ChoosePlan LV module? Wouldn't mind that but I feel like testing this function separately in the QuotaTest module is more convenient and understandable.

Fair point about the `:custom` recommendation. I was planning to leave the behaviour the same (enterprise tier is never highlighted as the recommended one) and in that case the `:custom` recommendation wouldn't be used. But on second thought, it does make sense to recommend Enterprise if the user is seeing something like this:

image

^no other option there really - might as well display this:

Dark mode:

image

Light mode (opted for a wider blue border around the box, as otherwise it's barely visible and looks weird):

image

> This version also checks feature access for business tier, which was omitted previously, not sure if intentionally.

Yeah that was intentional. There is currently no feature that's not included in the Business tier. Wouldn't hurt to add the additional check though, so that any new feature that gets added only for enterprise is also checked against the business tier.

I've addressed the suggestions in https://github.com/plausible/analytics/pull/4287/commits/ad9af81e1f958e2e9267f447f4f2e363309fc6fd. Let me know what you think 🙏🏼

aerosol: awesome. and great work spotting dead code

aerosol (approved):

PR #4300, opened 2024-07-02
Implement traffic drop notifications

aerosol:
### Changes

This PR implements traffic drop notifications, similar to already implemented spike notifications.

~~Postgres migration included that extends the `spike_notifications` table with a `type` field indicating whether the notification is tracking drops or spikes. Table name should be probably left intact - since we don't have a way to run a migration renaming it.~~

- [x] Extract migration to a separate PR 

It's important we don't send out drop notifications if e.g. traffic is barred via payment issues, hence the worker looks owner's `accept_traffic_until` up.

BTW the PR makes it no longer possible to set a negative threshold.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (approved):
Changes look good to me.

Since we're in a sort of housekeeping mode I would propose some further clean-ups for things that were touched in this PR:
* Make the `settings/email-reports` page a Liveview
* Remove `Double` dependency from traffic_change_notifier_tests.exs file

PR #4302, opened 2024-07-02
Add listing sites, goals and custom props to Sites API

zoldar:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


macobo:
100 seems quite low as a maximum limit and 10 as a default - why not 100 as default and 1000 as maximum? The query and response sizes are quite minimal and we're just inconveniencing users.

test/plausible_web/controllers/api/external_sites_controller_test.exs
```diff
@@ -527,6 +626,126 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
...
+
+      test "returns goals when present", %{conn: conn, site: site} do
+        goal1 = insert(:goal, %{site: site, page_path: "/login"})
+        goal2 = insert(:goal, %{site: site, event_name: "Signup"})
+        goal3 = insert(:goal, %{site: site, event_name: "Purchase", currency: "USD"})
+
+        conn = get(conn, "/api/v1/sites/goals?site_id=" <> site.domain)
+
+        assert json_response(conn, 200) == %{
+                 "goals" => [
```
macobo, line 653: Problem: users won't be able to construct a valid analytics request based off of this without reverse-engineering our code. We should return a goal name and id.

test/plausible_web/controllers/api/external_sites_controller_test.exs
```diff
@@ -527,6 +626,126 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
...
+                   %{
+                     "id" => goal3.id,
+                     "domain" => site.domain,
```
macobo, line 656: Why is this relevant for a given goal? Won't this be obvious from the request?

test/plausible_web/controllers/api/external_sites_controller_test.exs
zoldar: Yes, that is more about consistency with existing API. We could extend the goal payload with name though.

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -29,57 +69,50 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   end
 
   def get_site(conn, %{"site_id" => site_id}) do
-    site = Sites.get_for_user(conn.assigns[:current_user].id, site_id, [:owner, :admin])
+    case get_site(conn.assigns.current_user, site_id, [:owner, :admin]) do
+      {:ok, site} ->
+        json(conn, %{
+          domain: site.domain,
+          timezone: site.timezone,
+          allowed_custom_props: site.allowed_event_props || []
```
macobo, line 77: We should likely name this something else, e.g. `custom_properties`. I don't think our docs refer to `props` anywhere and we're moving away from the name everywhere (to dimensions) except for custom properties.

zoldar:
> 100 seems quite low as a maximum limit and 10 as a default - why not 100 as default and 1000 as maximum? The query and response sizes are quite minimal and we're just inconveniencing users.

Yes, I guess those could be bumped. The current limits are what's currently in effect for Plugin API.

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
zoldar: https://github.com/plausible/analytics/pull/4302/commits/63763f2f078c4e9d16e4499481797917e8dda442

test/plausible_web/controllers/api/external_sites_controller_test.exs
zoldar: Cleaned up in https://github.com/plausible/analytics/pull/4302/commits/645ef196cfdf1edb6698d5bf631159e426c15a61

test/plausible_web/controllers/api/external_sites_controller_test.exs
zoldar: Ditto https://github.com/plausible/analytics/commit/645ef196cfdf1edb6698d5bf631159e426c15a61

zoldar:
Bumped pagination limits in https://github.com/plausible/analytics/pull/4302/commits/93b3c571491f1571f741620ad6814663d434a59e

lib/plausible/goal/schema.ex
```diff
@@ -103,6 +103,7 @@ defimpl Jason.Encoder, for: Plausible.Goal do
     |> Map.put(:goal_type, goal_type)
     |> Map.take([:id, :goal_type, :event_name, :page_path])
     |> Map.put(:domain, domain)
+    |> Map.put(:name, to_string(value))
```
ukutaht, line 106: This should work fine for pageview goals and normal custom event goals. But for revenue goals `to_string` will add the currency to the goal name which is not a valid way to use goals in the Stats API.

Given a revenue goal like Purchase (currency=USD) the API will return:

```json
{
  "name": "Purchase (USD)"
}
```

When a better option would be:

```json
{
  "name": "Purchase"
  "currency": "USD"
}
```

I think that would be better because the API consumer can take the value of `name` and drop it into the Stats API without having to change anything.

ukutaht (commented):
Looks good! Just had one minor comment about revenue goal names.

lib/plausible/goal/schema.ex
zoldar: Addressed in https://github.com/plausible/analytics/pull/4302/commits/cb81b9d7f25f6257a37a318f159846e768824e9a

zoldar:
There's a docs update PR which should go along with this one https://github.com/plausible/docs/pull/521.

This one's ready and awaiting final sign-off \cc @ukutaht

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -2,12 +2,61 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
...
+        site
+        |> Plausible.Goals.for_site_query()
+        |> paginate(params, @pagination_opts)
+
+      json(conn, %{
+        goals:
+          Enum.map(page.entries, fn goal ->
+            %{
+              id: goal.id,
+              name: Goal.name(goal)
```
aerosol, line 44: 1. Please use `event_name` as we'll be extending goals with `display_name` soon-ish.
2. Revenue goals also come with currency, and pageviews have URLs. Maybe we should consider exposing that information.
 
See https://plausible.io/api/plugins/spec/swagger-ui#/default/PlausibleWeb.Plugins.API.Controllers.Goals.index and the `Goal` schema for reference.

aerosol: https://3.basecamp.com/5308029/buckets/26383192/messages/7596522947

aerosol: cc @ukutaht

aerosol: edited the top comment

zoldar: For that matter, I could use jason encoded form defined here https://github.com/plausible/analytics/blob/54cc82a404e31b32cb7209ff57fe241044869941/lib/plausible/goal/schema.ex#L110-L128 and additionally extend it with `display_name` field. WDYT?

aerosol: I guess. Technically it could be a phoenix view? Just wondering what to do with the vague `name` there. I'm going to be looking at this over the next couple of days, so maybe we can catch up and pair?

zoldar: Sure 👍

aerosol (approved):

lib/plausible/goal/schema.ex
aerosol: @ukutaht @zoldar The goals model is my favorite 😬

I think this is problematic and implies a breaking change(s) going forward.

Currently, the Stats API accepts a vague `event:goal` filter, as per the [docs](https://plausible.io/docs/stats-api#filtering):

```
event:goal==Signup|Visit+/register
```

Now that we're looking to introduce display names, in addition to the names under which events are sent, this is becoming tricky because:

1. There is no explicit support for handling Revenue Goals in the **Stats** API - it's either a pageview or anything else, with Revenue Goal being a special case of Custom Event. Feels like querying Revenue Goals through the Stats API, is done by an undocumented accident currently. There is not even a special `goal_type` defined in the existing responses. It seems also impossible to create a Revenue Goal using the Stats API.  
2. In the example above: `Signup` is in fact the `event_name` and `Visit /register` is a display name.
3. Neither `Signup` nor `Visit /register` should be understood as "display names" from the Stats API consumer's perspective, because the objective is to separate "names used in the tracking code" from "names seen in the dashboard" (per @metmarkosaric at https://3.basecamp.com/5308029/buckets/26383192/messages/7596522947#__recording_7596588712). Hence, in the APIs, primarily `event_name` or `page_path` should be used, along with `goal_type` if we're looking to officially support Revenue Goals.
4. Plugins API does the appropriate distinction by introducing explicit `goal_type` to all payloads.

To summarize, my suggestion is to:

1. Explicitly halt all accidental support for Revenue Goals in the Stats API. Stop returning them for now and respond with a descriptive error message that the support is being worked on.
2. Introduce `goal:type` filter to Stats API filters:

  ```
  event:goal==/register;goal:type=pageview
  event:goal==foo;goal:type=custom_event
  ``` 

  and later on:

  ```
  event:goal==Purchase;goal:type=revenue
  ```

3. To handle existing queries backwards, we need to do this horrible assumption:

  ```
  event:goal==Visit... (no type provided)    => assert goal:type=pageview
  event:goal==...      (no type provided)    => assert goal:type=custom_event
  ```

Only then we should update the Sites API:

1. to return structures like:

```json
{
  "event_name": "sign_up",
  "display_name": "Sign up",
  "goal_type": "custom_event"
}
```

but also accept `PUT` payload such as:

```json
{
  "event_name": "sign_up",
  "display_name": "Sign Up",
  "goal_type": "custom_event" # assumed, if the event_name does not start with "Visit"
}
```

Furthermore, even though the API docs for Revenue Goals don't exist, we do have some implementation in place, suggesting we are reducing their meaning to Custom Events at times:

  - https://github.com/plausible/analytics/blob/8a77968c341504e35a1447503004ac23edadc862/test/plausible_web/controllers/api/external_stats_controller/query_test.exs#L557

  - https://github.com/plausible/analytics/blob/8a77968c341504e35a1447503004ac23edadc862/test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs#L2855-L2856


I have no better ideas on how to work around it tbh.

aerosol (changes requested):
IMHO this is no-go for now, until we figure out https://github.com/plausible/analytics/pull/4302#discussion_r1677620812

lib/plausible/goal/schema.ex
ukutaht: Hmm yeah we should think through how goals are returned from Sites/Plugin APIs to work with the planned 'edit goals' feature.

My thinking was that we can treat the new `display_name` field as the main identifier for a goal. The `display_name` should be unique among all the goals for a site so I don't see the need for a goal type filter to be added.

### Custom event goals
```
## As returned from Sites/Plugin API
{
  "goal_type": "custom_event",
  "goal": {
    "event_name": "Signup",
    "display_name": "My Custom Goal Name"
  }
}

## Querying this through the Stats API

event:goal==My Custom Goal Name
```

### Revenue goals

```
## As returned from Sites/Plugin API
{
  "goal_type": "revenue",
  "goal": {
    "event_name": "Purchase",
    "name": "Complete Purchase Flow",
    "currency": "USD"
  }
}

## Querying this through the Stats API

event:goal==Complete Purchase Flow
```

### Pageview goals

```
## As returned from Sites/Plugin API
{
  "goal_type": "pageview",
  "goal": {
    "page_path": "/checkout/thank-you",
    "name": "Visit thank you page"
  }
}

## Querying this through the Stats API

event:goal==Visit thank you page
```

aerosol: @ukutaht Yeah, but my reasoning is, this goes against the decoupling aspect.
If the idea is to let the integrations use 

```html
<button class="plausible-event-name=Button1Click">Click Me</button>
```

in their code, and have it later mapped to "Our Main Marketing Objective" in the dashboard (meant for, and editable by, potentially different audience) it's not very consistent product-wise to expect API consumers treat "Our Main Marketing Objective" as an identifier...

ukutaht: >  "Our Main Marketing Objective" in the dashboard (meant for, and editable by, potentially different audience) it's not very consistent product-wise to expect API consumers treat "Our Main Marketing Objective" as an identifier...

In my model of the world the audience using Stats API are part of the marketing/analyst team in an organization. So if a marketer renames an event to "Our Main Marketing Objective" and then uses the Stats API to fill a spreadsheet or view a Looker Studio report, it should appear with the new goal name.

I agree it has the downside of potential breakage if a goal name is changed. But it's a matter of perspective - I can also 'break' an API integration by deleting the goal in question through the dashboard. The API response will change when the underlying data model changes. If this does prove to be a problem I would propose introducing a `goal_id` dimension so that clients can fetch and store a more stable identifier that doesn't change when the name changes.

ukutaht: In other words if we decouple between:
* Engineering -> integrating Plausible, tagging events
* Marketing -> configuring & consuming the dashboard

I would say the Stats API is more designed for the second audience, not the first.

aerosol: Thanks @ukutaht - I guess that simplifies things on our end and display name work is no longer blocked. Now it’s a matter of equipping the Stats/Sites APIs with official revenue goals support.

zoldar: Okay, I have changed goal in Sites API so that it always lists: `id`, `name`, `goal_type` (event|page), `event_name` and `page_path`. `currency` is unexposed for now. The `name` is either event name or "Visit {page_path}". It's also used to build `to_string` representation of the goal, which is used for display name in Plugin API (at the same time I have got rid of `Plausible.Goal.display_name` as an explicit function to reduce confusion, hopefully).

https://github.com/plausible/analytics/pull/4302/commits/6ada31d3ae65ae05b60ecd2e727e6415f49ae96e

Does that look ok to you @aerosol ?

aerosol: This is good, just `name` should be `display_name` - even if it breaks current accidental support for revenue goals (as it will include the currency too). Because going forward, this will be the name that marketing users will be able to change.

zoldar: https://github.com/plausible/analytics/pull/4302/commits/a092d27a13c9b441b01428a25a7939adbbafe667

aerosol (approved):

zoldar:
I have almost missed one important thing - it must be possible to get site details and list site goals by members who have only viewer access to given sites (with a valid Stats/Sites API token, of course). 

Done in https://github.com/plausible/analytics/pull/4302/commits/8b91312789e13f7e1e2af2fef1bacdc95c7124ee

\cc @aerosol

lib/plausible/goal/schema.ex
ukutaht: > (as it will include the currency too). Because going forward, this will be the name that marketing users will be able to change.

Why include the currency in `display_name`? I started this discussion because I thought it was better to not include it (https://github.com/plausible/analytics/pull/4302#discussion_r1668581322)

aerosol: Oops, I missed that. I guess no other reason than it's already there in the Plugins API under the same label.

ukutaht: Ah yeah that would be slightly inconsistent between the different APIs.

I've asked if Daan relies on the current behaviour (https://3.basecamp.com/5308029/buckets/32037438/chats/5956641161@7620983345)

zoldar: Ok, holding off until we settle it is safe to remove. If yes, we remove currency from display name altogether, correct?

zoldar: I'm wrapping it up with https://github.com/plausible/analytics/pull/4302/commits/a59186ea9d2bba0abbe881eebb53158638b4cdfe. The currency is no longer included in goal's computed display name. The goals list in "Goal Settings" view still displays the name with currency appended.

PR #4305, opened 2024-07-03
APIv2: TimeSeries using QueryBuilder, release `experimental_session_count`

macobo:
### Changes

This PR continues work on APIv2, changing TimeSeries module to use QueryBuilder under the hood. This simplifies operations significantly.

In addition:
- New `include.time_labels` option is added which includes the time range labels under `meta.time_labels`. This is useful when constructing a graph from the time series data.
- Release `experimental_session_count` which includes all sessions _active_ in a time range and for hourly/minute graphs smears them across all time buckets they were active in.


lib/plausible/stats/query.ex
```diff
@@ -231,9 +215,13 @@ defmodule Plausible.Stats.Query do
   end
 
   def set(query, keywords) do
-    query
-    |> struct!(keywords)
-    |> refresh_imported_opts()
+    new_query = struct!(query, keywords)
+
+    if Keyword.has_key?(keywords, :include_imported) do
```
ukutaht, line 220: Not a fan of this workaround. Where do we actually rely on the `refresh_imported_opts` behaviour? Maybe it makes sense to refresh the imported opts explicitly at the call site when needed.

macobo: Will handle separately

ukutaht (approved):

PR #4309, opened 2024-07-04
Stop listing UA views in Google data import flow

zoldar:
### Changes

As UA views have their data removed since July 1st and are effectively defunct, we stop listing them among properties and views for import. Full removal of backing logic will follow.

### Tests

### Changelog


RobertJoonas (approved):
👍

PR #4312, opened 2024-07-04
Remove Universal Analytics import logic

zoldar:
### Changes

This PR removes all code for importing Universal Analytics except bare minimum of import source definition needed for rendering existing, historical imports.


RobertJoonas (approved):
Must feel nice to delete so much code. Brilliant! ✨

PR #4318, opened 2024-07-08
Implement search in Details views

RobertJoonas:
### Changes

* Refactor the structure of a metric on the frontend - use a Metric class with special creator functions.
* Create a higher-order component for keeping the query context (such as the query object itself and additional information such as `revenueAvailable`, `importedDataInView`, `lastLoadTimestamp`)
* Create a new BreakdownModal component that includes the search functionality, and make all current "Details" view modals to use it.
  * Note: some API endpoints (goals, countries, regions, cities) do not support a contains filter. This PR will start using the new BreakdownModal in those reports too, but the search is explicitly disabled via `enableSearch={false}` prop.

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/components/query-context-hoc.js
```diff
@@ -0,0 +1,74 @@
...
+// * `lastLoadTimestamp` - used for displaying a tooltip with time passed since
+//   the last update in realtime components.
+
+export default function withQueryContext(WrappedComponent) {
+  return (props) => {
+    const { site, location } = props
+    
+    const [query, setQuery] = useState(parseQuery(location.search, site))
+    const [importedDataInView, setImportedDataInView] = useState(false)
+    const [revenueAvailable, setRevenueAvailable] = useState(false)
```
apata, line 28: **nitpick**: I had a closer look at `revenueAvailable,` and based on my reading, it's a calculated value. One of the principles of React state is for it not to contain anything duplicated or redundant ([reasoning](https://react.dev/learn/managing-state#choosing-the-state-structure)), and a value you can calculate is redundant. Therefore, I'm suggesting we remove the useState hook and calculate the value directly. If the calculation is expensive, we can use useMemo to limit its recalculation.

assets/js/dashboard/components/query-context-hoc.js
```diff
@@ -0,0 +1,74 @@
...
+        return goalFilters.some(([_op, _key, clauses]) => {
+          return clauses.includes(rg.event_name)
+        })
+      })
+
+      const singleCurrency = revenueGoalsInFilter.every((rg) => {
+        return rg.currency === revenueGoalsInFilter[0].currency
+      })
+
+      setRevenueAvailable(revenueGoalsInFilter.length > 0 && singleCurrency)
```
apata, line 54: **suggestion**: I'm loving the way you refactored some of the responsibility that was in component render code to a separate but colocated utility function (chooseMetrics). 

Would that be possible with this code as well? A separate unit-testable `function shouldShowRevenueAvailable(...)` would I think increase code readability and future maintainability.

assets/js/dashboard/components/query-context-hoc.js
RobertJoonas: Thanks, @apata. Makes perfect sense! I've created a new `revenueAvailable` predicate function in https://github.com/plausible/analytics/pull/4318/commits/601b849e2796a5739e2d3752a497cbb2f8609634

assets/js/dashboard/stats/modals/breakdown-modal.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    searchBox.addEventListener('keyup', handleKeyUp);
+
+    return () => {
+      searchBox.removeEventListener('keyup', handleKeyUp);
+    }
+  }, [])
+
+  const fetchData = useCallback(debounce(() => {
```
ukutaht, line 130: I don't think the debounce here is working as it should.

The idea with debouncing search queries is that if I write a phrase quickly into the search box, the intermediate results are not requested. Only the full phrase is searched once.

But when I write 'long' into the search box using our seeds, I see 4 requests in the network tab:
* Search with `page: 'l'`
* Search with `page: 'lo'`
* Search with `page: 'lon'`
* Search with `page: 'long'`


When it should only fire a single API request for:
* Search with `page: 'long'`

I think the debounce function should be used before the `onChange` handler fires so that the `search` prop is only updated after the debounce step.

ukutaht (changes requested):

assets/js/dashboard/query.js
```diff
@@ -134,6 +139,26 @@ export function filtersBackwardsCompatibilityRedirect() {
...
+// Returns a boolean indicating whether the given query includes a
+// non-empty goal filterset containing a single, or multiple revenue
+// goals with the same currency. Used to decide whether to render
+// revenue metrics in a dashboard report or not.
+export function revenueAvailable(query, site) {
+  const revenueGoalsInFilter = site.revenueGoals.filter((rg) => {
+    const goalFilters = getFiltersByKeyPrefix(query, "goal")
+    
+    return goalFilters.some(([_op, _key, clauses]) => {
+      return clauses.includes(rg.event_name)
```
macobo, line 151: Can't revenue goals be pageview goals?

assets/js/dashboard/stats/reports/list.js
```diff
@@ -260,7 +251,12 @@ export default function ListReport(props) {
           plot={metricToPlot}
         >
           <div className="flex justify-start px-2 py-1.5 group text-sm dark:text-gray-300 relative z-9 break-all w-full">
-            <FilterLink filterQuery={getFilterQuery(listItem)} onClick={props.onClick || noop}>
+            <FilterLink
+              query={props.query}
+              filterInfo={props.getFilterFor(listItem)}
+              onClick={props.onClick || noop}
```
macobo, line 257: Why is the || noop needed? Can't the FilterLink component handle undefined as others do usually?

assets/js/dashboard/stats/modals/props.js
```diff
@@ -1,137 +1,63 @@
...
+    return {
+      prefix: `${EVENT_PROPS_PREFIX}${propKey}`,
+      filter: ["is", `${EVENT_PROPS_PREFIX}${propKey}`, [listItem.name]]
+    }
+  }, [])
 
-    return (
-      <>
-        <h1 className="text-xl font-bold dark:text-gray-100">{specialTitleWhenGoalFilter(query, 'Custom Property Breakdown')}</h1>
+  const addSearchFilter = useCallback((query, s) => {
```
macobo, line 34: Nit: `s` -> `searchPhrase`.

assets/js/dashboard/stats/modals/locations-modal.js
```diff
@@ -0,0 +1,75 @@
...
+    
+    return [
+      metrics.createVisitors({renderLabel: (_query) => "Visitors" }),
+      currentView === 'countries' && metrics.createPercentage()
+    ].filter(m => !!m)
+  }
+  
+  let renderIcon
+
+  if (currentView === 'countries') {
```
macobo, line 54: Not sure this works as intended - IIRC using conditionals like this in react can cause issues.

I'd instead use `useMemo(() => ..., [currentView])` or extract subcomponent.

assets/js/dashboard/stats/modals/breakdown-modal.js
```diff
@@ -0,0 +1,307 @@
...
+    site,
+    query,
+    reportInfo,
+    metrics,
+    renderIcon,
+    getExternalLinkURL,
+    searchEnabled,
+    afterFetchData,
+    afterFetchNextPage
+  } = props
```
macobo, line 90: Nit: `props.something` is _almost_ not used below.

I'd move this deconstruction into the function header.

assets/js/dashboard/stats/modals/breakdown-modal.js
```diff
@@ -0,0 +1,307 @@
...
+  
+  const endpoint = `/api/stats/${encodeURIComponent(site.domain)}${reportInfo.endpoint}`
+  const isSearchEnabled = searchEnabled === false ? false : true
```
macobo, line 93: What's going on here?

Why not use `searchEnabled` prop directly? Is it to do with defaults.

assets/js/dashboard/query.js
RobertJoonas: Revenue goals are always custom event goals.

assets/js/dashboard/stats/modals/locations-modal.js
RobertJoonas: Moved the conditionals into the map defining the report info. https://github.com/plausible/analytics/pull/4318/commits/c739005e371db27aa4be0aac59e8e2a0e01a35b1

assets/js/dashboard/stats/modals/props.js
RobertJoonas: Did a few renamings and wrote out the whole word instead of `s` and `m`. https://github.com/plausible/analytics/pull/4318/commits/0efc3c9422a8470784c53484458b5fd8ccf5ac3d

assets/js/dashboard/stats/modals/breakdown-modal.js
RobertJoonas: Good call. Done in https://github.com/plausible/analytics/pull/4318/commits/016245f007ea640150642ac46b8c54d8249b5140

assets/js/dashboard/stats/modals/breakdown-modal.js
RobertJoonas: Yeah, it's supposed to be `true` by default. Found a better way to express that now in https://github.com/plausible/analytics/pull/4318/commits/78e44fa256db92b32c2d4b31c6fe5709f71b90fd

ukutaht (approved):

PR #4319, opened 2024-07-08
Limit excluded goals to 1000 in goal suggestions

zoldar:
### Changes

This is a fix for situations where large number of existing goals caused the suggestions query run over the limit of parameters in the CH query. As, worst case, this will cause suggesting goal which is already added (which is later validated anyway), it's safe to trim the list of exclusions.

The fix is not ideal but will address the immediate problem of goal settings failing to load for sites with large number of goals defined.


macobo (approved):
Do you have a good testcase for this in production? I'd like to revert this after fixing the underlying issue.

PR #4320, opened 2024-07-08
Link to the correct CRM page from User Settings > People

RobertJoonas:
### Changes

Currently the CRM links in User Settings > People always open up the user profile of the current user in the CRM. We want those links to take us to that user's page instead.


zoldar (approved):
🤦 LGTM 😅

PR #4278, opened 2024-06-25
Fix new tab links + shared links losing URL context

aerosol:
Another stab at #4276

PR #4323, opened 2024-07-09
Fix URLs breaking on date picker navigation

aerosol:
This prevents the date picker from updating the URL with data domain on each subsequent use. Only the query string should be updated.

Possibly regression introduced via #4278


assets/js/dashboard/query.js
```diff
@@ -49,10 +49,11 @@ export function parseQuery(querystring, site) {
 }
 
 export function navigateToQuery(history, queryFrom, newData) {
+  console.info('navigateToQuery', newData, queryFrom)
```
ukutaht, line 52: Dangling debug statement

ukutaht (approved):
❤️

assets/js/dashboard/query.js
aerosol: ah thanks

PR #4328, opened 2024-07-09
APIv2: visit:country_name, visit:region_name, visit:city_name dimensions

macobo:
This PR:
- Creates a migration to sync https://github.com/plausible/location package results into ClickHouse
- The sync happens daily when the library changes
- Adds new dictionary and ALIAS columns for convenient access to location names
- Adds support for visit:country_name, visit:region_name, visit:city_name dimensions in APIv2

Note the migrations/app check will fail in this PR - that's expected and OK in this case.

Depends on https://github.com/plausible/location/pull/25


mix.exs
```diff
@@ -93,7 +93,7 @@ defmodule Plausible.MixProject do
       {:hackney, "~> 1.8"},
       {:jason, "~> 1.3"},
       {:kaffy, "~> 0.10.2", only: [:dev, :test, :staging, :prod]},
-      {:location, git: "https://github.com/plausible/location.git"},
+      {:location, git: "https://github.com/plausible/location.git", branch: "all-all"},
```
macobo, line 96: TODO: Revert this once https://github.com/plausible/location/pull/25 is merged

lib/plausible/clickhouse_location_data.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.ClickhouseLocationData do
+  @moduledoc """
+  Schema for storing location id <-> translation mappins in Clickhouse
```
aerosol, line 3:
```suggestion
  Schema for storing location id <-> translation mappings in ClickHouse
```

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
+defmodule Plausible.DataMigration.LocationsSync do
+  @moduledoc """
+  ClickHouse locations data migration for storing location names in clickhouse.
+
+  Run regularly as plausible/locations data changes.
```
aerosol, line 5: Perhaps rephrase that, so it's clear that it's being run regularly by the background worker - this sounds like some user action is necessary.

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
...
+
+  def run() do
+    cluster? = Plausible.MigrationUtils.clustered_table?("sessions_v2")
+
+    {:ok, _} = run_sql("truncate-location-data-table", cluster?: cluster?)
+    {:ok, _} = run_sql("create-location-data-table", cluster?: cluster?)
+
+    countries =
+      Location.Country.all()
+      |> Enum.map(fn country -> %{type: "country", id: country.alpha_2, name: country.name} end)
```
aerosol, line 85: here and below I think it's worth pattern matching on `Location` structs, e.g. 

```suggestion
      |> Enum.map(fn %Location.Country{} = country -> %{type: "country", id: country.alpha_2, name: country.name} end
```

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
...
+
+    subdivisions =
+      Location.Subdivision.all()
+      |> Enum.map(fn subdivision ->
+        %{type: "subdivision", id: subdivision.code, name: subdivision.name}
+      end)
+
+    cities =
+      Location.City.all()
+      |> Enum.map(fn city -> %{type: "city", id: Integer.to_string(city.id), name: city.name} end)
```
aerosol, line 95:
```suggestion
      |> Enum.map(fn %Location.City{} = city -> %{type: "city", id: to_string(city.id), name: city.name} end)
```

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
...
+  @moduledoc """
+  ClickHouse locations data migration for storing location names in clickhouse.
+
+  Run regularly as plausible/locations data changes.
+
+  SQL files available at: priv/data_migrations/LocationsSync/sql
+  """
+  alias Plausible.ClickhouseLocationData
+
+  use Plausible.DataMigration, dir: "LocationsSync", repo: Plausible.IngestRepo
```
aerosol, line 11: Have you considered using `AsyncInsertRepo` here? I'm thinking maybe not sharing the ingest pool makes the intent cleaner.

Anyway, wondering if it's OK to use `Plausible.DataMigration` as a part of regular production runtime. Not sure what's the benefit, SQL templates?
One concern /could be irrelevant/ is, there's some stdout ANSI noise produced, that we normally don't do.

lib/plausible/stats/imported/sql/builder.ex
```diff
@@ -318,6 +318,10 @@ defmodule Plausible.Stats.Imported.SQL.Builder do
   defp filter_group_values(q, "visit:region"), do: where(q, [i], i.region != "")
   defp filter_group_values(q, "visit:city"), do: where(q, [i], i.city != 0 and not is_nil(i.city))
 
+  defp filter_group_values(q, "visit:country_name"), do: where(q, [i], i.country_name != "ZZ")
```
aerosol, line 321: I'm seeing `\0\0` [elsewhere](https://github.com/plausible/analytics/blob/location-sync/lib/plausible/stats/breakdown.ex#L194) in similar  clauses. Is it relevant here?

lib/workers/locations_sync.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Plausible.Workers.LocationsSync do
+  @moduledoc false
+
+  use Plausible.Repo
+  use Oban.Worker, queue: :locations_sync
```
aerosol, line 5: This module could use some tests + I think we should explicitly configure that queue. See `base_queues` in `runtime.exs`.

aerosol (commented):
Unsolicited review inline

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Plausible.IngestRepo.Migrations.PopulateLocationData do
+  use Ecto.Migration
+
+  def up do
+    try do
+      Location.load_all()
+    rescue
+      _ -> nil
```
aerosol, line 8: anything specific we're looking to ignore here?

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
...
+    case run_sql("get-location-data-table-comment") do
+      {:ok, %{rows: [[stored_version]]}} -> stored_version != Location.version()
+      _ -> true
+    end
+  end
+
+  def run() do
+    cluster? = Plausible.MigrationUtils.clustered_table?("sessions_v2")
+
+    {:ok, _} = run_sql("truncate-location-data-table", cluster?: cluster?)
```
aerosol, line 80: So for some unspecified period of time the location table is empty? Shall we expect varying API results at around midnight then? Wondering if building the dictionary based on Postgres table wouldn't be better to avoid the gaps?

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+    Plausible.DataMigration.LocationsSync.run()
```
aerosol, line 11: The migration is shared between static migrations and runtime cycles. Concern is, if the data migration code changes, we might not catch the regression since this module will be run only once in most environments having any data.

lib/plausible/stats/imported/sql/builder.ex
macobo: I don't think any of this is - artefact of old APIs. We'll (over time) remove these clauses or move them to controller -> frontend.

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
macobo: If it's already loaded it will error. We can safely ignore that. Will add a comment.

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,123 @@
...
+      table: "imported_locations",
+      column_name: "city_name",
+      type: "city",
+      input_column: "city"
+    }
+  ]
+
+  def out_of_date?() do
+    case run_sql("get-location-data-table-comment") do
+      {:ok, %{rows: [[stored_version]]}} -> stored_version != Location.version()
```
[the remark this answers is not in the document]
macobo: Yup.

lib/plausible/data_migration/locations_sync.ex
macobo: Postgres is not a good option 
- _more_ moving pieces
- it requires us to _ensure_ that postgres is reachable from clickhouse which is a pain in self-hosted setups
- complicates dictionary setup a lot
- we would still need the data version check unless we want to wipe/rewrite the the postgres table each time. Note that it's not guaranteed upsert is enough - entries may be deleted over time.

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
macobo: We have used this pattern before [1](https://github.com/plausible/analytics/blob/8c6234604a6e080fceec410d8f2eb5848d810b2d/priv/ingest_repo/migrations/20240222082911_sessions_v2_versioned_collapsing_merge_tree.exs#L5) [2](https://github.com/plausible/analytics/blob/8c6234604a6e080fceec410d8f2eb5848d810b2d/priv/ingest_repo/migrations/20230320094327_create_v2_schemas.exs#L13)

You're right that we need to be careful around changing the migration code - will add a note around it needing to keep behavior idempotent.

lib/plausible/data_migration/locations_sync.ex
macobo: Not sure if AsyncInsertRepo works here - we want to ensure that insert has succeeded before we recreate the dictionary.

> Anyway, wondering if it's OK to use Plausible.DataMigration as a part of regular production runtime. Not sure what's the benefit, SQL templates?

Yep. This is a complicated sequence of steps where some dynamicism is needed. Given that's what a data migration is this fits like a fist into an eye socket.

On ANSI noise I have no strong opinions. If you do - how would you feel about if we explicitly silenced it when running it in cron?

lib/workers/locations_sync.ex
macobo: Do we need a custom queue for this or can this run off of default?

Note this is a fast job - even when re-inserting everything it will take ~1s.

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
aerosol: I meant to specifically catch `{:error, :badarg}` maybe

lib/plausible/stats/filters/filters.ex
```diff
@@ -23,6 +23,9 @@ defmodule Plausible.Stats.Filters do
     :country,
     :region,
     :city,
+    :country_name,
```
ukutaht, line 26: Should these new filters be usable? If so, I think we should add some tests for the new behaviour

mix.exs
ukutaht, line 96: Did you branch this off current `main`? I know that @ruslandoga has been working on the location library so the `main` branch has some unreleased stuff. @ruslandoga do you think it's in a good/safe place to release now?

macobo: I did branch it off of the most recent `main`.

lib/plausible/stats/filters/filters.ex
macobo: They are. I was lazy and know that since using it as a `dimension` works it works here but will add a test.

lib/workers/locations_sync.ex
aerosol: I think you probably want to limit concurrent consumers to 1. But what I meant is the job will never execute as it is currently.

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
aerosol: If prior art is enough to justify using the (anti) pattern here, we might as well skip the comment - they're known to prevent nothing :D

lib/plausible/data_migration/locations_sync.ex
aerosol: > fits like a fist into an eye socket

Brain paused on that one

> On ANSI noise I have no strong opinions. If you do - how would you feel about if we explicitly silenced it when running it in cron?

Yeah I don't feel strongly about it either, just wanted to flag it.

lib/workers/locations_sync.ex
macobo: Thanks! This caught a different bug as well. TIL :facepalm:

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
macobo: It throws an ArgumentError, but updated, thanks!

aerosol: Same thing, depends if you catch or rescue

lib/plausible/stats/filters/filters.ex
macobo: Added tests

mix.exs
ruslandoga: https://github.com/plausible/location/pull/20 would probably need to be merged first. I'm not sure how good the current translations are.

lib/plausible/migration_utils.ex
```diff
@@ -13,4 +13,17 @@ defmodule Plausible.MigrationUtils do
...
     end
   end
+
+  # See https://clickhouse.com/docs/en/sql-reference/dictionaries#clickhouse for context
+  def dictionary_connection_params() do
+    Plausible.IngestRepo.config()
+    |> Enum.map(fn
+      {:database, database} -> "DB '#{database}'"
+      {:username, username} -> "USER '#{username}'"
+      {:password, password} -> "PASSWORD '#{password}'"
```
ruslandoga, line 23: Should this be escaped?

lib/plausible/data_migration/locations_sync.ex
```diff
@@ -0,0 +1,140 @@
...
```
ruslandoga, line 85: Should there be an ordering, newer = better? What if the library is updated and deployed twice in a single day?

macobo: > Should there be an ordering, newer = better?

I don't think it's for this migration to care about. All it knows is that this is out of sync with the main dependency.

> What if the library is updated and deployed twice in a single day?

Given that the `version` doesn't hold any semantic meaning here that's up to the library maintainers to keep managed, by changing the schema or something else. :)

ruslandoga: I meant a possibility of a race condition. I guess it would be resolved eventually. Or is it prevented somehow?

config/runtime.exs
```diff
@@ -566,7 +566,9 @@ base_queues = [
   analytics_exports: 1,
   notify_exported_analytics: 1,
   domain_change_transition: 1,
-  check_accept_traffic_until: 1
+  check_accept_traffic_until: 1,
+  clickhouse_clean_sites: 1,
+  locations_sync: 1
```
ruslandoga, line 571: Note that queues limits are per node, not per cluster. So in a cluster of three app nodes, three locations syncs could be performed at the same time. https://hexdocs.pm/oban/Oban.html#module-configuring-queues

priv/ingest_repo/migrations/20240709181437_populate_location_data.exs
ruslandoga: Other than prior art, is there a particular reason it can't be split into a normal Ecto migration that creates the dictionary and tables and a separate function in a context like Plausible.Locations that only does what's necessary to update the dictionary? These data migrations make it harder to keep CE easily updatable.

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
```diff
@@ -691,5 +694,53 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryImportedTest do
...
+
+      populate_stats(site, site_import.id, [
+        build(:pageview,
+          country_code: "GB",
+          # London
+          city_geoname_id: 2_643_743
+        ),
+        build(:pageview,
+          country_code: "CA",
+          # Different London
```
ruslandoga, line 712: This is more of a question to make sure I understand this part and the related Locations change. What is the expected behavior when the user filters by city=London. Do the stats for both Londons get returned or for only the latest one inserted into the dictionary?

macobo: > What is the expected behavior when the user filters by city=London. Do the stats for both Londons get returned or for only the latest one inserted into the dictionary? In which case we should probably prioritize the one with a larger population in Location.City.all/0.

If filtering my `["is", "visit:city_name", ["London"]]`, stats for all cities named london are returned as is tested here.

ruslandoga: > as is tested here.

I'm not seeing visit:city_name=London being passed into the requests. Where does it happen?

macobo: Sorry for the confusion. In this case it's used as a dimension = as a group by. There's 35 total pageviews - 1 from GB London + 1 from CA London + 33 from imports, even though all are from different countries.

ruslandoga: Is there a test for a request `["is", "visit:city_name", ["London"]]` filter?

macobo: > Does the dictionary use the value for the lookups instead of the key then?

Yes. [The dictionary uses the city.id and other identifiers for lookups](https://github.com/plausible/analytics/pull/4328/files#diff-30855ad8bd4a5454e0bb315f1d6376fa335a09c17f0db2735ef4774e25fbafadR5). See also https://clickhouse.com/docs/en/sql-reference/functions/ext-dict-functions

> So is there a test for a request with ["is", "visit:city_name", ["London"]] filter?

No test besides this one where labels collide, but [you can see a different test using the new dimensions as filters above](https://github.com/plausible/analytics/pull/4328/files#diff-3b7c6c9dbd8b3c38b6c9e09888ded35d80c9a6d624551a73959f48942bde47c5R2817).

ruslandoga: OK, I don't get it :) I'll try to reproduce the test I want on my own later. Thank you for trying to explain it to me!

ruslandoga: I realized what confused me. I thought that the dict was supposed to be used for filtering but it's not really. It's just doing the lookups and the WHERE clause becomes a scan but it's OK because ClickHouse.

Setup

```sql
CREATE TABLE location_data
(
    `type` LowCardinality(String),
    `id` String,
    `name` String
)
ENGINE = MergeTree;

INSERT INTO location_data (type, id, name) Values ('city', 123, 'London'), ('city', 234, 'London'), ('city', 456, 'not London');

CREATE OR REPLACE DICTIONARY location_data_dict
(
    `type` String,
    `id` String,
    `name` String
)
PRIMARY KEY type, id
SOURCE(CLICKHOUSE(TABLE location_data))
LIFETIME(MIN 0 MAX 0)
LAYOUT(COMPLEX_KEY_CACHE(SIZE_IN_CELLS 500000));
 
 CREATE TABLE data
(
    `page` String,
    `visits` UInt64,
    `city_id` String,
    `city` String ALIAS dictGet('location_data_dict', 'name', ('city', city_id))
)
ENGINE = MergeTree
ORDER BY tuple();

INSERT INTO data (page, visits, city_id) Values ('/a', 10, '123'), ('/b', 20, '234'), ('/c', 30, '456');
```

What I was confused about

```sql
SELECT * FROM data WHERE city = 'London';
```

This query does `dictGet` for all rows, including not London, and only then does the filtering which is OK since there are likely other fields which are actually used in the primary key.

macobo: Seems correct to me :)

lib/plausible/data_migration/locations_sync.ex
ruslandoga: Can this be changed with something like `Location.compare(v1, v2) :: :gt | :eq | :lt`? Then it would be easier to understand what guarantees the library needs to provide.

macobo: I'm not sure what advantage that would provide us for this functionality - are we future-proofing something unneccessarily?

ruslandoga: It's an easy way to prevent race conditions. The change is easy and makes the contract between the library and the app explicit.

macobo: I'd say want the app and library to be in sync over only passing forward. If we need to roll back the library version for some reason then having it behave differently for clickhouse vs out of it is surprising behavior.

In production we have a single version live at a time (except for a short period during deployments). I assume we can assume the same for self-hosted who will be upgrading rarely?

ruslandoga: I think these two examples would make things clearer:

Example one, one nodes is slow to update

0. Locations version is 1
1. Locations are updated, the new version is 2
2. Deployment, three nodes, one of the nodes fails to update or is slow to update
3. Midnight, dictionary is reloaded
4. Two healthy nodes update Locations version to 2
5. The node that is slow to update runs the dictionary reset last, Location version is back to 1

Example two, Locations is updated twice in a day

0. Locations version is 1
1. Locations are updated, the new version is 1
2. Deployment, three nodes again, one fails or is slow to update again
3. Midnight, dictionary is reloaded, but not really

Is it possible in reality? ¯\_(ツ)_/¯ My personal preference is that if it's easy to prevent, it should be prevented. In this case it's easy to prevent the possible problems by making the contract explicit.

macobo: > That library doesn't have a concept of rolling back. 

Yes, but users may revert PRs that update mix.exs. In your proposed solution this would result in the library version and clickhouse to go out of sync.

----

This thread is too long. If this is a blocker let's jump into a call and discuss this face-to-face.

ruslandoga: > In your proposed solution this would result in the library version and clickhouse to go out of sync.

Not if version=app_version_data_version. Reverting PRs still moves the version tag forward. But you are right, it's getting too long and my review was unsolicited anyway :) I'm marking this as resolved.

mix.exs
ruslandoga: To avoid blocking this PR for too long, I think https://github.com/plausible/location/pull/20 is safe to merge, I checked many of the translations in that PR manually with GPT and Google and they seem to be good enough.

macobo:
Note: Heading out to vacation. I'll leave it up to @ukutaht whether to push this to production or wait until I'm back whereupon I'll get this merged :)

lib/plausible/migration_utils.ex
macobo: Sadly not trivial to escape. Given this is provided by the person running plausible I don't see it making a huge difference if it were escaped - if an attacker can change plausible app config on disk, they can do anything already.

mix.exs
macobo: Merged the upstream PR!

config/runtime.exs
macobo: This sent me down a bit of a goose chase.

- https://hexdocs.pm/oban/Oban.html#module-periodic-jobs
- https://elixirforum.com/t/oban-periodic-cron-job-multi-node-question/43666

Crons don't need to worry about cross-node limits, but I guess it's better to list it here than not.

RobertJoonas (approved):

PR #4331, opened 2024-07-10
Fix React combobox debounce

RobertJoonas:
### Changes

In the `PlausibleCombobox` component (used in the filter menus and the `prop_key` picker), we're currently making too many API calls to fetch selectable options. Currently, we're calling the API for every keystroke. This PR fixes that - we will only make a request once 200ms has passed from the last keystroke (reusing the `useDebouncedCallback` custom hook introduced in #4318).

Also in this PR:

* Alter the `useDebouncedEffect` custom hook to not set a delay for the initial function call. Currently the search in all breakdown modals is waiting 300ms before calling the API for the first time. This fixes that.
* Remove redundant doubled "search" state from the BreakdownModal component.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -53,7 +53,9 @@ export default function PlausibleCombobox(props) {
   }
 
   useDebouncedEffect(() => {
-    fetchOptions()
+    if (isOpen) {
+      fetchOptions()
+    }
   }, [search], 200)
 
   const fetchOptions = useCallback(() => {
--- a/assets/js/dashboard/components/combobox.js
+++ b/assets/js/dashboard/components/combobox.js
@@ -1,7 +1,7 @@
 import React, { Fragment, useState, useCallback, useEffect, useRef } from 'react'
 import { Transition } from '@headlessui/react'
 import { ChevronDownIcon } from '@heroicons/react/20/solid'
-import debounce from 'debounce-promise'
+import { useDebouncedEffect } from '../custom-hooks.js'
 import classNames from 'classnames'
 
 function Option({isHighlighted, onClick, onMouseEnter, text, id}) {
@@ -39,7 +39,7 @@ export default function PlausibleCombobox(props) {
   const [options, setOptions] = useState([])
   const [isLoading, setLoading] = useState(false)
   const [isOpen, setOpen] = useState(false)
-  const [input, setInput] = useState('')
+  const [search, setSearch] = useState('')
   const [highlightedIndex, setHighlightedIndex] = useState(0)
   const searchRef = useRef(null)
   const containerRef = useRef(null)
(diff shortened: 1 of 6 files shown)


assets/js/dashboard/components/combobox.js
```diff
@@ -1,7 +1,7 @@
 import React, { Fragment, useState, useCallback, useEffect, useRef } from 'react'
 import { Transition } from '@headlessui/react'
 import { ChevronDownIcon } from '@heroicons/react/20/solid'
-import debounce from 'debounce-promise'
```
ukutaht, line 4: Can we remove the `debounce-promise` dependency now as well?

assets/js/dashboard/components/combobox.js
```diff
@@ -48,10 +48,26 @@ export default function PlausibleCombobox(props) {
...
+  useDebouncedEffect(() => {
+    if (isOpen) {
+      fetchOptions()
+    }
+  }, [search], 200)
+
+  const fetchOptions = useCallback(() => {
+    setLoading(true)
+
+    props.fetchOptions(search).then((loadedOptions) => {
```
macobo, line 64: I propose we use async over promise chaining here. Example: 

```javascript
const fetchOptions = useCallback(async () => {
  setLoading(true)

  const loadedOptions = await props.fetchOptions(search)

  setLoading(false)
  setHighlightedIndex(0)
  setOptions(loadedOptions)
}, [search])
```

assets/js/dashboard/components/combobox.js
```diff
@@ -48,10 +48,26 @@ export default function PlausibleCombobox(props) {
...
-    visibleOptions.push({value: input, label: input, freeChoice: true})
+  if (props.freeChoice && search.length > 0 && options.every(option => option.value !== search)) {
+    visibleOptions.push({value: search, label: search, freeChoice: true})
   }
 
+  useDebouncedEffect(() => {
+    if (isOpen) {
+      fetchOptions()
+    }
+  }, [search], 200)
```
macobo, line 59: Why is this 200 in one case and 300 in another? 

Proposal: Let's unify and make it a exported const somewhere.

assets/js/dashboard/components/combobox.js
RobertJoonas: I would leave it for now, as it's currently used all in pretty much every dashboard report. True, promise chaining might look a bit messy when there's multiple promises involved, but a single `.then()` clause doesn't hurt readability that much, I think?

assets/js/dashboard/components/combobox.js
ukutaht: Good point, agreed. The 'standard' delay could even be baked in to the`useDebouncedEffect` and not be configurable.

assets/js/dashboard/components/combobox.js
RobertJoonas: https://github.com/plausible/analytics/pull/4331/commits/9d9d66dde9f5aa4be0b08129a242fcee44a48a23

RobertJoonas:
Closed in favor of https://github.com/plausible/analytics/pull/4349

PR #4332, opened 2024-07-10
fix(suggestions): Properly provide suggestions

macobo:
Previously, if a filter was applied then filter modal suggestions would _always_ apply the previous suggestion. Now when say applying a `page` filter suggestions we remove the previous `page` filter.

### DIFF
--- a/assets/js/dashboard/util/filters.js
+++ b/assets/js/dashboard/util/filters.js
@@ -160,9 +160,16 @@ export function fetchSuggestions(apiPath, query, input, additionalFilter) {
 
 function queryForSuggestions(query, additionalFilter) {
   let filters = query.filters
-  if (additionalFilter && additionalFilter[2].length > 0) {
-    filters = filters.concat([additionalFilter])
+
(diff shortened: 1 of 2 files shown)


assets/js/dashboard/util/filters.js
```diff
@@ -160,9 +160,16 @@ export function fetchSuggestions(apiPath, query, input, additionalFilter) {
 
 function queryForSuggestions(query, additionalFilter) {
   let filters = query.filters
-  if (additionalFilter && additionalFilter[2].length > 0) {
-    filters = filters.concat([additionalFilter])
+
+  if (additionalFilter) {
+    const [_operation, filterKey, clauses] = additionalFilter
+
+    filters = filters.filter(([_op, key, _clauses]) => key != filterKey)
```
apata, line 167: **nitpick**: If we follow the single function, single responsibility principle to the max, this line could instead be function `omitFiltersByKey(...)`, which I think would increase code comprehension.

apata (approved):

PR #4333, opened 2024-07-10
fix(main-graph): fix 12mo and other comparisons with blanks

macobo:
Some comparisons return __blank__ values. After recent time series fix, this blew up when trying to parse __blank__ as a date.

This fixes the issue and adds a test for a period with __blank__ values


RobertJoonas (approved):
LGTM 👍

PR #4337, opened 2024-07-11
fix(TimeSeries): Work around weird time zones

macobo:
Previously APIv2 output didn't start at a full hour for these time zones
and main graph was blank

The core issue is that ClickHouse `timeSlots` is not time-zone
aware and works off of unix epoch - meaning that in time zones which
have an offset of 5:45 the "hours" reported would start at minute :45.

The fix is kind of silly - we now divide each hour into 4 and handle
things that way.

Related basecamp issue: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7590936581


lib/plausible/stats/sql/expression.ex
```diff
@@ -55,29 +55,35 @@ defmodule Plausible.Stats.SQL.Expression do
...
+    # :TRICKY: ClickHouse timeSlots works off of unix epoch and is not
+    #   timezone-aware. This means that for e.g. Asia/Katmandu (GMT+5:45)
+    #   to work, we divide time into 15-minute buckets and later combine these
+    #   via toStartOfHour
+    q
+    |> join(:array, [s], time_slot in time_slots(query, 15 * 60), as: :time_slot)
+    |> select_merge_as([s, time_slot: time_slot], %{
+      key => fragment("toStartOfHour(?)", time_slot)
     })
   end
```
apata, line 74: **praise:** the idea to increase granularity on timeslots and then recombine is genius!

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -1133,6 +1133,35 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
                %{"dimensions" => ["2021-01-01"], "metrics" => [1]}
              ]
     end
+
+    test "timeseries with quarter-hour timezone", %{conn: conn, user: user} do
+      # GMT+05:45
+      site = insert(:site, timezone: "Asia/Katmandu", members: [user])
+
+      populate_stats(site, [
+        build(:pageview, timestamp: ~N[2021-01-02 05:00:00]),
```
apata, line 1142: **question**: Really glad to see this test! Would we be able to provide the timestamps in local time in the test setup? Something like `build(:pageview, timestamp: convert_to_utc("Asia/Katmandu", ~N[2021-01-02 10:00:00])`? That way, verifying that the assertion is correct doesn't require mental arithmetic.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -1133,6 +1133,35 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
...
+        build(:pageview, timestamp: ~N[2021-01-02 06:00:00]),
+        build(:pageview, timestamp: ~N[2021-01-02 06:15:00]),
+        build(:pageview, timestamp: ~N[2021-01-02 06:30:00])
+      ])
+
+      conn =
+        post(conn, "/api/v2/query", %{
+          "site_id" => site.domain,
+          "metrics" => ["visits"],
+          "date_range" => ["2020-12-02", "2021-01-02"],
```
apata, line 1155: **meganitpick**: With nothing in the test setup or assertions connecting to this early (2020 vs 2021) start of range, it's a bit of a distraction.

I suggest the range to be a single day ´"date_range" => ["2021-01-02", "2021-01-02"],´ and the test setup to contain pageviews within and around that single day. 

We'd then assert that only the pageviews within the selected range are included in the results. 

Alternatively, if that exclusivity is sufficiently covered already, just changing the date range to a single day would remove the distraction.

macobo: This was a typo, thanks for noticing!

apata (approved):
I tested locally and it certainly seems to solve the issue for timezones that contain minute offsets. Tested with Katmandu, Marquesas, Rangoon. I don't know enough about Ecto & Elixir yet to comment on the necessity of the refactor, so you may want to seek another review as well if you feel it necessary.

PR #4338, opened 2024-07-11
Remove Timex.shift in app

ruslandoga:
### Changes

This PR starts [removing Timex](https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7479197128) by replacing [Timex.shift](https://hexdocs.pm/timex/Timex.html#shift/2) with [Duration functions](https://hexdocs.pm/elixir/Duration.html) (the biggest difference is plural vs singular time unit names, e.g. `minutes: 5` in Timex vs `minute: 5` in stdlib). The tests are kept unchanged for now to ensure that there are no unforeseen differences.

### Next Steps

Once this change is merged and verified over the next few days, the subsequent PR will update Timex.today to Date.utc_today. Then Timex.now to (Naive)DateTime.utc_now, then Timex.format to Calendar.strftime, and so one, always one change at a time. Then finally the tests would be updated and Timex dependency would be removed.

### Tests

### Changelog

### Documentation

### Dark mode


ruslandoga:
CI fails with :dialyzer.run error: Old PLT file /home/runner/work/analytics/analytics/priv/plts/dialyzer.plt

ukutaht (approved):

ruslandoga:
Not sure if related, but bounce rate seems too high for this PRs preview: https://pr-4354.review.plausible.io/dummy.site/

zoldar:
Yes, it does not seem to be related. Also, the number is basically 2^32 - 1 - I suspect some missed edge case of UInt32 casting in CH logic.

ruslandoga:
Or maybe it's a [toUInt32](https://github.com/plausible/analytics/blob/f09db60207990e1fa081de7b020599bcc562c13c/lib/plausible/stats/sql/expression.ex#L259) applied to a negative number.

```sql
1232e8f2a92c :) select toUInt32(-1);
--    ┌─toUInt32(-1)─┐
-- 1. │   4294967295 │ -- 4.29 billion
--    └──────────────┘
```

sevab:
Hi folks, just curious about the reason for removing Timex.

zoldar:
> > sed edge case of UInt32 casting in CH logic.
> 
> Or maybe it's a `toUInt32` [applied to a negative number.](https://github.com/plausible/analytics/blob/f09db60207990e1fa081de7b020599bcc562c13c/lib/plausible/stats/sql/expression.ex#L259) Maybe something went wrong during session caching / inserting.

Yes, that's exactly it 👍

ruslandoga:
👋 @sevab 

The linked Basecamp card said:

> ruslandoga: I was reading release notes for Elixir 1.17 and it adds a new module called [Duration. ](https://hexdocs.pm/elixir/1.17.0-rc.1/Duration.html)With this addition Timex can be fully replaced with stdlib functions.

So the reason is to have fewer dependencies :)

PR #4334, opened 2024-07-10
Alleviate props passing: institute UserContext, QueryContext, SiteContext

apata:
### Changes

The objective of the PR is to alleviate props passing in the application and pave the way for more DRY code. 

* Institutes SiteContext & useSiteContext hook
* Institutes QueryContext & useQueryContext hook
* Institutes UserContext & useUserContext hook

These contexts are not yet 100% in the app (some props passing remains), but in the interest of easy review, the most obvious and easy-to-change places have been tackled.

### Tests
- [ ] Automated tests have been added
- [ ] This PR does not require tests
- [x] Can't add FE tests, yet

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #4340, opened 2024-07-11
Remove query and site props passing from everywhere except class components

apata:
### Changes

Followup to https://github.com/plausible/analytics/pull/4334, to further alleviate issues from props passing

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/comparison-input.js
+++ b/assets/js/dashboard/comparison-input.js
@@ -86,7 +86,9 @@ function ComparisonModeOption({ label, value, isCurrentlySelected, updateMode, s
   )
 }
 
-function MatchDayOfWeekInput({ history, query, site }) {
+function MatchDayOfWeekInput({ history }) {
+  const site = useSiteContext()
+  const { query } = useQueryContext()
   const click = (matchDayOfWeek) => {
     storage.setItem(`comparison_match_day_of_week__${site.domain}`, matchDayOfWeek.toString())
     navigateToQuery(history, query, { match_day_of_week: matchDayOfWeek.toString() })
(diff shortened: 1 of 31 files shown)


assets/js/dashboard/comparison-input.js
```diff
@@ -86,7 +86,9 @@ function ComparisonModeOption({ label, value, isCurrentlySelected, updateMode, s
   )
 }
 
-function MatchDayOfWeekInput({ history, query, site }) {
+function MatchDayOfWeekInput({ history }) {
+  const site = useSiteContext()
+  const { query } = useQueryContext()
```
macobo, line 91: Would something like `useDashboardContext()` make sense which would encapsulate both site AND query? :thinking:

assets/js/dashboard/stats/imported-query-unsupported-warning.js
```diff
@@ -1,8 +1,10 @@
 import React from "react";
 import { ExclamationCircleIcon } from '@heroicons/react/24/outline'
 import FadeIn from "../fade-in";
+import { useQueryContext } from "../query-context";
```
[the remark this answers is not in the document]
apata: Thanks! Agreed, but I can't get prettier and eslint to work in the repository / in my IDE well. Not sure yet what the issue is with those, will come back to that if the eslint updates that I need to do for Typescript anyways don't fix it.

assets/js/dashboard/comparison-input.js
apata: Thanks!

I considered it initially, but my hunch is that many of these components don't really need either. 

I think the way forward is to actually remove some of these superfluous dependencies, and provide components with specific handlers for what they're supposed to do: in this example handleCompareExactMatch, handleCompareDayOfTheWeek. 

Such handlers would be declared higher up in the React tree and themselves dependent on a shared API component, which is dependent on the site.domain.

macobo (approved):
Didn't test it manually but the approach and code looks great! :ship:

PR #4357, opened 2024-07-16
Remove Timex.today

ruslandoga:
Continues #4338

### Changes

This PR continues [the removal of Timex](https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7479197128) by replacing [`Timex.today/0`](https://hexdocs.pm/timex/Timex.html#today/0) with `Date.utc_today/0` which was already being [delegated to.](https://github.com/bitwalker/timex/blob/c45b9a734074ac2a56355ec52ebf931b932223b7/lib/timex.ex#L47-L51) Additionally, it replaces a single instance of [`Timex.today/1`](https://hexdocs.pm/timex/Timex.html#today/1) with [similar `DateTime` functions.](https://github.com/bitwalker/timex/blob/c45b9a734074ac2a56355ec52ebf931b932223b7/lib/timex.ex#L53-L78) I'm using "similar" and not "equivalent" because now we raise an error on an invalid timezone instead of returning an error tuple, which previously could cause match errors downstream thus hiding the actual issue.

### Tests
- [x] The tests are still kept unchanged for now to ensure that there are no unforeseen differences.

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/mix/tasks/download_country_database.ex
+++ b/lib/mix/tasks/download_country_database.ex
@@ -13,7 +13,7 @@ defmodule Mix.Tasks.DownloadCountryDatabase do
   def run(_) do
     Application.ensure_all_started(:httpoison)
     Application.ensure_all_started(:timex)
-    this_month = Timex.today()
+    this_month = Date.utc_today()
     last_month = Date.shift(this_month, month: -1)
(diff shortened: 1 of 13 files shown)


aerosol (approved):
😍

zoldar (approved):
🔥

PR #4358, opened 2024-07-16
Fix present_index for daily, weekly and monthly intervals

ukutaht:
### Changes

In the main graph, the 'current' interval should be shown with a dotted line instead of a solid line. This is accomplished with the `present_index` variable which broke recently.

### Tests

### Changelog
- Since this behaviour was broken and fixed during a single CE release cycle, no changelog is needed.

### Documentation

### Dark mode


zoldar (approved):
I presume the point here is that time dimension value formats should align between here and https://github.com/plausible/analytics/blob/7015075e12d309fbf995c4678039a92caee681d6/lib/plausible/stats/time.ex#L58-L61 - which they do now 👍

PR #4361, opened 2024-07-16
Install query

apata:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/modals/breakdown-modal.js
```diff
@@ -84,24 +85,57 @@ export default function BreakdownModal({
...
+    getNextPageParam: (lastPageResults, _, lastPageIndex) => lastPageResults.length === LIMIT ? lastPageIndex + 1 : null,
+    initialPageParam: 1,
+    queryFn,
+  })
 
-  useMountedEffect(() => { debouncedFetchData() }, [search])
+  const queryClient = useQueryClient()
 
-  useMountedEffect(() => { fetchNextPage() }, [page])
+  useEffect(() => {
```
RobertJoonas, line 126: This `useEffect` here makes sure that we always cache only a single page of results.

If we imagine the following scenario:

1) user opens the modal
2) clicks load more 2 times
3) closes the modal
4) reopens the modal

Then,

* **without this `useEffect`** - after step 4) the user would instantly see the cached 3 pages of results. But since that data is stale, we'll still want to refresh all of it. That will take 3 API calls.
* **with this `useEffect`** - after step 4) we only get the first page of results from cache and make a single API call to refresh it.

So basically, it simply reduces load on the server.

assets/js/dashboard/stats/modals/breakdown-modal.js
```diff
@@ -84,24 +85,57 @@ export default function BreakdownModal({
...
-  const [initialLoading, setInitialLoading] = useState(true)
-  const [loading, setLoading] = useState(true)
   const [search, setSearch] = useState('')
-  const [results, setResults] = useState([])
-  const [page, setPage] = useState(1)
-  const [moreResultsAvailable, setMoreResultsAvailable] = useState(false)
-  const searchBoxRef = useRef(null)
 
-  useEffect(() => { fetchData() }, [])
+  const queryFn = async ({ pageParam, queryKey }) => {
```
ukutaht, line 95: Given that the name `query` is already very overloaded in this codebase, I wonder if it would be possible to encapsulate the `react-query` stuff so that we can call it by a different name. Maybe a custom hook like `useApiClient`?

ukutaht (approved):
Looks good! I really like the UX improvement from caching. One comment about naming below.

assets/js/dashboard/stats/modals/breakdown-modal.js
RobertJoonas: Agreed. I've created a new `useAPIClient` hook that also handles pagination. https://github.com/plausible/analytics/pull/4361/commits/0c53aec263af195b0ed77ff31100c579e8319fff.

Not every report needs pagination, so if we're planning to migrate other dashboard reports to use React Query, I was thinking we could create a new api hook, or make pagination optional somehow.

The Google Search Terms modal could nicely use the `useAPIClient` though, with built in pagination logic.

assets/js/dashboard/hooks/api-client.js
```diff
@@ -0,0 +1,79 @@
...
+
+// * `getRequestParams` - a function that takes the `key` prop as an argument,
+//   and returns `[query, params]` which will be used by `queryFn` that actually
+//   calls the API.
+
+// ### Optional props
+
+// * `afterFetchData` - a function to call after data has been fetched.
+
+// * `afterFetchNextPage` - a function to call after the next page has been fetched.
```
apata, line 27: **Ultranitpick, non-blocking:** if you use [JSDoc code comment format](https://jsdoc.app/about-getting-started), the IDE shows it in a nicer way when you hover the function where it is used. 
```
/**
 * A wrapper for the React Query library. Constructs the necessary options
 * (including pagination config) to pass into the `useInfiniteQuery` hook.
 *
 * ### Required props
 *
 * @param {Array} key - The key under which the global "query" instance will live. Should
 *   be passed as a list of two elements - `[endpoint, { query }]`. The object
 *   can also contain additional values (such as `search`) to be used by:
 *   1) React Query, to determine the uniqueness of the query instance
 *   2) the `getRequestParams` function to build the request params.
 *
 * @param {Function} getRequestParams - A function that takes the `key` prop as an argument,
 *   and returns `[query, params]` which will be used by `queryFn` that actually
 *   calls the API.
 *
 * ### Optional props
 *
 * @param {Function} [afterFetchData] - A function to call after data has been fetched.
 *
 * @param {Function} [afterFetchNextPage] - A function to call after the next page has been fetched.
 */
```

RobertJoonas: Makes sense, thanks! https://github.com/plausible/analytics/pull/4361/commits/68bf8826e2072e7e6861181cb8fd0c98c0de675e

PR #4378, opened 2024-07-22
Implement search and pagination in Google Keywords modal

RobertJoonas:
### Changes

This PR rebuilds the Google Keywords modal with search and pagintion functionality. It also changes the API structure to align better with other endpoints.

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/modals/google-keywords.js
```diff
@@ -1,118 +1,193 @@
...
+    status
+  } = useAPIClient({
+    key: [endpoint, {query, search}],
+    getRequestParams: (key) => {
+      const [_endpoint, {query, search}] = key
+      const params = { detailed: true }
+
+      return [query, search === '' ? params : {...params, search}]
+    },
+    initialPageParam: 0
```
ukutaht, line 46: Why are we starting at page 0 not 1?

assets/js/dashboard/stats/modals/google-keywords.js
```diff
@@ -1,118 +1,193 @@
-import React from "react";
-import { Link, withRouter } from 'react-router-dom'
+import React, { useState, useEffect, useRef } from "react";
```
ukutaht, line 1: Could this be refactored to use `BreakdownModal` or is it too different from the rest of the details views?

lib/plausible/google/http.ex
```diff
@@ -39,12 +39,15 @@ defmodule Plausible.Google.HTTP do
...
-  def list_stats(access_token, property, date_range, limit, search_console_filters) do
+  def list_stats(access_token, property, date_range, pagination, search_console_filters) do
+    {limit, page} = pagination
+
     params = %{
       startDate: Date.to_iso8601(date_range.first),
       endDate: Date.to_iso8601(date_range.last),
       dimensions: ["query"],
       rowLimit: limit,
+      startRow: page * limit,
```
ukutaht, line 50: Referring to my earlier comment about `initialPageParam` being 0 - I wonder if it would simplify things to use here `(page - 1) * limit` so the frontend can always start from page 1 (not 0).

ukutaht (approved):

assets/js/dashboard/stats/modals/google-keywords.js
RobertJoonas: This comes from the [Google Search Console API reference](https://developers.google.com/webmaster-tools/v1/searchanalytics/query):

image

assets/js/dashboard/stats/modals/google-keywords.js
RobertJoonas: It could. `GoogleKeywordsModal` is pretty much a copy of `BreakdownModal`, with small differences (pagination, error handling, applying search). Would be doable to unify those differences, but...

I thought it might be a good idea to keep it separate as we might want to build new functionalities into `BreakdownModal`, but not necessarily into `GoogleKeywordsModal`, and perhaps even vice versa.

One example: clicking on the column headers to order by a metric - GSC API doesn't support custom sorting unfortunately:
image.

lib/plausible/google/http.ex
RobertJoonas: hmm, yes it would mean that the pagination in the `useAPIClient` hook could just work with defaults. But it's only a matter of passing a different variable in the frontend, so it doesn't complicate things too much IMHO. WDYT?

assets/js/dashboard/stats/modals/google-keywords.js
ukutaht: Makes sense

PR #4381, opened 2024-07-23
Fix email report sentry errors

ukutaht:
### Changes

Fixes the following sentry crashes

https://sentry.plausible.io/organizations/plausible/issues/53/?project=2&query=is%3Aunresolved&referrer=issue-stream&stream_index=7

https://sentry.plausible.io/organizations/plausible/issues/52/?query=is%3Aunresolved&referrer=issue-stream&stream_index=23

### DIFF
--- a/lib/plausible_web/controllers/unsubscribe_controller.ex
+++ b/lib/plausible_web/controllers/unsubscribe_controller.ex
@@ -22,6 +22,10 @@ defmodule PlausibleWeb.UnsubscribeController do
     )
   end
 
+  def weekly_report(conn, _) do
+    render_error(conn, 400)
+  end
+
   def monthly_report(conn, %{"website" => website, "email" => email}) do
     site = Repo.get_by(Plausible.Site, domain: website)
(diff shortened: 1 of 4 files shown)


lib/plausible_web/controllers/unsubscribe_controller.ex
```diff
@@ -22,6 +22,10 @@ defmodule PlausibleWeb.UnsubscribeController do
     )
   end
 
+  def weekly_report(conn, _) do
+    render_error(conn, 400)
```
zoldar, line 26: This is weeeird, `Phoenix.ActionClauseError` should already be translated to HTTP 400 when `debug_errors` is set to `false` (anywhere outside dev env) 🤔

zoldar: Ahh, it indeed does return HTTP 400 screen but it still throws to sentry underneath.

zoldar (approved):

PR #4384, opened 2024-07-23
Install newest @tanstack/react-router

apata:
### Changes

This PR replaces `react-router-dom` with latest`@tanstack/react-router`. The reasoning is that we're using `@tanstack/query` to manage API response state in the application, and both packages depend on the same core logic package. We can expect these two to be complementing each other. 

- deprecates withRouter HOC
- replaces `PlausibleSearchParams` with a `stringifySearch` and `parseSearch` method, [performed automatically on location change by the new router](https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization)
  - NB! components are operating with and changing the parsed object
  - NB! setting search param value to `false` used to be the way to remove the search param, but I wasn't able to keep this behavior, now setting the value to `null` is the way to do it
- explicitly declares all app routes
  - makes navigating to each route use the route definition
- refactors navigation links and buttons to new navigate API

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/with-imported-switch.js
+++ b/assets/js/dashboard/stats/graph/with-imported-switch.js
@@ -14,22 +14,20 @@ function LinkOrDiv({ isLink, to, search, children }) {
 
 export default function WithImportedSwitch({ info }) {
   const { query } = useQueryContext();
-
-  if (!info?.visible) {
-    return null;
-  }
+  const {with_imported_switch, _imports_exist, _includes_imported} = info;
+  const switchedOn = query.with_imported;
   
-  const { togglable, tooltip_msg } = info
-  const enabled = togglable && query.with_imported
+  // const { togglable, tooltip_msg } = info
+  // const enabled = togglable && query.with_imported
   
   const iconClass = classNames("mt-0.5", {
-    "dark:text-gray-300 text-gray-700": enabled,
-    "dark:text-gray-500 text-gray-400": !enabled,
+    "dark:text-gray-300 text-gray-700": with_imported_switch.togglable,
+    "dark:text-gray-500 text-gray-400": !with_imported_switch.togglable,
   })
 
     return (
-      <div tooltip={tooltip_msg} className="w-4 h-4 mx-2">
-        <LinkOrDiv isLink={togglable} search={(search) => ({...search, with_imported: !enabled})}>
+      <div tooltip={with_imported_switch.tooltip_msg} className="w-4 h-4 mx-2">
+        <LinkOrDiv isLink={with_imported_switch.togglable} search={(search) => ({...search, with_imported: !switchedOn})}>
           <BarsArrowUpIcon className={iconClass} />
         </LinkOrDiv>
       </div>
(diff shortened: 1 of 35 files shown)


apata:
How much should each component with a navigation action know about the `query`? 

At the moment, their state is governed by the `query`, but they don't modify the `query` directly, they modify it by changing search params. I can see why, that's just what router utilities expect. 

However, because search params =/= `query`, each component needs to be provided with a remapper of what search params to set to affect the query as expected. This is the questionable bit and I'm wondering if this remapping can be avoided.

Just to understand what each components responsibility _should_ be, I created a small flow diagram. 

```mermaid
---
title: Dashboard state and navigation
---
flowchart TB
    subgraph ENTRY
        direction TB
        BrowserArrives([Browser navigates to page]) -->
        ServerRenders[Server responds with HTML with site data encoded] -->
        SiteIsParsed[Site data is decoded from HTML] -->
        ReactAppMounts[React app mounts with the parsed site]
    end
    subgraph INITIAL
        direction LR
        RouterMounts([Initial browser path and search params are parsed]) --> 
        QueryContextMounts[Parsed search params and saved or hardcoded defaults are made available as `query` within the app]
    end
    subgraph ANY LIST LIFECYCLE
        direction TB
        AnyListMounts([Based on the browser path and `query`, a list component makes an API call and renders list items]) --> 
        AnyNavigationLinkMounts[A list item is able to set the browser path and new search params, with the new URL visible on hover] -- link clicked --> 
        RouterReloads[New browser path and search params are parsed] -->
        QueryContextReloads[Query context recalculates query based on the new search value]
        --> AnyListMounts
    end

    ENTRY --> INITIAL
    INITIAL --> AnyListMounts
```

assets/js/dashboard/stats/pages/index.js
```diff
@@ -42,7 +43,7 @@ function EntryPages({ afterFetchData }) {
       getFilterFor={getFilterFor}
       keyLabel="Entry page"
       metrics={chooseMetrics()}
-      detailsLink={url.sitePath('entry-pages')}
+			detailsLinkProps={{to: entryPagesRoute.to, search: (search) => search}}
```
ukutaht, line 46: Something happened with whitespace here?

ukutaht (approved):

ukutaht:
> How much should each component with a navigation action know about the query?
> This is the questionable bit and I'm wondering if this remapping can be avoided.

Do you have a proposal for how it could be avoided?

In general I don't think it's a problem for components to know about `query`, especially now that we have the context for it.

assets/js/dashboard/stats/graph/with-imported-switch.js
```diff
@@ -1,38 +1,35 @@
...
+  })
 
     return (
-      <div tooltip={tooltip_msg} className="w-4 h-4 mx-2">
-        <LinkOrDiv isLink={togglable} target={target}>
+      <div tooltip={with_imported_switch.tooltip_msg} className="w-4 h-4 mx-2">
+        <LinkOrDiv isLink={with_imported_switch.togglable} search={(search) => ({...search, with_imported: !switchedOn})}>
           <BarsArrowUpIcon className={iconClass} />
         </LinkOrDiv>
       </div>
```
apata, line 33: todo: needs fix

PR #4391, opened 2024-07-25
Generate goal filter SQL from postgres goals

ukutaht:
With this PR we'll be generating the SQL for goal filters such that:
* When goals are present in either filters or requested dimensions, they are always preloaded to the query struct
* SQL for goal filters is always generated from goals that have been looked up from PostgreSQL
  * This will allow us to 'decouple' the goal name from the actual filters that get applied at the SQL-level. We'll be able to look up goals by the upcoming `display_name` field going forward
* Removes the `matches` filter type for goals. A follow-up PR will implement `contains`

These changes are motivated by:
* The need to be able to rename goals and look them up from postgres by arbitrary `display_name` field cc @aerosol 
* The need to be able to filter goals with the `contains` filter type on `goal.display_name` cc @RobertJoonas 

Also tagging @macobo because I think you should review this when you're back.


lib/plausible/stats/breakdown.ex
```diff
@@ -24,7 +23,6 @@ defmodule Plausible.Stats.Breakdown do
         order_by: infer_order_by(transformed_metrics, dimension),
         dimensions: transform_dimensions(dimension),
         filters: query.filters ++ dimension_filters(dimension),
-        preloaded_goals: QueryParser.preload_goals_if_needed(site, query.filters, [dimension]),
```
ukutaht, line 27: This is now done in `LegacyQueryBuilder`

lib/plausible/stats/imported/base.ex
```diff
@@ -178,12 +167,26 @@ defmodule Plausible.Stats.Imported.Base do
     ["imported_pages", "imported_custom_events"]
   end
 
-  defp do_decide_tables(%Query{filters: filters, dimensions: ["event:goal"]}) do
+  defp do_decide_tables(%Query{filters: filters, dimensions: ["event:goal"]} = query) do
```
ukutaht, line 170: Would like @RobertJoonas to review these changes... I *think* I got it mostly right but would appreciate extra 👀

lib/plausible/stats/imported/base.ex
```diff
@@ -193,33 +196,82 @@ defmodule Plausible.Stats.Imported.Base do
...
+  import Plausible.Stats.Filters.Utils, only: [page_regex: 1]
+
+  defp apply_filter(q, %Query{filters: filters} = query) do
+    Enum.reduce(filters, q, fn [operation, filter_key, clauses] = filter, q ->
       db_field = Filters.without_prefix(filter_key)
-      mapped_db_field = Map.get(@db_field_mappings, db_field, db_field)
-      condition = SQL.WhereBuilder.build_condition(mapped_db_field, filter)
 
-      where(q, ^condition)
+      if db_field == :goal do
```
ukutaht, line 242: This is kind of ugly at the moment. Very similar to how SQL is generated for goals in `SQL.WhereBuilder` but enough difference that it's hard to consolidate this logic.

lib/plausible/stats/query.ex
```diff
@@ -109,6 +109,15 @@ defmodule Plausible.Stats.Query do
     end)
   end
 
+  def preloaded_goals_for(query, operation, goal_name) when operation in [:is, :contains] do
+    Enum.filter(query.preloaded_goals, fn goal ->
+      case operation do
+        :is ->
```
ukutaht, line 115: This is where goals in the filters are matched with goals from Postgres. We can make this work with `:contains` as well as a follow-up.

test/plausible/stats/query_parser_test.exs
```diff
@@ -345,31 +345,33 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
-        ],
-        dimensions: [],
-        order_by: nil,
-        timezone: site.timezone,
-        include: %{imports: false, time_labels: false},
-        preloaded_goals: [{:page, "/thank-you"}, {:event, "Signup"}]
-      })
-    end
+  describe "event:goal filter validation" do
+    # test "valid filters", %{site: site} do
```
ukutaht, line 351: TODO: need to fix these tests

lib/plausible/stats/query.ex
ukutaht: And when goal names become editable, the `Plausible.Goal.display_name` can we replaced with a simple `goal.display_name`

test/plausible/stats/query_parser_test.exs
ukutaht: Fixed in https://github.com/plausible/analytics/pull/4391/commits/f31001d13ea65549a652dc27329c49143104a74e

lib/plausible/stats/breakdown.ex
RobertJoonas: Is the current plan to only focus on API v1?

lib/plausible/stats/imported/base.ex
```diff
@@ -178,12 +167,26 @@ defmodule Plausible.Stats.Imported.Base do
...
   end
 
-  defp do_decide_tables(%Query{filters: filters, dimensions: ["event:goal"]}) do
+  defp do_decide_tables(%Query{filters: filters, dimensions: ["event:goal"]} = query) do
     filter_props = Enum.map(filters, &Enum.at(&1, 1))
 
-    any_event_name_filters? = "event:name" in filter_props
-    any_page_filters? = "event:page" in filter_props
-    any_other_filters? = Enum.any?(filter_props, &(&1 not in ["event:page", "event:name"]))
+    filter_goals =
```
RobertJoonas, line 173: I believe this is missing an `Enum.filter` clause to first get the list of "event:goal" filters?

I'm also seeing some repetition - we're getting the `filter_goals` list in two functions. Perhaps we could extract that into a private function in this module?

lib/plausible/stats/query.ex
RobertJoonas: Since goal filters are applied very differently, compared to other filters, perhaps we could encapsulate this logic into a single place that deals with both:

* filtering the preloaded list of goals
* generating the actual SQL dynamic where clause

I've opened a PR targeting this branch with a proposal: https://github.com/plausible/analytics/pull/4399

lib/plausible/stats/imported/base.ex
RobertJoonas: https://github.com/plausible/analytics/pull/4399 makes this cleaner

lib/plausible/stats/breakdown.ex
ukutaht: No. The new apiv2 stuff already preloads goals into the `%Query{}` struct but the old endpoints (dashboard, apiv1) didn't before this PR. This PR makes is consistent so goals are always preloaded.

lib/plausible/stats/query.ex
ukutaht: I like it!

lib/plausible/stats/imported/base.ex
ukutaht: Good suggestion 👍 done in https://github.com/plausible/analytics/pull/4391/commits/702cbdbb14361a5b3482333c95667b2087669b90

RobertJoonas (commented):
Huge fan of this PR. I feel like this simple move solves so many problems with goal filters. Nicely done! 👏

Not approving yet though, as I believe some things are yet to be addressed here:

1) I've found a bug on staging where filter suggestions return a 500 with imported data
2) Some tests should be uncommented

lib/plausible/stats/legacy/legacy_query_builder.ex
```diff
@@ -25,6 +26,17 @@ defmodule Plausible.Stats.Legacy.QueryBuilder do
     query
   end
 
+  defp put_preloaded_goals(query, site) do
```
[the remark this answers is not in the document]
ukutaht: I thought about this as well and made a conscious decision to add some duplication because `QueryParser` and `LegacyQueryBuilder` have slightly different ways of going about building the struct. `QueryParser` doesn't build the `%Query{}` struct right away so all the helper functions just return data. But `LegacyQueryParser` helper functions operate on an already-built `%Query{}` struct by modifying it.

I prefer the way it's done in `QueryParser` and would like to keep it the way it is. We could refactor `LegacyQueryBuilder` to maybe use a similar flow as `QueryParser` but it doesn't make much sense to me since we're planning to remove the `LegacyQueryBuilder` stuff in the near future as well.

test/plausible_web/controllers/api/stats_controller/pages_test.exs
```diff
@@ -877,35 +877,36 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
...
+    #              "total_visitors" => 2
+    #            },
+    #            %{
+    #              "visitors" => 1,
+    #              "name" => "/blog",
+    #              "conversion_rate" => 100.0,
+    #              "total_visitors" => 1
+    #            }
+    #          ]
+    # end
```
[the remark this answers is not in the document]
ukutaht: Thanks! Done in https://github.com/plausible/analytics/pull/4391/commits/05266550b11fa458273f64b001f52d31f7f5a829

test/plausible_web/controllers/api/external_stats_controller/aggregate_test.exs
```diff
@@ -710,31 +710,32 @@ defmodule PlausibleWeb.Api.ExternalStatsController.AggregateTest do
...
+    #         "filters" => "event:page==/blog/2;event:goal==Visit /blog/**",
+    #         "with_imported" => "true"
+    #       })
+
+    #     assert json_response(conn, 200)["results"] == %{
+    #              "visitors" => %{"value" => 2},
+    #              "events" => %{"value" => 2},
+    #              "conversion_rate" => %{"value" => 100.0}
+    #            }
+    #   end
```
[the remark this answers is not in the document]
ukutaht: Thanks! Done in https://github.com/plausible/analytics/pull/4391/commits/05266550b11fa458273f64b001f52d31f7f5a829

ukutaht:
@RobertJoonas Thanks for noticing the 500 error! Fixed in https://github.com/plausible/analytics/pull/4391/commits/e05ef3936559257648081e8a1eb8ceb991901744

lib/plausible/stats/legacy/legacy_query_builder.ex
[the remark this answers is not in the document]
RobertJoonas: Makes sense!

RobertJoonas (approved):

aerosol:
Will take me a while to understand what happened here, but I've been able to edit a goal's display name and it just worked. So thanks for sparing me that bit. <3

PR #4398, opened 2024-07-27
Update `ua_inspector` fork to sanitize multi-part prereleases properly

zoldar:
### Changes

This PR bumps `ua_inspector` fork to a version with a patch for a corner case of versions with multiple prerelease parts, where each part has to be sanitized separately. Example failing user agent:

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 13_2_1) AppleWebKit/537.3666 (KHTML, like Gecko) Chrome/110.0.0.0.0 Safari/537.3666
```

The version that is currently causing failure: `110.0.0.0.0`.

The fix: https://github.com/plausible/ua_inspector/commit/cb8b41ae1ff2347c1e34b63aece46d4024abb9e9

It's quite possible that we might switch back to upstream soon once [upcoming](https://github.com/elixir-inspector/ua_inspector/commit/19d59201a7d699449e5e82f55716f6e797b5c55c) [changes](https://github.com/elixir-inspector/ua_inspector/commit/ce4cec1e3e786480706202a30c1d7a95eef4ea7a) get implemented and released. Before that though, we can at least switch to that temporary band-aid to avoid dropping events due to invalid version altogether.


RobertJoonas (approved):

apata:
From my side as well, I checked that it works using `mix send_pageview`. May I just ask if all test cases have been added in the PR? https://github.com/plausible/ua_inspector/commit/cb8b41ae1ff2347c1e34b63aece46d4024abb9e9

zoldar:
@apata good point, I'll add regression tests for event ingestion on our end 👍

apata (approved):

PR #4401, opened 2024-07-29
Fix flaky auth rate limit tests

zoldar:
### Changes

This PR adds a safeguard against flaky auth rate limiting due to fixed time rate limit buckets.

### Tests

### DIFF
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -259,7 +259,7 @@ defmodule PlausibleWeb.AuthController do
         )
 
       {:rate_limit, _} ->
-        maybe_log_failed_login_attempts("too many logging attempts for #{email}")
+        maybe_log_failed_login_attempts("too many login attempts for #{email}")
 
         render_error(
           conn,
@@ -514,7 +514,7 @@ defmodule PlausibleWeb.AuthController do
           {:ok, user}
         else
           {:rate_limit, _} ->
-            maybe_log_failed_login_attempts("too many logging attempts for #{user.email}")
+            maybe_log_failed_login_attempts("too many login attempts for #{user.email}")
 
             conn
             |> TwoFactor.Session.clear_2fa_user()
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -2,7 +2,7 @@ defmodule PlausibleWeb.AuthController do
   use PlausibleWeb, :controller
   use Plausible.Repo
 
-  alias Plausible.{Auth, RateLimit}
+  alias Plausible.Auth
   alias Plausible.Billing.Quota
   alias PlausibleWeb.TwoFactor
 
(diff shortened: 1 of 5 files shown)


test/support/conn_case.ex
```diff
@@ -37,6 +37,23 @@ defmodule PlausibleWeb.ConnCase do
...
+    # does time-based bucketing makes logic simpler
+    # at the cost of tests running into flakiness
+    # if the fixed time bucket changes right in the
+    # middle of test run, effectively resetting the
+    # rate limit counter.
+    if tags[:auth_rate_limit] do
+      now = System.system_time(:millisecond)
+      interval = 60_000
+
+      if rem(now, interval) != rem(now + 500, interval) do
```
RobertJoonas, line 52: I don't think this is working as it should here.

An inequality check will always return `true` here so we'd be sleeping 500ms before every `:auth_rate_limit tagged` test.

I think we should be checking whether `rem(now, interval)` **is greater than** `rem(now + 500, interval)` instead?

zoldar: https://github.com/plausible/analytics/pull/4401/commits/02666a9688842217499af1b53dfa289f04906519

test/support/conn_case.ex
```diff
@@ -37,6 +37,23 @@ defmodule PlausibleWeb.ConnCase do
...
+    # test result when specifically testing for
+    # auth rate limiting. The way `Plausible.RateLimit`
+    # does time-based bucketing makes logic simpler
+    # at the cost of tests running into flakiness
+    # if the fixed time bucket changes right in the
+    # middle of test run, effectively resetting the
+    # rate limit counter.
+    if tags[:auth_rate_limit] do
+      now = System.system_time(:millisecond)
+      interval = 60_000
```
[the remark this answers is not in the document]
zoldar: Done with additional clean up of rate limiting logic in auth context in https://github.com/plausible/analytics/pull/4401/commits/b657f594fcfeff17409b979e80b6c229abec7f5c and https://github.com/plausible/analytics/pull/4401/commits/02666a9688842217499af1b53dfa289f04906519

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -258,8 +258,8 @@ defmodule PlausibleWeb.AuthController do
           layout: {PlausibleWeb.LayoutView, "focus.html"}
         )
 
-      {:rate_limit, _} ->
-        maybe_log_failed_login_attempts("too many logging attempts for #{email}")
+      {:error, {:rate_limit, _}} ->
+        maybe_log_failed_login_attempts("too many login attempts for #{email}")
```
aerosol, line 262: :ok_hand:

test/support/conn_case.ex
```diff
@@ -37,6 +37,23 @@ defmodule PlausibleWeb.ConnCase do
...
 
+    # Tag intended for reducing risk of flaky
+    # test result when specifically testing for
+    # auth rate limiting. The way `Plausible.RateLimit`
+    # does time-based bucketing makes logic simpler
+    # at the cost of tests running into flakiness
+    # if the fixed time bucket changes right in the
+    # middle of test run, effectively resetting the
+    # rate limit counter.
+    if tags[:auth_rate_limit] do
```
aerosol, line 48: any reason not to use `Plausible.TestUtils.eventually/3` instead of indirectly slowing it down via test tags?

aerosol (approved):

test/support/conn_case.ex
zoldar: Indeed, `eventually` can be used for that by retrying the whole batch of requests in case of failure with a bigger than standard retry timeout. Refactored in https://github.com/plausible/analytics/pull/4401/commits/ec0221a070c8aef976fcea67bfa5250e8de9f988. Thanks ❤️

PR #4405, opened 2024-07-30
Improve how long URLs are displayed in dashboard

ukutaht:
### Changes

Improves readability for long URLs:
1. In modals, do not truncate URLs (screenshots below)
2. When displaying filters, show the full filter value in tooltip

Before:

Screenshot 2024-07-30 at 16 50 52

After:

Screenshot 2024-07-30 at 16 50 44

### Tests

### Changelog
- [x] A small change, no changelog needed

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/filters.js
+++ b/assets/js/dashboard/filters.js
@@ -10,11 +10,9 @@ import {
   cleanLabels,
   FILTER_MODAL_TO_FILTER_GROUP,
   formatFilterGroup,
-  formattedFilters,
   EVENT_PROPS_PREFIX,
-  getPropertyKeyFromFilterKey,
-  getLabel,
-  FILTER_OPERATIONS_DISPLAY_NAMES
+  plainFilterText,
+  styledFilterText
 } from "./util/filters";
 import { useQueryContext } from './query-context';
 import { useSiteContext } from './site-context';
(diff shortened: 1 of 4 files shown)


assets/js/dashboard/filters.js
```diff
@@ -40,7 +40,20 @@ function clearAllFilters(history, query) {
   );
 }
 
-function filterText(query, [operation, filterKey, clauses]) {
+function plainFilterText(query, [operation, filterKey, clauses]) {
```
ukutaht, line 43: This is very similar to the `styledFilterText` function, just without the JSX bold styling. Didn't find a good way to remove this duplication.

RobertJoonas: Yeah. That's tricky indeed. We could create another function that handles the logic of parsing the filter into "text parts" and then these parts could be used to generate either a plain string or JSX. I'm also fine with leaving it like that.

On another note: What do you think about using the same `plainFilterText` function to render a tooltip in BreakdownModals and ListReports, when list items are hovered? The tooltip could say something like: "Filter by: page is /long/247/0x0x0x0x........". Then it would make it possible in ListReports to see the full page path as well (for paths that don't fit on one line).

RobertJoonas (approved):
Left one comment inline, otherwise looks good to me 👍

assets/js/dashboard/filters.js
ukutaht: > What do you think about using the same plainFilterText function to render a tooltip in BreakdownModals and ListReports, when list items are hovered?

Love this suggestion. Done in https://github.com/plausible/analytics/pull/4405/commits/49a0bda8015e91b8bfddf54df52d817992c06fa9

PR #4408, opened 2024-07-31
Update react-router to version 6

apata:
### Changes

Redo of https://github.com/plausible/analytics/pull/4384.

It's almost exactly the same as before, except link and navigation search parameters aren't parsed and stringified at the router level, the transforms are applied in custom components. 

* deprecates withRouter HOC
* explicitly declares all app routes
* makes navigating to each route use the route definition
* replaces PlausibleSearchParams with a stringifySearch and parseSearch method, which are applied in the wrapped Link and navigate components AppNavigationLink and useAppNavigate and in QueryContextProvider
  * NB! components are operating with and changing the parsed object
  * NB! setting search param value to false used to be the way to remove the search param, but I wasn't able to keep this behavior, now setting the value to null is the way to do it
* refactors navigation links and buttons to new navigate API

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard.js
+++ b/assets/js/dashboard.js
@@ -2,7 +2,8 @@ import React from 'react';
 import { createRoot } from 'react-dom/client';
 import 'url-search-params-polyfill';
 
-import Router from './dashboard/router'
+import { RouterProvider } from 'react-router-dom';
+import { createAppRouter } from './dashboard/router'
 import ErrorBoundary from './dashboard/error-boundary'
 import * as api from './dashboard/api'
 import * as timer from './dashboard/util/realtime-update-timer'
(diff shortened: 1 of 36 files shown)


apata:
This time, it's tested to work with sites containing a forward slash: https://pr-4408.review.plausible.io/arturplausiblestaging.plausible.io%2Fdeep%2Fpath/?filters=((is,country,(RS)))&labels=(RS:Serbia)

ukutaht (approved):

ukutaht:
Getting this error on the preview deployment:

Screenshot 2024-07-31 at 13 47 41

apata:
> Getting this error on the preview deployment:
> 
> Screenshot 2024-07-31 at 13 47 41

Ah damn, browser difference in URLSearchParams implementation most likely. Getting the same with Safari. Checking

apata:
Refactored to one of the methods listed in the url search params polyfill library that we use (https://github.com/jerrybendy/url-search-params-polyfill?tab=readme-ov-file#foreach). That should be safer, even though entries should have been compatible (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/entries#browser_compatibility). Double checking with other browsers as well.

PR #4409, opened 2024-07-31
Refactor: remove transform_filters function from imported base

ukutaht:


apata:
Do I understand correctly that the fn had become superfluous over time?

ukutaht:
> Do I understand correctly that the fn had become superfluous over time?

Yes, it mostly became unneeded in https://github.com/plausible/analytics/pull/4391 and this PR removes it completely.

RobertJoonas (approved):
LGTM 👍

PR #4415, opened 2024-08-01
Edit goals with display names

aerosol:
### Changes

This PR introduces new capabilities of:

- editing goals
- setting up display names for goals

image

image

The display name is meant to decouple event names (used in tracking integration) from what dashboard/site settings users see. Especially useful in case of long/cryptic page paths, allowing marketing-oriented users quickly identify their meaning.

It's also used in the Stats/Sites API as the primary filtering identifier, replacing conditional naming used thus far. The API docs will need an update, a relevant PR will be referenced here.

By design, it's not possible to change the goal type when editing. So for example a custom event cannot be turned into a revenue goal or a pageview. We only allow editing initially provided set of data, otherwise goal deletion is recommended.

#### Migration

By default, all the pre-existing goals are migrated, such that:

- pageviews' display names become "Visit #{page_path}"
- custom events' display names identical to event name

Migrations extracted to: #4425 and #4426 

##### LiveView

The PR updates LiveView for listing and editing goals too. The form is now conditionally updated, depending on whether a `goal_id` has been passed (which effectively means edition attempt). For that reason, the optimization implemented via `PlausibleWeb.Live.Components.Modal` may not be fully applicable - the form is initially preloaded for new goal insertion, but on edit, it's updated via websocket messages to skip rendering irrelevant bits.

The optimizations mentioned were initially done to speed up modal loading in poor latency conditions. Those can be simulated locally by invoking  `liveSocket.enableLatencySim(500)` in the web inspector console, after navigating to goals site settings area.

There is one quirky bit there - applying `phash2` to DOM identifiers, so that AlpineJS picks up the changes made behind the scenes by LV. I couldn't find another way to do that. Without it, after navigating to goal settings and selecting the edit action first, the form modal won't be rendered correctly.

##### Minor UI changes

For convenience, in edit mode, clicking the display name for the first time will select the whole text input contents to allow quicker corrections.

The goals list now contains a whole lot of new information, so the list layout has been changed to accommodate them, at the same time aiming to skip display redundant info. For example, in case of pageview goal "/long_url" with display name "Visit /long_url", just the display name is shown, whilst in case of "/long_url" with display name "Pageview Alias", both will be visible.

#### Plugins API note

There should be no breaking changes other than displaying revenue goals differently in the Plugins API (used by our WordPress plugin) - the display name no longer contains currency, so "Purchase (EUR)" becomes "Purchase". This has been cleared with Daan and should cause no regressions with existing WordPress installations. 

#### How to test locally?

Re-seed the database with `mix ecto.reset`, visit goal settings, add & edit goals. Wherever goal names were used thus far, the display name should appear (dashboard, funnel settings).

#### What bugs I haven't solved yet

- [ ] ~~When editing a goal on chromium-based browsers, holding LMB to select the ComboBox dropdown contents and dragging outside of the modal window, and then releasing the LMB, triggers "click away" behaviour effectively closing the modal~~. https://github.com/alpinejs/alpine/discussions/807#discussioncomment-92461
- ~~[ ] In goal settings list, on Safari (that I don't have access to), the tooltips from `title=` attribute are not appearing on hover. Reason unknown yet. Those tooltips reportedly don't work in existing Shields sections as well.~~ not a bug, filed improvement issue separately

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -49,6 +49,11 @@ defmodule Plausible.Goal do
     |> maybe_drop_currency()
   end
 
+  @spec display_name(t()) :: String.t()
+  def display_name(goal) do
+    goal.display_name
+  end
+
   @spec type(t()) :: :event | :page
   def type(%{event_name: event_name}) when is_binary(event_name), do: :event
   def type(%{page_path: page_path}) when is_binary(page_path), do: :page
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -135,7 +135,7 @@ end
 
 defimpl String.Chars, for: Plausible.Goal do
   def to_string(goal) do
-    "#{goal.display_name}"
+    goal.display_name
   end
 end
 
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -8,6 +8,7 @@ defmodule Plausible.Goal do
   schema "goals" do
     field :event_name, :string
     field :page_path, :string
+    field :display_name, :string
 
     on_ee do
       field :currency, Ecto.Enum, values: Money.Currency.known_current_currencies()
@@ -22,7 +23,8 @@ defmodule Plausible.Goal do
     timestamps()
   end
 
-  @fields [:id, :site_id, :event_name, :page_path] ++ on_ee(do: [:currency], else: [])
+  @fields [:id, :site_id, :event_name, :page_path, :display_name] ++
+            on_ee(do: [:currency], else: [])
 
   @max_event_name_length 120
 
@@ -35,10 +37,10 @@ defmodule Plausible.Goal do
     |> cast_assoc(:site)
     |> update_leading_slash()
     |> validate_event_name_and_page_path()
-    |> update_change(:event_name, &String.trim/1)
-    |> update_change(:page_path, &String.trim/1)
+    |> maybe_put_display_name()
     |> unique_constraint(:event_name, name: :goals_event_name_unique)
     |> unique_constraint(:page_path, name: :goals_page_path_unique)
+    |> unique_constraint(:display_name, name: :goals_site_id_display_name_index)
     |> validate_length(:event_name, max: @max_event_name_length)
     |> check_constraint(:event_name,
       name: :check_event_name_or_page_path,
@@ -48,12 +50,8 @@ defmodule Plausible.Goal do
   end
 
   @spec display_name(t()) :: String.t()
-  def display_name(%{page_path: path}) when is_binary(path) do
-    "Visit " <> path
-  end
-
-  def display_name(%{event_name: name}) when is_binary(name) do
-    name
+  def display_name(goal) do
+    goal.display_name
   end
 
   @spec type(t()) :: :event | :page
@@ -76,6 +74,8 @@ defmodule Plausible.Goal do
   defp validate_event_name_and_page_path(changeset) do
     if validate_page_path(changeset) || validate_event_name(changeset) do
       changeset
+      |> update_change(:event_name, &String.trim/1)
+      |> update_change(:page_path, &String.trim/1)
     else
       changeset
       |> add_error(:event_name, "this field is required and cannot be blank")
@@ -100,6 +100,24 @@ defmodule Plausible.Goal do
       changeset
     end
   end
+
+  defp maybe_put_display_name(changeset) do
+    clause =
+      Enum.map([:display_name, :page_path, :event_name], &get_field(changeset, &1))
+
+    case clause do
+      [nil, path, _] when is_binary(path) ->
+        put_change(changeset, :display_name, "Visit " <> path)
+
+      [nil, _, event_name] when is_binary(event_name) ->
+        put_change(changeset, :display_name, event_name)
+
+      _ ->
+        changeset
+    end
+    |> update_change(:display_name, &String.trim/1)
+    |> validate_required(:display_name)
+  end
 end
 
 defimpl Jason.Encoder, for: Plausible.Goal do
@@ -110,14 +128,14 @@ defimpl Jason.Encoder, for: Plausible.Goal do
     |> Map.put(:goal_type, Plausible.Goal.type(value))
     |> Map.take([:id, :goal_type, :event_name, :page_path])
     |> Map.put(:domain, domain)
-    |> Map.put(:display_name, Plausible.Goal.display_name(value))
+    |> Map.put(:display_name, value.display_name)
     |> Jason.Encode.map(opts)
   end
 end
 
 defimpl String.Chars, for: Plausible.Goal do
   def to_string(goal) do
-    Plausible.Goal.display_name(goal)
+    "#{goal.display_name}"
   end
 end
 
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -49,11 +49,6 @@ defmodule Plausible.Goal do
     |> maybe_drop_currency()
   end
 
-  @spec display_name(t()) :: String.t()
-  def display_name(goal) do
-    goal.display_name
-  end
-
   @spec type(t()) :: :event | :page
   def type(%{event_name: event_name}) when is_binary(event_name), do: :event
   def type(%{page_path: page_path}) when is_binary(page_path), do: :page
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -64,15 +64,13 @@ defmodule Plausible.Goals do
     end
   end
 
-  def find_or_create(
-        site,
-        %{
-          "goal_type" => "event",
-          "event_name" => event_name,
-          "currency" => currency
-        } = params
-      )
+  def find_or_create(site, %{
+        "goal_type" => "event",
+        "event_name" => event_name,
+        "currency" => currency
+      } = params)
       when is_binary(event_name) and is_binary(currency) do
+
     with {:ok, goal} <- create(site, params, upsert?: true) do
       if to_string(goal.currency) == currency do
         {:ok, goal}
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -64,13 +64,15 @@ defmodule Plausible.Goals do
     end
   end
 
-  def find_or_create(site, %{
-        "goal_type" => "event",
-        "event_name" => event_name,
-        "currency" => currency
-      } = params)
+  def find_or_create(
+        site,
+        %{
+          "goal_type" => "event",
+          "event_name" => event_name,
+          "currency" => currency
+        } = params
+      )
       when is_binary(event_name) and is_binary(currency) do
-
     with {:ok, goal} <- create(site, params, upsert?: true) do
       if to_string(goal.currency) == currency do
         {:ok, goal}
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -52,9 +52,10 @@ defmodule Plausible.Goals do
   @spec update(Plausible.Goal.t(), map()) ::
           {:ok, Goal.t()} | {:error, Ecto.Changeset.t()} | {:error, :upgrade_required}
   def update(goal, params) do
-    with changeset <- Goal.changeset(goal, params),
-         site <- Repo.preload(goal, :site).site,
-         :ok <- maybe_check_feature_access(site, changeset),
+    changeset = Goal.changeset(goal, params)
+    site = Repo.preload(goal, :site).site
+
+    with :ok <- maybe_check_feature_access(site, changeset),
          {:ok, goal} <- Repo.update(changeset) do
       on_ee do
         {:ok, Repo.preload(goal, :funnels)}
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -10,10 +10,13 @@ defmodule Plausible.Goals do
 
   @spec get(Plausible.Site.t(), pos_integer()) :: nil | Plausible.Goal.t()
   def get(site, id) when is_integer(id) do
-    site
-    |> get_query()
-    |> where([g], g.id == ^id)
-    |> Repo.one()
+    q =
+      from g in Plausible.Goal,
+        where: g.site_id == ^site.id,
+        order_by: [desc: g.id],
+        where: g.id == ^id
+
+    Repo.one(q)
   end
 
   @spec create(Plausible.Site.t(), map(), Keyword.t()) ::
@@ -286,12 +289,5 @@ defmodule Plausible.Goals do
 
   defp maybe_trim(other) do
     other
-  end
-
-  defp get_query(site) do
-    from g in Plausible.Goal,
-      where: g.site_id == ^site.id,
-      order_by: [desc: g.id],
-      group_by: g.id
   end
 end
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -68,9 +68,8 @@ defmodule Plausible.Goals do
         "goal_type" => "event",
         "event_name" => event_name,
         "currency" => currency
-      })
+      } = params)
       when is_binary(event_name) and is_binary(currency) do
-    params = %{"event_name" => event_name, "currency" => currency}
 
     with {:ok, goal} <- create(site, params, upsert?: true) do
       if to_string(goal.currency) == currency do
@@ -90,15 +89,15 @@ defmodule Plausible.Goals do
     end
   end
 
-  def find_or_create(site, %{"goal_type" => "event", "event_name" => event_name})
+  def find_or_create(site, %{"goal_type" => "event", "event_name" => event_name} = params)
       when is_binary(event_name) do
-    create(site, %{"event_name" => event_name}, upsert?: true)
+    create(site, params, upsert?: true)
   end
 
   def find_or_create(_, %{"goal_type" => "event"}), do: {:missing, "event_name"}
 
-  def find_or_create(site, %{"goal_type" => "page", "page_path" => page_path}) do
-    create(site, %{"page_path" => page_path}, upsert?: true)
+  def find_or_create(site, %{"goal_type" => "page", "page_path" => _} = params) do
+    create(site, params, upsert?: true)
   end
 
   def find_or_create(_, %{"goal_type" => "page"}), do: {:missing, "page_path"}
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -8,6 +8,14 @@ defmodule Plausible.Goals do
   alias Plausible.Goal
   alias Ecto.Multi
 
+  @spec get(Plausible.Site.t(), pos_integer()) :: nil | Plausible.Goal.t()
+  def get(site, id) when is_integer(id) do
+    site
+    |> get_query()
+    |> where([g], g.id == ^id)
+    |> Repo.one()
+  end
+
   @spec create(Plausible.Site.t(), map(), Keyword.t()) ::
           {:ok, Goal.t()} | {:error, Ecto.Changeset.t()} | {:error, :upgrade_required}
   @doc """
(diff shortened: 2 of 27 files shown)


lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -45,54 +42,75 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
...
+                </button>
+
+                <button
+                  :if={goal.currency && !@revenue_goals_enabled?}
+                  id={"edit-goal-#{goal.id}-disabled"}
+                  disabled
+                  class="cursor-not-allowed"
+                >
+                  <Heroicons.pencil_square class="mr-4 feather feather-sm text-gray-400 dark:text-gray-600" />
+                </button>
```
apata, line 102: **meganitpick, non-blocking**: Since the margin of the edit button icon is considered part of the clickable button, the buttons to edit and delete are effectively different sizes, and there's no empty unclickable space between them. To solve, one could move the margin to the button or specify the horizontal gap in the parent flex container like `div class="flex items-center w-1/3 gap-x-4"` (https://tailwindcss.com/docs/gap). 

Screenshot 2024-08-06 at 12 25 08

apata: This should move the margin to button and solve it:
```suggestion
                <button
                  :if={!goal.currency || (goal.currency && @revenue_goals_enabled?)}
                  phx-click="edit-goal"
                  phx-value-goal-id={goal.id}
                  id={"edit-goal-#{goal.id}"}
                  class="mr-4"
                >
                  <Heroicons.pencil_square class="feather feather-sm text-indigo-800 hover:text-indigo-500 dark:text-indigo-500 dark:hover:text-indigo-300" />
                </button>

                <button
                  :if={goal.currency && !@revenue_goals_enabled?}
                  id={"edit-goal-#{goal.id}-disabled"}
                  disabled
                  class="mr-4 cursor-not-allowed"
                >
                  <Heroicons.pencil_square class="feather feather-sm text-gray-400 dark:text-gray-600" />
                </button>
```

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -216,13 +269,15 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
             type="button"
             x-on:click="active = !active; currency = ''"
             x-bind:aria-checked="active"
-            disabled={not @has_access_to_revenue_goals?}
+            disabled={(not @has_access_to_revenue_goals? or not is_nil(@goal)) && "disabled"}
```
apata, line 272: **nitpick, non-blocking**: Disabled is a boolean property, so we should be setting it to boolean
```suggestion
            disabled={not @has_access_to_revenue_goals? or not is_nil(@goal)}
```

test/plausible_web/live/goal_settings_test.exs
```diff
@@ -41,9 +41,14 @@ defmodule PlausibleWeb.Live.GoalSettingsTest do
       assert g3.currency
       assert resp =~ to_string(g3)
       assert resp =~ "Unlock Revenue Goals by upgrading to a business plan"
+
+      refute element_exists?(
+               resp,
+               ~s/button[phx-click="edit-goal"][phx-value-goal-id=#{g3.id}]#edit-goal-#{g3.id}/
+             )
```
apata, line 48: **ultranitpick, non-blocking**: by additionally asserting that the disabled version of the button is present, we'd give more of our UI intent to the reader

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -80,19 +95,20 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
         <div class="py-4" x-show="!tabSelectionInProgress">
           <PlausibleWeb.Components.Generic.button type="submit" class="w-full">
-            Add Goal →
+            <%= if @goal, do: "Update", else: "Add" %> Goal →
           </PlausibleWeb.Components.Generic.button>
         </div>
 
         <button
-          :if={@selected_tab == "custom_events" && @event_name_options_count > 0}
+          :if={@selected_tab == "custom_events" && @event_name_options_count > 0 && is_nil(@goal)}
```
apata, line 111: As I was reading this and similar expanded conditions, I started wondering if the whole render output could be conditional on `case is_nil(goal)`. 

Having the two render branches is not beneficial on its own, but in that scenario, if we also change the attributes of some of the other components to be more direct, we could I think cut down on complexity, while still not repeating ourselves. 

For example, rather than [needing to derive whether currency is switchable from "goal"](https://github.com/plausible/analytics/pull/4415/files#diff-5d8ab8130a88f328ae953c117f37965290817fd3280f173ee9dcaac46f2d37ecR261), that behaviour would be driven by `attr(:allow_switching_currency, :boolean)` and set/derived in the two render cases. 

This is super late in the game though, and should just be considered food for thought.

lib/plausible/goal/schema.ex
```diff
@@ -48,12 +50,8 @@ defmodule Plausible.Goal do
   end
 
   @spec display_name(t()) :: String.t()
-  def display_name(%{page_path: path}) when is_binary(path) do
-    "Visit " <> path
-  end
-
-  def display_name(%{event_name: name}) when is_binary(name) do
-    name
+  def display_name(goal) do
```
apata, line 53: **question, non-blocking**: Do we still get anything from this besides not needing to immediately refactor the couple of cases where its used?

apata (approved):

lib/plausible_web/live/goal_settings/list.ex
aerosol: Awesome, thanks!

lib/plausible_web/live/goal_settings/form.ex
aerosol: Hmm, pretty sure that change was deliberate as I ran into some issues when it was just `disabled` attribute set (without the actual value). Looking.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -181,7 +213,28 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+          </.error>
+        </div>
+
+        <div class="mt-2">
+          <.label for={"custom_event_display_name_input_#{@suffix}"}>
+            Display Name
+          </.label>
+
+          <.input
+            id="custom_event_display_name_input"
```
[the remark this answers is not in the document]
aerosol: I think the suffix comes from https://github.com/plausible/analytics/commit/d5a1d6765e0ea7dcdc7b5b28d561d8ccc282bdec and could have something to do with tabindex, so perhaps we should keep it? cc @zoldar

test/plausible_web/live/goal_settings_test.exs
aerosol: Thanks, good call 1243ebee3

lib/plausible/goal/schema.ex
aerosol: Exactly, not touching unrelated files just yet, I'll do the clean up in another PR. Thank you for noticing.

lib/plausible_web/live/goal_settings/form.ex
aerosol: Hmm, this sounds reasonable but I'm having trouble understanding exactly what do you mean by "two render branches" here. Would you mind illustrating it with a bit more (can be pseudo) code?

apata: I was thinking along these lines:
```ex
def render(assigns) do
  case is_nil(assigns.goal) do
    true ->
      ~H"""
        <div id={@id}>
          <.form
            :let={f}
            for={@form}
            phx-submit="save-goal"
            phx-target={@myself}
            x-data="{ tabSelectionInProgress: false }"
            >
          {/* more complex "new goal" form, with tabs and tab loading code, autoconfigure, etc */}
          {/* within this code block, any components can be given attributes that arise from the fact that this is the "new goal" form */}
          <.custom_event_fields can_toggle_revenue_goal={@has_access_to_revenue_goals} />
        </.form>
      </div>
      """

    false ->
      ~H"""
        <div id={@id}>
          <.form
            :let={f}
            for={@form}
            phx-submit="save-goal"
            phx-target={@myself}
          >
          {/* less complex "edit goal" form, no tabs, no autoconfigure */}
          {/* within this code block, any components can be given attributes that arise from the fact that this is the "edit goal" form */}
          <.custom_event_fields can_toggle_revenue_goal={false} />
        </.form>
      </div>
      """
  end
end
```

aerosol: Thanks @apata. Not entirely sure if this will work at runtime, will check. We could however split this into two separate LiveViews, e.g. `CreateForm` and `EditForm` with some degree of redundancy. Let me fiddle a bit and see if this is sensible direction in the context of this PR alone.

apata: It did work on my machine, but I didn't win much without making the subcomponents stupider as well: didn't get into that yet. With two liveviews, I'd be worried about ending up repeating the data preloading, form validation, error handling and other wiring logic. When it comes to repetitive stuff, I'd much rather it be HTML compositions than business logic.

aerosol: Perhaps https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-slots could be applicable here.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -181,7 +213,28 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+            x-on-selection-change="document.getElementById('custom_event_display_name_input').setAttribute('value', $event.detail.value.displayValue)"
+          />
+
+          <.error :for={msg <- Enum.map(@f[:event_name].errors, &translate_error/1)}>
+            <%= msg %>
+          </.error>
+        </div>
+
+        <div class="mt-2">
+          <.label for={"custom_event_display_name_input_#{@suffix}"}>
```
aerosol, line 227: Is there any significance to the suffix here or can it be skipped entirely?

lib/plausible_web/live/goal_settings/form.ex
apata: Ah, and we are already making use of it in the repository. That's a good feature and similar to React.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -236,17 +291,18 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
               ]}
-              id="enable-revenue-tracking"
+              id={"enable-revenue-tracking-#{:erlang.phash2(@f)}"}
             >
               Enable Revenue Tracking
             </span>
           </button>
 
-          <div x-show="active">
+          <div x-show="active" id={"revenue-input-#{:erlang.phash2(@f)}"}>
```
aerosol, line 300: Although the cost is low, these could be calculated just once. I haven't found any other way to make AlpineJS pick up the changes LiveView is making after clicking "edit" on first connected mount.

test/plausible_web/live/goal_settings/form_test.exs
```diff
@@ -31,39 +31,64 @@ defmodule PlausibleWeb.Live.GoalSettings.FormTest do
       lv = get_liveview(conn, site)
       html = render(lv)
 
-      [event_name_display, event_name, currency_display, currency_submit] =
-        find(html, "#goals-form-modalseq0 input")
-
+      event_name_display = find(html, "form input#event_name_input_modalseq0-tabseq0")
```
aerosol, line 34: This test feels terrible :/

test/support/factory.ex
```diff
@@ -107,8 +107,32 @@ defmodule Plausible.Factory do
     }
   end
 
-  def goal_factory do
-    %Plausible.Goal{}
+  def goal_factory(attrs) do
```
aerosol, line 110: only to support existing tests - ideally no factory is used for test goal creation. We can gradually switch to `Goals.create` instead

test/plausible_web/live/goal_settings/form_test.exs
aerosol: https://github.com/plausible/analytics/pull/4415/commits/54d2d92d486332ad22c448d5df782148ba3a998c

lib/plausible/goals/goals.ex
```diff
@@ -41,14 +49,30 @@ defmodule Plausible.Goals do
...
-      })
-      when is_binary(event_name) and is_binary(currency) do
-    params = %{"event_name" => event_name, "currency" => currency}
+  @spec update(Plausible.Goal.t(), map()) ::
+          {:ok, Goal.t()} | {:error, Ecto.Changeset.t()} | {:error, :upgrade_required}
+  def update(goal, params) do
+    with changeset <- Goal.changeset(goal, params),
+         site <- Repo.preload(goal, :site).site,
+         :ok <- maybe_check_feature_access(site, changeset),
+         {:ok, goal} <- Repo.update(changeset) do
```
zoldar, line 58:
```suggestion
    changeset = Goal.changeset(goal, params)
    site = Repo.preload(goal, :site).site

    with :ok <- maybe_check_feature_access(site, changeset),
         {:ok, goal} <- Repo.update(changeset) do
```

lib/plausible/goals/goals.ex
```diff
@@ -262,4 +286,11 @@ defmodule Plausible.Goals do
   defp maybe_trim(other) do
     other
   end
+
+  defp get_query(site) do
+    from g in Plausible.Goal,
+      where: g.site_id == ^site.id,
+      order_by: [desc: g.id],
+      group_by: g.id
```
zoldar, line 294: Is it necessary? 🤔

lib/plausible/goals/goals.ex
```diff
@@ -262,4 +286,11 @@ defmodule Plausible.Goals do
   defp maybe_trim(other) do
     other
   end
+
+  defp get_query(site) do
+    from g in Plausible.Goal,
```
zoldar, line 291: Perhaps better to inline this into `get/2` as it's only used there now?

lib/plausible_web/live/goal_settings.ex
```diff
@@ -43,7 +43,8 @@ defmodule PlausibleWeb.Live.GoalSettings do
        site_id: site_id,
        domain: domain,
        displayed_goals: socket.assigns.all_goals,
-       filter_text: ""
+       filter_text: "",
+       goal_id: nil
```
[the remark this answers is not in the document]
zoldar: `form_goal_id` perhaps? At least in the context of this (top-level) component.

lib/plausible_web/live/goal_settings.ex
```diff
@@ -103,6 +106,16 @@ defmodule PlausibleWeb.Live.GoalSettings do
     {:noreply, assign(socket, displayed_goals: new_list, filter_text: filter_text)}
   end
 
+  def handle_event("edit-goal", %{"goal-id" => goal_id}, socket) do
+    socket = socket |> assign(goal_id: goal_id) |> Modal.open("goals-form-modal")
```
zoldar, line 110: Casting `goal_id` to integer like it's done for "delete-goal" event would make handling the value in `Form` component a bit less hairy IMHO.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -33,13 +43,15 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+        selected_tab: selected_tab,
         tab_sequence_id: 0,
         site: site,
         has_access_to_revenue_goals?: has_access_to_revenue_goals?,
         existing_goals: assigns.existing_goals,
         on_save_goal: assigns.on_save_goal,
-        on_autoconfigure: assigns.on_autoconfigure
+        on_autoconfigure: assigns.on_autoconfigure,
+        goal_id: goal_id,
+        goal: goal_id && goal
```
zoldar, line 54:
```suggestion
        goal: goal
```

should be good enough

lib/plausible_web/live/goal_settings/form.ex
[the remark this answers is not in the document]
zoldar: `.input` isn't live component, so I think it's safe to remove the suffix from respective `.label`. It's meant for ensuring nested live components are fully reset when either switching tabs of cycling a modal wrapping them. Not an issue here.

lib/plausible_web/live/goal_settings/form.ex
zoldar: You could append existing `suffix` instead, if I'm getting it right? - It's changing with every modal open/close cycle.

aerosol:
> You could append existing suffix instead, if I'm getting it right? - It's changing with every modal open/close cycle.

hmm here's the patch I tried:

```patch
commit 7b013a8959f705fd772672a1031e138b0a71ece9
Author: Adam Rutkowski <hq@mtod.org>
Date:   Thu Aug 8 11:58:51 2024 +0200

    wip

diff --git a/lib/plausible_web/live/goal_settings/form.ex b/lib/plausible_web/live/goal_settings/form.ex
index 3923f5f82..01862ca91 100644
--- a/lib/plausible_web/live/goal_settings/form.ex
+++ b/lib/plausible_web/live/goal_settings/form.ex
@@ -149,7 +149,7 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
         suggest_fun={fn input, _options -> suggest_page_paths(input, @site) end}
         selected={if @goal && @goal.page_path, do: @goal.page_path}
         creatable
-        x-on-selection-change="document.getElementById('display_name_input').setAttribute('value', 'Visit ' + $event.detail.value.displayValue)"
+        x-on-selection-change="document.getElementById('pageview_display_name_input').setAttribute('value', 'Visit ' + $event.detail.value.displayValue)"
       />
 
       <.error :for={msg <- Enum.map(@f[:page_path].errors, &translate_error/1)}>
@@ -157,12 +157,12 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
       </.error>
 
       <div class="mt-2">
-        <.label for={"display_name_input_#{@suffix}"}>
+        <.label for="pageview_display_name_input">
           Display Name
         </.label>
 
         <.input
-          id="display_name_input"
+          id="pageview_display_name_input"
           field={@f[:display_name]}
           type="text"
           x-data="{ firstFocus: true }"
@@ -224,7 +224,7 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
         </div>
 
         <div class="mt-2">
-          <.label for={"custom_event_display_name_input_#{@suffix}"}>
+          <.label for="custom_event_display_name_input">
             Display Name
           </.label>
 
@@ -272,12 +272,12 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
             disabled={not @has_access_to_revenue_goals? or not is_nil(@goal)}
           >
             <span
-              id={"currency-switcher-1-#{:erlang.phash2(@f)}"}
+              id={"currency-switcher-1-#{@suffix}"}
               class="relative inline-flex h-6 w-11 flex-shrink-0 rounded-full border-2 border-transparent transition-colors duration-200 ease-in-out focus:outline-none focus:ring-2 focus:ring-indigo-600 focus:ring-offset-2"
               x-bind:class="active ? 'bg-indigo-600' : 'dark:bg-gray-700 bg-gray-200'"
             >
               <span
-                id={"currency-switcher-2-#{:erlang.phash2(@f)}"}
+                id={"currency-switcher-2-#{@suffix}"}
                 aria-hidden="true"
                 class="pointer-events-none inline-block h-5 w-5 transform rounded-full bg-white shadow ring-0 transition duration-200 ease-in-out"
                 x-bind:class="active ? 'dark:bg-gray-800 translate-x-5' : 'dark:bg-gray-800 translate-x-0'"
@@ -291,13 +291,13 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
                   else: "text-gray-500 dark:text-gray-300"
                 )
               ]}
-              id={"enable-revenue-tracking-#{:erlang.phash2(@f)}"}
+              id={"enable-revenue-tracking-#{@suffix}"}
             >
               Enable Revenue Tracking
             </span>
           </button>
 
-          <div x-show="active" id={"revenue-input-#{:erlang.phash2(@f)}"}>
+          <div x-show="active" id={"revenue-input-#{@suffix}"}>
             <.live_component
               id={"currency_input_#{@suffix}"}
               submit_name={@f[:currency].name}
```

Unfortunately this doesn't work well if the very first action you take is edit an existing revenue goal. The currency switcher is then disabled, whilst it should be enabled and allow currency change.

lib/plausible/goals/goals.ex
aerosol: done via https://github.com/plausible/analytics/pull/4415/commits/3cd2aea2504b173c1ec54828bbcee9a8a1ebe161

lib/plausible/goals/goals.ex
aerosol: noo 🤔
leftover from plugins api pagination I think

https://github.com/plausible/analytics/pull/4415/commits/3cd2aea2504b173c1ec54828bbcee9a8a1ebe161

PR #4417, opened 2024-08-01
Map lowercase tagged sources to capitalized form during ingestion

ukutaht:
### Changes

This work is in preparation of a bigger project - [Acquisition Channel](https://feedback.plausible.io/175).

This PR changes how we handle manually tagged acquisition data. When links are manually tagged, usually the `utm_source` parameter has a value like `google` or `facebook`. However, when we automatically detect sources based on the Referer header, we store the `source` field as `Google` or Facebook`. So then we have dashboards that have both sources `google` and `Google` where the first is manually tagged traffic and the second is automatically detected.

However, both are coming from the same source so we would like to merge the lowercase and the capitalized entry in sources.

The way I've decided to tackle it here is to build a lookup table from the `RefInspector` library that stores all sources in a table like:

```
lowercase(source) -> source
```

We use this table to lookup `google -> Google` during ingestion, for example.

The logic is that if the manually passed `utm_source` matches the lowercase form of any known traffic source, it gets transformed and stored as the capitalized form like automatically added sources.

We have also discussed a data migration for historical data to apply the same transformation. I'll look into that next.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/mix/tasks/send_pageview.ex
+++ b/lib/mix/tasks/send_pageview.ex
@@ -17,6 +17,7 @@ defmodule Mix.Tasks.SendPageview do
   @default_referrer "https://google.com"
   @default_event "pageview"
   @default_props "{}"
+  @default_queryparams ""
   @options [
     ip: :string,
     user_agent: :string,
@@ -28,7 +29,8 @@ defmodule Mix.Tasks.SendPageview do
     event: :string,
     props: :string,
     revenue_currency: :string,
-    revenue_amount: :string
+    revenue_amount: :string,
+    queryparams: :string
   ]
 
   def run(opts) do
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/send_pageview.ex
```diff
@@ -17,6 +17,7 @@ defmodule Mix.Tasks.SendPageview do
   @default_referrer "https://google.com"
```
ukutaht, line 17: Added this for manual testing

lib/plausible/ingestion/acquisition.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  def init() do
+    :ets.new(__MODULE__, [
+      :named_table,
+      :set,
+      :public,
+      {:read_concurrency, true}
+    ])
+
+    [{"referers.yml", map}] = RefInspector.Database.list(:default)
```
macobo, line 12: Q: Could and should this be built into RefInspector?

macobo (approved):

macobo:
This is missing a changelog entry.

lib/plausible/ingestion/acquisition.ex
ukutaht: > Q1: Could and should this be built into RefInspector?

What part exactly? I think RefInspector does maintain its own `ets` table but not in the exact form we need it (lowercase name -> original name)

> Q2: Is this database expected to change existing values in the future? Could that cause issues for our clients?

I expect entries to be added over time but not removed or modified. Should be safe.

PR #4418, opened 2024-08-01
Data migration for capitalizing sources

ukutaht:
### Changes

Data migration follow-up for https://github.com/plausible/analytics/pull/4417


macobo:
Should we run this automatically within a migration or suggest users run it for a self-hosted release?

ukutaht:
> Should we run this automatically within a migration or suggest users run it for a self-hosted release?

They way I did it, it will run automatically for self-hosters. I don't think a manual migration is necessary for this.

ukutaht:
Ran this on the staging/preview environment and seems to be working as expected

macobo (approved):

PR #4419, opened 2024-08-02
Add Details modals for all Devices reports

RobertJoonas:
### Changes

This PR adds a "Details" link to all the following dashboard reports:

* Browsers
* Browser versions
* Operating systems
* Operating system versions
* Screen sizes

All those modals will be using the `BreakdownModal` component under the hood, meaning that search functionality is also available (except for Screen Sizes, where it doesn't make sense since the number of entries is always 5 at most)

All the new Details views will display `bounce_rate` and `visit_duration` metrics (in additon to `visitors` and `%`).

### Tests

### Changelog

### Documentation

### Dark mode


apata (approved):

assets/js/dashboard/stats/modals/devices/browsers-modal.js
```diff
@@ -0,0 +1,48 @@
...
+  const getFilterInfo = useCallback((listItem) => {
+    return {
+      prefix: reportInfo.dimension,
+      filter: ["is", reportInfo.dimension, [listItem.name]]
+    }
+  }, [reportInfo.dimension])
+
+  const addSearchFilter = useCallback((query, searchString) => {
+    return addFilter(query, ['contains', reportInfo.dimension, [searchString]])
+  }, [reportInfo.dimension])
```
apata, line 31: **nitpick, non-blocking**: Since `reportInfo.dimension` is constant after the split to separate components, this function is not specific to the component: it doesn't require any inputs from the React component tree. `function queryWithSearchFactory(dimension) => (query, searchString) => parameters for API call`

RobertJoonas: True. Let's add that logic into the API module. Currently it's a bit tricky to do and would make things more hairy so let's do that as part of a bigger refactor.

PR #4427, opened 2024-08-08
Add Typescript

apata:
### Changes

* Updates node
* Updates tailwindcss
* Adds Typescript and related eslint plugins
* Since Esbuild with Typescript wasn't able to resolve the d3 dependency of the library that we use to render countries map, removed that library and rewrote the map on the latest d3
* Adds ThemeContext to be able to watch theme changes centrally
* Refactors some common components from `.js` to `.tsx` and formats them by adding the `/* @format */` pragma 

### Tests
- [x] Tested manually

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/navigation/use-app-navigate.tsx
```diff
@@ -0,0 +1,85 @@
+/* @format */
```
macobo, line 1: What is the plan for the format pragma? Should we just turn it on for all of tsx?

macobo:
There's a bunch of warnings in this PR (I assume new due to updating dependencies). Could we whitelist these with ignore comments to avoid busying other PRs?

assets/js/dashboard/navigation/use-app-navigate.tsx
apata: Checking if it’s possible. From my side, I struggled to arrive at a working stylelint eslint prettier configuration that didn’t require running a reformat of the whole codebase. I wanted to postpone full adoption until all engineers are comfortable running formatters locally.

apata: Prettier has the [insertPragma](https://prettier.io/docs/en/options.html#insert-pragma) option, but that is superseded by `requirePragma`: if I open a random unformatted .js file or any new .tsx file and run prettier format on it, it will look for the pragma and since it is not present, won't do anything: I still have to add the pragma manually.

apata:
@macobo Any further concerns/blockers?

macobo (approved):

PR #4451, opened 2024-08-19
Prepare for graph v2

RobertJoonas:
### Changes

Preparation for migrating the main_graph to API v2.

* Globally rename `date` interval to `day` to be consistent with the `time:*` dimensions in the new API. Docs update in Stats API is pending: https://github.com/plausible/docs/pull/530
* Stop returning the used `interval` from the `main_graph` controller action. The frontend already decides the interval based on what is valid, and the backend never overrides it - just validates it.
* Make it possible to pass a `date` param in API v2, relative to which the date range will be constructed.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/filters/query_parser.ex
```diff
@@ -116,6 +115,28 @@ defmodule Plausible.Stats.Filters.QueryParser do
 
   defp parse_clauses_list(filter), do: {:error, "Invalid filter '#{inspect(filter)}'"}
 
+  defp parse_date_range(_site, %{"date_range" => date_range})
+       when date_range in ["realtime", "30m"] do
+    {:ok, date_range}
+  end
+
+  defp parse_date_range(site, %{"date_range" => date_range, "date" => date})
```
macobo, line 123: If I understand this right this adds a bunch of new structures that are valid date ranges which is problematic - we don't want to expose this complexity for the users.

Q: Why do we need the date option in here?

lib/plausible/stats/time.ex
```diff
@@ -47,6 +26,27 @@ defmodule Plausible.Stats.Time do
     {first_datetime, last_datetime}
   end
 
+  defp realtime_utc_boundaries(site, now, period) do
```
macobo, line 29: It's worth inlining this function as it's only called once.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -227,8 +226,8 @@ defmodule PlausibleWeb.Api.StatsController do
       imports_exist: site.complete_import_ids != [],
       comparing_from: comparison_query && comparison_query.date_range.first,
       comparing_to: comparison_query && comparison_query.date_range.last,
-      from: query.date_range.first,
-      to: query.date_range.last
+      from: comparison_query && query.date_range.first,
```
macobo, line 229: comparison_query can be nil right? Would it be worth it to make the nil check explicit here?

Also `date_range.first` no longer works right in realtime cases so not sure this is even correct.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -116,6 +115,28 @@ defmodule Plausible.Stats.Filters.QueryParser do
 
   defp parse_clauses_list(filter), do: {:error, "Invalid filter '#{inspect(filter)}'"}
 
+  defp parse_date_range(_site, %{"date_range" => date_range})
```
macobo, line 118: Missing tests. Let's keep coverage of test/plausible/stats/query_parser_test.exs really really high.

test/plausible/stats/query_parser_test.exs
```diff
@@ -403,6 +403,8 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
 
   describe "date range validation" do
     test "parsing shortcut options", %{site: site} do
+      check_date_range("realtime", site, "realtime")
+      check_date_range("30m", site, "30m")
```
macobo, line 407: There are missing tests here - you added some other options above.

test/plausible/stats/query_parser_test.exs
```diff
@@ -358,7 +358,7 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
         ]
       }
 
-      assert {:ok, res} = parse(site, params, @today)
+      assert {:ok, res} = parse(site, Map.put(params, "date", @today))
```
macobo, line 361: The typing out of the whole query each time was intentional, as was the 3rd argument to parse. 

If queries now require and allow a "date" param each time then let's make it explicit rather than sneak it in.

I don't think users should be able to pass a "date" param though.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -116,6 +115,28 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+       when date_range in ["realtime", "30m"] do
+    {:ok, date_range}
```
macobo, line 120: There's a bunch of code that relies on being able to do `query.date_range.first` and `query.date_range.last`. This change breaks all of them, meaning we need to add exceptions to each of them or have hidden bugs. This sort of hidden assumptions make code painful to use.

Why do we need a separate type of value for realtime or 30m, is it ease-of-matching somewhere? Would using DateTime.range work? Would adding a new hidden but explicit query field work?

lib/plausible/stats/query_result.ex
```diff
@@ -33,6 +33,14 @@ defmodule Plausible.Stats.QueryResult do
     )
   end
 
+  defp result_date_range(date_range) when date_range in ["realtime", "30m"] do
+    date_range
```
macobo, line 37: This is painful for any API consumers since we weren't consistent with our return values in edge cases.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: This function call is only used to parse the `date` parameter. It delegates all further logic to `parse_date_range/3` which actually pattern matches on the allowed `date_range` params (and uses the parsed `date`, or `today(site)` if it wasn't provided)

> Q: Why do we need the date option in here? Wouldn't the "today" option from before still work well?

In the dashboard we want to be able to create queries with `date_range` shortcuts such as `week` or `month` which are relative to a `date` param. This shouldn't be exposed to the public API.

test/plausible/stats/query_parser_test.exs
RobertJoonas: As mentioned in my other comment, the `date` param is needed for the dashboard API requests.

> If queries now require and allow a "date" param each time then let's make it explicit rather than sneak it in.

Could you elaborate on this? How would you make it more explicit?

> I don't think users should be able to pass a "date" param though.

Agreed, as mentioned in the above comment - should not be exposed in the public API v2.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Ditto about the dropped commits - not relevant anymore in this PR.

test/plausible/stats/query_parser_test.exs
RobertJoonas: Ditto about the dropped commits - not relevant anymore in this PR.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Ditto about the dropped commits - not relevant anymore in this PR.

lib/plausible/stats/query_result.ex
RobertJoonas: Ditto about the dropped commits - not relevant anymore in this PR.

macobo (approved):

PR #4452, opened 2024-08-19
Refactor and consolidate user session logic

zoldar:
### Changes

This PR refactors and creates a groundwork for implementing persisted user sessions.

Review can be done commit-by-commit for convenience.

- User session management and retrieval is completely abstracted behind `PlausibleWeb.UserAuth`. The module takes care of managing cookies (currently the defaults session when and "logged_in" cookie shared with static site). There are also extension points introduced already for supporting persisted sessions and transition from the current (legacy) cookies to the new, token based ones which will be persisted using `Plausible.Auth.UserSession` (currently an embedded schema - just enough to support existing cookie auth).
- Cookies are now always renewed/reset, reducing the potential risk of [session fixation attacks](https://en.wikipedia.org/wiki/Session_fixation).
- There's a new assign introduced, called `live_socket_id`. Once the token based authentication is implemented, it will allow disconnecting LV clients on either explicit logout or after revoking the relevant token(s). More details in [the docs](https://hexdocs.pm/phoenix_live_view/security-model.html#disconnecting-all-instances-of-a-live-user). NOTE: As Phoenix PubSub is not fully setup, we might still not be able to fully benefit from that mechanism.
- The `current_user_id` value is no longer explicitly passed via `live_render`'s `session` option. All existing session keys are passed through implicitly, to this was redundant.
- `PlausibleWeb.Live.AuthContext` is now automatically populating `current_user` (and `current_user_session`) in LV context on mount - provided it's not provided by the `AuthPlug` already which is sooner in the overall pipeline.
- Both `Live.AuthContext` and `AuthPlug` ensure the shape of the loaded user is the same, always with an active subscription preloaded if there is one.
- Any direct references to `current_user_id` session value got removed from LV in favor of retrieving `current_user` implicitly set by the auth plug or context.
- The `current_user` from the context is used more consistently in a number of spots (mostly around billing where user was refetched and set under a different key, like `user`).

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -552,25 +552,25 @@ defmodule PlausibleWeb.AuthController do
   end
 
   defp render_settings(conn, opts) do
+    current_user = conn.assigns.current_user
     settings_changeset = Keyword.fetch!(opts, :settings_changeset)
     email_changeset = Keyword.fetch!(opts, :email_changeset)
-
-    user = Plausible.Users.with_subscription(conn.assigns[:current_user])
+    api_keys = Repo.preload(current_user, :api_keys).api_keys
 
     render(conn, "user_settings.html",
-      user: user |> Repo.preload(:api_keys),
+      api_keys: api_keys,
       settings_changeset: settings_changeset,
       email_changeset: email_changeset,
-      subscription: user.subscription,
-      invoices: Plausible.Billing.paddle_api().get_invoices(user.subscription),
-      theme: user.theme || "system",
-      team_member_limit: Quota.Limits.team_member_limit(user),
-      team_member_usage: Quota.Usage.team_member_usage(user),
-      site_limit: Quota.Limits.site_limit(user),
-      site_usage: Quota.Usage.site_usage(user),
-      pageview_limit: Quota.Limits.monthly_pageview_limit(user),
-      pageview_usage: Quota.Usage.monthly_pageview_usage(user),
-      totp_enabled?: Auth.TOTP.enabled?(user)
+      subscription: current_user.subscription,
+      invoices: Plausible.Billing.paddle_api().get_invoices(current_user.subscription),
+      theme: current_user.theme || "system",
+      team_member_limit: Quota.Limits.team_member_limit(current_user),
+      team_member_usage: Quota.Usage.team_member_usage(current_user),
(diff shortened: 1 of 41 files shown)


lib/plausible_web/live/sites.ex
```diff
@@ -112,10 +110,7 @@ defmodule PlausibleWeb.Live.Sites do
         >
           Total of <span class="font-medium"><%= @sites.total_entries %></span> sites
         </.pagination>
-        <.invitation_modal
-          :if={Enum.any?(@sites.entries, &(&1.entry_type == "invitation"))}
-          user={@user}
```
zoldar, line 117: This assign was completely unused in the modal component.

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -622,8 +622,7 @@ defmodule PlausibleWeb.AuthController do
     redirect_to = Map.get(params, "redirect", "/")
 
     conn
-    |> configure_session(drop: true)
-    |> delete_resp_cookie("logged_in")
+    |> UserAuth.log_out_user()
```
aerosol, line 625: Just to make sure - renewing is just as good as dropping the cookie now?

lib/plausible_web/live/auth_context.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  alias PlausibleWeb.UserAuth
+
+  defmacro __using__(_) do
+    quote do
+      on_mount unquote(__MODULE__)
+    end
+  end
+
+  def on_mount(:default, _params, session, socket) do
```
aerosol, line 19: nice :clap:

aerosol (approved):
Nice and clean :sparkles:

lib/plausible_web/controllers/auth_controller.ex
zoldar: Renewing _and_ clearing session is basically identical as dropping it in case of using COOKIE session store. It makes more of a difference in case of other stores (ETS, there's also Redis driver) where session data is stored server-side and session ID is stored in the cookie. Renewing then is indeed necessary to avoid malicious reuse of the same session ID (so-called session fixation attacks). Defaulting to renewal + clear is fine and additionally future-proof in case we ever want to switch the session store to avoid risk of potentially forgetting to update the reset strategy.

PR #4459, opened 2024-08-21
Onboarding improvements

aerosol:
### Changes

~~Migration will be extracted to a separate PR once the review is conducted.~~ [migration](https://github.com/plausible/analytics/pull/4486)

This PR aims to make the onboarding (and related flows, such as e.g. domain change) more helpful and visually consistent. The main functional change is the "customize installation" screen, where site type is determined early (WordPress/GTM/manual) and installation instructions are provided according to that inspection result. 

For manual and GTM installations, users can build the script snippet by selecting tracker extensions they're interested it (some of which will have side effects - e.g. including the "file downloads" extension will automatically create a relevant goal). 

As a consequence, the snippet window has been removed from general settings section (snippet may not be relevant at all, in case of WP installations) and replaced by 2-step flow allowing users to "review their installations". Similarly, the snippet window became no longer relevant during domain change.

By the way the following changes were also made:
  - focus.html layout has been removed, the distinction leads to unnecessary maintenance
  - site verification is no longer optional, `VERIFICATION_ENABLED` variable has been deprecated, it no longer has any effect. For self-hosted installations, verification will simply fail by default, unless browserless stack is set up. Instructions on how to do that for CE can be provided later on
  - minor visual rearrangements according to the new designs

Docs: major changes pending

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/site/installation_meta.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.Site.InstallationMeta do
+  @moduledoc """
+  Embedded schema for installation meta-data
+  """
+  use Ecto.Schema
+
+  @type t() :: %__MODULE__{}
+
+  embedded_schema do
```
aerosol, line 9: next up, we'll probably have "real domain" in here

zoldar (approved):
Nice work on getting rid of the focus layout btw ❤️ Left a couple comments, most of them mere suggestions.

lib/plausible_web/components/generic.ex
```diff
@@ -334,11 +336,45 @@ defmodule PlausibleWeb.Components.Generic do
...
+  end
+
+  attr :wrapper_class, :any, default: ""
+  attr :class, :any, default: ""
+  attr :icon?, :boolean, default: true
+  attr :position, :string, default: "bottom-10 margin-x-auto left-10 right-10"
+  slot :inner_block, required: true
+  slot :tooltip_content, required: true
+
+  def tooltip_non_sticky(assigns) do
```
apata, line 361: **suggestion, non-blocking**: since this is a bit repetitive, why not have an additional attribute on the tooltip?
```
  attr :sticky, :boolean, default: true
  attr :wrapper_class, :any, default: ""
  attr :class, :any, default: ""
  attr :icon?, :boolean, default: true
  attr :position, :string, default: "bottom-10 margin-x-auto left-10 right-10"
  slot :inner_block, required: true
  slot :tooltip_content, required: true

  def tooltip(assigns) do
    wrapper_data = if @sticky, do: "{sticky: false, hovered: false}", else: "{hovered: false}"
    show_inner = if @sticky, do: "hovered || sticky", else: "hovered"
    ~H"""
    <div x-data={wrapper_data} class={["tooltip-wrapper relative", @wrapper_class]}>
      <p
        x-on:click="sticky = true; hovered = true"
        x-on:click.outside="sticky = false; hovered = false"
        x-on:mouseover="hovered = true"
        x-on:mouseout="hovered = false"
        class={["cursor-pointer flex align-items-center", @class]}
      >
        <%= render_slot(@inner_block) %>
        <Heroicons.information_circle :if={@icon?} class="w-5 h-5 ml-2" />
      </p>
      <span
        x-show={show_inner}
        class={[
          "bg-gray-900 pointer-events-none absolute transition-opacity p-4 rounded text-sm text-white",
          @position
        ]}
      >
        <%= render_slot(List.first(@tooltip_content)) %>
      </span>
    </div>
    """
  end
```

lib/plausible_web/live/installation.ex
```diff
@@ -0,0 +1,463 @@
...
+      @script_config_params
+      |> Enum.into(%{}, fn param ->
+        {param, false}
+      end)
+      |> Map.merge(
+        params
+        |> Enum.filter(fn {key, value} -> key in @script_config_params and value == "on" end)
+        |> Enum.map(fn {key, _} -> {key, true} end)
+        |> Enum.into(%{})
+      )
```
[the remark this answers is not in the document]
aerosol: damn :D

lib/plausible_web/live/installation.ex
```diff
@@ -0,0 +1,463 @@
...
+    socket
+  end
+
+  defp update_script_config(socket, "outbound-links" = key, true) do
+    Plausible.Goals.create_outbound_links(socket.assigns.site)
+    update_script_config(socket, %{key => true})
+  end
+
+  defp update_script_config(socket, "outbound-links" = key, false) do
+    Plausible.Goals.delete_outbound_links(socket.assigns.site)
```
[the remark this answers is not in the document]
aerosol: Yes, confirmed with @metmarkosaric - the behavior is aligned with the WordPress plugin's. Historical records stay intact.

lib/plausible_web/live/installation.ex
```diff
@@ -0,0 +1,463 @@
...
+  @valid_qs_params @script_config_params ++ ["installation_type", "flow"]
+
+  def script_config_params, do: @script_config_params
+
+  def mount(
+        %{"website" => domain} = params,
+        %{"current_user_id" => user_id},
+        socket
+      ) do
+    site = Plausible.Sites.get_for_user!(user_id, domain)
```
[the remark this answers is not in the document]
aerosol: Thanks! 958edd5db

lib/plausible_web/live/verification.ex
```diff
@@ -8,75 +8,61 @@ defmodule PlausibleWeb.Live.Verification do
...
 
+  def mount(:not_mounted_at_router, %{"domain" => domain} = session, socket) do
+    mount(%{"website" => domain}, session, socket)
+  end
+
   def mount(
-        :not_mounted_at_router,
-        %{"domain" => domain} = session,
+        %{"website" => domain} = params,
+        session,
```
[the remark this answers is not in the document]
aerosol: hmm I'm still using `session` for initial assigns, but in tests only. 414d1a31e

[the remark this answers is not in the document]
aerosol: It's some remnant from the previous implementation where the LV was mounted isolated, I'll see how can I simplify this.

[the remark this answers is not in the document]
aerosol: 0f94bb678

lib/plausible_web/components/generic.ex
aerosol: Thanks for the suggestion @apata, pushed 2cfc131b0 and added the credit where it's due :)

aerosol:
Data migration for CE is failing: https://github.com/plausible/analytics/actions/runs/10612221725/job/29413425673?pr=4459

This is because `Plausible.DataMigration.SiteImports` relies on `Site` schema that gets migrated later on. How should we proceed in such case? cc @zoldar @ruslandoga

ruslandoga:
Ah, the issue seems a bit different. I think the fix here would be to not use Ecto schemas in data migrations and instead rely on "raw" `select`s.

aerosol:
hmm how about:

```diff
diff --git a/lib/plausible/data_migration/site_imports.ex b/lib/plausible/data_migration/site_imports.ex
index 8aaa03f3e..749fbeaa2 100644
--- a/lib/plausible/data_migration/site_imports.ex
+++ b/lib/plausible/data_migration/site_imports.ex
@@ -29,6 +29,7 @@ defmodule Plausible.DataMigration.SiteImports do
     sites_with_only_legacy_import =
       from(s in Site,
         as: :site,
+        select: %{id: s.id, imported_data: s.imported_data},
         where:
           not is_nil(s.imported_data) and fragment("?->>'status'", s.imported_data) == "ok" and
             not exists(site_import_query)
```

the test at `Plausible.DataMigration.SiteImportsTest` runs fine.

lib/plausible_web/flows.ex
```diff
@@ -0,0 +1,64 @@
...
+      "Verify installation"
+    ],
+    invitation: [
+      "Register",
+      "Activate account"
+    ],
+    provisioning: [
+      "Add site info",
+      "Install Plausible",
+      "Verify installation"
```
ruslandoga, line 31: 👋 @aerosol 

Is it possible to disable this somehow conditionally on CE? It might confuse some users:

Screenshot 2024-09-13 at 16 05 33

ruslandoga: E.g. we can enable it only if ghcr.io/browserless/chromium is running or if some extra ENV var is set.

config/runtime.exs
```diff
@@ -778,11 +778,6 @@ config :plausible, Plausible.PromEx,
   grafana: :disabled,
   metrics_server: :disabled
 
-config :plausible, Plausible.Verification,
-  enabled?:
-    get_var_from_path_or_env(config_dir, "VERIFICATION_ENABLED", "false")
```
ruslandoga, line 783: Seems like it was being done this way before. Has this configuration been moved elsewhere?

ruslandoga:
> For self-hosted installations, verification will simply fail by default, unless browserless stack is set up. Instructions on how to do that for CE can be provided later on

Ah, I see. I hope this won't cause too much confusion / questions on the forum :)

lib/plausible_web/templates/auth/login_form.html.heex
```diff
@@ -37,22 +37,23 @@
...
-          Don't have an account? <%= link("Register",
-            to: "/register",
-            class: "text-gray-800 dark:text-gray-50 underline"
-          ) %> instead.
-        </li>
-      <% end %>
-      <li>
+    <.focus_list>
+      <:item>
+        <%= if Keyword.fetch!(Application.get_env(:plausible, :selfhost), :disable_registration) == false do %>
```
ruslandoga, line 42: This renders an empty list element:

Screenshot 2024-09-13 at 16 20 10

lib/plausible_web/flows.ex
aerosol: Yeah, it felt like it was too costly to maintain a separate branch without verification. I think it's going to be easier going forward to set up instructions for self-hosters to run their own browserless stack. WYT?

lib/plausible_web/templates/auth/login_form.html.heex
aerosol: Yikes, sorry

lib/plausible_web/flows.ex
ruslandoga: This might be problematic for many users who decide to run on the smallest available instances (e.g. t2-micro on AWS):

```
[ec2-user@ip-172-31-46-71 ~]$ docker run --rm  ghcr.io/browserless/chromium
Unable to find image 'ghcr.io/browserless/chromium:latest' locally
latest: Pulling from browserless/chromium
602d8ad51b81: Pull complete
7e6a86bffb05: Pull complete
4f4fb700ef54: Pull complete
eadff0eab842: Pull complete
6b08750ce489: Pull complete
6931508ca973: Pull complete
e839281773da: Pull complete
d9fbbce37766: Pull complete
16df0a60c56d: Pull complete
daccdb15ea7d: Pull complete
872e2e99569f: Pull complete
9f74149fc518: Pull complete
35f28019a5e3: Pull complete
43a8a10c249b: Pull complete
b1b4ece4e245: Pull complete
fd5c44f4051d: Pull complete
62ed90b89bd0: Pull complete
0b23a1ec6119: Pull complete
2859dd417a71: Pull complete
a15b9ba04167: Pull complete
c76fa6cc8898: Pull complete
a6f34b5464b1: Pull complete
7aa3df8dfb9b: Pull complete
e15bbf55a158: Pull complete
9354a349fd43: Pull complete
362a19d48bd5: Pull complete
4708bbe9978e: Pull complete
4347e0d884d4: Extracting [==================================================>]  260.2MB/260.2MB
docker: failed to register layer: write /usr/local/bin/playwright-browsers/chromium-1134/chrome-linux/chrome: no space left on device.
```

ruslandoga: Since it's a one-time feature for self-hosters, I don't see much benefit in running idle Chrome in the background for it.

ruslandoga: Since it's possible to reach site settings from this page, shouldn't it be just a conditional render call when `not ce()?` for this "Verify Installation" step?

lib/plausible_web/templates/auth/login_form.html.heex
ruslandoga: @aerosol would it be OK for me to PR a fix?

lib/plausible_web/flows.ex
ruslandoga: @aerosol would it be OK for me to explore this conditional approach or should be postpone it until self-hosters' feedback? I.e. if nobody complains, then it would mean there is no problem.

lib/plausible_web/templates/auth/login_form.html.heex
aerosol: Of course, much appreciated, thank you

lib/plausible_web/flows.ex
aerosol: Hey, please make your call regarding this. I’ll be happy to review a PR. My reasoning was - for folks on small VPS boxes, I thought it’d be reasonable to say “in order for verification to work, you need more resources. alternatively verify your installation manually”. They are not missing much in such case. But if you feel it’s simple to make this step skippable based on configuration, go ahead.

PR #4460, opened 2024-08-21
Update Postgrex SSL config

ruslandoga:
### Changes

This PR removes `ssl_opts` that are no longer necessary after https://github.com/elixir-ecto/postgrex/pull/677 (and https://github.com/plausible/analytics/pull/4457)

### Tests
- [x] Config test has been updated

### Changelog
- [x] Changelog has been updated

### Documentation
- [x] Docs don't need updates (`DATABASE_CACERTFILE` is not documented right now)

### Dark mode


aerosol (approved):

aerosol (changes requested):
hmm CI says no

ruslandoga:
Hm, it seems like the behaviour has actually changed slightly. Previously `ssl_opts` were only applied when `ssl` connection was specified elsewhere, and now the new `ssl` opts actually force the `ssl` connection.

cnkk (approved):

PR #4461, opened 2024-08-21
Add types to dashboard query, make keybinds route specific

apata:
#### User facing
* Bugfix: Escape clears filters only when no modal is open
* Bugfix: Keybinds to set query periods only function when no modal is open

#### Other
* Adds types to dashboard query
* Refactors how the dashboard query is formed
* Rewires storing some query parameters from explicit functions within components to a navigation effect
  * Removes now superfluous navigation utilities and components
* Rolls query period and comparison mode into one component

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/components/dropdown.tsx
```diff
@@ -0,0 +1,123 @@
+/** @format */
```
ukutaht, line 1: In other places we've been using the components available in [HeadlessUI](https://headlessui.com/react/menu) instead of hand-rolling them. I think it would be preferable if we could also refactor the datepicker to use HeadlessUI then we don't need our own dropdown component at all.

ukutaht (approved):
Looks good overall!

assets/js/dashboard/components/dropdown.tsx
apata: I considered it when unifying the two dropdowns, but it wasn't apparent which one was better from UX perspective. 

The headless UI version doesn't support navigating within the dropdown with a tab, it overrides it to arrow keys instead. I'm no accessibility master but that didn't seem right, so it was the deciding factor to unify them to the hand rolled version. Could go back on that decision fairly easily though.

PR #4463, opened 2024-08-22
Implement token-based sessions

zoldar:
### Changes

Follow-up to https://github.com/plausible/analytics/pull/4452 implementing token-based sessions.

- A new table is added for storing sessions called `user_sessions` and used by `Auth.UserSession` schema
- The session struct stores user reference, token (randomly generated on creation), `last_used_at` and `timeout_at` timestamps
- Inactive session timeout is kept at 14 days, like before, just managed server-side
- New user sessions refresh is managed by a new plug, `UserSessionTouch` which also converts existing legacy sessions to new ones on the go
- There's a `CleanUserSessions` background job cleaning timed out sessions with an additional grace period of 7 days, just in case (sessions, when retrieved are always filtered by timeout, so that's not an issue)
- Retrievals of legacy sessions is logged with a warning. Logs will be scanned for occurrences to determine whether rate of legacy sessions goes down towards zero during/after 14 day period as expected
- Device name is detected using logic largely copied from `Ingestion.Event` pipeline. I have decided to not DRY these as they serve very different purposes.

## TODO

- [x] properly refresh valid and dispose of expired sessions
- [x] instrument retrievals of legacy session cookie (accomplished with simple logging)
- [x] more tests
- [x] ~investigate and fix borked "remember 2fa" cookie mechanism~ false alarm, it's fine
- [x] extract migration to a different PR (may be done post-review)

### Tests

### Changelog

### Documentation

### Dark mode


priv/repo/migrations/20240822095245_create_user_sessions.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Plausible.Repo.Migrations.CreateUserSessions do
+  use Ecto.Migration
+
+  def change do
+    create table(:user_sessions) do
+      add :user_id, references(:users, on_delete: :delete_all), null: false
+      add :token, :binary, null: false
```
apata, line 7: If the value of this works the same as a password / api key, should we not hash it for storage?

zoldar: Currently, client-side, this token is always stored in a hashed session cookie. The hash secret is stored in app environment (secret key base). If somebody gets their hands on a raw token, they can't do much with it client-side, without knowing secret key base - they can't reconstruct a valid session cookie from it alone.

lib/plausible_web/user_auth.ex
```diff
@@ -71,6 +75,26 @@ defmodule PlausibleWeb.UserAuth do
...
+    user_session
+  end
+
+  def touch_user_session(user_session, now \\ NaiveDateTime.utc_now(:second)) do
+    %{token: token, timeout_at: timeout_at, last_used_at: last_used_at} =
+      user_session
+      |> Auth.UserSession.touch_session(now)
+      |> Ecto.Changeset.apply_changes()
+
+    Repo.update_all(
```
ukutaht, line 90: Would it be possible to do this in a single database roundtrip using the regular `Repo.update` function here?

ukutaht: As in, replace the two roundtrips:
`Repo.update_all`
`Repo.reload`

with a single

`Repo.update`

ukutaht (commented):
Overall the implementation looks good to me.

I would like to review the DB queries we fire off during authentication. I tried loading a random settings screen with this PR and in the logs I counted:
4 SELECT requests from `user_sessions`
1 UPDATE to `user_sessions`
1 SELECT from `users`

I feel like we should be able to get by with
1 SELECT from `user_sessions` JOIN `users`
1 UPDATE to `user_sessions`

lib/plausible_web/user_auth.ex
zoldar: Improved with https://github.com/plausible/analytics/pull/4463/commits/defddfbcc9ccca1dd82b93e44415acc48bda2f2c. The original idea was to avoid crashing on concurrent updates or deletion of session. `Repo.update(..., allow_stale: true)` allows to accomplish that in a single query though.

zoldar:
@ukutaht I have revised session retrieval and modified logic so that session, user and last subscription are always fetched in a single query. I have also modified plug pipelines interacting with session to ensure they don't refetch the session or user unless necessary.

By the way, I have noticed that our external API endpoints (/api/event included) have `:fetch_session` and `AuthPlug` in their pipeline - I have removed both as none of those endpoints need the session (I'll do some double-checking for that still).

This limits the number of session related queries to minimum. When I open funnel settings, I can see 2 SELECTs of user session (one for dead view and one for live view, we can't get around that), and one UPDATE for user_session (via UserSessionTouch). All else is related to validating site access permission (user role, site membership etc.). If we want to address those, I'd opt for doing that in a separate PR.

lib/plausible_web/plugs/user_session_touch.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  def init(opts \\ []) do
+    opts
+  end
+
+  def call(conn, _opts) do
+    # NOTE: Needed only during transitional 14-day period
+    conn = UserAuth.convert_legacy_session(conn)
+
+    if user_session = conn.assigns[:current_user_session] do
```
aerosol, line 18: Wondering if we could spare some updates by checking if current session is "any close" to the end?

lib/plausible_web/router.ex
```diff
@@ -217,20 +228,25 @@ defmodule PlausibleWeb.Router do
   end
 
   scope "/api", PlausibleWeb do
-    pipe_through :api
+    scope assigns: %{} do
```
aerosol, line 231: what's the purpose of empty assigns initialization here?

lib/plausible_web/user_auth.ex
```diff
@@ -133,24 +173,91 @@ defmodule PlausibleWeb.UserAuth do
 
   defp get_user_token(session) do
     case Enum.map(["user_token", "current_user_id"], &Map.get(session, &1)) do
```
aerosol, line 175: Why do that if session keys can be directly matched?

lib/plausible_web/user_auth.ex
```diff
@@ -133,24 +173,91 @@ defmodule PlausibleWeb.UserAuth do
...
+  @unknown_label "Unknown"
+
+  defp get_device_name(%Plug.Conn{} = conn) do
+    conn
+    |> Plug.Conn.get_req_header("user-agent")
+    |> List.first()
+    |> get_device_name()
+  end
+
+  defp get_device_name(user_agent) when is_binary(user_agent) do
```
aerosol, line 215: What's the purpose of this? Can't we store UA as is? We don't seem to be querying by device.

aerosol (commented):
Good work! Keeping legacy sessions alive is a really nice touch.

lib/plausible_web/plugs/user_session_touch.ex
zoldar: Yeah, good idea. I'll see if I can limit updates to once a day at most.

lib/plausible_web/router.ex
zoldar: None, but `scope` expects non-empty opts. Could be `[]` just as well, probably. I'll adjust it.

lib/plausible_web/user_auth.ex
zoldar: This will be handly once we implement listing active sessions in user settings. The idea is giving user at least a general idea where a given session is coming from (computer, phone, tablet) so that they can manually invalidate it if they need to / want to.

lib/plausible_web/user_auth.ex
zoldar: Yeah, that was me brainfarting, fixed in https://github.com/plausible/analytics/pull/4463/commits/2080c4da57929e9419b1c290563add15758586d5, thanks 😄

lib/plausible_web/router.ex
zoldar: https://github.com/plausible/analytics/pull/4463/commits/2bb234c651726da4f85c947b735af7e79cf4c2c3

lib/plausible_web/plugs/user_session_touch.ex
zoldar: Ah, hold on, but we still probably want to keep `last_used_at` up to date. Showing last time the session was active is pretty common and is an extra data point to tell sessions apart when reviewing them (in addition to the device/os label/name). I could limit it to once a day as well, but then we'd could only display that time with day resolution. I'm not sure about what's better.

ukutaht: We actually already keep track of this actually and use it!

See: https://github.com/plausible/analytics/blob/master/lib/plausible_web/plugs/last_seen_plug.ex
and: https://github.com/plausible/analytics/blob/19ecd3d0ee6ecb8d801356034ccc37fc6637760f/lib/workers/send_check_stats_emails.ex#L15

I think the current 'at most once an hour' update logic is sound. It keeps the timestamp relatively up to date while limiting UPDATE queries to no more than once per hour. I'm proposing to port the same logic over to `UserSession` and remove `last_seen` from the `User` context.

zoldar: I'm not sure about removing `last_seen` entirely. After the 2 week period all the users active during that period will have their sessions converted. All the others though won't. Now if we simply drop `last_seen`, they will appear as if they have never logged in. Perhaps we could leave `user.last_seen` and bump it together with last active token's `last_used_at` (while getting rid of this legacy plug)? Given update frequency will be limited to at most once per hour, that shouldn't be a big deal.

ukutaht: Fair point about historical data.

> Perhaps we could leave user.last_seen and bump it together with last active token's last_used_at (while getting rid of this legacy plug)?

Yeah, I think that's a good solution!

zoldar: Done in https://github.com/plausible/analytics/pull/4463/commits/2e856a7eee1e73b10b741dac1a58c2da281516a7

zoldar:
@aerosol @ukutaht all feedback so far applied and PR ready for re-review.

ukutaht (approved):

lib/plausible_web/user_auth.ex
aerosol: OK cool, that makes sense, thanks!

aerosol (approved):

PR #4464, opened 2024-08-23
APIv2: JSON schema validation, separate internal and public API validation

macobo:
This PR:
- Adds JSON schema support to this repo
- Separates internal and public API validation

Note that the JSON schema in this repo will start to be copied over to /docs repo with upcoming changes.


lib/plausible/stats/json_schema.ex
```diff
@@ -0,0 +1,67 @@
...
+                           "matches"
+                         )
+                         |> JSONPointer.add!(
+                           "#/definitions/filter_entry/oneOf/0/items/0/enum/0",
+                           "does_not_match"
+                         )
+                         |> JSONPointer.add!("#/definitions/metric/oneOf/0", %{
+                           "const" => "time_on_page"
+                         })
+                         |> ExJsonSchema.Schema.resolve()
```
apata, line 30: Is this bit covered by any tests? This is maybe too paranoid, but I think we could future-proof this module a bit by adding a test to assert the exact value of both the internal and external schemas, since so much depends on them.

test/plausible_web/controllers/api/external_stats_controller/query_validations_test.exs
```diff
@@ -227,7 +230,7 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryValidationsTest do
         })
 
       assert json_response(conn, 400) == %{
-               "error" => "Metrics cannot be queried multiple times"
+               "error" => "#/metrics: Expected items to be unique but they were not."
```
apata, line 233: **nitpick, non-blocking**: Other errors in these tests don't end with sentence delimiter `.`. Since the others are reworded, perhaps we can reword them so they end with `.` for the sake of consistency?

RobertJoonas (approved):
Looks good to me 👍

lib/plausible/stats/json_schema.ex
macobo: Yes, it is: https://github.com/plausible/analytics/pull/4464/files#diff-3cdb27f0b3c7856a6665e78f3d6d7f60e39c2aa5a3a4f98fc3729cf334975c77R170-R186 and https://github.com/plausible/analytics/pull/4464/files#diff-3cdb27f0b3c7856a6665e78f3d6d7f60e39c2aa5a3a4f98fc3729cf334975c77R100-R108

test/plausible_web/controllers/api/external_stats_controller/query_validations_test.exs
macobo: Good idea, done!

lib/plausible/stats/json_schema.ex
apata: True, it's important to assert that the schema is actually applied by the backend. I just thought that it'd be good to verify that the two schemas are precisely what we expect them to be. 

I'm anticipating that we may want to change how we apply these overrides or which overrides we apply. 

If we have asserts in place on the exact output, we'll have even less to worry about. 

For example, imagine updating the JSONPointer 0.x package.

PR #4466, opened 2024-08-24
Run migrations in order across repos

ruslandoga:
### Changes

This PR makes [`/entrypoint.sh db migrate`](https://github.com/plausible/community-edition/blob/4e949e84f787702835ec0654c7f663dc2fb61dbe/docker-compose.yml#L28C63-L28C88) and `/entrypoint.sh db pending-migrations` order all pending migrations and run them in a single queue instead of per-repo.

Related card: https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7508896723

<details><summary>Contents of that card</summary>

<br/>

> Some migrations in PostgreSQL implicitly depend on migrations in ClickHouse (there is at least one such migration), and because PostgreSQL migrations run first, they fail because from their point of view ClickHouse schemas are now "outdated":
> 
> How migrations were added (example):
> - add table to PostgreSQL db
> - add column to ClickHouse table
> - run PostgreSQL data migration that reads from new column in ClickHouse
> 
> How migrations are run:
> - run all Plausible.Repo migrations
> - run all Plausible.IngestRepo migrations
> 
> Which results in this broken order:
> - add table to PostgreSQL db
> - run PostgreSQL data migration that reads from new column in ClickHouse <- fails
> - add column to ClickHouse table <- never runs
> 
> It might be possible to run them in a single queue / order in `db migrate` command:
> - collect all migrations to be run, order them by version / date
> - start all repos (Repo, IngestRepo, ClickhouseRepo)
> - execute them one by one

</details>

### Tests
- [x] Automated tests have been added (only for pending migrations listings, I'll try to add an actual `db migrate` test later)

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_release.ex
+++ b/lib/plausible_release.ex
@@ -25,6 +25,39 @@ defmodule Plausible.Release do
     IO.puts("Migrations successful!")
   end
 
+  # Unlike `migrate/0` above this function lists *all* pending migrations
+  # across all repos, sorts them into a single liss and only then runs them.
+  # This approach helps resolve dependencies between migrations across repos.
+  def interweave_sort_migrate do
+    prepare()
+
+    # interweave
+    all_pending =
+      Enum.flat_map(repos(), fn repo ->
+        Ecto.Migrator.migrations(repo)
+        |> Enum.filter(fn {status, _version, _name} -> status == :down end)
+        |> Enum.map(fn {_status, version, name} -> {repo, _migration = {version, name}} end)
+      end)
+
+    # sort
+    all_sorted =
+      Enum.sort_by(all_pending, fn {_repo, migration} -> migration end, :asc)
+
+    # migrate
+    Enum.each(all_sorted, fn {repo, {version, name}} ->
+      [{module, _binary}] =
+        Code.compile_file(
+          "#{version}_#{name}.exs",
+          _relative_to = Ecto.Migrator.migrations_path(repo)
+        )
+
+      {:ok, _, _} =
+        Ecto.Migrator.with_repo(repo, fn repo ->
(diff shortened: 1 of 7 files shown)


lib/plausible_release.ex
```diff
@@ -25,6 +25,54 @@ defmodule Plausible.Release do
     IO.puts("Migrations successful!")
   end
 
+  # Unlike `migrate/0` above this function:
```
ukutaht, line 28: I think we should add documentation for why we run migrations differently from the 'default' way.

ruslandoga: Done in https://github.com/plausible/analytics/pull/4466/commits/818401c4266f2beb6a6a1ce1f9bac12624d36dbb

ruslandoga: Done in https://github.com/plausible/analytics/pull/4466/commits/3d2e060c991fcdafb713dea8432040de1cf520cb

lib/plausible_release.ex
```diff
@@ -25,13 +25,127 @@ defmodule Plausible.Release do
...
+      pending
+    end)
+  end
+
   def pending_migrations do
     prepare()
     IO.puts("Pending migrations")
     IO.puts("")
     Enum.each(repos(), &list_pending_migrations_for/1)
   end
```
zoldar, line 117: I think we can get rid of this function now as it's unused (along with private function used by it).

zoldar: Yes! 🔥 No point in keeping it around IMHO.

ruslandoga: https://github.com/plausible/analytics/pull/4466/commits/de576a363b971365ef9393fcacd871c8bbbe7596

zoldar (approved):

PR #4469, opened 2024-08-27
Support realtime periods in API v2

RobertJoonas:
### Changes

* Parse query date ranges into a new `DateTimeRange` struct, including the timezone information.
* support `realtime` and `30m` date range shortcuts in the private API JSON schema
* make it possible to pass a `date` parameter (in the private API JSON schema) relative to which the `date_range` will be constructed.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/naive_datetime_range.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Plausible.Stats.NaiveDateTimeRange do
```
macobo, line 1: NaiveDateTimeRange is a mouthful and I can see myself getting confused by it vs `NaiveDateTime`. It also isn't clear what the naive part of it is.

Would `DateTimeRange` make just as much sense?

lib/plausible/stats/query_result.ex
```diff
@@ -24,6 +23,8 @@ defmodule Plausible.Stats.QueryResult do
         }
       end)
 
+    %{first: first, last: last} = NaiveDateTimeRange.to_date_range(query.date_range)
```
macobo, line 26: This is wrong right because realtime wouldn't be serialized as expected (IS08601 timestamp)

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -1,23 +1,20 @@
...
-  alias Plausible.Stats.JSONSchema
+  alias Plausible.Stats.{TableDecider, Filters, Query, Metrics, NaiveDateTimeRange, JSONSchema}
 
   @default_include %{
     imports: false,
     time_labels: false
   }
 
-  def parse(site, schema_type, params, now \\ nil) when is_map(params) do
+  def parse(site, schema_type, params, now_for_site \\ nil) when is_map(params) do
```
macobo, line 11: What is `now_for_site` and how is it different from `now`?

test/plausible/stats/query_parser_test.exs
```diff
@@ -478,33 +548,43 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+      check_date_range("12mo", site, @date_range_12mo)
+      check_date_range("year", site, @date_range_year)
+    end
+
+    test "parsing private internal API shortcut options for date_range", %{site: site} do
+      check_date_range("30m", site, @date_range_30m, :internal)
+      check_date_range("realtime", site, @date_range_realtime, :internal)
+    end
+
+    test "private internal API shortcut date_ranges are unavailable in the public schema", %{
```
macobo, line 566:
```suggestion
    test "30m and realtime date_ranges are unavailable in public API", %{
```

test/plausible/stats/query_parser_test.exs
```diff
@@ -478,33 +548,43 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+      check_date_range("day", site, @date_range_day)
+      check_date_range("7d", site, @date_range_7d)
+      check_date_range("30d", site, @date_range_30d)
+      check_date_range("month", site, @date_range_month)
+      check_date_range("6mo", site, @date_range_6mo)
+      check_date_range("12mo", site, @date_range_12mo)
+      check_date_range("year", site, @date_range_year)
+    end
+
+    test "parsing private internal API shortcut options for date_range", %{site: site} do
```
macobo, line 561:
```suggestion
    test "30m and realtime are available in internal API", %{site: site} do
```

test/plausible/stats/query_parser_test.exs
```diff
@@ -478,33 +548,43 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
   describe "date range validation" do
-    test "parsing shortcut options", %{site: site} do
-      check_date_range("day", site, Date.range(~D[2021-05-05], ~D[2021-05-05]))
-      check_date_range("7d", site, Date.range(~D[2021-04-29], ~D[2021-05-05]))
-      check_date_range("30d", site, Date.range(~D[2021-04-05], ~D[2021-05-05]))
-      check_date_range("month", site, Date.range(~D[2021-05-01], ~D[2021-05-31]))
-      check_date_range("6mo", site, Date.range(~D[2020-12-01], ~D[2021-05-31]))
-      check_date_range("12mo", site, Date.range(~D[2020-06-01], ~D[2021-05-31]))
-      check_date_range("year", site, Date.range(~D[2021-01-01], ~D[2021-12-31]))
+    test "parsing public shortcut options for date_range", %{site: site} do
```
macobo, line 551:
```suggestion
    test "parsing shortcut options", %{site: site} do
```

test/plausible/stats/query_parser_test.exs
```diff
@@ -514,6 +594,27 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+      check_date_range_with_date("day", site, @date_range_day)
+      check_date_range_with_date("7d", site, @date_range_7d)
+      check_date_range_with_date("30d", site, @date_range_30d)
+      check_date_range_with_date("month", site, @date_range_month)
+      check_date_range_with_date("6mo", site, @date_range_6mo)
+      check_date_range_with_date("12mo", site, @date_range_12mo)
+      check_date_range_with_date("year", site, @date_range_year)
+    end
+
+    test "date parameter is not available in the public API schema", %{site: site} do
```
macobo, line 608:
```suggestion
    test "date parameter is not available in the public API", %{site: site} do
```

test/plausible/stats/query_parser_test.exs
```diff
@@ -514,6 +594,27 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+      params = %{
+        "site_id" => site.domain,
+        "metrics" => ["visitors", "events"],
+        "date_range" => "month",
+        "date" => "2021-05-05"
+      }
+
+      check_error(params, site, "#/date: Schema does not allow additional properties.")
```
macobo, line 616: Nit: Let's keep a consistent test structure in the file, something like:

```elixir
check_error(params = %{
        "site_id" => site.domain,
        "metrics" => ["visitors", "events"],
        "date_range" => "month",
        "date" => "2021-05-05"
      }, site, "#/date: Schema does not allow additional properties.")
```

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -2912,11 +2917,11 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
       })
 
     assert json_response(conn, 200)["results"] == [
-             %{"dimensions" => ["2021-01-01 00:00:00", "Google"], "metrics" => [1]},
-             %{"dimensions" => ["2021-01-02 00:00:00", "Google"], "metrics" => [1]},
-             %{"dimensions" => ["2021-01-02 00:00:00", "Direct / None"], "metrics" => [1]},
-             %{"dimensions" => ["2021-01-03 00:00:00", "Direct / None"], "metrics" => [1]},
-             %{"dimensions" => ["2021-01-03 00:00:00", "Twitter"], "metrics" => [1]}
+             %{"dimensions" => ["2021-01-01", "Google"], "metrics" => [1]},
```
macobo, line 2920: Q: What changed here? This seems unrelated to everything else.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -104,16 +104,10 @@ defmodule PlausibleWeb.Api.StatsController do
     with {:ok, dates} <- parse_date_params(params),
          :ok <- validate_interval(params),
          :ok <- validate_interval_granularity(site, params, dates),
+         params <- realtime_period_to_30m(params),
          query = Query.from(site, params, debug_metadata(conn)),
```
macobo, line 108: In the end we're still using Query.from instead of Query.build, meaning that all the parser changes are unused anywhere. What's the plan there?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -217,21 +215,33 @@ defmodule PlausibleWeb.Api.StatsController do
...
       includes_imported: includes_imported?(query, comparison_query),
       imports_exist: site.complete_import_ids != [],
-      comparing_from: comparison_query && comparison_query.date_range.first,
-      comparing_to: comparison_query && comparison_query.date_range.last,
-      from: query.date_range.first,
-      to: query.date_range.last
+      comparing_from: comparing_from,
+      comparing_to: comparing_to,
+      from: Date.to_string(first_date),
+      to: Date.to_string(last_date)
```
macobo, line 240: I think we can avoid the assignment logic above and do something like:

```elixir
comparing_from: comparison_query && comparison_query.date_range.first |> Date.to_date
...
to: query.date_range |> Date.to_date
...
```

Is this not the case?

test/plausible/stats/query_parser_test.exs
```diff
@@ -1,32 +1,93 @@
...
+  @date_range_year %NaiveDateTimeRange{
+    first: ~N[2021-01-01 00:00:00],
+    last: ~N[2022-01-01 00:00:00]
+  }
+  @date_range_12mo %NaiveDateTimeRange{
+    first: ~N[2020-06-01 00:00:00],
+    last: ~N[2021-06-01 00:00:00]
+  }
+
+  def check_success(params, site, expected_result, schema_type \\ :public) do
```
macobo, line 48: Nit: I think having public/internal as a last argument is not ideal (because it muddles the assertion reading) but not an important issue.

lib/plausible/stats/comparisons.ex
```diff
@@ -61,52 +61,55 @@ defmodule Plausible.Stats.Comparisons do
       |> Keyword.put_new(:now, Timex.now(site.timezone))
       |> Keyword.put_new(:match_day_of_week?, false)
 
+    source_date_range = NaiveDateTimeRange.to_date_range(source_query.date_range)
```
macobo, line 64: You can't do comparisons over 30m/realtime? If so, this likely warrants a comment somewhere.

lib/plausible/stats/naive_datetime_range.ex
RobertJoonas: I've called it `NaiveDateTimeRange` as it's working with `NaiveDateTime` structs rather than `DateTime`.

The `date_range` in the query is always in `site.timezone` - however, we don't really care about what timezone it's in until we start constructing the `utc_boundaries`. Using `NaiveDateTime` makes that clear, so it's easier to work with shifting and comparing datetimes, without having to worry about timezone conversions and such.

> Would DateTimeRange make just as much sense?

Not sure if you're suggesting to stick with using `NaiveDateTime` under the hood and just call the module differently? To me it makes it look like it consists of `DateTime` structs, which actually expect a timezone to be created in the first place.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: I've just called it differently to make it a bit clearer that the `now` we're passing there is in `site.timezone` - not UTC.

macobo: Does it make a practical difference for the understanding of the code? If not, I'd name it `now` - less overhead for the reader of the code.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
RobertJoonas: 1. How we resolve time dimension is now working with `NaiveDateTime` structs instead of `Date`. Before, it got resolved to `time:hour` because:

```elixir
Timex.diff(~D[2021-01-03], ~D[2021-01-01], :hours) == 48
```

The actual datetime range that we end up querying with these dates is:

```elixir
%NaiveDateTimeRange{first: ~N[2021-01-01 00:00:00], last: ~N[2021-01-04 00:00:00]}
```

And this period is 72 hours, hence the `time` dimension gets resolved into `time:day` with the new code, which I believe is the expected behaviour?

2. I've also duplicated some of the events in order to make the returned order consistent. There are two sorting keys in this case - `time` and `metrics`. If both are equal, then the results are not guaranteed to have a consistent order. Maybe would make sense to add an alphabetical order to the returned non-time dimensions as well?

lib/plausible/stats/query_result.ex
RobertJoonas: Good point! how about **always** returning `date_range` as an ISO8601 timestamp?

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: The new API endpoints (such as `POST /:domain/main-graph-v2`) will start using `Query.build`. Will be introduced in a follow-up PR.

test/plausible/stats/query_parser_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/4469/commits/0876cbf3e5b165d990c02e8dc6707ae6e1451889

lib/plausible/stats/naive_datetime_range.ex
apata: Barging in here as I'm trying to understand the end to end logic for datetimes within the application. 

As per the [NaiveDateTime module docs](https://hexdocs.pm/elixir/NaiveDate)
> datetime may not actually exist in certain areas in the world even though it is valid.

Apart from the name, _can_ we actually use NaiveDateTime here?

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: The `NaiveDateTimeRange.to_date_range` shifts back a day in the last date if the time is at `~T[00:00:00]`.

For example, let's say that we're querying a `day` with a `date` of `2021-01-01`. This would be parsed into the following range:

```elixir
%NaiveDateTimeRange{first: ~N[2021-01-01 00:00:00], last: ~N[2021-01-02 00:00:00]}
```

Now if we just convert both first and last to a date (and then to string), we'll get:

```elixir
%{
  # ...
  from: "2021-01-01",
  to: "2021-01-02",
  # ...
}
```

which is wrong.

lib/plausible/stats/filters/query_parser.ex
apata: Is it a point in time (_a particular microsecond from Unix epoch_) or the label of that point in time in a particular locale (_twelve o'clock 27th August Tallinn / nine o'clock 27th August London_)? 

If it's a point in time, now === now for site.

It it's a locale sensitive label, why is it a label?

lib/plausible_web/controllers/api/stats_controller.ex
macobo: Q: Why not change this existing endpoint?

lib/plausible/stats/query_result.ex
macobo: The idea is that the returned query should (in theory) be usable as a input query.

So for dates I'd still return ISO8601 date, for timestamps ISO8601 timestamps (this will be internal only). Would that work?

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
macobo: Aha. So this is fixing a bug - previously deciding how many hours in range was incorrect.

Perhaps this also warrants a new test that shows the broken edge case in time_test.exs?

lib/plausible/stats/naive_datetime_range.ex
RobertJoonas: I've now changed it to work with DateTimes instead, which always include the timezone information.

A notable change in this is also that the last `DateTime` in the `DateTimeRange` is now the last `date` at 23:59:59 (instead of 00:00:00 on the next day). To make that work the same way as before, the ClickHouse queries are now returning events/sessions where `timestamp <= last_datetime` (intead of just `<`).

This change makes it easier for us to:

1. turn the last DateTime back into a Date when needed (e.g for comparisons). Just call `DateTime.to_date` instead of shifting the day back when the time is at 00:00:00
2. parse the last date into a datetime in `site.timezone`. For example, `2022-09-11T00:00:00` doesn't exist in `America/Santiago`. Setting that as the last datetime, we'd have to either move it to 01:00:00 on that day or 23:59:59 on the previous day. Converting that back into a date that was actually requested as `last` would become tricky. On the other hand, if last datetimes are always set to 23:59:59 on the last `date`. This will not be a problem or exception that we'll have to handle. Even if it doesn't exist in the timezone (which it, AFAIK, always does), we can shift the last datetime back and it will still remain the same `date`.

lib/plausible/stats/query_result.ex
RobertJoonas: I've now implemented this as discussed in our call.

The `query_result.date_range` will become a "timezoned" iso8601 timestamp, which does not follow any official standard (because there doesn't exist one), but solves the problem for us. The format is `"2023-03-03T00:00:00 Europe/Tallinn"`

The iso8601 timestamp does not include the UTC offset and is a "naive" timestamp. The timezone name is separated with a space which we can use to parse the DateTime into the exact timezone. (UTC offset alone doesn't help us here).

> The idea is that the returned query should (in theory) be usable as a input query.

This now works. I've also included a test for this in `ExternalStatsController.QueryTest` which takes the query and passes it as the body into another request, yielding the same results.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Named it back to `now` as it is now expected to be in UTC anyway.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: I'm planning to do a feature flag switch, eventually replacing the old endpoint with the new one.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Update: implemented your suggestion as it's now easier once we're parsing last datetime into last date on 23:59:59. https://github.com/plausible/analytics/pull/4469/commits/676b54cd2a80839151e3afb553731f1c77ad3b05

lib/plausible/stats/comparisons.ex
RobertJoonas: Added a comment in https://github.com/plausible/analytics/pull/4469/commits/0ec0ceb5357a1a39d5122151d1c3943c51ed3a96.

priv/json-schemas/query-api-schema.json
```diff
@@ -99,13 +99,17 @@
           "type": "array",
           "items": {
             "type": "string",
-            "pattern": "^\\d{4}-\\d{2}-\\d{2}$"
+            "pattern": "^\\d{4}-\\d{2}-\\d{2}(?:T\\d{2}:\\d{2}:\\d{2}\\s[A-Za-z/_]+)?$"
```
RobertJoonas, line 102: This pattern does currently not allow the standard `2023-03-03T00:00:00Z` format. Instead we expect timezone to be explicitly separated with a space.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -1,23 +1,23 @@
...
 
   @default_include %{
     imports: false,
     time_labels: false
   }
 
-  def parse(site, schema_type, params, now \\ nil) when is_map(params) do
+  def parse(site, schema_type, params, test_opts \\ []) when is_map(params) do
+    now = Keyword.get(test_opts, :now, DateTime.utc_now(:second))
+    date = Keyword.get(test_opts, :date, today(site))
```
macobo, line 13: Q: Why do we need both date and now parameters? Isn't `date = now |> to_date`?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -1,23 +1,23 @@
...
+
     with :ok <- JSONSchema.validate(schema_type, params),
-         {:ok, date} <- parse_date(site, Map.get(params, "date"), now),
+         now <- if(now, do: now, else: DateTime.utc_now(:second)),
```
macobo, line 16: I believe this if check can never fail right?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -110,84 +110,121 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+    with {:ok, from_datetime} <- datetime_from_timestamp(from),
+         {:ok, to_datetime} <- datetime_from_timestamp(to) do
+      {:ok, DateTimeRange.new!(from_datetime, to_datetime)}
+    else
+      _ -> {:error, "Invalid date_range '#{i([from, to])}'."}
+    end
+  end
+
+  defp datetime_from_timestamp(timestamp_string) do
+    with [timestamp, timezone] <- String.split(timestamp_string),
```
macobo, line 219: Can we avoid cooking our encoding scheme and use DateTime.from_iso8601 instead?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -110,84 +110,121 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+         {:ok, naive_datetime} <- NaiveDateTime.from_iso8601(timestamp),
+         {:ok, datetime} <- DateTime.from_naive(naive_datetime, timezone) do
+      {:ok, datetime}
+    else
+      _ -> {:error, :invalid}
```
macobo, line 224: Confusing - Can this error reach users? It's not formatted like any other error here.

If not - just dump the else branch.

lib/plausible_web/controllers/api/stats_controller.ex
macobo: I'd take a more eager approach here but that works for me. :+1:

test/plausible/stats/query_parser_test.exs
```diff
@@ -1,32 +1,92 @@
 defmodule Plausible.Stats.Filters.QueryParserTest do
```
macobo, line 1: This will have some unfortunate conflicts with https://github.com/plausible/analytics/pull/4480

test/plausible/stats/query_parser_test.exs
```diff
@@ -513,6 +612,121 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+        "site_id" => site.domain,
+        "date_range" => ["2021-02-03T00:00:00 UTC", "2021-02-04"],
+        "metrics" => ["visitors"]
+      }
+      |> check_error(
+        site,
+        "Invalid date_range '[\"2021-02-03T00:00:00 UTC\", \"2021-02-04\"]'."
+      )
+
+      # Datetime doesn't exist due to a gap in the given timezone
```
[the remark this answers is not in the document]
macobo: `Invalid date_range` seems fine.

test/plausible/stats/query_parser_test.exs
```diff
@@ -513,6 +612,121 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
 
       %{"site_id" => site.domain, "date_range" => ["21415-00", "eee"], "metrics" => ["visitors"]}
       |> check_error(site, "#/date_range: Invalid date range [\"21415-00\", \"eee\"]")
+
+      # Timestamps do not include timezone information
```
macobo, line 616: There's too many test cases in this test case.

Let's create separate test cases.

test/plausible/stats/query_parser_test.exs
```diff
@@ -1,32 +1,92 @@
...
+        include: %{imports: false, time_labels: false},
+        preloaded_goals: []
+      },
+      schema_type
+    )
+  end
+
+  # Checks parsing the date range relative to the `date` parameter (which is only
+  # available in the internal API schema) instead of using `@now`.
+  def check_date_range_with_date(date_range, site, expected_date_range) do
```
macobo, line 81: Nit: Can we make `date` an optional parameter of `check_date_range` instead?

E.g. `def check_date_range(date_params, site, expected_date_range, schema_type \\ public)

This reduces the number of abstractions we need to deal with reading tests.

lib/plausible/stats/query_result.ex
```diff
@@ -79,6 +81,15 @@ defmodule Plausible.Stats.QueryResult do
     |> Enum.reject(fn {_, value} -> is_nil(value) end)
     |> Enum.into(%{})
   end
+
+  defp to_iso_8601_with_timezone(%DateTime{time_zone: timezone} = datetime) do
```
macobo, line 85: Q: would `DateTime.to_iso8601(datetime, :extended)` not work here? This should make the parsing more standard as well.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Unfortunately not. The ISO8601 format alone does not provide the timezone information which we need to create the `DateTime` struct.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Good catch! This was just supposed to catch failures like `{:gap, date1, date2}`, but we can just dump the else branch indeed 👍

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: You're right, we can work with only `now`. https://github.com/plausible/analytics/pull/4469/commits/d3a9035ba634b7f7a858eb897b25b40e8accebc5

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: https://github.com/plausible/analytics/pull/4469/commits/d04bca5bb24fc66c6ca350ed2841582383cbd6f0

test/plausible/stats/query_parser_test.exs
RobertJoonas: Noted!

lib/plausible/stats/query_result.ex
RobertJoonas: This would include the UTC offset in the iso8601 string format, but not the timezone name that we need.

test/plausible/stats/query_parser_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/4469/commits/23e3160e8bcf32fac0440f9bec13972b3aeb526f

test/plausible/stats/query_parser_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/4469/commits/5c9bc6340f251f3fb07fb64ecd20cea9edf47c87

macobo (approved):
:1st_place_medal:

lib/plausible/stats/datetime_range.ex
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+  @doc """
+  Creates a `DateTimeRange` struct from the given `%Date{}` structs.
+
+  The first datetime will become the first date at 00:00:00, and the last datetime
+  will become the last date at 23:59:59. Both dates will be turned into `%DateTime{}`
+  structs in the given timezone.
+  """
+  def new!(%Date{} = first, %Date{} = last, timezone) do
```
ruslandoga, line 23: This seems to be prone to performing timezone conversion multiple times (stats_start_date -> native_stats_start_date (first conversion) -> new! (seconds conversion)). Would it be safer to always require DateTimes?

lib/plausible/stats/datetime_range.ex
```diff
@@ -0,0 +1,54 @@
...
+        {:ok, datetime} -> datetime
+        {:gap, _just_before, just_after} -> just_after
+        {:ambiguous, _first_datetime, second_datetime} -> second_datetime
+      end
+
+    last =
+      case DateTime.new(last, ~T[23:59:59], timezone) do
+        {:ok, datetime} -> datetime
+        {:gap, just_before, _just_after} -> just_before
+        {:ambiguous, first_datetime, _second_datetime} -> first_datetime
```
ruslandoga, line 35: Just curious: I wonder why lower values are picked for end of the range, and higher -- for the beginning? This makes the whole range smaller, potentially leaving some events out.

lib/plausible/stats/datetime_range.ex
ruslandoga: False alarm. Sorry! 🙏

lib/plausible/stats/datetime_range.ex
RobertJoonas: @ruslandoga, good question! The idea was to make sure that the date of the parsed datetime will always remain the same as the requested date.

One specific example: let's say the client requests a `date_range` of `["2022-09-11", "2022-09-11"]` and their timezone is `America/Santiago`. The first datetime will be parsed as follows:

```
iex(3)> DateTime.new(~D[2022-09-11], ~T[00:00:00], "America/Santiago")
{:gap, #DateTime<2022-09-10 23:59:59.999999-04:00 -04 America/Santiago>,
 #DateTime<2022-09-11 01:00:00-03:00 -03 America/Santiago>}
```

If we choose the first date (i.e. "just before"), we'll end up parsing the datetime into a different date ("2022-09-10") that was actually requested ("2022-09-11").

That means, when dealing with the date range later on in code (i.e. constructing the time labels for timeseries requests or doing comparisons), we'd run into errors, or we'd have to build some ugly exceptions to handle these cases.

That said, when allowing a flexible datetime format in the public API, where one could request a date range such that `2022-09-11 00:00:00` in "America/Santiago" is the last datetime, we'd still run into the same issue.

So perhaps the proper fix could be to always select `just_before` or `just_after` depending on which falls on the same date? \cc @macobo also bringing this to your attention as you're working on it.

ruslandoga: Ah, I see. I didn't think about conversion to dates, I assumed the filtering always happens on "full" timestamps. And I didn't know that there were timezones with gaps at midnight! Thank you for the explanation!

PR #4471, opened 2024-08-27
Add tests for FE date calculations, fix niggles with site context

apata:
**Hypothesis:** To correctly perform any date calculations, one needs to know the timezone: current offset is not enough, since the calculation may span the time when the offset changes. 

However, when needing only date level precision, assuming the BE returns the offset as expected, I couldn't find any issues. 

_(The first commit seemed to verify the hypothesis, but it had an incorrect expectation.)_


RobertJoonas (approved):
Thanks for this! LGTM 👍

PR #4476, opened 2024-08-28
Add better assets compression to CE

ruslandoga:
### Changes

This PR enables static assets compression in CE. And additionally adds a library with better Gzip and Brotli defaults.

Relevant card: https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7559344001

> Right now CE sends static assets uncompressed (unless the reverse proxy compresses them).
> 
> https://github.com/hauleth/phoenix_bakery digests statis assets (on mix phx.digest) to (better compressed) gzip and also brotli so that Plug.Static :gzip and :brotli options could be set to true which would make it serve well compressed static assets. This would be useful once we start removing reverse-proxies from the default CE setup.

### Tests

### Changelog

### Documentation

### Dark mode


aerosol (approved):

Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN mkdir /app
 WORKDIR /app
 
 # install build dependencies
-RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev && \
+RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli zstd && \
```
ruslandoga, line 23: Oops, actually `zstd` is not used yet. I'll need to remove it.

ruslandoga: Removed.

ukutaht (approved):

zoldar (approved):

PR #4480, opened 2024-08-28
APIv2: and/or/not support

macobo:
### Changes

While working on regex support I went down a rabbit hole and experimented with adding `and`/`or`/`not` support to the query API.

It was remarkably straight-forward, if applying the following limitations:
- `event:goal` filters can only be applied at the top level (due to business logic relating to goals)
- `event:hostname` filters can only be applied at the top level (due to business logic tracking visits across domains)
- When using `conversion_rate` and `group_conversion_rate`, custom properties can only be used at the top level

Outside of these limitations, other changes were straight-forward modifications to support nested structures:
- Filter manipulation and finding filters was moved outside of Query module 
- Ecto query building was updated
- JSON schemas were likewise modified 

Note there is no plan (yet) to support all this in the dashboard and only in query api v2


priv/json-schemas/query-api-schema.json
```diff
@@ -259,9 +259,7 @@
             {
               "$ref": "#/definitions/clauses"
             }
-          ],
-          "minItems": 3,
```
macobo, line 263: Might seem like an odd change, but without this autocompletion in monaco starts failing :(

Invalid filters don't pass parsing regardless of this.

apata: Super impressive PR! Couldn't resist checking it out and helping with what I could. l gave all the possible filter tree items in the schema `minItems` and `maxItems`, and that was enough to get the editor start offering autocomplete again. Find the code here: 
https://github.com/plausible/analytics/compare/and-or-not...and-or-not--item-limits?expand=1

macobo: That looks great! :heart: 

I propose we get this, the follow-up PR and @RobertJoonas's PR in in and then PR the refactored json schema afterwards to avoid merge and review pain.

Does that sound reasonable?

lib/plausible/stats/query.ex
```diff
@@ -141,11 +129,12 @@ defmodule Plausible.Stats.Query do
           :ok | {:error, :no_imported_data | :out_of_range | :unsupported_query | :not_requested}
   def ensure_include_imported(query, requested?) do
     cond do
+      not requested? -> {:error, :not_requested}
       is_nil(query.latest_import_end_date) -> {:error, :no_imported_data}
       query.period in ["realtime", "30m"] -> {:error, :unsupported_query}
       Date.after?(query.date_range.first, query.latest_import_end_date) -> {:error, :out_of_range}
       not Imported.schema_supports_query?(query) -> {:error, :unsupported_query}
-      not requested? -> {:error, :not_requested}
+      query.period == "realtime" -> {:error, :unsupported_query}
```
RobertJoonas, line 137: `query.period` is checked against "realtime" and "30m" on line 134 already.

```suggestion
```

lib/plausible/stats/query_optimizer.ex
```diff
@@ -152,10 +153,9 @@ defmodule Plausible.Stats.QueryOptimizer do
 
     filters =
       if "event:page" in query.dimensions do
-        query.filters
-        |> Enum.map(fn
-          [op, "event:page" | rest] -> [op, "visit:entry_page" | rest]
-          filter -> filter
+        Filters.transform_filters(query.filters, fn
```
RobertJoonas, line 156: I'm seeing that this is the only use of this function. The purpose is to keep the same filter but only override the `filter_key`.

IMO would be easier to understand the intent here with something like:

```elixir
Filters.transform_filter_keys(query.filters, %{"event:page" => "visit:entry_page"})
```

That way the caller wouldn't have to worry about returning `nil` to look deeper into the nested structures, etc. The function would do that by default. WDYT?

test/plausible/stats/query_parser_test.exs
```diff
@@ -917,7 +1060,7 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
         "site_id" => site.domain,
         "metrics" => ["visitors"],
         "date_range" => "all",
-        "filters" => [["is", "event:props:foobar", ["foo"]]]
+        "filters" => [["not", ["is", "event:props:foobar", ["foo"]]]]
```
RobertJoonas, line 1063: What's the purpose of this change? The test description currently gives a wrong idea about what's actually happening here.

lib/plausible/stats/query.ex
```diff
@@ -76,7 +76,13 @@ defmodule Plausible.Stats.Query do
...
   end
 
-  def remove_filters(query, prefixes) do
+  @doc """
+  Removes top level filters matching any of passed prefix from the query.
+
+  Note that this doesn't handle cases with AND/OR/NOT and as such is discouraged
+  from use.
+  """
+  def remove_top_level_filters(query, prefixes) do
```
ukutaht, line 85: Good candidate to move to `Filters` module as well

ukutaht (approved):

test/plausible/stats/query_parser_test.exs
macobo: Making sure that the validation still works for nested filters.

lib/plausible/stats/query.ex
macobo: Reason I kept it here is to avoid having the Filters module return queries and re-constructing filters object can be a pain.

test/plausible/stats/query_parser_test.exs
RobertJoonas: The filter that we're using is actually valid, but the test description claims it's not. The real error when parsing this query is that the site owner does not have access to custom props.

RobertJoonas (approved):

PR #4488, opened 2024-08-29
APIv2: Regex operations, consistent operators

macobo:
### Changes

Depends on https://github.com/plausible/analytics/pull/4480

This PR:
- Renames old `matches`/`does_not_match` operators to `matches_wildcard` and `matches_wildcard_not`. These have never been exposed to the frontend or to any user directly, only via APIv1 filtering syntax. As such we are free to rename these without breaking things
- Add support for `matches`/`matches_not` operators for query apiv2. These match the string against a regular expression, as defined in
https://github.com/google/re2/wiki/Syntax
- Rename `does_not_contain` to `contains_not` for consistency with other operators. Adds legacy support to dashboard so old filters don't break.

Note that `does_not_contain` support will be dropped in an upcoming PR once a day or so has passed and people have refreshed their dashboards.

Once this is merged I'll follow up with docs changes :)


lib/plausible/stats/filters/query_parser.ex
```diff
@@ -96,7 +98,16 @@ defmodule Plausible.Stats.Filters.QueryParser do
   defp parse_filter_key(filter), do: {:error, "Invalid filter '#{i(filter)}'."}
 
   defp parse_filter_rest(operator, filter)
-       when operator in [:is, :is_not, :matches, :does_not_match, :contains, :does_not_contain],
+       when operator in [
+              :is,
+              :is_not,
+              :match,
```
ukutaht, line 104: Naming: I think `matches` and `not_matches` would be more consistent here than `match` and `not_match`. For example we use `contains` over `contain`.

ukutaht: Also makes me think about consistency for negation. The consistent approaches would be IMO:
1. `not_matches` and `not_contains`
2. `does_not_match` and `does_not_contain`

I prefer the first since it's more succint. Maybe since APIv2 doesn't have any real clients yet we could rename `does_not_contain` to `not_contains` and use that form?

lib/plausible/stats/sql/where_builder.ex
```diff
@@ -241,34 +229,36 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
 
-  defp filter_field(db_field, [:does_not_match, _key, glob_exprs]) do
-    page_regexes = Enum.map(glob_exprs, &page_regex/1)
-
-    dynamic(
-      [x],
-      fragment("not(multiMatchAny(?, ?))", type(field(x, ^db_field), :string), ^page_regexes)
-    )
+  defp filter_field(db_field, [:not_matches_wildcard, dimension, clauses]) do
+    dynamic([], not (^filter_field(db_field, [:matches_wildcard, dimension, clauses])))
```
ukutaht, line 233: nice 👍

lib/plausible/stats/filters/query_parser.ex
macobo: Discussed this IRL and decided that `is_not`, `contains_not` and `matches_not` is the most consistent solution here.

Doing `contains_not` sadly requires a few more parsing and frontend changes. Will take a stab at this soon!

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -73,10 +73,14 @@ defmodule Plausible.Stats.Filters.QueryParser do
 
   defp parse_operator(["is" | _rest]), do: {:ok, :is}
   defp parse_operator(["is_not" | _rest]), do: {:ok, :is_not}
-  defp parse_operator(["matches" | _rest]), do: {:ok, :matches}
-  defp parse_operator(["does_not_match" | _rest]), do: {:ok, :does_not_match}
+  defp parse_operator(["match" | _rest]), do: {:ok, :match}
+  defp parse_operator(["match_not" | _rest]), do: {:ok, :match_not}
```
RobertJoonas, line 77: As I understand we're planning to rename the following:

* `match` -> `matches`
* `match_not` -> `matches_not`

Will this happen in a follow up PR?

Also, what's the reason for supporting both `match` (i.e. `matches`) and `matches_wildcard`?

macobo: Oops. Good catch, the PR was in a weird state. We now have `matches` and `matches_not`.

`matches_wildcard` is the wildcard matching operation we had in apiv1 which is not going anywhere but is not exposed in apiv2.

ukutaht (approved):

lib/plausible/stats/sql/where_builder.ex
```diff
@@ -194,21 +194,14 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
     dynamic(
-      [t],
-      has_key(t, column_name, ^prop_name) and
-        fragment(
-          "not(arrayExists(k -> match(?, k), ?))",
-          get_by_key(t, column_name, ^prop_name),
-          ^regexes
-        )
+      [],
+      not (^filter_custom_prop(prop_name, column_name, [:matches_wildcard, dimension, clauses]))
```
RobertJoonas, line 200: I believe this changes the current behaviour.

Currently, the `where` clause would evaluate to `false` for an event that doesn't have that `prop_key` at all. If we negate the whole thing, we would also return those events. Same comment applies for the `not_contains` filter on custom props.

Was this change desired?

lib/plausible/stats/sql/where_builder.ex
```diff
@@ -241,34 +229,36 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
+  defp filter_field(db_field, [:is, _dimension, clauses]) do
+    list = Enum.map(clauses, &db_field_val(db_field, &1))
     dynamic([x], field(x, ^db_field) in ^list)
   end
 
-  defp filter_field(db_field, [:is_not, _key, list]) do
-    list = Enum.map(list, &db_field_val(db_field, &1))
-    dynamic([x], field(x, ^db_field) not in ^list)
+  defp filter_field(db_field, [:is_not, dimension, clauses]) do
+    dynamic([], not (^filter_field(db_field, [:is, dimension, clauses])))
```
RobertJoonas, line 261: Why don't we need the `db_field_val()` call here anymore? It's still used with the `:is` filter.

lib/plausible/stats/sql/where_builder.ex
macobo: You are right and this indicates an important test is missing that this wasn't caught. Will add it and update the PR.

lib/plausible/stats/sql/where_builder.ex
```diff
@@ -220,19 +228,14 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
     dynamic(
-      [t],
-      has_key(t, column_name, ^prop_name) and
-        fragment(
-          "not(multiSearchAny(?, ?))",
-          get_by_key(t, column_name, ^prop_name),
-          ^clauses
-        )
+      [],
+      not (^filter_custom_prop(prop_name, column_name, [:contains, dimension, clauses]))
```
RobertJoonas, line 234: Behavioural change here too (same as with `matches_not` that you've fixed)

lib/plausible/stats/sql/where_builder.ex
RobertJoonas: Oh, right... My bad! 🤦🏻‍♂️

lib/plausible/stats/sql/where_builder.ex
macobo: Frustrating that the test suite didn't catch this. Added appropriate tests for contains/contains_not as well.

PR #4489, opened 2024-08-29
Acquisition channel

ukutaht:
### Changes

This PR will detect and store the [acquisition channel](https://feedback.plausible.io/175) for each session. The way we detect channels is based on GA4 channels as documented [here](https://support.google.com/analytics/answer/9756891?hl=en).

I've also added rules to detect as paid search when:
* Source is `Google` and `gclid` url parameter is present
* Source is `Bing` and `msclkid` url parameter is present

For now the plan is just to store this data and evaluate whether it looks correct. In a follow-up PR I will add an 'Acquisition Channel` report to the UI and make this data available from the Stats API as well.

I'm planning to extract the DB migration before deploying this PR.

### Tests

### Changelog
Will add once the feature is visible to users

### Documentation
Will add once the fetaure is visible to users

### Dark mode


lib/plausible/ingestion/acquisition.ex
```diff
@@ -24,4 +29,149 @@ defmodule Plausible.Ingestion.Acquisition do
...
+  end
+
+  defp video_source?(nil), do: false
+
+  defp video_source?(source) do
+    @source_categories[String.downcase(source)] == "SOURCE_CATEGORY_VIDEO"
+  end
+
+  defp paid_medium?(medium) do
+    Regex.match?(~r/^(.*cp.*|ppc|retargeting|paid.*)$/, medium || "")
```
zoldar, line 175: won't the `cp` pattern be too susceptible to false positives?

lib/plausible/ingestion/acquisition.ex
```diff
@@ -1,5 +1,10 @@
 defmodule Plausible.Ingestion.Acquisition do
   @moduledoc false
+  @source_categories Application.app_dir(:plausible, "priv/ga4-source-categories.csv")
```
zoldar, line 3: Would be good to declare `priv/ga4-source-categories.csv` as an external resource https://hexdocs.pm/elixir/Module.html#module-external_resource

Like here, for example: https://github.com/plausible/analytics/blob/master/lib/plausible/stats/json_schema.ex#L9

zoldar (approved):

macobo (approved):

lib/plausible/ingestion/acquisition.ex
ukutaht: Hmm yeah good point. It's copied verbatim from GA4 documentation:

Screenshot 2024-09-05 at 11 32 41

So I'm inclined to keep it as is

lib/plausible/ingestion/acquisition.ex
ukutaht: Makes sense! Done in https://github.com/plausible/analytics/pull/4489/commits/c3d5e7f8c7c5c15a61919a9de6b6557ec401e7e6

PR #4493, opened 2024-08-30
Enforce sequential processing of session events

zoldar:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/session/cache_store.ex
```diff
@@ -2,18 +2,27 @@ defmodule Plausible.Session.CacheStore do
...
-      updated_session = update_session(found_session, event)
-      buffer.insert([%{found_session | sign: -1}, %{updated_session | sign: 1}])
-      persist_session(updated_session)
-    else
-      new_session = new_session_from_event(event, session_attributes)
-      buffer.insert([new_session])
-      persist_session(new_session)
-    end
+  def on_event(event, session_attributes, prev_user_id, buffer_insert \\ &WriteBuffer.insert/1) do
+    Plausible.Cache.Adapter.with_lock!(
```
ruslandoga, line 8: Wouldn't this approach make session ingestion essentially single-threaded? It seems like having a single GenServer handle all the requests could create a potential bottleneck.

zoldar: Good point. Like you say, we already use that locking mechanism implicitly and the lock state map in the cache process shouldn't change in any significant way. Either way, I have added a TODO item to instrument lock wait time, so that we can keep an eye on how much time is usually spent in that state (either due to contentious locking situation - probably rare and limited to customers with excessive numbers of events emitted - or the lookup time alone).

zoldar: https://github.com/plausible/analytics/pull/4493/commits/126d18c4b0242811aa22c54a6c9527b85f054c5d

ruslandoga (approved):

aerosol (approved):

Issue #4257, opened 2024-06-21
Can't view stats for a longish time range


jonahgreenthal:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

When I try to view page stats for a time range longer than about six months, the page spins seemingly indefinitely (even overnight) without actually producing data.

This instance is self-hosted on a `t2.micro` EC2 instance. One possibility is that calculating the aggregation is too resource-intensive, though `top` shows that there is plenty of available memory (and even more available swap space). If that is the issue, though, there should at least be a reasonable error message, though realistically, I'd be comfortable waiting (just not indefinitely). 

I would be happy to do additional work to try to narrow this down, troubleshoot, etc. However, I do not know anything about Plausible internals.

### Expected behavior

The report for the indicated time range should load, preferably in a reasonable amount of time.

### Screenshots

![image](https://github.com/plausible/analytics/assets/737172/a7756f1f-d47a-42ed-bcbd-94dc80cec480)

### Environment

```markdown
- OS: Windows 10
- Browser: Firefox
- Browser Version: 127.0
```

ruslandoga:
👋 @jonahgreenthal 

According to the browser console the query times out after ten seconds. You can might get more info from the server logs:

```console
$ cd hosting # or wherever you cloned plausible/community-edition
$ docker compose logs
```

There are two possible improvements:
- make timeout configurable or remove it completely in self-hosted
- show an error in UI and stop the spinning wheel

jonahgreenthal:
Thanks! Here's what I see in the logs:
```
plausible_plausible_events_db_1  | 2024.06.21 13:07:38.363143 [ 328 ] {58dfd4a2-fe00-4cec-a90c-90d9d45f3a49} <Error> DynamicQueryHandler: Cannot send exception to client: Code: 24. DB::Exception: Cannot write to ostream at offset 712. (CANNOT_WRITE_TO_OSTREAM), Stack trace (when copying this message, always include the lines below):
plausible_plausible_events_db_1  | 
plausible_plausible_events_db_1  | 0. DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int, bool) @ 0xa3fb75a in /usr/bin/clickhouse
[...]

PR #4494, opened 2024-08-30
Make ClickhouseRepo task timeouts configurable in CE

ruslandoga:
### Changes

This PR makes CE's ClickhouseRepo tasks reuse timeouts defined on `CLICKHOUSE_DATABASE_URL`

For example, setting `CLICKHOUSE_DATABASE_URL=http://plausible_events_db:8123/plausible_events_db?timeout=120000` would make the task timeout value equal eight minutes [(480000=url_timeout*4).](https://github.com/plausible/analytics/pull/4494#discussion_r1740639595) It has already been used to solve some timeout issues before: https://github.com/plausible/analytics/discussions/3311 - so it makes sense to standardize on this approach rather than introduce extra ENV vars. 

Related issues and discussions, where increasing the timeout value for tasks might help:
- https://github.com/plausible/analytics/issues/4257
- https://github.com/plausible/analytics/discussions/4360
- https://github.com/plausible/analytics/discussions/4285

Some examples for how the timeout param in URL affects Repo config:

```elixir
# for export CLICKHOUSE_DATABASE_URL=http://localhost:8123/plausible_events_db
iex> Keyword.fetch!(Plausible.ClickhouseRepo.config(), :timeout)
#==> 15000 <- this is the default from db_connection

# for export CLICKHOUSE_DATABASE_URL="http://localhost:8123/plausible_events_db?timeout=120000"
iex> Keyword.fetch!(Plausible.ClickhouseRepo.config(), :timeout)
#==> 120000 <- this would make tasks timeout after 120000*4=480000 or eight minutes
```

### Tests
- [ ] Not sure how (or if) to test this, maybe `config_test.exs` can be updated, but for a custom repo like in https://github.com/plausible/analytics/pull/4466

### Changelog
- [x] Does not need a new entry, I'll just reply in the discussions above that the task timeouts can be configured in `CLICKHOUSE_DATABASE_URL`

### Documentation

### Dark mode


lib/plausible/clickhouse_repo.ex
```diff
@@ -23,9 +25,17 @@ defmodule Plausible.ClickhouseRepo do
 
     max_concurrency = Keyword.get(opts, :max_concurrency, 3)
 
+    task_timeout =
+      on_ee do
+        @task_timeout
+      else
+        ch_timeout = Keyword.fetch!(config(), :timeout)
+        max(ch_timeout * 2, @task_timeout)
```
ruslandoga, line 33: I'm doubling the timeout value from the repo just to make sure we don't timeout the task before the db_connection. It could be `max(ch_timeout + 1, @task_timeout)` but I thought that `* 2` is safer.

zoldar: Yeah, let's adjust it to 4 👍

zoldar:
> Not sure how (or if) to test this

@ruslandoga Perhaps starting a repo process without name and using https://hexdocs.pm/ecto/Ecto.Repo.html#c:put_dynamic_repo/1 would enable testing it (the original repo pid/name would have to be put back at the end of the test though).

lib/plausible/clickhouse_repo.ex
ruslandoga: https://github.com/plausible/analytics/pull/4494/commits/f331f58ac59b7d5572c125ab9d2a0c3b87823dc4

ruslandoga:
I've added the test in https://github.com/plausible/analytics/pull/4494/commits/f8675e077acf5d02a22223b461283c5537894ac1. It turned out complicated, though. I'll try to clean it up a bit and remove the need for exit trapping.

zoldar (approved):

PR #4495, opened 2024-08-30
Fix invalid parameter destructuring in `navigate` calls

zoldar:
### Changes

The destructuring results in ignoring existing search params which are not modified by a given navigation call - in case of removing filters it's date range and in case of time range selection from graph, it's filters.

### Tests
- [x] Tested manually


apata (approved):

apata:
Nice job catching this! 💯

PR #4491, opened 2024-08-30
Auto HTTPS in CE

ruslandoga:
### Changes

This PR starts using https://github.com/sasa1977/site_encrypt in CE to generate (and keep up-to-date) TLS certificates using https://certbot.eff.org.

The goal is to simplify CE deployments by removing the reverse-proxy requirement.

### Trying it out locally

- Launch a local CA like [Pebble:](https://github.com/letsencrypt/pebble)

  ```console
  $ docker run --rm --name pebble -d -p 14000:14000 -p 15000:15000 -e "PEBBLE_VA_NOSLEEP=1" ghcr.io/letsencrypt/pebble
  ```

  ---

  **UPDATE:** since Docker Desktop 4.34.0 host networking can be used on non-Linux machines:   https://docs.docker.com/engine/network/drivers/host/?uuid=ebc0dd6c-ae8d-43ac-aa17-81671a3a76dd%0A#docker-desktop, that would remove the need for `host.docker.internal` workaround below.

  ---

- Install Certbot, e.g. with `brew install certbot`

- Install new CE deps, e.g. with `MIX_ENV=ce mix deps.get`

- Edit config/.env.dev and .iex.exs so as not to interfere with custom ENV used in the next step:

  <sub><kbd>config/.env.dev</kbd></sub>
  ```diff
  - BASE_URL=http://localhost:8000
    SECURE_COOKIE=false
    DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/plausible_dev
    CLICKHOUSE_DATABASE_URL=http://127.0.0.1:8123/plausible_events_db
    ...
  ```

  <sub><kbd>.iex.exs</kbd></sub>
  ```diff
    ...

    import_if_available(Ecto.Query)
    import_if_available(Plausible.Factory)
  
  - Logger.configure(level: :warning)
  
    ...
  ```
  
- Launch Plausible with the following ENV:
  <sub><kbd>console</kbd></sub>
  ```console
  $ export BASE_URL=https://host.docker.internal:5001
  $ export HTTP_PORT=5002
  $ export HTTPS_PORT=5001
  $ export ACME_DIRECTORY_URL=https://localhost:14000/dir
  $ export DATA_DIR=./tmp
  $ export MIX_ENV=ce_dev
  $ mix phx.server
  ```

  ---

[...]

PR #4502, opened 2024-09-02
Support order_by in URL params of breakdown endpoints

apata:
### Changes

* Adds support for order_by in URL params of breakdown endpoints

### Tests

### Changelog
- [x] Does not need a changelog entry

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -20,6 +20,7 @@ defmodule Plausible.Stats.Breakdown do
       Query.set(
         query,
         metrics: transformed_metrics,
+        # Concat client requested order with default order, overriding only if client explicitly requests it
         order_by:
           Enum.concat(query.order_by || [], infer_order_by(transformed_metrics, dimension))
           |> Enum.uniq_by(&elem(&1, 0)),
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -20,7 +20,9 @@ defmodule Plausible.Stats.Breakdown do
       Query.set(
         query,
         metrics: transformed_metrics,
-        order_by: infer_order_by(transformed_metrics, dimension),
+        order_by:
+          Enum.concat(query.order_by || [], infer_order_by(transformed_metrics, dimension))
+          |> Enum.uniq_by(&elem(&1, 0)),
         dimensions: transform_dimensions(dimension),
         filters: query.filters ++ dimension_filters(dimension),
         v2: true,
@@ -174,7 +176,8 @@ defmodule Plausible.Stats.Breakdown do
     end)
   end
 
-  defp infer_order_by(metrics, "event:goal"), do: [{metric_to_order_by(metrics), :desc}]
+  defp infer_order_by(metrics, "event:goal"),
+    do: [{metric_to_order_by(metrics), :desc}]
 
   defp infer_order_by(metrics, dimension),
     do: [{metric_to_order_by(metrics), :desc}, {dimension, :asc}]
--- a/lib/plausible/stats/legacy/legacy_query_builder.ex
+++ b/lib/plausible/stats/legacy/legacy_query_builder.ex
@@ -1,9 +1,12 @@
 defmodule Plausible.Stats.Legacy.QueryBuilder do
-  @moduledoc false
+  @moduledoc """
+  Module used to parse URL search params to a valid Query, used to power the API for the dashboard.
+  @deprecated
+  """
 
   use Plausible
 
-  alias Plausible.Stats.{Filters, Interval, Query, DateTimeRange}
+  alias Plausible.Stats.{Filters, Interval, Query, DateTimeRange, Metrics}
 
   def from(site, params, debug_metadata) do
     now = DateTime.utc_now(:second)
@@ -16,6 +19,7 @@ defmodule Plausible.Stats.Legacy.QueryBuilder do
       |> put_interval(params)
       |> put_parsed_filters(params)
       |> put_preloaded_goals(site)
+      |> put_order_by(params)
       |> Query.put_experimental_reduced_joins(site, params)
       |> Query.put_imported_opts(site, params)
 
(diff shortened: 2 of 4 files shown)


test/plausible_web/controllers/api/stats_controller/sources_test.exs
```diff
@@ -498,6 +498,91 @@ defmodule PlausibleWeb.Api.StatsController.SourcesTest do
...
+      ])
+
+      filters = Jason.encode!([["is", "event:page", ["/page1"]]])
+      conn = get(conn, "/api/stats/#{site.domain}/sources?filters=#{filters}")
+
+      assert json_response(conn, 200)["results"] == [
+               %{"name" => "Google", "visitors" => 2},
+               %{"name" => "DuckDuckGo", "visitors" => 1}
+             ]
+    end
```
apata, line 525: Just an additional sanity check for already existing functionality :)

lib/plausible/stats/legacy/legacy_query_builder.ex
```diff
@@ -164,6 +165,11 @@ defmodule Plausible.Stats.Legacy.QueryBuilder do
     end
   end
 
+  defp put_order_by(query, %{} = params) do
+    order_by = Map.get(params, "order_by", nil) |> Filters.parse_order_by()
```
macobo, line 169: It isn't clear `Filters.parse_order_by` is part of 'legacy' parsing code that we can delete soon and shouldn't be used outside of it.

Instead of calling Filters module we should either:
a) Make `Plausible.Stats.Filters.QueryParser.parse_order_by` public
or 
b) Make it a `defp` in this class instead of in the filters module

lib/plausible/stats/filters/filters.ex
```diff
@@ -92,4 +93,56 @@ defmodule Plausible.Stats.Filters do
...
+  def parse_order_by(order_by) when is_binary(order_by) do
+    case Jason.decode(order_by) do
+      {:ok, parsed} when is_list(parsed) ->
+        parsed
+        |> Enum.flat_map(&parse_order_by_pair/1)
+
+      {:ok, _} ->
+        []
+
+      {:error, _err} ->
```
macobo, line 129: What's the condition it may raise here? What expected case is this branch protecting against?

lib/plausible/stats/breakdown.ex
```diff
@@ -20,7 +20,9 @@ defmodule Plausible.Stats.Breakdown do
       Query.set(
         query,
         metrics: transformed_metrics,
-        order_by: infer_order_by(transformed_metrics, dimension),
+        order_by:
+          Enum.concat(query.order_by || [], infer_order_by(transformed_metrics, dimension))
+          |> Enum.uniq_by(&elem(&1, 0)),
```
macobo, line 25: This is quite complex.

Why not: `query.order_by || infer_order_by(transformed_metrics, dimension)`?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - New /debug/clickhouse route for super admins which shows information on clickhouse queries executed by user
 - Typescript support for `/assets`
 - Testing framework for `/assets`
+- Accept `order_by` search parameter in dashboard breakdown endpoints
```
macobo, line 21: Not sure we should document this in the changelog given it isn't a user-visible change and users should not be using these endpoints. Gonna leave it up to your best judgement though :)

lib/plausible/stats/legacy/legacy_query_builder.ex
aerosol: `nil` is also default when no value is found, so `Map.get(params, "order_by")` suffices, as well as `params["order_by"]`

lib/plausible/stats/filters/filters.ex
apata: Thanks! Adding the invalid JSON doctest. Because non-array and invalid json code paths are the same, will roll them into `_ ->` case.

lib/plausible/stats/legacy/legacy_query_builder.ex
apata: Thanks for the suggestions! Went with B, but got a warning on doc string (and therefore my doctests) being ignored on private functions, had to keep it `def`.

lib/plausible/stats/breakdown.ex
apata: Ah damn, and I was hoping this is elegant! 

I'm doing it this way to keep the default inferred order by always active but lowest priority, unless specifically overridden in the dashboard (`Enum.uniq_by` taking the first uniques in the array).

I am anticipating this help to prevent jumpy sorts on the FE side. I'll keep the || approach in mind for when it doesn't.

CHANGELOG.md
apata: Thanks! Thought it was mostly for the CE users, removing.

lib/plausible/stats/breakdown.ex
macobo: A bit suspect, but okay. I think it's worth adding a comment to the effect of why the concat here (and no-where else) so the historic context doesn't get lost.

macobo (approved):
All good (bar missing comment). :1st_place_medal:

ruslandoga:
The new test seems to fail CI in https://github.com/plausible/analytics/pull/4491

test/plausible_web/controllers/api/stats_controller/sources_test.exs
```diff
@@ -498,6 +498,108 @@ defmodule PlausibleWeb.Api.StatsController.SourcesTest do
...
+        ),
+        build(:pageview, referrer_source: "A"),
+        build(:pageview, referrer_source: "A"),
+        build(:pageview, referrer_source: "Z")
+      ])
+
+      order_by_asc = Jason.encode!([["visit_duration", "asc"], ["visit:source", "desc"]])
+
+      conn1 =
+        get(conn, "/api/stats/#{site.domain}/sources?detailed=true&order_by=#{order_by_asc}")
```
ruslandoga, line 582: Is there any time filtering happening here? If so, should the timestamps in pageviews be dynamic?

ruslandoga: Seems like it filters the last month by default, since it started failing today.

PR #4503, opened 2024-09-02
Fix property filter suggestions 500 error when property hasn't been selected

macobo:
Broken by APIv2 refactors.

Relevant sentry error: https://sentry.plausible.io/organizations/plausible/issues/448/?query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=4

### Changelog


aerosol (approved):
I couldn't repro anymore with this branch. Would it be too much work to turn filter parsing failures into 4xx client errors though?

macobo:
> Would it be too much work to turn filter parsing failures into 4xx client errors though?

Old code uses `Query.from` which doesn't handle any parsing failures and is closely tied to stats apiv1. We'll be moving away from this soonish so the time could be spent just refactoring it to use `Query.build` instead. I just might go ahead and do that here in a follow-up today.

PR #4504, opened 2024-09-02
Dont include imports for time:hour and time:minute dimensions

macobo:
Also include more information about import warnings in query results.

This mostly affects Stats API v2.

Documentation changes incoming.


lib/plausible/stats/query_result.ex
```diff
@@ -70,7 +70,10 @@ defmodule Plausible.Stats.QueryResult do
 
   defp meta(query) do
     %{
-      warning:
+      imports_included: if(query.include.imports, do: query.include_imported, else: nil),
```
RobertJoonas, line 73: I'm a little confused by this line here. I think it should always be a boolean. Wouldn't it work simply like this?

```suggestion
      imports_included: query.include.imports,
```

RobertJoonas (commented):
The main graph doesn't display any warning or message about the main graph excluding imported data due to the applied interval. I'll fix that in this PR.

lib/plausible/stats/query_result.ex
macobo: Oops. This doesn't work and the tests luckily caught that. Intent is to signal whether imports were included successfully with this flag.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Testing framework for `/assets`
 
 ### Removed
+- Stop returning imported stats in timeseries queries for intervals shorter than `day`
```
RobertJoonas, line 23: This looks like a duplicate changelog entry for the same thing. Let's drop it.

```suggestion
```

RobertJoonas (approved):
Other than the duplicate changelog entry it's good to go! 👍

PR #4505, opened 2024-09-02
Fix password reset link

aerosol:


zoldar (approved):

macobo (approved):

macobo:
Could this rename also cause problems because of any emails we might have sent out recently?

aerosol:
> Could this rename also cause problems because of any emails we might have sent out recently?

No, route is intact

PR #4509, opened 2024-09-02
Proper fix for realtime timeseries queries

RobertJoonas:
### Changes

Follow up to https://github.com/plausible/analytics/pull/4508 with a proper fix that will also work for API v2.

The solution is actually much simpler than what I essentially did in that PR. The problem was that we were selecting the time dimensions by shifting the timestamps into `query.timezone` **which was always `site.timezone`, and yet the actual `query.date_range` for realtime was in UTC**.

So eventually the bug surfaced in `Plausible.Stats.Timeseries.build_timeseries_result` from something like this:

```elixir
%Plausible.Stats.QueryResult{
  # returned dimensions are in site.timezone
  results: [%{metrics: [1], dimensions: ["2024-09-03 08:30:00"]}],
  # the time labels are created with a minute step starting from `query.date_range.first` and are in UTC
  meta: %{
    time_labels: ["2024-09-03 05:20:00", "2024-09-03 05:21:00", ..., "2024-09-03 05:49:00"]
  },
  ...
}

```

**The fix**

`query.timezone` does not have to be equivalent to `site.timezone`, so the fix is simply to make sure that `query.timezone` is in UTC for realtime queries. Doing that, we'll be converting from UTC to UTC when selecting the time dimensions, which is a no-op.

What I've done now is rely on `query.date_range.first` to determine `query.timezone` in both `Legacy.QueryBuilder` and `Query.build`. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/filters/query_parser.ex
+++ b/lib/plausible/stats/filters/query_parser.ex
@@ -31,7 +31,7 @@ defmodule Plausible.Stats.Filters.QueryParser do
            date_range: date_range,
            dimensions: dimensions,
            order_by: order_by,
-           timezone: site.timezone,
+           timezone: date_range.first.time_zone,
            preloaded_goals: preloaded_goals,
            include: include
          },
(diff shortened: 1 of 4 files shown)


ukutaht (approved):

aerosol (approved):
oh that's way more elegant

PR #4528, opened 2024-09-04
Replace native tooltips with simplified, generic component

aerosol:
### Changes

This change simplifies the tooltip component by reducing its input attributes to a minimum. The component is then used to replace native browser tooltips. 

List of tooltip locations:
  - entries on all shield lists 
  - plan box when usage exceeds selected plan configuration
  - slow import 
  - script extensions on installation screen 

### Tests

### Changelog

### Documentation

### Dark mode
- x ] The UI has been tested both in dark and light mode


lib/plausible_web/components/generic.ex
```diff
@@ -352,26 +348,28 @@ defmodule PlausibleWeb.Components.Generic do
     assigns = assign(assigns, wrapper_data: wrapper_data, show_inner: show_inner)
 
     ~H"""
-    <div x-data={@wrapper_data} class={["tooltip-wrapper relative", @wrapper_class]}>
-      <p
+    <div x-data={@wrapper_data} class="tooltip-wrapper w-full relative">
+      <div
+        x-cloak
+        x-show={@show_inner}
+        class="tooltip-content z-[1000] bg-gray-900 rounded text-white absolute bottom-24 sm:bottom-7 left-0 sm:w-72 p-4 text-sm font-medium"
```
apata, line 355: **nitpick:** "bottom-24 sm:bottom-7" the first class here causes the tooltip to be placed quite high above the item on mobile. It makes it harder to see which element it's about in the shields list.

apata (approved):

apata:
General accessibility comment: hover tooltip bad :D 

In the shields list, with voiceover active, I wasn't able to get the reader to read out the "Added at...". May seem counterintuitive, but the tooltip wrapper should probably be a unstyled button that toggles the tooltip: that way, the wrapper is focusable with Tab key navigation, like other interactive elements.

lib/plausible_web/components/generic.ex
aerosol: I have trouble deciding what the default behaviour should be on mobile - especially when it comes to the choose plan screen. Do you have any suggestions?

apata: Ah, so that spacing was optimized for the plan screen, which I didn't check. There won't be a perfect value then probably.

aerosol:
oh wait no

aerosol:
No, I don't understand how the wrapper should be a button, halp

apata:
> Oof, fair point. I'll see if I can pull that one off, although, now that you mention it, I guess accessibility in general is a project on its own.


> 

I tried exactly that actually when I made the original comment, and at least in the shields table it became focusable. The screenreader offered the ability to press the button. When I took it up on that offer, it inserted the tooltip into the dom and since it was part of the button now, read it out "added by ... at ... <flag> <country>".

apata:
The question is whether the styles break and how badly: if badly, then as you say a11y is a project in its own right.

aerosol:
The styling doesn't break but it didn't make sense semantically - the button wasn't clickable after all? I think I'll let it be for now, until we figure out a plan on how to tackle the matter globally perhaps.

PR #4531, opened 2024-09-04
Migration for acquisition channel

ukutaht:
Migration extracted from https://github.com/plausible/analytics/pull/4489

### DIFF
--- a/priv/ingest_repo/migrations/20240829092858_add_acquisition_channel.exs
+++ b/priv/ingest_repo/migrations/20240829092858_add_acquisition_channel.exs
@@ -1,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.AddAcquisitionChannel do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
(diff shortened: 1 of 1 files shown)


priv/ingest_repo/migrations/20240829092858_add_acquisition_channel.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Plausible.IngestRepo.Migrations.AddAcquisitionChannel do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
+      add :channel, :"LowCardinality(String)"
```
macobo, line 6: Is a channel user-provided or a fixed list? LowCardinality can cause a issues if the cardinality isn't low = not a more-or-less fixed list.

ukutaht: Yes, I think the cardinality is something like 10 and will not grow.

zoldar (approved):

macobo (approved):

PR #4533, opened 2024-09-05
[NOT NOW]: WIP graph to API v2

RobertJoonas:
The aim of this work was to make a start with migrating the dashboard to API v2, adding a new controller action for `POST /main-graph` that would use `Query.build` rather than `Query.from` (`Legacy.QueryBuilder`).

However, we've decided not to continue with this for now, due to several complications:

* APIv2 is currently missing functionality that would still have to be included in the controller:
  * Comparisons
  * Revenue goals
* There is no clear vision yet for how the frontend should make these requests (e.g. `POST` vs `GET`) and it makes more sense to work on migrating the whole dashboard to API v2 at once, taking into account the requirements of other endpoints, not just `main_graph`.

In conclusion, this is not something we want to rush into, and there's no real benefit in doing it at this stage.

Nevertheless, some of the code in this PR might become useful, so I'm keeping this PR open as a draft until necessary.

### DIFF
--- a/assets/js/dashboard/api.js
+++ b/assets/js/dashboard/api.js
@@ -1,5 +1,6 @@
 import { formatISO } from './util/date'
-import { serializeApiFilters } from './util/filters'
+import { addPrefixes, serializeApiFilters } from './util/filters'
+import { apiPath } from './util/url'
 
 let abortController = new AbortController()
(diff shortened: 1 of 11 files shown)


macobo:
Let's close for now - made a note of this work and will dig it up when converting main_graph.

PR #4540, opened 2024-09-06
Remove support for legacy user sessions

zoldar:
### Changes

This PR removes support for legacy sessions entirely. They are treated the same as "no valid token" case.

Additionally, on password reset, all user's sessions are deleted and their LV sockets are disconnected (note: this has limited effect when the pubsub is not connected between the nodes in multi-node scenario).

### Tests


lib/plausible_web/user_auth.ex
```diff
@@ -79,6 +65,18 @@ defmodule PlausibleWeb.UserAuth do
     end
   end
 
+  @spec revoke_all_user_sessions(Auth.User.t()) :: :ok
+  def revoke_all_user_sessions(user) do
+    {_count, tokens} =
+      Repo.delete_all(
+        from us in Auth.UserSession, where: us.user_id == ^user.id, select: us.token
```
aerosol, line 72: select is probably unnecessary here?

lib/plausible_web/user_auth.ex
```diff
@@ -79,6 +65,18 @@ defmodule PlausibleWeb.UserAuth do
...
+      )
+
+    Enum.each(tokens, fn token ->
+      PlausibleWeb.Endpoint.broadcast(live_socket_id(token), "disconnect", %{})
```
aerosol, line 76: nice! awaiting empd integration 🤘 is this a built-in message that's handled elsewhere?

aerosol (approved):
great

lib/plausible_web/user_auth.ex
zoldar: `select` here makes the DELETE query return tokens without selecting them separately - they are then iterated over to broadcast "disconnect LV" message.

lib/plausible_web/user_auth.ex
zoldar: Yeah, this is built-in: https://hexdocs.pm/phoenix_live_view/0.15.3/security-model.html#disconnecting-all-instances-of-a-given-live-user

zoldar:
@plausible/developers is https://github.com/plausible/analytics/pull/4540/commits/77d1a1035658915f9afe538afc5fb9a3da0ec905 the proper way to force dashboard refresh when any API request returns with 404 or should that be accomplished differently?

Reasoning for the change:

Currently, when the token expires, until dashboard page is fully refreshed, the user has no feedback. Failed API requests either don't change anything (/current-visitors query which is run periodically) or get stuck in loading/spinning state (filter suggestions for instance). Hence the idea to tread 404 error differently by force reloading the dashboard so that changed token state (or removed site?) is accounted for sooner without user wondering why everything is half-broken.

PR #4541, opened 2024-09-06
Reassert tuples in query JSON schema

apata:
### Changes

This reasserts the tuples in the json schemas, so the autocomplete / autoverify in the playground would inform users when they are constructing invalid filters.
* Some whitespace changes due to prettier.
* Slight refactor to lift out filtering operation definitions (benefit: simpler references)

### Tests

### Changelog

### Documentation

### Dark mode

### Schema verification without this PR:
Screenshot 2024-09-06 at 13 06 13

### Schema verification with this PR:
Screenshot 2024-09-06 at 13 07 21

### DIFF
--- a/lib/plausible/stats/json_schema.ex
+++ b/lib/plausible/stats/json_schema.ex
@@ -17,11 +17,11 @@ defmodule Plausible.Stats.JSONSchema do
   @internal_query_schema @raw_public_schema
                          # Add overrides for things allowed in the internal API
                          |> JSONPointer.add!(
-                           "#/definitions/filter_entry/oneOf/0/items/0/enum/0",
+                           "#/definitions/filter_operation_without_goals/enum/0",
                            "matches_wildcard"
                          )
                          |> JSONPointer.add!(
-                           "#/definitions/filter_entry/oneOf/0/items/0/enum/0",
+                           "#/definitions/filter_operation_without_goals/enum/0",
                            "matches_wildcard_not"
                          )
                          |> JSONPointer.add!("#/definitions/metric/oneOf/0", %{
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/json_schema.ex
```diff
@@ -17,11 +17,11 @@ defmodule Plausible.Stats.JSONSchema do
   @internal_query_schema @raw_public_schema
                          # Add overrides for things allowed in the internal API
                          |> JSONPointer.add!(
-                           "#/definitions/filter_entry/oneOf/0/items/0/enum/0",
+                           "#/definitions/filter_operation_negative/enum/0",
```
macobo, line 20: Naming it `filter_operation_negative` doesn't make sense and this line highlights why.

The intent is to disallow some operations against event:goal filters.

macobo (commented):
This looks great. It would be good to have some more tests under test/plausible/stats/query_parser_test.exs to show the need for some of the new additions (additionalItems, minItems) or the error message changing.

macobo:
Was testing this out. It introduces a regression:
image

`event:goal` should not be autocompleted in this scenario.

On master, this does not propose `event:goal`:
image

apata:
> Was testing this out. It introduces a regression: image
> 
> `event:goal` should not be autocompleted in this scenario.
> 
> On master, this does not propose `event:goal`: image

Thanks for checking this out, well spotted! I removed the deprecated constant from the schema and contains_not autocomplete is working as intended again. Shall we just delay the merge until the week is up?

apata:
Since we got rid of the deprecated const in master, the autocomplete in this is working again. Any other blockers for merging this, @macobo ?

PR #4554, opened 2024-09-09
Verification: Add super-admin report displaying diagnostics

aerosol:
https://github.com/user-attachments/assets/e57ac18a-d92f-4331-b50a-e922ed0e8608

Yet another debug/troubleshooting option for super-admins

### DIFF
--- a/lib/plausible_web/live/components/verification.ex
+++ b/lib/plausible_web/live/components/verification.ex
@@ -10,18 +10,21 @@ defmodule PlausibleWeb.Live.Components.Verification do
 
   import PlausibleWeb.Components.Generic
 
-  attr :domain, :string, required: true
+  attr(:domain, :string, required: true)
 
-  attr :message, :string, default: "We're visiting your site to ensure that everything is working"
-  attr :super_admin?, :boolean, default: false
-  attr :finished?, :boolean, default: false
-  attr :success?, :boolean, default: false
-  attr :verification_state, Plausible.Verification.State, default: nil
-  attr :interpretation, Plausible.Verification.Diagnostics.Result, default: nil
-  attr :attempts, :integer, default: 0
-  attr :flow, :string, default: ""
-  attr :installation_type, :string, default: nil
-  attr :awaiting_first_pageview?, :boolean, default: false
+  attr(:message, :string,
+    default: "We're visiting your site to ensure that everything is working"
+  )
+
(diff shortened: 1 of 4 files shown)


lib/plausible_web/live/components/verification.ex
```diff
@@ -10,17 +10,21 @@ defmodule PlausibleWeb.Live.Components.Verification do
 
   import PlausibleWeb.Components.Generic
 
-  attr :domain, :string, required: true
+  attr(:domain, :string, required: true)
```
aerosol, line 13: argh, lsp races, will revert

lib/plausible_web/live/components/verification.ex
```diff
@@ -121,6 +122,34 @@ defmodule PlausibleWeb.Live.Components.Verification do
...
+                class="bg-yellow-100 dark:text-gray-800"
+              >
+                As a super-admin, you're eligible to see diagnostics details. Click to expand.
+              </a>
+            </p>
+            <div x-show="showDiagnostics" x-cloak>
+              <.focus_list>
+                <:item
+                  :for={{diag, value} <- Map.from_struct(@verification_state.diagnostics)}
+                  :if={@verification_state && @super_admin?}
```
zoldar, line 144: Isn't this `:if` redundant given wrapping `div` contains this condition already?

lib/plausible_web/live/components/verification.ex
```diff
@@ -121,6 +122,34 @@ defmodule PlausibleWeb.Live.Components.Verification do
               </.styled_link>
             </:item>
           </.focus_list>
+          <div
+            :if={@verification_state && @super_admin? && @finished?}
```
zoldar, line 126: Wouldn't `@super_admin? and @finished` be enough?

zoldar (approved):
Nice ✨

lib/plausible_web/live/components/verification.ex
aerosol: It is, thanks! bfd6a4a8b

lib/plausible_web/live/components/verification.ex
aerosol: Hm yes, should be, but technically nothing enforces it otherwise at the component level, so I'd rather keep it defensive here.

PR #4555, opened 2024-09-09
Safeguard queries relying on `sign` from faulty old session entries

zoldar:
### Changes

Since #4493 got deployed, faulty session records stopped appearing. This PR introduces a change to bounce computation (for native stats), which safeguards against negative bounce rate value overflow for historical records, ultimately caused by those races occurring in the past. 

### Tests

### DIFF
--- a/lib/plausible/stats/sql/expression.ex
+++ b/lib/plausible/stats/sql/expression.ex
@@ -252,40 +252,33 @@ defmodule Plausible.Stats.SQL.Expression do
     wrap_alias([], %{
       bounce_rate:
         fragment(
-         # :TRICKY: Before PR #4493, we could have sessions where sum(is_bounce * sign) is negative, leading to an underflow and >100% bounce rate. This works around that issue. 
+          # :TRICKY: Before PR #4493, we could have sessions where `sum(is_bounce * sign)`
+          # is negative, leading to an underflow and >100% bounce rate. This works around
+          # that issue.
           "toUInt32(greatest(ifNotFinite(round(sumIf(is_bounce * sign, ?) / sumIf(sign, ?) * 100), 0), 0))",
           ^condition,
           ^condition
         ),
-      __internal_visits: fragment("toUInt32(greatest(sum(sign), 0))")
+      __internal_visits: fragment("toUInt32(sum(sign))")
     })
   end
 
   def session_metric(:visits, _query) do
     wrap_alias([s], %{
-      visits: fragment("toUInt64(round(greatest(sum(?), 0) * any(_sample_factor)))", s.sign)
+      visits: fragment("toUInt64(round(sum(?) * any(_sample_factor)))", s.sign)
     })
   end
 
   def session_metric(:pageviews, _query) do
     wrap_alias([s], %{
       pageviews:
-        fragment(
-          "toUInt64(round(greatest(sum(? * ?), 0) * any(_sample_factor)))",
-          s.sign,
-          s.pageviews
-        )
+        fragment("toUInt64(round(sum(? * ?) * any(_sample_factor)))", s.sign, s.pageviews)
     })
   end
 
   def session_metric(:events, _query) do
     wrap_alias([s], %{
-      events:
-        fragment(
-          "toUInt64(round(greatest(sum(? * ?), 0) * any(_sample_factor)))",
-          s.sign,
-          s.events
-        )
+      events: fragment("toUInt64(round(sum(? * ?) * any(_sample_factor)))", s.sign, s.events)
     })
   end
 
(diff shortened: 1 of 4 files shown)


lib/plausible/stats/sql/expression.ex
```diff
@@ -252,30 +252,39 @@ defmodule Plausible.Stats.SQL.Expression do
     wrap_alias([], %{
       bounce_rate:
         fragment(
-          "toUInt32(ifNotFinite(round(sumIf(is_bounce * sign, ?) / sumIf(sign, ?) * 100), 0))",
+          "toUInt32(greatest(ifNotFinite(round(sumIf(is_bounce * sign, ?) / sumIf(sign, ?) * 100), 0), 0))",
```
macobo, line 255: Is this correct? The undeflow only happens on `toUInt32` cast right?

lib/plausible/stats/sql/expression.ex
```diff
@@ -252,30 +252,39 @@ defmodule Plausible.Stats.SQL.Expression do
...
           ^condition,
           ^condition
         ),
-      __internal_visits: fragment("toUInt32(sum(sign))")
+      __internal_visits: fragment("toUInt32(greatest(sum(sign), 0))")
```
macobo, line 259: This is not needed - sum(sign) is still always correct even with the incorrect merging. Let's keep it as is.

macobo:
Let's only update `bounce_rate` unless others can be verified to be buggy in the scenario we know about. These sorts of checks are impossible to remove after the fact and we don't want to carry this complexity forever.

lib/plausible/stats/sql/expression.ex
macobo, line 255:
```suggestion
         # :TRICKY: Before PR #4493, we could have sessions where sum(is_bounce * sign) is negative, leading to an underflow and >100% bounce rate. This works around that issue. 
          "toUInt32(greatest(ifNotFinite(round(sumIf(is_bounce * sign, ?) / sumIf(sign, ?) * 100), 0), 0))",
```

lib/plausible/stats/sql/expression.ex
zoldar: I'm sorry, I don't follow. How is it possible that `sum(sign)` returns correct result while `sum(is_bounce * sign)` doesn't?

Also, currently, all query expressions with `sign` in `Plausible.Exports` already apply `greatest(...,0)`, including visit duration, visits etc. So, as it currently stands, the base query expressions are out of sync with them.

zoldar:
I've settled on only altering bounce rate expression as it's the only metric so far that visibly manifests as broken in production due to historical race condition occurrences.

I have tried to write a regression test (reproducing broken sessions state is still possible by manually manipulating session cache store). Unfortunately, I have stumbled upon an issue where Clickhouse on Mx Mac (regardless if it's native ARM binary or dockerized linux instance) behaves differently when casting negative Float64 to UInt32, namely:

On Mac:

```

04a87f302e40 :) select toUInt32(-100.)

SELECT toUInt32(-100.)

Query id: e1f39adf-c373-4fc9-a599-cf528ccc3034

   ┌─toUInt32(-100.)─┐
1. │               0 │
   └─────────────────┘

1 row in set. Elapsed: 0.031 sec. 
```

while on Linux for instance (as well as staging and production):

```

19e6aea7d224 :) select toUInt32(-100.0)

SELECT toUInt32(-100.)

Query id: 58a7d96e-e4c0-4923-9b78-757d22b87209

   ┌─toUInt32(-100.)─┐
1. │      4294967196 │ -- 4.29 billion
   └─────────────────┘

1 row in set. Elapsed: 0.007 sec. 
```

Becuase of this difference in behavior I cannot make the regression to actually fail when running it on Mac. Regardless, I'm leaving the test as it reliably fails on CI when regression is reintroduced.

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
```diff
@@ -3156,6 +3156,63 @@ defmodule PlausibleWeb.Api.ExternalStatsController.BreakdownTest do
                ]
              }
     end
+
+    test "bounce rate calculation handles invalid session data gracefully", %{
```
macobo, line 3160: Note: This module will likely disappear over time. Prefer test/plausible_web/controllers/api/external_stats_controller/query_test.exs in the future!

macobo (approved):

test/plausible_web/controllers/api/external_stats_controller/breakdown_test.exs
zoldar: https://github.com/plausible/analytics/pull/4555/commits/aa5c9eb3dcd83ba8e34b5ce226241e9b1c34a674

PR #4558, opened 2024-09-09
Update to the copy

metmarkosaric:
Update to the copy to include a note about billable pageviews. hopefully it doesn't make the copy too long...

### DIFF
--- a/lib/plausible_web/live/installation.ex
+++ b/lib/plausible_web/live/installation.ex
@@ -295,7 +295,7 @@ defmodule PlausibleWeb.Live.Installation do
         <label for={"check-#{@variant}"}>
           <%= @label %>
         </label>
-        <div class="ml-2">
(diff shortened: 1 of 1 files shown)


zoldar:
Looks ok in desktop view but feels a bit too spread vertically on mobile now?

https://github.com/user-attachments/assets/0e20888d-1eb8-4c0a-a4da-120ed26a3ef2

metmarkosaric:
true @zoldar! any way to make it nicer?

zoldar:
@metmarkosaric It's a bit tricky as the tooltip icon is already a link to the external page. As there's no other way to trigger the tooltip than by touch in this context, it's kind of pointless IMHO. Maybe we should do away with tooltip entirely here for smaller screens?

metmarkosaric:
@zoldar sure, that works

zoldar (approved):

PR #4565, opened 2024-09-10
APIv2: Make imports_skip_reason flag not appear when imports not requested

macobo:
Minor change to the aesthetics of the response. No reason to include this when imports were not requested.

Noticed while updating the docs.


lib/plausible/stats/query_result.ex
```diff
@@ -74,7 +74,10 @@ defmodule Plausible.Stats.QueryResult do
     %{
       imports_included: if(query.include.imports, do: query.include_imported, else: nil),
       imports_skip_reason:
-        if(query.skip_imported_reason, do: Atom.to_string(query.skip_imported_reason), else: nil),
+        if(query.include.imports and query.skip_imported_reason,
+          do: Atom.to_string(query.skip_imported_reason),
+          else: nil
+        ),
```
aerosol, line 80:
```suggestion
        if(query.include.imports and query.skip_imported_reason,
          do: to_string(query.skip_imported_reason)
        ),
```

aerosol (approved):
minor cosmetic suggestion inline

PR #4568, opened 2024-09-11
Make CE HTTPS use current Mozilla recommendations

ruslandoga:
### Changes

This PR continues #4491 and makes CE HTTPS follow [Mozilla recommendations](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29) for "Intermediate" compatibility.

---

ChatGPT [says](http://chatgpt.com/share/199c7def-9e22-45dd-8862-278a7b14d0f3) this configuration suite would cover approximately 95-98% of global web clients.

> This configuration would cover the vast majority of browsers, mobile devices, and modern clients, with only some very old clients (e.g., Windows XP or older Android versions) potentially being excluded.

It also excludes iPhones with iOS 8 and below.

---

Compared to Plug's `compatible` suite (used in https://github.com/plausible/analytics/pull/4491), this configuration:
- supports TLS 1.3
- drops TLS 1, TLS 1.1
- enforces eccs order
- supports x25519 ecc
- uses currently recommended ciphers

---

Current SSL Labs results: https://www.ssllabs.com/ssltest/analyze.html?d=who.copycat.fun

---

Compared to [Plausible Cloud,](https://www.ssllabs.com/ssltest/analyze.html?d=plausible.io) this configuration:
- drops ciphers that SSL Labs marks as WEAK and that Mozilla doesn't recommend
- uses ecdsa instead of rsa for keys (mostly because it is the new certbot default)


config/runtime.exs
```diff
@@ -329,14 +329,93 @@ config :plausible, PlausibleWeb.Endpoint,
 # maybe enable HTTPS in CE
 if config_env() in [:ce, :ce_dev, :ce_test] do
   if https_port do
-    https_opts = [
-      port: https_port,
-      ip: listen_ip,
-      cipher_suite: :compatible,
```
ruslandoga, line 335: Plug's `compatible` cipher suite is not compatible enough: https://www.ssllabs.com/ssltest/analyze.html?d=who.copycat.fun

I'd like the defaults for CE to be similar to https://plausible.io: https://www.ssllabs.com/ssltest/analyze.html?d=plausible.io

config/runtime.exs
```diff
@@ -329,12 +329,33 @@ config :plausible, PlausibleWeb.Endpoint,
...
-      transport_options: [socket_opts: [log_level: :warning]]
-    ]
+    # the following configuration is based on https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29
+    # except we make the server choose the cipher preference
```
ruslandoga, line 333: Because that's what https://plausible.io (or rather Bunny CDN) is doing.

config/runtime.exs
```diff
@@ -329,12 +329,35 @@ config :plausible, PlausibleWeb.Endpoint,
...
+        eccs: [:x25519, :secp256r1, :secp384r1],
+        supported_groups: [:x25519, :secp256r1, :secp384r1],
+        ciphers: [
+          # Mozilla recommended cipher suites (TLS 1.3)
+          ~c"TLS_AES_128_GCM_SHA256",
+          ~c"TLS_AES_256_GCM_SHA384",
+          ~c"TLS_CHACHA20_POLY1305_SHA256",
+          # Mozilla recommended cipher suites (TLS 1.2)
+          ~c"ECDHE-ECDSA-AES128-GCM-SHA256",
+          ~c"ECDHE-RSA-AES128-GCM-SHA256",
```
ruslandoga, line 351: RSA ones can be removed since certbot generates ECDSA keys by default, and we don't provide an option to configure it.

ruslandoga: https://github.com/plausible/analytics/pull/4568/commits/16f84d25be2534886766a134c98cebfd25d762bb

zoldar (approved):

PR #4570, opened 2024-09-11
Limit docs query API, return 401 for API routes

macobo:
### Changes

After this:
- Docs Query API routes are not queriable by viewers/public
- API routes return a JSON error over html page

Solves https://3.basecamp.com/5308029/buckets/37310803/card_tables/cards/7782351091

There might be a follow-up docs change to tweak behavior when user has insufficient access or isn't logged in.

### Tests

### Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ All notable changes to this project will be documented in this file.
 - Improve Google error messages in CE plausible/analytics#4485
 - Better compress static assets in CE plausible/analytics#4476
 - Return domain-less cookies in CE plausible/analytics#4482
-- Internal stats API routes now return 401 over 404 in case of invalid access.
+- Internal stats API routes now return a JSON error over HTML in case of invalid access.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 - Improve Google error messages in CE plausible/analytics#4485
 - Better compress static assets in CE plausible/analytics#4476
 - Return domain-less cookies in CE plausible/analytics#4482
+- Internal stats API routes now return 401 over 404 in case of invalid access.
 
 ### Fixed
 
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -66,7 +66,10 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
   defp fail(conn) do
     case get_format(conn) do
       "json" ->
-        PlausibleWeb.Api.Helpers.unauthorized(conn, "User does not have sufficient access.")
+        PlausibleWeb.Api.Helpers.not_found(
+          conn,
+          "Site does not exist or user does not have sufficient access."
+        )
 
       _ ->
         PlausibleWeb.ControllerHelpers.render_error(conn, 404) |> halt
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -1,5 +1,6 @@
 defmodule PlausibleWeb.AuthorizeSiteAccess do
   import Plug.Conn
+  import Phoenix.Controller, only: [get_format: 1]
   use Plausible.Repo
 
   def init([]), do: [:public, :viewer, :admin, :super_admin, :owner]
@@ -18,7 +19,7 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
       shared_link_auth && Repo.get_by(Plausible.Site.SharedLink, slug: shared_link_auth)
 
     if !site do
-      PlausibleWeb.ControllerHelpers.render_error(conn, 404) |> halt
+      fail(conn)
     else
       user_id =
         case PlausibleWeb.UserAuth.get_user_session(conn) do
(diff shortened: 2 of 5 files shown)


lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -2,10 +2,16 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
...
 
-  def init([]), do: [:public, :viewer, :admin, :super_admin, :owner]
-  def init(allowed_roles), do: allowed_roles
+  alias PlausibleWeb.Api.Helpers, as: H
 
-  def call(conn, allowed_roles) do
+  def init(options) do
+    {
+      Keyword.get(options, :allowed_roles, [:public, :viewer, :admin, :super_admin, :owner]),
+      Keyword.get(options, :api, false)
```
apata, line 10: **meganitpick, non-blocking**: api is a bit general term: it won't be obvious that the error response style (html or json) is wired on it. 

If we need to branch for error formats, I'd advise respecting the request "accept" header. 

To prevent JSON APIs from serving html errors, I'd also reject requests that don't carry the accept header expected for the route. It's a bit of work adding the request accept validation to each pipeline / route, but it's standard (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406).

apata (approved):

lib/plausible_web/plugs/authorize_site_access.ex
macobo: What you're suggesting makes sense but seems rabbit-holey. I'll keep this in mind but not go down this route (yet).

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -2,10 +2,16 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
   import Plug.Conn
   use Plausible.Repo
 
-  def init([]), do: [:public, :viewer, :admin, :super_admin, :owner]
-  def init(allowed_roles), do: allowed_roles
+  alias PlausibleWeb.Api.Helpers, as: H
```
aerosol, line 5: just import?

lib/plausible_web/plugs/authorize_site_access.ex
aerosol: Whether it comes from the API or not should be differentiated based on conn rather than compile-time marker?

Halting the connection with HTTP status should result with error view being applied based on accept header as @apata suggests (via `PlausibleWeb.ErrorView`, no?).

macobo: @aerosol sorry I'm lost, what is it you're suggesting?

aerosol: let me try something, will get back to you

zoldar: We can use `(Phoenix.Controller).get_format(conn)` to get the negotiated content type set by `:accept` plug, with a fallback, just in case, like this:

```elixir
format = get_format(conn) || "html"
```

and use it to determine which error format to send ("json" or "html" in this case) - that should remove the need for explicit `api` option I think. We could go even further and modify `PlausibleWeb.ControllerHelpers.render_error/2,3` to detect format on its own. However, as is, this plug returns 404 for web/HTML and 401 for API/JSON - maybe we could settle on 404 for both for instance?

macobo: Thanks @zoldar! I was initially hesitant since we hadn't done anything like this in the codebase and I was unsure of the complexity. Your proposal helps though.

> However, as is, this plug returns 404 for web/HTML and 401 for API/JSON - maybe we could settle on 404 for both for instance?

I'd argue it makes sense. Consider the two usecases:
1. User is not logged in and tries to open a settings page - 404
2. User is logged in and does not have access to that site - 404 still makes sense but is a bit more dubious
3. Someone has logged in but browser requests internal stats API - I'd argue 401 makes more sense, but can be convinced otherwise.
4. Logged in user tries to use [Stats APIv2 playground](https://plausible.io/docs/stats-api-playground) but mistypes the site_id - returning 404 wouldn't make sense in that scenario and would confuse the UX of the playground. A 401 - Forbidden with an error message makes more sense.

zoldar: What's the difference between 2 and 3? Anyways, in all 4 cases we are talking about authenticated user who has _no_ access to a given site (regardless whether they try to request it delibarately or due to a typo), correct? The whole point of returning 404 is to avoid leaking the fact that stats for a given site are in the system to any unauthorized party. With that in mind, I'd argue 404 is the appropriate error to return in all those cases.

macobo: I've changed it to return 404 over 401 now and changed the error message to accomodate. It still feels off but simplifies the PR significantly. Thanks!

zoldar (approved):

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 - Improve Google error messages in CE plausible/analytics#4485
 - Better compress static assets in CE plausible/analytics#4476
 - Return domain-less cookies in CE plausible/analytics#4482
+- Internal stats API routes now return 401 over 404 in case of invalid access.
```
apata, line 42: typo: not accurate any more, right?

apata (changes requested):

CHANGELOG.md
zoldar: ah, well spotted 🙈

apata:
```suggestion
```

macobo: Good eyes. Changed this to `Internal stats API routes now return a JSON error over HTML in case of invalid access.`

PR #4571, opened 2024-09-11
APIv2: Support `timezone` query parameter 

macobo:
This parameter determines what time zone the query result is in and how custom date ranges are parsed.

Mentioned in a user request.


macobo:
Discussed this with Uku and decided to close this since:
1. This might interfere with future pre-aggregation plans
2. We don't completely understand the usecase

PR #4572, opened 2024-09-11
APIv2: pagination

macobo:
### Changes

Depends on https://github.com/plausible/analytics/pull/4571 to avoid merge conflicts.

This PR:
- Adds new `pagination` clause to queries to limit/offset results
- Adds new `include.total_rows` option which includes total number of rows under `meta.total_rows` in apiv2

Uses offset-based pagination for efficiency together with upcoming Looker integration

Documentation will be added separately.

### DIFF
--- a/lib/plausible/stats/filters/query_parser.ex
+++ b/lib/plausible/stats/filters/query_parser.ex
@@ -5,7 +5,13 @@ defmodule Plausible.Stats.Filters.QueryParser do
 
   @default_include %{
     imports: false,
-    time_labels: false
+    time_labels: false,
+    total_rows: false
+  }
+
+  @default_pagination %{
+    limit: 10_000,
+    offset: 0
   }
 
   def parse(site, schema_type, params, now \\ nil) when is_map(params) do
@@ -22,6 +28,7 @@ defmodule Plausible.Stats.Filters.QueryParser do
          {:ok, dimensions} <- parse_dimensions(Map.get(params, "dimensions", [])),
          {:ok, order_by} <- parse_order_by(Map.get(params, "order_by")),
          {:ok, include} <- parse_include(Map.get(params, "include", %{})),
+         {:ok, pagination} <- parse_pagination(Map.get(params, "pagination", %{})),
          preloaded_goals <- preload_goals_if_needed(site, filters, dimensions),
          query = %{
            metrics: metrics,
@@ -31,7 +38,8 @@ defmodule Plausible.Stats.Filters.QueryParser do
            order_by: order_by,
            timezone: timezone,
            preloaded_goals: preloaded_goals,
-           include: include
+           include: include,
+           pagination: pagination
          },
          :ok <- validate_order_by(query),
          :ok <- validate_custom_props_access(site, query),
@@ -310,18 +318,15 @@ defmodule Plausible.Stats.Filters.QueryParser do
   defp parse_order_direction(entry), do: {:error, "Invalid order_by entry '#{i(entry)}'."}
 
   defp parse_include(include) when is_map(include) do
-    with {:ok, parsed_include_list} <- parse_list(include, &parse_include_value/1) do
-      include = Map.merge(@default_include, Enum.into(parsed_include_list, %{}))
-
-      {:ok, include}
-    end
+    {:ok, Map.merge(@default_include, atomize_keys(include))}
   end
 
-  defp parse_include_value({"imports", value}) when is_boolean(value),
-    do: {:ok, {:imports, value}}
+  defp parse_pagination(pagination) when is_map(pagination) do
+    {:ok, Map.merge(@default_pagination, atomize_keys(pagination))}
+  end
 
-  defp parse_include_value({"time_labels", value}) when is_boolean(value),
-    do: {:ok, {:time_labels, value}}
+  defp atomize_keys(map),
+    do: Map.new(map, fn {key, value} -> {String.to_existing_atom(key), value} end)
 
   defp parse_filter_key_string(filter_key, error_message \\ "") do
     case filter_key do
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -20,9 +20,11 @@ defmodule Plausible.Stats.Query do
             preloaded_goals: [],
             include: %{
               imports: false,
-              time_labels: false
+              time_labels: false,
+              total_rows: false
             },
-            debug_metadata: %{}
+            debug_metadata: %{},
+            pagination: nil
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{Filters, Imported, Legacy}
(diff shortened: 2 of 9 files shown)


apata:
If we're indeed moving the dashboard to API v2, I have concerns with offset pagination. If the query period includes the present moment, and there's a continuous stream of events in between page requests, then the responses start including duplicates. See here for test case: https://github.com/plausible/analytics/compare/apiv2-pagination...apiv2-pagination-dups?expand=1.

This would not be rare either, as we support several query periods that include the present moment. 

The way to avoid this is to have cursor based pagination: in its simplest form (_I haven't yet worked out whether this particular form would work for our data_), we'd return to the client the data timestamp that the current page ends with. The client then uses that timestamp cursor in requests for the next page. In the backend we'd load entries starting from that timestamp.

apata:
Alternatively, staying with offset pagination, when we get page one (offset 0) for query periods including the present moment, the BE can determine what the present moment is for this request, and return it next to data. 

For subsequent pages, the client would then include that moment in the request, and the BE will cap the otherwise moving query period to that point. For example, it'd treat the period of Month to Date as Month to Date Up To Moment in Request. 

That would leave it up to the client to decide when the moment is too stale for a particular view and when to start request from offset 0 again.

macobo:
Thanks for the feedback @apata! I haven’t written down all the background for this technical decision so let’s take the opportunity to do so.

**Background - ClickHouse constraints**

The bottleneck for our API is ClickHouse and the amount of data read from there. For most (and costliest) arbitrary queries, LIMIT/OFFSET does not speed the queries up - it still needs to read the whole relevant dataset and aggregate things together. Luckily ClickHouse is great at that sort of workload, but this informs how we do pagination in that larger limits are better than many small requests.

Note if there’s hotspots in the future we may want to create e.g. materialized views to pre-aggregate or elsehow pre-index some frequent queries, but this won’t be needed for a long while yet.

**Background - Looker integration**

One of the motivations for the APIv2 project was building a better looker integration (in the works). When speaking with the contractor, he mentioned a technical constraint in Looker - doing sequential requests (as you would with cursors) is slow since there’s a significant overhead (~1s) between each request.

As such, getting the total number of rows to request and being able to fire off all subsequent requests at once required for a good user experience.

**Background - Analytical calculations**

Analytical queries are mostly calculated on the fly - unlike e.g. product pages or other CRUD type datasets, the results of the queries are not precalculated or -determined. 

I am unaware of any way to efficiently implement cursor-based pagination for analytics queries without adding a lot of extra complexity to our backend. 

However the same problem can be addressed by using the same exact *past* time range between requests with the assumption that no past data is being ingested.

**Background - APIv1 vs APIv2 design philosophy**

This is less important than other constraints. 

Stats APIv1 used limit/page-no for pagination. APIv2 is trying to be more SQL and LISP-like to simplify querying for our users. As such I’ve opted for limit/offset over limit/page

---

With this context, I hope you agree that LIMIT/OFFSET makes sense for APIv2.

Your feedback seems to be partially motivated around the question of how to provide great user experience in the details model with the load more button.

My proposal for solving that: Let’s cheat in the UI. Let’s load up to 10k items into the modal but only show as much as makes sense for the user. If they click more (or “show all”) we load data from memory if possible and maybe do a subsequent request if really needed.

If that doesn’t work we can also start returning the *exact* datetime of the request in query response `meta.query.date_range[1]`. This way subsequent requests can use that as needed. But I would discourage from going down that route due to it being an unplanned feature.

apata:
Thank you for the detailed explanation! There's definitely enough reasons to justify going with limit / offset for now.

With offset pagination and the planned request strategy for Looker, supposing a complex query, total output count of 1000 and limit of 100, wouldn't we be starting ~9 concurrent attempts to compute the data, even though we've just computed it once already? 

It seems to follow that the computation result would ideally be temporarily persisted. If we pass the reference to this temporarily persisted result to the client and the client includes it in subsequent requests, recomputes could be avoided. Note that the client could still paginate through this persisted result using limit/offset.

macobo:
> With offset pagination and the planned request strategy for Looker, supposing a complex query, total output count of 1000 and limit of 100, wouldn't we be starting ~9 concurrent attempts to compute the data, even though we've just computed it once already?
> 
> It seems to follow that the computation result would ideally be temporarily persisted. If we pass the reference to this temporarily persisted result to the client and the client includes it in subsequent requests, recomputes could be avoided. Note that the client could still paginate through this persisted result using limit/offset.

The default limit in this PR is 10000, which should be enough for most purposes. If users are going a lot further beyond that, it might indicate they're using the product in an unintended way and we should figure out if and how to support that.

Also while theoretically nice, advanced caching isn't easy to build/maintain with our limited resources. I also posit there's a lot more bang-for-buck utilizing the database well and tuning it and our schemas over increasing the complexity a lot on application side with application-side caching solutions.

Aside: immediate subsequent queries are already faster thanks to clickhouse page cache as long as the queries hit the same server.

apata (approved):

ukutaht:
I'm struggling to imagine how timestamp-based cursor pagination would even work...let's say my 'Top Pages' report for today this

/page1 - 3 pageviews
/page2 - 2 pageview
/page3 - 1 pageview

Now if I request with page size 2 I will get the first two entries and the current timestamp as the cursor. I want `/page3` to show up in the second page as well, even though the event didn't happen in the timeframe since the last request

apata:
> I'm struggling to imagine how timestamp-based cursor pagination would even work...let's say my 'Top Pages' report for today this
> 
> /page1 - 3 pageviews /page2 - 2 pageview /page3 - 1 pageview
> 
> Now if I request with page size 2 I will get the first two entries and the current timestamp as the cursor. I want `/page3` to show up in the second page as well, even though the event didn't happen in the timeframe since the last request

You're right that timestamp based cursor would not work for the aggregations we serve. 

It's just theory, because the reason for limit and offset is now well-explained and the bottleneck is probably elsewhere, but for cursor-paginating a list ordered by descending pageviews, the cursor could be the pageviews and name of the last item served to the client. 

The results for the next page will be where pageviews <= cursor.pageviews and name is not cursor.name.

ukutaht:
Right yeah, that would work.

I don't have a strong opinion either way on this. Normal performance concerns do not apply here because we're paginating the results of a GROUP BY which needs to be executed fully regardless of the page that's requested. So a cursor would not reduce the amount of data scanned as usual. It might have some benefits in getting a more stable result set I suppose.

Since there's a strong use-case for OFFSET/LIMIT in the Looker Studio connector, it makes sense to go with that.

apata:
> Right yeah, that would work.
> 
> I don't have a strong opinion either way on this. Normal performance concerns do not apply here because we're paginating the results of a GROUP BY which needs to be executed fully regardless of the page that's requested. So a cursor would not reduce the amount of data scanned as usual. It might have some benefits in getting a more stable result set I suppose.
> 
> Since there's a strong use-case for OFFSET/LIMIT in the Looker Studio connector, it makes sense to go with that.

apata:
Oops

macobo:
Oops x2 - sorry did not intend to merge, will open a new PR :sweat_smile:

PR #4574, opened 2024-09-12
Automatically generate Typescript types for v2 API query schema 

apata:
### Changes

This proposal makes use of the query schema enforced in the backend for API v2 to generate types for the dashboard. By using these autogenerated types, we can more easily ensure that the dashboard is in sync with the backend. 

### Tests

### Changelog

### Documentation

### Dark mode


.github/workflows/node.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         node-version: ${{steps.versions.outputs.nodejs}}
     - run: npm install --prefix ./assets
     - run: npm install --prefix ./tracker
+    - run: npm run generate-types --prefix ./assets && git diff --exit-code -- ./assets/js/types
```
apata, line 28: Note: This particular line regenerates the types and compares the output with the current commit, failing if there are any differences. This (rather clumsily) ensures that the generated types are always in sync with the schema.

apata:
This is currently generating types from the v2 API **public schema**, since that is a static json file. For the dashboard, we'd actually want to generate types from the private schema, which is currently created by the server at compile time. 

To fix, the easiest thing would be to have the private schema in the static json file and create the public schema dynamically from it.

assets/js/types/query-api.d.ts
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Domain of site to query
+   */
+  site_id: string;
+  /**
+   * List of metrics to query
+   *
+   * @minItems 1
+   */
+  metrics: [Metric, ...Metric[]];
```
macobo, line 108: Why not `Metric[]`? :thinking:

assets/js/types/query-api.d.ts
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Date range to query
+   */
+  date_range: "all" | "day" | "7d" | "30d" | "month" | "6mo" | "12mo" | "year" | [string | string, string | string];
```
macobo, line 112: I wonder if we can improve the generated types here? E.g. `[ISODate, IsoDate] | [ISOTimestamp, ISOTimestamp]` and use e.g. branded types for both?

assets/js/types/query-api.d.ts
```diff
@@ -0,0 +1,130 @@
...
+   */
+  filters?: FilterTree[];
+  /**
+   * How to order query results
+   */
+  order_by?: OrderByEntry[];
+  include?: {
+    time_labels?: boolean;
+    imports?: boolean;
+    [k: string]: unknown;
```
macobo, line 128: Hmm. Unknown include options shouldn't be valid - bug in the schema file likely?

macobo (commented):
This is really cool! Left some initial thoughts :heart_eyes: 

Also feel free to flip the JSON schema file being internal in this PR.

assets/js/types/query-api.d.ts
apata: I suspect Metric[] permits empty array

assets/js/types/query-api.d.ts
apata: Good idea, I suspect we can alias strings, checking!

macobo: You can, e.g. `type ISODate = string`.

However branded types take it a step further providing more type-safety after casting an existing string:  https://egghead.io/blog/using-branded-types-in-typescript

lib/plausible/stats/json_schema.ex
```diff
@@ -8,33 +8,34 @@ defmodule Plausible.Stats.JSONSchema do
...
 
-  @raw_public_schema Application.app_dir(:plausible, "priv/json-schemas/query-api-schema.json")
-                     |> File.read!()
-                     |> Jason.decode!()
+  @raw_internal_schema Application.app_dir(:plausible, "priv/json-schemas/query-api-schema.json")
+                       |> File.read!()
+                       |> Jason.decode!()
 
-  @public_query_schema ExJsonSchema.Schema.resolve(@raw_public_schema)
+  @raw_public_schema (with {:ok, s1, _} <-
```
macobo, line 15: This is hard to read and the error handling isn't valuable since errors here would end up as compile-time errors anyways.

Suggestion:

```elixir
defp remove!(schema, pointer) do
  {:ok, schema} = JSONPointer.remove(schema, pointer)
  schema
end
```

And set this up as a chain instead.

lib/plausible/stats/json_schema.ex
```diff
@@ -8,33 +8,34 @@ defmodule Plausible.Stats.JSONSchema do
...
-                         })
-                         |> JSONPointer.add!("#/definitions/date_range/oneOf/0", %{
-                           "const" => "30m"
-                         })
-                         |> JSONPointer.add!("#/definitions/date_range/oneOf/0", %{
-                           "const" => "realtime"
-                         })
-                         |> JSONPointer.add!("#/properties/date", %{"type" => "string"})
-                         |> ExJsonSchema.Schema.resolve()
+  @internal_query_schema ExJsonSchema.Schema.resolve(@raw_internal_schema)
```
macobo, line 37: Note: If I understand things correct variable is kept in memory. As such if we can get away with using only 3 variables over 4 (as previously) let's prefer that approach.

.github/workflows/node.yml
macobo: Note you might also want to add this to the precommit hooks? See: .pre-commit-config.yaml

priv/json-schemas/query-api-schema.json
```diff
@@ -56,7 +64,52 @@
...
+      "type": "array",
+      "additionalItems": false,
+      "minItems": 2,
+      "maxItems": 2,
+      "markdownDescription": "A list of two ISO8601 dates or datetimes.",
+      "examples": [
+        ["2024-09-01", "2024-09-30"],
+        ["2024-01-01T00:00:00+03:00", "2024-01-02T12:00:00+03:00"]
+      ],
+      "items": [
```
macobo, line 76: This new schema is incorrect - you should not mix [date, datetime] and [datetime,date].

lib/plausible/stats/json_schema.ex
apata: Thanks! I'm not super familiar with Elixir patterns yet, and after checking the [docs for module attributes](https://hexdocs.pm/elixir/module-attributes.html) I still don't fully get it. 

Why are we using module attributes for this in the first place? 

Is it because we want to be sure whether our schema variants are parseable and resolvable? Is that it, or is there a server memory optimization aspect to it as well?

lib/plausible/stats/json_schema.ex
apata: Thanks! That would have been a good change from the previous, but I couldn't really trust that we could reliably maintain the position-based remove strategy. I'm now omitting JSON nodes that have "$comment": "private".

priv/json-schemas/query-api-schema.json
apata: Roger that, managed to fix:
Screenshot 2024-09-17 at 14 24 39

lib/plausible/stats/json_schema.ex
ukutaht: My take on this: the main reason to use module attributes in general is because they are evaluated at compile-time as opposed to runtime. So by putting it a module attribute we can do the file read, json parsing, etc at compile time so it doesn't run on every request.

macobo: Exactly what uku said :)

lib/plausible/stats/json_schema_tools.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Plausible.Stats.JSONSchemaTools do
```
macobo, line 1: I'm not a fan of this separate module under Plausible.Stats since it's not related to stats at all and is closely coupled with json_schema.ex. I'd inline it to the other module or figure out a better organization for this.

lib/plausible/stats/json_schema.ex
```diff
@@ -5,37 +5,20 @@ defmodule Plausible.Stats.JSONSchema do
...
 
-  @raw_public_schema Application.app_dir(:plausible, "priv/json-schemas/query-api-schema.json")
-                     |> File.read!()
-                     |> Jason.decode!()
-
+  @raw_internal_schema Application.app_dir(:plausible, "priv/json-schemas/query-api-schema.json")
+                       |> File.read!()
+                       |> Jason.decode!()
+  @raw_public_schema JSONSchemaTools.traverse(@raw_internal_schema, fn
+                       %{"$comment" => "private"} -> :remove
```
macobo, line 16: `private` is a bad name for this given it's for the `internal_schema`. Suggestion: make the comment `internal_schema_only` for clarity.

lib/plausible/stats/json_schema_tools.ex
```diff
@@ -0,0 +1,44 @@
...
+  @moduledoc """
+  Module for traversing and modifying JSON schemas.
+  """
+
+  @type json :: map() | list() | String.t() | number() | boolean() | nil
+  @type transform_fun :: (json() -> json() | :remove)
+
+  @spec traverse(map(), transform_fun()) :: map() | :remove
+  def traverse(json, fun) when is_map(json) do
```
macobo, line 10: While a nice abstraction, We don't right now need all the power this exposes and I am not sure we ever will. Let's  expose a single function named something like `remove_by_comment`.

test/plausible_web/controllers/api/internal_controller/schema_for_docs_test.exs
```diff
@@ -0,0 +1,19 @@
...
+      conn = get(conn, "/api/docs/query/schema.json")
+      response = json_response(conn, 200)
+
+      assert %{"$schema" => "http://json-schema.org/draft-07/schema#", "type" => "object"} =
+               response
+    end
+
+    test "schema does not contain nodes with private comments", %{conn: conn} do
+      conn = get(conn, "/api/docs/query/schema.json")
+      refute response(conn, 200) =~ ~s/"$comment":"private"/
```
macobo, line 16: Nit/suggestion: 
```suggestion
      refute response(conn, 200) =~ ~s/"realtime"/
```

apata: Hmm, that doesn't prove that all intended nodes were removed.

lib/plausible/stats/json_schema.ex
macobo: I'll look into perhaps doing this in a macro in a follow-up.

lib/plausible/stats/json_schema_tools.ex
macobo: Proposal: `Plausible.Stats.JSONSchema.Utils` - put it in a subfolder?

test/plausible_web/controllers/api/internal_controller/schema_for_docs_test.exs
apata: Makes sense!

lib/plausible/stats/json_schema.ex
apata: True! Went with "only :internal", so a search for the atom will turn it up

macobo (approved):

PR #4580, opened 2024-09-16
Dashboard: Fix countries/regions/cities tooltips

macobo:
Previously these tooltips said `Add filter: Country is undefined` because new labels were yet to be in `query`.

cc @apata - curious on your thoughts on if or how best test this

Basecamp: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7816925449


apata (approved):
On testing, besides unit tests for adding and subtracting all kinds of filters, for this particular case I'd probably go one level up and render the ListReport with react test library, then expect the links to be present with the appropriate title. See query-dates.test.tsx for an example how to use react testing library for this.

PR #4582, opened 2024-09-16
Drop event explicitly on session lock timeout 

zoldar:
### Changes

Currently, the `CacheStore.on_event` call for timed out lock case returns `:ok` instead of session, which leads to a pretty opaque error when trying to merge session data with event.

This change makes timeout handling explicit and drops the timed out event with `:lock_timeout` set as reason.

### Tests

### DIFF
--- a/lib/plausible/cache/adapter.ex
+++ b/lib/plausible/cache/adapter.ex
@@ -113,11 +113,18 @@ defmodule Plausible.Cache.Adapter do
       []
   end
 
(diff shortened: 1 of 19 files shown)


aerosol (approved):
LGTM, although I like to understand why the intermittent test failures are still visible on Linux test runs specifically. The probability of congestion is much higher during tests, however it feels like something related to scheduling.

PR #4583, opened 2024-09-16
Refactor `AuthorizeSiteAccess`

zoldar:
### Changes

This PR is refactoring `AuthorizeSiteAccess`, adding some extra compile- and runtime checks along the way. The range of accepted inputs was revised as well.

### Tests

### DIFF
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -3,12 +3,24 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
   Plug restricting access to site and shared link, when present.
   """
 
-  import Plug.Conn
-  import Phoenix.Controller, only: [get_format: 1]
   use Plausible.Repo
 
-  def init([]), do: [:public, :viewer, :admin, :super_admin, :owner]
-  def init(allowed_roles), do: allowed_roles
+  import Plug.Conn
+  import Phoenix.Controller, only: [get_format: 1]
+
+  @all_roles [:public, :viewer, :admin, :super_admin, :owner]
+
+  def init([]), do: @all_roles
+
+  def init(allowed_roles) do
+    unknown_roles = allowed_roles -- @all_roles
+
+    if unknown_roles != [] do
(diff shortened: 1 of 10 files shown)


lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -58,21 +60,51 @@ defmodule PlausibleWeb.AuthorizeSiteAccess do
...
+    end
+  end
+
+  defp maybe_get_shared_link(conn, site) do
+    slug = conn.path_params["slug"] || conn.params["auth"]
+    site_id = site.id
+
+    if is_binary(slug) do
+      case Repo.get_by(Plausible.Site.SharedLink, slug: slug) do
+        %{site_id: ^site_id} = shared_link -> {:ok, shared_link}
```
ukutaht, line 84: Can the `site_id == ^site_id` condition be applied in the SQL query itself? i.e.

```elixir
if link = Repo.get_by(Plausible.Site.SharedLink, slug: slug, site_id: site.id) do
  {:ok, link}
else
  error_not_found(conn)
end
```

ukutaht (commented):
Opportunity to remove another DB roundtrip: https://github.com/plausible/analytics/commit/9ed96632adf44c211d462ac4b144514634d1720f

I think the code in my proposal is less elegant than the PR. But since auth code runs for pretty much every controller action, I like taking some extra care with performance there.

lib/plausible_web/plugs/authorize_site_access.ex
zoldar: Thanks, fixed in https://github.com/plausible/analytics/pull/4583/commits/9db87fc2677a9acf54aaad08580f80c52205c9ab. Not sure what was going through my head when writing this...

zoldar:
> Opportunity to remove another DB roundtrip: https://github.com/plausible/analytics/commit/9ed96632adf44c211d462ac4b144514634d1720f
>
> I think the code in my proposal is less elegant than the PR. But since auth code runs for pretty much every controller action, I like taking some extra care with performance there.

Agreeed. Applied your change with a minor adjustment of naming.

ukutaht (approved):

PR #4585, opened 2024-09-17
Add feature-flagged channels UI

ukutaht:
### Changes

This PR adds dashboard reports for the new `channel` dimension.

Remaining work on channels before go-live:
* Add channels to dashboard CSV export
* Add channels to full CSV export/import
* Import channels from GA4

### Tests

### Changelog
Feature is not ready yet

### Documentation
Feature is not ready yet

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/modals/sources.js
+++ b/assets/js/dashboard/stats/modals/sources.js
@@ -22,6 +22,9 @@ const VIEWS = {
       )
     }
   },
+  channels: {
+    info: { title: 'Top Acquisition Channels', dimension: 'channel', endpoint: '/channels', dimensionLabel: 'Channel', defaultOrder: ["visitors", SortDirection.desc] }
+  },
   utm_mediums: {
     info: { title: 'Top UTM Mediums', dimension: 'utm_medium', endpoint: '/utm_mediums', dimensionLabel: 'UTM Medium', defaultOrder: ["visitors", SortDirection.desc] }
   },
--- a/assets/js/dashboard/stats/sources/source-list.js
+++ b/assets/js/dashboard/stats/sources/source-list.js
@@ -67,6 +67,41 @@ function AllSources({ afterFetchData }) {
   )
 }
 
+function Channels({ afterFetchData }) {
+  const { query } = useQueryContext();
+  const site = useSiteContext();
+
+  function fetchData() {
+    return api.get(url.apiPath(site, '/channels'), query, { limit: 9 })
+  }
+
+  function getFilterFor(listItem) {
+    return {
+      prefix: 'channel',
+      filter: ["is", "channel", [listItem['name']]]
+    }
+  }
+
+  function chooseMetrics() {
+    return [
+      metrics.createVisitors({ meta: { plot: true } }),
+      hasGoalFilter(query) && metrics.createConversionRate(),
+    ].filter(metric => !!metric)
+  }
+
+  return (
+    <ListReport
+      fetchData={fetchData}
+      afterFetchData={afterFetchData}
+      getFilterFor={getFilterFor}
+      keyLabel="Channel"
+      metrics={chooseMetrics()}
+      detailsLinkProps={{ path: sourcesRoute.path, search: (search) => search }}
+      color="bg-blue-50"
+    />
+  )
+}
+
 function UTMSources({ tab, afterFetchData }) {
   const { query } = useQueryContext();
   const site = useSiteContext();
@@ -79,7 +114,7 @@ function UTMSources({ tab, afterFetchData }) {
     utm_content: utmContentsRoute,
     utm_term: utmTermsRoute,
     }[tab]
-    
+
   function fetchData() {
     return api.get(url.apiPath(site, utmTag.endpoint), query, { limit: 9 })
   }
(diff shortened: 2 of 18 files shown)


priv/json-schemas/query-api-schema.json
```diff
@@ -190,6 +190,7 @@
         "event:page",
         "event:hostname",
         "visit:source",
+        "visit:channel",
```
macobo, line 193: Sanity check: Do we immediately want to expose this to all API users? If not, the place to add it is lib/plausible/stats/json_schema.ex @internal_query_schema

ukutaht: Good point.

Removing it from the APIs would be a bit annoying since I want to test this feature mainly through the APIs rather than dashboard endpoints which will probably go away soon. Basically I would have to rewrite a bunch of tests now and then undo that work when we add `channel` to the APIs.

I think it's OK to leave it as is because the only way for people to discover this dimension would be from APIv2 docs editor and that page is annotated with a `beta` notice.

apata: Jumping in here because I'm working on [something adjacent to this](https://github.com/plausible/analytics/pull/4574).

As far as I understand, the query schema represents to the client what they can do. Ideally, it's as accurate as possible. 

There's currently two variants of it: public and internal. I see that after this PR, there'd be four variants: public, internal, public with channels flag, internal with channels flag. Assuming I build the segments feature behind a feature flag, there'd be even more variants.

Maybe we should have static set of definitions and then compose particular schemas for particular clients? Besides branching the schemas for particular flags and internal/public combinations, I can imagine serving a lower pagination limit to low tier customers for example.

ukutaht: I may be misunderstanding but in my PR there are still only two schemas: private and public. I just added the `channel` dimension to both, feature flag does not affect the schema.

I agree it would make sense if it did, though!

In Elixir we could massage the schema at runtime based on feature flags. But I'm not sure the typescript types PR would work if the schema is dynamic at runtime.

macobo: I don't think we should to worry about feature-flagging the schema - note it's all compiled.

I think doing it like this is fair given we're fast approaching a point where this is stable, just wanted to make sure we're aware of the trade-off. :)

Merging master to get the preview env working and looking at UI/UX then.

assets/js/dashboard/stats/modals/sources.js
```diff
@@ -22,6 +22,9 @@ const VIEWS = {
       )
     }
   },
+  channels: {
+    info: { title: 'Top Acquisition Channels', dimension: 'channel', endpoint: '/channels', dimensionLabel: 'Channel', defaultOrder: ["visitors", SortDirection.desc] }
```
macobo, line 26: Should this be channels (plural)? Having it be the only singular thing in the UI stands out.

image

assets/js/dashboard/stats/sources/source-list.js
```diff
@@ -67,6 +67,41 @@ function AllSources({ afterFetchData }) {
...
+
+  function chooseMetrics() {
+    return [
+      metrics.createVisitors({ meta: { plot: true } }),
+      hasGoalFilter(query) && metrics.createConversionRate(),
+    ].filter(metric => !!metric)
+  }
+
+  return (
+    <ListReport
```
macobo, line 93: Bug: Searching for Direct doesn't work in this report.

macobo (commented):
Looks great to me, small issues highlighted in comments. Once these are resolved, happy to :heavy_check_mark:

priv/json-schemas/query-api-schema.json
apata: Yep, I meant the 4 variants in theory, not in fact. 

The potential benefit of doubling down on the schemas in the backend and composing them (_not massaging one schema, as we do now_) for particular situations is to go from custom situational validation logic to schema-based validation logic. 

The reason I'm harping on this is because I've started sketching out the "segment as filter" logic and I see that sticking to the two variants of the schemas we already have ([see $comments in the file](https://github.com/plausible/analytics/compare/sketch-segment-is-schema?expand=1)) will likely have me writing custom validation logic in various places. 

As far as typegen goes, it's really only meant to help us write better FE code faster: we should probably generate the types for it from the most expansive public schema possible. One could compose it and write it to disk at compile time, then generate types from it.

assets/js/dashboard/stats/modals/sources.js
ukutaht: Agreed. https://github.com/plausible/analytics/pull/4585/commits/db272a27800192de834913d888e7d8f359870213

assets/js/dashboard/stats/sources/source-list.js
ukutaht: This is a wider issue than just channels. Logged it here as a more general bug: https://3.basecamp.com/5308029/buckets/26383192/card_tables/5431762648

ukutaht:
@macobo since the `contains` filter was already broken for `Direct / None` values before this PR I would consider it a separate issue to be fixed in a dedicated PR.

PR #4589, opened 2024-09-17
Fix applying city filters

RobertJoonas:
### Changes

Currently, when clicking on a city in the Locations report to filter by it, the city code will become an integer in the filter.

However, when applying the city filter from the filter modal, the city code will become a string in the filter, because the value returned from city filter suggestions is a string.

When applying filters only from the filter modal, it works. But when applying a filter first by clicking on a city in the Locations report, then editing the filter, and adding another city - it breaks because string and integer types are mixed.

This PR fixes that by returning integers instead of strings from the filter suggestions API. It's also consistent with the cities breakdown report where integers are returned.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/filter_suggestions.ex
+++ b/lib/plausible/stats/filter_suggestions.ex
@@ -119,7 +119,7 @@ defmodule Plausible.Stats.FilterSuggestions do
       city = Location.get_city(c)
 
       %{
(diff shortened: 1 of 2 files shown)


zoldar (approved):
🙌

ukutaht (approved):

PR #4597, opened 2024-09-19
Allow configuring `AuthorizeSiteAccess` plug site param

apata:
### Changes

- Allow auth site access plug to specify that domain is found in request body at some key, needed for api/docs/query endpoint where :domain or :website does not appear in path params
- Nuisance: Stop typescript --watch command clearing previous server stdout when running make server

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -84,6 +84,7 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
 
   defp get_domain(conn, site_param) do
     domain = conn.params[site_param]
+
     if is_binary(domain) do
       {:ok, domain}
     else
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -10,32 +10,29 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
 
   @all_roles [:public, :viewer, :admin, :super_admin, :owner]
 
-  def init([]), do: @all_roles
+  def init([]), do: {@all_roles, nil}
 
   def init(allowed_roles) when is_list(allowed_roles) do
+    init({allowed_roles, nil})
+  end
+
+  def init({allowed_roles, conn_params_domain_accessor})
+      when is_list(allowed_roles) do
     unknown_roles = allowed_roles -- @all_roles
 
     if unknown_roles != [] do
       raise ArgumentError, "Unknown allowed roles configured: #{inspect(unknown_roles)}"
     end
 
-    allowed_roles
+    if !is_binary(conn_params_domain_accessor) && !is_nil(conn_params_domain_accessor) do
+      raise ArgumentError, "Invalid site param given: #{inspect(conn_params_domain_accessor)}"
+    end
+
+    {allowed_roles, conn_params_domain_accessor}
   end
 
-  def init({allowed_roles, conn_params_domain_accessor})
-      when is_list(allowed_roles) and is_binary(conn_params_domain_accessor) do
-    {init(allowed_roles), conn_params_domain_accessor}
-  end
-
-  def call(conn, options) do
+  def call(conn, {allowed_roles, conn_params_domain_accessor}) do
     current_user = conn.assigns[:current_user]
-
-    {allowed_roles, conn_params_domain_accessor} =
-      if is_tuple(options) do
-        options
-      else
-        {options, nil}
-      end
 
     with {:ok, %{site: site, role: membership_role}} <-
            get_site_with_role(conn, current_user, conn_params_domain_accessor),
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -83,7 +83,8 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
   end
 
   defp get_domain(conn, site_param) do
-    if is_binary(domain = conn.params[site_param]) do
+    domain = conn.params[site_param]
+    if is_binary(domain) do
       {:ok, domain}
     else
       error_not_found(conn)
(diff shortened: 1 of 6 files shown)


lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -63,12 +76,16 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
...
+  defp get_domain_from_conn(conn, nil) do
+    conn.path_params["domain"] || conn.path_params["website"]
+  end
+
+  defp get_domain_from_conn(conn, conn_params_domain_accessor) do
+    conn.params[conn_params_domain_accessor]
+  end
+
+  defp get_site_with_role(conn, current_user, conn_params_domain_accessor) do
+    domain = get_domain_from_conn(conn, conn_params_domain_accessor)
```
zoldar, line 88: When the param is not provided at all (`site_id`, for instance), the plug will crash when executing the query, AFAICT. Is `site_id` optional here? If not, perhaps we should return 404, like in other cases?

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -22,10 +22,23 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     allowed_roles
   end
 
-  def call(conn, allowed_roles) do
+  def init({allowed_roles, conn_params_domain_accessor})
```
zoldar, line 25: IMHO explicit map instead of a tuple is better for readability.

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -12,7 +12,7 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
 
   def init([]), do: @all_roles
 
-  def init(allowed_roles) do
+  def init(allowed_roles) when is_list(allowed_roles) do
```
zoldar, line 15: 👍

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -22,10 +22,23 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     allowed_roles
   end
 
-  def call(conn, allowed_roles) do
+  def init({allowed_roles, conn_params_domain_accessor})
+      when is_list(allowed_roles) and is_binary(conn_params_domain_accessor) do
```
zoldar, line 26: Maybe `domain_param`? Or even `site_param`?

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -22,10 +22,23 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
...
+  def call(conn, options) do
     current_user = conn.assigns[:current_user]
 
-    with {:ok, %{site: site, role: membership_role}} <- get_site_with_role(conn, current_user),
+    {allowed_roles, conn_params_domain_accessor} =
+      if is_tuple(options) do
+        options
+      else
+        {options, nil}
+      end
```
zoldar, line 38: Instead of pattern matching in `call` we could return consistent/normalized args from `init` already (via map?)

apata: Ah, I tried that but then that would have me refactor the args to AuthoriseSiteAccess.call

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -63,12 +76,16 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     end
   end
 
-  defp get_site_with_role(conn, current_user) do
-    # addition is flimsy, do we need an extra argument on plug init to control where we look for the domain?
-    domain =
-      conn.path_params["domain"] || conn.path_params["website"] ||
-        (conn.method == "POST" && conn.path_info == ["api", "docs", "query"] &&
-           conn.params["site_id"])
```
ruslandoga, line 71: 👋 

Was just passing through and saw this comment :) I wonder if we can use `:assings` on scope or route for the relevant paths and set some value there and then based on that value perform different lookups?

Alternatively, if the logic in this plug can be split in two:
- extract site plug
- authorize site plug

zoldar: Passing anonymous function to plug config is something I've been fiddling with initially, but then I got reminded `init` is evaluated at compile time, and supporting passing stuff like anonymous functions at compile time is AFAIK quite a mess 😬 (like in case on [abinsthe](https://elixirforum.com/t/escaping-anonymous-function-in-module-attribute/60664/2)).

The scope with `private` sounds neat but it also makes the configration a bit too replaced from the actual use site - and the relationship not obvious. IMHO keeping config close to the plug is better from maintenance perspective - even if the failure mode is relatively safe (missing/stale parameter would just cause rejection, no accidentally allowing the request).

ruslandoga: > The scope with private sounds neat but it also makes the configration a bit too replaced from the actual use site - and the relationship not obvious.

In my mid it's the opposite -- the use site is the router.

test/plausible_web/controllers/api/internal_controller/docs_query_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule PlausibleWeb.Api.InternalController.DocsQueryTest do
-  use PlausibleWeb.ConnCase, async: true
+  use PlausibleWeb.ConnCase, async: false
```
aerosol, line 2: what's the reason for this change?

aerosol: ah, patch env. welp.

aerosol: wait, there's no env patching in that test suite

aerosol (approved):

test/plausible_web/controllers/api/internal_controller/docs_query_test.exs
zoldar: https://github.com/plausible/analytics/pull/4597/commits/61be18219040275f9c5e70a27d1155f5385b6a69

PR #4599, opened 2024-09-19
Channel and source data updates

ukutaht:
### Changes

Some adjustments on how we handle and categorize source/channel data based on manual testing by @metmarkosaric.

Largely this includes 2 kinds of changes:
1. Changes to how source data is categorized (i.e. transforming source `perplexity.ai` to `Perplexity`). These also include a migration to apply these changes to historical data
2. Changes to how channels are derived from source data (i.e. `Telegram` should be `Social` instead of `Referral`). These will only apply going forward. Dealing with historical and already collected channel data is not in scope of this PR. Since channels are not live to customers yet it's ok to have them in a slightly broken state for a while. A follow-up PR will deal with historical channel data.

Another change is that I've introduced `SOURCE_CATEGORY_EMAIL` which is not present in the GA4 channel grouping definitions. The goal is that sessions with source=Gmail will have channel=email rather than channel=referral.

Most of the changes are tests. 

### Tests

### Documentation
- Channels will be documented once ready to launch to customers

### Dark mode


priv/ingest_repo/migrations/20240919112752_remap_sources.exs
```diff
@@ -0,0 +1,49 @@
+defmodule Plausible.IngestRepo.Migrations.RemapSources do
+  use Ecto.Migration
+
+  @mappings %{
```
ukutaht, line 4: I could have used the `@custom_sources` and `@mappings_overrides` module attributes here, but decided to duplicate the list in case it changes in the future. Don't want migrations to reference dynamic code in general.

lib/plausible/ingestion/acquisition.ex
```diff
@@ -1,10 +1,38 @@
...
+  @custom_source_categories [
+    {"hacker news", "SOURCE_CATEGORY_SOCIAL"},
+    {"yahoo!", "SOURCE_CATEGORY_SEARCH"},
+    {"gmail", "SOURCE_CATEGORY_EMAIL"},
+    {"telegram", "SOURCE_CATEGORY_SOCIAL"},
+    {"slack", "SOURCE_CATEGORY_SOCIAL"},
+    {"producthunt", "SOURCE_CATEGORY_SOCIAL"},
+    {"github", "SOURCE_CATEGORY_SOCIAL"},
+    {"steamcommunity.com", "SOURCE_CATEGORY_SOCIAL"},
+    {"statics.teams.cdn.office.net", "SOURCE_CATEGORY_SOCIAL"},
```
macobo, line 19:
```suggestion
    {"statics.teams.cdn.office.net", "SOURCE_CATEGORY_SOCIAL"}, # codespell:ignore
```

This should do the trick according to docs: https://github.com/codespell-project/codespell?tab=readme-ov-file#inline-ignore

lib/plausible/ingestion/acquisition.ex
```diff
@@ -1,10 +1,38 @@
 defmodule Plausible.Ingestion.Acquisition do
   @moduledoc false
```
macobo, line 2:
```suggestion
  @moduledoc """
  This module is responsible for figuring out aquisition channel from event referrer_source.
  
  Acquisition channel is the marketing channel where people come from and convert and help
  users to understand and improve their marketing flow.
  
  Note it uses priv/ga4-source-categories.csv as a source, which comes from XXX.
  """
```

lib/plausible/ingestion/acquisition.ex
```diff
@@ -1,10 +1,38 @@
...
   @external_resource "priv/ga4-source-categories.csv"
+  @mapping_overrides [
```
macobo, line 4: It's worth adding a comment as to why this is needed.

lib/plausible/ingestion/acquisition.ex
```diff
@@ -116,10 +148,11 @@ defmodule Plausible.Ingestion.Acquisition do
...
   end
 
-  @email_tags ["email", "e-mail", "e_mail", "e mail"]
-  defp email?(request) do
-    String.contains?(request.query_params["utm_source"] || "", @email_tags) or
-      String.contains?(request.query_params["utm_medium"] || "", @email_tags)
+  @email_tags ["email", "e-mail", "e_mail", "e mail", "newsletter"]
+  defp email?(request, source) do
+    email_source?(source) or
+      String.contains?(String.downcase(request.query_params["utm_source"] || ""), @email_tags) or
```
macobo, line 154: We are quite inconsistent with checking whether things are downcased or not in this file, which is likely a source of bugs.

Suggestion:
1. Downcase `source` in `get_channel`
2. Add function query_param(request, name) which downcases and handles `|| ""`

macobo (commented):
Looking good! I added some suggestions.

lib/plausible/ingestion/acquisition.ex
ukutaht: Thanks, the module is actually doing two completely separate things: consolidating sources and generating acquisiton channel. I think i'll split it into 2 modules and write a separate doc for why the source mapping/consolidation thing is needed at all.

lib/plausible/ingestion/acquisition.ex
ukutaht: [Done](https://github.com/plausible/analytics/pull/4599/commits/4bdd45e8dce44aa81bb46433033eaf2f7c98aa75)

lib/plausible/ingestion/acquisition.ex
ukutaht: Added a fairly comprehensive doctest to demostrate what's going on: https://github.com/plausible/analytics/pull/4599/commits/34bff5ad17a88c97c0f2e4e2d22418b0fb570d41

ukutaht: Also renamed `mapping_overrides` to `custom_sources` and extracted into a JSON file

macobo (approved):

PR #4601, opened 2024-09-20
Skip Browserless checks in CE

ruslandoga:
### Changes

Attempt at skipping Browserless checks in CE: https://github.com/plausible/analytics/pull/4459#discussion_r1758491643

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/verification/checks.ex
+++ b/lib/plausible/verification/checks.ex
@@ -12,14 +12,26 @@ defmodule Plausible.Verification.Checks do
 
   require Logger
 
-  @checks [
+  default_checks = [
     Checks.FetchBody,
     Checks.CSP,
     Checks.ScanBody,
     Checks.Snippet,
-    Checks.SnippetCacheBust,
+    Checks.SnippetCacheBust
+  ]
+
+  ee_checks = [
     Checks.Installation
   ]
+
+  checks =
+    if Plausible.ee?() do
+      default_checks ++ ee_checks
+    else
+      default_checks
+    end
+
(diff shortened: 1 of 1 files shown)


lib/plausible/verification/checks.ex
```diff
@@ -12,15 +12,27 @@ defmodule Plausible.Verification.Checks do
...
   require Logger
 
-  @checks [
+  default_checks = [
     Checks.FetchBody,
     Checks.CSP,
     Checks.ScanBody,
     Checks.Snippet,
-    Checks.SnippetCacheBust,
+    Checks.SnippetCacheBust
```
aerosol, line 20: without `Checks.Installation`, verification won't ever succeed so in that sense this is  equivalent to querying a non-existing browserless endpoint.

ruslandoga: Is there a way around it? I assumed the verification succeeds after all defined checks pass.

aerosol: sort of - but `plausible_installed?` is crucial (and even then - it's not binary, if you skim through the diagnostics module), which comes from the browserless check. Take a look at `Plausible.Verification.Diagnostics` it's where the decision about user-facing result is made.

Feels also pointless to show verification success to the user if you don't run verification.

We could consider disabling it altogether again, by removing `"live "/:website/verification", Verification, :verification, as: :site" route` in CE and following up on the consequences. `PlausibleWeb.Flows` comes to mind as the next thing to deal with...

PR #4604, opened 2024-09-23
Skip verification in CE

ruslandoga:
### Changes

Another attempt at https://github.com/plausible/analytics/pull/4459#discussion_r1758491643

### Tests

### Changelog

### Documentation

### Dark mode
Screenshot 2024-09-23 at 17 14 21
Screenshot 2024-09-23 at 17 14 36

### DIFF
--- a/lib/plausible_web/live/verification.ex
+++ b/lib/plausible_web/live/verification.ex
@@ -1,9 +1,10 @@
 defmodule PlausibleWeb.Live.Verification do
   @moduledoc """
   LiveView coordinating the site verification process.
-  Onboarding new sites, renders a standalone component. 
+  Onboarding new sites, renders a standalone component.
   Embedded modal variant is available for general site settings.
   """
+  use Plausible
   use PlausibleWeb, :live_view
   use Phoenix.HTML
 
@@ -28,13 +29,14 @@ defmodule PlausibleWeb.Live.Verification do
     private = Map.get(socket.private.connect_info, :private, %{})
 
     super_admin? = Plausible.Auth.is_super_admin?(socket.assigns.current_user)
+    has_pageviews? = has_pageviews?(site)
 
     socket =
       assign(socket,
         site: site,
         super_admin?: super_admin?,
         domain: domain,
-        has_pageviews?: has_pageviews?(site),
+        has_pageviews?: has_pageviews?,
         component: @component,
         installation_type: params["installation_type"],
         report_to: self(),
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/verification.ex
```diff
@@ -45,28 +47,65 @@ defmodule PlausibleWeb.Live.Verification do
...
+      if connected?(socket) do
+        launch_delayed(socket)
+      end
+    end
+
+    on_ee do
+      {:ok, socket}
+    else
+      # on CE we skip the verification process and instead,
+      # we just wait for the first pageview to be recorded
```
ruslandoga, line 60: This is not perfect (e.g. review flow redirects to stats instead of showing success) but should do for the next release.

lib/plausible_web/live/verification.ex
```diff
@@ -45,28 +47,65 @@ defmodule PlausibleWeb.Live.Verification do
...
-      super_admin?={@super_admin?}
-    />
-    """
+  on_ce do
+    defp awaiting_pageviews(assigns) do
+      ~H"""
+      <PlausibleWeb.Components.Generic.focus_box>
+        <div class="flex items-center">
+          <div class="block pulsating-circle"></div>
+          <p class="ml-8">Awaiting a page view ...</p>
```
aerosol, line 104:
```suggestion
          <p class="ml-8">Awaiting your first pageview...</p>
```

Minor, to make it similar to what we communicate elsewhere

lib/plausible_web/live/verification.ex
```diff
@@ -45,28 +47,65 @@ defmodule PlausibleWeb.Live.Verification do
...
+        else
+          schedule_pageviews_check(socket)
+        end
+
+      {:ok, socket}
     end
+  end
 
-    {:ok, socket}
+  on_ee do
```
aerosol, line 72: Any reason to prefer separate render clauses to `:if={ee?()}` in the relevant tag?

lib/plausible_web/live/verification.ex
```diff
@@ -45,28 +47,65 @@ defmodule PlausibleWeb.Live.Verification do
...
 
-    if connected?(socket) do
-      launch_delayed(socket)
+    on_ee do
+      if connected?(socket) do
+        launch_delayed(socket)
+      end
+    end
+
+    on_ee do
```
aerosol, line 56: There's `on_ee` above, do we need this to be separate?

aerosol (approved):
LGTM, minor questions inline

lib/plausible_web/live/verification.ex
ruslandoga: I had it like this at first, but then changed to "a page view" to make it somewhat relevant in the flow=review.

lib/plausible_web/live/verification.ex
ruslandoga: More readable this way.

lib/plausible_web/live/verification.ex
ruslandoga: `render/1` clauses are tricky in lv. This way there is nothing to worry about change tracking. I didn't want to do extensive testing so went with a safer approach -- compile render clauses conditionally.

lib/plausible_web/live/verification.ex
ruslandoga: I guess the review flow is sort-of broken anyway, so we can call this check the "first pageview" there as well. Thank you!

ruslandoga: I think in the future CE might check not just the pageview count, but actually wait for a pageview to arrive after the check starts.

PR #4606, opened 2024-09-23
Mua: support 465 port connections

ruslandoga:
This PR adds support for SMTPS connections and also makes Mua use SMTPS when relay port is 465.

Related: https://github.com/plausible/analytics/discussions/4584

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -641,8 +641,18 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, ssl: [middlebox_comp_mode: middlebox_comp_mode]
 
     if relay = get_var_from_path_or_env(config_dir, "SMTP_HOST_ADDR") do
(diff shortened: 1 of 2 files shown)


config/runtime.exs
```diff
@@ -641,8 +641,21 @@ case mailer_adapter do
     config :plausible, Plausible.Mailer, ssl: [middlebox_comp_mode: middlebox_comp_mode]
 
     if relay = get_var_from_path_or_env(config_dir, "SMTP_HOST_ADDR") do
-      port = get_int_from_path_or_env(config_dir, "SMTP_HOST_PORT", 25)
-      config :plausible, Plausible.Mailer, relay: relay, port: port
+      port = get_int_from_path_or_env(config_dir, "SMTP_HOST_PORT", 587)
```
ruslandoga, line 644: Also changing default here to port=587 since that's what relays listen on usually.

zoldar (approved):

PR #4607, opened 2024-09-23
Check usage for users with active/cancelled/past due subscriptions

aerosol:
So that over limit e-mails and grace period is handled regardless - they're still subscribers.

### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/workers/check_usage.ex
+++ b/lib/workers/check_usage.ex
@@ -39,6 +39,7 @@ defmodule Plausible.Workers.CheckUsage do
     last_subscription_query =
       from(s in Subscription,
         order_by: [desc: s.inserted_at],
+        where: s.user_id == parent_as(:user).id,
         where:
           s.status in [
             ^Subscription.Status.active(),
@@ -57,10 +58,12 @@ defmodule Plausible.Workers.CheckUsage do
     active_subscribers =
       Repo.all(
         from(u in User,
-          join: s in subquery(last_subscription_query),
-          on: s.user_id == u.id,
+          as: :user,
+          inner_lateral_join: s in subquery(last_subscription_query),
+          on: true,
           left_join: ep in Plausible.Billing.EnterprisePlan,
           on: ep.user_id == u.id,
+          order_by: u.id,
           preload: [subscription: s, enterprise_plan: ep]
         )
       )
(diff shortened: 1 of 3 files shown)


lib/workers/check_usage.ex
```diff
@@ -43,7 +43,12 @@ defmodule Plausible.Workers.CheckUsage do
           on: s.user_id == u.id,
           left_join: ep in Plausible.Billing.EnterprisePlan,
           on: ep.user_id == u.id,
-          where: s.status == ^Subscription.Status.active(),
+          where:
+            s.status in [
+              ^Subscription.Status.active(),
+              ^Subscription.Status.past_due(),
+              ^Subscription.Status.deleted()
+            ],
```
zoldar, line 51: I wonder - shouldn't we pick the most recent subscription per user instead? The way it's done now, won't it evaluate each past subscription in cases where user was switching plans over time?

aerosol: omg so right, I'll see what the implications are tomorrow

aerosol: It won't evaluate _each_ past subscription, but may make decision (e.g. whether to send an email or not) based on the wrong one (e.g. cancelled). I've fixed the query via 9ddc7bdcc but I still need a test that will fail without it

aerosol: Pushed [bead241](https://github.com/plausible/analytics/pull/4607/commits/bead241a4ce9c70a09446c2f957e353f21f872f0) and [c630119](https://github.com/plausible/analytics/pull/4607/commits/c630119af2fa2a4abcf2fe5aef62763dd6b6242a) - WYT?

lib/workers/check_usage.ex
```diff
@@ -36,19 +36,31 @@ defmodule Plausible.Workers.CheckUsage do
...
+          least(day_of_month(s.last_bill_date), day_of_month(last_day_of_month(^yesterday))) ==
+            day_of_month(^yesterday),
+        limit: 1
+      )
+
     active_subscribers =
       Repo.all(
         from(u in User,
-          join: s in Plausible.Billing.Subscription,
+          join: s in subquery(last_subscription_query),
```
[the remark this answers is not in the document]
aerosol: https://github.com/plausible/analytics/pull/4607/commits/1602e3ad6918b881bce97bd565991ac22dea9648

zoldar (approved):
✨

PR #4624, opened 2024-09-25
Experimental pageleave events

RobertJoonas:
### Changes

With a future goal of implementing the `scroll_depth` metric, this PR introduces an experimental script variant for tracking `pageleave` events. 

Sending analytics data when the page is unloaded is unreliable by nature. Hence the first objective is to find out the percentage of pageleave events that we can expect (100% would mean that the number of pageleaves equals the number of pageviews).

If the percentage is reasonable, the next step would be to track a `scroll_depth` parameter with `pageleave` events, and eventually start reporting the average of that value on the dashboard, per page.

Here's a more technical description of the changes introduced in this PR:

* Implement a `script.pageleave.js` script variant.
  * The script will start triggering `pageleaves` on the `pagehide` browser event. We're only sticking with `pagehide` (and not `beforeunload`) because adding listeners to the `beforeunload` event will prevent placing the page in bfcache, therefore negatively affecting website performance ([reference doc](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)).
  * The script will also trigger `pageleaves` on history and hash-based SPA navigation. This will be tested at plausible.io/docs which uses history-based navigation. 
* Start using the new variant on plausible.io (three repositories):
  * plausible/analytics (this PR)
  * plausible/docs (will open a PR once this gets merged)
  * plausible/website (will open a PR once this gets merged)
* Consider `pageleave` a "special" event in many cases:
  * Prohibit adding goals with this event name
  * Do not count it towards the `events` metric
  * Do not count it towards "billable pageviews"
  * Do not let it affect the current `time_on_page`
  * When ingesting `pageleave` events, make sure it just gets the `session_id`, but **does not update the session** itself. At a later iteration we probably want to update `is_bounce` and `duration` as well.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible_web/dogfood.ex
+++ b/extra/lib/plausible_web/dogfood.ex
@@ -15,9 +15,9 @@ defmodule PlausibleWeb.Dogfood do
 
   def script_url() do
     if Application.get_env(:plausible, :environment) in ["prod", "staging"] do
-      "#{PlausibleWeb.Endpoint.url()}/js/script.manual.pageview-props.tagged-events.js"
+      "#{PlausibleWeb.Endpoint.url()}/js/script.manual.pageview-props.tagged-events.pageleave.js"
     else
-      "#{PlausibleWeb.Endpoint.url()}/js/script.local.manual.pageview-props.tagged-events.js"
+      "#{PlausibleWeb.Endpoint.url()}/js/script.local.manual.pageview-props.tagged-events.pageleave.js"
     end
   end
 
--- a/lib/plausible/goal/schema.ex
+++ b/lib/plausible/goal/schema.ex
@@ -72,25 +72,43 @@ defmodule Plausible.Goal do
   end
 
   defp validate_event_name_and_page_path(changeset) do
-    if validate_page_path(changeset) || validate_event_name(changeset) do
-      changeset
-      |> update_change(:event_name, &String.trim/1)
-      |> update_change(:page_path, &String.trim/1)
-    else
-      changeset
-      |> add_error(:event_name, "this field is required and cannot be blank")
-      |> add_error(:page_path, "this field is required and must start with a /")
+    case {validate_page_path(changeset), validate_event_name(changeset)} do
+      {:ok, _} ->
+        update_change(changeset, :page_path, &String.trim/1)
+
+      {_, :ok} ->
+        update_change(changeset, :event_name, &String.trim/1)
+
+      {{:error, page_path_error}, {:error, event_name_error}} ->
+        changeset
+        |> add_error(:event_name, event_name_error)
+        |> add_error(:page_path, page_path_error)
     end
   end
 
   defp validate_page_path(changeset) do
     value = get_field(changeset, :page_path)
-    value && String.match?(value, ~r/^\/.*/)
+
+    if value && String.match?(value, ~r/^\/.*/) do
+      :ok
+    else
+      {:error, "this field is required and must start with a /"}
+    end
   end
 
   defp validate_event_name(changeset) do
     value = get_field(changeset, :event_name)
-    value && String.match?(value, ~r/^.+/)
+
+    cond do
+      value == "pageleave" ->
+        {:error, "The event name 'pageleave' is reserved and cannot be used as a goal"}
+
+      value && String.match?(value, ~r/^.+/) ->
+        :ok
+
+      true ->
+        {:error, "this field is required and cannot be blank"}
+    end
   end
 
   defp maybe_drop_currency(changeset) do
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -34,6 +34,8 @@ defmodule Plausible.Ingestion.Event do
           | :site_page_blocklist
           | :site_hostname_allowlist
           | :verification_agent
+          | :lock_timeout
+          | :no_session_for_pageleave
 
   @type t() :: %__MODULE__{
           domain: String.t() | nil,
@@ -375,6 +377,9 @@ defmodule Plausible.Ingestion.Event do
           event
           | clickhouse_event: ClickhouseEventV2.merge_session(event.clickhouse_event, session)
         }
+
+      {:error, :no_session_for_pageleave} ->
+        drop(event, :no_session_for_pageleave)
 
       {:error, :timeout} ->
         drop(event, :lock_timeout)
--- a/test/plausible/billing/quota_test.exs
+++ b/test/plausible/billing/quota_test.exs
@@ -723,6 +723,27 @@ defmodule Plausible.Billing.QuotaTest do
              } = Quota.Usage.monthly_pageview_usage(user)
     end
 
+    test "pageleave events are not counted towards monthly pageview usage" do
+      user = insert(:user) |> Plausible.Users.with_subscription()
+      site = insert(:site, members: [user])
+      now = NaiveDateTime.utc_now()
+
+      populate_stats(site, [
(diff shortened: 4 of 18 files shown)


aerosol (approved):

lib/plausible/ingestion/event.ex
```diff
@@ -34,6 +34,8 @@ defmodule Plausible.Ingestion.Event do
           | :site_page_blocklist
           | :site_hostname_allowlist
           | :verification_agent
+          | :lock_timeout
```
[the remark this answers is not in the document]
zoldar: ❤️

aerosol:
Also, @RobertJoonas - would you consider mainlining a feature flag for it?

lib/plausible/goal/schema.ex
```diff
@@ -72,25 +72,43 @@ defmodule Plausible.Goal do
...
+      {:error, "this field is required and must start with a /"}
+    end
   end
 
   defp validate_event_name(changeset) do
     value = get_field(changeset, :event_name)
-    value && String.match?(value, ~r/^.+/)
+
+    cond do
+      value == "pageleave" ->
```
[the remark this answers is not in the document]
RobertJoonas: > FWIW, I checked on prod and there's no events/goals named like this.

Thanks! I did check goals myself too. Checking events was on my list as well - good to know! 🙏🏼

> Is the only reason for this safeguard that those events are not billable?

Pretty much. Other reasons might emerge later on, but this one was enough to prohibit it already. Otherwise people could use it as loophole and send us "free" events.

> This made me try entering a 'pageview' goal. Shall we also prevent it from being added?

I think that makes sense, but doesn't hurt to allow it. I haven't checked but I think suspect we might find some goals with this event name.

test/plausible/billing/quota_test.exs
```diff
@@ -723,6 +723,27 @@ defmodule Plausible.Billing.QuotaTest do
...
 
+    test "pageleave events are not counted towards monthly pageview usage" do
+      user = insert(:user) |> Plausible.Users.with_subscription()
+      site = insert(:site, members: [user])
+      now = NaiveDateTime.utc_now()
+
+      populate_stats(site, [
+        build(:event, timestamp: Timex.shift(now, days: -8), name: "custom"),
+        build(:pageview, user_id: 199, timestamp: Timex.shift(now, days: -5, minutes: -2)),
+        build(:event, user_id: 199, timestamp: Timex.shift(now, days: -5), name: "pageleave")
```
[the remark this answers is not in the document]
RobertJoonas: Sounds reasonable, I'll keep that comment in mind for later, when the pageleaves prove to work for us.

tracker/src/plausible.js
```diff
@@ -27,6 +28,42 @@
...
+    if (navigator.sendBeacon) {
+      var blob = new Blob([JSON.stringify(payload)], { type: 'text/plain' });
+      navigator.sendBeacon(endpoint, blob)
+    }
+  }
+
+  function registerPageLeaveListener(url) {
+    if (listeningPageLeave) { return }
+
+    window.addEventListener('pagehide', function () {
```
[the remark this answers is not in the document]
RobertJoonas: No. Only when the browser/tab is closed or the user navigates away from the page.

RobertJoonas:
> Also, @RobertJoonas - would you consider mainlining a feature flag for it? (at the query end)

I don't think it's necessary. There are no behavioural changes introduced in this PR. I'm also fairly confident that the extra `e.name != "pageleave"` and `count(*)` -> `countIf(e.name != "pageview")` clauses will not have any major impact on the query performance. Similar checks are used for pageviews already. In any case, I'll be waiting for @macobo's review on this as well before merging.

extra/lib/plausible_web/dogfood.ex
```diff
@@ -15,9 +15,9 @@ defmodule PlausibleWeb.Dogfood do
 
   def script_url() do
     if Application.get_env(:plausible, :environment) in ["prod", "staging"] do
-      "#{PlausibleWeb.Endpoint.url()}/js/script.manual.pageview-props.tagged-events.js"
+      "#{PlausibleWeb.Endpoint.url()}/js/script.manual.pageview-props.tagged-events.pageleave.js"
```
macobo, line 18: Note: If we need to revert it would make sense to revert here only - otherwise ingestion changes could cause trouble.

tracker/compile.js
```diff
@@ -26,7 +26,7 @@ function compilefile(input, output, templateVars = {}) {
   }
 }
 
-const base_variants = ["hash", "outbound-links", "exclusions", "compat", "local", "manual", "file-downloads", "pageview-props", "tagged-events", "revenue"]
+const base_variants = ["hash", "outbound-links", "exclusions", "compat", "local", "manual", "file-downloads", "pageview-props", "tagged-events", "revenue", "pageleave"]
```
macobo, line 29: Note/non-blocking: This name is likely to change once we enter the beta period w/ time-on-page.

tracker/src/plausible.js
```diff
@@ -10,6 +10,7 @@
   var scriptEl = document.currentScript;
   {{/if}}
   var endpoint = scriptEl.getAttribute('data-api') || defaultEndpoint(scriptEl)
+  var dataDomain = scriptEl.getAttribute('data-domain')
```
macobo, line 13: Potential behavioral change: if the script tag gets data-domain dynamically, this might break behavior.

Just a note in case, this isn't likely to break anything. Let's keep it in mind if we get any bug reports.

tracker/src/plausible.js
```diff
@@ -129,25 +171,41 @@
   {{#unless manual}}
     var lastPage;
 
-    function page() {
+    {{#if pageleave}}
+    var lastUrl = location.href
+
+    function pageLeave() {
```
macobo, line 177: Suggestion: Let's name this pageLeaveSPA - avoids a bit of confusion when reading the code.

tracker/src/plausible.js
```diff
@@ -129,25 +171,41 @@
...
-    window.addEventListener('hashchange', page)
+    window.addEventListener('hashchange', onSPANavigation)
     {{else}}
     var his = window.history
     if (his.pushState) {
       var originalPushState = his['pushState']
       his.pushState = function() {
         originalPushState.apply(this, arguments)
-        page();
+        onSPANavigation();
```
macobo, line 206: Q: Does history.pushState trigger pagehide? 

If so, the intended behavior here isn't clear from the code, perhaps we should add a comment.

macobo (commented):
Great work! Added a few comments around the script. Once these are resolved, let's :shipit:

tracker/src/plausible.js
RobertJoonas: > Q: Does history.pushState trigger pagehide?

AFAIK - not necessarily. But I'm sure there are cases where these events happen in very close succession to each other (even more so with Window popstate event). Those cases might result in two pageleave events being sent for a single pageview. I didn't consider that problematic, but I can see how it would complicate things...

I've now added a simple flag + timeout logic to prevent sending multiple pageleaves: https://github.com/plausible/analytics/pull/4624/commits/ce4ab9bab2390185e23c00b72fa1edbafb0ee314

tracker/src/plausible.js
RobertJoonas: Good call! https://github.com/plausible/analytics/pull/4624/commits/a2e7e7acd12711b4fa37b56b0a5c0c340dd2ac60

macobo (approved):

PR #4628, opened 2024-09-26
Try to make ingest lock timeout tests less flaky

zoldar:
### Changes

Can't tell for sure as this test fail is hard to reproduce, but I suspect there might be situations where reaching session processing step in the first ingest takes longer than the sleep before spinning up the second ingest.

### DIFF
--- a/test/plausible/ingestion/event_test.exs
+++ b/test/plausible/ingestion/event_test.exs
@@ -312,7 +312,7 @@ defmodule Plausible.Ingestion.EventTest do
                )
     end)
 
-    Process.sleep(100)
+    Process.sleep(200)
 
     assert {:ok, %{buffered: [], dropped: [dropped]}} = Event.build_and_buffer(second_request)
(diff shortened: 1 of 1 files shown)


aerosol (approved):
code_server overhead in interactive mode again? regardless, perhaps let's settle on `eventually` ?

zoldar:
@aerosol hmmm, I could track whether there's con_cache lock present with "eventually" in tight loop and only attempt the second ingest after that's satisfied. I'll see how feasible that is.

aerosol:
doesn't have to be con_cache specifically, since we're preloading the module in test_helper now

aerosol:
What I did to track it down previously:

1. Run iex in MIX_ENV=test
2. Set up flameon https://github.com/DockYard/flame_on/tree/main - looking at how @ruslandoga did it at https://github.com/ruslandoga/plausible/pull/301
3. Run tests of interests from iex via Mix.Task.run

PR #4648, opened 2024-10-02
Saved segments spike

apata:
This PR outlines schema and API structure for saved segments. Not intended to be merged as is.

### DIFF
--- a/lib/plausible/segment/schema.ex
+++ b/lib/plausible/segment/schema.ex
@@ -1,0 +1,30 @@
+defmodule Plausible.Segment do
+  @moduledoc """
+  Schema for segments. Segments are saved filter combinations.
+  """
+  use Plausible
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @type t() :: %__MODULE__{}
+
+  @derive {Jason.Encoder,
+           only: [:id, :name, :description, :segment_data, :inserted_at, :updated_at]}
+
+  schema "segments" do
+    field :name, :string
+    field :description, :string
+    field :segment_data, :map
+    belongs_to :site, Plausible.Site
+    many_to_many :users, Plausible.Auth.User, join_through: Plausible.SegmentCollaborator
+
+    timestamps()
+  end
+
+  def changeset(segment, attrs) do
+    segment
+    |> cast(attrs, [:name, :description, :segment_data, :site_id])
+    |> validate_required([:name, :segment_data, :site_id])
+    |> foreign_key_constraint(:site_id)
+  end
+end
--- a/lib/plausible/segment/schema.ex
+++ b/lib/plausible/segment/schema.ex
@@ -26,5 +26,16 @@ defmodule Plausible.Segment do
     |> cast(attrs, [:name, :description, :segment_data, :site_id])
     |> validate_required([:name, :segment_data, :site_id])
     |> foreign_key_constraint(:site_id)
+    |> validate_segment_data()
+  end
+
+  defp validate_segment_data(changeset) do
+    case get_field(changeset, :segment_data) do
+      %{"filters" => filters} when is_list(filters) ->
+        changeset
+
+      _ ->
+        add_error(changeset, :segment_data, "must contain property \"filters\" with array value")
+    end
   end
 end
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -5,6 +5,7 @@ defmodule Plausible.Stats.Breakdown do
   Avoid adding new logic here - update QueryBuilder etc instead.
   """
 
+  use Plausible.Repo
   use Plausible.ClickhouseRepo
   use Plausible
   use Plausible.Stats.SQL.Fragments
@@ -18,16 +19,28 @@ defmodule Plausible.Stats.Breakdown do
         %Query{dimensions: [dimension], order_by: order_by} = query,
         metrics,
         pagination,
-        _opts \\ []
+        opts \\ []
       ) do
+    get_available_segments = fn ->
+      case Keyword.get(opts, :conn) do
+        %{assigns: %{current_user: current_user}} ->
+          Repo.all(
+            from(i in Plausible.SegmentCollaborator,
+              join: segment in assoc(i, :segment),
+              select: {segment.id, segment},
+              where: i.user_id == ^current_user.id,
+              where: segment.site_id == ^site.id
+            )
+          )
+          |> Map.new()
+
+        _ ->
+          %{}
+      end
+    end
+
     transformed_metrics = transform_metrics(metrics, dimension)
     transformed_order_by = transform_order_by(order_by || [], dimension)
-
-    fake_segments =
-      Enum.into(
-        Enum.map(1..5, fn i -> {i, %{segment_data: %{filters: [["is", "visit:entry_page", ["/docs/#{i}"]]]}}} end),
-        %{}
-      )
 
     query_with_metrics =
       Query.set(
@@ -43,7 +56,7 @@ defmodule Plausible.Stats.Breakdown do
         # Allow pageview and event metrics to be queried off of sessions table
         legacy_breakdown: true
       )
-      |> QueryOptimizer.optimize(available_segments: fake_segments)
+      |> QueryOptimizer.optimize(get_available_segments: get_available_segments)
 
     q = SQL.QueryBuilder.build(query_with_metrics, site)
 
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -23,6 +23,12 @@ defmodule Plausible.Stats.Breakdown do
     transformed_metrics = transform_metrics(metrics, dimension)
     transformed_order_by = transform_order_by(order_by || [], dimension)
 
+    fake_segments =
+      Enum.into(
+        Enum.map(1..5, fn i -> {i, %{segment_data: %{filters: [["is", "visit:entry_page", ["/docs/#{i}"]]]}}} end),
+        %{}
+      )
+
     query_with_metrics =
       Query.set(
         query,
@@ -37,7 +43,7 @@ defmodule Plausible.Stats.Breakdown do
         # Allow pageview and event metrics to be queried off of sessions table
         legacy_breakdown: true
       )
-      |> QueryOptimizer.optimize()
+      |> QueryOptimizer.optimize(available_segments: fake_segments)
 
     q = SQL.QueryBuilder.build(query_with_metrics, site)
 
--- a/lib/plausible/stats/filters/filters_parser.ex
+++ b/lib/plausible/stats/filters/filters_parser.ex
@@ -81,6 +81,9 @@ defmodule Plausible.Stats.Filters.FiltersParser do
            "Invalid visit:country filter, visit:country needs to be a valid 2-letter country code."}
         end
 
+      {"segment", false} when all_integers? ->
+        {:ok, [list]}
+
       {_, true} ->
         {:ok, [list]}
 
@@ -113,6 +116,9 @@ defmodule Plausible.Stats.Filters.FiltersParser do
         else
           {:error, error_message}
         end
+
+      "segment" ->
+        {:ok, filter_key}
 
       _ ->
         {:error, error_message}
--- a/lib/plausible/stats/filters/filters_parser.ex
+++ b/lib/plausible/stats/filters/filters_parser.ex
@@ -6,7 +6,21 @@ defmodule Plausible.Stats.Filters.FiltersParser do
   alias Plausible.Stats.Filters
   alias Plausible.Helpers.ListTraverse
 
-  defp i(value), do: inspect(value, charlists: :as_lists)
+  @segment_filter_key "segment"
+  def segment_filter_key(), do: @segment_filter_key
+
+  @filter_entry_operators [
+    :is,
+    :is_not,
+    :matches,
+    :matches_not,
+    :matches_wildcard,
+    :matches_wildcard_not,
+    :contains,
+    :contains_not
+  ]
+
+  @filter_tree_operators [:not, :and, :or]
 
   def parse_filters(filters) when is_list(filters) do
     ListTraverse.parse_list(filters, &parse_filter/1)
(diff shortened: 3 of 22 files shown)


test/plausible/stats/query_parser_test.exs
```diff
@@ -491,6 +491,7 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
         "metrics" => ["visitors"],
         "date_range" => "all",
         "filters" => [
+          ["is", "segment", [200]],
```
macobo, line 494: Since it's not clear whether this is intended as the full code: Nit: separate test.

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
```diff
@@ -0,0 +1,180 @@
...
+               "error" => "Segment not found with ID \"#{segment_id}\""
+             }
+    end
+
+    test "serves 200 with segment when user is collaborating on the segment", %{
+      conn: conn,
+      site: site,
+      user: user
+    } do
+      name = "foo"
```
macobo, line 66: Nit: I don't think DRY-ing the variables here helps here - it just ends up with a wordier test where you need to scroll back-and-forth.

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
```diff
@@ -0,0 +1,180 @@
...
+        })
+
+      response = json_response(conn, 200)
+
+      assert %{
+               "role" => "owner",
+               "segment" => %{
+                 "description" => nil,
+                 "name" => ^name,
+                 "segment_data" => ^segment_data
```
macobo, line 122: Nit: This is confusing to follow - it isn't clear what the response actually looks like. I suggest the following structure.

```
segment = from(s in segments, where: %{ site_id: ^site_id }) |> Repo.one()

assert json_response(conn, 200) == %{
               "role" => "owner",
               "segment" => %{ ... }
}
```

And using fully-hard-coded values

lib/plausible/stats/breakdown.ex
```diff
@@ -18,8 +19,26 @@ defmodule Plausible.Stats.Breakdown do
         %Query{dimensions: [dimension], order_by: order_by} = query,
         metrics,
         pagination,
-        _opts \\ []
+        opts \\ []
       ) do
+    get_available_segments = fn ->
+      case Keyword.get(opts, :conn) do
```
macobo, line 25: Q: Why is this in legacy breakdown code and no-where else? 

Feedback: Having a mysterious conn option passed to this module where assigns need to be in a specific structure couples things way too close to the controller and mixes responsibilities.

Suggestion: Store `user_id` on Query object, use that in here.

Problem: I don't think filtering by user_id is at all correct here and we should only filter on site_id. This is the sharing links problem I've mentioned before.

lib/plausible/stats/filters/filters_parser.ex
```diff
@@ -0,0 +1,134 @@
...
+    :matches_not,
+    :matches_wildcard,
+    :matches_wildcard_not,
+    :contains,
+    :contains_not
+  ]
+
+  @filter_tree_operators [:not, :and, :or]
+
+  def parse_filters(filters) when is_list(filters) do
```
macobo, line 25: This doesn't apply the JSON schema which is a problem.

lib/plausible/stats/filters/filters_parser.ex
```diff
@@ -0,0 +1,134 @@
+defmodule Plausible.Stats.Filters.FiltersParser do
```
macobo, line 1: I'm not sold on separating out this module from QueryParser due to it making it more likely coverage and code quality starts to slip and utils are now used from all over the place..

We could instead expose a single `validate_filters(...)` method on QueryParser.

lib/plausible/stats/filters/segments.ex
```diff
@@ -0,0 +1,72 @@
...
+  alias Plausible.Stats.Filters
+  alias Plausible.Stats.Filters.FiltersParser
+
+  @spec has_segment_filters?(list()) :: boolean()
+  def has_segment_filters?(filters),
+    do: Filters.filtering_on_dimension?(filters, FiltersParser.segment_filter_key())
+
+  @spec expand_segments_to_constituent_filters(list(), map()) ::
+          {:ok, list()} | {:error, any()}
+  def expand_segments_to_constituent_filters(filters, segments) do
```
macobo, line 11: Why a separate module for this vs putting it in the QueryOptimizer? The responsibilities of this vs QueryOptimizer are unclear.

lib/plausible/stats/filters/segments.ex
```diff
@@ -0,0 +1,72 @@
...
+    case segment_filter_index = find_top_level_segment_filter_index(filters) do
```
macobo, line 12: This seems quite unwieldy and bakes in the unverified assumption that there's only ever a single segment filter.

Why not:

```
Enum.flat_map(filters, fn 
  [:is, "segment", clauses] =  -> expand_segment_filter(clauses)
  filter -> [filter]
end)
```

lib/plausible/stats/filters/segments.ex
```diff
@@ -0,0 +1,72 @@
...
+
+        expanded_filters =
+          Enum.concat(
+            Enum.map(segment_id_clauses, fn segment_id ->
+              with {:ok, segment_data} <- get_segment_data(segments, segment_id),
+                   {:ok, %{filters: filters}} <-
+                     validate_segment_data(segment_data) do
+                filters
+              else
+                {:error, :segment_not_found} ->
```
macobo, line 28: This code should not be raising. Instead if segment does not exist, parsing in QueryParser should fail.

This way we can provide users proper feedback AND have sane boundaries between code that can throw and code that is just manipulating data structures.

See also `preload_goals_if_needed` and goal-related solutions in there. I suggest that we similarly store preloaded segments on the query object.

priv/repo/migrations/20240918171300_create_segments_and_segment_users.exs
```diff
@@ -0,0 +1,29 @@
+defmodule Plausible.Repo.Migrations.CreateSegmentsAndSegmentUsers do
+  use Ecto.Migration
+
+  def change do
+    create table(:segments) do
+      add :name, :string, null: false
+      add :segment_data, :map, null: false
+      add :site_id, references(:sites, on_delete: :delete_all), null: false
+
+      add :description, :text
```
macobo, line 10: Is this nullable? If so, let's be explicit.

priv/repo/migrations/20240918171300_create_segments_and_segment_users.exs
```diff
@@ -0,0 +1,29 @@
+defmodule Plausible.Repo.Migrations.CreateSegmentsAndSegmentUsers do
+  use Ecto.Migration
+
+  def change do
+    create table(:segments) do
+      add :name, :string, null: false
+      add :segment_data, :map, null: false
```
macobo, line 7: Why not `add :segment_filters, :array, null: false` or something equivelent? What are we winning by nesting the data under filters and creating an abstraction?

priv/json-schemas/query-api-schema.json
```diff
@@ -335,10 +340,29 @@
...
       ]
     },
+    "filter_for_segment": {
+      "type": "array",
+      "additionalItems": false,
+      "minItems": 3,
+      "maxItems": 3,
+      "items": [
+        {
+          "$ref": "#/definitions/filter_operation_for_segments"
```
macobo, line 350: Nit: Let's inline the operation given there's only one and we're not repeating the ref anywhere.

```suggestion
          "const": "is"
```

priv/repo/migrations/20240918171300_create_segments_and_segment_users.exs
```diff
@@ -0,0 +1,29 @@
...
+      add :site_id, references(:sites, on_delete: :delete_all), null: false
+
+      add :description, :text
+
+      timestamps()
+    end
+
+    create index(:segments, [:segment_data], using: :gin)
```
macobo, line 15: What will be using this index?

priv/repo/migrations/20240918171300_create_segments_and_segment_users.exs
```diff
@@ -0,0 +1,29 @@
...
+    create index(:segments, [:site_id])
+
+    create table(:segment_collaborators, primary_key: false) do
+      add :role, :string, null: false
```
macobo, line 19: String columns like this are suspect but it seems this is ORM-generated right?

lib/plausible/segment/schema.ex
```diff
@@ -0,0 +1,41 @@
...
+    segment
+    |> cast(attrs, [:name, :description, :segment_data, :site_id])
+    |> validate_required([:name, :segment_data, :site_id])
+    |> foreign_key_constraint(:site_id)
+    |> validate_segment_data()
+  end
+
+  defp validate_segment_data(changeset) do
+    case get_field(changeset, :segment_data) do
+      %{"filters" => filters} when is_list(filters) ->
```
macobo, line 34: Should validate here that it's valid and doesn't contain any nested segment references right?

macobo (commented):
Some initial thoughts. :rocket:

lib/plausible/stats/breakdown.ex
apata: Well spotted! I was just about to ask for help on this in team chat. Rather than storing the user_id, what if I store the loader function in the Query object?

lib/plausible/segment/schema.ex
apata: Agreed! It should be validated, but at what level? I didn't want to make this module depend on query / filters parsing. How about doing that in the API controller?

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
apata: Thanks! I struggled with the timestamps in this test file. Any tips on how to omit them from the comparison while remaining brief?

lib/plausible/stats/filters/filters_parser.ex
apata: Agreed! I'm not happy with this split either. As I was doing it, I noticed that the module definitions didn't fully line up with the responsibilities in the modules: Stats.Filters.QueryParser, Stats.Filters.

lib/plausible/segment/schema.ex
macobo: Doing it in the controller seems valid and best (and if it's already validated then sorry for missing it!).

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
macobo: Good question - elixir assertions don't have any clever shorthands like any(datetime) for assert. I'll dig into existing tests for controllers and see how they have solved it. :thinking: Will provide an answer tomorrow!

lib/plausible/segment/schema.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Plausible.Segment do
```
macobo, line 1: Suggestion: move this under `lib/plausible/segment.ex` - the current placement makes the module hard to find.

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
macobo: Dug a bit.

We don't have a usecase yet for exposing updated_at or inserted_at at the API level. Looking at similar models like lib/plausible/auth/invitation.ex and lib/plausible/site.ex we don't expose them either. I'd remove exposing them.

I'd rewrite the test as follows:

```elixir

  describe "POST /internal-api/:domain/segments" do
    setup [:create_user, :create_new_site, :log_in]

    test "creates segment successfully", %{conn: conn, site: site} do
      conn =
        post(conn, "/internal-api/#{site.domain}/segments", %{
          "segment_data" => %{"filters" => [["is", "visit:entry_page", ["/blog"]]]},
          "name" => "Blog entry"
        })

      segment = Plausible.Repo.one(Plausible.Segment)

      assert json_response(conn, 200) == %{
               "role" => "owner",
               "segment" => %{
                 "id" => segment.id,
                 "name" => "Blog entry",
                 "segment_data" => %{"filters" => [["is", "visit:entry_page", ["/blog"]]]},
                 "description" => nil,
               }
             }
    end
  end
```

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
```diff
@@ -0,0 +1,180 @@
...
+    end
+
+    test "serves 200 with segment when user is collaborating on the segment", %{
+      conn: conn,
+      site: site,
+      user: user
+    } do
+      name = "foo"
+      description = "bar"
+      segment_data = %{"filters" => []}
```
macobo, line 68: Let's use realistic test data to simulate how a user might use the API. Since the data doesn't matter for most tests, I suggest defining `@segment_data %{"filters" => [["is", "visit:entry_page", ["/blog"]]]}` at the top level and re-using it in most tests.

test/plausible_web/controllers/api/internal_controller/segments_controller_test.exs
```diff
@@ -0,0 +1,180 @@
...
+      inserted_at = "2024-09-01T10:00:00"
+      updated_at = inserted_at
+      role = "owner"
+
+      %{id: segment_id} =
+        insert(:segment,
```
macobo, line 74: Rather than insert it here, would it make sense to use the other APIs to create these to simulate user behavior?

lib/plausible/stats/breakdown.ex
apata: This turned out to be not necessary.

priv/repo/migrations/20240918171300_create_segments_and_segment_users.exs
apata: This was a premature optimization to lookup segments containing a filter like [is, goal, [...,<previous goal name>,...]] when a goal is being updated. Maybe not needed though.

PR #4653, opened 2024-10-07
Fix revenue metrics if custom display name is used

macobo:
Seems since https://github.com/plausible/analytics/pull/4415 or https://github.com/plausible/analytics/pull/4391, revenue metrics have not been correctly displayed on the dashboard if the goal display name has been changed from the default.

Discovered while refactoring this code, wanted to PR the fix separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 ### Fixed
+- Revenue metrics are displayed correctly after goal has been renamed
 
 ## v2.1.3 - 2024-09-26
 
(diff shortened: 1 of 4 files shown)


aerosol:
Ref https://github.com/plausible/analytics/pull/4391

aerosol (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 ### Fixed
+- Revenue metrics are displayed correctly after goal has been renamed
```
ruslandoga, line 16: I'm not 100% sure, but I think changelog is mostly for CE, and revenue features are not enabled there. Should this still be included in v2.1.4 release notes?

aerosol: @ruslandoga thanks for spotting - yes, changelog is for CE. There is no reason to mention revenue goals indeed.

ruslandoga: Ok if I remove it in https://github.com/plausible/analytics/pull/4660?

aerosol: of course

PR #4656, opened 2024-10-08
Trigger pageviews on pages loaded from bfcache

RobertJoonas:
### Changes

While experimenting with pageleave events we've discovered that pageviews are not being fired when a page is loaded from bfcache.

This PR fixes that, for now, only in the context of the pageleave script variant. I'm leaving the default script untouched for now as the pageleave variant will also be merged into the default script at some point.

The fix itself is to listen for the `pageshow` event and fire a pageview if the `event.persisted` property is `true`.

Note that we only add the 'pageshow' listener if the script is **not in manual mode**. If it's manual, the listener needs to be added manually as well.

Since we're using the manual variant ourselves (in both this repo, and in plausible/website):

1. the PR at hand updates `_tracking.html` to add the 'pageshow' listener
2. [this PR](https://github.com/plausible/website/pull/662) updates the tracking in the plausible/website repo

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/layout/_tracking.html.heex
+++ b/lib/plausible_web/templates/layout/_tracking.html.heex
@@ -32,6 +32,13 @@
       }
 
       plausible('pageview', defaultEventOptions)
+
+      window.addEventListener('pageshow', function(event) {
+        if (event.persisted) {
+          // Page was restored from bfcache - trigger a pageview
+          plausible('pageview', defaultEventOptions);
+        }
+      })
     </script>
   <% end %>
 <% end %>
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -231,6 +231,15 @@
     } else {
       page()
     }
+
+    {{#if pageleave}}
+    window.addEventListener('pageshow', function(event) {
+      if (event.persisted) {
+        // Page was restored from bfcache - trigger a pageview
+        page();
+      }
+    })
+    {{/if}}
   {{/unless}}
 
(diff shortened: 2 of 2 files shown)


lib/plausible_web/templates/layout/_tracking.html.heex
```diff
@@ -32,6 +32,13 @@
       }
 
       plausible('pageview', defaultEventOptions)
+
+      window.addEventListener('pageshow', function(event) {
```
macobo, line 36: Q: why are we adding it here? Is this leftover debugging code?

tracker/src/plausible.js
```diff
@@ -231,6 +231,15 @@
     } else {
       page()
     }
+
+    {{#if pageleave}}
+    window.addEventListener('pageshow', function(event) {
+      if (event.persisted) {
```
macobo, line 237: Non-blocking Q: Would _only_ sending if `!event.persisted` for `pagehide` also do the trick?

There's some business impact on this change and that would provide us a lever that wouldn't cost users money.

macobo (approved):

tracker/src/plausible.js
RobertJoonas: Yeah, I believe it's possible to accomplish something like this. But sending or not sending pageleaves would have to depend on whether 'pageshow' happened with `event.persisted` or not. So basically set a flag here on line 238 like `disableSendingPageleave = true`.

PR #3336, opened 2023-09-09
log mailer errors

ruslandoga:
### Changes

Log mailer errors both to console and Sentry. The way these errors are sent to Sentry should stay the same ~~,but I'll need to try it out on my instance first.~~ (done, and they are still fingerprinted under the same issue)

#### Before
Screenshot 2023-09-09 at 10 55 57

#### After
Screenshot 2023-09-09 at 10 56 11

REF: https://github.com/plausible/analytics/discussions/3335#discussioncomment-6953510

There are more places in the codebase where the logs are suppressed and only Sentry receives the error; if this PR works out well, I'll update those other places as well.

### Tests
- [x] This PR does not require tests

### Changelog
- [x] Entry has been added to changelog

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #4657, opened 2024-10-08
Log Oban errors (and still report them to Sentry)

ruslandoga:
This PR adds logging for Oban errors. The reasoning is that most self-hosters don't have Sentry configured which makes these errors invisible.

Here's how it'd look after this PR:

Screenshot 2024-10-08 at 16 23 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 
 - Add ability to review and revoke particular logged in user sessions
 - Add ability to change password from user settings screen
+- Add error logs for background jobs plausible/analytics#4657
 
 ### Removed
 
--- a/lib/oban_error_reporter.ex
+++ b/lib/oban_error_reporter.ex
@@ -22,32 +22,17 @@ defmodule ObanErrorReporter do
       |> Map.merge(measure)
 
     on_job_exception(job)
-
-    on_ce do
-      log_ce_error(meta, extra)
-    end
-
-    Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
+    capture_error(meta, extra)
   end
 
   defp handle_event([:oban, :notifier, :exception], _timing, meta) do
     extra = Map.take(meta, ~w(channel payload)a)
-
-    on_ce do
-      log_ce_error(meta, extra)
-    end
-
-    Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
+    capture_error(meta, extra)
   end
 
   defp handle_event([:oban, :plugin, :exception], _timing, meta) do
     extra = Map.take(meta, ~w(plugin)a)
-
-    on_ce do
-      log_ce_error(meta, extra)
-    end
-
-    Sentry.capture_exception(meta.reason, stacktrace: meta.stacktrace, extra: extra)
+    capture_error(meta, extra)
   end
 
   defp on_job_exception(%Oban.Job{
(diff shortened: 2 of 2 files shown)


macobo (approved):

macobo:
It's worth adding something to changelog as well!

aerosol:
IMHO it's ok not to do the CE branching. It doesn't hurt to log regardless of the build.

ruslandoga:
I thoughts so too, but decided to go with CE-only since otherwise EE would receive duplicate Sentry reports since error logs get sent there too: https://github.com/plausible/analytics/blob/e7332d95b251f64122e8afb67e7985028e7554de/config/runtime.exs#L33-L35

aerosol:
Just replace Sentry.capture_exception that's beneath with Logger.error then?

ruslandoga:
The reports would be a bit different (https://github.com/plausible/analytics/pull/3336), is that OK?

ruslandoga:
<details><summary>Sentry.capture_exception screenshot</summary>

from-sentry

</details>

<details><summary>Logger.error screenshot</summary>

from-logger

</details>

The differences I see are:
- Sentry.capture_message groups / fingerprints by stacktrace instead of error message (this info is at the very bottom of the page)
- using Sentry.capture_message renders stacktrace in a nice table :)

Also, to make it work, I needed to modify the Logger call to

```elixir
Logger.error(
  Exception.format(:error, meta.reason, meta.stacktrace),
  sentry: %{extra: extra}
)
```

which removes the "Background job failed" message and makes the actual log appear like this

```
16:52:52.129 [error] ** (ArgumentError) The email address `plausible@localhost` is invalid.

    (mail 0.3.1) lib/mail/renderers/rfc_2822.ex:134: Mail.Renderers.RFC2822.validate_address/1
    (mail 0.3.1) lib/mail/renderers/rfc_2822.ex:141: Mail.Renderers.RFC2822.render_address/1
    (mail 0.3.1) lib/mail/renderers/rfc_2822.ex:97: Mail.Renderers.RFC2822.render_header/2
    (elixir 1.17.1) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (elixir 1.17.1) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (mail 0.3.1) lib/mail/renderers/rfc_2822.ex:174: Mail.Renderers.RFC2822.render_headers/2
    (mail 0.3.1) lib/mail/renderers/rfc_2822.ex:56: Mail.Renderers.RFC2822.render_part/2
    (bamboo_mua 0.2.2) lib/bamboo_mua.ex:57: Bamboo.Mua.deliver/2
    (bamboo 2.3.0) lib/bamboo/mailer.ex:208: Bamboo.Mailer.deliver_now/4
    (bamboo 2.3.0) lib/bamboo/mailer.ex:231: Bamboo.Mailer.deliver_now!/4
    (plausible 0.0.1) lib/workers/notify_exported_analytics.ex:36: Plausible.Workers.NotifyExportedAnalytics.perform/1
    (oban 2.17.2) lib/oban/queue/executor.ex:129: Oban.Queue.Executor.perform/1
    (oban 2.17.2) lib/oban/queue/executor.ex:74: Oban.Queue.Executor.call/1
    (elixir 1.17.1) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
    (elixir 1.17.1) lib/task/supervised.ex:36: Task.Supervised.reply/4
```

which is probably less intuitive for self-hosters.

ruslandoga:
Seems like I can make both look the same by passing `crash_reason: {meta.reason, meta.stacktrace}` to Logger. Then I would also be able to to modify the message to be more user-friendly since it won't be used for fingerprinting.

aerosol (approved):

ruslandoga:
I've updated the PR and now it logs a "self-hoster-friendly" message but reports a more structured Stnry event, the same way as before.

PR #4659, opened 2024-10-08
APIv2: Revenue metrics

macobo:
### Changes

This PR brings revenue metrics to APIv2. They are still not yet publicly available but are now fully functional for dashboard use.

Notes:
- No behavioral changes for users
- Revenue currencies are stored on `Query` similar to goals with associated business logic
- Revenue metric access is validated properly
- Old behavior of removing revenue metrics when unavailable is done via a Query flag and QueryOptimizer behavior.

Needed to bring breakdowns frontend to APIv2.

### DIFF
--- a/extra/lib/plausible/stats/goal/revenue.ex
+++ b/extra/lib/plausible/stats/goal/revenue.ex
@@ -45,6 +45,9 @@ defmodule Plausible.Stats.Goal.Revenue do
     Plausible.Billing.Feature.RevenueGoals.check_availability(site.owner) == :ok
   end
 
+  # :NOTE: Legacy queries don't have metrics associated with them so work around the issue by assuming
+  #   revenue metric was requested.
+  defp requested?([]), do: true
   defp requested?(metrics), do: Enum.any?(metrics, &(&1 in @revenue_metrics))
 
   defp get_goal_dimension_revenue_currency(query, dimension_values) do
@@ -111,6 +114,9 @@ defmodule Plausible.Stats.Goal.Revenue do
   end
 
   def cast_revenue_metrics_to_money(results, _), do: results
+
+  # :TODO: Avoid double-casting while working on the feature.
+  def maybe_cast_metric_to_money(%Money{} = value, _, _), do: value
 
   def maybe_cast_metric_to_money(value, metric, currency) do
     if currency && metric in @revenue_metrics do
--- a/extra/lib/plausible/stats/goal/revenue.ex
+++ b/extra/lib/plausible/stats/goal/revenue.ex
@@ -13,7 +13,7 @@ defmodule Plausible.Stats.Goal.Revenue do
   end
 
   def preload_revenue_currencies(site, goals, metrics, goal_filters?) do
-    if requested?(metrics) and length(goals) > 0 and revenue_goals_available?(site) do
+    if requested?(metrics) and length(goals) > 0 and available?(site) do
       goal_currency_map =
         Map.new(goals, fn goal -> {Plausible.Goal.display_name(goal), goal.currency} end)
 
@@ -40,7 +40,7 @@ defmodule Plausible.Stats.Goal.Revenue do
     end
   end
 
-  defp revenue_goals_available?(site) do
+  defp available?(site) do
     site = Plausible.Repo.preload(site, :owner)
     Plausible.Billing.Feature.RevenueGoals.check_availability(site.owner) == :ok
   end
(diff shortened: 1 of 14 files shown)


lib/plausible/stats/filters/query_parser.ex
```diff
@@ -405,14 +410,22 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
-  def preload_goals_if_needed(site, filters, dimensions) do
+  def preload_goals_if_needed(site, metrics, filters, dimensions) do
```
RobertJoonas, line 413: A companion to my other suggestion - also rename this function to `preload_needed_goals`.

RobertJoonas (commented):
I think I found a small bug here (revenue metrics should show up in the Top Stats as €0.0 instead of blank):

https://pr-4659.review.plausible.io/plausible.io?filters=((is,goal,(Payment)))&period=day&keybindHint=E&date=2024-10-08

Otherwise the PR looks good to me. Left a couple of suggestions inline though.

lib/plausible/goals/filters.ex
```diff
@@ -34,18 +34,37 @@ defmodule Plausible.Goals.Filters do
     end)
   end
 
+  def preload(goals, filters) do
```
apata, line 37: **suggestion, non-blocking**: When used as Goals.Filters.preload, it's not obvious what records are getting preloaded. Bit more verbose, but how's `Goals.Filters.reduce_goals_by_filters`?

extra/lib/plausible/stats/goal/revenue.ex
```diff
@@ -2,83 +2,74 @@ defmodule Plausible.Stats.Goal.Revenue do
...
+  def available?(site) do
+    site = Plausible.Repo.preload(site, :owner)
+    Plausible.Billing.Feature.RevenueGoals.check_availability(site.owner) == :ok
   end
 
-  def cast_revenue_metrics_to_money(results, _), do: results
+  # :NOTE: Legacy queries don't have metrics associated with them so work around the issue by assuming
+  #   revenue metric was requested.
+  def requested?([]), do: true
+  def requested?(metrics), do: Enum.any?(metrics, &(&1 in @revenue_metrics))
```
apata, line 66: **suggestion, non-blocking**: I understand this will be temp, but could we detect the legacy case here by having the query.v2 as a second arg? `Revenue.requested?(query.metrics, query.v2)`

apata (approved):
Nice work!

lib/plausible/goals/filters.ex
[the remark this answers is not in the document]
macobo: Good suggestion!

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -460,6 +473,25 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
+  on_ee do
+    alias Plausible.Stats.Goal.Revenue
+
+    defdelegate preload_revenue_currencies(site, preloaded_goals, metrics, dimensions),
+      to: Plausible.Stats.Goal.Revenue
```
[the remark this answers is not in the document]
macobo: Yes, but I preferred the full name for explicitness. `defdelegate` is quite rare in our codebase so I wanted the code-reader to have a straight-forward experience.

lib/plausible/goals/filters.ex
macobo: Robert above had a similar thought. Updated!

extra/lib/plausible/stats/goal/revenue.ex
macobo: Doesn't seem straight-forward to introduce it without polluting preload call with extra args as well.

macobo:
@RobertJoonas good catch, this also caught an untested edge case. Fixed and updated the tests.

test/plausible_web/controllers/api/stats_controller/top_stats_test.exs
```diff
@@ -1488,6 +1489,71 @@ defmodule PlausibleWeb.Api.StatsController.TopStatsTest do
...
+        build(:event,
+          name: "Payment",
+          revenue_reporting_amount: Decimal.new(1_000),
+          revenue_reporting_currency: "USD"
+        ),
+        build(:event, name: "Signup"),
+        build(:event, name: "Signup")
+      ])
+
+      filters = Jason.encode!(%{goal: "Payment|Signup"})
```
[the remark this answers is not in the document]
macobo: Fair. I used this for the consistency with the rest of the file, but should likely refactor all of it at this point as it doesn't reflect actual dashboard calls. Adding it to my mental TODO.

extra/lib/plausible/stats/goal/revenue.ex
apata: Ah, roger that. The idea was to get rid of the note, but unless we passed the query itself as the only arg, it wouldn't be obvious.

RobertJoonas (approved):

PR #4667, opened 2024-10-10
Fix experimental pageleaves script (SPAs)

RobertJoonas:
### Changes

Fixes two bugs with the experimental script:

1. After the initial pageview is triggered, the current script sets an eventListener to trigger pageleave with the **URL of the initial page**, when `pagehide` occurs. That doesn't make sense in Single Page Apps. Instead, every time SPA navigation takes place, that URL should change to that of the last pageview.

2. Currently, pageleave events in SPAs are being fired regardless of exclusion rules (incl localhost and bot traffic). The second commit of this PR fixes that by adding an additional condition: only fire pageleaves on SPA navigation if `listeningPageLeave` is `true`. That value will never become true until a pageview has been sent.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -190,7 +190,7 @@
       {{/unless}}
       
       {{#if pageleave}}
-      if (isSPANavigation) {
+      if (isSPANavigation && listeningPageLeave) {
         triggerPageLeave();
         currentURL = location.href;
       }
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -2,7 +2,6 @@
   'use strict';
 
   var location = window.location
-  var currentURL = location.href
   var document = window.document
 
   {{#if compat}}
@@ -32,6 +31,11 @@
   {{#if pageleave}}
   // :NOTE: Tracking pageleave events is currently experimental.
 
+  // Keeps track of the URL to be sent in the pageleave event payload.
+  // Should get updated on pageviews triggered manually with a custom
+  // URL, and on SPA navigation.
+  var currentPageLeaveURL = location.href
+
   // Multiple pageviews might be sent by the same script when the page
   // uses client-side routing (e.g. hash or history-based). This flag
   // prevents registering multiple listeners in those cases.
@@ -52,7 +56,7 @@
     var payload = {
       n: 'pageleave',
       d: dataDomain,
-      u: currentURL,
+      u: currentPageLeaveURL,
     }
 
     {{#if hash}}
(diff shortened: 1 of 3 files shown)


tracker/src/plausible.js
```diff
@@ -116,10 +115,13 @@
...
+    payload.u = currentURL
     {{#if manual}}
-    payload.u = options && options.u ? options.u : location.href
-    {{else}}
-    payload.u = location.href
+    if (options && options.u) {
+      payload.u = options.u
+
+      if (isPageview) {currentURL = options.u}
+    }
```
RobertJoonas, line 124: This part is a little tricky, and there's actually a behavioural change here which I'm not sure if we should introduce or not.

Basically, when the url for a pageview is overridden with `script.manual.js`, e.g. `plausible('pageview', {u: 'https://plausible.io/:dashboard'})` (let's say that the real url is `https://plausible.io/plausible.io`), then with this change, any custom event that gets sent during that pageview, without an explicitly passed custom location, will also get that same anonymized URL. 

For example. Let's say `plausible('Filter Menu: Open')` is called during that pageview.

* Currently, before this change: the URL is sent as `https://plausible.io/plausible.io`
* After this change: the URL will be sent as `https://plausible.io/:dashboard`

On one hand, it would make sense that custom events sent during the same pageview get the same page path as the pageview. On the other hand, it might not be obvious that it works that way.

To keep the existing behaviour in place we could just add something like:

```javascript
if (options && options.u) {
  // ...
}

payload.u = location.href
```

tracker/src/plausible.js
```diff
@@ -96,7 +95,7 @@
     var dataIncludeAttr = scriptEl && scriptEl.getAttribute('data-include')
     var dataExcludeAttr = scriptEl && scriptEl.getAttribute('data-exclude')
 
-    if (eventName === 'pageview') {
+    if (isPageview) {
```
macobo, line 98: Reading this code, shouldn't we also exclude page leaves?

RobertJoonas: pageleaves are never sent via the `trigger` function. There are several differences in how these requests are sent, which is why they both have their own functions.

tracker/src/plausible.js
RobertJoonas: Had a call with Uku and decided to revert the behavioural change in the end. Not a good enough reason to change the behaviour - it's true that it sounds like an improvement, but from the API perspective, it looks like a hidden functionality which is not obvious.

About the tests - I've only added tests for the manual extension at the moment. Planning to work on pageleave script testing too, but there are so many different scenarios to test, so I'd like to get the script fix live and out of the way. Tests for pageleave will come in a separate PR later :)

macobo (approved):
Still not 100% confident in this (more testing around pageleave would be nice) but let's ship and monitor!

Issue #4249, opened 2024-06-19
Bamboo.Mua configuration fix for some mailing servers


Striffly:
### Past Issues Searched

- [X] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [X] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

Hello,

I recently migrated Plausible from version 2.0.0 to 2.1.0 (and now 2.1.1), and thus switched to Bamboo.Mua.

Since then, I have been encountering problems with receiving Plausible emails, which have become impossible.

To debug the issue, here are the commands I executed:
```text
email = Bamboo.Email.new_email(from: PlausibleWeb.Email.mailer_email_from(), to: "xxx@hotmail.fr", subject: "test html_body", html_body: "test html_body")
Plausible.Mailer.deliver_now(email)

email = Bamboo.Email.new_email(from: PlausibleWeb.Email.mailer_email_from(), to: "xxx@hotmail.fr", subject: "test text_body", text_body: "test text_body")
Plausible.Mailer.deliver_now(email)

email = Bamboo.Email.new_email(from: PlausibleWeb.Email.mailer_email_from(), to: "xxx@hotmail.fr", subject: "test text_body html_body", text_body: "test text_body html_body", html_body: "test text_body html_body")
Plausible.Mailer.deliver_now(email)
```

To be sure of my results, I performed these commands each time with two different destination emails.
In all cases :
- The logs returned by the console do not indicate any errors and return a table of this type :
```text
{:ok,
 %Bamboo.Email{
   from: {"xxx", "noreply@xxx.fr"},
   to: [nil: "xxx@hotmail.fr"],
   cc: [],
   bcc: [],
   subject: "test text_body html_body",            <==== depending of command
   html_body: "test text_body html_body",       <==== depending of command
   text_body: "test text_body html_body",        <==== depending of command
   headers: %{},
   attachments: [],
   assigns: %{},
   private: %{},
   blocked: false
 }}

```
- Only emails sent with both the `text_body` and `html_body` parameters (and not just one or the other) are correctly delivered.
I deduce that on the Plausible side, one of the two parameters is not defined at the time of sending.

I discussed this issue on this link: https://github.com/plausible/analytics/discussions/4125#discussioncomment-9572216
[...]

PR #4674, opened 2024-10-14
Autogenerate text emails

ruslandoga:
### Changes

This PR autogenerates text versions of non-MJML emails.

Closes https://github.com/plausible/analytics/issues/4249

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/email.ex
```diff
@@ -460,7 +460,39 @@ defmodule PlausibleWeb.Email do
...
+    |> Floki.text()
+  end
+
+  defp textify_link(node) do
+    case node do
+      {"a", attrs, children} ->
+        {"href", href} = List.keyfind!(attrs, "href", 0)
+        text = Floki.text(children)
+        text_and_link = IO.iodata_to_binary([text, " (", href, ?)])
+        {"p", attrs, [text_and_link]}
```
ruslandoga, line 492: The result looks like this:

- Click here to upgrade your subscription (http://localhost:8000/billing/choose-plan).
- In your account settings (http://localhost:8000/settings/billing/subscription), you'll find an overview of your usage and limits.
- etc.

lib/plausible_web/email.ex
```diff
@@ -460,7 +460,50 @@ defmodule PlausibleWeb.Email do
...
+    else
+      email |> html_body(html) |> text_body(textify(html))
+    end
+  end
+
+  on_ce do
+    defp textify(html) do
+      html
+      |> Floki.parse_document!()
+      |> Floki.traverse_and_update(&textify_link/1)
```
aerosol, line 483: haha clever. any reason not to do it on EE?

ruslandoga: Didn't want to make EE emails slower :) And EE probably doesn't need it since Postmark might be doing something similar already.

ruslandoga: Ah, no. Postmark only sends HTML. But then, I guess it's still OK since we only had problems with CE users.

aerosol: > Didn't want to make EE emails slower :)

Is there a significant slow down? If so, I'm sure you'd be tempted to generate those text versions statically with a mix task instead or come up with something even more clever ;D

I think including text versions boosts our reputation? Either way, even Postmark recommends it https://postmarkapp.com/guides/transactional-email-best-practices#11-include-well-formatted-plain-text-versions so I really don't think we should be that pedantic about build separation here. cc @ukutaht

ruslandoga: I wanted to sort of pre-compile the text templates into functions like EEx does, but then got paranoid about dynamic parts staying HTML ... I can look into it more. I'll also run some benchmarks to see how much of a performance hit the current "textifying" approach is.

aerosol: haha, ace!

ukutaht: Nice work @ruslandoga!

Yeah I agree with @aerosol, adding text emails to EE will be useful I think. And I don't have any performance concerns here. Maybe I'm overconfident but I don't see how email formatting would be a bottleneck in any way.

lib/plausible_web/templates/email/export_success.html.heex
```diff
@@ -1,6 +1,5 @@
 Your <%= Plausible.product_name() %> export for <%= @site.domain %> is now ready for download.
-Please click <a href={@download_url}>here</a>
-to start the download process.
+<p>Please click <a href={@download_url}>here</a> to start the download process.</p>
```
ruslandoga, line 2: Seems like `Floki.text` preserves formatting within elements. So to avoid too much whitespace from the HEEX formatting change, HTML email templates would need to be updated to use proper p/span/etc. tags, not just plain text.

ruslandoga: Without this p tag change, in the text version, this email is rendered as

```
Hey Jane,



Your Plausible CE export for é-0.example.com is now ready for download.
Please click here (http://localhost:8000/%C3%A9-0.example.com/download/export)
to start the download process.





--

http://localhost:8000

```

ruslandoga: And after wrapping the line in a p tag

```
Hey Jane,



Your Plausible CE export for é-0.example.com is now ready for download.
Please click here (http://localhost:8000/%C3%A9-0.example.com/download/export) to start the download process.



--

http://localhost:8000

```

ruslandoga: I think the ideal version would be

```
Hey Jane,

Your Plausible CE export for é-0.example.com is now ready for download.
Please click here (http://localhost:8000/%C3%A9-0.example.com/download/export) to start the download process.

--

http://localhost:8000
```

lib/plausible_web/email.ex
```diff
@@ -459,17 +459,82 @@ defmodule PlausibleWeb.Email do
...
+    end
+  end
+
+  defp traverse_and_textify({tag, attrs, children}) do
+    {tag, attrs, traverse_and_textify(children)}
   end
+
+  defp traverse_and_textify(other), do: other
+
+  # this is slow but easy to understand:
```
ruslandoga, line 520: Or would something like this be even better?

```elixir
defp collapse_whitespace(text) do
  text
  |> String.split("\n")
  |> Enum.map(fn line ->
    line
    |> String.replace(~r/\s+/u, " ")
    |> String.trim()
  end)
  |> Enum.join("\n")
end
```

lib/plausible_web/email.ex
```diff
@@ -459,17 +459,82 @@ defmodule PlausibleWeb.Email do
...
+        text
+      else
+        IO.iodata_to_binary([text, " (", href, ?)])
+      end
+    else
+      {tag, attrs, children}
+    end
+  end
+
+  defp traverse_and_textify({tag, attrs, children}) do
```
ruslandoga, line 514: `<pre>` should probably be skipped. In case Plausible emails start including code snippets in the future.

lib/plausible_web/email.ex
ukutaht: Yes I think the one in your comment is more understandable

ruslandoga: https://github.com/plausible/analytics/pull/4674/commits/720d280379edde29c60eb17fe10a55136eb8a6f9

lib/plausible_web/email.ex
ruslandoga: `collapse_whitespace/1` would still break it, so maybe next time.

ukutaht (approved):

PR #4678, opened 2024-10-15
Set root domain `(none)` for ingests without hostname

aerosol:
### Changes

This is not a bug, but more of a micro-optimization. If the hostname is known to be `(none)`, there's no need try and compute root domain. 

:eyes: At the same, it looks like ingestion had a dead clause trying to match `nil` hostname. `Plausible.Ingestion.Request` should reject such cases before they reach `Plausible.Ingestion.Event` processing. 

Ref https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/7916301991

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -541,6 +541,7 @@ defmodule Plausible.Ingestion.Event do
     end
   end
 
+  defp get_root_domain("(none)"), do: "(none)"
   defp get_root_domain(nil), do: "(none)"
 
   defp get_root_domain(hostname) do
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -542,7 +542,6 @@ defmodule Plausible.Ingestion.Event do
   end
 
   defp get_root_domain("(none)"), do: "(none)"
-  defp get_root_domain(nil), do: "(none)"
 
   defp get_root_domain(hostname) do
     case :inet.parse_ipv4_address(String.to_charlist(hostname)) do
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -194,11 +194,11 @@ defmodule Plausible.Ingestion.Request do
   defp put_hostname(changeset) do
     host =
       case Changeset.get_field(changeset, :uri) do
-        %{host: host} when is_binary(host) and host != "" -> host
+        %{host: host} when is_binary(host) and host != "" -> sanitize_hostname(host)
         _ -> "(none)"
       end
 
-    Changeset.put_change(changeset, :hostname, sanitize_hostname(host))
+    Changeset.put_change(changeset, :hostname, host)
   end
 
   @max_props 30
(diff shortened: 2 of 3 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -541,6 +541,7 @@ defmodule Plausible.Ingestion.Event do
     end
   end
 
+  defp get_root_domain("(none)"), do: "(none)"
   defp get_root_domain(nil), do: "(none)"
```
aerosol, line 545: :thinking: at this point hostname shouldn't ever be `nil` - rejected at the requested creation level at best

aerosol: 24185ef69

test/plausible/ingestion/event_test.exs
```diff
@@ -397,4 +397,20 @@ defmodule Plausible.Ingestion.EventTest do
...
+      domain: site.domain,
+      name: "pageview",
+      url: "/no/hostname"
+    }
+
+    conn = build_conn(:post, "/api/events", payload)
+    assert {:ok, request} = Request.build(conn)
+
+    assert {:ok, %{buffered: [event]}} = Event.build_and_buffer(request)
+    assert event.clickhouse_event.hostname == "(none)"
```
macobo, line 414: For event properties, utm tags, referrers, channels and other cases we handle this at a query level by making empty strings have special meaning. See lib/plausible/stats/sql/where_builder.ex#db_field_val and lib/plausible/stats/sql/expression.ex#select_dimension

We should solve this similarly.

aerosol: sure, feel free to file a ticket

aerosol: heh, you've edited your comment, there was no question before..

Our first `(none)` hostname in the db dates back to 2021-09. What you're suggesting implies rewriting those AFAIU and is not something I intended working on.

macobo: Would it be possible to solve this similarly or is the issue tied to user session hashing?

If we do it this way, we would also need to migrate historical data, this requires extra space on disk and I argue it makes for more complicated code.

aerosol: > or is the issue tied to user session hashing?

the only issue being solved here is not parsing `(none)` as if it was an IP address

lib/plausible/ingestion/request.ex
```diff
@@ -295,6 +295,8 @@ defmodule Plausible.Ingestion.Request do
     sanitize_hostname(hostname)
   end
 
+  def sanitize_hostname("(none)"), do: "(none)"
```
ukutaht, line 298: `sanitize_hostname` is used for both hostname and domain but the `"(none)"` clause is specific to the hostname field. I think the code would be a bit more cohesive if instead of this new clause we move the call to `sanitize_hostname` from line [201 ](https://github.com/plausible/analytics/blob/0099164cdca60f546dc0d0ebe39aace0bba87afe/lib/plausible/ingestion/request.ex#L201)to [197](https://github.com/plausible/analytics/blob/0099164cdca60f546dc0d0ebe39aace0bba87afe/lib/plausible/ingestion/request.ex#L197). Then we don't need this new clause.

ukutaht (approved):

lib/plausible/ingestion/request.ex
aerosol: ed17b1e9a

PR #4680, opened 2024-10-15
fix(revenue): Fix selecting and graphing revenue goals with comparisons

macobo:
### Changes

This PR fixes two bugs:
1. Revenue metrics were not selectable if filtering by revenue goal with custom display names.
2. Fix timeseries crash when graphing a revenue metric with comparison

Surprisingly, comparisons for revenue metrics graphing had no tests so these were added as well.


lib/plausible/stats/query_runner.ex
```diff
@@ -254,7 +254,18 @@ defmodule Plausible.Stats.QueryRunner do
...
+    |> Enum.map(fn metric -> empty_metric_value(metric) end)
+  end
+
+  on_ee do
+    defp empty_metric_value(metric)
+         when metric in [:total_revenue, :average_revenue],
+         do: nil
+  end
+
+  defp empty_metric_value(_), do: 0
```
apata, line 268: **question, non-blocking**: what's the difference between nil and 0 metric value?

apata (approved):

PR #4687, opened 2024-10-16
Remove space before periods in site setup success email

ruslandoga:
This PR removes space before periods when they come after links by moving the period inside the link: https://github.com/plausible/analytics/pull/4675/files#r1802568680

Screenshot 2024-10-16 at 15 16 09

PR #4690, opened 2024-10-16
Use anchor tags in email templates

ruslandoga:
Context: https://github.com/plausible/analytics/pull/4687#issuecomment-2416131752

### DIFF
--- a/lib/plausible_web/templates/email/google_analytics_import.html.heex
+++ b/lib/plausible_web/templates/email/google_analytics_import.html.heex
@@ -3,7 +3,7 @@
     @site_import.start_date
   ) %> to <%= date_format(@site_import.end_date) %>
   <br /><br />
-  <.styled_link href={@link}>Click here</.styled_link>
+  <a href={@link}>Click here</a>
   to view your dashboard.
 <% else %>
   Unfortunately, your Google Analytics import for <%= @site_import.site.domain %> did not complete successfully. Sorry about that!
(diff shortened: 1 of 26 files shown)


lib/plausible_web/templates/email/google_analytics_import.html.heex
```diff
@@ -3,7 +3,7 @@
     @site_import.start_date
   ) %> to <%= date_format(@site_import.end_date) %>
   <br /><br />
-  <.styled_link href={@link}>Click here</.styled_link>
+  <a href={@link}>Click here</a>
```
ruslandoga, line 6: I replaced with `styled_link` as well since it we don't have access to Tailwind classes in emails.

aerosol (approved):
Thank you @ruslandoga

aerosol:
@ruslandoga looks like format is needed 😓

lib/plausible_web/templates/email/ownership_transfer_request.html.heex
```diff
@@ -1,15 +1,15 @@
...
   to view and respond to the invitation.
 <% else %>
-  <.unstyled_link href={
-    Routes.auth_url(PlausibleWeb.Endpoint, :register_form, invitation: @invitation.invitation_id)
-  }>
-    Click here
-  </.unstyled_link>
-  to create your account. <br /><br />
+  <a
+    phx-no-format
```
ruslandoga, line 7: There might be a bug in HEEX formatter since I couldn't get this tag to preserve its content without adding `phx-no-foramt`. Seems to be because the anchor is inside an if/else block.

aerosol: Perhaps we can disable formatting for email files?

ruslandoga: Oh, I didn't know this was an option :)

However, this anchor tag is the only one that has this problem, so it might be easier to leave this as is. I'd like to make it work with the default configuration.

aerosol (approved):

PR #4691, opened 2024-10-16
Replace br with p in emails

ruslandoga:
This makes the email more readable as they get formatted more predictably. And this is also how GitHub renders HTML emails.


aerosol:
Nice, I think we should give it a try!

PR #4697, opened 2024-10-17
Comparisons pagination fixes

macobo:
### Changes

Depends on https://github.com/plausible/analytics/pull/4692

Currently, when doing comparisons when breaking down by e.g. `visit:browser` dimension, zeroes would be incorrectly applied when small limits are in place. This is because the top 10 browsers today might not be the same as in previous period.

This PR fixes that by OR-ing together all dimension values from main result in comparison filter. Additionally, this filter is ignored when calculating totals as is needed for percentage/conversion rate calculations.

Rather than test every old endpoint separately, this PR adds a test-only endpoint for testing query logic as well.

### DIFF
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -57,6 +57,45 @@ defmodule Plausible.Stats.Comparisons do
     |> maybe_include_imported(source_query)
   end
 
+  @doc """
+  Builds comparison query after executing `main` query.
+
+  When querying for comparisons with dimensions and pagination, extra
+  filters are added to ensure comparison query returns same set of results
+  as main query.
+  """
+  def add_comparison_filters(comparison_query, main_results_list) do
+    comparison_filters =
+      Enum.flat_map(main_results_list, &build_comparison_filter(&1, comparison_query))
+
+    comparison_query
+    |> add_query_filters(comparison_filters)
+  end
+
+  defp add_query_filters(query, []), do: query
+
+  defp add_query_filters(query, [filter]) do
+    Query.add_filter(query, [:ignore_in_totals_query, filter])
+  end
+
+  defp add_query_filters(query, filters) do
+    Query.add_filter(query, [:ignore_in_totals_query, [:or, filters]])
+  end
+
+  defp build_comparison_filter(%{dimensions: dimension_labels}, query) do
+    query_filters =
+      query.dimensions
+      |> Enum.zip(dimension_labels)
+      |> Enum.reject(fn {dimension, _label} -> String.starts_with?(dimension, "time") end)
+      |> Enum.map(fn {dimension, label} -> [:is, dimension, [label]] end)
+
+    case query_filters do
+      [] -> []
+      [filter] -> [filter]
+      filters -> [[:and, filters]]
+    end
+  end
+
   defp get_comparison_date_range(source_query, %{mode: "year_over_year"} = options) do
     source_date_range = Query.date_range(source_query)
 
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -58,7 +58,7 @@ defmodule Plausible.Stats.Comparisons do
   end
 
   @doc """
-  Builds comparison query after executing `main` query.
+  Builds comparison query that specifically filters for values appearing in the main query results.
 
   When querying for comparisons with dimensions and pagination, extra
   filters are added to ensure comparison query returns same set of results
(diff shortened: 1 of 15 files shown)


test/plausible/stats/comparisons_test.exs
```diff
@@ -243,4 +224,112 @@ defmodule Plausible.Stats.ComparisonsTest do
...
+      result_query =
+        Comparisons.add_comparison_filters(query, [
+          %{
+            dimensions: ["Chrome", "99.9", "2024-01-01"],
+            metrics: [123]
+          },
+          %{
+            dimensions: ["Firefox", "12.0", "2024-01-01"],
+            metrics: [123]
+          }
```
apata, line 285: **meganitpick, non-blocking:** This important test may be easier to understand if we lift this into a named variable like `main_query_results = [...]`

lib/plausible/stats/query.ex
```diff
@@ -93,7 +93,7 @@ defmodule Plausible.Stats.Query do
   def remove_top_level_filters(query, prefixes) do
     new_filters =
       Enum.reject(query.filters, fn [_, filter_key | _rest] ->
-        Enum.any?(prefixes, &String.starts_with?(filter_key, &1))
+        is_binary(filter_key) and Enum.any?(prefixes, &String.starts_with?(filter_key, &1))
```
apata, line 96: **question:** Is this change connected with the fix, or just an additional guard that is good to have?

lib/plausible/stats/comparisons.ex
```diff
@@ -57,6 +57,45 @@ defmodule Plausible.Stats.Comparisons do
...
+
+  defp add_query_filters(query, []), do: query
+
+  defp add_query_filters(query, [filter]) do
+    Query.add_filter(query, [:ignore_in_totals_query, filter])
+  end
+
+  defp add_query_filters(query, filters) do
+    Query.add_filter(query, [:ignore_in_totals_query, [:or, filters]])
+  end
```
apata, line 83: Super impressed how a fairly complex thing can be resolved so cleanly! 

It took me a few tries until I got it though, and I'm wondering if the atom name :ignore_in_totals_query has anything to do with it. It gets me thinking about totals, not about selecting for values that appear in the original query.

lib/plausible/stats/comparisons.ex
```diff
@@ -57,6 +57,45 @@ defmodule Plausible.Stats.Comparisons do
     |> maybe_include_imported(source_query)
   end
 
+  @doc """
+  Builds comparison query after executing `main` query.
```
apata, line 61:
```suggestion
  Builds comparison query that specifically filters for values appearing in the main query results.
```

lib/plausible/stats/comparisons.ex
```diff
@@ -57,6 +57,45 @@ defmodule Plausible.Stats.Comparisons do
...
+
+  defp add_query_filters(query, filters) do
+    Query.add_filter(query, [:ignore_in_totals_query, [:or, filters]])
+  end
+
+  defp build_comparison_filter(%{dimensions: dimension_labels}, query) do
+    query_filters =
+      query.dimensions
+      |> Enum.zip(dimension_labels)
+      |> Enum.reject(fn {dimension, _label} -> String.starts_with?(dimension, "time") end)
```
apata, line 89: **nitpick, non-blocking**: I noticed this `String.starts_with?(..., "time")` logic is duplicated in several places. Is there a good place to define a shared `is_time_dimension?(d)`?

lib/plausible_web/router.ex
```diff
@@ -208,16 +208,22 @@ defmodule PlausibleWeb.Router do
...
 
-  scope "/api/v2", PlausibleWeb.Api, assigns: %{api_scope: "stats:read:*"} do
+  scope "/api/v2", PlausibleWeb.Api, assigns: %{api_scope: "stats:read:*", schema_type: :public} do
     pipe_through [:public_api, PlausibleWeb.Plugs.AuthorizePublicAPI]
 
     post "/query", ExternalQueryApiController, :query
+
+    if Mix.env() in [:test, :ce_test] do
+      scope assigns: %{schema_type: :internal} do
+        post "/query-internal", ExternalQueryApiController, :query
```
apata, line 218: **suggestion, non-blocking**: If we include the word `test` in the endpoint URL, it's extra clear that it's to be maintained for testing purposes only.

lib/plausible/stats/query.ex
macobo: It is related - previously this branch was never hit with `or/and/not` queries and it blew up after comparisons started adding them here. :)

apata (approved):

PR #4704, opened 2024-10-21
Channels: Write `click_id_param` column

macobo:
### Changes

This follows up on https://github.com/plausible/analytics/pull/4703 and starts writing `click_id_source` column.

Valid values currently are for example gclid and msclkid, this will be set when pageview comes and has an appropritate click_id query param set denoting it's a paid search. This will be used to calculate acquisition_channel in clickhouse, see https://github.com/plausible/analytics/pull/4701.


ukutaht (approved):
Looks good!

PR #4714, opened 2024-10-22
Improve GA4 import error handling

zoldar:
### Changes

This PR improves error handling in GA4 imports.

- Fetch errors are consistently logged with "warning" level
- socket and 5xx errors are treated as recoverable (even after retry-with-backoff fails) - the import is resumed after and hour, like in case of rate limit being exceeded

### Tests


RobertJoonas (approved):
🙌

PR #4719, opened 2024-10-23
Breakdown comparisons: Tooltip UX, arrows 

macobo:
### Changes

This PR:
- Updates arrows across the app
image

- Fixes a small bug when calculating bounce_rate in comparisons
- Returns main and comparison date ranges in breakdown endpoints
- Updates comparison tooltip designs to include comparison date ranges
image

- Removes wavy arrow from being shown when no change in breakdown lists
- Makes Graph tooltip render via react for consistency and easier reuse of arrows
image

Notes:
- I also experimented with varying stroke widths for arrows, but IMO it did not look good. Feel free to build upon this: https://github.com/plausible/analytics/pull/4719/commits/23a97ba506d2fd9042cde598639bb75a28ddaa33
- The screenshots show some oddity around comparison date range calculations. ~~This seems like an older bug, investigating separately.~~ This comes from day-of-week matching, not actioning this.

<details><summary>Click here for more screenshots</summary>

### 1. Arrows
image
image
image
image
image
image

</details>

### DIFF
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -43,17 +43,22 @@ defmodule Plausible.Stats.Breakdown do
   end
 
   def formatted_date_ranges(query) do
-    comparison_date_range =
-      if query.include.comparisons do
-        Comparisons.get_comparison_query(query, query.include.comparisons).utc_time_range
-      else
-        nil
-      end
-
-    %{
-      date_range: format_date_range(query.utc_time_range, query.timezone),
-      comparison_date_range: format_date_range(comparison_date_range, query.timezone)
+    formatted = %{
+      date_range_label: format_date_range(query.utc_time_range, query.timezone)
     }
+
+    if query.include.comparisons do
+      comparison_date_range =
+        Comparisons.get_comparison_query(query, query.include.comparisons).utc_time_range
+
+      Map.put(
+        formatted,
+        :comparison_date_range_label,
+        format_date_range(comparison_date_range, query.timezone)
+      )
+    else
+      formatted
+    end
   end
 
   defp build_breakdown_result(query_result, query, metrics) do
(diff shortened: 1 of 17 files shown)


lib/plausible/stats/breakdown.ex
```diff
@@ -135,4 +156,23 @@ defmodule Plausible.Stats.Breakdown do
...
+  end
+
+  defp string_format_date_range(%Date.Range{first: first, last: last})
+       when first.year == last.year do
+    "#{Calendar.strftime(first, "%d %b")} - #{Calendar.strftime(last, "%d %b")}"
+  end
+
+  defp string_format_date_range(%Date.Range{first: first, last: last}) do
+    "#{Calendar.strftime(first, "%d %b %Y")} - #{Calendar.strftime(last, "%d %b %Y")}"
+  end
```
apata, line 177: **nitpick, non-blocking:** This is a little different from what the FE formatDateRange (that's used in top-stats.js and date picker) yields. 
Some differences:
* FE doesn't render leading zeros on days
* FE renders short years
* FE renders year whenever it's not the current year

Here's an example of how omitting the year when first.year == last.year may be confusing:
Screenshot 2024-10-23 at 13 05 52

apata (approved):

lib/plausible/stats/breakdown.ex
apata: On the same subject, the trim trailing fixes labels that we get from the BE, but the ones in top graph remain wrong (Oct 1 to Oct 31).

macobo: Thanks!

- I've updated the date formats
- I opted to use long formats since the reason behind short formats above seems to be to save space in the tight top stats
- I didn't trim other labels as it would change existing behavior (IMO intentional behavior)

PR #4720, opened 2024-10-23
Fix custom property value filter suggestions

RobertJoonas:
### Changes

This has been broken since filtering by multiple custom props became possible. 

To reproduce the bug:

1. Go to the live demo
2. Open the filter modal and apply a filter like "browser_language is en-US"
3. Click on the filter card above the dashboard to edit the filter (the filter modal will open again)
4. Click "+ Add another"
5. Select a property other than `browser_language`, e.g. `logged_in`
6. Click on the prop value field to fetch suggestions
7. BUG: suggestions are fetched for `browser_language` rather than `logged_in`

image

The current logic in the `/:domain/suggestions` endpoint for custom prop values is not accounting for the fact that there might be multiple filters with the `event:props` prefix.

This PR fixes that, introducing a new endpoint with a `prop_key` path parameter which indicates for what `prop_key` we actually want the value suggestions returned.

The old endpoint will be removed in a follow-up PR.

This PR also adds test coverage to the searching functionality for both `prop_key` and `prop_value` suggestions.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
-- Fix returning filter suggestions for custom property values in the dashbaord Filter modal
+- Fix returning filter suggestions for multiple custom property values in the dashboard Filter modal
 
 ## v2.1.4 - 2024-10-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+- Fix returning filter suggestions for custom property values in the dashbaord Filter modal
```
macobo, line 11:
```suggestion
- Fix returning filter suggestions for multiple custom property values in the dashboard Filter modal
```

macobo (approved):

PR #4724, opened 2024-10-23
Remove experimental_reduced_joins flag

macobo:


macobo:
Quite a bit of test breakage. Will investigate later.

PR #4729, opened 2024-10-23
Show filters on a separate line

apata:
### Changes

* Shows Filters on a separate line if feature flag `saved_segments` is true
* Makes minor adjustments to elements on the first line (rolled out to all, too minor to guard with FF)
  * icons won't shrink on small screens
  * elements have fixed height at `h-9`
* Makes the dropdown component less opinionated: more configurable. 

### Tests

### Changelog
- [x] Changelog entry not needed, because changes are behind FF

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/filters.js
+++ b/assets/js/dashboard/filters.js
@@ -208,7 +208,7 @@ function Filters() {
       const filterCount = query.filters.length
       return (
         <>
-          <AdjustmentsVerticalIcon className="-ml-1 mr-1 h-4 w-4" aria-hidden="true" />
+          <AdjustmentsVerticalIcon className="-ml-1 mr-1 h-4 w-4 shrink-0" aria-hidden="true" />
           {filterCount} Filter{filterCount === 1 ? '' : 's'}
         </>
       )
@@ -216,7 +216,7 @@ function Filters() {
 
     return (
       <>
-        <MagnifyingGlassIcon className="-ml-1 mr-1 h-4 w-4 md:h-4 md:w-4" aria-hidden="true" />
+        <MagnifyingGlassIcon className="-ml-1 mr-1 h-4 w-4 shrink-0" aria-hidden="true" />
         {/* This would have been a good use-case for JSX! But in the interest of keeping the breakpoint width logic with TailwindCSS, this is a better long-term way to deal with it. */}
         <span className="sm:hidden">Filter</span><span className="hidden sm:inline-block">Filter</span>
       </>
--- a/assets/js/dashboard/nav-menu/filter-pills-list.tsx
+++ b/assets/js/dashboard/nav-menu/filter-pills-list.tsx
@@ -1,0 +1,96 @@
+/** @format */
+
+import React, { DetailedHTMLProps, HTMLAttributes } from 'react'
+import { useQueryContext } from '../query-context'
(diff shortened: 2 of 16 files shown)


assets/js/dashboard/filters.js
```diff
@@ -208,15 +208,15 @@ function Filters() {
       const filterCount = query.filters.length
```
apata, line 208: **note**: This whole file can be deleted once the feature is rolled out to all.

assets/js/dashboard/nav-menu/filter-pills-list.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const PILL_X_GAP = 16
+export const PILL_Y_GAP = 8
+
+/** Restricts output to slice of DashboardQuery['filters'], or makes the output outside the slice invisible */
+type Slice = {
+  /** The beginning index of the specified portion of the array. If start is undefined, then the slice begins at index 0. */
+  start?: number
+  /** The end index of the specified portion of the array. This is exclusive of the element at the index 'end'. If end is undefined, then the slice extends to the end of the array. */
+  end?: number
```
apata, line 24: **note:** This is the same behavior and API as Array.slice and the JSDoc hint is copied from that function. 

I'm not the happiest with this component's prop definition, but I didn't see the point of fiddling with it any longer.

assets/test-utils/index.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const customRender = (ui: ReactNode, options: RenderOptions) => {
+  const output = render(ui, options)
+  registerPartialTailwindStyle()
+  return output
+}
+
+export * from '@testing-library/react'
+export { customRender as render }
```
macobo, line 25: Q: Could this cause issues for tests that don't require tailwind?

assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -0,0 +1,218 @@
...
+              ) : null}
+            </ToggleDropdownButton>
+          )}
+        <ClearAction />
+      </div>
+    </div>
+  )
+}
+
+export const ClearAction = () => (
```
macobo, line 206: Nit: The positioning of this seems too far with the flag on:

image

macobo (commented):
Looks good to me, flag works well.

Nit: It's also worthwhile including interesting screenshots into pure-FE PRs. This helps reviewers to make sure they're not missing something important.

assets/test-utils/index.ts
apata: Thanks for the question! All tests that use render will benefit from the DOM-under-test being more like the real thing. For other unit tests, since this fn isn't called, there's no effect.

assets/js/dashboard/nav-menu/filters-bar.tsx
apata: I agree that it looks a bit lonely right now, but the design spec is that it's a button that's 9 wide and 9 tall, and the whole bar has a horizontal gap of 4. 

On my dev branch for Saved Segments, when there's stuff to the right of it, it looks better. Since it's behind a feature flag, I think we can circle back to fine-tuning individual elems once segments are in place as well.

macobo (approved):

PR #4731, opened 2024-10-24
Delay "Add Goal" button activation till wss connects

aerosol:
Perhaps still better than indefinite spinner in high latency conditions...

reading material: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7954549712

### DIFF
--- a/lib/plausible_web/live/goal_settings/list.ex
+++ b/lib/plausible_web/live/goal_settings/list.ex
@@ -27,7 +27,7 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
         >
           Add Goal
         </.button>
-        <.button :if={connected?(@socket)} id="add-goal-button" mt?={false}>
+        <.button :if={not connected?(@socket)} id="add-goal-button" mt?={false}>
           Add Goal
         </.button>
       </.filter_bar>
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -26,6 +27,9 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
         >
           Add Goal
         </.button>
+        <.button :if={not connected?(@socket)} id="add-goal-button" mt?={false}>
```
zoldar, line 30: Shouldn't that be

```suggestion
        <.button :if={connected?(@socket)} id="add-goal-button" mt?={false}>
```

instead? 🤔

aerosol: :man_facepalming:

aerosol: lmao github broke - original code:

image

zoldar: I take that back, the condition was valid, I have not connected the dots that it's a button in _inactive_ state 🤡

zoldar (approved):

PR #4744, opened 2024-10-28
Add tests for the pageleave script

RobertJoonas:
### Changes

* Add a test suite to verify `pageleave` events sent in different scenarios
* Update Playwright to the latest version (1.48), and tweak the config a bit to make it work with the VSCode plugin
* Add a new NPM script to run test with a possibility of skipping tracker file generation.
* Speed up the tracker test suite - wait for 3s for requests instead of 10s. I think it was 10s because we were previously using BrowserStack and the tests were executed remotely. Given that Playwright is now running tests locally, there's no need to wait that much.

### Tests

### Changelog

### Documentation

### Dark mode


tracker/dev-compile/can-skip-compile.js
```diff
@@ -0,0 +1,52 @@
...
+const LAST_HASH_FILEPATH = path.join(__dirname, './last-hash.txt')
+
+// Re-compilation is only required if any of these files have been changed. 
+const COMPILE_DEPENDENCIES = [
+  path.join(__dirname, '../compile.js'),
+  path.join(__dirname, '../src/plausible.js'),
+  path.join(__dirname, '../src/customEvents.js')
+]
+
+async function currentHash() {
```
apata, line 14: **nitpick, non-blocking**: does not need to be async.

apata (approved):

tracker/dev-compile/can-skip-compile.js
RobertJoonas: Oh, right. Thanks!  https://github.com/plausible/analytics/pull/4744/commits/91180e989b930cdfb2e47b108c86f40059360f96

PR #4745, opened 2024-10-28
Add size reporter GH action

RobertJoonas:
### Changes

Currently opted for a selection of filenames that are "interesting to know". Printing out file sizes for every file is pretty noisy (> 2000 lines) and irrelevant, I think. The default `plausible.js` script size is the one that matters.


aerosol (approved):
nice!

PR #4746, opened 2024-10-28
Limit maximum date to today in calendars

apata:
### Changes

Limits maximum date in calendars to today.

### Tests
- [x] Tests added

### Changelog
- [x] This PR does makes a user-facing change, but the change is very small

### Documentation

### Dark mode


ukutaht:
The design principle we've landed on is that everything in the dashboard should be scoped to the timezone of the site itself, ignoring the client's time zone completely. So we should be limiting the datepicker to "today" in the timezone that the site is configured.

With this implementation the "today" limit will change depending on the time zone I set my computer to. It's slightly confusing to me because `toISOString` should make it UTC but the calendar library is probably parsing it back in local timezone? Anyways, in this example I set my machine to NZ summer time (where "now" is 31st Oct) while the site stayed in Tallinn time (where "now" is 30 Oct). I would expect the date picker to limit to 30 Oct but it's 31st:

Screenshot 2024-10-31 at 00 10 17

apata:
> The design principle we've landed on is that everything in the dashboard should be scoped to the timezone of the site itself, ignoring the client's time zone completely. So we should be limiting the datepicker to "today" in the timezone that the site is configured.

Thanks for pointing it out, @ukutaht! It should be fixed now.

ukutaht (approved):

PR #4752, opened 2024-10-29
[NOT READY] Playwright

macobo:
Please ignore for now.


apata:
Good idea to have e2e tests! Since you mentioned having issues with the seeded data, how about we run it without seeded data? 

We could run a mock website at localhost:80 with a particular version of the tracker script installed. Tests would look something like: 

* Visit Plausible-under-test.
* Create Plausible account.
* Log in.
* Add site localhost. 
* Visit localhost/blog.
* Visit Plausible dashboard.
* Expect /blog to appear in Pages report.

macobo:
Note that this PR is not ready yet - I will open it up as a proposal once it is!

PR #4758, opened 2024-10-31
Fix sites list for needs to upgrade condition check

zoldar:


metmarkosaric (approved):

macobo (approved):
Could we add a test as well?

zoldar:
@macobo I'll add a test covering this case in a follow-up as it was clearly missed.

PR #4762, opened 2024-10-31
Hard-code sample rate based on fractional_hardcoded_sample_rate flag

macobo:
We found cases where using a numeric sample rate would cause issues when joining two tables due to different _sample_factor. Ref: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/7973456592#__recording_7978780711

The proper fix is to use fractional sample rates everywhere, but this is a whole project due to not wanting to sample small sites. For now, hard-code sample rate for specific sites having the issue [while we work on a larger fix](https://3.basecamp.com/5308029/buckets/39750953/messages/7978775089).

### DIFF
--- a/extra/lib/plausible/stats/sampling.ex
+++ b/extra/lib/plausible/stats/sampling.ex
@@ -17,8 +17,9 @@ defmodule Plausible.Stats.Sampling do
     end
   end
 
-  @spec add_query_hint(Ecto.Query.t(), pos_integer()) :: Ecto.Query.t()
-  def add_query_hint(%Ecto.Query{} = query, threshold) when is_integer(threshold) do
+  @spec add_query_hint(Ecto.Query.t(), pos_integer() | float()) :: Ecto.Query.t()
+  def add_query_hint(%Ecto.Query{} = query, threshold)
+      when is_integer(threshold) or is_float(threshold) do
     from(x in query, hints: unsafe_fragment(^"SAMPLE #{threshold}"))
   end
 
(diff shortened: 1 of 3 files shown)


extra/lib/plausible/stats/sampling.ex
```diff
@@ -17,8 +17,9 @@ defmodule Plausible.Stats.Sampling do
     end
   end
 
-  @spec add_query_hint(Ecto.Query.t(), pos_integer()) :: Ecto.Query.t()
-  def add_query_hint(%Ecto.Query{} = query, threshold) when is_integer(threshold) do
+  @spec add_query_hint(Ecto.Query.t(), pos_integer() | float()) :: Ecto.Query.t()
+  def add_query_hint(%Ecto.Query{} = query, threshold)
+      when is_integer(threshold) or is_float(threshold) do
```
zoldar, line 22: A minor detail, feel free to ignore - `is_number/1` should work here too https://www.erlang.org/doc/apps/erts/erlang#is_number/1

extra/lib/plausible/stats/sampling.ex
```diff
@@ -27,15 +28,32 @@ defmodule Plausible.Stats.Sampling do
...
-        "infinite" -> :infinite
-        value -> String.to_integer(value)
+        nil ->
+          site_default_threshold(site)
+
+        "infinite" ->
+          :infinite
+
+        value_string ->
+          {value, _} = Float.parse(value_string)
```
zoldar, line 43: Just to double check - is CH ok with >1 sample values being passed as floats (with trailing .0)?

zoldar (approved):

extra/lib/plausible/stats/sampling.ex
macobo: Good check! Tested, it works though not confident of the semantics.

PR #4763, opened 2024-10-31
Run full test suite against `read_team_schemas` FF on CI

zoldar:
### Changes

This PR introduces full test suite runs with `read_team_schemas` feature flag up and down on CI. It's meant to reduce the risk of regressions during future further changes.

We are replacing the existing setup for `experimental_reduced_joins` which wasn't actively used and will be removed some time in the future. This way we don't make CI run longer than necessary.

### Tests

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -6,14 +6,12 @@ end
 Mox.defmock(Plausible.HTTPClient.Mock, for: Plausible.HTTPClient.Interface)
 Application.ensure_all_started(:double)
 
-# Temporary flag to test `experimental_reduced_joins` flag on all tests.
-if System.get_env("TEST_EXPERIMENTAL_REDUCED_JOINS") == "1" do
-  FunWithFlags.enable(:experimental_reduced_joins)
+# Temporary flag to test `read_team_schemas` flag on all tests.
+if System.get_env("TEST_READ_TEAM_SCHEMAS") == "1" do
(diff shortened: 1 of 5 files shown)


zoldar:
CI failed with regression reintroduced, like expected: https://github.com/plausible/analytics/actions/runs/11611306731/job/32332366332?pr=4763

test/test_helper.exs
```diff
@@ -6,15 +6,13 @@ end
 Mox.defmock(Plausible.HTTPClient.Mock, for: Plausible.HTTPClient.Interface)
 Application.ensure_all_started(:double)
 
-# Temporary flag to test `experimental_reduced_joins` flag on all tests.
-if System.get_env("TEST_EXPERIMENTAL_REDUCED_JOINS") == "1" do
-  FunWithFlags.enable(:experimental_reduced_joins)
+# Temporary flag to test `read_team_schemas` flag on all tests.
+if System.get_env("TEST_READ_TEAM_SCHEMAS") == "1" do
+  FunWithFlags.enable(:read_team_schemas)
```
macobo, line 11: Let's enable both flags please! The previous one isn't yet safe to remove as requires a bit of work on my end with an upcoming project.

zoldar: https://github.com/plausible/analytics/pull/4763/commits/5daaf674734a71f646bd31da7cfc31e9205a24f1

macobo (approved):

PR #4766, opened 2024-11-03
Remove support@plausible.io contact from CE on site domain error

ruslandoga:
Context: https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7982856649


lib/plausible/site.ex
```diff
@@ -82,9 +83,15 @@ defmodule Plausible.Site do
 
   def new(params), do: changeset(%__MODULE__{}, params)
 
-  @domain_unique_error """
-  This domain cannot be registered. Perhaps one of your colleagues registered it? If that's not the case, please contact support@plausible.io
-  """
+  @domain_unique_error on_ee do
+    """
+    This domain cannot be registered. Perhaps one of your colleagues registered it? If that's not the case, please contact support@plausible.io
```
ruslandoga, line 88: This is the one mentioned in the ticket.

lib/plausible_web/controllers/billing_controller.ex
```diff
@@ -78,7 +78,11 @@ defmodule PlausibleWeb.BillingController do
 
       _ ->
         msg =
-          "Something went wrong with loading your plan change information. Please try again, or contact us at support@plausible.io if the issue persists."
+          if Plausible.ee?() do
+            "Something went wrong with loading your plan change information. Please try again, or contact us at support@plausible.io if the issue persists."
```
ruslandoga, line 82: I think instead of "treating symptoms" like I did here, the billing pages can be removed completely from CE. Right now they seem to be hidden but still accessible directly.

Screenshot 2024-11-03 at 15 32 52

ruslandoga: I think I'll remove these changes, and open another PR to only route `/billing` requests on EE.

ruslandoga: Removed these changes in https://github.com/plausible/analytics/pull/4766/commits/051c43c8edd038d4f599e44772da9aec304f3039

ruslandoga: And created a card to hide the billings routes: https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7986914457

ukutaht (approved):

PR #4782, opened 2024-11-06
Channels: Backfill click_id_param based on historical data

macobo:
This will allow channels calculated in clickhouse to be accurate.

Tested via this query:

```sql
select referrer_source, count() as t, transform(referrer_source, ['Google', 'Bing'], ['gclid', 'msclkid'], '')
from events_v2
where timestamp > now() - toIntervalDay(30) AND channel = 'Paid Search' AND click_id_param = ''
group by referrer_source
order by t desc
limit 5
```

### DIFF
--- a/priv/ingest_repo/migrations/20241106113750_backfill_click_id_param.exs
+++ b/priv/ingest_repo/migrations/20241106113750_backfill_click_id_param.exs
@@ -1,0 +1,24 @@
+defmodule Plausible.IngestRepo.Migrations.BackfillClickIdParam do
+  use Ecto.Migration
+
+  def up do
+    events_sql = """
+      ALTER TABLE events_v2
+      UPDATE click_id_param = transform(referrer_source, ['Google', 'Bing'], ['gclid', 'msclkid'], '')
+      WHERE channel = 'Paid Search' AND click_id_param = ''
+    """
+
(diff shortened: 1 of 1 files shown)


ukutaht:
Looks like this will add the click ID param for every event/session with `source=Google/Bing and channel=Paid Search`. But traffic from those sources can be paid search for 2 reasons other than click_id_param:
1. Paid UTM medium
2. Paid utm_source like "google-ads"

I think the migration should exclude events/sessions where either of the two conditions above are met. It's impossible to know whether those sessions did have a click ID param or not, and I don't think we should add it if we can't guarantee it was there.

ukutaht:
Also I think the migration can be limited in time to where we started collecting channel data in Elixir which was a couple months ago. `channel="Paid Search"` will never match earlier than that.

macobo:
Yikes, good catch. Checking for paid sources/utms is right now not trivial in the migration. As such, closing this PR for now - I think this makes more sense once the channel functions are created and available in other migrations.

Will return to this once the other wing of the work has wrapped up.

macobo:
New query here: https://3.basecamp.com/5308029/buckets/39750953/card_tables/cards/7998407512, planning on running this only in prod once functions are live.

PR #4788, opened 2024-11-07
Extract schema transitions under delegated namespace

aerosol:
### Changes

See if a separate namespace helps with reasoning about transitioning new schema reads

### DIFF
--- a/lib/plausible_web/live/sites.ex
+++ b/lib/plausible_web/live/sites.ex
@@ -6,11 +6,7 @@ defmodule PlausibleWeb.Live.Sites do
   use PlausibleWeb, :live_view
   import PlausibleWeb.Live.Components.Pagination
 
-  alias Plausible.Auth
-  alias Plausible.Site
   alias Plausible.Sites
-  alias Plausible.Site.Memberships.Invitations
-  alias Plausible.Teams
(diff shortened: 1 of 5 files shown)


lib/plausible_web/live/sites.ex
```diff
@@ -726,32 +695,12 @@ defmodule PlausibleWeb.Live.Sites do
...
-    user_or_team =
-      if Teams.read_team_schemas?(user) do
-        case Teams.get_by_owner(user) do
-          {:ok, team} -> team
-          {:error, _} -> nil
-        end
-      else
-        user
-      end
+  defdelegate check_feature_access(site, user), to: Plausible.Teams.Adapter.Read.Ownership
```
zoldar, line 700: Hmm why this and not `Plausible.Teams.Adapter.Read.Billing` ?

aerosol: I thought it's only relevant in context of ownership transfer. Even the function signature has `new_owner` as the key parameter.

zoldar: Ah, fair point, makes sense

zoldar (approved):

PR #4791, opened 2024-11-07
Average Scroll Depth Metric

RobertJoonas:
### Changes

- [x] Modify the `pageleave` script variant to start tracking scroll depth
- [x] Start ingesting pageleaves with a `scroll_depth` value from 0 to 100
- [x] Add querying logic for `scroll_depth` (also expose it in Query API v2 under a user-based feature flag)
- [x] Show scroll depth on the dashboard - Top Pages > Details (always); Top Stats / Main Graph (only when page filter applied)
- [ ] TODO: Configure feature flag access to scroll depth on the dashboard
- [ ] TODO: Extract migration

There are still other things missing (e.g. making scroll depth sortable, Dashboard CSV export, Full CSV import/export, dealing with scroll depth when imported data is included, etc...), but since it's feature flagged anyway, we could already merge the bigger chunk of this and iron out the smaller details later. Then we'll already have some real scroll depth data to work with.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/modals/breakdown-modal.tsx
+++ b/assets/js/dashboard/stats/modals/breakdown-modal.tsx
@@ -135,7 +135,7 @@ export default function BreakdownModal<TListItem extends { name: string }>({
       {
         label: reportInfo.dimensionLabel,
         key: 'name',
-        width: 'w-48 md:w-80 flex items-center break-all',
+        width: 'w-48 md:w-full flex items-center break-all',
         align: 'left',
         renderItem: (item) => (
           <NameCell
--- a/lib/plausible/stats/sql/special_metrics.ex
+++ b/lib/plausible/stats/sql/special_metrics.ex
@@ -16,6 +16,7 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
     |> maybe_add_percentage_metric(site, query)
     |> maybe_add_global_conversion_rate(site, query)
     |> maybe_add_group_conversion_rate(site, query)
+    |> maybe_add_scroll_depth(site, query)
   end
 
   defp maybe_add_percentage_metric(q, site, query) do
@@ -119,6 +120,71 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
     else
       q
     end
+  end
+
+  def maybe_add_scroll_depth(q, site, query) do
+    cond do
+      :scroll_depth not in query.metrics -> q
+      query.dimensions == [] -> add_aggregate_scroll_depth(q, site, query)
+      true -> add_group_scroll_depth(q, site, query)
+    end
+  end
+
+  defp add_aggregate_scroll_depth(q, site, query) do
+    max_per_visitor_q =
+      Base.base_event_query(site, query)
+      |> where([e], e.name == "pageleave")
+      |> select([e], %{
+        user_id: e.user_id,
+        max_scroll_depth: max(e.scroll_depth)
+      })
+      |> group_by([e], e.user_id)
+
+    scroll_depth_q =
+      subquery(max_per_visitor_q)
+      |> select([p], fragment("toUInt8(round(ifNotFinite(avg(?), 0)))", p.max_scroll_depth))
+
+    select_merge_as(q, [e], %{scroll_depth: subquery(scroll_depth_q)})
+  end
+
+  defp add_group_scroll_depth(q, site, query) do
+    max_per_visitor_q =
+      Base.base_event_query(site, query)
+      |> where([e], e.name == "pageleave")
+      |> select([e], %{
+        user_id: e.user_id,
+        max_scroll_depth: max(e.scroll_depth)
+      })
+      |> SQL.QueryBuilder.build_group_by(:events, query)
+      |> group_by([e], e.user_id)
+
+    dim_shortnames = Enum.map(query.dimensions, fn dim -> shortname(query, dim) end)
+
+    dim_select =
+      dim_shortnames
+      |> Enum.map(fn dim -> {dim, dynamic([p], field(p, ^dim))} end)
+      |> Map.new()
+
+    dim_group_by =
+      dim_shortnames
+      |> Enum.map(fn dim -> dynamic([p], field(p, ^dim)) end)
+
+    scroll_depth_q =
+      subquery(max_per_visitor_q)
+      |> select([p], %{
+        scroll_depth: fragment("toUInt8(round(ifNotFinite(avg(?), 0)))", p.max_scroll_depth)
+      })
+      |> select_merge(^dim_select)
+      |> group_by(^dim_group_by)
+
+    join_on_dim_condition =
+      dim_shortnames
+      |> Enum.map(fn dim -> dynamic([_e, ..., s], selected_as(^dim) == field(s, ^dim)) end)
+      |> Enum.reduce(fn condition, acc -> dynamic([], ^acc and ^condition) end)
+
+    q
+    |> join(:left, [e], s in subquery(scroll_depth_q), on: ^join_on_dim_condition)
+    |> select_merge_as([_e, ..., s], %{scroll_depth: fragment("any(?)", s.scroll_depth)})
   end
 
   # `total_visitors_subquery` returns a subquery which selects `total_visitors` -
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -399,7 +399,7 @@ defmodule PlausibleWeb.Api.StatsController do
     metrics =
       cond do
         page_filter? && query.include_imported -> metrics
-        page_filter? -> metrics ++ [:bounce_rate, :time_on_page]
+        page_filter? -> metrics ++ [:bounce_rate, :scroll_depth, :time_on_page]
         true -> metrics ++ [:views_per_visit, :bounce_rate, :visit_duration]
       end
 
@@ -418,7 +418,8 @@ defmodule PlausibleWeb.Api.StatsController do
             nil -> 0
             value -> value
           end
-        )
+        ),
+        top_stats_entry(current_results, "Scroll depth", :scroll_depth, graphable?: true)
       ]
       |> Enum.filter(& &1)
 
@@ -1522,11 +1523,20 @@ defmodule PlausibleWeb.Api.StatsController do
       end
 
     requires_goal_filter? = metric in [:conversion_rate, :events]
+    has_goal_filter? = Filters.filtering_on_dimension?(query, "event:goal")
 
-    if requires_goal_filter? and !Filters.filtering_on_dimension?(query, "event:goal") do
-      {:error, "Metric `#{metric}` can only be queried with a goal filter"}
-    else
-      {:ok, metric}
+    requires_page_filter? = metric == :scroll_depth
+    has_page_filter? = Filters.filtering_on_dimension?(query, "event:page")
+
+    cond do
+      requires_goal_filter? and not has_goal_filter? ->
+        {:error, "Metric `#{metric}` can only be queried with a goal filter"}
+
+      requires_page_filter? and not has_page_filter? ->
+        {:error, "Metric `#{metric}` can only be queried with a page filter"}
+
+      true ->
+        {:ok, metric}
     end
   end
 
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -814,9 +814,16 @@ defmodule PlausibleWeb.Api.StatsController do
     query = Query.from(site, params, debug_metadata(conn))
 
     extra_metrics =
-      if params["detailed"],
-        do: [:pageviews, :bounce_rate, :time_on_page],
-        else: []
+      cond do
+        params["detailed"] && !query.include_imported ->
+          [:pageviews, :bounce_rate, :time_on_page, :scroll_depth]
+
+        params["detailed"] ->
+          [:pageviews, :bounce_rate, :time_on_page]
+
+        true ->
+          []
+      end
 
     metrics = breakdown_metrics(query, extra_metrics)
     pagination = parse_pagination(params)
--- a/priv/ingest_repo/migrations/20241028142653_add_scroll_depth_to_events.exs
+++ b/priv/ingest_repo/migrations/20241028142653_add_scroll_depth_to_events.exs
@@ -1,0 +1,9 @@
+defmodule Plausible.IngestRepo.Migrations.AddScrollDepthToEvents do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
+      add :scroll_depth, :UInt8
+    end
+  end
+end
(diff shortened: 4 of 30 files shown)


priv/ingest_repo/migrations/20241028142653_add_scroll_depth_to_events.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.IngestRepo.Migrations.AddScrollDepthToEvents do
+  use Ecto.Migration
+
+  def change do
+    alter table(:events_v2) do
+      add :scroll_depth, :UInt8
+    end
```
RobertJoonas, line 7: Should we also consider a db constraint like `scroll_depth = 0 UNLESS name = 'pageleave'`?

priv/json-schemas/query-api-schema.json
```diff
@@ -269,6 +269,10 @@
         {
           "const": "average_revenue",
           "$comment": "only :internal"
+        },
+        {
+          "const": "scroll_depth",
+          "markdownDescription": "The percentage of page content (y-axis) that was visible to an average visitor. Requires: an `event:page` filter or dimension"
```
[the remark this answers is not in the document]
macobo: There's a `/api/v2/query-internal-test` endpoint (see test/plausible_web/controllers/api/external_stats_controller/query_comparisons_test.exs for examples) which allows querying the internal schema.

Let's use that over updating the endpoint logic.

priv/ingest_repo/migrations/20241028142653_add_scroll_depth_to_events.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.IngestRepo.Migrations.AddScrollDepthToEvents do
```
macobo, line 1: Let's split this PR:
1. Tracker changes
2. Migration
3. (Potentially) Ingestion changes.
4. API and dashboard changes

Reasoning: 1-2 should be relatively safe to merge but 3 and 4 might cause other issues and is hard to review with other changes.

priv/ingest_repo/migrations/20241028142653_add_scroll_depth_to_events.exs
macobo: No - slows down ingestion and worst case drop batches of data if the constraint fails as we use batch ingestion.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -399,7 +399,7 @@ defmodule PlausibleWeb.Api.StatsController do
     metrics =
       cond do
         page_filter? && query.include_imported -> metrics
-        page_filter? -> metrics ++ [:bounce_rate, :time_on_page]
+        page_filter? -> metrics ++ [:bounce_rate, :scroll_depth, :time_on_page]
```
macobo, line 402: We should only add scroll-depth conditional on a feature flag.

Reasoning: Avoids unintentionally breaking the dashboard if there's an edge case where the query breaks and gives a quick tool to revert the change if needed.

test/plausible/ingestion/event_test.exs
```diff
@@ -334,6 +334,40 @@ defmodule Plausible.Ingestion.EventTest do
     assert dropped.drop_reason == :no_session_for_pageleave
   end
 
+  test "saves scroll depth for a pageleave event" do
```
macobo, line 337: This is quite convoluted and doesn't give full confidence that this works end-to-end due to the mocking below. Let's add at least 2 tests to test/plausible_web/controllers/api/external_controller_test.exs: one for no param having event.scroll_depth 0 and other with param having it set as something else.

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -121,6 +122,71 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
 
+  def maybe_add_scroll_depth(q, site, query) do
+    cond do
+      :scroll_depth not in query.metrics -> q
+      query.dimensions == [] -> add_aggregate_scroll_depth(q, site, query)
+      true -> add_group_scroll_depth(q, site, query)
+    end
+  end
+
+  defp add_aggregate_scroll_depth(q, site, query) do
```
macobo, line 133: The naming is confusing here. Can we think of better terms than `aggregate_scroll_depth` vs `group_scroll_depth`? Can we put into words why these even need to be different which might lead us to the neccessary insight?

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -121,6 +122,71 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
+      |> group_by([e], e.user_id)
+
+    scroll_depth_q =
+      subquery(max_per_visitor_q)
+      |> select([p], fragment("toUInt8(round(ifNotFinite(avg(?), 0)))", p.max_scroll_depth))
+
+    select_merge_as(q, [e], %{scroll_depth: subquery(scroll_depth_q)})
+  end
+
+  defp add_group_scroll_depth(q, site, query) do
```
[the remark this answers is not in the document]
macobo: It's impossible to know what the best way to write this query is without having the column live in production first. As suggested above, let's deploy the migration first.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -1716,6 +1805,187 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
            ]
   end
 
+  test "breakdown by event:page with scroll_depth metric", %{conn: conn, site: site} do
+    t0 = ~N[2020-01-01 00:00:00]
+    [t1, t2, t3] = for i <- 1..3, do: NaiveDateTime.add(t0, i, :minute)
```
macobo, line 1810: Q: Why this way vs just hard-coding the dates below?

I personally dislike 'programming' the data in my test and prefer hard-coded data as it makes debugging tests easier since you don't need to jump back-and-forth in the code.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -1716,6 +1805,187 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
...
+      build(:pageview, user_id: 34, pathname: "/blog", timestamp: t2),
+      build(:pageleave, user_id: 34, pathname: "/blog", timestamp: t3, scroll_depth: 60),
+      build(:pageview, user_id: 56, pathname: "/blog", timestamp: t0),
+      build(:pageleave, user_id: 56, pathname: "/blog", timestamp: t1, scroll_depth: 100)
+    ])
+
+    conn =
+      post(conn, "/api/v2/query", %{
+        "site_id" => site.domain,
+        "metrics" => ["visitors", "pageviews", "scroll_depth"],
```
macobo, line 1830: Nit: Let's remove the other metrics as they're irrelevant here?

tracker/src/plausible.js
```diff
@@ -48,13 +48,46 @@
...
+
+  function getCurrentScrollDepthPx() {
+    var viewportHeight = window.innerHeight || docEl.clientHeight || 0
+    var scrollTop = window.scrollY || docEl.scrollTop || bodyEl.scrollTop || 0
+
+    return currentDocumentHeight <= viewportHeight ? currentDocumentHeight : scrollTop + viewportHeight
+  }
+
+  var currentDocumentHeight = getDocumentHeight()
+  var maxScrollDepthPx = getCurrentScrollDepthPx()
```
macobo, line 73: Saving anything and only calculating it at the top level is inherently tricky since it might mess with back-forward cache.

tracker/src/plausible.js
```diff
@@ -48,13 +48,46 @@
...
+  })
+
   function triggerPageLeave() {
     if (pageLeaveSending) {return}
     pageLeaveSending = true
     setTimeout(function () {pageLeaveSending = false}, 500)
 
     var payload = {
       n: 'pageleave',
+      sd: Math.round((maxScrollDepthPx / currentDocumentHeight) * 100),
```
macobo, line 90: Let's have tests for this as well.

macobo (commented):
Added a bunch of initial thoughts. We should likely pair when the opportunity arises.

assets/js/dashboard/stats/modals/breakdown-modal.tsx
```diff
@@ -135,7 +135,7 @@ export default function BreakdownModal<TListItem extends { name: string }>({
       {
         label: reportInfo.dimensionLabel,
         key: 'name',
-        width: 'w-48 md:w-80 flex items-center break-all',
+        width: 'w-48 md:w-full flex items-center break-all',
```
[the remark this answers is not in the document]
macobo: Is this a current bug? If so, it might be wise to push up a quick 1-line PR for this.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
RobertJoonas: The primary reason was that otherwise `mix format` pushes every `build(:pageview)` or `build(:pageleave)`onto multiple lines and the readability gets significantly worse.

tracker/src/plausible.js
RobertJoonas: This state is actually function-scoped, since the whole Plausible script is wrapped inside a function that is immediately called. So I don't think it interferes with bfcache in any way.

tracker/src/plausible.js
RobertJoonas: Absolutely!

assets/js/dashboard/stats/modals/breakdown-modal.tsx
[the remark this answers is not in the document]
RobertJoonas: Wouldn't call it a bug, but the dimension label column taking the full width would make the most sense IMO.

Currently you can see this in the live demo:

image

The page url could perfectly fit onto a single line here:

image


>  If so, it might be wise to push up a quick 1-line PR for this

Sure! https://github.com/plausible/analytics/pull/4809

tracker/src/plausible.js
```diff
@@ -48,13 +48,46 @@
   // flag prevents sending multiple pageleaves in those cases.
   var pageLeaveSending = false
 
+  var bodyEl = document.body
+  var docEl = document.documentElement
```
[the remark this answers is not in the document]
RobertJoonas: Good catch! I was indeed aiming for a smaller minified output here.

The real minified file size can be decreased by extracting these variables (from 2257B to 2186B). But the compressed sizes are actually even ~2B greater, and that's what matters more. So I agree - let's drop the vars. 👍

PS: Will do that in another PR.

tracker/src/plausible.js
```diff
@@ -48,13 +48,46 @@
...
+  }
+
+  function getCurrentScrollDepthPx() {
+    var viewportHeight = window.innerHeight || docEl.clientHeight || 0
+    var scrollTop = window.scrollY || docEl.scrollTop || bodyEl.scrollTop || 0
+
+    return currentDocumentHeight <= viewportHeight ? currentDocumentHeight : scrollTop + viewportHeight
+  }
+
+  var currentDocumentHeight = getDocumentHeight()
```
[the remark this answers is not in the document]
RobertJoonas: Getting the document height at the time of scroll is not an option, as we need to have a value even when scrolling does not happen at all.

My initial hunch was to read the document height only once - when the pageleave event is sent. But that doesn't work because on SPA navigation, pageleaves will use the height of the next page.

That's why we need to keep some kind of state representing the "current document height".

How about just adding an extra line after line 72:
```javascript
window.addEventListener("load", () => currentDocumentHeight = getDocumentHeight());
});
```

tracker/src/plausible.js
RobertJoonas: https://github.com/plausible/analytics/pull/4826/commits/8ae1a816eadcb3653634618c9db4d15783241186

tracker/src/plausible.js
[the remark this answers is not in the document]
RobertJoonas: https://github.com/plausible/analytics/pull/4826/commits/5d064898674da1e988ee4b1fa63e9b2be3e5fb76

tracker/src/plausible.js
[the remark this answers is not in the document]
RobertJoonas: > How about just adding an extra line after line 72

I've now  done this in https://github.com/plausible/analytics/pull/4826/commits/a5b22fac376c02f5a00c841f7ba1510b689f4da3. Also added a test.

> Shouldn't we take page height at time of scroll rather than when this happens to be initialized?

TBH, I don't think reevaluating `currentDocumentHeight` on `scroll` event makes sense. The only way the document height could change on scroll is if additional resources are being loaded (e.g. infinite scroll). but we're only using the `currentDocumentHeight` to calculate the percentage, which is pretty much useless in such cases. 

For example, 75% scroll depth could mean that the user scrolled 75% of the first page, but at the same time, it could be that they scrolled to the end of the document three times and loaded more content every time (i.e. actually consuming 3/4 full pages of content)

To me it sounds more useful to consider all users who consumed more than one page of content as 100% scroll depth. Of course, measuring scroll depth in pixels would be by far more useful in these cases. Providing that option to the users still needs to be figured out.

lib/plausible/stats/sql/special_metrics.ex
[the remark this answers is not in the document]
RobertJoonas: Roger that! https://github.com/plausible/analytics/pull/4827

test/plausible/ingestion/event_test.exs
RobertJoonas: Totally agreed - `external_controller_test.exs` is 100% the better place to test this, end to end. I have opened a [new PR](https://github.com/plausible/analytics/pull/4828) with extracted ingestion changes, where I've replaced this convoluted test with multiple tests in `external_controller_test.exs`.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: 👍 Done in https://github.com/plausible/analytics/pull/4832

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: Currently decided to still move it into the same function (in https://github.com/plausible/analytics/pull/4832), as there was a bug anyway with selecting the `scroll_depth` metric on its own. The "aggregate" one didn't work like that - it requires a join anyway.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
RobertJoonas: 👍 Done in https://github.com/plausible/analytics/pull/4832.

PS: Led me to an important discovery (ditto my other comment) - turned out that the "aggregate" function couldn't handle selecting only `metrics: ["scroll_depth"]`

RobertJoonas:
Thanks for the awesome review @macobo! This PR was a useful POC but I'm closing it now in favour of the four smaller PRs (as you suggested). I've left links to relevant commits under the discussions here. Let's continue those discussions in each of those PRs separately :)

PR #4797, opened 2024-11-11
Saved segments/create table

apata:
### Changes

Creates the saved segments table.
```
         List of data types
 Schema |     Name     | Description 
--------+--------------+-------------
 public | segment_type | 

                                           Table "public.segments"
    Column    |              Type              | Collation | Nullable |               Default                
--------------+--------------------------------+-----------+----------+--------------------------------------
 id           | bigint                         |           | not null | nextval('segments_id_seq'::regclass)
 name         | character varying(255)         |           | not null | 
 type         | segment_type                   |           | not null | 'personal'::segment_type
 segment_data | jsonb                          |           | not null | 
 site_id      | bigint                         |           | not null | 
 owner_id     | bigint                         |           |          | 
 inserted_at  | timestamp(0) without time zone |           | not null | 
 updated_at   | timestamp(0) without time zone |           | not null | 
Indexes:
    "segments_pkey" PRIMARY KEY, btree (id)
    "segments_site_id_index" btree (site_id)
Foreign-key constraints:
    "segments_owner_id_fkey" FOREIGN KEY (owner_id) REFERENCES users(id)
    "segments_site_id_fkey" FOREIGN KEY (site_id) REFERENCES sites(id) ON DELETE CASCADE
```

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20241107120000_create_segments.exs
+++ b/priv/repo/migrations/20241107120000_create_segments.exs
@@ -2,14 +2,16 @@ defmodule Plausible.Repo.Migrations.CreateSegments do
   use Ecto.Migration
 
   def change do
+    execute(
+      "CREATE TYPE segment_type AS ENUM ('personal', 'site')",
+      "DROP TYPE segment_type"
+    )
+
     create table(:segments) do
       add :name, :string, null: false
-      add :personal, :boolean, default: true, null: false
+      add :type, :segment_type, null: false, default: "personal"
       add :segment_data, :map, null: false
       add :site_id, references(:sites, on_delete: :delete_all), null: false
-
-      # owner_id is null (aka segment is dangling) when the original owner is deassociated from the site
-      # the segment is dangling until another user edits it: the editor becomes the new owner
       add :owner_id, references(:users, on_delete: :nothing), null: true
 
       timestamps()
--- a/priv/repo/migrations/20241107120000_create_segments.exs
+++ b/priv/repo/migrations/20241107120000_create_segments.exs
@@ -1,0 +1,20 @@
+defmodule Plausible.Repo.Migrations.CreateSegments do
+  use Ecto.Migration
+
+  def change do
+    create table(:segments) do
+      add :name, :string, null: false
+      add :personal, :boolean, default: true, null: false
+      add :segment_data, :map, null: false
+      add :site_id, references(:sites, on_delete: :delete_all), null: false
+      # owner_id is null (aka segment is dangling) when the original owner is deassociated from the site
+      # the segment is dangling until another user edits it: the editor becomes the new owner
+      add :owner_id, references(:users, on_delete: :nothing), null: true
+
+      timestamps()
+    end
+
+    create index(:segments, [:segment_data], using: :gin)
+    create index(:segments, [:site_id])
+  end
+end
--- a/priv/repo/migrations/20241107120000_create_segments.exs
+++ b/priv/repo/migrations/20241107120000_create_segments.exs
@@ -7,6 +7,7 @@ defmodule Plausible.Repo.Migrations.CreateSegments do
       add :personal, :boolean, default: true, null: false
       add :segment_data, :map, null: false
       add :site_id, references(:sites, on_delete: :delete_all), null: false
+
       # owner_id is null (aka segment is dangling) when the original owner is deassociated from the site
       # the segment is dangling until another user edits it: the editor becomes the new owner
       add :owner_id, references(:users, on_delete: :nothing), null: true
--- a/priv/repo/migrations/20241107120000_create_segments.exs
+++ b/priv/repo/migrations/20241107120000_create_segments.exs
@@ -14,7 +14,6 @@ defmodule Plausible.Repo.Migrations.CreateSegments do
       timestamps()
     end
 
-    create index(:segments, [:segment_data], using: :gin)
     create index(:segments, [:site_id])
   end
 end


priv/repo/migrations/20241107120000_create_segments.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Plausible.Repo.Migrations.CreateSegments do
+  use Ecto.Migration
+
+  def change do
+    create table(:segments) do
+      add :name, :string, null: false
+      add :personal, :boolean, default: true, null: false
+      add :segment_data, :map, null: false
```
macobo, line 8: Q: We had a discussion earlier about segment_data vs just storing filters. Where did we end up on that discussion?

priv/repo/migrations/20241107120000_create_segments.exs
```diff
@@ -0,0 +1,20 @@
...
+      add :site_id, references(:sites, on_delete: :delete_all), null: false
+
+      # owner_id is null (aka segment is dangling) when the original owner is deassociated from the site
+      # the segment is dangling until another user edits it: the editor becomes the new owner
+      add :owner_id, references(:users, on_delete: :nothing), null: true
+
+      timestamps()
+    end
+
+    create index(:segments, [:site_id])
```
macobo, line 18: Probably also want to index on owner_id?

priv/repo/migrations/20241107120000_create_segments.exs
apata: I tried splitting segment_data into separate labels and filters columns, but ran into issues with registering the schema with Ecto: [for array type, it expects the inner type](https://hexdocs.pm/ecto/Ecto.Schema.html#module-primitive-types), which for filters is a tuple, where the third member, clauses, is another array.

priv/repo/migrations/20241107120000_create_segments.exs
apata: I considered that but I decided against it at this point, not to prematurely optimize.

priv/repo/migrations/20241107120000_create_segments.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Plausible.Repo.Migrations.CreateSegments do
+  use Ecto.Migration
+
+  def change do
+    create table(:segments) do
+      add :name, :string, null: false
+      add :personal, :boolean, default: true, null: false
```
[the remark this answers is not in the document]
apata: For the first iteration, I think we'll have 2 cases. If we end up needing more cases, we can always migrate to a "type" column down the road.

priv/repo/migrations/20241107120000_create_segments.exs
apata: Here's how the schema looks like right now: https://github.com/plausible/analytics/compare/saved-segments/add-schema?expand=1

macobo: The problem seems to be that array maps to the native string type. 

I'm not a fan of stuffing multiple things into a single structure like this. I propose:

```
add :filters, :map, null: false
add :labels, :map, null: false
```

If it needs to be a json object, add a `value` key at the top level for both?

priv/repo/migrations/20241107120000_create_segments.exs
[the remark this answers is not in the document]
macobo: These migrations are generally a pain and so is dealing with booleans in code. I suggest switching to an enum: https://hexdocs.pm/ecto/Ecto.Enum.html - even in the 2-option case this will IMO result in cleaner code.

priv/repo/migrations/20241107120000_create_segments.exs
apata: > The problem seems to be that array maps to the native string type.

Can you explain this further? 

> If it needs to be a json object, add a value key at the top level for both?

This did come to my mind also, but then accessing the value in application will still be verbose, and we'll probably need custom serialization (unless we want to deliver `{filters: {value: [[...], [...]]}, labels: {value: {...}} }`. 

Since labels without filters don't really stand on their own, I'm not seeing the benefits of having a separate column for them.

macobo: What are examples of valid `segment_data` maps? Especially curious on the labels side.

> we'll probably need custom serialization

Ecto has good tools for that I believe for doing bidirectional mapping I believe. Not confident though.

> I'm not seeing the benefits of having a separate column for them.

My concern is that oblique blobs in the schema are a nighmare to maintain and build around. You need to start going in and checking different places to figure out what the `real` schema is.

My thinking is that `filters` column bypasses that since we can say this is the (static) filter structure we use elsewhere, but I'm not confident in what problem the `labels` is solving and how it fits in here.

----

Worth jumping into a call? Ping me on basecamp if so.

apata: With your suggestion, wouldn't we still need to hunt down the real schema? 

I need to store the labels used for the filters in the segment, because they're needed for showing the filter pills correctly. The FE doesn't have tools for getting the labels from scratch for an arbitrary filter set. We'll probably want something like that in the future, but it's a sidetrack that I hoped to avoid for now.

macobo: Examples you listed in basecamp: https://github.com/plausible/analytics/compare/saved-segments/add-schema?expand=1#diff-847456b6d98fe97df3b85ff4c16a2c089cffdf168a061b607f2d9fb548e43a27R69-R80

Okay. Given that labels is something we can safely get rid of with a bit of work, let's keep this as-is.

priv/repo/migrations/20241107120000_create_segments.exs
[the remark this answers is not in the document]
apata: Can you give an example of a potential issue with dealing booleans in code? As for cleaner code, I'm not so sure about that. Atoms are convenient in the backend, but using enums in the FE can be quite verbose. 

```ts
import SegmentType from '../../segments.ts' 
// ...
if (segment.type === SegmentType.personal) {
// ...
}
```

vs

```ts
if (segment.personal) {
// ...
}
```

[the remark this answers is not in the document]
macobo: I prefer the first option. Elixir sample as well:

```
handle(segment.personal, segment.segment_data)

def handle(true, segment_data), do: ...
def handle(false, segment_data), do: ...
```

vs
```
handle(segment.type, segment.segment_data)

def handle(:personal, segment_data), do: ...
def handle(:team, segment_data), do: ...
```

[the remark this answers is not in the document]
apata: Hmm, it's not something that I'm currently branching functions on (see here: https://github.com/plausible/analytics/tree/saved-segments-spec/). The branching is happening in Ecto queries. 

As for the example you gave, I wonder whether a better spec would actually be not to split the segment to pieces in function arguments. That seems to have been a successful strategy with Plausible.Query.  
```ex
  @spec handle(Plausible.Segment.t()) :: :ok

  def handle(%Plausible.Segment{personal: true} = segment) do end
  def handle(%Plausible.Segment{personal: false} = segment) do end

```

macobo (approved):

PR #4798, opened 2024-11-11
Channels: Migration to add materialized column, backfill code

macobo:
This change adds `acqusition_channel` columns to events_v2 and sessions_v2 tables. These columns are materialized - we don't ingest into them directly. Instead they're calculated based on other columns.

The data migration changes now allow to also backfill the column.

Tested the ability to change definitions by changing the function definitions and re-running the migration with backfill. Confirmed that the underlying data changed as expected.

Before merging this:
- Merge https://github.com/plausible/analytics/pull/4789

After this:
- Run migration on cloud
- Backfill `click_id_param` on cloud
- Kick off a backfill
- Spot-check the results and impact on ingestion, etc
- PR a change to stop ingesting into the old column, to read from new one

### DIFF
--- a/lib/plausible/data_migration/acquisition_channel.ex
+++ b/lib/plausible/data_migration/acquisition_channel.ex
@@ -2,6 +2,15 @@ defmodule Plausible.DataMigration.AcquisitionChannel do
   @moduledoc """
   Creates dictionaries and functions to calculate acquisition channel in ClickHouse
 
+  Creates `acquisition_channel` columns in `events_v2` and `sessions_v2` tables.
+  Run via `Plausible.DataMigration.AcquisitionChannel.run(options)`
+
+  Options:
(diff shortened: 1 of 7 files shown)


lib/plausible/data_migration/acquisition_channel.ex
```diff
@@ -2,25 +2,86 @@ defmodule Plausible.DataMigration.AcquisitionChannel do
...
+
+    cond do
+      Keyword.get(opts, :add_column) ->
+        alter_data_tables(
+          "acquisition_channel_add_materialized_column",
+          on_cluster_statement,
+          opts
+        )
+
+      Keyword.get(opts, :update_column, true) ->
```
ukutaht, line 44: Why/when would we need to update the column like this?

ukutaht (approved):

PR #4800, opened 2024-11-11
Channels: Read from new column, don't write

macobo:
### Changes

Follow-up to https://github.com/plausible/analytics/pull/4798

Once this change is in, we'll read from the new `acquisition_channel` column.


ukutaht (approved):
Looks good.

PR #4801, opened 2024-11-11
Channels: Fix ON CLUSTER behavior

macobo:
CREATE TABLE AS SELECT syntax did not work on cluster.

Instead, let's do a normal insert. For safety and to avoid timing issues, ensure that TRUNCATE/INSERT waits for data to be in sync in all the replicas.


ukutaht (approved):
ugh, why doesn't clickhouse just have a simple `Map` datatype. This is way too much work to get a static lookup table going :/ Anyways, looks good.

macobo:
Agreed. I'd take a simple way to create a dictionary from static data. :(

PR #4803, opened 2024-11-11
Solve noisy warnings about `a negative range was inferred for Date.range/2`

macobo:
`query.now` is in utc but the date range was in the querys timezone

This is visible in the pattern on the graph, with peaks around UTC midnight:

image

Ref: https://3.basecamp.com/5308029/buckets/39750953/card_tables/cards/7999429743

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -23,7 +23,7 @@ defmodule Plausible.Stats.Query do
             pagination: nil
 
   require OpenTelemetry.Tracer, as: Tracer
-  alias Plausible.Stats.{Filters, Imported, Legacy}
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/query.ex
```diff
@@ -53,9 +53,11 @@ defmodule Plausible.Stats.Query do
     date_range = Plausible.Stats.DateTimeRange.to_date_range(query.utc_time_range, query.timezone)
 
     if Keyword.get(options, :trim_trailing) do
+      today = query.now |> DateTime.shift_zone!(query.timezone) |> DateTime.to_date()
```
apata, line 56: Good catch! Though it would be nice to see a test case covering this.

PR #4805, opened 2024-11-12
Register segments schema

apata:
### Changes

Registers segments schema with Ecto. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segment.ex
+++ b/lib/plausible/segment.ex
@@ -1,0 +1,95 @@
+defmodule Plausible.Segment do
+  @moduledoc """
+  Schema for segments. Segments are saved filter combinations.
+  """
+  use Plausible
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @segment_types [:personal, :site]
+
+  @type t() :: %__MODULE__{}
+
+  @derive {Jason.Encoder,
+           only: [
+             :id,
+             :name,
+             :type,
+             :segment_data,
+             :owner_id,
+             :inserted_at,
+             :updated_at
+           ]}
+
+  schema "segments" do
+    field :name, :string
+    field :type, Ecto.Enum, values: @segment_types
+    field :segment_data, :map
+
+    # owner ID can be null (aka segment is dangling) when the original owner is deassociated from the site
+    # the segment is dangling until another user edits it: the editor becomes the new owner
+    belongs_to :owner, Plausible.Auth.User, foreign_key: :owner_id
+    belongs_to :site, Plausible.Site
+
+    timestamps()
+  end
+
+  def changeset(segment, attrs) do
+    segment
+    |> cast(attrs, [
+      :name,
+      :segment_data,
+      :site_id,
+      :type,
+      :owner_id
+    ])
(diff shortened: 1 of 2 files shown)


macobo (approved):
This looks good to me, though I would love another set of :eyes: for elixir expertiese. cc @aerosol @RobertJoonas @zoldar

lib/plausible/segment.ex
```diff
@@ -0,0 +1,95 @@
...
+  @derive {Jason.Encoder,
+           only: [
+             :id,
+             :name,
+             :type,
+             :segment_data,
+             :owner_id,
+             :inserted_at,
+             :updated_at
+           ]}
```
aerosol, line 22: what's the reason for keeping this?

lib/plausible/segment.ex
```diff
@@ -0,0 +1,95 @@
...
+
+  def changeset(segment, attrs) do
+    segment
+    |> cast(attrs, [
+      :name,
+      :segment_data,
+      :site_id,
+      :type,
+      :owner_id
+    ])
```
aerosol, line 45: use put_assoc rather than literate foreign keys casting?

lib/plausible/segment.ex
```diff
@@ -0,0 +1,95 @@
...
+           ]}
+
+  schema "segments" do
+    field :name, :string
+    field :type, Ecto.Enum, values: @segment_types
+    field :segment_data, :map
+
+    # owner ID can be null (aka segment is dangling) when the original owner is deassociated from the site
+    # the segment is dangling until another user edits it: the editor becomes the new owner
+    belongs_to :owner, Plausible.Auth.User, foreign_key: :owner_id
```
aerosol, line 31: This shouldn't overlap with teams then.

aerosol (approved):

lib/plausible/segment.ex
apata: Ah, I actually didn't notice that this is the first changeset in the repo that is using the foreign_key_constraint rather than put_assoc. 

In the spike that I did for segments, I didn't end up needing to actually load these associations, but maybe I will at one point, maybe to check if site (_is within a team that_) has a Business subscription. 

When I do, I can update the foreign_key_constraint()'s to put_assoc() without much impact, right? Or do you suggest I start with put_assoc right away?

lib/plausible/segment.ex
apata: Not at the schemas level, but I've yet to settle on the way to implement nulling when the user is deassociated from one or more sites.

lib/plausible/segment.ex
aerosol: Right OK, it seemed unused in the context of this PR.

lib/plausible/segment.ex
aerosol: Yeah, I think you'll need to preload anyway to ensure no random IDs are injected, depending on how your API works. Your call, just wanted to flag this.

aerosol (approved):

PR #4806, opened 2024-11-12
Make "Details" search case insensitive

aerosol:
This is done by introducing `icontains` and `icontains_not` filter operator variants, so that no unexpected behaviour is accidentally triggered.

Dashboard searches for:

  - top/entry/exit pages
  - cities
  - conversions
  - sources (including UTM)
  - props
  - browsers/OS

use `icontains` explicitly.

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated: https://github.com/plausible/docs/pull/562

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Dashboard shows comparisons for all reports
+- Support for `icontains` and `icontains_not` operators in Stats API filters
 
 ### Removed
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Dashboard shows comparisons for all reports
-- Support for `icontains` and `icontains_not` operators in Stats API filters
+- Support for `icontains` and `icontains_not` operators in Stats API filters for case-insensitive searches.
+  Those are enabled by default in "Details" search inputs on the dashboard.
 
 ### Removed
 ### Changed
--- a/assets/js/dashboard/stats/modals/conversions.js
+++ b/assets/js/dashboard/stats/modals/conversions.js
@@ -28,7 +28,7 @@ function ConversionsModal() {
   }, [reportInfo.dimension])
 
   const addSearchFilter = useCallback((query, searchString) => {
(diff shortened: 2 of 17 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Dashboard shows comparisons for all reports
+- Support for `icontains` and `icontains_not` operators in Stats API filters
```
apata, line 8: **nitpick, non-blocking:** This could use a bit more explanation unless icontains is an industry standard term that I'm just ignorant of. Maybe add another line that in some modals, search is now case insensitive? 

On that note actually, why not make icontains search default for all dashboard breakdowns? Maybe it's just me, but I was manually testing and was surprised that it's not enabled for browsers for example.

aerosol: > Maybe add another line that in some modals, search is now case insensitive?

12eef49e4

> Maybe it's just me, but I was manually testing and was surprised that it's not enabled for browsers for example.

https://github.com/plausible/analytics/pull/4806/commits/7c9b828744cc72f5ccb057ecc305d17723e98f9a - overlooked that part

apata (approved):
This looks very good to me.

aerosol:
> The one thing that comes to mind is to list the new operators

There is a docs PR cross-referenced here

apata:
> > The one thing that comes to mind is to list the new operators
> 
> There is a docs PR cross-referenced here

Yep, good work! Realized as soon as I scrolled up :)

ukutaht (commented):
In the [Filters RFC](https://github.com/plausible/knowledge_base/blob/main/src/planning/rfcs/0005-filters_v2.md#examples) we designed for a case sensitivity modifier (see example nr 3) for filters.

I still think that's the preferred solution because it will prevent the explosion of possible filter operators. If we have case-insensitive `contains` and `contains_not` then it would also make sense to have a case-insensitive `is`, `is_not`, `matches`, `matches_not`, and in the future stuff like `starts_with`, `starts_with_not`, `ends_with`, `ends_with_not`, etc.

aerosol:
@ukutaht ah OK, this is much heavier lifting then, potentially involving more URL/linking quirks. I'll drop this for now.

PR #4810, opened 2024-11-12
WIP: flatten ?filters=(...) to ?f=...&f=...

apata:
### Changes

Maybe solves issue with Plausible dashboard links not being rendered in Twitter DMs and other places? 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/query.ts
+++ b/assets/js/dashboard/query.ts
@@ -154,6 +154,25 @@ export function filtersBackwardsCompatibilityRedirect(
   }
 }
 
+// Called once when dashboard is loaded load. Checks whether old filter style is used and if so,
+// updates the filters and updates location
+export function filtersBackwardsCompatibilityRedirect2(
+  windowLocation: Location,
+  windowHistory: History
+) {
+  const searchRecord = parseSearch(windowLocation.search)
+  const getValue = (k: string) => searchRecord[k]
+
+  if (getValue('filters')) {
+    windowHistory.pushState(
+      {},
+      '',
+      `${windowLocation.pathname}${stringifySearch(searchRecord)}`
+    )
+  }
+}
+
+
 // Returns a boolean indicating whether the given query includes a
 // non-empty goal filterset containing a single, or multiple revenue
 // goals with the same currency. Used to decide whether to render
--- a/assets/js/dashboard.tsx
+++ b/assets/js/dashboard.tsx
@@ -9,7 +9,7 @@ import { createAppRouter } from './dashboard/router'
 import ErrorBoundary from './dashboard/error/error-boundary'
 import * as api from './dashboard/api'
 import * as timer from './dashboard/util/realtime-update-timer'
-import { filtersBackwardsCompatibilityRedirect } from './dashboard/query'
+import { filtersBackwardsCompatibilityRedirect, filtersBackwardsCompatibilityRedirect2 } from './dashboard/query'
 import SiteContextProvider, {
   parseSiteFromDataset
 } from './dashboard/site-context'
@@ -42,6 +42,13 @@ if (container && container.dataset) {
     } catch (e) {
       console.error('Error redirecting in a backwards compatible way', e)
     }
+
+    try {
+      filtersBackwardsCompatibilityRedirect2(window.location, window.history)
+    } catch (e) {
+      console.error('Error redirecting in a backwards compatible way', e)
+    }
+
 
     const router = createAppRouter(site)
 
--- a/assets/js/dashboard/util/url.ts
+++ b/assets/js/dashboard/util/url.ts
@@ -95,11 +95,23 @@ export function isSearchEntryDefined(
 export function stringifySearch(
   searchRecord: Record<string, unknown>
 ): '' | string {
-  const definedSearchEntries = Object.entries(searchRecord || {})
+  const { filters, ...rest } = searchRecord || {}
+  const definedSearchEntries = Object.entries(rest)
     .map(stringifySearchEntry)
     .filter(isSearchEntryDefined)
 
   const encodedSearchEntries = definedSearchEntries.map(encodeSearchParamEntry)
+
+  if (Array.isArray(filters) && filters.length) {
+    return `?${filters
+      .map((f) => {
+        const [operator, dimension, ...clauses] = f
+        const serializedFilter = [operator, dimension, ...clauses.map(encodeURIComponentPermissive)].join(',')
+        return ['f', '=', serializedFilter].join('')
+      })
+      .concat(encodedSearchEntries)
+      .join('&')}`
+  }
 
   return encodedSearchEntries.length ? `?${encodedSearchEntries.join('&')}` : ''
 }
@@ -148,6 +160,21 @@ export function parseSearchFragment(
 export function parseSearch(searchString: string): Record<string, unknown> {
   const urlSearchParams = new URLSearchParams(searchString)
   const searchRecord: Record<string, unknown> = {}
-  urlSearchParams.forEach((v, k) => (searchRecord[k] = parseSearchFragment(v)))
+  const filters: unknown[] = []
+  urlSearchParams.forEach((v, k) => {
+    if (k !== 'f') {
+      searchRecord[k] = parseSearchFragment(v)
+      return
+    }
+    const parsedFilter = v.split(',').map(parseSearchFragment)
+    const [operator, dimension, ...clauses] = parsedFilter as Array<unknown>
+    filters.push([operator, dimension, clauses])
+  })
+  if (filters.length) {
+    return {
+      ...searchRecord,
+      filters
+    }
+  }
   return searchRecord
 }


ruslandoga:
👋 

Sorry for a drive-by comment, but if it's possible to set the attribute key to `f[]=` then it would get automatically parsed into a list in Phoenix. Otherwise, I think a custom parser for query string would need to be implemented.

apata:
> 👋
> 
> Sorry for a drive-by comment on a draft PR, but if you set the attribute key to `f[]=` then it would get automatically decoded into a list in Plug / Phoenix. Otherwise a custom decoder for query strings would need to be implemented.
> 
> More info: https://hexdocs.pm/plug/Plug.Conn.Query.html

Thanks for bringing it up @ruslandoga ! At the moment, this format must be recognizable only to the FE and ideally humans, though. I think of it as URL-stored dashboard state. 

Currently, the BE receives applied dashboard filters in URL-encoded stringified JSON, and there's an extra transform taking place https://github.com/plausible/analytics/blob/master/assets/js/dashboard/util/filters.js#L231 before that, which we're not planning to get rid of (in the interest of readable URLs).

ukutaht:
The idea with the new filters format was to also support nesting in case we need it in the future. Stuff like:

```
[
  ["or", [
    ["and", [
      ["is", "visit:source",  "Google", {case_sensitive: false}],
      ["is", "visit:utm_medium", "is", "cpc"]
    ]],
    ["and", [
      ["is", "visit:source", "Facebook", {case_sensitive: false}],
      ["is", "visit:utm_medium", "paid_social"]
    ]]
  ]]
]
```

This format seems to make that a bit harder. Did you consider that as well @apata ?

apata:
Good question, @ukutaht! I expect there to be a bunch of ways to do this because we know the key words: and, or, not, all operators and all operator options.

```json
[
  [
    "or",
    [
      [
        "and",
        [
          [
            "is",
            "visit:source",
            ["Google", "Bing"],
            { "case_sensitive": false }
          ],
          ["is", "visit:utm_medium", "is", ["cpc"]]
        ]
      ],
      [
        "and",
        [
          ["is", "visit:source", ["Facebook"], { "case_sensitive": false }],
          ["is", "visit:utm_medium", ["paid_social"]]
        ]
      ]
    ]
  ],
  ["not", ["is", "visit:country", ["Estonia"]]],
  ["contains", "event:goal", ["trial"], { "case_sensitive": false }]
]
```

For this edgy filter tree, an URL in the new style might be

```
?f=or/2&f=and/2&f=is/i,source,Google,Bing&f=is,utm_medium,cpc&f=and/2&f=is/i,source,Facebook&f=is,utm_medium,paid_social&f=not/1&f=is,country,Estonia&f=contains/i,goal,trial
```

This would be reconstructable thanks to fixed order of f= in the URL and known group lengths.

```ts
// url.split("&").map(i => i.split("=")).filter(i => i[0] === "f").map(i => i[1]) -->
[
  "or/2",
  "and/2",
  "is/i,source,Google,Bing",
  "is,utm_medium,cpc",
  "and/2",
  "is/i,source,Facebook",
  "is,utm_medium,paid_social",
  "not/1",
  "is,country,Estonia",
  "contains/i,goal,trial",
]
// reconstruct recursively
```

ukutaht:
Would be good to get another review/opinion - @aerosol @zoldar @macobo thoughts on this?

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -0,0 +1,179 @@
...
+    : []
+
+  const serializedLabels = Object.entries(labels ?? {}).map(
+    (entry) => `${LABEL_URL_PARAM_NAME}=${serializeLabelsEntry(entry)}`
+  )
+
+  return `?${serializedFilters.concat(serializedLabels).concat(definedSearchEntries).join('&')}`
+}
+
+export function parseSearch(searchString: string): Record<string, unknown> {
```
macobo, line 34: Bug: `parseSearch('')` currently returns `{"": 'undefined'}`. This results in a bunch of `&=undefined` artefacts in urls whenever navigating around.

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Serializes and flattens filters array item.
+ * Examples:
+ * ["is", "entry_page", ["/blog", "/news"]] -> "is,entry_page,/blog,/news"
+ */
+export function serializeFilter([operator, dimension, clauses]: Filter) {
+  const serializedFilter = [
+    encodeURIComponentPermissive(operator, ':/'),
+    encodeURIComponentPermissive(dimension, ':/'),
+    ...clauses.map((clause) =>
```
macobo, line 104: We recently added support for case insensitivity in APIv2 via modifiers (https://github.com/plausible/docs/pull/567). 
While we haven't exposed this in the frontend, it's something we should do once we figure out the UX. I'm not sure whether this function is extendable to allow for that structure though.

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/**
+ * Serializes and flattens filters array item.
+ * Examples:
+ * ["is", "entry_page", ["/blog", "/news"]] -> "is,entry_page,/blog,/news"
+ */
+export function serializeFilter([operator, dimension, clauses]: Filter) {
+  const serializedFilter = [
+    encodeURIComponentPermissive(operator, ':/'),
```
macobo, line 102: Let's extract a constant for `':/'`, e.g.

```
const NOT_ESCAPED = ':/'
```

assets/js/dashboard/util/url-search-params.ts
apata: I believe it is: https://github.com/plausible/analytics/pull/4810#issuecomment-2572633688

apata: > Note that in general modifiers are not always flags and may contain arbitrary data structures within.

The filter modifiers and structures we support is up to us, so I'm not clear on what you mean by arbitrary.

That said, I can imagine something like the following working:

```ts
-> `is\non_flag_modifier_name:${encodeURIComponent(JSON.stringify(complex_structure))},<dimension>,<clause1>,<clause2>`
```

That's stretching the limits of readability though.

ukutaht: > I'm not clear on what you mean by arbitrary.

@macobo and I were just bouncing around some ideas about how to extend filtering capabilities in the future which would utilize more modifiers. Note that these new modifiers will not necessarily be built in the short term but having the ability to extend filters with this kind of stuff in the future would be really useful.

Arbitrary date range example: 
Screenshot 2025-01-09 at 18 24 13

More flags:
Screenshot 2025-01-09 at 18 24 22

See this discussion: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/7793417773

As we were discussing it did cross my mind that this may not work well with the proposed URL structure here.,,

apata: Thanks for the extra examples! What makes these modifiers arbitrary?

We know what keys to look for, and even what the possible values can be.

As I see it, filter clauses are the only arbitrary value currently: their values are unpredictable to us, and may overlap with any of our option names or option configurations. 

That's why in this solution, they are expanded to be the last members of the comma separated array: `f=<known word at known position>/<optional known words and values>,<known word at known position>,...unpredictable words...`

However, from my side, I'm not married to this solution. There's various others that are totally unexplored at the moment, like enriching jsonurl etc. It's just that "working well" will be hard to solve for. Short, easy to implement, linkable, readable are some of the things that I optimized for. I personally think expandable is also one of the qualities, but I see that doubt remains.

macobo: I think something analogous to what you proposed in https://github.com/plausible/analytics/pull/4810#discussion_r1908818932 should work for the original question (how to handle more complex data structures in filter modifiers and should unblock us for now.

macobo (commented):
Behaviorally this is great and I appreciate the extensive tests!

I didn't do extensive testing on the backwards redirect. 

I also had some minor suggestions around code structure to avoid the backwards redirects causing issues in the future. Let me know if you'd prefer me to push to the branch to implement the suggestions.

assets/js/dashboard/query.ts
```diff
@@ -100,33 +104,46 @@ export function postProcessFilters(filters: Array<Filter>): Array<Filter> {
...
 
-// Called once when dashboard is loaded load. Checks whether old filter style is used and if so,
-// updates the filters and updates location
-export function filtersBackwardsCompatibilityRedirect(
-  windowLocation: Location,
-  windowHistory: History
-) {
-  const searchRecord = parseSearch(windowLocation.search)
-  const getValue = (k: string) => searchRecord[k]
+export function maybeGetRedirectTargetFromLegacySearchParams(
```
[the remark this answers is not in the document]
apata: Good idea. Added with slight modifications.

assets/js/dashboard/query.ts
```diff
@@ -100,33 +104,46 @@ export function postProcessFilters(filters: Array<Filter>): Array<Filter> {
...
+  windowLocation: Location
+): null | string {
```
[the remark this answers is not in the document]
apata: I agree that what you propose is better, but since this the first redirect wasn't properly tested, I opted against it originally: in review, any slight breaking changes could have slipped past.

assets/js/dashboard/util/legacy-jsonurl-url-search-params.ts
```diff
@@ -0,0 +1,80 @@
+/* @format */
```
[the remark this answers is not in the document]
apata: You're right that utils can get messy, but I think I'd prefer it solved in a separate refactor that tackles code organization only. 

There's relations between url, api, url search params, query,  filters that our current code organization isn't fully capturing. Just now I had to solve some circular import dependency issue.

macobo (approved):

PR #4814, opened 2024-11-12
Import acquisiton channel from GA4

ukutaht:
A relatively straightforward addition of channel import from GA4. Thank @zoldar @RobertJoonas for making the imports code easy to work with, all in all this was quite painless to implement :)

Migration: https://github.com/plausible/analytics/pull/4815

### Tests

### Changelog
Will add changelog when feature is ready to launch

### Documentation
Will add docs when feature is ready to launch

### Dark mode


test/support/factory.ex
```diff
@@ -135,21 +135,29 @@ defmodule Plausible.Factory do
...
-  def pageview_factory do
-    Map.put(event_factory(), :name, "pageview")
+  def pageview_factory(attrs) do
+    Map.put(event_factory(attrs), :name, "pageview")
   end
 
-  def event_factory do
+  def event_factory(attrs) do
+    if Map.get(attrs, :acquisition_channel) do
+      raise "Acquistion channel cannot be written directly since it's a materialized column."
```
ukutaht, line 144: FYI @macobo - I was really confused why my test setup wasn't working 😀

test/support/factory.ex
```diff
@@ -135,21 +135,29 @@ defmodule Plausible.Factory do
...
       site_id: Enum.random(1000..10_000),
       pathname: "/",
       timestamp: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second),
       user_id: SipHash.hash!(hash_key(), Ecto.UUID.generate()),
       session_id: SipHash.hash!(hash_key(), Ecto.UUID.generate())
     }
+
+    event
+    |> merge_attributes(attrs)
+    |> evaluate_lazy_attributes()
```
macobo, line 160: Q: Why is this needed?

ukutaht: In order to get ["full control of factory"](https://hexdocs.pm/ex_machina/readme.html#full-control-of-factory) and inspect the attributes at runtime. Basically I changed from from more to less syntactic sugar with the library in order to get access to the attributes passed into the factory function.

lib/plausible/stats/imported/sql/expression.ex
```diff
@@ -231,6 +231,18 @@ defmodule Plausible.Stats.Imported.SQL.Expression do
     })
   end
 
+  defp select_group_fields(q, "visit:channel", key, _query) do
```
macobo, line 234: We could merge this with

```
  defp select_group_fields(q, dimension, key, _query)
       when dimension in ["visit:device", "visit:browser"] do
```

Just above right

lib/plausible/stats/imported/imported.ex
```diff
@@ -197,6 +197,7 @@ defmodule Plausible.Stats.Imported do
 
   @filter_suggestions_mapping %{
     referrer_source: :source,
+    acquisition_channel: :channel,
```
macobo, line 200: Oof. This took me a bit to unravel. `filter_name` in lib/plausible/stats/filter_suggestions.ex is actually db_column, and we need to convert one db column to the other here.

This is slightly annoying coupling and would be better if we passed the dimension name here instead but probably not worth tackling in this PR.

lib/plausible/imported/google_analytics4.ex
```diff
@@ -175,6 +175,7 @@ defmodule Plausible.Imported.GoogleAnalytics4 do
       import_id: import_id,
       date: get_date(row),
       source: row.dimensions |> Map.fetch!("sessionSource") |> parse_source(),
+      channel: row.dimensions |> Map.fetch!("sessionDefaultChannelGroup"),
```
macobo, line 178: This makes the assumption that our channels map 1-1 to GA4's. This seems to be the case, but for future code travellers it would be great to add a note about this.

macobo (approved):
Looks great, though I didn't touch GA4 import.

Probably needs a rebase after https://github.com/plausible/analytics/pull/4815 is merged.

lib/plausible/stats/imported/imported.ex
ukutaht: > This is slightly annoying coupling and would be better if we passed the dimension name here instead but probably not worth tackling in this PR.

Agreed with both points.

lib/plausible/stats/imported/sql/expression.ex
ukutaht: Thanks!

PR #4816, opened 2024-11-13
Follow-up to Solve noisy warnings about a negative range was inferred for Date.range/2

macobo:
Follow-up to https://github.com/plausible/analytics/pull/4803

We could still get warnings about negative date ranges when the date being queried is in the future.

This could happen if the users local time is in the future for some reason or they manually edit the url.

Ref: https://3.basecamp.com/5308029/buckets/39750953/card_tables/cards/7999429743

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -57,18 +57,18 @@ defmodule Plausible.Stats.Query do
 
       Date.range(
         date_range.first,
-        clamp(today, date_range.first, date_range.last)
+        clamp(today, date_range)
       )
     else
       date_range
     end
   end
 
-  defp clamp(date, lower_bound, upper_bound) do
+  defp clamp(date, date_range) do
     cond do
-      Date.compare(date, lower_bound) == :lt -> lower_bound
-      Date.compare(date, upper_bound) == :gt -> upper_bound
-      true -> date
+      date in date_range -> date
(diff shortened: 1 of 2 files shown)


lib/plausible/stats/query.ex
```diff
@@ -57,15 +57,19 @@ defmodule Plausible.Stats.Query do
...
       )
     else
       date_range
     end
   end
 
-  defp earliest(a, b) do
-    if Date.compare(a, b) in [:eq, :lt], do: a, else: b
+  defp clamp(date, lower_bound, upper_bound) do
+    cond do
```
aerosol, line 68: minor: there's hardly ever a good reason to prefer cond over case

macobo: Q: How would you write this with case?

My initial thought seems worse than the cond version:
```
case date do
  date when Date.compare(date, lower_bound) == :lt -> lower_bound
  # ...
end
```

aerosol: ah right, you can't do arbitrary function calls in guards. in that case (🥁), a slightly more readable version would be:

```elixir
  defp clamp(date, range) do
    case do
      date in range -> date
      Date.before?(date, range.first) -> range.first
      Date.after?(date, range.last) -> range.last
    end
  end
```

aerosol (approved):

PR #4820, opened 2024-11-13
Channels UI polish

ukutaht:
### Changes

* Renames 'All' to 'Sources' in Top Sources report
* Reorders the tabs so that channels is on the left
* Introduces a kind of UI hierarchy in Top Sources the same way we have for locations. When the user clicks on a channel, the UI is automatically switched to the 'Sources' tab.
* Change the title of the report depending on which tab is selected


assets/js/dashboard/stats/sources/source-list.js
```diff
@@ -154,9 +165,23 @@ export default function SourceList() {
   const [currentTab, setCurrentTab] = useState(storedTab || 'all')
   const [loading, setLoading] = useState(true)
   const [skipImportedReason, setSkipImportedReason] = useState(null)
+  const previousQuery = usePrevious(query);
```
ukutaht, line 168: I copied how this was implemented in Locations report, this time with hooks. The idea is that when the user is removing a `channel` filter while on the Top Sources tab, the UI should switch to 'Top Channels'. This gives the UI sort of a hierarchical drilldown feel.

@apata what do you think about this implementation? Any better ideas?

apata: If it works, it works :) 

As for alternatives, I think we could wire up our dashboard report tabs to be governed by [history state](https://developer.mozilla.org/en-US/docs/Web/API/History/state), exposed by [React Router](https://reactrouter.com/en/main/hooks/use-navigate#optionsstate). 

This state can be set in any app navigate call, or in any app link ([I'm using it to toggle segment edit mode for example](https://github.com/plausible/analytics/blob/saved-segments-spec/assets/js/dashboard/segments/segments-dropdown.tsx#L194-L207)).

As I imagine it, clicking on a tab directly would do `navigate({search: s => s, state: {sourcesReportTab: <clicked-tab>}}, {replace: true})`. (_We'd probably want to replace, because otherwise Back button takes you to previously selected tab._)

The channel drilldown links within the channel report would not only set a new search value, they would also do `state: {sourceListTab: 'all'}`. 

We'd read the current tab with useLocation in SourceList. If no tab is defined in location state, we'd read the preferred tab from local storage, falling back to a default value if neither is present. 

We'd also run a simple useEffect in SourceList that saves the tab to localStorage when it changes.

apata: Wait, I read the original comment again. 
I see I misunderstood the trigger: it's the removal of a filter, not adding of a filter. 

My alternative wouldn't work as described then, unless we wire it up globally (because there's multiple ways for the Channel filter to be removed). Previous and current search and therefore previous and current filters are available in getNavigateToOptions in use-app-navigate.tsx.

assets/js/dashboard/hooks/use-previous.ts
```diff
@@ -0,0 +1,15 @@
+import { useRef, useEffect } from 'react';
+
+function usePrevious<T>(value: T): T | undefined {
+  const ref = useRef<T | undefined>(undefined);
+
+  useEffect(() => {
+    // Update the ref with the current value after render
+    ref.current = value;
+  });
```
apata, line 9: **nitpick, non-blocking**: This effect runs more often than it has to. Why don't we say that it depends on `[value]` changing?

assets/js/dashboard/stats/sources/source-list.js
```diff
@@ -154,9 +165,23 @@ export default function SourceList() {
...
+      if (!previousQuery) return false
+
+      return getFiltersByKeyPrefix(previousQuery, filterName).length > 0 &&
+        getFiltersByKeyPrefix(query, filterName).length == 0
+    }
+
+    if (currentTab == 'all' && isRemovingFilter('channel')) {
+      setTab('channels')()
+    }
+  }, [query])
```
apata, line 183: **issue, non-blocking**: Effects that don't list all their dependencies can run with unpredictable values. At the moment, in the review env, all seems to be well though.

apata (approved):

assets/js/dashboard/hooks/use-previous.ts
ukutaht: Thanks. fixed in https://github.com/plausible/analytics/pull/4820/commits/ed4ebee1c09783dfa2228961fdd1dba8e2c01935

assets/js/dashboard/stats/sources/source-list.js
ukutaht: I think I'm going to lean on this part of your response and not get to deep into this for now 😀

> If it works, it works :)

PR #4824, opened 2024-11-14
Hotfix Direct / None modal

apata:
### Changes

1) Passing a react-router path param with a forward slash in it, like "Direct / None", breaks the route. There's two routes that have params that take user input (filter modals has fixed param values), and this PR encodes such params. 
2) useParams gives us a decoded value for the params, therefore we need to re-encode them to make successful API calls. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 
 - Fix returning filter suggestions for multiple custom property values in the dashboard Filter modal
 - Fix typo on login screen
+- Fix Direct / None details modal not opening
 
 ## v2.1.4 - 2024-10-08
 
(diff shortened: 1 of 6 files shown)


ukutaht (approved):
Question: are you encoding conditionally here so we can keep URLs prettier if they don't need encoding?

apata:
> Question: are you encoding conditionally here so we can keep URLs prettier if they don't need encoding?

I didn't want to risk creating another problem where previously working dashboard links break, or API requests differ from previous requests. Wanted to get a hotfix live as a priority.

With a bit more investigation and tests, we could have answers if there is any risk encoding in _all_ scenarios, and to what degree we can maintain nice links.

PR #4826, opened 2024-11-14
Extracted tracker changes from Scroll Depth PR

RobertJoonas:
### Changes

This PR extracts the tracker changes from https://github.com/plausible/analytics/pull/4791, and addressing the feedback left there as well.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -48,6 +48,38 @@
   // flag prevents sending multiple pageleaves in those cases.
   var pageLeaveSending = false
 
+  var bodyEl = document.body
+  var docEl = document.documentElement
+
+  function getDocumentHeight() {
+    return Math.max(
+      bodyEl.scrollHeight || 0,
+      bodyEl.offsetHeight || 0,
+      bodyEl.clientHeight || 0,
+      docEl.scrollHeight || 0,
+      docEl.offsetHeight || 0,
+      docEl.clientHeight || 0
+    )
+  }
+
+  function getCurrentScrollDepthPx() {
+    var viewportHeight = window.innerHeight || docEl.clientHeight || 0
+    var scrollTop = window.scrollY || docEl.scrollTop || bodyEl.scrollTop || 0
+
+    return currentDocumentHeight <= viewportHeight ? currentDocumentHeight : scrollTop + viewportHeight
+  }
+
+  var currentDocumentHeight = getDocumentHeight()
+  var maxScrollDepthPx = getCurrentScrollDepthPx()
+
+  document.addEventListener('scroll', function() {
+    var currentScrollDepthPx = getCurrentScrollDepthPx()
+
+    if (currentScrollDepthPx > maxScrollDepthPx) {
+      maxScrollDepthPx = currentScrollDepthPx
+    }
+  })
+
   function triggerPageLeave() {
     if (pageLeaveSending) {return}
     pageLeaveSending = true
@@ -55,6 +87,7 @@
 
     var payload = {
       n: 'pageleave',
+      sd: Math.round((maxScrollDepthPx / currentDocumentHeight) * 100),
       d: dataDomain,
       u: currentPageLeaveURL,
     }
@@ -202,6 +235,8 @@
       if (isSPANavigation && listeningPageLeave) {
         triggerPageLeave();
         currentPageLeaveURL = location.href;
+        currentDocumentHeight = getDocumentHeight()
+        maxScrollDepthPx = getCurrentScrollDepthPx()
       }
       {{/if}}
 
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -48,23 +48,20 @@
   // flag prevents sending multiple pageleaves in those cases.
   var pageLeaveSending = false
 
-  var bodyEl = document.body
-  var docEl = document.documentElement
-
   function getDocumentHeight() {
     return Math.max(
-      bodyEl.scrollHeight || 0,
-      bodyEl.offsetHeight || 0,
-      bodyEl.clientHeight || 0,
-      docEl.scrollHeight || 0,
-      docEl.offsetHeight || 0,
-      docEl.clientHeight || 0
+      document.body.scrollHeight || 0,
+      document.body.offsetHeight || 0,
+      document.body.clientHeight || 0,
+      document.documentElement.scrollHeight || 0,
+      document.documentElement.offsetHeight || 0,
+      document.documentElement.clientHeight || 0
     )
   }
 
   function getCurrentScrollDepthPx() {
-    var viewportHeight = window.innerHeight || docEl.clientHeight || 0
-    var scrollTop = window.scrollY || docEl.scrollTop || bodyEl.scrollTop || 0
+    var viewportHeight = window.innerHeight || document.documentElement.clientHeight || 0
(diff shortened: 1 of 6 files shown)


tracker/test/scroll-depth.spec.js
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable playwright/expect-expect */
+/* eslint-disable playwright/no-skipped-test */
+const { clickPageElementAndExpectEventRequests, mockRequest } = require('./support/test-utils')
+const { test } = require('@playwright/test');
```
macobo, line 4: Nit: Semi-colon usage in specs is quite inconsistent. I suggest we don't use unneeded semicolons.

tracker/test/scroll-depth.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+test.describe('scroll depth', () => {
+  test.skip(({browserName}) => browserName === 'webkit', 'Not testable on Webkit');
+
+  test('sends scroll_depth in the pageleave payload when navigating to the next page', async ({ page }) => {
+    const pageviewRequestMock = mockRequest(page, '/api/event')
+    await page.goto('/scroll-depth.html');
+    await pageviewRequestMock;
+
+    // eslint-disable-next-line no-undef
```
macobo, line 15: Nit: Let's disable this file-wide.

tracker/src/plausible.js
```diff
@@ -48,13 +48,47 @@
...
+    var scrollTop = window.scrollY || document.documentElement.scrollTop || document.body.scrollTop || 0
+
+    return currentDocumentHeight <= viewportHeight ? currentDocumentHeight : scrollTop + viewportHeight
+  }
+
+  var currentDocumentHeight = getDocumentHeight()
+  var maxScrollDepthPx = getCurrentScrollDepthPx()
+  
+  window.addEventListener('load', function () {
+    currentDocumentHeight = getDocumentHeight()
```
macobo, line 73: Q: Shouldn't we also refresh this on scroll / when we go deeper than before?

Thinking of scenarios where content slowly loads, e.g. SPA.

tracker/test/scroll-depth.spec.js
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable playwright/expect-expect */
+/* eslint-disable playwright/no-skipped-test */
+const { clickPageElementAndExpectEventRequests, mockRequest } = require('./support/test-utils')
+const { test } = require('@playwright/test');
+const { LOCAL_SERVER_ADDR } = require('./support/server');
+
+test.describe('scroll depth', () => {
```
macobo, line 7: These are great tests. :clap:

tracker/test/scroll-depth.spec.js
RobertJoonas: I think the rule can still be useful when writing code, so instead, I added `window` and `document` as "globals" in eslint rules, so that this specific rule is quiet about these specific var names (https://github.com/plausible/analytics/pull/4826/commits/0f0d600185c510d900a2740ae88a283e596c0a5e).

Also added two other playwright specific rules in https://github.com/plausible/analytics/pull/4826/commits/7892a17090bd9d6060940a70693c3bac1585038e:

* Do not error when using `page.waitForTimeout` - this is something we often need because we're testing the script in a completely isolated way
* Consider `expectCustomEvent` and `clickPageElementAndExpectEventRequests` functions to be `expect` statements too

tracker/src/plausible.js
RobertJoonas: I wouldn't do it on scroll event due to the reasons I mentioned [here](https://github.com/plausible/analytics/pull/4791#discussion_r1843782333):

> TBH, I don't think reevaluating currentDocumentHeight on scroll event makes sense. The only way the document height could change on scroll is if additional resources are being loaded (e.g. infinite scroll). but we're only using the currentDocumentHeight to calculate the percentage, which is pretty much useless in such cases.

> For example, 75% scroll depth could mean that the user scrolled 75% of the first page, but at the same time, it could be that they scrolled to the end of the document three times and loaded more content every time (i.e. actually consuming 3/4 full pages of content)

> To me it sounds more useful to consider all users who consumed more than one page of content as 100% scroll depth. Of course, measuring scroll depth in pixels would be by far more useful in these cases. Providing that option to the users still needs to be figured out.

But you bring a valid point about dynamically loaded content (e.g. a XHR request after window load). In these cases, updating documentHeight on scroll wouldn't really help either - we want the document height to change after that content is loaded, even if the user doesn't scroll at all.

To tackle this, I've implemented a limited time interval to re-evaluate the documentHeight after the initial HTML has been loaded. I've currently set it to update every 200ms during 3 seconds (that is 15 times in total, 16 if you count the initial one on Window `load`).

https://github.com/plausible/analytics/pull/4826/commits/8eec711d1ced02086d8acb7ebd9fe5a3b556b1d3

macobo: Okay. Let's go with this rough approach `currentDocumentHeight`

tracker/src/plausible.js
```diff
@@ -48,13 +48,56 @@
...
+  var maxScrollDepthPx = getCurrentScrollDepthPx()
+
+  window.addEventListener('load', function () {
+    currentDocumentHeight = getDocumentHeight()
+
+    // Update the document height again after every 200ms during the
+    // next 3 seconds. This makes sure dynamically loaded content is
+    // also accounted for.
+    var count = 0
+    var interval = setInterval(function () {
```
macobo, line 79: This seems super hacky and I don't trust it. 3 seconds might not be enough on a bad connection

Could we update this on scroll instead?

RobertJoonas: Still need to keep the `setInterval` because the user might not scroll at all. But sure, we can also do that on scroll. https://github.com/plausible/analytics/pull/4826/commits/6bc3a297b2ef0e3ddfda90af068464a59b0dcad3

tracker/src/plausible.js
```diff
@@ -48,13 +48,57 @@
...
+      currentDocumentHeight = getDocumentHeight()
```
macobo, line 80: I wonder if we should also update maxScrollDepthPx here as well to account for any weird reflows.

tracker/test/scroll-depth.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    // During the first 3 seconds, the script periodically updates document height
+    // to account for dynamically loaded content. Since we want to test document
+    // height also getting updated on scroll, we need to just wait for 3 seconds.
+    await page.waitForTimeout(3100)
+
+    // scroll to the bottom of the page
+    await page.evaluate(() => window.scrollBy(0, document.body.scrollHeight))
+    
+    // Wait until documentHeight gets increased by the fixture JS
+    await page.waitForSelector('#more-content')
```
macobo, line 81: Note: It might have been simpler to evaluate JS on the page appending a div/img rather than rely on timers which might be slightly finnicky.

macobo (approved):
Small notes but :+1:

tracker/test/scroll-depth.spec.js
RobertJoonas: Good note! The timeout in the test fixture was intended only for debugging. I've removed it now. It now actually works by simply appending a 2000px div to the document.

PR #4839, opened 2024-11-19
fractional_sample_rate: Vary sample rate according to traffic estimation

macobo:
Our old sampling mechanism used SAMPLE 20000000 syntax. This was wonderful since it allowed essentially dynamic sampling based on the data being queried. However this ran into many issues relating to JOINs and sample rate being different for different tables.

Instead, we now start to dynamically vary sample rates fractionally.

At query time we check the time window being queried and estimate how many rows this query might reach. For large queries, we then dynamically decide the sample rate.

For getting the traffic estimate for a site, we have a new SampingCache class which queries `ingest_counters`.

Future work:
- The query being cached is slightly expensive and can be sped up with a ClickHouse projection. This will be added in a follow-up PR.
- Reducing sample rate if `query.filters` is set.

### DIFF
--- a/extra/lib/plausible/stats/sampling_cache.ex
+++ b/extra/lib/plausible/stats/sampling_cache.ex
@@ -50,7 +50,6 @@ defmodule Plausible.Stats.SamplingCache do
   end
 
   def thirty_days_ago() do
-    Date.utc_today()
-    |> Date.add(-30)
+    Date.shift(Date.utc_today(), day: -30)
   end
 end
--- a/extra/lib/plausible/stats/sampling_cache.ex
+++ b/extra/lib/plausible/stats/sampling_cache.ex
@@ -1,0 +1,56 @@
+defmodule Plausible.Stats.SamplingCache do
+  @moduledoc """
+  Cache storing estimation for events ingested by a team in the past month.
+
+  Used for sampling rate calculations in Plausible.Stats.Sampling.
+  """
+  alias Plausible.Ingestion
+  alias Plausible.Stats.Sampling
+
+  import Ecto.Query
+  use Plausible.Cache
+
+  @cache_name :stats_sampling_cache
+
+  @impl true
+  def name(), do: @cache_name
+
+  @impl true
+  def child_id(), do: :cache_stats_sampling
+
+  @impl true
+  def repo(), do: Plausible.ClickhouseRepo
+
+  @impl true
+  def count_all() do
+    base_db_query()
+    |> repo().all()
+    |> length()
+  end
+
+  @impl true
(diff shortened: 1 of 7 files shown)


test/plausible/stats/sampling_cache_test.exs
```diff
@@ -0,0 +1,64 @@
...
+            metric: "buffered"
+          }
+        ])
+
+        {:ok, _} = start_test_cache(test)
+
+        assert SamplingCache.get(1) == nil
+        assert SamplingCache.get(2) == 22_000_000
+        assert SamplingCache.get(3) == nil
+        assert SamplingCache.get(4) == nil
```
aerosol, line 51: Here you need to be specific if you need to hit the cache you've just started. 

```suggestion
        assert SamplingCache.get(1, force?: true, cache_name: test) == nil
        assert SamplingCache.get(2, force?: true, cache_name: test) == 22_000_000
        assert SamplingCache.get(3, force?: true, cache_name: test) == nil
        assert SamplingCache.get(4, force?: true, cache_name: test) == nil
```

Otherwise `get_from_source/1` is called for convenience - for cases when starting the cache isn't relevant for code under test.

aerosol: In a similar fashion, probably good to test `count_all` or even `refresh_all` to guard them from regression.

aerosol (approved):
There's a small problem with cache tests, but otherwise looks good.  Happy to see counters and cache stitched together for greater good.

test/plausible/stats/sampling_cache_test.exs
```diff
@@ -45,10 +45,10 @@ defmodule Plausible.Stats.SamplingCacheTest do
 
         {:ok, _} = start_test_cache(test)
```
aerosol, line 46: oops, best to `refresh_all` before hitting the cache?

test/plausible/stats/sampling_cache_test.exs
macobo: Thanks! Updated this.

extra/lib/plausible/stats/sampling_cache.ex
```diff
@@ -0,0 +1,56 @@
...
+  def child_id(), do: :cache_stats_sampling
+
+  @impl true
+  def repo(), do: Plausible.ClickhouseRepo
+
+  @impl true
+  def count_all() do
+    base_db_query()
+    |> repo().all()
+    |> length()
```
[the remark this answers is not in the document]
macobo: This wouldn't speed up the query and would result in extra code to manage - I don't think it's worth it.

PR #4846, opened 2024-11-21
Log mailer errors in CE

ruslandoga:
### Changes

This PR improves Plausible.Mailer error messages in CE.
It tries to keep changes to a minimum and only changes how exceptions are logged.

Similar to https://github.com/plausible/analytics/pull/4657

Requested in https://github.com/plausible/analytics/discussions/4844#discussioncomment-11321492

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/mailer.ex
```diff
@@ -28,6 +28,16 @@ defmodule Plausible.Mailer do
     end
   end
 
+  defp handle_error(error) when is_exception(error) do
+    # this message is ignored by Sentry, but it's useful in CE
+    Logger.error("Failed to send e-mail:\n\n " <> Exception.format(:error, error),
+      # Sentry report is built entirely from crash_reason
+      crash_reason: {error, _stacktrace = []}
```
ruslandoga, line 35: I can add `message: "Failed to send e-mail"` option to make it appear the same way as before in Sentry.

PR #4856, opened 2024-11-26
CI: Reduce separate jobs being executed for elixir

macobo:
We're not expecting a lot of incompatibilities with pg 15 so it makes sense to combine these with reading team schemas


aerosol (approved):
Any reason to test against pg16 at all currently? As far as I can see, all envs run 15? 🤔

cnkk (approved):

PR #4857, opened 2024-11-26
Perf: Improve performance of top stats query with long time windows

macobo:
We usually don't read events/pageviews from sessions table as sessions can start before or after the time window of the query. This is particularly pronounced with realtime queries.

After this change we start reading events/pageviews from sessions table for longer time periods as the inaccuracies introduced by the session timing is minimized this way.

The end result is that queries for e.g. top stats on dashboard load for 7 days or longer periods only requires reading sessions table.

### DIFF
--- a/lib/plausible/stats/table_decider.ex
+++ b/lib/plausible/stats/table_decider.ex
@@ -62,10 +62,6 @@ defmodule Plausible.Stats.TableDecider do
     end
   end
 
-  # Note: This is inaccurate when filtering but required for old backwards compatibility
-  defp metric_partitioner(%Query{legacy_breakdown: true}, :pageviews), do: :either
-  defp metric_partitioner(%Query{legacy_breakdown: true}, :events), do: :either
-
   defp metric_partitioner(_, :conversion_rate), do: :either
   defp metric_partitioner(_, :group_conversion_rate), do: :either
   defp metric_partitioner(_, :visitors), do: :either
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/table_decider.ex
```diff
@@ -75,12 +71,24 @@ defmodule Plausible.Stats.TableDecider do
...
+  defp metric_partitioner(query, metric) when metric in [:pageviews, :events] do
+    query_days = DateTime.diff(query.utc_time_range.last, query.utc_time_range.first, :day)
+
+    cond do
+      # Note: This is inaccurate when filtering but required for old backwards compatibility
+      legacy_breakdown?(query) -> :either
+      # When querying large time window aggregates, session-level event/pageview counts
+      # are accurate enough: the amount of sessions starting/ending outside of time window
+      # doesn't matter as much.
+      query_days > 5 and empty?(query.filters) and empty?(query.dimensions) -> :either
```
apata, line 87: This condition would be true for some periods on virgin dashboards (of users that have just started using Plausible). When the counts are in the single digits, the count changing when cycling through various periods may be mistaken for a bug.

lib/plausible/stats/table_decider.ex
```diff
@@ -97,6 +105,9 @@ defmodule Plausible.Stats.TableDecider do
 
   defp dimension_partitioner(_, _), do: :either
 
+  defp legacy_breakdown?(%Query{} = query), do: query.legacy_breakdown
+  defp legacy_breakdown?(query) when is_map(query), do: query[:legacy_breakdown]
```
apata, line 109: This looks like a new condition: does that mean we missed a few legacy breakdowns cases previously?

lib/plausible/stats/table_decider.ex
apata: It may be an acceptable tradeoff, but maybe it's possible to describe a condition where the differences in counts really don't matter? For whom is the events table query most expensive actually?

lib/plausible/stats/table_decider.ex
macobo: Didn't matter in practice.

TableDecider might be called with a map instead of a full query during query building for validation purposes (see #validate_no_metrics_filters_conflict).

lib/plausible/stats/table_decider.ex
macobo: Discussed this live. 

A problem with this approach is that when user looks at montly view and sees that $DAY had 9 visitors and after clicking on $DAY sees they had 8 visitors, this might trigger some WTFs from our users.

Arturs suggestion was to change this logic to only _not_ read from sessions in realtime view. 

I think this is valid, but want to ponder this a bit more and figure out the implications. Marking as a draft for now.

PR #4880, opened 2024-12-05
Fix: Add missing SAMPLE BY to query

macobo:
### Changes

There was a branch in code where sampling was never triggered. This fixes that issue.

### DIFF
--- a/lib/plausible/stats/sql/query_builder.ex
+++ b/lib/plausible/stats/sql/query_builder.ex
@@ -60,7 +60,7 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
         )
 
       on_ee do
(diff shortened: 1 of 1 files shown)


lib/plausible/stats/sql/query_builder.ex
```diff
@@ -59,6 +59,10 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
           group_by: s.session_id
         )
 
+      on_ee do
+        sessions_q = Plausible.Stats.Sampling.add_query_hint(q, query)
```
macobo, line 63:
```suggestion
        sessions_q = Plausible.Stats.Sampling.add_query_hint(sessions_q, query)
```

ukutaht (approved):

PR #4882, opened 2024-12-05
Remove LegacyDashboardFilterParser

macobo:
### Changes

LegacyDashboardFilterParser has not been hit for a while now, this PR removes it completely by updating a bunch of tests.

It also removes an unused feature from referrer_drilldown endpoint.


lib/plausible/stats/filters/utils.ex
```diff
@@ -1,7 +1,6 @@
 defmodule Plausible.Stats.Filters.Utils do
   @moduledoc """
-  Contains utility functions shared between `LegacyDashboardFilterParser`
-  and `StatsAPIFilterParser`.
+  Contains utility functions for `StatsAPIFilterParser`.
```
ukutaht, line 3: Suggestion: inline this module to `StatsAPIFilterParser`

ukutaht (approved):

PR #4885, opened 2024-12-06
Log Mailer errors

ruslandoga:
### Changes

This PR improves Plausible.Mailer error messages in CE.

Requested in https://github.com/plausible/analytics/discussions/4844#discussioncomment-11321492

### DIFF
--- a/lib/plausible/mailer.ex
+++ b/lib/plausible/mailer.ex
@@ -2,34 +2,21 @@ defmodule Plausible.Mailer do
   use Bamboo.Mailer, otp_app: :plausible
   require Logger
 
-  @type result() :: :ok | {:error, :hard_bounce} | {:error, :unknown_error}
-
-  @spec send(Bamboo.Email.t()) :: result()
+  @spec send(Bamboo.Email.t()) :: :ok | {:error, :unknown_error}
   def send(email) do
-    case deliver_now(email) do
-      {:ok, _email} -> :ok
-      {:ok, _email, _response} -> :ok
-      {:error, error} -> handle_error(error)
-    end
-  end
+    try do
+      deliver_now!(email)
+    rescue
+      e ->
+        # this message is ignored by Sentry, only appears in logs
+        log = "Failed to send e-mail:\n\n  " <> Exception.format(:error, e, __STACKTRACE__)
+        # Sentry report is built entirely from crash_reason
+        crash_reason = {e, __STACKTRACE__}
 
-  defp handle_error(%{response: response}) when is_binary(response) do
-    case Jason.decode(response) do
-      {:ok, %{"ErrorCode" => 406}} ->
-        {:error, :hard_bounce}
-
-      {:ok, response} ->
-        Logger.error("Failed to send e-mail", sentry: %{extra: %{response: inspect(response)}})
+        Logger.error(log, crash_reason: crash_reason)
         {:error, :unknown_error}
-
-      {:error, _any} ->
-        Logger.error("Failed to send e-mail", sentry: %{extra: %{response: inspect(response)}})
-        {:error, :unknown_error}
+    else
+      _sent_email -> :ok
     end
-  end
-
-  defp handle_error(error) do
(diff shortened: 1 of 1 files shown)


lib/plausible/mailer.ex
```diff
@@ -28,8 +29,22 @@ defmodule Plausible.Mailer do
     end
   end
 
-  defp handle_error(error) do
-    Logger.error("Failed to send e-mail", sentry: %{extra: %{response: error}})
-    {:error, :unknown_error}
+  on_ee do
```
ruslandoga, line 32: I went with `on_ee` because I don't fully understand the error reporting in EE (https://github.com/plausible/analytics/pull/4846#discussion_r1851650174) and this is the easiest workaround.

ukutaht: I can confirm that the `%{extra: %{response: error}}` stuff does not end up in our Sentry instance. This is quite confusing because their docs seem to suggest that extra information can be added via Logger metadata: https://hexdocs.pm/sentry/Sentry.LoggerHandler.html#module-examples.

In general it looks like error reporting here is fairly broken and I think we should fix it across the board in this PR instead of just in CE.

aerosol: Needs to be `inspect(error)` I suppose.

ukutaht: > Needs to be inspect(error) I suppose.

Who tracks errors in the error tracker? 😀 yeah, could be that it cannot handle a nested structure like this.

> I think Sentry works as expected ... Maybe it's just that handle_error is never called in Plausible: https://github.com/plausible/analytics/pull/4846#discussion_r1851650174

Yeah I think it's very possible that the [first function](https://github.com/plausible/analytics/blob/e82ae6599d085129a4f5668212329cc7db2116d3/lib/plausible/mailer.ex#L16-L29) clause never gets executed. But the [second one should](https://github.com/plausible/analytics/blob/e82ae6599d085129a4f5668212329cc7db2116d3/lib/plausible/mailer.ex#L31-L34) and it includes a `response` key which never ends up in Sentry.

ruslandoga: Right, seems like it's more complicated than that.

Do you have `"Failed to send-email"` error in Sentry? What are they grouped by? Do you have these warnings in logs?

```
16:06:33.533 [warning] Event dropped due to being a duplicate of a previously-captured event.
```

ruslandoga: And yes, `sentry.extra` Logger metadata needs to be a flat map.

ruslandoga: > In general it looks like error reporting here is fairly broken and I think we should fix it across the board in this PR instead of just in CE.

@ukutaht I tried doing this in my previous attempt but without having access and seeing the actual way the errors are being reported right now, I don't think I have enough information to do it well. So I'll keep it CE-specific, at least for the upcoming CE release.

zoldar (approved):

ruslandoga:
@ukutaht double checking if it's OK to merge?

lib/plausible/mailer.ex
```diff
@@ -2,34 +2,21 @@ defmodule Plausible.Mailer do
...
-        Logger.error("Failed to send e-mail", sentry: %{extra: %{response: inspect(response)}})
+        Logger.error(log, crash_reason: crash_reason)
         {:error, :unknown_error}
+    else
+      _success -> :ok
     end
   end
-
-  defp handle_error(error) do
-    Logger.error("Failed to send e-mail", sentry: %{extra: %{response: inspect(error)}})
```
ruslandoga, line 32: BTW, turns out inspect wasn't necessary. The problem was (is?) in the Sentry logger handler, not nested extra. Sentry logger backend handles nested extra correctly.

ukutaht (approved):

ruslandoga:
Possible fix for hard bounces Bamboo.ApiError Sentry flood: https://github.com/plausible/bamboo_postmark/pull/3

PR #4889, opened 2024-12-09
Average Scroll Depth Metric: handle missing data better

RobertJoonas:
### Changes

* Start ingesting missing scroll depth values as 255 (max `UInt8`) rather than 0 for pageleave events
* Display "-" instead of 0 when no scroll depth data in range
* Add a data migration setting scroll depth of pageleaves to 255 if it's 0

Reasoning:

A scroll depth of 0 currently represents a missing value, but it can also be a valid scroll depth in case of extremely short viewports and long documents. Also, 0's are currently counted towards average which we don't want to do in case of missing values

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/data_migration/missing_scroll_depth.ex
+++ b/lib/plausible/data_migration/missing_scroll_depth.ex
@@ -1,0 +1,18 @@
+defmodule Plausible.DataMigration.MissingScrollDepth do
+  @moduledoc """
+  Set scroll_depth to 255 (max UInt8) for all pageleave events where it's 0.
+  """
+
+  def run() do
+    Plausible.IngestRepo.query!(
+      """
+      ALTER TABLE events_v2
+      #{Plausible.MigrationUtils.on_cluster_statement("events_v2")}
+      UPDATE scroll_depth = 255
+      WHERE name = 'pageleave' AND scroll_depth = 0
+      """,
+      [],
+      timeout: 60_000
+    )
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/plausible/data_migration/missing_scroll_depth.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Plausible.DataMigration.MissingScrollDepth do
```
macobo, line 1: Q: Why is this a data migration not a normal one? Do we want to maintain it long-term?

RobertJoonas: I just figured it's the preferred way of doing it since are no schema changes involved. When included in the normal migration pipeline, it would run every time when running migrations from scratch. Thought that would be unnecessary.

RobertJoonas: > Do we want to maintain it long-term?

Nope - we could also just run `Plausible.DataMigration.MissingScrollDepth.run()` in the remote shell and delete this file later on.

macobo (approved):

lib/plausible/data_migration/missing_scroll_depth.ex
macobo: Note we could also just run the same sql directly in clickhouse, but will leave it up to you.

PR #4891, opened 2024-12-10
Add Saved Segments UI (variant D)

apata:
### Changes

Adds saved segments related menus and modals. Almost all of it is conditional on feature flag saved_segments: true. 

### Tests

### Changelog
- [x] This PR does not make a user-facing change (for those without the FF)

### Documentation

### Dark mode


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -397,6 +411,30 @@ defmodule PlausibleWeb.StatsController do
       end)
       |> Map.new()
 
+  defp get_members(site_role, %Plausible.Site{} = site)
```
apata, line 414: This is a workaround to avoid resolving segment.owner_id to segment.owner_name in the backend. Instead, we forward users <id>:<name> map to the front end.

assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -232,9 +257,8 @@ export const FiltersBar = ({ accessors }: FiltersBarProps) => {
 
 const ClearAction = () => (
   <AppNavigationLink
-    title="Clear all filters"
     className={classNames(
-      'self-start button h-9 !px-3 !py-2 flex !bg-red-500 dark:!bg-red-500 hover:!bg-red-600 dark:hover:!bg-red-700 whitespace-nowrap'
+      'button flex self-start h-9 !px-3 !bg-red-500 dark:!bg-red-500 hover:!bg-red-600 dark:hover:!bg-red-700 whitespace-nowrap'
```
ukutaht, line 261: I think the `.button` class should be considered legacy and not used for any new stuff. This applies to pretty much everything in `app.css`. These were added due to missing features in tailwind that would not allow us to use tailwind directly. But tailwind has been developed a lot and pretty much everything is possible now by just using tailwind classes.

assets/js/dashboard/segments/segment-authorship.tsx
```diff
@@ -0,0 +1,55 @@
...
+) => {
+  if (!site.members) {
+    return ''
+  }
+
+  if (!owner_id || !site.members[owner_id]) {
+    return '(Removed User)'
+  }
+
+  // if (owner_id === user.id) {
```
ukutaht, line 20: Should this be uncommented? Deleted?

lib/plausible_web/controllers/stats_controller.ex
ukutaht: Why not resolve it on the backend?

ukutaht (approved):
Looks good overall but the quality of my review is meh. Touches a bunch of code I haven't worked with in a long time, plus the PR is pretty big.

assets/js/dashboard/nav-menu/filters-bar.tsx
apata: True, I saw that it was sort of wonky, and I'd be ok with getting rid of it. At the moment, I think let's leave it to mark which components should ideally share style. 

Separate topic, but I had a notion that these custom classes made using TW @apply in app.css could be used to share class combinations between server-rendered pages and React dashboard. 

Now that we have the storybook as well, it's especially tempting to try to codify and share these basic styles, like texts, notices, button types, dropdown backgrounds.

lib/plausible_web/controllers/stats_controller.ex
apata: Mostly because this worked ok and the PR was getting too bloated already.

assets/js/dashboard/nav-menu/filters-bar.tsx
ukutaht: Hmm yes they could be used for sharing...

For me though, one of the biggest things I like about tailwind is not having to come up with class names, and especially names for modifier classes :) I prefer the style where template and style code are combined and named together as a component (like `<.button theme={"plain"} disabled="true">` over `.button .button--plain .button--disabled`).

lib/plausible_web/controllers/stats_controller.ex
ukutaht: Got it. I think the logic would be more cohesive if the owners were resolved in the same query as segments themselves. But I agree with not making the PR any bigger than it is.

PR #4893, opened 2024-12-11
Revenue metrics in APIv2

macobo:
### Changes

This PR exposes revenue metrics in APIv2 (only on ee).

There are a few scenarios where revenue metrics can't be calculated. On the dashboard we automagically remove these metrics but this isn't a great fit for APIv2. Instead, we return nulls and produce warnings under `response.meta`. This is similar to imports where we still return results for the rest of the query. 

Ref: https://3.basecamp.com/5308029/buckets/39750953/card_tables/cards/8070760266

Docs PR: https://github.com/plausible/docs/pull/571. Will wait for a :+1: from Looker integration before merging.


RobertJoonas (approved):
Looks good to me! 🙌

And thanks for introducing the `metric_warnings` in this format. I believe we can nicely use that for scroll depth as well when imported data is not being included. :)

PR #4908, opened 2024-12-16
Migration: Add scroll_depth to imported_pages

RobertJoonas:
### Changes

This PR only contains the migration adding the `scroll_depth UInt8` column to the `imported_pages` ClickHouse table.

The column will have a default value of 255 which hints that the value is non-existent (a valid scroll depth has to be from 0 to 100). The alternatives default values are:

* 0 - cannot use because it can be a valid value in rare cases
* NULL - bad performance in ClickHouse

When querying scroll depth from imported data, we'll simply filter out rows where `scroll_depth` is more than 100.

### DIFF
--- a/priv/ingest_repo/migrations/20241216133031_add_scroll_depth_to_imported_pages.exs
+++ b/priv/ingest_repo/migrations/20241216133031_add_scroll_depth_to_imported_pages.exs
@@ -1,0 +1,17 @@
+defmodule Plausible.IngestRepo.Migrations.AddScrollDepthToImportedPages do
+  use Ecto.Migration
+
+  def up do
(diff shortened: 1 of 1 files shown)


apata (approved):
Is it irreversible actually? Seems to me that we could drop it just as easily as we add it.

ukutaht (approved):

RobertJoonas:
> Is it irreversible actually? Seems to me that we could drop it just as easily as we add it.

You're right - I just figured that rolling it back would not be necessary in any case. The column has a default value, and it's not in the schema yet, so there shouldn't by any issues. I'll also run it on staging first and complete a full export -> import flow.

PR #4910, opened 2024-12-16
Fix membership role action and checks under Site Settings > People

zoldar:
### Changes

This logic will be soon revised (and replaced) once we commence with implementing teams support.

This is a fix for a regression where Admin role update under Site Settings > People is failing due to wrong mapping.

### DIFF
--- a/lib/plausible_web/templates/site/settings_people.html.heex
+++ b/lib/plausible_web/templates/site/settings_people.html.heex
@@ -127,7 +127,7 @@
                           :update_role_by_user,
                           @site.domain,
                           membership.user.id,
-                          "admin"
+                          "editor"
                         )
                       }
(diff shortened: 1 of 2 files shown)


lib/plausible_web/templates/site/settings_people.html.heex
```diff
@@ -142,7 +142,7 @@
                         </p>
                       </div>
 
-                      <%= if membership.role == :admin do %>
+                      <%= if membership.role == "admin" do %>
```
zoldar, line 145: Currently `membership.role` type here is inconsistent - for owner, it's an atom, for the rest, it's a string. We'll correct that on subsequent rewrite.

aerosol (approved):

PR #4862, opened 2024-11-28
Drop Logger backends

ruslandoga:
### Changes

This PR switches from Logger backends to `:logger` handlers.

Card: https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/7610404292

Some time later, once or if there is a need for structured logging, we can also replace ExJsonFormatter with [something else.](https://github.com/plausible/analytics/pull/4855)

### Tests
- [x] This PR does not require tests

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #4914, opened 2024-12-17
update Sentry

ruslandoga:
### Changes

This PR updates Sentry client to the latest version to take advantage of the recent Sentry.LoggerHandler improvements.


ruslandoga:
Alternatively, we can revert #4862

lib/plausible_web/views/email_view.ex
```diff
@@ -16,7 +16,7 @@ defmodule PlausibleWeb.EmailView do
     Calendar.strftime(date, "%-d %b %Y")
   end
 
-  def sentry_link(trace_id, dsn \\ Sentry.Config.dsn()) do
+  def sentry_link(trace_id, dsn \\ Sentry.get_dsn()) do
```
ruslandoga, line 19: Was `Sentry.Config.dsn()` ever a binary?

lib/sentry_filter.ex
```diff
@@ -10,17 +10,6 @@ defmodule Plausible.SentryFilter do
   def before_send(%{original_exception: %Plug.CSRFProtection.InvalidCSRFTokenError{}}), do: false
   def before_send(%{original_exception: %Plug.Static.InvalidPathError{}}), do: false
 
-  def before_send(
-        %{exception: [%{type: "Clickhousex.Error"}], original_exception: %{code: code}} = event
```
ruslandoga, line 14: Clickhousex is no longer being used.

ruslandoga:
There are too many changes that I'm not sure about. I think it would be safer to revert to LoggerBackend and handle the upgrade later.

https://3.basecamp.com/5308029/buckets/29267832/card_tables/cards/8135181616

ruslandoga:
https://github.com/plausible/analytics/pull/4919

PR #4916, opened 2024-12-17
Add return_to behaviour to 404 -> login flow

ukutaht:
### Changes

It's quite annoying at the moment to:
1. Receive a link to a dashboard (maybe with filters)
2. Get 404
3. Go to login
4. Enter credentials
5. Be redirected to /sites instead of the original link

This PR implements a `return_to` query param used in login flow, including when 2fa challenge is presented. This way the user lands on the intended page.

We currently have a session-based `login_dest` variable but it only works for known routes which dashboards are not. It could have been extended for this use-case but I felt like using query params is better when session state isn't strictly necessary.


lib/plausible_web/templates/error/404_error.html.heex
```diff
@@ -1,11 +1,15 @@
 <div class="container flex flex-col items-center text-center mt-24">
   <h1 class="text-5xl font-black dark:text-gray-100"><%= @status %></h1>
   <div class="mt-4 text-xl dark:text-gray-100">Oops! There's nothing here</div>
-  <div class="text-xl dark:text-gray-100">
-    Trying to access your dashboard? You may need to log in again to see it
-  </div>
-  <div class="mt-6 flex">
-    <.button_link href={Routes.auth_path(@conn, :login_form)} class="mr-4">Log in</.button_link>
+  <%= if !@conn.assigns[:current_user] do %>
```
ukutaht, line 4: The CTA to log in only makes sense if there is no current user

lib/plausible_web/templates/error/404_error.html.heex
```diff
@@ -1,11 +1,20 @@
...
+  <% end %>
+  <div class="mt-4 flex">
+    <%= if !@conn.assigns[:current_user] do %>
+      <.button_link
+        href={Routes.auth_path(@conn, :login_form, return_to: url_path(@conn))}
+        class="mr-4"
+      >
+        Log in
+      </.button_link>
+    <% end %>
```
zoldar, line 17: As it's heex already, we can use a bit less noisy syntax here:

```suggestion

  <div :if={!@conn.assigns[:current_user]} class="text-xl dark:text-gray-100">
    Trying to access your dashboard? You may need to log in again to see it
  </div>
  <div class="mt-4 flex">
    <.button_link
      :if={!@conn.assigns[:current_user]}
      href={Routes.auth_path(@conn, :login_form, return_to: url_path(@conn))}
      class="mr-4"
    >
      Log in
    </.button_link>
```

lib/plausible_web/templates/auth/login_form.html.heex
```diff
@@ -33,6 +33,8 @@
       <div class="text-red-500 mt-4"><%= @conn.assigns[:error] %></div>
     <% end %>
 
+    <.input type="hidden" field={f[:return_to]} value={@conn.assigns[:return_to]} />
```
zoldar, line 36: I'm not 100% sure now, but given "return_to" always comes in params, this could be just:

```
<.input type="hidden" field={f[:return_to]} />
```

and we could stop passing it around explicitly in assigns maybe? 🤔

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -324,12 +330,13 @@ defmodule PlausibleWeb.AuthController do
     end
   end
 
-  def verify_2fa_form(conn, _) do
+  def verify_2fa_form(conn, params) do
```
zoldar, line 333: I assume verify_2fa_recovery_code(_form) is omitted intentionally as an edge case not worth covering?

zoldar (commented):
Looks good - left some minor suggestions. Before the final 🦭 I'll take it for a short spin on preview app though.

lib/plausible_web/templates/auth/login_form.html.heex
zoldar: This might also help with cases where we miss to pass the param around to assigns, like failed verify_2fa code submission - we loose `return_to` param there. This will be definitely more common than 2FA recovery code case 😅

apata (changes requested):
@ukutaht I'm a bit concerned that as-is this can be used to craft malicious links. 
https://pr-4916.review.plausible.io/login?return_to=https:%2F%2Fexample.com%2Fhack%2Fplausible.io
If on the return URL is a copy of the Plausible login page, the user may naively enter their details again, inadvertently sharing their details with a hacker.

zoldar:
@apata good point. The `return_to` URL could be sanitized before final redirect, scraping any with non-matching host part.

lib/plausible_web/templates/auth/login_form.html.heex
ukutaht: > and we could stop passing it around explicitly in assigns maybe? 🤔

Ah yes, super nice! I didn't realize forms worked this way. I've removed the explicit assigns. I've also made sure that `return_to` is carried along in the failed form submissions: https://github.com/plausible/analytics/pull/4916/commits/b4448bd6ad70d8e3d650e8af2e62c81e5a1251df

lib/plausible_web/controllers/auth_controller.ex
ukutaht: Yes, exactly.

ukutaht:
> I'm a bit concerned that as-is this can be used to craft malicious links.

Thanks, that's a good point! I've taken a stab at it [here](https://github.com/plausible/analytics/pull/4916/commits/2aced045dde3126df705567f69b0032dab329763). A few points about this commit:
* It changes the `Phoenix.Controller.redirect` function call to use `to: ` argument over `external: ` which only allows local redirects
* When an external URL is provided to `redirect` it will crash with a 500. Should be rare but still bad UX that I wanted to avoid. So I added a small validation our side as well which will ignore `return_to` unless it starts with a `"/"`
* Note that it's still possible to craft a malicious URL that starts with a double slash `"//"`, and in that case we will 500 but I don't expect this to happen often if at all
* @zoldar is there any reason we've been using `external` over `to? I'm not sure why we would need to redirect to an external URL here...

zoldar:
@ukutaht 

Ad 5. this was originally a workaround for issue with internal redirects failing when there's a url encoded UTF8 character in the path (the domain of the site, in particular): https://github.com/plausible/analytics/pull/3560 - phoenix was validating the "to" URLs a bit overzealously. The phoenix version we currently are on (1.7.14) has this behavior changed via https://github.com/phoenixframework/phoenix/commit/4ca6ca200973c940a7e412404e735fba9d56640e. This should no longer be an issue and we should be able to get back to using internal "to" redirects.

apata:
I'm happy with that, @ukutaht! But because this feature allows redirecting to a page that has itself url search params (https://pr-4916.review.plausible.io/login?return_to=/sites%3Ffoo%3Dbar), any other open redirect vulnerability that we might have would still be exploitable. Before we release it, we should double check that we don't have any.

ukutaht:
Thanks both @zoldar @apata. If it's good to go from your side @apata please change your review to approved so this can be merged :)

apata (approved):

apata:
Sorry for the wait, @ukutaht! I think I was actually expecting to see a fix on the failing CI checks, but then I forgot about the whole thing for the holidays.

PR #4920, opened 2024-12-18
Remove old `Plausible.Auth.Invitation` references

ruslandoga:
This -- together with `Membership.t()` -- fails CI for some of my PRs: https://github.com/plausible/analytics/actions/runs/12373633123/job/34534378559

I'll open a separate PR for `Membership.t()`

### DIFF
--- a/lib/plausible/site/memberships/create_invitation.ex
+++ b/lib/plausible/site/memberships/create_invitation.ex
@@ -4,7 +4,7 @@ defmodule Plausible.Site.Memberships.CreateInvitation do
   transfers.
   """
 
-  alias Plausible.Auth.{User, Invitation}
+  alias Plausible.Auth.User
   alias Plausible.Site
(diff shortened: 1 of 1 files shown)


ruslandoga:
Seems like Dialyzer in CI is broken as it doesn't catch 

```
lib/plausible/site/memberships/accept_invitation.ex:36:33:unknown_type
Unknown type: Plausible.Site.Membership.t/0.
________________________________________________________________________________
```

which is still  a problem in this PR.

zoldar (approved):
TIL about ability to split PLT ❤️ Definitely worth considering. I'll add a housekeeping note on this.

ruslandoga:
> Definitely worth considering. I'll add a housekeeping note on this.

It might make sense to disable Dialyzer until it's done. Right now it provides false assurance.

PR #4921, opened 2024-12-18
Remove old `Plausible.Site.Membership` references

ruslandoga:
Please see https://github.com/plausible/analytics/pull/4920 for context.

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -376,6 +376,7 @@ defmodule Plausible.Sites do
 
   defp owned_sites_query(user) do
     from(s in Site,
+      # TODO that schema doesn't exist
       join: sm in Site.Membership,
       on: sm.site_id == s.id,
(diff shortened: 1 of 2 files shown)


lib/plausible/sites.ex
```diff
@@ -376,6 +376,7 @@ defmodule Plausible.Sites do
 
   defp owned_sites_query(user) do
     from(s in Site,
+      # TODO that schema doesn't exist
       join: sm in Site.Membership,
```
ruslandoga, line 380: This seems to be used in HelpScout integration.

ruslandoga: Ah, seems like these functions are abandoned. The Teams ones are being used instead now.

ruslandoga: https://github.com/plausible/analytics/pull/4921/commits/af5f75661ef75dce3058d6e1b4806294f080921c

PR #4925, opened 2024-12-18
Backfill teams - CE

aerosol:


lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,661 @@
...
+        select: %{
+          email: si.email,
+          role: si.role,
+          invitation_id: si.invitation_id,
+          inserted_at: si.inserted_at,
+          updated_at: si.updated_at,
+          site: s,
+          inviter: inv
+        }
+        # preload: [inviter: inv, site: s]
```
zoldar, line 404: 🔥

zoldar (approved):

aerosol:
@ruslandoga the belief here is, running this (with `dry_run?: false`) should be the only step required for self-hosters to update to current master. Would you mind giving it a look? Thanks

ruslandoga:
👋 

Could you please summarise the changes here? I don't know what to be looking at :)

lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,660 @@
...
+    dry_run? = Keyword.get(opts, :dry_run?, true)
+
+    # Teams backfill
+    db_url =
+      System.get_env(
+        "TEAMS_MIGRATION_DB_URL",
+        Application.get_env(:plausible, Plausible.Repo)[:url]
+      )
+
+    @repo.start(db_url, pool_size: 2 * @max_concurrency)
```
ruslandoga, line 29: Can this be reduced? Some free-tier / entry level hosted PostgreSQL instances limit the pool size to 25.

lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,660 @@
...
+
+    backfill(dry_run?)
+  end
+
+  defp backfill(dry_run?) do
+    # Orphaned teams
+
+    orphaned_teams =
+      from(
+        t in Plausible.Teams.Team,
```
ruslandoga, line 39: Since CE can run this migration at any point in the future, would it be safer to hard code table names at the moment of the next CE release?

ruslandoga: I think we have already had a similar problem with sites.

lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,660 @@
...
+
+        log(
+          "Max sites: #{Enum.max_by(grouped, fn {_, sites} -> length(sites) end) |> elem(1) |> length()}"
+        )
+
+      _ ->
+        :pass
+    end)
+    |> Enum.with_index()
+    |> Task.async_stream(
```
ruslandoga, line 442: Does it have to be async?

lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,660 @@
...
+    |> Enum.with_index()
+    |> Enum.each(fn {{guest_invitation, site_invitation}, idx} ->
+      guest_invitation
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.put_change(:role, translate_role(site_invitation.role))
+      |> Ecto.Changeset.put_change(:invitation_id, site_invitation.invitation_id)
+      |> Ecto.Changeset.put_change(:updated_at, guest_invitation.updated_at)
+      |> @repo.update!()
+
+      if rem(idx, 1000) == 0 do
```
ruslandoga, line 622: It can probably be lowered for CE. Or just print each site that's being updated right now.

lib/plausible/data_migration/backfill_teams.ex
```diff
@@ -0,0 +1,660 @@
...
+        "TEAMS_MIGRATION_DB_URL",
+        Application.get_env(:plausible, Plausible.Repo)[:url]
+      )
+
+    @repo.start(db_url, pool_size: 2 * @max_concurrency)
+
+    backfill(dry_run?)
+  end
+
+  defp backfill(dry_run?) do
```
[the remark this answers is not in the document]
ruslandoga: Otherwise it seems like if it fails half-way, it would always fail in the future due to insert conflicts.

ruslandoga:
Btw, does it mean the Teams is ready to be released and I don't need to cherry-pick v2.1.5? Or should we still wait? If we should wait, then I think it's better to merge this migration right before the CE release to avoid problems with the schemas.

lib/plausible/data_migration/backfill_teams.ex
aerosol: Yes, this is adapted from what we've been running on prod.

lib/plausible/data_migration/backfill_teams.ex
aerosol: Because it's compiled within `lib/`, I think we'll get warnings sooner?

lib/plausible/data_migration/backfill_teams.ex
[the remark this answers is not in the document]
aerosol: The only reason for multiple transactions is to speed up execution time for many sites. Any specific insert conflicts you're predicting? We've been re-running it successfully. It will only catch up with records that require attention on subsequent executions. Setting up the pool size to one should make it possible to wrap in a single transaction, even with parallel `Task`s spawned I think.

lib/plausible/data_migration/backfill_teams.ex
aerosol: Only if we want to utilize the pool, shouldn't matter for small CE instances. I think we can remove any parallel processing then.

lib/plausible/data_migration/backfill_teams.ex
aerosol: Absolutely :+1:

aerosol:
> I think this can stay open until the next full CE release.

Great, thanks. Roughly, when will that be?

aerosol:
> Could you please summarise the changes / possible problems here? I don't know what to be looking out for :)

Gah I'm sorry for not providing much context.
We're working on introducing Teams to Plausible, that will, among other things, change how memberships and subscriptions are managed. That significantly affects the underlying data model. For users that are sole guests, or sole owners, not much should change even at the UI level, so we've been trying to remap data from the old model to the new one gradually, without disruptions, before actually implementing any of the new features. So we've switched all the reads first (to use backfilled/mirrored data) and then started writing to those new tables exclusively, silently deprecating the old schema. For CE it's hopefully much simpler: shutdown, run quick backfill, bring it back up with upgraded code.

lib/plausible/data_migration/backfill_teams.ex
ruslandoga: I don't think there would be any warnings.

lib/plausible/data_migration/backfill_teams.ex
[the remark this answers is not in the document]
ruslandoga: Any of the `insert!` `delete!` etc. calls can raise and stop the execution thus leaving the DB in a potentially inconsistent state, where some records have already been inserted and inserting them again would cause a conflict. I don't know the specifics of this migration so I can't immediately provide examples. But simply wrapping it in a single transaction is easy and safe. So I suggest just using transactions.

lib/plausible/data_migration/backfill_teams.ex
ruslandoga: > I think we can remove any parallel processing then.

Let's do this, I think it would be easier to understand and debug if it goes wrong for one of the self-hosters.

lib/plausible/data_migration/backfill_teams.ex
aerosol: > I think we have already had a similar problem with sites.

Sorry, what problem exactly?

ruslandoga: I cannot find it but basically there was a similar query that was using a schema and no `select` and it was failing in CE because some column in the Ecto schema didn't yet exist in the database because the migration for that column was being run afterwards. I think you were the one who fixed it by adding an explicit `select:` :)

ruslandoga: The problem with these is that they only warn / raise at runtime.

ruslandoga: Found it: https://github.com/plausible/analytics/pull/4459#issuecomment-2317152294

lib/plausible/data_migration/backfill_teams.ex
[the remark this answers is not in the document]
zoldar: This backfill procedure was specifically designed in a away where it can be resumed at any point. Operation where consistency needs to be ensured across multiple db operations are wrapped up in a transaction. We can of course wrap it all up in a single transaction as well, if you prefer.

lib/plausible/data_migration/backfill_teams.ex
zoldar: I wonder if Elixir 1.18 would surface this at compile time 🤔

lib/plausible/data_migration/backfill_teams.ex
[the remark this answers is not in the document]
ruslandoga: Yes, let's wrap it in a single transaction and remove async opera.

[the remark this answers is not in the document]
ruslandoga: If it would be easier, I can make the changes I would want to see and then re-request review from the core team to make sure the intended logic is preserved.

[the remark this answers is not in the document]
aerosol: @ruslandoga oh my, just sat down to this PR but then saw your comment. Very much appreciated.

ruslandoga (approved):

ruslandoga:
Seems like the migration fails due to missing https://github.com/plausible/analytics/pull/4991

PR #4926, opened 2024-12-18
Redo migration:

RobertJoonas:
### Changes

This PR adds a migration that drops the currently unused `imported_pages.scroll_depth` column, and re-adds it with a different type.

UInt8 is good for representing an average value on an event basis, but not for already aggregated data. The new type is Int64 because:

* we still need to represent a missing value (this will be -1, hence we'll be using `Int` rather than `UInt`)
* the new value will represent a sum across all visitors, which makes it easier to calculate in the "merge_imported" queries.

### DIFF
--- a/priv/ingest_repo/migrations/20241218102326_change_scroll_depth_type_in_imported_pages.exs
+++ b/priv/ingest_repo/migrations/20241218102326_change_scroll_depth_type_in_imported_pages.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.IngestRepo.Migrations.AddScrollDepthToImportedPages do
+defmodule Plausible.IngestRepo.Migrations.DropAndAddScrollDepthToImportedPages do
   use Ecto.Migration
 
   @on_cluster Plausible.MigrationUtils.on_cluster_statement("imported_pages")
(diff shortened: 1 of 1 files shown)


priv/ingest_repo/migrations/20241218102326_change_scroll_depth_type_in_imported_pages.exs
```diff
@@ -0,0 +1,27 @@
...
+    execute """
+    ALTER TABLE imported_pages
+    #{@on_cluster}
+    DROP COLUMN scroll_depth
+    """
+
+    execute """
+    ALTER TABLE imported_pages
+    #{@on_cluster}
+    ADD COLUMN scroll_depth Int64 DEFAULT -1
```
apata, line 16: WDYT about naming it scroll_depth_sum to prevent confusion between this and the one in our other models?

RobertJoonas: Thought about it, but there's also time on page which is a sum as well, and doesn't have any indication in the name either, so I thought that would be consistent.

PR #4932, opened 2024-12-31
Average Scroll Depth Metric: another imported schema migration

RobertJoonas:
### Changes

When merging imported data into stats queries, we cannot simply divide `scroll_depth` / `visitors` to get the imported scroll depth. That’s wrong because `visitors` is based on pageviews, not pageleaves. Therefore, we’ll need to export an additional metric - `pageleave_visitors` which we can use specifically for that calculation.

Note: This field should also be used for our new time on page in the future

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
Is this for importing data from another Plausible instance, GA4, or both?

RobertJoonas:
> Is this for importing data from another Plausible instance, GA4, or both?

Only for importing from another Plausible instance.

ukutaht (approved):

PR #4933, opened 2024-12-31
Fix docker build + FE test

RobertJoonas:
### Changes

* Hotfix for resolving a mismatch between NPM and Node JS versions in `build-private-images.yml`.
* Fix a FE test that started failing due to year change from 2024 to 2025.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,7 @@ RUN mkdir /app
 WORKDIR /app
 
 # install build dependencies
-RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli && \
-  npm install npm@latest -g
+RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli
 
 COPY mix.exs ./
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -20,8 +20,7 @@ RUN mkdir /app
 WORKDIR /app
 
 # install build dependencies
-RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli && \
-  npm install npm@latest -g
+RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli
```
apata, line 23: As a hotfix, what you propose seems good!
However, I think best practice would be to build also in the container with the version we're using locally (as defined in .tool-versions and read in various worfklows using the version output by marocchino/tool-versions-action).

apata (approved):

Dockerfile
RobertJoonas: Makes sense! In that case we'll need to install node from a source other than apk. I'll merge this one to unblock others and find a solution to use the node version from .tool-versions.

PR #4942, opened 2025-01-06
Install nodejs specced in tool-versions in Dockerfile

RobertJoonas:
### Changes

Follow-up to https://github.com/plausible/analytics/pull/4933.

We'll use `nvm` to install the same version of nodejs as specified in the `.tool-versions` file. Also, rather than always installing the latest version of NPM (as it was before #4933), nvm will install the one that's bundled with the specific nodejs version (the same happens via `asdf install`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,10 @@ RUN mkdir /app
 WORKDIR /app
 
 # install build dependencies
-RUN apk add --no-cache git yarn python3 ca-certificates wget gnupg make gcc libc-dev brotli bash \
+RUN apk add --no-cache git yarn python3 ca-certificates wget gnupg make gcc g++ libc-dev brotli bash \
   && wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash \
+  && export NVM_DIR="$HOME/.nvm" \
+  && [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,7 +23,11 @@ RUN mkdir /app
 WORKDIR /app
 
 # install build dependencies
-RUN apk add --no-cache git nodejs yarn python3 npm ca-certificates wget gnupg make gcc libc-dev brotli
+RUN apk add --no-cache git yarn python3 ca-certificates wget gnupg make gcc g++ libc-dev brotli bash \
+  && wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash \
+  && export NVM_DIR="$HOME/.nvm" \
+  && [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" \
+  && nvm install ${NODE_VERSION}
```
apata, line 30: Why not skip the middleman and wget the version we know we want directly from https://nodejs.org/dist/?

RobertJoonas: The problem is that all those distributions are incompatible with Alpine Linux as it's using [musl libc](https://musl.libc.org). Musl-compatible builds have been [requested before](https://github.com/nodejs/build/issues/1140), but it looks like so far there are only unofficial ones available.

RobertJoonas:
Closing in favour of https://github.com/plausible/analytics/pull/4945

PR #4944, opened 2025-01-07
Teams setup schema

aerosol:
Once https://github.com/plausible/analytics/pull/4943 is merged

### DIFF
--- a/lib/plausible/teams/team.ex
+++ b/lib/plausible/teams/team.ex
@@ -25,6 +25,9 @@ defmodule Plausible.Teams.Team do
     field :accept_traffic_until, :date
     field :allow_next_upgrade_override, :boolean, default: false
 
+    field :setup_complete, :boolean, default: false
+    field :setup_at, :naive_datetime
+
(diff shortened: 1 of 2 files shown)


zoldar (approved):

lib/plausible/teams/team.ex
```diff
@@ -25,6 +25,9 @@ defmodule Plausible.Teams.Team do
     field :accept_traffic_until, :date
     field :allow_next_upgrade_override, :boolean, default: false
 
+    field :setup_complete, :boolean, default: false
```
ukutaht, line 28: Question: what's the purpose of the boolean if there's also a timestamp? Would the presence of the timestamp, or lack thereof, not provide the same information as the boolean?

aerosol: Boolean is straightforward to match, no dealing with null tri-state. Timestamp is just for troubleshooting/auditing purposes for now and might go away in the future.

aerosol: Another way of looking at it: the app never cares about the actual date specifics.

PR #4945, opened 2025-01-07
[WIP] Pin nodejs version to be the same as .tool-versions

macobo:
Proposed follow-up to https://github.com/plausible/analytics/pull/4933

Alpine index only exposes a single version of nodejs, which this PR now explicitly pins against. See also: https://pkgs.alpinelinux.org/packages?name=nodejs-current&branch=v3.20&repo=&arch=x86_64&origin=&flagged=&maintainer=

### DIFF
--- a/.github/workflows/build-private-images-ghcr.yml
+++ b/.github/workflows/build-private-images-ghcr.yml
@@ -61,14 +61,14 @@ jobs:
         run: echo ${{ steps.docker_build.outputs.digest }}
 
       - name: Notify team on failure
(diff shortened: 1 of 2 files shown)


RobertJoonas (approved):
Wow, I should have searched Alpine packages by `nodejs*` 🙈

This sounds like a good fix to me, thanks! Whoever updates node in the future needs to be mindful anyway, so not a big deal if it's not automatically read from the .tool-versions.

cnkk (approved):

PR #4947, opened 2025-01-07
Upgrade deps, add storybook

ukutaht:
### Changes

Adds [phoenix storybook](https://github.com/phenixdigital/phoenix_storybook) dependency along with example pages for dropdowns and basic inputs. In order to make it work I had to upgrade some deps:
* phoenix_ecto to 4.5
* phoenix_html to 4.1
* phoenix_live_view to 1.0

There were some breaking changes which I'll take notes on in self-review

There are a couple known issues with storybook installation:
1. The storybook area (`./storybook`) uses its own JS bundle which is using pre-1.0 version of liveview. This causes a console warning but no real-world issues as far as I can tell. A new version of storybook should be released this week with support for Liveview 1.0
2. For some reason, when you navigate to a page with Alpine.JS component (Input -> Dropdown) it causes some error from AlpineJS. When you load the page with AlpineJS directly it works fine. I wasn't able to find a fix.


lib/plausible/billing/ecto/feature.ex
```diff
@@ -44,35 +44,4 @@ defmodule Plausible.Billing.Ecto.FeatureList do
   def cast(list), do: Ecto.Type.cast(type(), list)
   def load(list), do: Ecto.Type.load(type(), list)
   def dump(list), do: Ecto.Type.dump(type(), list)
```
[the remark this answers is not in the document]
ukutaht: Yes my bad, refactored in https://github.com/plausible/analytics/pull/4947/commits/ef692ed3736519c58f1f48f5d478ef7533cacc61

lib/plausible/billing/ecto/limit.ex
```diff
@@ -19,22 +19,4 @@ defmodule Plausible.Billing.Ecto.Limit do
 
   def dump(:unlimited), do: {:ok, -1}
   def dump(other), do: Ecto.Type.dump(:integer, other)
```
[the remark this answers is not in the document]
ukutaht: https://github.com/plausible/analytics/pull/4947/commits/ef692ed3736519c58f1f48f5d478ef7533cacc61

aerosol (commented):
Cool work :clap:. 

CI failures aside, I think this could benefit from being a `MIX_ENV=dev`-only dependency/runtime. Or do you intentionally want to run it on prod servers? IMHO it's unnecessary build clutter for prod images.

I've tried looking into the Alpine misbehavior, but Alpine-specific setup you did seems a-OK.

ukutaht:
> CI failures aside

About that - the new HeeX engine wants us to convert all the templates from `<%= var %>` to `{var}` with `mix format`. However this will make the PR much harder to review. I was thinking I'd wait to get reviews and push the `mix format` change at the last minute before merging.

> I think this could benefit from being a MIX_ENV=dev-only dependency/runtime. Or do you intentionally want to run it on prod servers? IMHO it's unnecessary build clutter for prod images.

Yeah it feels like a dev thing to me as well but there were a couple points that made me include it in docker builds:
* I was thinking it would be nice to make the components easily accessible for Liina at https://plausible.io/storybook
* And also to be able to see in a preview environment when components are worked on and how they differ from prod: https://pr-4947.review.plausible.io/storybook/dropdown

Maybe just including in staging would be a decent compromise for now. But since preview builds are also made with `MIX_ENV=prod` it would only be a runtime exclude so not really address the 'build clutter' issue which I agree with. BTW I found that we have a couple instances in the codebase checking for `MIX_ENV=staging` which, as far as I can tell, is never the case.

Anyways - I definitely think having storybook in preview environments is useful but I'm agnostic about prod. Given all that - do you think we should exclude from prod at runtime?

apata:
I think being able to view the storybook without running the app locally is an important part for its adoption: when following the storybook is inconvenient, it's unlikely to help in unifying user experience within the app.

That said, I'd also be wary of exposing it from prod servers: as I see it, the increase in attack surface / ways for our prod servers to fail is not worth the convenient plausible.io/storybook URL. Have it on staging?

storybook/input.story.exs
```diff
@@ -0,0 +1,46 @@
+defmodule MyAppWeb.Storybook.Input do
+  use PhoenixStorybook.Story, :component
+
+  # required
+  def function, do: &PlausibleWeb.Live.Components.Form.input/1
```
apata, line 5: In dark mode, the input text is black in the storybook, even though it correctly renders as off-white in Settings views. Poked around a bit, but I couldn't figure out what style is missing. FWIW, in both places they are affected by a tailwind reset, which sets the color to inherit.

Screenshot 2025-01-09 at 16 35 42

ukutaht:
I've pushed the large `mix format` [commit](https://github.com/plausible/analytics/pull/4947/commits/a1b98687ac7fbc37ca269f458290a28db770adb1).

PS: Merging this will likely cause conflicts with other PRs that change templates

storybook/input.story.exs
apata: That indeed makes the text the same both in the storybook and app, but doesn't address the root issue: storybook component container style is not consistent with app root style, causing this and potentially other differences. 

Ideally, we wouldn't need to add styles to components that already look good in the app, just to get them to look right in the storybook. 

Doesn't need to be solved right now, but worth making a note of somewhere.

config/runtime.exs
```diff
@@ -969,3 +969,5 @@ unless s3_disabled? do
     exports_bucket: s3_env_value.("S3_EXPORTS_BUCKET"),
     imports_bucket: s3_env_value.("S3_IMPORTS_BUCKET")
 end
+
+config :phoenix_storybook, enabled: env !== "prod"
```
apata, line 973: nitpick: There's nothing to verify that this is in effect. Just to be safe, can we make storybook routes conditional in the router as well?

ukutaht: > Just to be safe, can we make storybook routes conditional in the router as well?

No. The router macros are used at compile-time while the `ENVIRONMENT=staging` is a runtime config variable.

storybook/input.story.exs
ukutaht: Agree overall. I went with the band-aid because I was unable to find the root cause of the inconsistency. I would happily fix the consistency if I knew how to. Open for suggestions.

apata (approved):
All seems to work as expected in the review env.

PR #4957, opened 2025-01-10
shorten the alert

metmarkosaric:
Bogdan mentioned that this alert gets cuts in a browser as there's not enough space to display it in full so here's my attempt at shortening it


zoldar:
There's a tooltip with a clarification instead now

image



I think we can close it then @metmarkosaric ?

PR #4958, opened 2025-01-10
Handle non-string binary input before postgres

apata:
### Changes

Stops internal server errors stemming from binary but non-string input reaching postgres queries.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugs/authorize_site_access.ex
+++ b/lib/plausible_web/plugs/authorize_site_access.ex
@@ -128,8 +128,12 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     end
   end
 
+  defp valid_path_fragment?(fragment), do: is_binary(fragment) and String.valid?(fragment)
+
   defp get_domain(conn, nil) do
-    if domain = conn.path_params["domain"] do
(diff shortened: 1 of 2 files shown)


lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -128,8 +128,12 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     end
   end
 
+  defp valid_path_fragment?(fragment), do: is_binary(fragment) and String.valid?(fragment)
```
aerosol, line 131: when is this not binary?

apata: It won't be binary when Plugs.AuthorizeSiteAccess is configured to use any param except the default path param `:domain`, and a non-binary value is given in `conn.params[:site_param]`. 

The question this PR attempts to provide an answer for is what to do when it is binary but non-string, like https://plausible.io/%c0%ae. I propose that it should behave the same way as https://plausible.io/invalid.

ukutaht (approved):

PR #4964, opened 2025-01-13
Average Scroll Depth Metric: Track custom props on pageleaves

RobertJoonas:
### Changes

_NOTE: Reviewing this PR commit by commit is highly recommended._

The main intent of this PR is to make scroll depth data available also in queries where a custom property dimension is used. That means that we'll start tracking custom props on pageleaves as well. The implementation is fairly straight forward - simply keep track of the props that were sent with the pageview and send them along with the pageleave as well.

In addition, this PR:

* Fixes an important bug with tracking pageleaves on ignored pages (a test asserting the broken behaviour was accidentally passing due to pageleaves being throttled)
* Refactors the `currentPageleaveURL` logic - I overcomplicated things in the beginning. 
* Refactors the tracker test suite quite a bit

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -12,9 +12,15 @@
   var endpoint = scriptEl.getAttribute('data-api') || defaultEndpoint(scriptEl)
   var dataDomain = scriptEl.getAttribute('data-domain')
 
-  function onIgnoredEvent(reason, options) {
+  function onIgnoredEvent(eventName, reason, options) {
     if (reason) console.warn('Ignoring Event: ' + reason);
     options && options.callback && options.callback()
+
+    {{#if pageleave}}
+    if (eventName === 'pageview') {
+      currentPageLeaveIgnored = true
+    }
+    {{/if}}
   }
 
   function defaultEndpoint(el) {
@@ -31,6 +37,7 @@
   {{#if pageleave}}
   // :NOTE: Tracking pageleave events is currently experimental.
 
+  var currentPageLeaveIgnored
   var currentPageLeaveURL = location.href
   var currentPageLeaveProps = {}
 
@@ -90,7 +97,7 @@
   })
 
   function triggerPageLeave() {
-    if (pageLeaveSending) {return}
+    if (pageLeaveSending || currentPageLeaveIgnored) {return}
     pageLeaveSending = true
     setTimeout(function () {pageLeaveSending = false}, 500)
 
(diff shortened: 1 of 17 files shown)


tracker/src/plausible.js
```diff
@@ -220,6 +222,9 @@
       if (request.readyState === 4) {
```
macobo, line 222: Now we're revisiting this: Is this the right place to trigger page leave listeners? 

Imagine scenario with slooow connection to plausible.io:
1. t=0ms user opens page `/`
2. t=5ms `pageview` gets triggered, sent to us
3. t=1000ms user clicks, navigates to `/foo` via history (no full page reload). New request gets triggered
4. t=2000ms plausible events API finally responds with success for step 2. We register `registerPageLeaveListener` 
5. t=3000ms plausible events API finally responds with success for step 3.

We would only send 1 pageleave for /foo in this scenario right? Shouldn't we register the listener immediately?

tracker/test/support/test-utils.js
```diff
@@ -44,33 +50,22 @@ const mockManyRequests = function(page, path, numberOfRequests) {
...
+ * requests that should or should not have been made after doing a page
+ * action (e.g. navigating to the page, clicking a page element, etc). 
  * 
  * This function accepts subsets of request bodies (the JSON payloads) as
  * arguments, and compares them with the bodies of the requests that were
  * actually made. For a body subset to match a request, all the key-value
  * pairs present in the subset should also appear in the request body.
  */
-exports.clickPageElementAndExpectEventRequests = async function (page, locatorToClick, expectedBodySubsets, refutedBodySubsets = []) {
+exports.pageActionAndExpectEventRequests = async function (page, pageActionFn, expectedBodySubsets, refutedBodySubsets = [], eventsToAwait = null) {
```
macobo, line 63: Calls of this are slightly hard to follow. On our call tomorrow let's spend a minute to see if we can tweak the signature to be a bit easier to follow when called.?

First thought: Make this only take a single object as an argument and call it with something like:

```js
pageActionAndExpectEventRequests({
  action: () => page.click('#navigate-away'),
  expectRequests: [{n: 'pageview'}],
  page
})
```

We might be able to shorten the request this way.

macobo (commented):
In general this is looking great!

However I'm slightly suspicious of the general pageleave lifecycle - let's pair tomorrow and regain the confidence.

tracker/src/plausible.js
macobo: Will take a look at this as part of https://3.basecamp.com/5308029/buckets/39034214/card_tables/cards/8210482322

tracker/test/support/test-utils.js
RobertJoonas: https://github.com/plausible/analytics/pull/4964/commits/92912c343a028b3b2325beac88efd07cd635f18a

macobo (approved):

PR #4969, opened 2025-01-14
Revert "Upgrade deps, add storybook (#4947)"

ukutaht:
This reverts commit 6c30f62d5d5f07c93f02328f0643b93c2ef6bcb4.


apata (approved):
This works for me locally, approving! However, as an alternative, I submit a forward patch https://github.com/plausible/analytics/pull/4971, which I believe to have lower risk and cost overall.

PR #4972, opened 2025-01-15
Remove channels feature flag

ukutaht:


macobo:
Channels are not currently mentioned in the changelog. We should also add an entry there.

macobo (approved):
LGTM though changelog should likely get updated

ukutaht:
Good point, thank you. I've added a changelog

macobo (approved):
:shipit:

PR #4974, opened 2025-01-15
Keep caches alive in case of postgres unavailability

aerosol:
This wraps cache refreshes with a pokemon catch, so that whatever happens with postgres, the cache supervision tree remains stable.


zoldar (approved):

RobertJoonas (approved):
👍

PR #4980, opened 2025-01-16
APIv2: Behavioral filtering

macobo:
### Changes

This PR adds support for behavioral filtering by adding two new operators `has_done` and `has_not_done` to APIv2.

These operators can be used to filter whether users have completed (or not completed) a given at any point in their session. This can be used to exclude goals or check whether a user has done an additional action in their session.

In the following example, we could pageviews and visitors on /pricing page, but only by users who _have_ opened support chat but have not registered or logged in either before or after visiting the page.

```json
{
  "site_id": "dummy.site",
  "metrics": ["visitors", "pageviews"],
  "date_range": "7d",
  "filters": [
    ["is", "event:page", ["/pricing"]],
    ["has_done", ["is", "event:goal", ["Open Support Chat"]]],
    ["has_not_done", ["is", "event:goal", ["Register", "Login"]]]
  ]
}

```

~~Docs PR: https://github.com/plausible/docs/pull/575~~ The operators are currently internal-only.
Basecamp ref: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/7793417773

## What's to be done

I'll open another PR to add support for users checking whether user has not done a goal. This turned out to be a bigger rabbit hole due to UI issues hence the split.

### DIFF
--- a/lib/plausible/goals/filters.ex
+++ b/lib/plausible/goals/filters.ex
@@ -4,6 +4,31 @@ defmodule Plausible.Goals.Filters do
   import Ecto.Query
   import Plausible.Stats.Filters.Utils, only: [page_regex: 1]
 
+  alias Plausible.Stats.Filters
+
+  @doc """
+  Preloads goals data if needed for query-building and related work.
+  """
+  def preload_needed_goals(site, dimensions, filters) do
+    if Enum.member?(dimensions, "event:goal") or
+         Filters.filtering_on_dimension?(filters, "event:goal") do
+      goals = Plausible.Goals.for_site(site)
+
+      %{
+        # When grouping by event:goal, later pipeline needs to know which goals match filters exactly.
+        # This can affect both calculations whether all goals have the same revenue currency and
+        # whether we should skip imports.
+        matching_toplevel_filters: goals_matching_toplevel_filters(goals, filters),
+        all: goals
+      }
+    else
+      %{
+        all: [],
+        matching_toplevel_filters: []
+      }
+    end
+  end
+
   @doc """
   Translates an event:goal filter into SQL. Similarly to other `add_filter` clauses in
   `Plausible.Stats.SQL.WhereBuilder`, returns an `Ecto.Query.dynamic` expression.
@@ -26,7 +51,7 @@ defmodule Plausible.Goals.Filters do
 
     Enum.reduce(clauses, false, fn clause, dynamic_statement ->
       condition =
-        query.preloaded_goals
+        query.preloaded_goals.all
         |> filter_preloaded(filter, clause)
         |> build_condition(imported?)
 
@@ -34,9 +59,11 @@ defmodule Plausible.Goals.Filters do
     end)
   end
 
-  def preload_needed_goals(site, filters) do
-    goals = Plausible.Goals.for_site(site)
+  defp filter_preloaded(goals, filter, clause) do
+    Enum.filter(goals, fn goal -> matches?(goal, filter, clause) end)
+  end
 
+  defp goals_matching_toplevel_filters(goals, filters) do
     Enum.reduce(filters, goals, fn
       [_, "event:goal" | _] = filter, goals ->
         goals_matching_any_clause(goals, filter)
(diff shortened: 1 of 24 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1625,6 +1629,10 @@ defmodule PlausibleWeb.Api.StatsController do
 
   defp realtime_period_to_30m(params), do: params
 
+  defp toplevel_goal_filter?(query) do
+    Filters.filtering_on_dimension?(query, "event:goal", max_depth: 0)
+  end
```
apata, line 1634: General question, can be taken offline: why is this a concern, but not `event:name`?

lib/plausible/goals/filters.ex
```diff
@@ -4,6 +4,31 @@ defmodule Plausible.Goals.Filters do
...
+    if Enum.member?(dimensions, "event:goal") or
+         Filters.filtering_on_dimension?(filters, "event:goal") do
+      goals = Plausible.Goals.for_site(site)
+
+      %{
+        # When grouping by event:goal, later pipeline needs to know which goals match filters exactly.
+        # This can affect both calculations whether all goals have the same revenue currency and
+        # whether we should skip imports.
+        matching_toplevel_filters: goals_matching_toplevel_filters(goals, filters),
+        all: goals
```
ukutaht, line 22: Where is `all` used and why?

lib/plausible_web/controllers/api/stats_controller.ex
macobo: So the dashboard has toggles for behavior based on if there's a goal filter or not. Mostly for checking whether to query/show conversion_rate metric. This needed to change since conversion_rate is not calculatable with a behavioral goal filter (e.g. has not done some goal).

event:name and other event dimensions don't have this sort of special-casing.

lib/plausible/goals/filters.ex
macobo: Good question. 

I needed to change the structure of `preloaded_goals` since we need to know what goals match the top-level is/contains filter for group by and other logics as it will affect what gets displayed (e.g. do all have the same currency, whether we can include imports. For this reason `matching_toplevel_filters` exists.

Behavioral filters `has_done` and `has_not_done` do not impact these bits of business logic but we still need to know what condition(s) these goals have when building the Ecto Query. Hence the `all` key. This gets used only in this same file when building the Ecto query:

```elixir
    Enum.reduce(clauses, false, fn clause, dynamic_statement ->
      condition =
        query.preloaded_goals.all
        |> filter_preloaded(filter, clause)
        |> build_condition(imported?)

      dynamic([e], ^condition or ^dynamic_statement)
    end)
```

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -472,16 +474,52 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
+  defp validate_behavioral_filters(query) do
```
apata, line 477: **suggestion, nonblocking:** Since this doesn't use anything else other than query.filters as input, I think it would be better if filters was the argument. That way, we could doctest it.

macobo: We would need tests at a unit (parser) level anyways to be confident in the parsing for completeness and I am against exposing 'private' methods just for testing. Given this module is the entry point into the system, I'm trying to maintain really high coverage on the whole thing.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -472,16 +474,52 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+    nested_behavioral_filter? =
+      query.filters
+      |> Filters.traverse(0, fn behavioral_depth, operator ->
```
apata, line 480: **nitpick, non-blocking:** In this function, we're traversing filters twice because dimensions_used_in_filters is also calling traverse. In both cases, we're looking for behavioral filters, so potentially unifiable? E.g. in the first traversal exit early when encountering a non-event dimension filter.

lib/plausible/stats/filters/query_parser.ex
apata: From my perspective, a doc and doctest with examples gives immediate info on what the function is about without me needing to execute the function as I'm reading it. 

This is especially valuable if we're looking at a double traversal of a tree-structured property of one of the most complex structs in the repo, with several different possible outcomes, no spec and no pattern matching in arguments.

I think the fact that we have to make the function public is a nuisance, but small price to pay for clarity. No cost, if we'd move the innards of this function to Plausible.Stats.Filters.validate_behavioral_filters(filters) and just wrap it in this module thinly. 

That said, it's definitely mergable as is.

apata (approved):

PR #4982, opened 2025-01-16
Expand segments in queries

apata:
### Changes

This PR introduces functions that handle expanding segments in queries to constituent filters.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/filters.ex
+++ b/lib/plausible/segments/filters.ex
@@ -5,6 +5,8 @@ defmodule Plausible.Segments.Filters do
   alias Plausible.Segments
   alias Plausible.Stats.Filters
 
+  @max_segment_filters_count 10
+
   @doc """
   Finds unique segment IDs used in query filters.
 
@@ -16,8 +18,6 @@ defmodule Plausible.Segments.Filters do
     {:error, "Invalid filters. You can only use up to 10 segment filters in a query."}
   """
   def get_segment_ids(filters) do
-    max_segment_filters_count = 10
-
     ids =
       filters
       |> Filters.traverse()
@@ -26,9 +26,9 @@ defmodule Plausible.Segments.Filters do
         _ -> []
       end)
 
-    if length(ids) > max_segment_filters_count do
+    if length(ids) > @max_segment_filters_count do
       {:error,
-       "Invalid filters. You can only use up to #{max_segment_filters_count} segment filters in a query."}
+       "Invalid filters. You can only use up to #{@max_segment_filters_count} segment filters in a query."}
     else
       {:ok, Enum.uniq(ids)}
     end
--- a/lib/plausible/segments/filters.ex
+++ b/lib/plausible/segments/filters.ex
@@ -1,0 +1,105 @@
+defmodule Plausible.Segments.Filters do
+  alias Plausible.Segments
+  alias Plausible.Stats.Filters
+
+  @doc """
+  Finds unique segment IDs used in query filters.
+
+  ## Examples
+    iex> get_segment_ids([[:not, [:is, "segment", [10, 20]]], [:contains, "visit:entry_page", ["blog"]]])
+    {:ok, [10, 20]}
+
+    iex> get_segment_ids([[:and, [[:is, "segment", Enum.to_list(1..6)], [:is, "segment", Enum.to_list(1..6)]]]])
+    {:error, "Invalid filters. You can only use up to 10 segment filters in a query."}
+  """
+  def get_segment_ids(filters) do
+    max_segment_filters_count = 10
+
+    ids =
+      filters
+      |> Filters.traverse()
+      |> Enum.flat_map(fn
+        {[_operation, "segment", clauses], _depth} -> clauses
+        _ -> []
+      end)
+
+    if length(ids) > max_segment_filters_count do
+      {:error,
+       "Invalid filters. You can only use up to #{max_segment_filters_count} segment filters in a query."}
+    else
+      {:ok, Enum.uniq(ids)}
+    end
+  end
+
+  def preload_needed_segments(%Plausible.Site{} = site, filters) do
+    with {:ok, segment_ids} <- get_segment_ids(filters),
+         {:ok, segments} <-
+           Segments.get_many(
+             site,
+             segment_ids,
+             fields: [:id, :segment_data]
+           ),
+         {:ok, segments_by_id} <-
+           {:ok,
+            Enum.into(
+              segments,
+              %{},
+              fn %Segments.Segment{id: id, segment_data: segment_data} ->
+                case Filters.QueryParser.parse_filters(segment_data["filters"]) do
+                  {:ok, filters} -> {id, filters}
+                  _ -> {id, nil}
+                end
+              end
+            )},
+         :ok <-
+           if(Enum.any?(segment_ids, fn id -> is_nil(Map.get(segments_by_id, id)) end),
+             do: {:error, "Invalid filters. Some segments don't exist or aren't accessible."},
+             else: :ok
+           ) do
+      {:ok, segments_by_id}
+    end
+  end
+
+  defp replace_segment_with_filter_tree([_, "segment", clauses], preloaded_segments) do
+    if length(clauses) === 1 do
+      [[:and, Map.get(preloaded_segments, Enum.at(clauses, 0))]]
+    else
+      [[:or, Enum.map(clauses, fn id -> [:and, Map.get(preloaded_segments, id)] end)]]
+    end
+  end
+
+  defp replace_segment_with_filter_tree(_filter, _preloaded_segments) do
+    nil
+  end
+
+  @doc """
+  ## Examples
+
+    iex> resolve_segments([[:is, "visit:entry_page", ["/home"]]], %{})
+    {:ok, [[:is, "visit:entry_page", ["/home"]]]}
+
+    iex> resolve_segments([[:is, "visit:entry_page", ["/home"]], [:is, "segment", [1]]], %{1 => [[:contains, "visit:entry_page", ["blog"]], [:is, "visit:country", ["PL"]]]})
+    {:ok, [
+      [:is, "visit:entry_page", ["/home"]],
+      [:and, [[:contains, "visit:entry_page", ["blog"]], [:is, "visit:country", ["PL"]]]]
+    ]}
+
+    iex> resolve_segments([[:is, "segment", [1, 2]]], %{1 => [[:contains, "event:goal", ["Singup"]], [:is, "visit:country", ["PL"]]], 2 => [[:contains, "event:goal", ["Sauna"]], [:is, "visit:country", ["EE"]]]})
+    {:ok, [
+      [:or, [
+        [:and, [[:contains, "event:goal", ["Singup"]], [:is, "visit:country", ["PL"]]]],
+        [:and, [[:contains, "event:goal", ["Sauna"]], [:is, "visit:country", ["EE"]]]]]
+      ]
+    ]}
+  """
+  def resolve_segments(original_filters, preloaded_segments) do
+    if map_size(preloaded_segments) > 0 do
+      {:ok,
+       Filters.transform_filters(original_filters, fn f ->
+         replace_segment_with_filter_tree(f, preloaded_segments)
+       end)}
+    else
+      {:ok, original_filters}
+    end
+  end
+end
--- a/lib/plausible/segments/filters.ex
+++ b/lib/plausible/segments/filters.ex
@@ -1,4 +1,7 @@
 defmodule Plausible.Segments.Filters do
+  @moduledoc """
+  This module contains functions that enable resolving segments in filters.
+  """
   alias Plausible.Segments
   alias Plausible.Stats.Filters
 
--- a/lib/plausible/stats/filters/filters.ex
+++ b/lib/plausible/stats/filters/filters.ex
@@ -130,14 +130,11 @@ defmodule Plausible.Stats.Filters do
   to ignore and look deeper.
 
   ## Examples
-
-    iex> Filters.transform_filters([[:is, "visit:entry_page", ["/blog"]]], fn f -> [f] end)
-    [[:is, "visit:entry_page", ["/blog"]]]
-
-    iex> Filters.transform_filters([[:is, "visit:entry_page", ["/blog"]]], fn _f -> nil end)
-    [[[:is, "visit:entry_page", ["/blog"]]]]
-
-
+    iex> Filters.transform_filters([[:is, "visit:os", ["Linux"]], [:and, [[:is, "segment", [1]], [:is, "segment", [2]]]]], fn
+    ...>    [_, "segment", _] -> [[:is, "segment", ["changed"]]]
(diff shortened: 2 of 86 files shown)


lib/plausible/stats/filters/query_parser.ex
```diff
@@ -34,6 +34,9 @@ defmodule Plausible.Stats.Filters.QueryParser do
          utc_time_range = raw_time_range |> DateTimeRange.to_timezone("Etc/UTC"),
          {:ok, metrics} <- parse_metrics(Map.get(params, "metrics", [])),
          {:ok, filters} <- parse_filters(Map.get(params, "filters", [])),
+         {:ok, preloaded_segments} <- preload_needed_segments(site, filters),
+         {:ok, filters} <-
+           resolve_segments(filters, preloaded_segments),
```
macobo, line 39: Q: I remember we had a chat about this but remind me, why can't we do this in QueryOptimizer? Validations?

lib/plausible/stats/filters/filters.ex
```diff
@@ -128,6 +128,16 @@ defmodule Plausible.Stats.Filters do
...
   query and must return a list of nodes to replace it with or nil
   to ignore and look deeper.
+
+  ## Examples
+
+    iex> Filters.transform_filters([[:is, "visit:entry_page", ["/blog"]]], fn f -> [f] end)
+    [[:is, "visit:entry_page", ["/blog"]]]
+
+    iex> Filters.transform_filters([[:is, "visit:entry_page", ["/blog"]]], fn _f -> nil end)
+    [[[:is, "visit:entry_page", ["/blog"]]]]
```
apata, line 138: @macobo This one case has no bearing on the pull request (ended up relying on the transformer from the case above), except that I noticed it when developing and couldn't figure out why it's like this.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -443,6 +452,108 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+  ## Examples
+    iex> get_segment_ids([[:not, [:is, "segment", [10, 20]]], [:contains, "visit:entry_page", ["blog"]]])
+
+    {:ok, [10, 20]}
+
+    iex> get_segment_ids([[:and, [[:is, "segment", Enum.to_list(1..6)], [:is, "segment", Enum.to_list(1..6)]]]])
+
+    {:error, "Invalid filters. You can only use up to 10 segment filters in a query."}
+  """
+  def get_segment_ids(filters) do
```
macobo, line 467: Problem: I really want to keep methods as hidden as possible and the API for `QueryParser` as narrow as possible. Having public methods that are not really public makes all module design kind of pointless. As such we should either move this into a separate module or remove the doctests and make sure parser_tests cover the outward behavior 100%.

lib/plausible/stats/filters/query_parser.ex
apata: From validations, definitely `validate_custom_props_access`, `validate_metrics`, probably others. 

Also, preloading goals wouldn't work as expected. 

As I see it, there's every reason to do it as early as possible to reject any stale (_using segments that don't exist_) and invalid queries (_using incompatible segments, or segments with other filters that they are incompatible with_) as early as possible. 

If the issue is with the expanded segments appearing in query response, we can carry the original_filters in the Query struct to be shown instead.

macobo, line 39: Let's create a new module e.g. `lib/plausible/segments/filters.ex` with `def expand_segments(site, filters)` which returns the expanded segments and move the added code there.

This keeps segments logic "local", allows to add tests/doctests to the interface nicely for special cases and keeps internals of how segments work out of this module. It's also similar to how goals and revenue metrics do things.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -443,6 +452,106 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+    iex> get_segment_ids([[:and, [[:is, "segment", Enum.to_list(1..6)], [:is, "segment", Enum.to_list(1..6)]]]])
+    {:error, "Invalid filters. You can only use up to 10 segment filters in a query."}
+  """
+  def get_segment_ids(filters) do
+    max_segment_filters_count = 10
+
+    ids =
+      filters
+      |> Filters.traverse()
+      |> Enum.filter(fn {[_, filter_key | _rest], _depth} -> filter_key === "segment" end)
```
macobo, line 471: Naming nit: prefer `dimension` over `filter_key`. There's some confusion around this (originally we went with filter_key) but new code should use `dimension`. I'll get around to cleaning this confusion up from the code.

lib/plausible/stats/filters/filters.ex
```diff
@@ -154,7 +164,7 @@ defmodule Plausible.Stats.Filters do
     end
   end
 
-  defp traverse(filters, depth \\ -1) do
+  def traverse(filters, depth \\ -1) do
```
macobo, line 167: Annoyingly, usages of this as-is is going to cause annoying conflicts with https://github.com/plausible/analytics/pull/4980.

To reduce headaches depending on the order of the merge I suggest copying the implementation from that PR and maybe adding a default to state/state_transformer.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -443,6 +452,106 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+      |> Enum.map(fn {[_, _filter_key, clauses], _depth} -> clauses end)
+      |> Enum.concat()
```
macobo, line 473: Suggestion:

```elixir
|> Enum.flat_map(
  {[_operation, "segment", clauses], _depth} -> clauses
  _ -> []
)
```

Not only does it avoid filter + map + concat confusion, but it's also more at-a-glance future-proof should filter structure change.

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -443,6 +452,106 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+    iex> resolve_segments([[:is, "segment", [1, 2]]], %{1 => [[:contains, "event:goal", ["Singup"]], [:is, "visit:country", ["PL"]]], 2 => [[:contains, "event:goal", ["Sauna"]], [:is, "visit:country", ["EE"]]]})
+    {:ok, [
+      [:or, [
+        [:and, [[:contains, "event:goal", ["Singup"]], [:is, "visit:country", ["PL"]]]],
+        [:and, [[:contains, "event:goal", ["Sauna"]], [:is, "visit:country", ["EE"]]]]]
+      ]
+    ]}
+  """
+  def resolve_segments(original_filters, preloaded_segments) do
+    if Map.keys(preloaded_segments) !== [] do
```
macobo, line 545: Nit: `map_size(...) == 0`

lib/plausible/stats/filters/query_parser.ex
macobo: Suggested solution: https://github.com/plausible/analytics/pull/4982#discussion_r1918348174

lib/plausible/stats/filters/filters.ex
macobo: Both examples here are kind of not useful or how the function is intended to be used. Let's remove them - no docs is better than misleading docs.

priv/json-schemas/query-api-schema.json
```diff
@@ -439,11 +437,32 @@
         }
       ]
     },
+    "filter_for_segment": {
```
macobo, line 440: Since this isn't marked as internal, we should also update API documentation

apata: Right! Marked as internal for the development phase.

test/plausible/stats/query_parser_test.exs
```diff
@@ -2,10 +2,11 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
   use Plausible
```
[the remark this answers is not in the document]
apata: Since it's an only :internal value until released, I won't be able to add v2 endpoint tests to this. Ok to do on release? I should test resolving filters on the dashboard /api/stats/:domain/... routes, but I'd like to do it in a separate PR I think: might be a bit of a rabbit hole, since those are still on Query.from.

lib/plausible/stats/filters/query_parser.ex
apata: Done, though I am a bit worried that this creates the situation that any module anywhere can include filters related logic.

lib/plausible/stats/filters/query_parser.ex
ukutaht: Should we start a campaign to be able to use @doc and @doctest in Elixir private functions? 😀

lib/plausible/stats/filters/filters.ex
apata: I refactored to an example of how it's supposed to be used. That informed me that it's likely a bug, and I've put in the fix.

lib/plausible/stats/filters/filters.ex
apata: Can we cross that bridge when we come to it?

lib/plausible/stats/filters/query_parser.ex
apata: @ukutaht The heat of the discussion https://elixirforum.com/t/is-it-necessary-to-warn-about-the-doc-attribute-for-a-private-function/19187 warned me off from any sort of campaigning on that subject.

test/plausible/stats/query_parser_test.exs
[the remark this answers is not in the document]
macobo: APIv2 tests can be added for internal filters by using this route: `/api/v2/query-internal-test`

lib/plausible/stats/filters/filters.ex
macobo: Yup - just wanted to point out an annoyance you might run into trying to get this merged.

test/plausible/stats/query_parser_test.exs
```diff
@@ -2028,4 +2029,149 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
       })
     end
   end
+
+  describe "filtering with segments" do
```
macobo, line 2033: These look gr

Let's also add cases that validate that other validations are done after segment resolving:
1. Custom props are used but the feature is inaccessible
2. conversion_rate metric with event:goal filter coming from segment
3. `views_per_visit` with event:page filter

lib/plausible/segments/filters.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  ## Examples
+    iex> get_segment_ids([[:not, [:is, "segment", [10, 20]]], [:contains, "visit:entry_page", ["blog"]]])
+    {:ok, [10, 20]}
+
+    iex> get_segment_ids([[:and, [[:is, "segment", Enum.to_list(1..6)], [:is, "segment", Enum.to_list(1..6)]]]])
+    {:error, "Invalid filters. You can only use up to 10 segment filters in a query."}
+  """
+  def get_segment_ids(filters) do
+    max_segment_filters_count = 10
```
macobo, line 19: Nitpick: Probably better as a module attribute?

lib/plausible/stats/filters/filters.ex
macobo: Great, thanks for fixing!

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -34,6 +34,10 @@ defmodule Plausible.Stats.Filters.QueryParser do
          utc_time_range = raw_time_range |> DateTimeRange.to_timezone("Etc/UTC"),
          {:ok, metrics} <- parse_metrics(Map.get(params, "metrics", [])),
          {:ok, filters} <- parse_filters(Map.get(params, "filters", [])),
+         {:ok, preloaded_segments} <-
+           Plausible.Segments.Filters.preload_needed_segments(site, filters),
+         {:ok, filters} <-
+           Plausible.Segments.Filters.resolve_segments(filters, preloaded_segments),
```
macobo, line 40: Nitpick, not blocking: Given that `preloaded_segments` is only used in `resolve_segments`, we could combine the two into `Plausible.Segments.Filters.resolve_segments(site, filters)`.

test/plausible/stats/query_parser_test.exs
apata: What if we do a loop in most tests in this file asserting that the result is the same whether the filterset is given within a segment or naked?

macobo: Optimize tests for readability over DRY. I think meta-programming in this case would detract from being easy to read and follow. You are correct that this wouldn't catch _all_ validations but the goal would be to assert that validations run _after_ segment expansion which does not have any coverage.

Thinking about it, just 1-2 would be sufficient as (3) is the same case as (1).

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -4471,4 +4471,117 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
...
+          "site_id" => site.domain,
+          "filters" => [["is", "segment", [segment.id]]],
+          "date_range" => "all",
+          "metrics" => ["events"]
+        })
+
+      assert json_response(conn, 200) == %{
+               "results" => [%{"dimensions" => [], "metrics" => [3]}],
+               "meta" => %{},
+               "query" => %{
```
macobo, line 4574: This will break tomorrow and in general cause a lot of breakage as the "query" as we add more defaults/options to query. 

In general I'd say to follow the same pattern as other tests in this file and verify `"results"` key only. However in this case `query.filters` is also relevant - perhaps do it like this?

```elixir
assert json_response(conn, 200)["results"] == [...]
assert json_response(conn, 200)["query"]["filters"] == [...]
```

apata: > This will break tomorrow and in general cause a lot of breakage as the "query" as we add more defaults/options to query.

Isn't this exactly the reason to validate the whole response?

lib/plausible/stats/filters/query_parser.ex
apata: I considered that, but I think it's better to align the names and order with existing functions in this pipeline.

macobo: Goals and revenue are preloaded separately because they're used later in the query pipeline (hence attached to Query) when constructing the Ecto query and API response. That said, not a blocker.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
apata: As we discussed in the call, I've updated it for the sake of consistency.

test/plausible/stats/query_parser_test.exs
apata: Roger that. Added cases 1 and 2 to showcase / cover that validation runs on resolved filters.

macobo (approved):

PR #4983, opened 2025-01-16
Excluding goals in dashboard

macobo:
### Changes

Depends on https://github.com/plausible/analytics/pull/4980

This PR adds support for filtering out sessions where a goal has been completed (excluding goals).

Along the way:
- Users can now add multiple goal filters (similar to properties) 
- Refactored how filter text is calculated (typescript + removed duplicated logic)
- `has_not_done` is shown as a special operator in the URL for compatibility with new style URLs and expanded when making the FE API request.

Basecamp ref: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/7793417773

## Screenshots

image
image
image


apata:
Impressive to see the whole thing working! Also, great job with the fixes as well.

https://pr-4983.review.plausible.io/plausible.io?filters=((is,goal,(Signup)),(is,browser,(Microsoft%2BEdge)))&period=7d&keybindHint=W
https://pr-4983.review.plausible.io/plausible.io?filters=((has_not_done,goal,(Signup)),(is,browser,(Microsoft%2BEdge)))&period=7d&keybindHint=W

assets/js/dashboard/util/filters.js
```diff
@@ -229,22 +183,25 @@ export function cleanLabels(filters, labels, mergedFilterKey, mergedLabels) {
...
+function serializeFilter([operation, filterKey, clauses, ...modifiers]) {
+  let apiFilterKey = `visit:${filterKey}`
+  if (
+    filterKey.startsWith(EVENT_PROPS_PREFIX) ||
+    EVENT_FILTER_KEYS.has(filterKey)
+  ) {
+    apiFilterKey = `event:${filterKey}`
+  }
+  if (operation === FILTER_OPERATIONS.has_not_done) {
+    return ['has_not_done', ['is', apiFilterKey, clauses, ...modifiers]]
```
apata, line 199: **issue:** I think we should introduce the concept of nested filters on the dashboard, instead of patching in `has_not_done` as an operator with the goal dimension. 

While it certainly works in this case, it creates yet another difference between the FE representation of a filter and the BE representation. 

Current differences at the filter level are manageable (I hope):
* dimension is shortened in the FE
* clauses are flattened

Other planned differences (debatable):
* filter modifiers are moved from last position to the position next to operation

apata: It's a bunch of work, and how exactly it should look like in the UI is debatable, but I believe we should take this leap forward now, rather than later. 

https://github.com/plausible/analytics/pull/4810#issuecomment-2572633688

macobo: > how exactly it should look like in the UI is debatable

The intended UI is already nailed down - we don't want to be rebuilding the filters UI/UX for the sake of this change. AS such, the only effect of this refactor would be in the code rather than visible for the user.

The cost of doing it like this is potentially causing further issues when we rebuild the filters UI/UX, especially around doing redirects. This is a valid concern but given we already have experience around this and we already are doing filter transformations before sending requests to the API, I feel like the trade-off is reasonable.

Note there's a third alternative which would be introducing a `has_not_done_goal` internal operator in the backend. However this has it's own drawbacks in both validation (similar issue to segments) and us sometimes recommending users inspect query payloads to reproduce queries in APIv2.

In summary, we have 3 options:
| Summary | Benefit | Drawback |
| -- | -- | -- |
| 1. Handling it in filter transformations in FE. | least code | potential future complexity or bugs around redirects and filters when refactoring filtering logic |
| 2. Updating FE code to handle nesting | keeps systems more aligned | tricky change, likely a 40+ file change with particular complexity in the Goals modal |
| 3. Internal operator. | simple from a FE perspective | Significantly trickier BE behavioral filters and validations code, conflicts with our advice to clients |

Given these trade-offs I'm still partial to (1). How would you feel if I derisked option a bit by adding this context + documenting the prefixing logic to `serializeApiFilters` as documentation?

apata: UI-wise, just to give some ugly visual aids to what I'm saying above

Screenshot 2025-01-16 at 14 59 41

and group to be represented as an area with one or more filter pills in it, possibly collapsible

Screenshot 2025-01-16 at 15 13 26

apata: Of course, we might just decide that we want this released as is, as soon as possible, but it means in the future supporting one more special redirect scenario: where `?f=has_done,event:goal,Signup,Register` must be recognized as a legal filter.

apata: Ah, I was so focused on getting my message and visual aids across, I didn't notice the reply in between. Checking

macobo: >  This feature would still be deployable and usable over the API while we bring the FE up to speed.

This would not be sufficient - the feature we're building here was to add this filtering capability to the dashboard.

> It seems to me that as soon as we want NOT or OR capability with filters, we'd have to embark on the nested filters support anyways.

You are right, but we don't want to add that capability right now.

Adding advanced filtering capabilities is a whole project onto itself which needs careful consideration on UX side so we'd be able to both serve advanced users while not turning into Google Analytics query builder. We will likely embark on this journey at some point, but it needs a proper plan.

We discussed the UX of this feature with Uku on basecamp and on a call and decided exposing it as presented in this PR is a good compromise until we decide to tackle that.

> Also, have you considered the cost over time of understanding and maintaining the various legacy systems that we have? Since they're going to be around for a long time (at the moment, I don't think we have any cutoffs planned), even a very small cost adds up. The more of these legacy nuances we have, the harder it will be to make progress.

[I did consider it](https://github.com/plausible/analytics/pull/4983#discussion_r1918510291) and given we don't know what form filtering will take in the UI long-term I don't see another option. Speculatively coding up a more complex solution will run into the same problem where where it will turn into legacy as soon as we discover the real requirements.

> A possible fourth option would be to keep the current patched-in UI, but implement support for nested filters in URL params and have the combobox apply it in its realistic nested format right away.

The UX presented above is quite confusing - so we're trading off user experience in a core functionality for a speculative benefit down the line. That's a bad trade-off and not something we want to do.

macobo: I've added extra documentation + tests for `serializeApiFilters` in https://github.com/plausible/analytics/pull/4983/commits/2b2d4ee3ec6ef3d51312bad84132eac7995978d8

apata (approved):

PR #4992, opened 2025-01-20
Remove tailwind important config

aerosol:
Not sure how important it is 🥁 - introduced via https://github.com/plausible/analytics/pull/4947 - but it fixes jumpy external links on hover.
@ukutaht advise please

### DIFF
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -14,7 +14,6 @@ module.exports = {
     "max-w-screen-xl"
   ],
   darkMode: 'class',
-  important: '.plausible',
   theme: {
     container: {
       center: true,


aerosol:
@apata why need preview env? looks fine locally

aerosol:
the class was explicitly added in: https://github.com/plausible/analytics/pull/4947/files#diff-ea9fe738a4d8fa8906d6e4024f4654da49980aa3a91da1248a9af233f507db09R11 and https://github.com/plausible/analytics/pull/4947/files#diff-29df2e888b7f4797d7a855f73f87f80c0a63b951552bf455eb7b481bb3ee15b8L2 (h/t @zoldar)

aerosol:
Found this https://github.com/phenixdigital/phoenix_storybook/blob/main/guides/sandboxing.md

aerosol:
Hot take: because storybook has 0 users currently, and we have at least one annoyed by this dashboard quirk, I think we should merge and figure out implications later on, if anyone feels like it.

macobo (approved):
Uku is out for the week so it makes sense to me to revert this part for now.

PR #4994, opened 2025-01-21
Refactor: Remove `filter_key` terminology from backend

macobo:
### Changes

In a previous PR, using both `dimension` and `filter_key` to refer to filter dimension caused confusion. This PR refactors the code to only refer to dimension or filter_dimension.

Depends on https://github.com/plausible/analytics/pull/4980


apata (approved):

macobo:
Thanks. Not sure how I missed table_decider. The query.ex one is related to metrics in honeycomb - given it might break someones dashboard or saved query, don't want to touch it right now.

PR #4998, opened 2025-01-21
replace Sentry.capture_message with Logger.error where appropriate for CE

ruslandoga:
This PR attempts to make CE errors better but replacing some `Sentry.capture_message` calls with (to be) equivalent `Logger.error` ones. `Sentry.capture_message` in billing, Paddle, HelpScout and other EE-only contexts are left as-is.


lib/plausible/ingestion/counters.ex
```diff
@@ -84,11 +84,13 @@ defmodule Plausible.Ingestion.Counters do
           {_, _} = AsyncInsertRepo.insert_all(Record, records)
         catch
           _, thrown ->
-            Sentry.capture_message(
+            Logger.error(
```
ruslandoga, line 87: This one can possibly be improved with exceptions (CE gets a more descriptive log, Sentry gets a stacktrace):

```elixir
try do
  AsyncInsertRepo.insert_all(Record, records)
rescue
  e ->
    msg = Exception.format(:error, e, __STACKTRACE__)
    Logger.error("Caught an error when trying to flush ingest counters.\n\n  " <> msg, crash_reason: {e, __STACKTRACE__}, sentry: %{extra: %{number_of_records: Enum.count(records)}})
end
```

lib/plausible_web/views/email_view.ex
```diff
@@ -20,15 +20,15 @@ defmodule PlausibleWeb.EmailView do
...
-      |> Map.replace(:query, search_query)
-      |> URI.to_string()
-    else
-      ""
+    case dsn do
+      {endpoint_uri, _public_key, _secret_key} when is_binary(endpoint_uri) ->
+        URI.parse(endpoint_uri)
+        |> Map.replace(:path, path)
+        |> Map.replace(:query, search_query)
+        |> URI.to_string()
```
ruslandoga, line 28: With the current Sentry client version and 

```sh
export SENTRY_DSN=https://7f16d5d6ee70465789e082bd09481556@o1012425.ingest.sentry.io/6643873
```
this returns

```elixir
iex> Sentry.Config.dsn
#==> {"https://o1012425.ingest.sentry.io/api/6643873/envelope/",
#==> "7f16d5d6ee70465789e082bd09481556", nil}

iex> PlausibleWeb.EmailView.sentry_link("trace_id")
#==> "https://o1012425.ingest.sentry.io/organizations/sentry/issues/?query=trace_id"
```

PR #5000, opened 2025-01-22
Scroll depth: Register `pageleave` listener earlier

macobo:
Currently, pageleave listener is only registered when a successful response is received from plausible API.

After this change pageleave listener is registered immediately when pageview is triggered, hopefully increasing capture rate.

TODO after deploy:
- [ ] Purge bunny cache
- [ ] Measure impact (taking into account pageleaves without pageviews)

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -6,3 +6,4 @@ taht
 referer
 referers
 statics
+firs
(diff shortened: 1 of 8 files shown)


RobertJoonas (approved):
Looks good to me. Let's see how it goes 👍

PR #5001, opened 2025-01-22
WIP: Speed up tracker CI by caching dependencies

macobo:
Tries to speed up tracker CI by caching node_modules and playwright browsers.


macobo:
This does not work due to webkit doing its own magic: https://github.com/microsoft/playwright/issues/30538

PR #5005, opened 2025-01-22
Test out fetch-with-keepalive on plausible.io

macobo:
Fetch with keepalive is a
[widely-supported](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) which indicates whether the browser will keep the associated request alive if the page that initiated it is unloaded before the request is complete.

We're hoping it will improve event capture rates for `pageleave` and `pageview` events when the user closes the tab

To use it, we also need to start using `fetch` (with fallback to xhr).

For extra safety, we will only deploy this on `plausible.io` initially. This will ensure that if there are issues we will be able to react without affecting any other customers.

I tested this locally by adding temporary logging to the script:
1. Modifying lib/plausible_web/templates/layout/_tracking.html.heex to track locally
2. Making fetch the default and running the test suite.

TODO after this PR:
- [ ] Companion docs PR
- [ ] Purge bunny cache
- [ ] Make fetch the default request method without data-property

### DIFF
--- a/lib/plausible_web/templates/layout/_tracking.html.heex
+++ b/lib/plausible_web/templates/layout/_tracking.html.heex
@@ -5,6 +5,7 @@
       data-api={PlausibleWeb.Dogfood.api_destination()}
       data-domain={PlausibleWeb.Dogfood.domain(@conn)}
       src={PlausibleWeb.Dogfood.script_url()}
+      data-allow-fetch
     >
     </script>
     <script>
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -11,6 +11,7 @@
   {{/if}}
   var endpoint = scriptEl.getAttribute('data-api') || defaultEndpoint(scriptEl)
   var dataDomain = scriptEl.getAttribute('data-domain')
+  var allowFetch = scriptEl.hasAttribute('data-allow-fetch')
 
   function onIgnoredEvent(eventName, reason, options) {
     if (reason) console.warn('Ignoring Event: ' + reason);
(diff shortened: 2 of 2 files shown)


lib/plausible_web/templates/layout/_tracking.html.heex
```diff
@@ -5,6 +5,7 @@
       data-api={PlausibleWeb.Dogfood.api_destination()}
       data-domain={PlausibleWeb.Dogfood.domain(@conn)}
       src={PlausibleWeb.Dogfood.script_url()}
+      data-allow-fetch
```
RobertJoonas, line 8: Wouldn't it give a clearer comparison if we'd duplicate the script:

1) the current script would remain as it is
2) a new script will start sending events to a new dashboard (e.g. fetchtest.plausible.io) with `fetch`

We could then compare the actual number between the two sites. Otherwise it depends on the traffic trends as well.

tracker/src/plausible.js
```diff
@@ -11,6 +11,7 @@
   {{/if}}
   var endpoint = scriptEl.getAttribute('data-api') || defaultEndpoint(scriptEl)
   var dataDomain = scriptEl.getAttribute('data-domain')
+  var allowFetch = scriptEl.hasAttribute('data-allow-fetch')
```
RobertJoonas, line 14: How about doing this in a script extension instead, in order to not increase the file size for anyone yet?

macobo: Marked it conditional on `pageleave` flag.

lib/plausible_web/templates/layout/_tracking.html.heex
macobo: Creating a new script variant and creating a different dashbord for it is overkill:
- Fetch is supported everywhere but IE (https://caniuse.com/?search=fetch) and accounts for 0.003% of usage. Realistically I'm not worried about this blowing up - most other analytics services are also using `fetch` at this point.
- On comparison side, we have a good baseline already based on `pageleave` data we're already capturing. Once this is deployed, bunny cache cleared and 24 hours is over, we will start seeing whether this makes a significant difference or not.

Aside: I'm planning on making `fetch` the default transport in a separate push after this.

RobertJoonas (approved):
👍

PR #5008, opened 2025-01-22
Implement support for multiple team owners and multiple teams per user 

zoldar:
This PR addresses enabling support for following across the whole application:

- more than one user as an owner of a single team
- more than one non-guest membership per user

Summary of changes:

- Teams are exposed in CRM now and all team related fields were moved there from User CRM with a link to it
- Enterprise Plan CRM refers to team directly as well now though the team is still searchable via owner's name or email; the workarounds for limited Kaffy association selection are left in place because default way of using team name only when listing and looking up are not sufficient; A faulty feature selection prefill on enterprise plan create was fixed BTW
- The Editor role is now checked for consistently across all `AuthorizeSiteAccess` calls instead of implicit translation of legacy `Admin` in the input roles to `Editor`
- Team is now can now be explicitly provided during site creation service. In Sites API it's exposed as an optional argument, `team_id` (required though if user is an owner on more than one team already). If the argument is not passed, an implicit team is used or created if there isn't one already. In case of adding a website from the dashboard, a currently selected team is used. The controller action has an option to accept "team_id" but it's not exposed in the UI yet; This leads to a problematic corner case mentioned below
- User deletion is prevented when user is the only owner on a setup team; the handling logic will be extended in a follow-up
- API key provisioning is checked against Stats API availability only if the user is an owner of at most 1 team; otherwise, they are allowed to create the API regardless of feature availability on any of the owned teams; The feature availability is checked at the time the key is used anyway; We may consider removing the check at the provisioning time entirely - currently it's only preserved for "legacy" behavior
- notification emails directed at site or team owner are now sent to all owners at once, if there's more than one
- when handling Paddle webhook for subscription creation for a user without a team yet (a case of non-member user with pending site ownership transfer), we always default to creating a new "My Team" even if there are an owner on existing one to avoid ending up with 2 active subscriptions on one existing team
- site invitation accept routine is now passed a team the user is currently switched to. This is only relevant for site ownership transfer; the transfer will be attempted with a current team (and will fail early if permissions are insufficient)
- when checking team member usage quota, only one owner is subtracted from the usage calculation even there are multiple
- when site transfer ownership is executed, all existing owners are added as guest editors to the transferred site (if they aren't members on the other team already)
- direct site ownership transfer from CRM is now accepting optional "Team Identifier" for gracefully handling cases of users owning more than one team; the identifier is exposed in Teams CRM when viewing a particular team
- there's a team switcher in the upper right menu now; the currently switched team is stored in session cookie; the switcher is only shown when user is team member on more than one team
- there are now 3 assigns populated by `AuthPlug` and `AuthContext` - `my_team` (existing), `current_team` and `multiple_teams?`; `multiple_teams?` is a flag used to determine whether to show the team switcher in the upper menu; `current_team` is a team set via team switcher and stored under `current_team_id` in session cookie; if there's none set, it falls back to `my_team`; `my_team` is either the same as `current_team` is user's team role is `owner` or the first found team where the user is an owner otherwise - they may differ
- Listing sites via /sites is now scoping the list with current team; That applies to only full memberships -  sites where user is a guest and guest invitations are all visible regardless of which team the user is currently switched to
- Listing sites via Sites API has an option to scope via team identifier with `team_id` parameter

There still is a number of corner cases to consider, most likely in a follow-up:

- HelpScout integration implicitly falls back on the first team found owned by a user when they are an owner of more than one team; the integration will probably get extended with a way to list those teams and let support folks switch easily between them
- Notes exposed in HelpScout integration are still on user only; it will probably make sense to have them on a team as well
- when trying to create a site as a viewer member of another team (without any other team membership, including implicit one), the user is basically locked out of creating a team of their own until the are removed from the current one; We might want to provide an "exit hatch" where, on failed attempt, we suggest user an option to create a site under their own account
- team Identifier is not yet explicitly exposed anywhere; given Sites API has now ability use it, the identifier should be exposed somehow - via Team Switcher in the UI and/or with additional endpoint in Sites API
- ~when a user has their own implicit team and gets added as a team member to another team, when switched to their team, they still see CTA option and can't access team member management until they complete it. It feels a bit weird but maybe it's fine?~ Leaving it as is for now.
- ~the /sites view still lists all sites the user has access to, regardless of which team they are switched to; the view will probably be scoped to the currently switched team, though the Sites API endpoint should probably still list everything, with an optional team ID parameter?~ Added
- ~add ability to choose team to transfer ownership to in CRM~

## TODO

- [x] set `is_autocreated` to `false` either on role downgrade or team membership upsert where role is downgraded (or at least double-check the implications of leaving it on)
- [x] hide team switcher when there's only one team to choose from
- [x] list owned teams in user CRM edit form for easy navigation
- [x] add as view-only team CRM  edit form: subscription plan, subscription status, grace period
- [x] properly enable lookup by team identifier in Enterprise Plan CRM edit picker
- [ ] more manual testing, possibly extending automated test suite

### Changes

Depends on https://github.com/plausible/analytics/pull/5004

### Tests

### DIFF
--- a/extra/lib/plausible/help_scout.ex
+++ b/extra/lib/plausible/help_scout.ex
@@ -90,6 +90,12 @@ defmodule Plausible.HelpScout do
             plan = Billing.Plans.get_subscription_plan(team.subscription)
             {team, team.subscription, plan}
 
+          {:error, :multiple_teams} ->
+            [team | _] = Plausible.Teams.Users.owned_teams(user)
+            team = Plausible.Teams.with_subscription(team)
+            plan = Billing.Plans.get_subscription_plan(team.subscription)
+            {team, team.subscription, plan}
+
           {:error, :no_team} ->
             {nil, nil, nil}
         end
--- a/extra/lib/plausible/help_scout.ex
+++ b/extra/lib/plausible/help_scout.ex
@@ -108,6 +108,17 @@ defmodule Plausible.HelpScout do
           Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :show, :auth, :user, user.id)
         end
 
+      sites_link =
+        if team do
+          Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :index, :sites, :site,
+            custom_search: team.identifier
+          )
+        else
+          Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :index, :sites, :site,
+            custom_search: user.email
+          )
+        end
+
       {:ok,
        %{
          email: user.email,
@@ -117,10 +128,7 @@ defmodule Plausible.HelpScout do
          plan_label: plan_label(subscription, plan),
          plan_link: plan_link(subscription),
          sites_count: Plausible.Teams.owned_sites_count(team),
-         sites_link:
-           Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :index, :sites, :site,
-             custom_search: user.email
-           )
+         sites_link: sites_link
        }}
     end
   end
--- a/extra/lib/plausible/help_scout.ex
+++ b/extra/lib/plausible/help_scout.ex
@@ -91,6 +91,7 @@ defmodule Plausible.HelpScout do
             {team, team.subscription, plan}
 
           {:error, :multiple_teams} ->
+            # NOTE: We might consider exposing the other teams later on
             [team | _] = Plausible.Teams.Users.owned_teams(user)
             team = Plausible.Teams.with_subscription(team)
             plan = Billing.Plans.get_subscription_plan(team.subscription)
--- a/extra/lib/plausible/help_scout.ex
+++ b/extra/lib/plausible/help_scout.ex
@@ -101,13 +101,19 @@ defmodule Plausible.HelpScout do
             {nil, nil, nil}
         end
 
+      status_link =
+        if team do
+          Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :show, :teams, :team, team.id)
+        else
+          Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :show, :auth, :user, user.id)
+        end
+
       {:ok,
        %{
          email: user.email,
          notes: user.notes,
          status_label: status_label(team, subscription),
-         status_link:
-           Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :show, :auth, :user, user.id),
+         status_link: status_link,
          plan_label: plan_label(subscription, plan),
          plan_link: plan_link(subscription),
          sites_count: Plausible.Teams.owned_sites_count(team),
--- a/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
+++ b/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
@@ -30,7 +30,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
     user = conn.assigns.current_user
 
     with {:ok, site_id} <- expect_param_key(params, "site_id"),
-         {:ok, site} <- get_site(user, site_id, [:owner, :admin, :viewer]) do
+         {:ok, site} <- get_site(user, site_id, [:owner, :admin, :editor, :viewer]) do
       page =
         site
         |> Plausible.Goals.for_site_query()
@@ -96,7 +96,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   end
 
   def get_site(conn, %{"site_id" => site_id}) do
-    case get_site(conn.assigns.current_user, site_id, [:owner, :admin, :viewer]) do
+    case get_site(conn.assigns.current_user, site_id, [:owner, :admin, :editor, :viewer]) do
       {:ok, site} ->
         json(conn, %{
           domain: site.domain,
@@ -122,7 +122,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
 
   def update_site(conn, %{"site_id" => site_id} = params) do
     # for now this only allows to change the domain
-    with {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin]),
+    with {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin, :editor]),
          {:ok, site} <- Plausible.Site.Domain.change(site, params["domain"]) do
       json(conn, site)
     else
@@ -139,7 +139,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   def find_or_create_shared_link(conn, params) do
     with {:ok, site_id} <- expect_param_key(params, "site_id"),
          {:ok, link_name} <- expect_param_key(params, "name"),
-         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin]) do
+         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin, :editor]) do
       shared_link = Repo.get_by(Plausible.Site.SharedLink, site_id: site.id, name: link_name)
 
       shared_link =
@@ -173,7 +173,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   def find_or_create_goal(conn, params) do
     with {:ok, site_id} <- expect_param_key(params, "site_id"),
          {:ok, _} <- expect_param_key(params, "goal_type"),
-         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin]),
+         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin, :editor]),
          {:ok, goal} <- Goals.find_or_create(site, params) do
       json(conn, goal)
     else
@@ -191,7 +191,7 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   def delete_goal(conn, params) do
     with {:ok, site_id} <- expect_param_key(params, "site_id"),
          {:ok, goal_id} <- expect_param_key(params, "goal_id"),
-         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin]),
+         {:ok, site} <- get_site(conn.assigns.current_user, site_id, [:owner, :admin, :editor]),
          :ok <- Goals.delete(goal_id, site) do
       json(conn, %{"deleted" => true})
     else
--- a/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
+++ b/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
@@ -8,16 +8,18 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
   alias Plausible.Sites
   alias Plausible.Goal
   alias Plausible.Goals
+  alias Plausible.Teams
   alias PlausibleWeb.Api.Helpers, as: H
 
   @pagination_opts [cursor_fields: [{:id, :desc}], limit: 100, maximum_limit: 1000]
 
   def index(conn, params) do
+    team = Teams.get(params["team_id"])
     user = conn.assigns.current_user
 
     page =
       user
-      |> Sites.for_user_query()
+      |> Sites.for_user_query(team)
       |> paginate(params, @pagination_opts)
 
     json(conn, %{
--- a/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
+++ b/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
@@ -60,8 +60,9 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
 
   def create_site(conn, params) do
     user = conn.assigns.current_user
+    team = Plausible.Teams.get(params["team_id"])
 
-    case Sites.create(user, params) do
+    case Sites.create(user, params, team) do
       {:ok, %{site: site}} ->
         json(conn, site)
 
@@ -71,6 +72,20 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
         |> json(%{
           error:
             "Your account has reached the limit of #{limit} sites. To unlock more sites, please upgrade your subscription."
+        })
+
+      {:error, _, :permission_denied, _} ->
+        conn
+        |> put_status(403)
+        |> json(%{
+          error: "You can't add sites to the selected team."
+        })
+
+      {:error, _, :multiple_teams, _} ->
+        conn
+        |> put_status(400)
+        |> json(%{
+          error: "You must select a team with 'team_id' parameter."
         })
 
       {:error, _, changeset, _} ->
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -7,6 +7,7 @@ defmodule Plausible.Auth do
   use Plausible.Repo
   alias Plausible.Auth
   alias Plausible.RateLimit
+  alias Plausible.Teams
 
   @rate_limits %{
     login_ip: %{
@@ -71,9 +72,9 @@ defmodule Plausible.Auth do
 
   def delete_user(user) do
     Repo.transaction(fn ->
-      case Plausible.Teams.get_by_owner(user) do
-        {:ok, team} ->
-          for site <- Plausible.Teams.owned_sites(team) do
+      case Teams.get_by_owner(user) do
+        {:ok, %{setup_complete: false} = team} ->
+          for site <- Teams.owned_sites(team) do
             Plausible.Site.Removal.run(site)
           end
 
@@ -84,13 +85,37 @@ defmodule Plausible.Auth do
           )
 
           Repo.delete!(team)
+          Repo.delete!(user)
 
-        _ ->
-          :skip
+        {:ok, team} ->
+          check_can_leave_team!(team)
+          Repo.delete!(user)
+
+        {:error, :multiple_teams} ->
+          check_can_leave_teams!(user)
+          Repo.delete!(user)
+
+        {:error, :no_team} ->
+          Repo.delete!(user)
       end
 
-      Repo.delete!(user)
+      :deleted
     end)
+  end
+
+  defp check_can_leave_teams!(user) do
+    user
+    |> Teams.Users.owned_teams()
+    |> Enum.reject(&(&1.setup_complete == false))
+    |> Enum.map(fn team ->
+      check_can_leave_team!(team)
+    end)
+  end
+
+  defp check_can_leave_team!(team) do
+    if Teams.Memberships.owners_count(team) <= 1 do
+      Repo.rollback(:is_only_team_owner)
+    end
   end
 
   on_ee do
@@ -108,7 +133,7 @@ defmodule Plausible.Auth do
           {:ok, Auth.ApiKey.t()} | {:error, Ecto.Changeset.t() | :upgrade_required}
   def create_api_key(user, name, key) do
     team =
-      case Plausible.Teams.get_by_owner(user) do
+      case Teams.get_by_owner(user) do
         {:ok, team} -> team
         _ -> nil
       end
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -132,17 +132,12 @@ defmodule Plausible.Auth do
   @spec create_api_key(Auth.User.t(), String.t(), String.t()) ::
           {:ok, Auth.ApiKey.t()} | {:error, Ecto.Changeset.t() | :upgrade_required}
   def create_api_key(user, name, key) do
-    team =
-      case Teams.get_by_owner(user) do
-        {:ok, team} -> team
-        _ -> nil
-      end
-
     params = %{name: name, user_id: user.id, key: key}
     changeset = Auth.ApiKey.changeset(%Auth.ApiKey{}, params)
 
-    with :ok <- Plausible.Billing.Feature.StatsAPI.check_availability(team),
-         do: Repo.insert(changeset)
+    with :ok <- check_stats_api_available(user) do
+      Repo.insert(changeset)
+    end
   end
 
   @spec delete_api_key(Auth.User.t(), integer()) :: :ok | {:error, :not_found}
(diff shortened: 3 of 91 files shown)


extra/lib/plausible/help_scout.ex
```diff
@@ -90,17 +90,30 @@ defmodule Plausible.HelpScout do
             plan = Billing.Plans.get_subscription_plan(team.subscription)
             {team, team.subscription, plan}
 
+          {:error, :multiple_teams} ->
```
zoldar, line 93: This case will be handled better on further iterations. For now we just fall back on first team the user owns.

extra/lib/plausible/help_scout.ex
```diff
@@ -90,17 +90,30 @@ defmodule Plausible.HelpScout do
...
+            team = Plausible.Teams.with_subscription(team)
+            plan = Billing.Plans.get_subscription_plan(team.subscription)
+            {team, team.subscription, plan}
+
           {:error, :no_team} ->
             {nil, nil, nil}
         end
 
+      status_link =
+        if team do
```
zoldar, line 105: If user does not own any team yet, we fall back on linking directly to the User CRM

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -60,8 +60,9 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
 
   def create_site(conn, params) do
     user = conn.assigns.current_user
+    team = Plausible.Teams.get(params["team_id"])
 
-    case Sites.create(user, params) do
+    case Sites.create(user, params, team) do
```
zoldar, line 65: `Sites(user, params, nil)` will preserve the current behavior of adding the site to the existing implicit team or creating one.

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -73,6 +74,20 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
             "Your account has reached the limit of #{limit} sites. To unlock more sites, please upgrade your subscription."
         })
 
+      {:error, _, :permission_denied, _} ->
```
zoldar, line 77: The new error cases will only occur on explicit team selection or once user is an owner of more than one team at once.

lib/plausible_web/live/auth_context.ex
```diff
@@ -31,11 +32,35 @@ defmodule PlausibleWeb.Live.AuthContext do
...
+          case current_team &&
+                 Plausible.Teams.Memberships.team_role(current_team, context.current_user) do
+            {:ok, :owner} -> true
+            _ -> false
+          end
+
+        case {current_team_owner?, current_team, context.current_user} do
+          {_, nil, nil} -> nil
+          {true, %Teams.Team{}, _} -> current_team
+          {_, _, %{team_memberships: [%{team: team} | _]}} -> team
```
zoldar, line 54: TODO: Ensure team memberships are sorted consistently in `UserAuth`

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -297,7 +297,7 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
     ~H"""
     <div class="mt-6 space-y-3" x-data={@js_data}>
       <PlausibleWeb.Components.Billing.Notice.premium_feature
-        billable_user={@site.owner}
+        billable_user={List.first(@site.owners)}
```
zoldar, line 300: TODO: Ensure `site.owners` (and other similar relationships) order entries consistently.

lib/plausible_web/live/auth_context.ex
zoldar: https://github.com/plausible/analytics/pull/5008/commits/73347394e436311e6e06e12238e5d5106960bb63

lib/plausible_web/live/goal_settings/form.ex
zoldar: https://github.com/plausible/analytics/pull/5008/commits/73347394e436311e6e06e12238e5d5106960bb63

lib/plausible/auth/auth.ex
```diff
@@ -71,9 +72,9 @@ defmodule Plausible.Auth do
 
   def delete_user(user) do
     Repo.transaction(fn ->
-      case Plausible.Teams.get_by_owner(user) do
-        {:ok, team} ->
-          for site <- Plausible.Teams.owned_sites(team) do
+      case Teams.get_by_owner(user) do
```
zoldar, line 75: The team is deleted along with the user only if its' setup is not complete yet. Otherwise, user can only be deleted if they aren't only only owner on any of the teams they have that role in.

lib/plausible/billing/billing.ex
```diff
@@ -236,7 +245,8 @@ defmodule Plausible.Billing do
       )
 
     if plan do
-      api_keys = from(key in Plausible.Auth.ApiKey, where: key.user_id == ^team.owner.id)
+      owner_ids = Enum.map(team.owners, & &1.id)
+      api_keys = from(key in Plausible.Auth.ApiKey, where: key.user_id in ^owner_ids)
```
zoldar, line 249: Altering hourly request limits on subscription update is tricky when each user can be a member/owner on more than one team. For now the logic here is left intact but with teams out, this might lead to weird inconsistencies in some cases. Perhaps we should consider storing hourly request limit on either team or subscription instead? Especially _if_ the Stats API capability is checked against the site (and, by extension, team) already?

lib/plausible/site.ex
```diff
@@ -47,8 +47,8 @@ defmodule Plausible.Site do
     has_one :google_auth, GoogleAuth
     has_one :weekly_report, Plausible.Site.WeeklyReport
     has_one :monthly_report, Plausible.Site.MonthlyReport
-    has_one :ownership, through: [:team, :ownership]
-    has_one :owner, through: [:team, :owner]
+    has_many :ownerships, through: [:team, :ownerships], preload_order: [asc: :id]
```
zoldar, line 50: Preload order is meant to enforce consistent ordering of owners. This is desirable when, for instance, we fall back on passing only the onet owner. This currently applies to all plan notification banner components accepting `billable_user` .

lib/plausible/site/memberships.ex
```diff
@@ -5,7 +5,9 @@ defmodule Plausible.Site.Memberships do
 
   alias Plausible.Site.Memberships
 
-  defdelegate accept_invitation(invitation_id, user), to: Memberships.AcceptInvitation
+  defdelegate accept_invitation(invitation_id, user, team \\ nil),
```
zoldar, line 8: Invitation acceptance and (direct) ownership transfer now accept explicit team the site is transferred to. In the UI, it's the team the user is currently switched to. NOTE: Double check a case of user without a team trying to accept the transfer. They should be notified they must choose a plan first - but the failure must be graceful, so the eligibility check must be done earlier.

lib/plausible/sites.ex
```diff
@@ -163,16 +163,25 @@ defmodule Plausible.Sites do
     )
   end
 
-  def create(user, params) do
+  def create(user, params, team \\ nil) do
```
zoldar, line 166: Team creation behaves as it was before if no team is provided but it will add a site under an explicitly passed team when present and permitted.

lib/plausible/sites.ex
```diff
@@ -304,9 +313,7 @@ defmodule Plausible.Sites do
     locked
   end
 
-  def get_for_user!(user, domain, roles \\ [:owner, :admin, :viewer]) do
-    roles = translate_roles(roles)
-
+  def get_for_user!(user, domain, roles \\ [:owner, :admin, :editor, :viewer]) do
```
zoldar, line 316: Editor role is now explicitly accounted for in all site-scoped checks.

lib/plausible/teams.ex
```diff
@@ -156,6 +137,25 @@ defmodule Plausible.Teams do
     end
   end
 
+  @spec force_create_my_team(Auth.User.t()) :: Teams.Team.t()
+  def force_create_my_team(user) do
+    {:ok, team} =
+      Repo.transaction(fn ->
+        Repo.update_all(
```
zoldar, line 144: This update is done to avoid hitting uniqueness constraint. In practice, forced "My Team" creation is only used in the special case of creating a subscription for a user without a team at time of the checkout (already mentioned earlier).

lib/plausible/teams/billing.ex
```diff
@@ -382,7 +382,7 @@ defmodule Plausible.Teams.Billing do
   def team_member_usage(nil, _), do: 0
 
   def team_member_usage(team, opts) do
-    {:ok, owner} = Teams.get_owner(team)
+    [owner | _] = Repo.preload(team, :owners).owners
```
zoldar, line 385: We only exclude one owner from team member usage.

lib/plausible/teams/team_admin.ex
```diff
@@ -0,0 +1,221 @@
...
+      },
+      lock: %{
+        name: "Lock",
+        action: fn _, team -> lock(team) end
+      }
+    ]
+  end
+
+  def delete(_conn, %{data: _team}) do
+    # TODO: Implement custom team removal
```
zoldar, line 85: To be addressed in a follow-up.

lib/plausible_web/controllers/admin_controller.ex
```diff
@@ -8,18 +8,12 @@ defmodule PlausibleWeb.AdminController do
...
-          team
-          |> Teams.with_subscription()
-          |> Plausible.Repo.preload(:owner)
-
-        {:error, :no_team} ->
-          nil
-      end
+      team_id
+      |> Teams.get()
+      |> Repo.preload([:owners, team_memberships: :user])
```
zoldar, line 16: NOTE: Probably lost `with_subscription` here.

lib/plausible_web/controllers/admin_controller.ex
```diff
@@ -74,21 +64,39 @@ defmodule PlausibleWeb.AdminController do
...
-    id = params["user_id"]
+  def team_by_id(conn, params) do
+    id = params["team_id"]
 
     entry =
       Repo.one(
-        from u in Plausible.Auth.User,
-          where: u.id == ^id,
-          select: fragment("concat(?, ?, ?, ?)", u.name, " (", u.email, ")")
+        from t in Plausible.Teams.Team,
```
zoldar, line 72: The queries are meant to build meaningful names for teams in the context of CRM. The general idea is:

- if the team name is "My Team", we only display the name(s) and email(s) of owner(s)
- otherwise we display team name with owner names in suffix

lib/plausible_web/controllers/invitation_controller.ex
```diff
@@ -7,7 +7,10 @@ defmodule PlausibleWeb.InvitationController do
        [:owner, :editor, :admin] when action in [:remove_invitation]
 
   def accept_invitation(conn, %{"invitation_id" => invitation_id}) do
-    case Plausible.Site.Memberships.accept_invitation(invitation_id, conn.assigns.current_user) do
+    current_user = conn.assigns.current_user
+    team = conn.assigns.current_team
```
zoldar, line 11: When accepting the site transfer invite, we use the team the user is currently switched to. It may happen they may not be permitted to add a site there, which is handled explicitly.

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -26,12 +26,13 @@ defmodule PlausibleWeb.SiteController do
   end
 
   def create_site(conn, %{"site" => site_params}) do
-    team = conn.assigns.my_team
+    current_team = conn.assigns.current_team
+    team = Plausible.Teams.get(site_params["team_id"]) || current_team
```
zoldar, line 30: The UI part of resolving "wrong team" issue is not implemented yet. One way the user can act on it, is switching to another team. However, if they don't have any other team where they have permission to create sites, they are basically locked out from creating their own team. This calls for adding an extra "resolution" step where user can pick to create their own team in such case.

lib/plausible_web/live/auth_context.ex
```diff
@@ -31,11 +32,35 @@ defmodule PlausibleWeb.Live.AuthContext do
           _ -> nil
         end
       end)
+      |> assign_new(:team_from_session, fn _ ->
```
zoldar, line 35: Both, `AuthContext` and `AuthPlug` provide 2 other assigns in addition to `my_team` - `current_team` and `multiple_teams?`. The `current_team` stores the team the user is currently switched to. `multuple_teams?` is used to determine whether site switcher should be rendered at all. `my_team` is always a team the user is an owner in (first pick when there are more and the explicitly picked team is not owned by them)

lib/plausible_web/live/sites.ex
```diff
@@ -674,20 +674,14 @@ defmodule PlausibleWeb.Live.Sites do
     )
   end
 
-  defp extract_invitations(sites, user) do
+  defp extract_invitations(sites, team) do
```
zoldar, line 677: Site transfers are checked against limits of the team the user is currently switched to.

lib/plausible_web/plugs/authorize_public_api.ex
```diff
@@ -134,11 +134,7 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPI do
   end
 
   defp verify_site_access(api_key, site) do
-    team =
-      case Plausible.Teams.get_by_owner(api_key.user) do
-        {:ok, team} -> team
-        _ -> nil
-      end
+    team = Repo.preload(site, :team).team
```
zoldar, line 137: This changes the existing behavior - we check Stats API availability against queried site's team instead of the team the user owns.

lib/plausible_web/user_auth.ex
```diff
@@ -142,6 +142,7 @@ defmodule PlausibleWeb.UserAuth do
         left_lateral_join: ts in subquery(last_team_subscription_query),
         on: true,
         where: us.token == ^token and us.timeout_at > ^now,
+        order_by: t.id,
```
zoldar, line 145: Ensuring consistent ordering for `my_team` selection in `AuthPlug` and `AuthContext`.

lib/plausible_web/controllers/admin_controller.ex
zoldar: https://github.com/plausible/analytics/pull/5008/commits/8709c9b2ff9d5eb01e897fc12412e2ffb29119f8

lib/plausible_web/controllers/admin_controller.ex
```diff
@@ -177,7 +209,7 @@ defmodule PlausibleWeb.AdminController do
 
       sites_link =
         Routes.kaffy_resource_url(PlausibleWeb.Endpoint, :index, :sites, :site,
-          custom_search: team.owner.email
+          custom_search: List.first(team.owners).email
```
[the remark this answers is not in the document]
zoldar: https://github.com/plausible/analytics/pull/5008/commits/86fd5ccc8d80ad14aa710dfe17a9711704b1bdc5

lib/plausible/site/admin.ex
```diff
@@ -150,12 +153,15 @@ defmodule Plausible.SiteAdmin do
     {:error, "Please select at least one site from the list"}
   end
 
-  defp transfer_ownership_direct(_conn, sites, %{"email" => email}) do
+  defp transfer_ownership_direct(_conn, sites, %{"email" => email} = params) do
```
zoldar, line 156: CRM must have a way to provide a team identifier. NOTE: Given it's direct ownership transfer, the "new_owner" does not make much sense - picking destination team would be more appropriate.

zoldar: Added ability to provide (optional) team identifier when doing direct transfer in https://github.com/plausible/analytics/pull/5008/commits/b6faceef839c56140fc975cd723a6f4b173326bb

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -20,13 +21,24 @@ defmodule PlausibleWeb.AuthPlug do
...
         user = user_session.user
 
-        team =
-          case user.team_memberships do
-            [%{team: team}] ->
-              team
+        current_team =
+          conn
+          |> Plug.Conn.get_session("current_team_id")
+          |> Plausible.Teams.get()
```
ukutaht, line 27: Can't we get the current team from `user.team_memberships.team` and avoid this database query?

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -20,13 +21,24 @@ defmodule PlausibleWeb.AuthPlug do
...
+          |> Teams.with_subscription()
+          |> Repo.preload(:owners)
+
+        current_team_owner? =
+          case current_team && Teams.Memberships.team_role(current_team, user) do
```
ukutaht, line 32: Same question as above - isn't this data already available via `user.team_memberships`?

metmarkosaric:
> I'm not sure if it's necessary to keep the old owners around as guests. @metmarkosaric curious what you think

I'm not sure i understand the context of this. currently if you remove a team member who has an owner role, we don't remove them but just demote them to be guests. is that it? 

I don't think that's necessary. after someone is converted from a guest to a full team member we can forget that they were ever a guest in the past on the sites that the team owns. and if they ever get removed from a team, they are removed and don't keep any access

lib/plausible_web/plugs/auth_plug.ex
```diff
@@ -35,7 +47,9 @@ defmodule PlausibleWeb.AuthPlug do
         conn
         |> assign(:current_user, user)
         |> assign(:current_user_session, user_session)
-        |> assign(:my_team, team)
+        |> assign(:my_team, my_team)
```
ukutaht, line 50: The way `my_team` vs `current_team` is handled doesn't seem completely sound... I was able locally to get into a state where I can see settings for a team I am not currently switched to:

Screenshot 2025-02-18 at 12 49 29

lib/plausible/billing/billing.ex
```diff
@@ -138,9 +140,16 @@ defmodule Plausible.Billing do
         Teams.get!(team_id)
 
       {:user_id, user_id} ->
```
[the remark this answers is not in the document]
zoldar: It's a legit case where a user without any sites yet (and without a team) has a pending site ownership transfer and has to subscribe to a suitable plan before before they can accept it.

lib/plausible_web/plugs/auth_plug.ex
zoldar: The current struct coming from `UserAuth` has only `owner` memberships preloaded there - we'd have to change it so that it preloads all team memberships and filter after the fact.

lib/plausible_web/plugs/auth_plug.ex
zoldar: Ditto about sparse preload coming from `UserAuth`. We can adjust that though, I think. I'll have a look.

lib/plausible_web/plugs/auth_plug.ex
zoldar: Yeah, my_team<->current_team usage is not fully fleshed out yet and was meant to be addressed when implementing enforcement of team permissions on plug level (along with introducing BIlling role).

zoldar:
> I don't think that's necessary. after someone is converted from a guest to a full team member we can forget that they were ever a guest in the past on the sites that the team owns. and if they ever get removed from a team, they are removed and don't keep any access

Is that something we can change right away so that the behavior changes for site transfers going forward?

lib/plausible_web/templates/stats/site_locked.html.heex
```diff
@@ -44,14 +44,14 @@
           <div class="mt-3 text-gray-600 dark:text-gray-300 text-center">
             <p>
               This dashboard is currently locked and cannot be accessed. The site owner
-              <b>{@site.owner.email}</b>
+              <b>{List.first(@site.owners).email}</b>
               must upgrade their subscription plan in order to
               unlock the stats.
             </p>
             <div class="mt-6 text-sm text-gray-500">
               <p>Want to pay for this site with the account you're logged in with?</p>
```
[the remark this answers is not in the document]
zoldar: Okay to do that right away or should the removal be gated behind FF until release?

metmarkosaric:
> Is that something we can change right away so that the behavior changes for site transfers going forward?

for site transfers it seems more useful to keep access. in a lot of cases, a freelancer or agency would setup the site and only then transfer site ownership to a client. they generally need to keep access to the site even after the client starts owning/paying for it so the way it works now is great for that

deleting a team member seems like a different use case. if someone deletes a team member they would expect this team member not to stay anywhere and not to have access to any of the sites that the team owns

metmarkosaric:
> So I presume, we should leave the current behavior of leaving former owners as guests on the transferred site if they aren't members on that other team already?

yeah this works for me

lib/plausible_web/plugs/auth_plug.ex
zoldar: Turns out I forgot preloaded teams include all non-guest memberships 😂 Anyways, I have made an optimization where we keep fetching everything that's needed in a single query in https://github.com/plausible/analytics/pull/5008/commits/22e91db2286fd0870fcb4f94fb352faa523af4b7. I'm keeping the current behavior of team selection for `my_team` with a note to revise it once we properly introduce "My Personal Sites".

lib/plausible_web/plugs/auth_plug.ex
zoldar: Addressed in https://github.com/plausible/analytics/commit/22e91db2286fd0870fcb4f94fb352faa523af4b7

lib/plausible_web/templates/stats/site_locked.html.heex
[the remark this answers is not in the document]
ukutaht: Ah right, better to keep the current message as long as teams are still feature flagged.

Can do FF or just keep it as a post-release followup. Either one works.

ukutaht (approved):

lib/plausible_web/templates/stats/site_locked.html.heex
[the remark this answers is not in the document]
zoldar: Hiding it when teams FF is on in https://github.com/plausible/analytics/pull/5008/commits/f45de8904b687a65a61fc3a63d7d3c9fd10a3147

PR #5009, opened 2025-01-23
Fix: has_not_done breakdown modals

macobo:
### Changes

Fixes a bug introduced in https://github.com/plausible/analytics/pull/4983:
1. Add a filter to "Goal is not X"
2. Open a breakdown modal
3. Be met with an error

The root cause was us checking whether there's a goal filter applied to display conversion_rate/percentage metrics. has_not_done means these filters should not be applied.


RobertJoonas (approved):
LGTM 👍

PR #5013, opened 2025-01-23
Add a special goal: "WP Form Completions"

RobertJoonas:
### Changes

https://3.basecamp.com/5308029/buckets/32037438/card_tables/cards/6013863433#__recording_8250028578

Adds a new special goal (`WP Form Completion`) with a special property (`path`). This will behave exactly the same as the `404` goal (that has the `path` property too):

* It will be included in the full Plausible CSV export (`imported_custom_events.csv`) with the `path` prop
* it will be included from imported data in stats queries if it exists (from imports where it doesn't, e.g. GA4, we will silently not return any data for this goal)
* In the dashboard, clicking on the goal in the goals breakdown, the "special goal contents" will pop up in the conversions tab:

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/custom_prop_breakdown_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/custom_prop_breakdown_test.exs
@@ -1184,7 +1184,7 @@ defmodule PlausibleWeb.Api.StatsController.CustomPropBreakdownTest do
       :ok
     end
 
-    for special_prop <- ["url", "path", "search_query"] do
+    for special_prop <- Plausible.Props.internal_keys() do
       test "returns breakdown for the internally used #{special_prop} prop key", %{
         site: site,
         conn: conn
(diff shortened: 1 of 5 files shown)


test/plausible_web/controllers/api/stats_controller/custom_prop_breakdown_test.exs
```diff
@@ -1179,31 +1179,31 @@ defmodule PlausibleWeb.Api.StatsController.CustomPropBreakdownTest do
...
       :ok
     end
 
-    test "returns breakdown for internally used prop keys", %{conn: conn, site: site} do
-      [%{"visitors" => 1, "name" => "one"}] =
-        conn
-        |> get("/api/stats/#{site.domain}/custom-prop-values/url?period=day")
-        |> json_response(200)
-        |> Map.get("results")
+    for special_prop <- Plausible.Props.internal_keys() do
```
aerosol, line 1187: Maybe worth adding it also to the remaining tests mentioning "WP Search Query" for example?

I'm seeing `props_test.exs` and `query_imported_test.exs`

aerosol (approved):
LGTM, just one comment wrt potentially better coverage

test/plausible_web/controllers/api/stats_controller/custom_prop_breakdown_test.exs
RobertJoonas: added to `props_test.exs` - https://github.com/plausible/analytics/pull/5013/commits/b34359947560e934567513d7e53130f2f842278d.

In `query_imported_test.exs` it's already covered via `for goal_name <- Plausible.Imported.goals_with_url() do ...`

PR #5015, opened 2025-01-24
APIv2: validate combining event-only metrics with session-only dimensions

macobo:
Currently the following query results in a 500:

```json
{
  "site_id": "plausible.io",
  "metrics": ["visitors", "events", "pageviews"],
  "dimensions": ["visit:exit_page"],
  "date_range": "7d"
}
```

This adds proper validation for that case that was previously missing.

In the future we might be able to remove _some_ of these validations by selecting the appropriate value(s) from sessions table.

Basecamp ref: https://3.basecamp.com/5308029/buckets/39750953/card_tables/cards/8251923578

### DIFF
--- a/test/plausible/stats/table_decider_test.exs
+++ b/test/plausible/stats/table_decider_test.exs
@@ -152,7 +152,35 @@ defmodule Plausible.Stats.TableDeciderTest do
     end
   end
 
-  defp make_query(filter_dimensions, dimensions \\ []) do
+  describe "validate_no_metrics_dimensions_conflict" do
+    for {metrics, dimensions, expected} <- [
+          {[], [], :ok},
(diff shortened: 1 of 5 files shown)


test/plausible/stats/table_decider_test.exs
```diff
@@ -152,7 +152,35 @@ defmodule Plausible.Stats.TableDeciderTest do
     end
   end
 
-  defp make_query(filter_dimensions, dimensions \\ []) do
+  describe "validate_no_metrics_dimensions_conflict" do
+    for {metrics, dimensions, expected} <- [
```
apata, line 156: Nice readable unit test!

apata (approved):

PR #5019, opened 2025-01-27
Proposal: New testing utility `assert_matches`

macobo:
This PR introduces a new testing utility called `assert_matches` to more flexibly test our API responses.

## Motivation

In previous reviews, @apata and I discussed how best test API responses for fields that either aren't deterministic or relevant for the test at hand (e.g. inserted_at/updated_at or APIv2 "query"). Removing these fields creates a lot of boilerplate code in tests, making them harder to read and violating the broad one-assert-per-test rule.

This is not an elixir-specific problem with various solutions in other languages. The one I like is "structural testing". For example, jest has [`expect.any(type)`](https://jestjs.io/docs/expect#expectanyconstructor) and other utility functions that don't require exact equality. RSpec in ruby world has a [whole slate of matchers for similar dynamic checking](https://www.rubydoc.info/gems/rspec-expectations/RSpec/Matchers).

## Proposal

`assert_matches api_response, expected_structure`

If expected structure is fixed and all fields known ahead of time, this would behave exactly like `assert api_response == expected_structure`

However `assert_matches` also allows to only assert the types of some values. Consider this example:

```elixir
        assert_matches response, %{
          "id" => expect_any(:integer),
          "name" => "Some segment",
          "type" => "personal",
          "segment_data" => %{"filters" => [["is", "visit:entry_page", ["/blog"]]]},
          "owner_id" => user.id,
          "inserted_at" => expect_any(:string),
          "updated_at" => expect_any(:string)
        }
```

In this example, for the case `id`, `inserted_at` and `updated_at` types are asserted but the value is not checked to be equal to some precise predetermined value.

If there's an issue (e.g. id is a string), this would report the following diff, highlighting only the fields that are incorrect.
image

## How to review this PR

If you think this proposal is worth it without reviewing the code fully, feel free to add a :+1: reaction to the PR.

If you want to review "meat" of the PR is in [test/support/assert_matches.ex](https://github.com/plausible/analytics/pull/5019/files#diff-148c449601b36a602063fda037f9d43ab402b15373fa075a6ceab37ec5718532)

This PR made no behavioral changes, but I went ahead and refactored all APIv2 query tests to also check `meta` and `query` response values. This added some extra coverage for cases that were previously somewhat untested.

### DIFF
--- a/test/support/assert_matches.ex
+++ b/test/support/assert_matches.ex
@@ -1,4 +1,6 @@
 defmodule Plausible.AssertMatches do
+  @moduledoc false
+
   @doc """
   Custom assertion function that allows for flexible matching of expected results.
 
--- a/test/support/assert_matches.ex
+++ b/test/support/assert_matches.ex
@@ -15,9 +15,9 @@ defmodule Plausible.AssertMatches do
 
   """
   def assert_matches(data, expected) do
-    {equivelent?, left, right} = check_matches(data, expected)
+    {equivalent?, left, right} = check_matches(data, expected)
 
-    if not equivelent? do
+    if not equivalent? do
       raise ExUnit.AssertionError,
         left: left,
         right: right,
@@ -43,8 +43,8 @@ defmodule Plausible.AssertMatches do
         match_maps(value, expected)
 
       {a, b} when is_tuple(a) and is_tuple(b) ->
-        {equivelent?, left, right} = match_lists(Tuple.to_list(a), Tuple.to_list(b))
-        {equivelent?, List.to_tuple(left), List.to_tuple(right)}
+        {equivalent?, left, right} = match_lists(Tuple.to_list(a), Tuple.to_list(b))
+        {equivalent?, List.to_tuple(left), List.to_tuple(right)}
 
       {value, expected} ->
         {value == expected, value, expected}
@@ -56,9 +56,9 @@ defmodule Plausible.AssertMatches do
   defp match_lists(left, []), do: {false, left, []}
 
   defp match_lists([left | left_rest], [right | right_rest]) do
-    {equivelent?, left_acc, right_acc} = match_lists(left_rest, right_rest)
+    {equivalent?, left_acc, right_acc} = match_lists(left_rest, right_rest)
     {eq, l, r} = check_matches(left, right)
-    {equivelent? and eq, [l | left_acc], [r | right_acc]}
+    {equivalent? and eq, [l | left_acc], [r | right_acc]}
   end
 
   defp match_maps(left, expected) do
@@ -72,12 +72,12 @@ defmodule Plausible.AssertMatches do
   defp match_maps([], _, _), do: {true, %{}, %{}}
 
   defp match_maps([key | keys], left, expected) do
-    {equivelent?, left_result, right_result} = match_maps(keys, left, expected)
+    {equivalent?, left_result, right_result} = match_maps(keys, left, expected)
 
     case {left, expected} do
       {%{^key => left_value}, %{^key => right_value}} ->
         {eq, l, r} = check_matches(left_value, right_value)
-        {equivelent? and eq, Map.put(left_result, key, l), Map.put(right_result, key, r)}
+        {equivalent? and eq, Map.put(left_result, key, l), Map.put(right_result, key, r)}
 
       {%{^key => left_value}, _} ->
         {false, Map.put(left_result, key, left_value), right_result}
--- a/test/support/assert_matches.ex
+++ b/test/support/assert_matches.ex
@@ -1,0 +1,109 @@
+defmodule Plausible.AssertMatches do
+  @doc """
+  Custom assertion function that allows for flexible matching of expected results.
+
+  This is useful when testing APIs, where some response fields are not deterministic.
+
+  Examples:
+
+      iex> assert_matches %{ key: 1 }, %{ key: expect_any(:integer) }
+
+      iex> assert_matches %{ string: "abc" }, %{ string: "abc" }
+
+      iex> # assert_matches %{ key: 1 }, %{ key: expect_any(:string) }
+      # This would raise ExUnit.AssertionError with a diff.
+
+  """
+  def assert_matches(data, expected) do
+    {equivelent?, left, right} = check_matches(data, expected)
+
+    if not equivelent? do
+      raise ExUnit.AssertionError,
+        left: left,
+        right: right,
+        message: "Values did not match"
+    end
+  end
+
+  defp check_matches(value, expected) do
+    case {value, expected} do
+      # Main case: If expected value is a function, we assume it's a predicate passed from expect_any
+      {value, predicate_fn} when not is_function(value) and is_function(predicate_fn) ->
+        if predicate_fn.(value) do
+          # :TRICKY: To make ex_unit not highlight the predicate_fn, pretend it was the value passed from left
+          {true, value, value}
+        else
+          {false, value, predicate_fn}
+        end
+
+      {value, expected} when is_list(value) and is_list(expected) ->
+        match_lists(value, expected)
+
+      {value, expected} when is_map(value) and is_map(expected) ->
+        match_maps(value, expected)
+
+      {a, b} when is_tuple(a) and is_tuple(b) ->
+        {equivelent?, left, right} = match_lists(Tuple.to_list(a), Tuple.to_list(b))
+        {equivelent?, List.to_tuple(left), List.to_tuple(right)}
+
+      {value, expected} ->
+        {value == expected, value, expected}
+    end
+  end
+
+  defp match_lists([], []), do: {true, [], []}
+  defp match_lists([], right), do: {false, [], right}
+  defp match_lists(left, []), do: {false, left, []}
+
+  defp match_lists([left | left_rest], [right | right_rest]) do
+    {equivelent?, left_acc, right_acc} = match_lists(left_rest, right_rest)
+    {eq, l, r} = check_matches(left, right)
+    {equivelent? and eq, [l | left_acc], [r | right_acc]}
+  end
+
+  defp match_maps(left, expected) do
+    keys =
+      Map.keys(left)
+      |> Enum.concat(Map.keys(expected))
+
+    match_maps(keys, left, expected)
+  end
+
+  defp match_maps([], _, _), do: {true, %{}, %{}}
+
+  defp match_maps([key | keys], left, expected) do
+    {equivelent?, left_result, right_result} = match_maps(keys, left, expected)
+
+    case {left, expected} do
+      {%{^key => left_value}, %{^key => right_value}} ->
+        {eq, l, r} = check_matches(left_value, right_value)
+        {equivelent? and eq, Map.put(left_result, key, l), Map.put(right_result, key, r)}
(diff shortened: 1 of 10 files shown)


apata (approved):
beautiful!

test/support/assert_matches.ex
```diff
@@ -0,0 +1,111 @@
...
+  defp match_lists([left | left_rest], [right | right_rest]) do
+    {equivalent?, left_acc, right_acc} = match_lists(left_rest, right_rest)
+    {eq, l, r} = check_matches(left, right)
+    {equivalent? and eq, [l | left_acc], [r | right_acc]}
+  end
+
+  defp match_maps(left, expected) do
+    keys =
+      Map.keys(left)
+      |> Enum.concat(Map.keys(expected))
```
zoldar, line 69:
```suggestion
      |> Enum.concat(Map.keys(expected))
      |> Enum.uniq()
```

to save on redundant checks?

test/support/assert_matches.ex
```diff
@@ -0,0 +1,111 @@
...
+  def expect_any(:atom), do: &is_atom/1
+  def expect_any(:string), do: &is_binary/1
+  def expect_any(:integer), do: &is_integer/1
+  def expect_any(:float), do: &is_float/1
+  def expect_any(:boolean), do: &is_boolean/1
+  def expect_any(:map), do: &is_map/1
+  def expect_any(:list), do: &is_list/1
+  def expect_any(:any), do: fn _ -> true end
+
+  def expect_any(type, predicate_fn) do
```
zoldar, line 106: For `:string` it could be special-cased for regex (can be matched by `%Regex{}` to exclude accidental string use) as a predicate and checked with `String.match?(value, regex)` - it would further save on boilerplate.

aerosol:
>  Removing these fields creates a lot of boilerplate code in tests, 

1. What do you mean by "removing" here and how does removing (?) create boilerplate?
2. Also, are you saying that verifying if "`inserted_at` is any string" is valuable assertion to keep explicitly at all times? 

> violating the broad one-assert-per-test rule.

TIL about the rule. Just went ahead to check if elixir itself obeys it - nope.
The "easier to read" spectrum is in the eye of the beholder. I'm on this end I suppose: https://elixirforum.com/t/exunit-and-the-one-assertion-per-test-pattern/46561/3 - even in that thread, it seems the consensus is that for integration tests the rule doesn't apply smoothly.

```elixir
assert_matches response, %{
```

Normally the pattern matching bindings are on the left, which you might want to consider:

```elixir
%{x: y} = m

match? %{x: y}, m
```

macobo:
@aerosol 

> Also, are you saying that verifying if "inserted_at is any string" is valuable assertion to keep explicitly at all times?

It is worth checking whether it's set in the API response or not if it's part of the API contract.

We could also add a more special-case `is_any(:iso8601_string)` or something in the future - the goal right now was to get the baseline in place.

> TIL about the rule. Just went ahead to check if elixir itself obeys it - nope.

Don't take this rule too seriously - having too many asserts is often a sign that the test is doing too much, but it's not a hard rule to follow at all times. 

> What do you mean by "removing" here and how does removing (?) create boilerplate?

Consider this assertion

```elixir
    assert_matches json_response(conn, 200), %{
      "results" => [%{"metrics" => [3], "dimensions" => []}],
      "meta" => %{},
      "query" => %{
        "metrics" => ["pageviews"]
        "filters" => [],
        "dimensions" => [],
        "site_id" => site.domain,
        "date_range" => [AssertMatches.expect_any(:string), AssertMatches.expect_any(:string)],
        "order_by" => AssertMatches.expect_any(:list),
        "pagination" => AssertMatches.expect_any(:map),
        "include" => %{}
      })
    }
```

(I've inlined the `response_query` helper for the sake of clarity)

The equivelent test to assert the same things would be something like:

```elixir
%{ "results" => results, "meta" => meta, "query" => query } = json_response(conn, 200)

assert results == [%{"metrics" => [3], "dimensions" => []}],
assert meta == meta

assert Map.drop(query, ["date_range", "order_by", "pagination"]) == %{
        "metrics" => ["pageviews"]
        "filters" => [],
        "dimensions" => [],
        "site_id" => site.domain,
        "include" => %{}
}

keys = query |> Map.keys() |> Enum.sort()
assert keys == Enum.sort(["metrics", "filters", "dimensions", "site_id", "date_range", "order_by", "pagination", "include"])

assert is_list(query["date_range"]) and length(query["date_range"]) == 2
assert is_binary(query["date_range"][0])
assert is_binary(query["date_range"][1])

assert is_map(query["order_by"])
assert is_list(query["order_by"])
```

This is a ton of code that would need to be duplicated into every test or extracted into a helper or not covered at all. 
- The helper would work, but would need to be different for every use-case and this proposal generalizes how this could be checked
- Not covering it is usually acceptable but will cause questions what to test and when

I prefer the solution(s) that this solution, jest and rspec offers because of it's descriptiveness - I can glance at the code and at-a-glance understand what the intended response structure is even with more nested data structures.

zoldar:
The proposal looks nice, however @aerosol has raised a good point. We are making a heavy tradeoff of basically throwing all the powers of pattern-matching out the window with this particular implementation. We won't always want to check the value exhaustively, espcially in the case of structs. 

Ideally, it would be be great to implement these `expect` checks as things embeddable in the match pattern directly, something like:

```elixir
assert_matches %{field: ^any_string(), number: ^any_integer(& &1 > 20)} = %{field: "some string", number: 42, ignored: true}
```

This could be internally expanded into consecutive matches where a top level match is done first:

```elixir
assert %{field: match_1, number: match_2} = value
assert is_string(match_1)
assert is_integer(match_2)
assert match_2_predicate_fn.(match2)
```

aerosol:
> Consider this assertion
(...)
> The equivelent test to assert the same things would be something like:
(...)


@macobo Correct me if I'm wrong, but this is equivalent:

```elixir

assert %{"results" => [%{"metrics" => [3], "dimensions" => []}],
      "meta" => %{},
      "query" => %{
        "metrics" => ["pageviews"]
        "filters" => [],
        "dimensions" => [],
        "site_id" => site.domain,
        "date_range" => [_, _] = date_range,
        "order_by" => [_|_],
        "pagination" => %{},
        "include" => %{}
      }) = response


assert Enum.all?(date_range, &is_binary/1)
```

macobo:
> @macobo Correct me if I'm wrong, but this is equivalent:

That's really cool and almost works, covering the use-case I was trying to solve for. Here's actual working code:

```elixir
    site_domain = site.domain
    assert %{"results" => [%{"metrics" => [3], "dimensions" => []}],
      "meta" => %{},
      "query" => %{
        "metrics" => ["pageviewss"],
        "filters" => [],
        "dimensions" => [],
        "site_id" => ^site_domain,
        "date_range" => [_, _] = date_range,
        "order_by" => [_|_],
        "pagination" => %{},
        "include" => %{}
      }} = json_response(conn, 200)

    assert Enum.all?(date_range, &is_binary/1)
```

Two issues with this:
1. Pins only work for variables, hence the need to extract `site_domain` into a separate variable
2. It doesn't check there aren't any _more_ keys under nested maps

(2) is something we can very well live with. I'm however curious if there's some elegant solution for (1) besides extracting a variable above every match or having another separate assert below.

> @zoldar: Ideally, it would be be great to implement these expect checks as things embeddable in the match pattern directly, something like:

One problem with this sort of approach would be the output when the test fails being harder to follow. Both this proposal and @aerosol's counter-proposal above have the benefit of ex_unit creating nice highlighted diffs for nested object that tell exactly where the error is coming from.

zoldar:
> One problem with this sort of approach would be the output when the test fails being harder to follow. Both this proposal and @aerosol's counter-proposal above have the benefit of ex_unit creating nice highlighted diffs for nested object that tell exactly where the error is coming from.

Not necessarily. Metadata in AST could be potentially exploited/manipulated to make resulting assert error point at the expression in the pattern instead of the implicitly generated consecutive pattern. We could experiment and determine whether this is achievable.

> Pins only work for variables, hence the need to extract site_domain into a separate variable

That's exactly what came to my mind as well haha. Also worth an experiment, because I get the impression it's entirely feasible, especially given `Ecto.Query` does this since ever.

aerosol:
> Pins only work for variables, hence the need to extract site_domain into a separate variable

Yeah this is annoying. Although in practice, this is usually (idiomatically) solved with pattern matching as well.

```elixir
test "foo", %{site: %{domain: domain}} do
...
end

# or

%{domain: domain} = new_site()
```

> It doesn't check there aren't any more keys under maps

Arguably, this is hardly ever a problem in API tests - and you rarely want to update 20 tests because a new field has been added. 

For exhaustive checks of API return types, I like what OpenAPI offers. We use it in the Plugins API tests, e.g.:

```elixir
  resp = conn
        |> authenticate(site.domain, token)
        |> get(url)
        |> json_response(200)
        |> assert_schema("SharedLink", spec())
```

Where the schema (not the test!) is responsible for defining the types and payload properties. This has the advantage of also being enforced at runtime. 

> Diff highlighting (...)

Sure. What I'm saying is, I'm not opposed to the proposal, it can be improved for sure (as per @zoldar's comments), but at the same time I don't want us to use it religiously, where standard pattern matching can be leveraged to express the intent.

zoldar:
https://github.com/plausible/analytics/compare/assert-experiment?expand=1 here's an experiment I did to determine viability of approach with extending pattern match.

It's currently limited to a reporting a single predicate-based error at a time, but this can be worked out (probably with an extra layer of macro rewriting).

image

Unfortunately, due to the fact we have to assign the values from the original pattern match before we can check them, if any non-predicate patterns fail, they will always show without failed predicate patterns highlighted. They will always show after the "normal" pattern errors are fixed though. I think that would be an acceptable trade-off.

WDYT?

zoldar:
I've played around with the idea some more and managed to make the macro more expressive. For instance, with a test like this:

```elixir
  test "foobar" do
    n = %{z: 2}

    assert_matches(
      %{
        a: ^any(:integer, &(&1 > 2)),
        b: ^any(:string, ~r/baz/),
        d: [_ | _],
        e: ^~r/invalid/,
        f: ^n.z,
        g: ^&is_float/1
      } = %{
        a: 1,
        b: "twofer",
        c: :other,
        d: [1, 2, 3],
        e: "another string",
        f: 1,
        g: 4.2
      }
    )
  end
```

results in a following output now:

image

apata:
@zoldar I checked it out locally, and it's quite impressive! I still think we should be able to say that the output is X and no more than X. The following test passes at the moment:
```ex
  test "extraneous keys" do
    n = %{z: 2}
    assert_matches(
      %{
        f: ^n.z
      } = %{
        f: 2,
        __stack_trace: "...",
        __env: "..."
      }
    )
  end
```

zoldar:
@apata It's still possible, as you can pin any value to compare against:

```elixir
    n = %{z: 2}

    assert_matches(%{wrap: ^%{f: n.z}} = %{wrap: %{f: 2, _env: "..."}, other: :baz})
```

image

macobo:
The testing proposal from @zoldar looks great. Thanks for tackling that :muscle: 

Since continuing this PR with a new implementation + specification would cause confusion I'll close this PR and let @zoldar open a new one with the new proposal (no need to re-port any tests that were ported here).

PR #5020, opened 2025-01-27
Tracker: Use fetch with keepalive over beacon for pageleaves

macobo:
### Changes

:facepalm: I didn't realize pageleaves had their own tracking code. After this change fetch with keepalive is used instead of beacon.

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -114,10 +114,7 @@
     payload.h = 1
     {{/if}}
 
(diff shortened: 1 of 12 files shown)


RobertJoonas (approved):
Wow, can't believe I missed that too 🤯

Looks good to me other than the CI - including `data-allow-fetch` in all snippets of the relevant html test fixtures should do the trick.

PR #5022, opened 2025-01-27
Replace second line for filters with UI from segments variant D

apata:
### Changes

* Removes the second line for filters that's shown when saved_segments feature flag is true, replacing it with the single line version from segments variant D and the multicolumned filter menu from the same.
  * For users that don't have the feature flag (pretty much all of them), no visual changes are expected, except a few tiny tweaks that deal with what happens in the top bar in the overflow situation. 
* Refactors filter pills so their drop shadows wouldn't be cut off by overflow: hidden.
* Creates non-interactive filter pill and pill list versions (in anticipation of segment modals).

### Tests

### Changelog

### Documentation

### Dark mode

### Visuals
pr-5022 review plausible io_plausible io_f=is,page,__dashboard f=is,country,US f=is,region,US-IL l=US,United%20States l=US-IL,Illinois (1)


assets/js/dashboard/stats/current-visitors.js
```diff
@@ -56,7 +53,10 @@ export default function CurrentVisitors({ tooltipBoundary }) {
       >
         <AppNavigationLink
           search={(prev) => ({ ...prev, period: 'realtime' })}
-          className="h-9 flex items-center ml-1 mr-auto text-xs md:text-sm font-bold text-gray-500 dark:text-gray-300"
+          className={classNames(
```
apata, line 56: Final positioning is now with className prop

assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -1,38 +1,38 @@
...
 import { useQueryContext } from '../query-context'
+import { AppNavigationLink } from '../navigation/use-app-navigate'
+import { BUFFER_FOR_SHADOW_PX } from './filter-pill'
 
-const SEE_MORE_GAP_PX = 16
+const BUFFER_RIGHT_PX = 16 - BUFFER_FOR_SHADOW_PX - PILL_X_GAP
+const BUFFER_LEFT_PX = 16 - BUFFER_FOR_SHADOW_PX
 const SEE_MORE_WIDTH_PX = 36
+const SEE_MORE_RIGHT_MARGIN_PX = BUFFER_FOR_SHADOW_PX + PILL_X_GAP
+const SEE_MORE_LEFT_MARGIN_PX = BUFFER_FOR_SHADOW_PX
```
apata, line 20: The idea here is to have a constant separation equivalent to gap-x-4 (16) between pills and top bar left actions, top bar right actions.

These calculations are needed, because pills include a margin around them (BUFFER_FOR...), which if not taken into account, causes misalignment.

Width calculations include the pills and the gaps: 
`[pill1]  [pill2]  [pill3]` is cut right before [pill 2] ->
`[pill1]  `
 
That's why the buffers are different for left and right.

assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -123,96 +127,110 @@ export const FiltersBar = () => {
...
+            style={{
+              width: SEE_MORE_WIDTH_PX,
+              marginLeft: SEE_MORE_LEFT_MARGIN_PX,
+              marginRight: SEE_MORE_RIGHT_MARGIN_PX
+            }}
+            className="md:relative"
+            ref={seeMoreRef}
+            dropdownContainerProps={{
+              ['title']: opened ? 'Show less' : 'Show more',
+              ['aria-controls']: 'more-filters-menu',
```
ukutaht, line 191: We already have [headlessui](https://headlessui.com/)  as a dependency. Consider using the [dropdown menu component](https://headlessui.com/react/menu) over a hand-rolled dropdown to get things like aria properties for free.

ukutaht:
Problem: when a filter is added the page jumps vertically by a little bit, but enough to be annoying for me:

https://github.com/user-attachments/assets/c8ab4246-6597-47c1-ae92-8105d600507a

assets/js/dashboard/nav-menu/filter-pill.tsx
```diff
@@ -1,49 +1,87 @@
...
-          {children}
-        </span>
-      </AppNavigationLink>
-      <button
-        title={`Remove filter: ${plainText}`}
-        className="flex items-center h-full px-2 mr-1 cursor-pointer hover:text-indigo-700 dark:hover:text-indigo-500 "
-        onClick={() => onRemoveClick()}
+    <div className={className}>
+      <div
+        style={{ margin: BUFFER_FOR_SHADOW_PX }}
```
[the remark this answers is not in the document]
apata: I need the numeric values for calculations: I can't generate the custom classes dynamically because https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names. I could declare the string class names like BUFFER_CLASS = "m-[4px]", which TW detects properly, and then parse these class name strings to get the px values, but it's a bit more messy. 

Ideally, of course, I'd not need these weird calculations, and it's probably possible to get rid of them when either
A) scrapping the partially expanded filters mode (as in designs) in favor of current version: all shown or none shown
B) refactoring filter pills and pill lists more

assets/js/dashboard/nav-menu/filters-bar.tsx
apata: Agreed with the principle overall. Alas, headlessui doesn't align with accessibility standards (tab navigation between items) and that's why I refactored datepicker out of it last year. I wanted to keep new UI aligned with datepicker, thus I kept this custom component. 

Maybe we want to align with the storybook dropdown instead?

ukutaht: > headlessui doesn't align with accessibility standards (tab navigation between items)

What standard specifically? AFAIK accessible menus should be navigable via arrows and items should have tabindex="-1" which headlessUI does https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

apata: You're right, MDN is clear on that! My gut feel had solidified to a standard since the last time we talked about this: https://github.com/plausible/analytics/pull/4461#discussion_r1730930573

Still, the question remains which components to refactor and when. There's a few on the dashboard, but also https://plausible.io/storybook/dropdown is on tab navigation.

apata: This turned out to be a bit more involved than I hoped, but I think I managed to get some good results. I've prepped an additional PR against this branch:
https://github.com/plausible/analytics/pull/5037

ukutaht (approved):

ukutaht:
Approved assuming https://github.com/plausible/analytics/pull/5037 gets merged

PR #5024, opened 2025-01-28
Implement `assert_matches` with expanded expressions support in match assertions

zoldar:
### Changes

This is a proposal for implementing a more expressive match assertion mechanism. It was prompted by the original idea by @macobo in #5019.

The idea here is that the pin (^) operator does not only rebind existing binding in the scope but also allows embedding basically any other expression to match against the given part of the pattern. Normal pattern matching is also supported and both can be mixed. The only caveat so far is that when normal patterns fail, only they are listed in the error even if there are potentially failing expressions. However, once the normal pattern is fixed, they surface.

Currently, the following expressions can be pinned:

* `any(:atom)`
* `any(:string)`
* `any(:binary)`
* `any(:integer)`
* `any(:float)`
* `any(:boolean)`
* `any(:map)`
* `any(:list)`
* all above variants of `any` with a one argument predicate function accepting value and returning a boolean, like: `any(:integer, & &1 > 20)`
* a special case of `any(:string, ~r/regex pattern/)` checking that value is a string and matches a pattern
* shorthand version of the above, `~r/regex pattern`
* any artibrary one argument function returning a boolean, like `&is_float/1` or `&(&1 < 40 or &1 > 300)`
* `exact(expression)` where expression is compared using equality, so that can enforce full equality inside a pattern, like: `exact(%{foo: 2})` which will fail if the value is something like `%{foo: 2, other: "something}`
* any other arbitrary expression which is compared the way as if it was wrapped with `exact` - this allows "interpolating" values from schemas and maps without rebinding like `user.id` (instead of having to rebind to `user_id` first)

```elixir
    n = %{z: 2}

    assert_matches %{
                     a: ^any(:integer, &(&1 > 2)),
                     b: ^any(:string, ~r/baz/),
                     d: [_ | _],
                     e: ^~r/invalid/,
                     f: ^n.z,
                     g: ^(&is_float/1),
                     h: ^exact(%{foo: :bar})
                   } = %{
                     a: 1,
                     b: "twofer",
                     c: :other,
                     d: [1, 2, 3],
                     e: "another string",
                     f: 1,
                     g: 4.2,
                     h: %{foo: :bar, other: "stuff"}
                   }
  end
```

image

Although the macro seems to be relatively feature complete, more testing against some real world scenarios must be done before merging.

### DIFF
--- a/test/support/assert_matches.ex
+++ b/test/support/assert_matches.ex
@@ -17,10 +17,16 @@ defmodule Plausible.AssertMatches do
     * `any(:string)`
     * `any(:binary)`
     * `any(:integer)`
+    * `any(:pos_integer)`
+    * `any(:number)`
     * `any(:float)`
     * `any(:boolean)`
     * `any(:map)`
     * `any(:list)`
+    * `any(:tuple)`
+    * `any(:iso8601_date)`
+    * `any(:iso8601_datetime)`
+    * `any(:iso8601_naive_datetime)`
     * all above variants of any with a one argument predicate function accepting
       value and returning a boolean, like: `any(:integer, & &1 > 20)`
     * a special case of `any(:string, ~r/regex pattern/)` checking that value is
@@ -344,10 +350,45 @@ defmodule Plausible.AssertMatches do
     def any(:string), do: &is_binary/1
     def any(:binary), do: &is_binary/1
     def any(:integer), do: &is_integer/1
+    def any(:number), do: &is_number/1
     def any(:float), do: &is_float/1
     def any(:boolean), do: &is_boolean/1
     def any(:map), do: &is_map/1
     def any(:list), do: &is_list/1
+    def any(:tuple), do: &is_tuple/1
+
+    def any(:pos_integer) do
+      fn value ->
+        is_integer(value) and value > 0
+      end
+    end
+
+    def any(:iso8601_date) do
+      fn value ->
+        case Date.from_iso8601(value) do
+          {:ok, _} -> true
+          _ -> false
+        end
+      end
+    end
+
+    def any(:iso8601_datetime) do
+      fn value ->
+        case DateTime.from_iso8601(value) do
+          {:ok, _} -> true
+          _ -> false
+        end
+      end
+    end
+
+    def any(:iso8601_naive_datetime) do
+      fn value ->
+        case NaiveDateTime.from_iso8601(value) do
+          {:ok, _} -> true
+          _ -> false
+        end
+      end
+    end
 
     def any(:string, %Regex{} = regex) do
       fn value ->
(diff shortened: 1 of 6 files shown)


test/support/assert_matches.ex
```diff
@@ -0,0 +1,214 @@
...
+      end
+    end
+
+    def regex(regex) do
+      fn value ->
+        String.match?(value, regex)
+      end
+    end
+
+    def exact(expr) do
```
macobo, line 208: Suggestion: Let's name this `exactly`. When testing this out this seemed to roll better off the tongue/keyboard.

zoldar: https://github.com/plausible/analytics/pull/5024/commits/8b2e5f44722520a5f1f6eef87e9067084fa77905

zoldar:
@macobo suggested it would be good to have a way to enforce exact set of keys within a potentially nested map pattern.

I _think_ it's feasible with an extra pass in the macro, and the syntax for such pattern would look roughly like:

```elixir
      assert_matches  %{
        "results" => [%{"dimensions" => ["some phrase"], "metrics" => [1, 1, 100.0]}],
        "meta" => ^strict_map(%{
          "imports_included" => false,
          "imports_skip_reason" => "unsupported_query",
          "imports_warning" =>
            ~r/Imported stats are not included in the results because query parameters are not supported/
        }),
        "query" =>
          response_query(site, %{
            "metrics" => ["visitors", "events", "conversion_rate"],
            "dimensions" => ["event:props:search_query"],
            "filters" => [
              ["is", "event:goal", ["WP Search Queries"]]
            ],
            "include" => %{"imports" => true}
          })
      } = json_response(conn, 200)
```

the idea here is that ^strict_map pins are matched first (with a check boiling down to `map_size(value) == count_of_keys_in_pattern`). Then the `^strict_map` is removed from the pattern with its contents in the AST and passed on to the existing check. I'll have a stab at it some time later on.

zoldar:
Alright, I think I got it functionally at least, the code's still not pretty:

```elixir
        assert_matches ^strict_map(%{
                         "id" => ^any(:integer),
                         "name" => "Some segment",
                         "type" => ^"#{unquote(type)}",
                         "segment_data" =>
                           ^strict_map(%{"filters" => [["is", "visit:entry_page", ["/blog"]]]}),
                         "owner_id" => ^user.id,
                         "inserted_at" => ^any(:string),
                         # "updated_at" => ^any(:string)
                       }) = response
```

The report:

image

zoldar:
Also, added some additional built-in checks for common data types and date/time string formats, because why not:

```elixir
        assert_matches ^strict_map(%{
                         "id" => ^any(:pos_integer),
                         "name" => "Some segment",
                         "type" => ^"#{unquote(type)}",
                         "segment_data" =>
                           ^strict_map(%{"filters" => [["is", "visit:entry_page", ["/blog"]]]}),
                         "owner_id" => ^user.id,
                         "inserted_at" => ^any(:iso8601_naive_datetime),
                         "updated_at" => ^any(:iso8601_naive_datetime)
                       }) = response
```

zoldar:
@apata @macobo It's ready if you'd like to give it a spin.

PR #5025, opened 2025-01-28
Use `fetch` over `XMLHttpRequest` in plausible.js

macobo:
`fetch` allows us to use the keepalive flag, which more reliably delivers data to us. It's also required for reliable scroll depth capturing down the line.

The only major browser not to support fetch is Internet Explorer. Clients who care about IE traffic can use the `.compat` variant of our script (documented in
https://plausible.io/docs/script-extensions#scriptcompatjs). As of this commit, IE makes up less than ~0.003% of incoming traffic to plausible.

@ukutaht should we make an announcement on the product side or is the changelog sufficient?


ukutaht (approved):
Looks good 👍 I think the changelog entry is sufficient.

PR #5027, opened 2025-01-29
Update goal in segment

ukutaht:
When a goal's display name is updated, this PR makes it so the segments containing that goal are also updated accordingly

### DIFF
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -59,11 +59,12 @@ defmodule Plausible.Goals do
     site = Repo.preload(goal, :site).site
 
     with :ok <- maybe_check_feature_access(site, changeset),
-         {:ok, goal} <- Repo.update(changeset) do
+         {:ok, updated_goal} <- Repo.update(changeset),
+         :ok <- Plausible.Segments.update_goal_in_segments(goal, updated_goal) do
       on_ee do
-        {:ok, Repo.preload(goal, :funnels)}
+        {:ok, Repo.preload(updated_goal, :funnels)}
       else
-        {:ok, goal}
+        {:ok, updated_goal}
       end
     end
   end
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -64,9 +64,9 @@ defmodule Plausible.Goals do
            {:ok, updated_goal} <- Repo.update(changeset),
            :ok <- Plausible.Segments.update_goal_in_segments(site, goal, updated_goal) do
         on_ee do
-          {:ok, Repo.preload(updated_goal, :funnels)}
+          Repo.preload(updated_goal, :funnels)
         else
-          {:ok, updated_goal}
+          updated_goal
         end
       end
     end)
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -56,17 +56,20 @@ defmodule Plausible.Goals do
           {:ok, Goal.t()} | {:error, Ecto.Changeset.t()} | {:error, :upgrade_required}
   def update(goal, params) do
     changeset = Goal.changeset(goal, params)
-    site = Repo.preload(goal, :site).site
 
-    with :ok <- maybe_check_feature_access(site, changeset),
-         {:ok, updated_goal} <- Repo.update(changeset),
-         :ok <- Plausible.Segments.update_goal_in_segments(goal, updated_goal) do
-      on_ee do
-        {:ok, Repo.preload(updated_goal, :funnels)}
-      else
-        {:ok, updated_goal}
+    Repo.transaction(fn ->
+      site = Repo.preload(goal, :site).site
+
+      with :ok <- maybe_check_feature_access(site, changeset),
+           {:ok, updated_goal} <- Repo.update(changeset),
+           :ok <- Plausible.Segments.update_goal_in_segments(goal, updated_goal) do
+        on_ee do
+          {:ok, Repo.preload(updated_goal, :funnels)}
+        else
+          {:ok, updated_goal}
+        end
       end
-    end
+    end)
   end
 
   def find_or_create(
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -62,7 +62,7 @@ defmodule Plausible.Goals do
 
       with :ok <- maybe_check_feature_access(site, changeset),
            {:ok, updated_goal} <- Repo.update(changeset),
-           :ok <- Plausible.Segments.update_goal_in_segments(goal, updated_goal) do
+           :ok <- Plausible.Segments.update_goal_in_segments(site, goal, updated_goal) do
         on_ee do
           {:ok, Repo.preload(updated_goal, :funnels)}
         else
--- a/lib/plausible/segments/segments.ex
+++ b/lib/plausible/segments/segments.ex
@@ -168,8 +168,12 @@ defmodule Plausible.Segments do
 
       updated_segment_data = Map.put(segment.segment_data, "filters", updated_filters)
 
-      Segment.changeset(segment, %{segment_data: updated_segment_data})
-      |> Repo.update!()
+      from(
+        s in Segment,
+        where: s.id == ^segment.id,
+        update: [set: [segment_data: ^updated_segment_data]]
+      )
+      |> Repo.update_all([])
     end
 
     :ok
--- a/lib/plausible/segments/segments.ex
+++ b/lib/plausible/segments/segments.ex
@@ -136,6 +136,8 @@ defmodule Plausible.Segments do
         %Plausible.Goal{} = stale_goal,
         %Plausible.Goal{} = updated_goal
       ) do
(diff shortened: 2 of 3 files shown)


test/plausible/goals_test.exs
```diff
@@ -117,6 +117,37 @@ defmodule Plausible.GoalsTest do
...
+        "type" => "site",
+        "segment_data" => %{
+          "filters" => [
+            ["is", "event:page", ["/blog"]],
+            ["is", "event:goal", ["Signup from nav", "Signup"]],
+            ["is", "event:props:variant", ["A"]]
+          ]
+        },
+        "name" => "Segment name"
+      })
```
apata, line 137: **nitpick, non-blocking:** In some other cases, we're using Plausible.Factory to populate data. `insert(:segment, owner: user, site: site, ...)` would save you the trouble of declaring user role.

test/plausible/goals_test.exs
```diff
@@ -117,6 +117,37 @@ defmodule Plausible.GoalsTest do
...
+    Goals.update(goal1, %{"display_name" => "SIGNUP"})
+    updated_segment = Repo.reload!(segment)
+
+    assert updated_segment.segment_data == %{
+             "filters" => [
+               ["is", "event:page", ["/blog"]],
+               ["is", "event:goal", ["Signup from nav", "SIGNUP"]],
+               ["is", "event:props:variant", ["A"]]
+             ]
+           }
```
apata, line 148: **issue**: Noticed that we also override segment `updated_at`. 

If that's intentional, it would be good to assert it here as well, with something like `segment.updated_at < updated_segment.updated_at`.

Alternatively, we may want to not override updated_at on these automatic updates. May be confusing when the message in the UI (`Last updated at <stamp> by <name>`) doesn't match with their usual online time. 

Or we could also just rephrase it in the UI: `Last updated at <stamp>, belongs to <name>` or similar.

lib/plausible/segments/segments.ex
```diff
@@ -131,6 +131,50 @@ defmodule Plausible.Segments do
...
 
+  def update_goal_in_segments(
+        %Plausible.Site{} = site,
+        %Plausible.Goal{} = stale_goal,
+        %Plausible.Goal{} = updated_goal
+      ) do
+    goal_filter_regex =
+      ~s(.*?\\["is",\s*"event:goal",\s*\\[.*?"#{Regex.escape(stale_goal.display_name)}".*?\\]\\].*?)
+
+    IO.inspect(stale_goal)
```
apata, line 142: Flagging this just in case

test/plausible/goals_test.exs
```diff
@@ -117,6 +117,37 @@ defmodule Plausible.GoalsTest do
     assert goal2.display_name == "Homepage"
   end
 
+  test "update/2 also updates any segments the goal is a part of" do
```
apata, line 120: **ultranitpick, non-blocking**: test says any segments, but there's only one segment the case

lib/plausible/segments/segments.ex
apata: credo also flags it, resolving

test/plausible/goals_test.exs
aerosol: The reason we're using contextual interfaces for Goals setup (above) is that much of the insertion logic is abstracted away. I'm not a big fan of using factories for anything slightly complex - otherwise an unverified contract emerges (we assume insert(s) made in tests are equivalent to actual creation routine). Not sure how it's like with segments, but it is certainly more reliable to use `Goals.create` for instance. Just my 2c.

lib/plausible/goals/goals.ex
```diff
@@ -56,16 +56,20 @@ defmodule Plausible.Goals do
...
     changeset = Goal.changeset(goal, params)
-    site = Repo.preload(goal, :site).site
 
-    with :ok <- maybe_check_feature_access(site, changeset),
-         {:ok, goal} <- Repo.update(changeset) do
-      on_ee do
-        {:ok, Repo.preload(goal, :funnels)}
-      else
-        {:ok, goal}
+    Repo.transaction(fn ->
```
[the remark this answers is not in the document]
apata: Migrating segments automatically on goal name change is a bit of a workaround to not having goal IDs in filters. 
We're betting on the fact that when user is composing filters like `[is, event:goal, <goal-display-name>]`, they don't mean `[contains, event:goal, <goal-display-name>]`.

On deleting goals, the plan was to just let the user know that it's in <count> segments and that these will not filter right until a goal with the same name is added back again, or the segments manually changed.

lib/plausible/segments/segments.ex
```diff
@@ -131,6 +131,50 @@ defmodule Plausible.Segments do
     end
   end
 
+  def update_goal_in_segments(
+        %Plausible.Site{} = site,
+        %Plausible.Goal{} = stale_goal,
+        %Plausible.Goal{} = updated_goal
+      ) do
+    goal_filter_regex =
+      ~s(.*?\\["is",\s*"event:goal",\s*\\[.*?"#{Regex.escape(stale_goal.display_name)}".*?\\]\\].*?)
```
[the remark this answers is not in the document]
ukutaht: This is the way it's formatted, pulled straight out of postgres:

```
[["is", "event:goal", ["Outbound Link: Click"]]]
```

The current regex matches this correctly. It feels a bit brittle but at the same time if the library we use changes its formatting then we would get a test failure.

lib/plausible/goals/goals.ex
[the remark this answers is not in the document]
ukutaht: > Wrapping this with Repo.transaction means, that it now returns {:ok, result} on success - so {:ok, {:ok, goal}} while {:ok, goal} is expected.

Thanks for the heads up @aerosol, fixed in https://github.com/plausible/analytics/pull/5027/commits/f5498824595bac98f5081caaa597dd01197d3dfd

> I'm out of the loop with this, but AFAIU Segments don't set constraints on any goals they depend on, so should we also handle deleting a goal?

Having discussed it we didn't land on a good idea on how to deal with deletion. The user has to decide whether to also delete the segments that contain the goal, or manually fix them.

Related to this discussion I realized that not doing anything will render the segment filters invalid. @apata I think we've been working off the assumption that by leaving a non-existing goal in the segment we would get an empty dashboard. But now I remember that there's a validation step that makes stats queries with non-existing goals invalid.

My proposal for this is to remove the goal validation and instead return 0s when a non-existing goal is used in a stats query. 2 reasons why:
1. More consistent with other filters where we return 0s for non-existing entries
2. Longer term plan is to move to a system where custom event goals show up automatically, mixed with goals that require manual setup (pageview goals, upcoming scroll depth goals). With that in mind we won't be able to know during query-building whether a goal exists or not, just like we don't know whether a page or source exists during query-building.

test/plausible/goals_test.exs
ukutaht: Expanded the test to make sure it can update multiple segments: https://github.com/plausible/analytics/pull/5027/commits/7c5a436fefc3c3b252fe6e4aaf995e815359b056

lib/plausible/goals/goals.ex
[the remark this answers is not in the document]
apata: We do get an empty dashboard https://plausible.io/plausible.io?f=is,goal,definitely_no_exist, but over the v2 Stats API, we get 
```
{
  "error": "The goal `definitely_no_exist` is not configured for this site. Find out how to configure goals here: https://plausible.io/docs/stats-api#filtering-by-goals"
}
```

lib/plausible/segments/segments.ex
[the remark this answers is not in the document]
apata: Hmm, so it doesn't matter how many clauses there are, how lengthy the clauses are, or how deep the filter tree is? Because we'd get the warning from our tests only if the wrapping appears for the test case.

apata (approved):

lib/plausible/segments/segments.ex
```diff
@@ -131,6 +131,50 @@ defmodule Plausible.Segments do
     end
   end
 
+  def update_goal_in_segments(
+        %Plausible.Site{} = site,
+        %Plausible.Goal{} = stale_goal,
+        %Plausible.Goal{} = updated_goal
+      ) do
+    goal_filter_regex =
```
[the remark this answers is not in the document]
ukutaht: Ugh, I spent quite a bit of time trying to replace the regex with a `jsonpath` query in postgres. It works great from the console but `Postrex.DefaultTypes` doesn't support it and would require a custom Ecto type to work. So I opted for leaving the regex and commenting it here: https://github.com/plausible/analytics/pull/5027/commits/b80aa2d254a14e177a84174e9b2e2eba19e07764

test/plausible/goals_test.exs
ukutaht: You're right, I think it's better to avoid bumping the timestamp: https://github.com/plausible/analytics/pull/5027/commits/b9fcc223f33220c859d91199dc6c0c994b123b82

aerosol (approved):
Let's go! :rocket: 

PS: @ukutaht If you have any notes/snippets from your jsonpath attempts, please share so maybe I could have a look.

PR #5029, opened 2025-01-29
Page Scroll Goals

RobertJoonas:
### Required migration steps

Migrating from the old unique index to the new one is a bit tricky, requiring multiple steps:

1. Migration: Add `scroll_threshold` column to `Goal` and add the new unique index. https://github.com/plausible/analytics/pull/5033
2. Minimal schema change: Make Ecto aware of the new unique_constraint. https://github.com/plausible/analytics/pull/5034
3. Migration: Drop the old index. https://github.com/plausible/analytics/pull/5036

Once the old index is dropped, this PR is ready to go.

### Changes

This PR implements page scroll goals.

A page scroll goal can be created by simply adding a value into an optional `scroll_threshold` field when creating a pageview goal:

image

A user can filter and break down page scroll goals like any other goals.

_TODO: figure out what to do with the total conversions metric_

### Tests

### Changelog

### Documentation

### Dark mode


priv/repo/migrations/20250128161815_add_scroll_threshold_to_goals.exs
```diff
@@ -0,0 +1,39 @@
+defmodule Plausible.Repo.Migrations.AddScrollThresholdToGoals do
```
macobo, line 1: We should add a constraint that scroll_threshold can only be set if page_path is set.

priv/repo/migrations/20250128161815_add_scroll_threshold_to_goals.exs
```diff
@@ -0,0 +1,39 @@
...
+
+  # Plausible.Repo.Migrations.GoalsUnique
+  @old_index unique_index(
+               :goals,
+               [:site_id, :page_path],
+               where: "page_path IS NOT NULL",
+               name: :goals_page_path_unique
+             )
+
+  @new_index unique_index(
```
macobo, line 15: Sanity check: Multiple scroll goals for the same page path make sense, right?

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -193,6 +196,17 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
         x-data="{ firstFocus: true }"
         x-on:focus="if (firstFocus) { $el.select(); firstFocus = false; }"
       />
+
+      <.input
```
macobo, line 200: TODO (Myself): Check this is properly optional when testing.

lib/plausible/stats/sql/query_builder.ex
```diff
@@ -139,18 +139,61 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
...
+      types: goal_types,
+      event_names: goal_event_names,
+      page_regexes: goal_page_regexes,
+      scroll_thresholds: goal_scroll_thresholds
+    } =
+      query.preloaded_goals.matching_toplevel_filters
+      |> Plausible.Goals.decompose()
 
     from(e in q,
-      join: goal in Expression.event_goal_join(events, page_regexes),
```
macobo, line 146: This requires further work on my end. Some initial thoughts:

- `event_goal_join` is now unused. Could we move this huge expression + maybe the decompose call out of this file?
- Are we doing repetitive operations per events_v2 row for each site goal in the CASE below? If possible we should avoid that work somehow as it's costly CPU-wise - e.g. by doing concat instead of filter?
- Defining variables in WHERE clauses is generally an anti-pattern especially with the JOIN - figure out if we can move the whole expression to a JOIN.

I'll post a new comment once I have actionable suggestions or a proposal.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -4156,6 +4156,169 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
...
+          "site_id" => site.domain,
+          "metrics" => ["visitors", "events", "conversion_rate"],
+          "filters" => [["is", "event:page", ["/blog"]]],
+          "date_range" => "all",
+          "dimensions" => ["event:goal"]
+        })
+
+      assert json_response(conn, 200)["results"] == [
+               %{"dimensions" => ["Visit /blog"], "metrics" => [2, 2, 100.0]},
+               # TODO: make it possible to query total conversions from pageleave events
```
macobo, line 4202: TODOs in tests are generally not actionable or won't come up refactoring. Let's move the TODO to e.g. lib/plausible/stats/sql/expression.ex

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
```diff
@@ -4156,6 +4156,169 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTest do
...
+          user_id: 56,
+          pathname: "/blog/jane-post",
+          timestamp: ~N[2021-01-01 00:00:10],
+          scroll_depth: 75,
+          "meta.key": ["author"],
+          "meta.value": ["jane"]
+        )
+      ])
+
+      # Double-dimension breakdown
```
macobo, line 4273: Let's create separate tests instead of multiple ones in the same one. Huge tests are a pain to understand and debug.

I assume the reason they're bunched up here is because of shared test data setup? If so, worth considering using `setup [:foo]` and describe blocks, maybe even splitting these tests out to a separate file as we do with e.g. test/plausible_web/controllers/api/external_stats_controller/query_goal_dimension_test.exs

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -36,15 +40,108 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
...
+      insert(:goal, %{
+        site: site,
+        page_path: "/blog/posts/1",
+        scroll_threshold: 75,
+        display_name: "Scroll 75 /blog/posts/1"
+      })
+
+      conn = get(conn, "/api/stats/#{site.domain}/conversions?period=day&date=2020-01-01")
+
+      # TODO: make it possible to query `events` metric
```
macobo, line 132: TODO in tests: Consider moving this to lib/plausible/stats/sql/expression.ex

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -193,6 +196,17 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+        :if={Plausible.Stats.ScrollDepth.feature_visible?(@site, @current_user)}
+        label="Scroll Depth Threshold"
+        field={@f[:scroll_threshold]}
+        type="number"
+        value={if @goal && @goal.scroll_threshold > -1, do: @goal.scroll_threshold, else: nil}
+        min="0"
+        max="100"
+        step="1"
+      />
```
macobo, line 209: This element doesn't have placeholder text indicating it's optional. As-is I think it will confuse users into thinking they need to input something.

macobo (commented):
Some initial thoughts on the PR. Will dig into the query side of this in-depth in another pass.

lib/plausible/goals/filters.ex
```diff
@@ -106,33 +106,44 @@ defmodule Plausible.Goals.Filters do
...
     Enum.reduce(filtered_goals, false, fn goal, dynamic_statement ->
-      case goal do
-        nil ->
+      cond do
+        is_nil(goal) ->
           dynamic([e], ^dynamic_statement)
 
-        %Plausible.Goal{event_name: event_name} when is_binary(event_name) ->
-          dynamic([e], e.name == ^event_name or ^dynamic_statement)
+        Plausible.Goal.type(goal) == :event ->
```
macobo, line 113: Nit: calling type 3 times is slightly unneccessary.

Given we already have `page_scroll_goal_condition`/`pageview_goal_condition` consider creating `defp goal_condition(type, goal, imported?)` and pattern matching within that.

lib/plausible/stats/sql/query_builder.ex
```diff
@@ -139,18 +139,61 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
...
+          notEmpty(
+            arrayFilter(
+              goal_index ->
+                CASE
+                  WHEN ?[goal_index] = 'event' THEN
+                      ? = ?[goal_index]
+                  WHEN ?[goal_index] = 'page' THEN
+                      ? = 'pageview' AND match(?, ?[goal_index])
+                  ELSE
+                      ? = 'pageleave' AND match(?, ?[goal_index]) AND ? >= ?[goal_index]
```
macobo, line 165: Note: This does not handle the scroll_depth=255 case we added for invalid data.

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
RobertJoonas: From my experience with testing Plausible stats queries, it's even more of a pain if multiple tests share the same `populate_stats` setup because it moves the `populate_stats` (the whole logic of the test) even further from the actual test.

I've simply duplicated the same populate stats code into every test separately in https://github.com/plausible/analytics/pull/5029/commits/1848d4475cbb9b9d7a52ab86fc469fbc964f39ef. Also moved to `query_goal_dimension_test.exs` as you suggested.

priv/repo/migrations/20250128161815_add_scroll_threshold_to_goals.exs
RobertJoonas: Good call! I've done that in the `Goal` schema. https://github.com/plausible/analytics/pull/5029/commits/7efca9202590bd5ad9cfda580e570235f1afc10b

test/plausible_web/controllers/api/external_stats_controller/query_test.exs
RobertJoonas: Done in https://github.com/plausible/analytics/pull/5029/commits/d19fe6553161981c1e142436b2a1be9b1ba570f7

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: I didn't want to get too involved with the UI yet, but the point is valid. For now, I've added just the text `(optional)` above the field.

What I imagined as the end result was something like revenue currency field in the custom event goal creation form - you flip a switch and a field becomes visible. It could then display 90 as the placeholder (GA4 built-in scroll event threshold).

lib/plausible/goals/filters.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5029/commits/ec5f801635a50d52372676350ce913b9185a7468

lib/plausible/stats/sql/expression.ex
```diff
@@ -196,6 +196,7 @@ defmodule Plausible.Stats.SQL.Expression do
     })
   end
 
+  # TODO: make it possible to query events from pageleave events (total conversions for page scroll goals)
```
macobo, line 199: I think this is doable if we also pass `query` to this method.

Basically something like:

```
if goal_dimension?(query) do
  count()
else 
  countIf(? != 'pageleave')
end
```

Reasoning:
- If we are grouping by goal, we will be filtering each match by the appropriate event already
- If we're not, then events is already kind of undefined and can match both custom events and pageviews, but by contract we don't want to include pageviews

This may change if we go with engagements? I wonder if events is the right name for what we want - perhaps it's actually number of visits? As such I don't think it's worth building this out until we have the data to know which way we're going.

lib/plausible/stats/sql/query_builder.ex
RobertJoonas: Oh yes. Thanks for noting that! https://github.com/plausible/analytics/pull/5029/commits/1ac92fbe7601ad2696af7eac80e983e85cd244f5

lib/plausible/stats/sql/expression.ex
RobertJoonas: > This may change if we go with engagements? I wonder if events is the right name for what we want - perhaps it's actually number of visits? As such I don't think it's worth building this out until we have the data to know which way we're going.

Exactly what I was thinking. Let's see what we'll land on.

lib/plausible/stats/sql/query_builder.ex
macobo: @RobertJoonas I pushed an update to this branch that:
1. Moves the join code back under `event_goal_join`
2. Uses `multiMatchAllIndices`, `arrayIntersect` and mapping to find the matching goals.
3. Refactored `Plausible.Goals.decompose` -> `Plausible.Stats.Goals.goal_join_data(query)` to be more in-line with the rest of the logic and placing the code close to what it depends on.

Note I did not update imports query to follow same pattern. Logic being that was hard to do and not urgent since the data in that table is already rolled up so the simpler code is a boon.

(I'll keep an eye on tests and push until they pass).

lib/plausible_web/live/goal_settings/form.ex
macobo: Should I skip reviewing the UI/user flow in-depth for now until a follow-up PR?

lib/plausible/stats/goals.ex
```diff
@@ -106,33 +151,43 @@ defmodule Plausible.Goals.Filters do
...
+  end
 
-    page_condition =
-      if String.contains?(page_path, "*") do
-        dynamic([e], fragment("match(?, ?)", field(e, ^db_field), ^page_regex(page_path)))
-      else
-        dynamic([e], field(e, ^db_field) == ^page_path)
-      end
+  defp goal_condition(:scroll, goal, _) do
+    pathname_condition = page_path_condition(goal.page_path, _imported? = false)
```
macobo, line 168: Is it possible to end up here with imported? = true? If so, this should return `false` in that case right? 
If it's not, let's make the method signature reflect the invariant that `imported = true`.

lib/plausible/stats/goals.ex
```diff
@@ -106,33 +151,43 @@ defmodule Plausible.Goals.Filters do
...
+    name_condition = dynamic([e], e.name == "pageleave")
+
+    scroll_condition =
+      dynamic([e], e.scroll_depth <= 100 and e.scroll_depth >= ^goal.scroll_threshold)
+
+    dynamic([e], ^pathname_condition and ^name_condition and ^scroll_condition)
```
macobo, line 174: I find the nesting causes more confusion than it helps with. What do you think about something like the following: 

```elixir
dynamic([e],
  e.name == "pageleave" and
  e.scroll_depth >= ^goal.scroll_threshold and e.scroll_depth <= 100 and
  ^path_condition(...)
```

lib/plausible/stats/goals.ex
```diff
@@ -106,33 +151,43 @@ defmodule Plausible.Goals.Filters do
...
+  end
+
+  defp goal_condition(:page, goal, imported?) do
+    pathname_condition = page_path_condition(goal.page_path, imported?)
+    name_condition = if imported?, do: true, else: dynamic([e], e.name == "pageview")
```
macobo, line 179: I think two separate defs would be a lot cleaner in this case.

```elixir
def goal_condition(:page, goal, true = _imported), do: page_path_condition(goal.page_path, true)

def goal_condition(:page, goal, false = _imported) do 
  dynamic([e], e.name == "pageview" and ^page_path_condition(goal.page_path, false))
end
```

test/plausible_web/controllers/api/external_stats_controller/query_goal_dimension_test.exs
```diff
@@ -130,4 +130,365 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryGoalDimensionTest do
              ]
     end
   end
+
+  describe "page scroll goals" do
```
macobo, line 134: I think we're missing tests in test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs for:
1. Tests filtering with scroll goals filter
2. Tests doing event:goal group by with scroll goals

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: User flow for *creating a goal* can be ignored at the current stage, yes. However, the user flow of how that goal behaves on the dashboard after it's been created is a crucial part of this PR.

lib/plausible/stats/goals.ex
RobertJoonas: Depends on the readers' preference. To me personally, extracting long boolean expressions like these into named variables helps grasp the intent quicker when reading the code. Changing this would only be cosmetic, so I'd leave it like that for now if it's OK with you.

lib/plausible/stats/goals.ex
RobertJoonas: Agreed! https://github.com/plausible/analytics/pull/5029/commits/9c51ba801e39fb2af93d23d18580b21af7cf22fc

test/plausible_web/controllers/api/external_stats_controller/query_goal_dimension_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/5029/commits/f8914e6961e60123070890a1649433069d25717c

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
```diff
@@ -1358,8 +1358,162 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryImportedTest do
...
+      insert(:goal,
+        site: site,
+        page_path: "/blog",
+        scroll_threshold: 50,
+        display_name: "Scroll /blog 50"
+      )
+
+      insert(:goal, site: site, page_path: "/blog**")
+
+      site_import = insert(:site_import, site: site)
```
macobo, line 1379: Redundant code in all the tests - you could do above 
```
    test "rejects imported data with warning when page scroll goal is filtered by", %{
      conn: conn,
      site: site,
      site_import: site_import
    } do
```

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
```diff
@@ -1358,8 +1358,162 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryImportedTest do
...
+      assert %{
+               "results" => results,
+               "meta" => %{
+                 "imports_included" => false,
+                 "imports_skip_reason" => "unsupported_query",
+                 "imports_warning" => @unsupported_query_warning
+               }
+             } = json_response(conn, 200)
+
+      assert results == [%{"dimensions" => [], "metrics" => [1, 0, 100.0]}]
```
macobo, line 1414: Nitpick: This could be inlined in the previous assert? Or is this intentional?

test/plausible_web/controllers/api/external_stats_controller/query_goal_dimension_test.exs
macobo: I think we're still missing a test showing the behavior with imported data + conversion rate + event group by.

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
RobertJoonas: This is more explicit. Inlining would be pattern matching which means there could be more keys in the map. Not a huge issue to inline it but not a concern to keep it this way either.

RobertJoonas:
> I think we're still missing a test showing the behavior with imported data + conversion rate + goal group by.

@macobo test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs:1468 is testing it. If you meant to say imported data + *group* conversion rate + goal group by - that's not possible with imported data, as mixing multiple dimensions will make the query ineligible for imported data.

macobo (approved):
Would be happy if the testing nit were resolved but other than that looks great!

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/5029/commits/2ab0ff9ffe8da0ff687f7f383ed009dd4f3de8ed

PR #5031, opened 2025-01-29
Don't preload goal settings modal to avoid issues with opening in edit mode

zoldar:
This PR fixes an issue where, on a fresh view load, when trying to open edit dialog for an existing goal, the form is not loaded properly. The preloading optimisation (meant load modal contents in the background in a  state defined on page load - in this case, an empty form) breaks it, as first open does not force contents to be reloaded when they should. The modal preopen routine can probably be improved to account for this case with a switch of sorts but currently this is the simplest solution for that particular case.

All the other current uses of the modal component do not have "edit" state currently (Funnel settings still rely on a custom-made solution).

### DIFF
--- a/lib/plausible_web/live/components/modal.ex
+++ b/lib/plausible_web/live/components/modal.ex
@@ -131,7 +131,12 @@ defmodule PlausibleWeb.Live.Components.Modal do
 
   @impl true
   def update(assigns, socket) do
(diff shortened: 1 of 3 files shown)


zoldar:
Converting this PR back to draft as there's a test failing still.

zoldar:
It's finally fixed and ready for review.

aerosol (approved):

PR #5032, opened 2025-01-29
Looker studio

RobertJoonas:
### Changes

Companion PR: https://github.com/plausible/website/pull/742 (should be merged together)

* Adds "Looker Studio Connector" as a plan benefit equivalent to the Stats API feature, meaning that:
  * for all plans after the 3rd generation, it's considered a business feature
  * for grandfathered (v1, v2, v3) plans, it's considered a growth feature
* Adds a new "Google Looker Studio Connector" section into site settings > integrations with just a few bits of info and a docs link. Will be displayed only in Enterprise Edition, as the connector is unavailable for self-hosters.

image

image

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/templates/site/settings_integrations.html.heex
```diff
@@ -5,6 +5,8 @@
     search_console_domains={@search_console_domains}
   />
 
+  <PlausibleWeb.Components.Site.SettingsLookerStudio.render />
```
aerosol, line 8:
```suggestion
 <PlausibleWeb.Components.Site.SettingsLookerStudio.render :if={ee?()} />
```

would something like this work instead of an empty render clause?

RobertJoonas: Yes, that works, but only if we define the function on both `ee` and `ce`. Otherwise, a compile warning will surface on `ce` about this function not existing.

RobertJoonas: https://github.com/plausible/analytics/pull/5032/commits/bd86634a80f6d3cbf1f185e988826c7ca318e377

aerosol (approved):

PR #5035, opened 2025-01-30
Scroll depth: update `site.scroll_depth_visible_at` in ingestion

macobo:
We decide what metrics to show dynamically based on whether site has sent us scroll depth data.

Previously, this data was queried dynamically from ClickHouse on page loads, but this approach adds 50-200ms to _every_ page load until a site has sent us scroll depth data.

Instead, we now update `site.scroll_depth_visible_at` in ingestion. To limit impact on ingestion:
1. `scroll_depth_visible_at` is now cached along with other ingestion site data.
2. updates are processed using ets, making sure only one database update is ever done per site
3. handles errors gracefully if postgres is ever in e.g. read-only mode or otherwise unavailable.

I'm curious whether I'm taking a sane approach with the GenServer here. cc @zoldar for feedback on that front.

### DIFF
--- a/lib/plausible/ingestion/scroll_depth_visible_at.ex
+++ b/lib/plausible/ingestion/scroll_depth_visible_at.ex
@@ -30,7 +30,7 @@ defmodule Plausible.Ingestion.ScrollDepthVisibleAt do
 
   defp attempt_update_repo(site_id) do
     Repo.update_all(
-      from(s in Plausible.Site, where: s.id == ^site_id),
+      from(s in Plausible.Site, where: s.id == ^site_id and is_nil(s.scroll_depth_visible_at)),
       set: [
         scroll_depth_visible_at: DateTime.utc_now()
       ]
--- a/lib/plausible/ingestion/scroll_depth_visible_at.ex
+++ b/lib/plausible/ingestion/scroll_depth_visible_at.ex
@@ -1,5 +1,11 @@
 defmodule Plausible.Ingestion.ScrollDepthVisibleAt do
-  @moduledoc false
+  @moduledoc """
+  GenServer that updates the `scroll_depth_visible_at` column for a site when needed.
+
+  This is called in a hot loop in ingestion, so it:
+  1. Only updates the database once per site (if SiteCache doesn't know about visibility yet)
+  2. Does not retry the update if it fails, to be retried on server restart
+  """
 
   use GenServer
   require Logger
@@ -7,14 +13,9 @@ defmodule Plausible.Ingestion.ScrollDepthVisibleAt do
 
   import Ecto.Query
 
-  @initial_state %{
-    updated_sites: MapSet.new(),
-    pending: MapSet.new()
-  }
-
-  def start_link(opts) do
+  def start_link(opts \\ []) do
     opts = Keyword.merge(opts, name: __MODULE__)
-    GenServer.start_link(__MODULE__, @initial_state, opts)
+    GenServer.start_link(__MODULE__, opts, opts)
   end
 
   def mark_scroll_depth_visible(site_id) do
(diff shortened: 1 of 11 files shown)


lib/plausible/ingestion/scroll_depth_visible_at.ex
```diff
@@ -0,0 +1,55 @@
+defmodule Plausible.Ingestion.ScrollDepthVisibleAt do
```
zoldar, line 1: Given the GenServer is used exclisively for state, perhaps we could use an ETS table instead? It should also be more efficient to query against.

Setting the table up is relatively simple and can be done on application startup (like in the case of `Plausible.Ingestion.Source`):

```elixir
  def init() do
    :ets.new(__MODULE__, [
      :named_table,
      :set,
      :public,
      {:read_concurrency, true},
      {:write_concurrency, true}
    ])
  end
```

The check function would check the site_id against the ETS table and spawn the update operation and insert it into the ETS table at the same time if the entry is missing. Something roughly like:

```
case :ets.lookup(__MODULE__, site_id) do
  [{_, _}] -> 
    :skip

  _ ->
    :ets.insert(__MODULE__, {site_id, site_id})
    Task.start(fn -> attempt_update_repo(site_id) end)
end
```

We could get more fancy and remove the entry if the query process fails for whatever reason (in the `rescue` block), but that's probably an overkill.

Another possibility is using `Plausible.Cache` along with `Plausible.Cache.Warmer` and implement a dedicated cache following that interface for the purpose. It has an added benefit of keeping the cache "warmed up", reducing potential excess of DB queries even more.

macobo: > Given the GenServer is used exclisively for state, perhaps we could use an ETS table instead? It should also be more efficient to query against.

Yeah, this should work and would behave pretty much identically. If you think it's worth it from an efficiency standpoint I'll go ahead and make this change.

> We could get more fancy and remove the entry if the query process fails for whatever reason (in the rescue block), but that's probably an overkill.

I intentionally didn't do this w/ the current implementation to avoid creating a thundering herd effect - if the db is down, this could lead to failures causing more errors. To do this properly we'd need to implement a exponential backoff system or make this a queue type thing.

> Another possibility is using Plausible.Cache along with Plausible.Cache.Warmer and implement a dedicated cache following that interface for the purpose. It has an added benefit of keeping the cache "warmed up", reducing potential excess of DB queries even more.

I considered this, but this would require an inordinate amount of work on clickhouse schema side or we'd end up with inefficient queries creating a lot of undue load. Given one-off usecase schemas are a pain to maintain long-term doing it in ingestion seemed like the more sane approach.

lib/plausible/ingestion/scroll_depth_visible_at.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  def mark_scroll_depth_visible(site_id) do
+    if :ets.insert_new(__MODULE__, {site_id}) do
+      Task.start(fn -> attempt_update_repo(site_id) end)
+    end
+  end
+
+  defp attempt_update_repo(site_id) do
+    Repo.update_all(
+      from(s in Plausible.Site, where: s.id == ^site_id),
```
zoldar, line 33: Given we now only care about timestamp being present and not having the value in particular range (or within a threshold), we could add ` where: is_nil(s.scroll_depth_visible_at)` to further avoid redundant table writes.

lib/plausible/ingestion/scroll_depth_visible_at.ex
zoldar: Yeah, that works, I think 😄 Good call on using `insert_new` 👍

zoldar (approved):
✨

test/plausible/ingestion/scroll_depth_visible_at_test.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Plausible.Ingestion.ScrollDepthVisibleAtTest do
+  use Plausible.DataCase
+  use Plausible.Teams.Test
+
+  test "mark_scroll_depth_visible" do
```
RobertJoonas, line 5: This test can never fail. We should assert on the return value of `TestUtils.eventually`

RobertJoonas (approved):

PR #5037, opened 2025-01-30
Replace hand-rolled dropdown on the dashboard and menu-related UI fixes

apata:
### Changes

* Removes handrolled dropdowns and replaces them with headlessui Popovers: https://headlessui.com/v1/react/popover

  * Also tried to use headlessui Menu component for datepicker, but unfortunately those were too opinionated to work well with our keybinds. 
  
  * Had to refactor the datepicker: visually that means extremely minor changes (first group doesn't have top padding)
  
* Fixes the issue with calendars positioning off screen, aligning their right edge to right of dropdown or right of visible page on smaller screens.

  * Removes flatpickr position styles from CSS and applies position styles with TW.
  
* Fixes issues with Dashboard modals not receiving focus

* Fixes issues with Escape clearing filters instead of closing opened menu / popover

  * This was a bit hacky due to headlessui API: we have to stop Escape propagating from Popover.Button (which receives focus on escaping from Popover). If the button isn't blurred immediately, as this PR does, user would have to tab off the button for the clear all filters keybind to work.  
  
* Fixes issues with multiple dropdowns being openable at the same time

* Optimises top bar components re-rendering too many times for no reason

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/nav-menu/query-periods/query-period-menu.tsx
```diff
@@ -0,0 +1,208 @@
...
+        {...popover.transition.props}
+        className={classNames(
+          'mt-2',
+          popover.transition.classNames.fullwidth,
+          dropdownState === DropdownState.CALENDAR
+            ? 'md-left-auto'
+            : 'md:left-auto md:w-56'
+        )}
+      >
+        <Popover.Panel
```
ukutaht, line 155: Nit: The fact that `dropdownState` changes immediately when the menu is closed sort of negates the close transition. The menu disappears immediately and only the shadow is transitioned away which looks weird if you really pay attention.

Ideally the dropdownState would change to closed after the transition is finished. Or maybe it would be easier to make the different states their own separate `Popover.Panel` components but I don't know how that would affect the Menu -> Calendar transition.

Anyways, not a biggie but something I noticed.

assets/js/dashboard/stats/graph/interval-picker.js
```diff
@@ -147,8 +151,12 @@ export function IntervalPicker({ onIntervalUpdate }) {
             enterTo="opacity-100 scale-100"
```
ukutaht, line 151: Can we use the same consolidated transition props here as the popovers to make sure they're consistent?

assets/js/dashboard/stats/modals/modal.js
```diff
@@ -95,3 +98,12 @@ export default function ModalWithRouting(props) {
   const onClose = props.onClose ?? (() => navigate({ path: rootRoute.path, search: (s) => s }))
   return <Modal {...props} onClose={onClose} />
 }
+
+const FocusOnMount = ({focusableRef}) => {
```
ukutaht, line 102: 👍 Nice

Not suggesting to do this now but using HeadlessUI Dialog would give this for free and also trap focus. Currently when you tab through everything in the modal the focus can go to the background since focus is not trapped.

ukutaht: As you can see I'm a fan of that library :) Love that it handles all the nitty-gritty of accessibility, focus, and keyboard interactions while leaving styling completely up to us.

ukutaht (approved):
Looks good. I had some comments about further improvements that could be made but this PR already does a lot. Not suggesting any of this needs to be part of this PR.

> If the button isn't blurred immediately, as this PR does, user would have to tab off the button for the clear all filters keybind to work.

Yeah. OTOH it breaks the expected behaviour that I can:
1. Open dropdown
2. Close via escape key
3. Open again via enter

I think the ideal behaviour would be that `stopPropagation` only runs when the popover is open. The event would be allowed to propagate if popover is closed without blurring the button. In that case both escape and enter would work as expected after closing the dropdown. I tried using the `open` render prop to achieve this but the sequence of events didn't allow me to achieve it (the `open` prop is already false when the key event handler runs).

Not a blocker for sure, it's a tiny detail.

assets/js/dashboard/nav-menu/query-periods/query-period-menu.tsx
apata: Well spotted! I thought it was barely noticeable and was just happy to have finally gotten the following transitions working at all with keyboard and clicks. 

- Closed -> Menu
- Menu -> Closed
- Menu -> Calendar
- Calendar -> Closed
- Closed -> Calendar

Will look into it once more: I actually didn't try swapping out the panel as you suggest OR driving Transition and Panel directly with the state.

ukutaht: Yeah, it works great as is, just a very minor nitpick. Don't spend too much time polishing it :)

PR #5039, opened 2025-02-03
Tracker: Track only engagements (pageleave extension)

macobo:
After this change we'll only track `engagement` events from tracker. 

This does not rename the extension - too much work for little gain, in the end this code will get rolled into main script. Possibility of breaking tracking for us or for beta users by accident.

**Reviewers:** Let's discuss the deployment plan in [basecamp]( https://3.basecamp.com/5308029/buckets/39034214/card_tables/cards/8283769943) since it involves multiple PRs.


tracker/src/plausible.js
```diff
@@ -100,52 +100,39 @@
...
-    // Note that `currentPageMaxEngagementScrollDepth` default of -1 ensures that at least one
+    // Note that `currentEngagementMaxScrollDepth` default of -1 ensures that at least one
     // engagement event is sent
-    if (!currentPageLeaveIgnored && currentPageMaxEngagementScrollDepth < maxScrollDepthPx) {
-      currentPageMaxEngagementScrollDepth = maxScrollDepthPx
-      triggerEngagementEvent('engagement')
-    }
+    if (!engagementCooldown && !currentEngagementIgnored && currentEngagementMaxScrollDepth < maxScrollDepthPx) {
+      currentEngagementMaxScrollDepth = maxScrollDepthPx
+      setTimeout(function () {engagementCooldown = false}, 300)
```
RobertJoonas, line 109: Q: Why change the engagement cooldown to 300ms?

RobertJoonas (approved):

tracker/src/plausible.js
macobo: Engagements had no cooldowns prior to this PR. I set this as 300ms to maximize engagements where user opens site and bounces really fast while balancing against the redirect usecase.

PR #5043, opened 2025-02-03
Stop segment queries when not needed

apata:
### Changes

When investigating some traces, we found that there's more database queries happening than needed. This removes one of them, which was running on all stats requests.

### Tests
- [x] Existing controller tests should cover this

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/segments.ex
+++ b/lib/plausible/segments/segments.ex
@@ -39,7 +39,13 @@ defmodule Plausible.Segments do
 
   @spec get_many(Plausible.Site.t(), list(pos_integer()), Keyword.t()) ::
           {:ok, [Segment.t()]}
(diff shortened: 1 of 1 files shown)


lib/plausible/segments/segments.ex
```diff
@@ -39,7 +39,13 @@ defmodule Plausible.Segments do
 
   @spec get_many(Plausible.Site.t(), list(pos_integer()), Keyword.t()) ::
           {:ok, [Segment.t()]}
-  def get_many(%Plausible.Site{} = site, segment_ids, opts) when is_list(segment_ids) do
+  def get_many(%Plausible.Site{} = _site, segment_ids, _opts)
```
ukutaht, line 42: I think pattern matching directly instead of using `when` would work here (?)

```elixir
def get_many(%Plausible.Site{} = _site, [], _opts) do
```

PR #5048, opened 2025-02-05
Separate liveness and readiness checks

ukutaht:
### Changes

Adds two new endpoints:
1. `/api/system/health/live` does nothing more than return 200 OK with a JSON body `{"ok": "true"}`
2. `/api/system/health/ready` checks all the necessary dependencies for accepting traffic (identical to old `/api/health` behaviour):
  2.1 Postgres reachable
  2.2 Clickhouse reachable
  2.3 Ingestion caches ready

The old `/api/health` endpoint is still usable to give time to migrate external checks.

Technically ingestion can keep working as long as caches are loaded but postgres is unreachable. Future possibility is separating ingestion readiness check from app readiness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Details modal search inputs are now case-insensitive.
 - Improved report performance in cases where site has a lot of unique pathnames
 - Plausible script now uses `fetch` with keepalive flag as default over `XMLHttpRequest`. This will ensure more reliable tracking. Reminder to use `compat` script variant if tracking Internet Explorer is required.
+- The old `/api/health` healtcheck is soft-deprecated in favour of separate `/api/system/live` and `/api/system/ready` checks
 
 ### Fixed
 
--- a/lib/plausible_web/controllers/api/system_controller.ex
+++ b/lib/plausible_web/controllers/api/system_controller.ex
@@ -1,5 +1,6 @@
 defmodule PlausibleWeb.Api.SystemController do
   use PlausibleWeb, :controller
+  require Logger
 
   def info(conn, _params) do
     build =
@@ -20,9 +21,20 @@ defmodule PlausibleWeb.Api.SystemController do
     json(conn, %{ok: true})
   end
 
+  @task_timeout 15_000
   def readiness(conn, _params) do
+    postgres_health_task =
+      Task.async(fn ->
+        Ecto.Adapters.SQL.query(Plausible.Repo, "SELECT 1", [])
+      end)
+
+    clickhouse_health_task =
+      Task.async(fn ->
+        Ecto.Adapters.SQL.query(Plausible.ClickhouseRepo, "SELECT 1", [])
+      end)
+
     postgres_health =
-      case Ecto.Adapters.SQL.query(Plausible.Repo, "SELECT 1", []) do
+      case Task.await(postgres_health_task, @task_timeout) do
         {:ok, _} ->
           "ok"
 
@@ -32,7 +44,7 @@ defmodule PlausibleWeb.Api.SystemController do
       end
 
     clickhouse_health =
-      case Ecto.Adapters.SQL.query(Plausible.ClickhouseRepo, "SELECT 1", []) do
+      case Task.await(clickhouse_health_task, @task_timeout) do
         {:ok, _} ->
           "ok"
 
--- a/lib/plausible_web/controllers/api/system_controller.ex
+++ b/lib/plausible_web/controllers/api/system_controller.ex
@@ -1,0 +1,55 @@
+defmodule PlausibleWeb.Api.SystemController do
+  use PlausibleWeb, :controller
+
+  def info(conn, _params) do
+    build =
+      :plausible
+      |> Application.get_env(:runtime_metadata)
+      |> Keyword.take([:version, :commit, :created, :tags])
+      |> Map.new()
+
+    geo_database = Plausible.Geo.database_type() || "(not configured)"
+
+    json(conn, %{
+      geo_database: geo_database,
+      build: build
+    })
+  end
+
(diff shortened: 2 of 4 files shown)


lib/plausible_web/controllers/api/system_controller.ex
```diff
@@ -0,0 +1,55 @@
+defmodule PlausibleWeb.Api.SystemController do
```
apata, line 1: The split of SystemController from ExternalController makes sense. 

The simpler liveness check is worth trying out at least. We can switch back from it fast, if it's not working out for us.

lib/plausible_web/controllers/api/system_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  def liveness(conn, _params) do
+    json(conn, %{ok: true})
+  end
+
+  def readiness(conn, _params) do
+    postgres_health =
+      case Ecto.Adapters.SQL.query(Plausible.Repo, "SELECT 1", []) do
+        {:ok, _} -> "ok"
+        e -> "error: #{inspect(e)}"
```
apata, line 27: I noticed we deliver the raw error over HTTP, and I think we shouldn't. It could be revealing more than we intend. Let's deliberately log it, but deliver something opaque like "error".

lib/plausible_web/router.ex
```diff
@@ -294,8 +294,14 @@ defmodule PlausibleWeb.Router do
 
       post "/event", Api.ExternalController, :event
       get "/error", Api.ExternalController, :error
-      get "/health", Api.ExternalController, :health
-      get "/system", Api.ExternalController, :info
+      # Remove this once all external checks are migration to new /system/health/* checks
+      get "/health", Api.SystemController, :readiness
+    end
+
+    scope "/system" do
```
apata, line 301: Just triple checking here that we don't want any plugs on this?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Details modal search inputs are now case-insensitive.
 - Improved report performance in cases where site has a lot of unique pathnames
 - Plausible script now uses `fetch` with keepalive flag as default over `XMLHttpRequest`. This will ensure more reliable tracking. Reminder to use `compat` script variant if tracking Internet Explorer is required.
+- The old `/api/health` healtcheck is soft-deprecated in favour of separate `/api/system/live` and `/api/system/ready` checks
```
apata, line 23: Are we expecting anyone except us to use this endpoint? It's not part of the versioned API (`v1` / `v2`), which makes me assume not. If not, do we need a change log for this one? Or is it to help CE?

aerosol:
What decision(s) are we willing to make based on HTTP availability alone? IMHO it's the cache(s) readiness check that should be separated.

BTW, omission on my part, only two caches are currently checked against. We should include user agents, sessions, country rules, page rules, hostname rules too. 

Checking cache size is not mandatory, but convenient way of makinge sure there's no exit thrown out of a key lookup, at least. Without those marked green we should not expose any node via LB.

ukutaht:
> What decision(s) are we willing to make based on HTTP availability alone?

The readiness probe is used to decide whether to route traffic to the container or not.
The liveness probe is used by kubernetes to decide whether to restart a container or not.

As I understand it the liveness probe should only fail in fatal scenarios that the application cannot recover by itself. Database downtime, for example, can be recovered without a restart which is the main idea with this separation.

> BTW, omission on my part, only two caches are currently checked against. We should include user agents, sessions, country rules, page rules, hostname rules too.
> Checking cache size is not mandatory, but convenient way of making sure there's no exit thrown out of a key lookup, at least. Without those marked green we should not expose any node via LB.

I'll add all the caches to the readiness probe.

ukutaht:
@aerosol from your basecamp comment

> Happy to work on any /api/health improvements allowing us to achieve better orchestration. Basically my take is we should start by refining our restart strategy to „let it crash more before we take it down”, to have more meaningful traces of any underlying conditions. 

This is exactly what the separate liveness probe does. The app will only be restarted when the container is unreachable or is reachable and cannot even respond with 200 OK.

aerosol:
@ukutaht yes, but given we let the traffic in once caches are initially ready, which /correct me if I'm wrong/ is currently not enforced? cc @cnkk

ukutaht:
Sorry I don't understand that sentence @aerosol

lib/plausible_web/controllers/api/system_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+      geo_database: geo_database,
+      build: build
+    })
+  end
+
+  def liveness(conn, _params) do
+    json(conn, %{ok: true})
+  end
+
+  def readiness(conn, _params) do
```
apata, line 23: Can we optimise this by running the checks concurrently?

aerosol:
> Sorry I don't understand that sentence

Sorry, hope all cleared up in basecamp. I think we might give it a go. Please consider merging https://github.com/plausible/analytics/pull/5057 - not critical, mostly cosmetic though.

aerosol (approved):

lib/plausible_web/controllers/api/system_controller.ex
ukutaht: Good catch https://github.com/plausible/analytics/pull/5048/commits/69bbae491870ec4cb0734c48db555564f6b24e8a

lib/plausible_web/router.ex
ukutaht: The only plug we're leaving out compared to before is `accepts :json`. We're not parsing JSON in SystemController so should be good

CHANGELOG.md
ukutaht: It's not really part of the public, documented, versioned APIs, no.

PR #5051, opened 2025-02-06
Scroll depth dashboard warnings (imported data)

RobertJoonas:
### Changes

Return a `scroll_depth` warning under `meta.metric_warnings` in the API response if imported data was included in the response, but no scroll depth data was found.

In the dashboard, we'll display warnings in two places (where the scroll depth metric is surfaced) - Top Stats and Top Pages > Details.

Screen recordings of how the dashboard tooltips work can be found at: https://3.basecamp.com/5308029/buckets/39034214/card_tables/cards/8130992150

This PR also changes `Plausible.DataMigration.SiteImports` to stop relying on the up-to-date SiteImport schema, defining an inline snapshot of the schema instead. Also does the same thing with the `imported_*` CH tables - defining a fixed list of imported tables that existed at that time.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -66,7 +66,7 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         )}
 
         {stat.name === 'Scroll depth' &&
-          stat.warning_code === 'no_imported_scroll_depth' && (
+          data.metric_warnings.scroll_depth === 'no_imported_scroll_depth' && (
             <p className="font-normal text-xs whitespace-nowrap">
               * Does not include imported data
             </p>
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -68,7 +68,7 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         {stat.name === 'Scroll depth' &&
           stat.warning_code === 'no_imported_scroll_depth' && (
             <p className="font-normal text-xs whitespace-nowrap">
-              * Includes only native data
+              * Does not include imported data
             </p>
           )}
       </div>
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -66,7 +66,7 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         )}
 
         {stat.name === 'Scroll depth' &&
-          data.metric_warnings.scroll_depth === 'no_imported_scroll_depth' && (
+          data.meta.metric_warnings?.scroll_depth?.code === 'no_imported_scroll_depth' && (
             <p className="font-normal text-xs whitespace-nowrap">
               * Does not include imported data
             </p>
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -66,7 +66,8 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         )}
 
         {stat.name === 'Scroll depth' &&
-          data.meta.metric_warnings?.scroll_depth?.code === 'no_imported_scroll_depth' && (
+          data.meta.metric_warnings?.scroll_depth?.code ===
+            'no_imported_scroll_depth' && (
             <p className="font-normal text-xs whitespace-nowrap">
               * Does not include imported data
             </p>
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -64,6 +64,12 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
             <SecondsSinceLastLoad lastLoadTimestamp={lastLoadTimestamp} />s ago
           </p>
         )}
+
+        {stat.name === 'Scroll depth' && stat.warning_code === 'no_imported_scroll_depth' && (
+          <p className="font-normal text-xs whitespace-nowrap">
+            * Includes only native data
+          </p>
+        )}
       </div>
     )
   }
@@ -114,6 +120,9 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         {statDisplayName}
         {statExtraName && (
           <span className="hidden sm:inline-block ml-1">{statExtraName}</span>
+        )}
+        {stat.warning_code && (
+          <span className="inline-block ml-1">*</span>
         )}
       </div>
     )
--- a/assets/js/dashboard/stats/graph/top-stats.js
+++ b/assets/js/dashboard/stats/graph/top-stats.js
@@ -65,11 +65,12 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
           </p>
         )}
 
-        {stat.name === 'Scroll depth' && stat.warning_code === 'no_imported_scroll_depth' && (
-          <p className="font-normal text-xs whitespace-nowrap">
-            * Includes only native data
-          </p>
-        )}
+        {stat.name === 'Scroll depth' &&
+          stat.warning_code === 'no_imported_scroll_depth' && (
+            <p className="font-normal text-xs whitespace-nowrap">
+              * Includes only native data
+            </p>
+          )}
       </div>
     )
   }
@@ -121,9 +122,7 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
         {statExtraName && (
           <span className="hidden sm:inline-block ml-1">{statExtraName}</span>
         )}
-        {stat.warning_code && (
-          <span className="inline-block ml-1">*</span>
-        )}
+        {stat.warning_code && <span className="inline-block ml-1">*</span>}
       </div>
     )
   }
--- a/assets/js/dashboard/stats/modals/breakdown-modal.tsx
+++ b/assets/js/dashboard/stats/modals/breakdown-modal.tsx
@@ -237,7 +237,7 @@ const ExternalLinkIcon = ({ url }: { url?: string }) =>
 
 const getMetricWarning = (metric: Metric, meta: BreakdownResultMeta | null) => {
   const warnings = meta?.metric_warnings
-  
+
   if (warnings) {
     const code = warnings[metric.key]?.code
 
--- a/assets/js/dashboard/stats/modals/breakdown-modal.tsx
+++ b/assets/js/dashboard/stats/modals/breakdown-modal.tsx
@@ -240,7 +240,7 @@ const getMetricWarning = (metric: Metric, meta: BreakdownResultMeta | null) => {
     const code = meta.metric_warnings[metric.key]
 
     if (code == 'no_imported_scroll_depth') {
-      return 'Includes only native data'
+      return 'Does not include imported data'
     }
   }
 }
--- a/assets/js/dashboard/stats/modals/breakdown-modal.tsx
+++ b/assets/js/dashboard/stats/modals/breakdown-modal.tsx
@@ -152,6 +152,7 @@ export default function BreakdownModal<TListItem extends { name: string }>({
           key: m.key,
           width: m.width,
           align: 'right',
+          metricWarning: getMetricWarning(m, meta),
           renderValue: (item) => m.renderValue(item, meta),
           onSort: m.sortable ? () => toggleSortByMetric(m) : undefined,
           sortDirection: orderByDictionary[m.key]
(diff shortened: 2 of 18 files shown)


assets/js/dashboard/stats/graph/top-stats.js
```diff
@@ -64,6 +64,13 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
             <SecondsSinceLastLoad lastLoadTimestamp={lastLoadTimestamp} />s ago
           </p>
         )}
+
+        {stat.name === 'Scroll depth' &&
+          stat.warning_code === 'no_imported_scroll_depth' && (
+            <p className="font-normal text-xs whitespace-nowrap">
+              * Includes only native data
```
macobo, line 71:
```suggestion
              * Does not include imported data
```

I don't think "native" terminology is not mentioned anywhere in the app and would be confusing.

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
```diff
@@ -1362,6 +1367,210 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryImportedTest do
...
+    test "returns warning when a site_import without scroll depth is in queried range", %{
+      conn: conn,
+      site: site
+    } do
+      site_import =
+        insert(:site_import, site: site, start_date: ~D[2021-02-01], end_date: ~D[2021-02-28])
+
+      populate_stats(site, site_import.id, [
+        build(:pageview, user_id: 123, pathname: "/", timestamp: ~N[2021-02-01 00:00:00]),
+        build(:pageleave,
```
macobo, line 1380: This needs a rebase

lib/plausible/stats/query_result.ex
```diff
@@ -162,7 +164,30 @@ defmodule Plausible.Stats.QueryResult do
     end
   end
 
-  defp metric_warning(_metric, _query), do: nil
+  @no_imported_scroll_depth_metric_warning %{
+    code: :no_imported_scroll_depth,
+    warning: "No imports with scroll depth data were found"
+  }
+
+  defp metric_warning(:scroll_depth, %{imports_included: true} = _meta, %QueryRunner{} = runner) do
```
macobo, line 172: We generally read whether imports are included via `query.include_imported`. Could we use a similar approach here instead of relying on meta?

This would keep imports-related code nicely greppable.

lib/plausible/stats/query_result.ex
```diff
@@ -162,7 +164,30 @@ defmodule Plausible.Stats.QueryResult do
...
+    %QueryRunner{site: site, main_query: main_query, comparison_query: comparison_query} = runner
+
+    imports_in_main_range = Imported.completed_imports_in_query_range(site, main_query)
```
macobo, line 175: Problem: QueryResult is supposed to be a dumb data class which does does no additional queries and work besides relying on the data passed to it. 

After this change it would do database queries. 

Rather than doing work in one place and another, could we preload imports in query_parser and rely on them throughout the query pipeline (including for building query.include_imported)?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -414,35 +428,29 @@ defmodule PlausibleWeb.Api.StatsController do
...
 
-  defp top_stats_entry(current_results, name, key, opts \\ []) do
-    if current_results[key] do
-      formatter = Keyword.get(opts, :formatter, & &1)
-      value = get_in(current_results, [key, :value])
+  defp top_stats_entry(current_results, meta, stat) do
+    if current_results[stat.key] do
+      formatter =
+        if stat.key == :time_on_page do
+          &if is_nil(&1), do: 0, else: &1
```
macobo, line 444: This seems to me too clever code that makes things hard to modify in the future and causes incompatibilities between stats API and dashboards.

What are we trying to achieve here?

assets/js/dashboard/stats/modals/breakdown-modal.tsx
```diff
@@ -233,3 +234,13 @@ const ExternalLinkIcon = ({ url }: { url?: string }) =>
       </svg>
     </a>
   ) : null
+
+const getMetricWarning = (metric: Metric, meta: BreakdownResultMeta | null) => {
+  if (meta && meta.metric_warnings) {
+    const code = meta.metric_warnings[metric.key]
+
+    if (code == 'no_imported_scroll_depth') {
+      return 'Includes only native data'
```
macobo, line 243: I don't think "native" terminology is not mentioned anywhere in the app and would be confusing.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -459,6 +467,15 @@ defmodule PlausibleWeb.Api.StatsController do
     end
   end
 
+  defp maybe_put_metric_warning(entry, meta, :scroll_depth) do
```
macobo, line 470: This change makes it quite a bit harder to change from our current response structure to the apiv2 one and adds more differences between our dashboard and APIv2 APIs.

Could we instead return a "meta" key at the top level?

lib/plausible/stats/query_result.ex
RobertJoonas: It won't actually do any db queries since `:site_imports` is already preloaded at this point - in PlausibleWeb.Plugs.AuthorizeSiteAccess we call `Repo.preload(site, :completed_imports)` which leaves completed imports available for the rest of the `conn` pipeline. If needed in QueryParser, the completed imports can be accessed under the `site` struct as easily.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: This behaviour was already there and I decided not to touch it at this point.

assets/js/dashboard/stats/graph/top-stats.js
RobertJoonas: Good point. Changed now!

test/plausible_web/controllers/api/external_stats_controller/query_imported_test.exs
RobertJoonas: Done.

lib/plausible/stats/query_result.ex
RobertJoonas: The question of whether imported data was included or not gets answered based on two queries - the main query, and the comparison query. We're already doing that when computing `meta.imports_included`. This is why I've decided to read that field directly here, instead of doing the exact same thing again.

assets/js/dashboard/stats/modals/breakdown-modal.tsx
RobertJoonas: Yup. Changed this.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: I've changed the response to have a `metric_warnings` field in the top level. The reason I didn't go with the `meta` field for the time being is that there are many other fields that could also be considered "meta". I'd leave that for a separate project. But I agree that introducing a new field in every top stat in the list will make it harder to migrate to API v2.

lib/plausible/stats/query_result.ex
macobo: Ah, this was changed in https://github.com/plausible/analytics/pull/5011. I think either the new logic is wrong or we're tackling this the wrong way and this business logic should be rolled into query.include_imported. We should make a call.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -888,9 +901,17 @@ defmodule PlausibleWeb.Api.StatsController do
         pages |> to_csv(cols)
       end
     else
+      response_meta = Stats.Breakdown.formatted_date_ranges(query)
+
+      response_meta =
```
macobo, line 906: Stats.breakdown already returns a `meta` object. Can't we return that directly with 0 transformations in the controller?

lib/plausible_web/controllers/api/stats_controller.ex
macobo: Q: Where? That this was already done this way is not obvious from the diff.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -888,9 +901,17 @@ defmodule PlausibleWeb.Api.StatsController do
...
+        case metric_warnings(meta) do
+          warnings when map_size(warnings) == 0 -> response_meta
+          warnings -> Map.put(response_meta, :metric_warnings, warnings)
+        end
+
       json(conn, %{
         results: pages,
-        meta: Stats.Breakdown.formatted_date_ranges(query),
+        meta: response_meta,
```
macobo, line 914:
```suggestion
        meta: Map.merge(meta, Stats.Breakdown.formatted_date_ranges(query))
```

lib/plausible/stats/query_result.ex
macobo: We discussed this in the call. One idea we arrived at:
1. Add something like `comparison_date_range` to Query object in QueryParser
2. Use that date range when deciding `query.include_imported`
3. Use that date range in QueryRunner as well
4. Simplify QueryResult

All this in a separate PR

assets/js/dashboard/stats/graph/top-stats.js
```diff
@@ -64,6 +64,13 @@ export default function TopStats({ data, onMetricUpdate, tooltipBoundary }) {
             <SecondsSinceLastLoad lastLoadTimestamp={lastLoadTimestamp} />s ago
           </p>
         )}
+
+        {stat.name === 'Scroll depth' &&
```
macobo, line 68: Minor: We might be able to use `stat.graph_metric` here and make this a bit more easily findable.

RobertJoonas: with the changes in https://github.com/plausible/analytics/pull/5051/commits/18dcaa4ef2ac5da885318cd1a9482638bbdf5fe7, searching by `scroll_depth`, this place will show up in the results :)

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Done. https://github.com/plausible/analytics/pull/5051/commits/18dcaa4ef2ac5da885318cd1a9482638bbdf5fe7

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: In the end went with passing meta directly. https://github.com/plausible/analytics/pull/5051/commits/18dcaa4ef2ac5da885318cd1a9482638bbdf5fe7

lib/plausible/data_migration/site_imports.ex
```diff
@@ -37,7 +71,7 @@ defmodule Plausible.DataMigration.SiteImports do
       |> Repo.all(log: false)
 
     site_imports =
-      from(i in Imported.SiteImport, where: i.status == ^SiteImport.completed())
+      from(i in SiteImportSnapshot, where: i.status == ^SiteImport.completed())
```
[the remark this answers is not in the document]
RobertJoonas: Done. I've removed any mention of the original SiteImport schema. https://github.com/plausible/analytics/pull/5051/commits/f25e61e06e6b6dc2e12f29b7048d2d5e7d7045f3

lib/plausible/data_migration/site_imports.ex
```diff
@@ -9,20 +9,51 @@ defmodule Plausible.DataMigration.SiteImports do
...
+    schema "site_imports" do
+      field :start_date, :date
+      field :end_date, :date
+      field :label, :string
+      field :source, Ecto.Enum, values: [:universal_analytics, :google_analytics_4, :csv, :noop]
+      field :status, Ecto.Enum, values: [:pending, :importing, :completed, :failed]
+      field :legacy, :boolean, default: false
+
+      belongs_to :site, Plausible.Site
+      belongs_to :imported_by, Plausible.Auth.User
```
zoldar, line 30: I think we could do away with this relationship as it's unused. We could even swap `belongs_to :user` for `field :user_id, :integer` but leaving it as is shouldn't hurt either as we don't actively query or cast against them anyway.

zoldar (approved):
LGTM ✨ left one entirely optional suggestion.

lib/plausible/data_migration/site_imports.ex
RobertJoonas: I tried getting rid of these actually, but a few operations in this script got overly complicated (e.g. using Ecto.Fragment to call `jsonb_set`, manual `iso_8601` <> `Date` conversions, etc). Using the Site and User schemas in these queries doesn't hurt that much, as you said, so let's leave it like that for the time being.

PR #5063, opened 2025-02-11
Scroll depth: Engagement events should refresh in-memory session cache

macobo:
This will drop less events when session was actually active but didn't have any pageviews/custom events arrive.

Ref: https://3.basecamp.com/5308029/buckets/39034214/card_tables/cards/8314968048


RobertJoonas (approved):
LGTM 👍

PR #5065, opened 2025-02-11
Fix mismatch between session and event user_ids after salts are rotated

macobo:
Consider the following scenario:
1. We receive a pageview from a user A
2. Salts are rotated after day end
3. We receive another pageview

Previously, this would result in:
- 2 sessions with different user_ids
- 2 events where each would have different user_ids

This fixes the problem by removing the previous faulty session updating
and forcing the event.user_id to be updated after we found a session
with the previous salt.


ukutaht (approved):

ukutaht:
Whoops, thanks for catching it :)

PR #5066, opened 2025-02-12
Improve Page Scroll Goals UX

RobertJoonas:
### Changes

Improve the UX of adding a scroll goal, moving it into a different tab, implementing automatic display name fill, and more. The UX can be tested in the review environment.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -196,16 +208,91 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
 
       <.input
-        :if={Plausible.Stats.ScrollDepth.feature_visible?(@site, @current_user)}
-        label="Scroll Depth Threshold (optional)"
+        id={"scroll_threshold_input_#{@suffix}"}
+        required
         field={@f[:scroll_threshold]}
         type="number"
-        value={if @goal && @goal.scroll_threshold > -1, do: @goal.scroll_threshold, else: nil}
         min="0"
```
macobo, line 260: Should this be 1?

At 0 my question is why not use a page goal?

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -196,16 +208,91 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+        }
+        """
+      end
+
+    assigns = assign(assigns, :js, js)
+
+    ~H"""
+    <div id="scroll-form" class="py-2" x-data={@js} {@rest}>
+      <.label for={"scroll_threshold_input_#{@suffix}"}>
+        Scroll Percentage Threshold (0-100)
```
macobo, line 252: It _feels_ like there could be a useful tooltip we could add here with context what this means.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -418,6 +510,25 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+        @selected? && "shadow-inner font-medium bg-indigo-600 text-white",
+        !@selected? && "dark:text-gray-100 text-gray-800"
+      ]}
+      id="scroll-tab"
+      x-on:click={!@selected? && "tabSelectionInProgress = true"}
+      phx-click="switch-tab"
+      phx-value-tab="scroll"
+      phx-target={@myself}
+    >
+      Scroll
```
macobo, line 527: I wonder if this should say "Scroll depth" instead. I can see myself getting confused as a user.

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: I think rather than a tooltip, an info paragraph (like in the custom events tab) would perhaps be better? Anyway, not sure if we need extra info there? (cc @metmarkosaric)

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: So the suggestion is to Rename "Scroll" to "Scroll depth" here:

image

cc @metmarkosaric thoughts?

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: image

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: Agreed. https://github.com/plausible/analytics/pull/5066/commits/838c5d6814945f23242c17a1e4a142c2761e685b

RobertJoonas: image

lib/plausible_web/live/goal_settings/form.ex
metmarkosaric: "Scroll depth" is fine. does it fit on mobile screens etc?

lib/plausible_web/live/goal_settings/form.ex
metmarkosaric: I think it's pretty clear due to the headings but it doesn't hurt to add an info paragraph. how about:

> Scroll Depth goals allow you to see how many people scroll beyond your desired scroll depth percentage threshold. Learn more in [our docs](https://plausible.io/docs/scroll-depth).

maybe then it also makes sense to add an info paragraph to Pageview goals as well? 

> Pageview goals allow you to measure how many people visit a specific page or section of your site. Learn more in [our docs](https://plausible.io/docs/pageview-goals).

RobertJoonas: Makes sense @metmarkosaric. I'll do that as part of the documentation task: https://3.basecamp.com/5308029/buckets/39034214/card_tables/cards/8195964494

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: Sounds good!

> does it fit on mobile screens etc?

Even on the smallest screens there's no overflow and it looks pretty okay (on most newer phones though, even "Custom Event" won't split onto two lines)

image

macobo (approved):

cnkk:
@cubic-dev-ai review this pull request

PR #5067, opened 2025-02-12
Speed up salts lookup by serving them off ets

aerosol:
A tiny patch to bypass synchronous gen server calls when looking up salts. Updates are still serialized.

### DIFF
--- a/lib/plausible/session/salts.ex
+++ b/lib/plausible/session/salts.ex
@@ -7,37 +7,54 @@ defmodule Plausible.Session.Salts do
       fn ->
         clean_old_salts()
 
+        __MODULE__ =
+          :ets.new(__MODULE__, [
+            :named_table,
+            :set,
+            :protected,
+            {:read_concurrency, true}
+          ])
+
         salts =
           Repo.all(from s in "salts", select: s.salt, order_by: [desc: s.inserted_at], limit: 2)
 
-        case salts do
-          [current, prev] ->
-            %{previous: prev, current: current}
+        state =
+          case salts do
+            [current, prev] ->
+              %{previous: prev, current: current}
 
-          [current] ->
-            %{previous: nil, current: current}
+            [current] ->
+              %{previous: nil, current: current}
 
-          [] ->
-            new = generate_and_persist_new_salt()
-            %{previous: nil, current: new}
-        end
+            [] ->
+              new = generate_and_persist_new_salt()
+              %{previous: nil, current: new}
+          end
+
+        true = :ets.insert(__MODULE__, {:state, state})
+        :ok
       end,
       name: __MODULE__
     )
   end
 
   def fetch() do
-    Agent.get(__MODULE__, & &1)
+    [state: state] = :ets.lookup(__MODULE__, :state)
+    state
   end
 
   def rotate() do
-    Agent.update(__MODULE__, fn %{current: current} ->
+    Agent.update(__MODULE__, fn :ok ->
+      current = fetch().current
       clean_old_salts()
 
-      %{
+      state = %{
         current: generate_and_persist_new_salt(),
         previous: current
       }
+
+      true = :ets.insert(__MODULE__, {:state, state})
+      :ok
     end)
   end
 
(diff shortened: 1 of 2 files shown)


lib/plausible/session/salts.ex
```diff
@@ -7,37 +7,54 @@ defmodule Plausible.Session.Salts do
...
-            %{previous: nil, current: new}
-        end
+        true = :ets.insert(__MODULE__, {:state, state})
+        :ok
       end,
       name: __MODULE__
     )
   end
 
   def fetch() do
```
macobo, line 41: This module currently lacks any unit tests. I think with these changes it makes sense to add some?

lib/plausible/session/salts.ex
```diff
@@ -7,37 +7,54 @@ defmodule Plausible.Session.Salts do
...
-    Agent.get(__MODULE__, & &1)
+    [state: state] = :ets.lookup(__MODULE__, :state)
+    state
   end
 
   def rotate() do
-    Agent.update(__MODULE__, fn %{current: current} ->
+    Agent.update(__MODULE__, fn :ok ->
```
macobo, line 47: Q: Do we need the agent at all anymore? `rotate()` is called within an oban worker already and there's no concurrency concerns here.

aerosol: It makes sense to do it here because Agent is the owner of that table so we let it manage its lifecycle. The only reason we have Oban in the mix is a decent guarantee of "exactly once at midnight" cron-like behaviour, that is otherwise hard to achieve with non-persistent processing. The table is protected, because we don't need public access (there is no need for concurrent writes).

lib/plausible/session/salts.ex
```diff
@@ -7,37 +7,54 @@ defmodule Plausible.Session.Salts do
...
+        true = :ets.insert(__MODULE__, {:state, state})
+        :ok
       end,
       name: __MODULE__
     )
   end
 
   def fetch() do
-    Agent.get(__MODULE__, & &1)
+    [state: state] = :ets.lookup(__MODULE__, :state)
```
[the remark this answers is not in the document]
aerosol: ETS and [persistent term](https://www.erlang.org/doc/apps/erts/persistent_term.html) are as fast as you can get in erlang. There is no process registry (or mailboxes) involved, serialization isn't enforced. Single process throughput is limited, you need to spawn a pool and handle routing/partitioning to shed the load. 

Potentially interesting:

- https://www.erlang.org/docs/23/man/ets
- https://www.erlang.org/docs/23/efficiency_guide/introduction
- https://learnyousomeerlang.com/ets
- https://dockyard.com/blog/2017/05/19/optimizing-elixir-and-phoenix-with-ets
- https://hexdocs.pm/elixir/main/ets.html
- https://learnyousomeerlang.com/clients-and-servers
- https://hexdocs.pm/elixir/erlang-term-storage.html

macobo (approved):
I'm okay with merging this as-is, though some tests wouldn't hurt. :star2:

aerosol:
@macobo I'll add tests :coffee:

aerosol:
@macobo See https://github.com/plausible/analytics/pull/5067/commits/9dadb83d7817fc4d9c4a0ed614e0147d58623265 - that good enough?

aerosol:
errr, there's some DB related issue on CI

aerosol:
> I don't really see the need to keep the Agent around, though. I think we can just get rid of it. 

Something must own the table though, and I don't like the application supervisor to do it.

lib/plausible/session/salts.ex
```diff
@@ -2,55 +2,75 @@ defmodule Plausible.Session.Salts do
...
+      {:ok, name}
     end)
   end
 
-  defp generate_and_persist_new_salt() do
+  defp generate_and_persist_new_salt(now) do
     salt = :crypto.strong_rand_bytes(16)
 
-    Repo.insert_all("salts", [%{salt: salt, inserted_at: DateTime.utc_now()}])
+    Repo.insert_all("salts", [%{salt: salt, inserted_at: now}])
```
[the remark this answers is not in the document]
aerosol: Or did you mean 2aa3664559 ?

aerosol:
@ukutaht trying to keep this PR scoped, GenServer is now in use

zoldar (approved):

PR #5071, opened 2025-02-13
Scroll depth: Rename imported_pages columns to reflect whats desired

macobo:
Pre-requisite to https://github.com/plausible/analytics/pull/5072, see that PR for deployment plan.

The new columns are:
- `total_scroll_depth UInt64`
- `total_scroll_depth_visits UInt64`

This should hopefully make it clear in queries that these columns are used as numerator and denominator.

This PR changes code and migrations at the same time. Generally this can cause issues, but the code modifications just remove the ability to _read_ the columns that were changed. As such, this is okay to merge (will do so via admin priviledges).

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/pages_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/pages_test.exs
@@ -676,6 +676,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
              }
     end
 
+    @tag skip: "To be re-enabled in the next PR"
     test "calculates scroll_depth from native and imported data combined", %{
       conn: conn,
       site: site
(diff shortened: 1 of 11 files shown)


test/plausible_web/controllers/api/stats_controller/pages_test.exs
```diff
@@ -1149,6 +1153,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
                "no_imported_scroll_depth"
     end
 
+    @tag skip: "To be re-enabled in the next PR"
```
RobertJoonas, line 1156: Why are we skipping this test? Doesn't seem related to imported scroll depth.

macobo: Hmm, not sure anymore. Maybe mistagged? In any case, this tag is removed in https://github.com/plausible/analytics/pull/5072 which will go live shortly after merging this.

RobertJoonas (approved):

PR #5073, opened 2025-02-13
Implement support for cache partitioning for sessions

zoldar:
### Changes

This PR introduces cache partitioning mechanism in `Plausible.Cache.Adapter` and uses it for `sessions`.

### Tests

### DIFF
--- a/test/plausible/cache_test.exs
+++ b/test/plausible/cache_test.exs
@@ -2,6 +2,7 @@ defmodule Plausible.CacheTest do
   use Plausible.DataCase, async: true
 
   alias Plausible.Cache
+  alias Plausible.Cache.Adapter
   import ExUnit.CaptureLog
 
   defmodule ExampleCache do
--- a/test/plausible/cache_test.exs
+++ b/test/plausible/cache_test.exs
@@ -2,7 +2,6 @@ defmodule Plausible.CacheTest do
   use Plausible.DataCase, async: true
 
   alias Plausible.Cache
-  alias Plausible.Cache.Adapter
   import ExUnit.CaptureLog
 
   defmodule ExampleCache do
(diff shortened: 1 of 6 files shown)


aerosol (commented):
This is good. One caveat: grafana metrics will be gone? Currently we enumerate cache metric names in `Plausible.PromEx.Plugins.PlausibleMetrics` (via `execute_cache_metrics/0` periodic call).

zoldar:
Maybe we could sum the stats across the clustered tables when sending them at https://github.com/plausible/analytics/blob/master/lib/plausible/telemetry/plausible_metrics.ex#L121-L124 ? That would still leave the open option to monitor each partition separately with gathering left intact.

aerosol:
@zoldar I dislike application env here, but couldn't figure out a better way. Added a lower level test to ensure compatibility (https://github.com/plausible/analytics/pull/5073/commits/fb3a549e7dca9dfe3176f5f28bb057b9feada58b)

aerosol:
And no, I don't think we need separate metrics for each partition. That's effectively monitoring consistent hashing distribution, kind of pointless.

aerosol (approved):

zoldar:
> @zoldar I dislike application env here, but couldn't figure out a better way

I'm not liking it too but I don't have any better idea either, perhaps other than baking the partitions number in at compile time.

test/plausible/cache_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule Plausible.CacheTest do
   use Plausible.DataCase, async: true
 
   alias Plausible.Cache
+  alias Plausible.Cache.Adapter
```
zoldar, line 5: 🔥?

aerosol: e0af2e00dc

aerosol (approved):

PR #5078, opened 2025-02-17
Use host network with `make clickhouse`

macobo:
Without this I couldn't get S3 exports to work locally as the ClickHouse container wasn't able to access minio.


zoldar (approved):

ruslandoga:
Should we also add a comment to https://docs.docker.com/engine/network/drivers/host/?uuid=ebc0dd6c-ae8d-43ac-aa17-81671a3a76dd%0A#docker-desktop? On my Mac it wasn't enabled by default. And we can probably remove port bindings.

PR #5082, opened 2025-02-17
Fix bug: bfcache pageviews not firing

RobertJoonas:
### Changes

This is a follow-up to https://github.com/plausible/analytics/pull/4656 to actually fix the issue.

The problem in the current implementation is that it conflicts with the `if (lastPage === location.pathname) return;` check. It intends to ignore pageviews on navigations in SPAs where the pathname didn't actually change, but it also ends up ignoring the bfcache pageviews. Fix: only do that check on SPA navigation.

### Tests

### Changelog

### Documentation

### Dark mode


macobo (approved):
I couldn't find a good way to simulate testing this (besides firing dummy events which is... suspect). 

The logic looks good to me. Let's revert the test stuff and ship? :shipit:

tracker/test/fixtures/simple.html
```diff
@@ -6,11 +6,14 @@
...
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Plausible Playwright tests</title>
-  <script defer src="/tracker/js/plausible.local.js"></script>
 </head>
 
 <body>
-  HELLO!
+  <script>
+    window.addEventListener('pageshow', (e) => {
+      console.log(e.persisted)
```
ukutaht, line 14: Is this leftover from some manual testing?

tracker/src/plausible.js
```diff
@@ -275,7 +275,7 @@
 
     function page(isSPANavigation) {
       {{#unless hash}}
-      if (lastPage === location.pathname) return;
+      if (isSPANavigation && lastPage === location.pathname) return;
```
ukutaht, line 278: tbh I don't even remember why this check is in there in the first place. Any idea?

tracker/test/pageview.spec.js
```diff
@@ -10,4 +10,12 @@ test.describe('Basic installation', () => {
     expect(plausibleRequest.url()).toContain('/api/event')
     expect(plausibleRequest.postDataJSON().n).toEqual('pageview')
   })
+
+  test('bfcache', async ({ page }) => {
```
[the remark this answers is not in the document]
RobertJoonas: Ditto - WIP.

tracker/src/plausible.js
RobertJoonas: Prevents pageviews from being sent on history navigation when the pathname didn't change. Not 100% sure this is a correct example, but our own dashboard state is persisted in `window.history` so without this check we would send a pageview every time a filter is applied or date range is changed, etc.

PR #5085, opened 2025-02-18
Extracted migration from PR #5084 (createCustomEventArray udf)

macobo:
See #5084  for full information.


RobertJoonas (commented):
Does this db function bring a performance benefit as well? Or is it just an attempt to improve the query readability?

macobo:
Yes, I expect it to affect performance at least in theory.

TL;DR: ClickHouse is quite dumb in terms of how it does things. `if(indexOf(...) > 0, [indexOf(...) + 7], [])` would evaluate the same `indexOf` twice every row.

RobertJoonas (approved):

PR #5089, opened 2025-02-18
Restrict public CSV export access

zoldar:
### Changes

Allow access to CSV export only to authenticated users having access to the website.

### Tests

### DIFF
--- a/assets/js/dashboard/stats/graph/stats-export.js
+++ b/assets/js/dashboard/stats/graph/stats-export.js
@@ -3,11 +3,17 @@ import * as api from '../../api';
 import { getCurrentInterval } from "./interval-picker";
 import { useSiteContext } from "../../site-context";
 import { useQueryContext } from "../../query-context";
+import { useUserContext } from '../../user-context';
 
 export default function StatsExport() {
   const site = useSiteContext();
   const { query } = useQueryContext();
   const [exporting, setExporting] = useState(false)
+  const user = useUserContext();
+
+  if (!user.loggedIn) {
+    return null;
+  }
 
   function startExport() {
     setExporting(true)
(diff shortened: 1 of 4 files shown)


assets/js/dashboard/stats/graph/stats-export.js
```diff
@@ -1,17 +1,25 @@
...
-  const site = useSiteContext();
-  const { query } = useQueryContext();
+  const site = useSiteContext()
+  const { query } = useQueryContext()
   const [exporting, setExporting] = useState(false)
 
+  const user = useUserContext()
+  const adminAccess = ['owner', 'admin', 'editor', 'super_admin'].includes(
+    user.role
+  )
```
apata, line 18: nitpick: BE I think allows viewer as well. You can do 
```
  const user = useUserContext()
  if (!user.loggedIn) {return null}
```

aerosol (approved):

assets/js/dashboard/stats/graph/stats-export.js
zoldar: Looking at https://github.com/plausible/analytics/blob/restrict-csv-export/assets/js/dashboard.tsx#L59-L64 and https://github.com/plausible/analytics/blob/restrict-csv-export/assets/js/dashboard/user-context.tsx#L4-L14, I'm not exactly sure how to match against "public" role properly. Help @apata ? 😅

aerosol: let's merge asap, this isn't critical

zoldar: We are going to merge it as is, to see if it has any measurable impact while the pref issues still persist. We can work out a proper way to hiding it in the UI in a follow-up.

PR #5096, opened 2025-02-19
Ignore excess logger errors from Cowboy when sending to Sentry

zoldar:
### Changes

After a recent dependency bump, we have observed excess of new errors in Sentry combing from cowboy in a following format:

```
Ranch listener PlausibleWeb.Endpoint.HTTP, connection process #PID<...>, 
stream .. had its request process #PID<...> exit with reason ...
```

After some digging, it turned out that between cowboy 2.12 and 2.13 there's https://github.com/ninenines/cowboy/commit/fbd680f0f6fb5fe99c4345b838cb42aa1afb3283 which plugs an apparent missed pattern case for and starts triggering error messages. This in turn seems to escape the special-cased pattern matching of ranch/cowboy errors in https://github.com/getsentry/sentry-elixir/commit/a14e3f85748e694492bb51d59fb1df9e95bbac16.

This patch is meant as a temporary measure meant to reduce the noise in sentry.

### Tests

### DIFF
--- a/lib/sentry_filter.ex
+++ b/lib/sentry_filter.ex
@@ -23,9 +23,7 @@ defmodule Plausible.SentryFilter do
     %{event | fingerprint: ["ingestion_request"]}
   end
 
-  def before_send(
-        %{extra: %{logger_level: :error}, message: %{formatted: "Ranch listener" <> rest}} = event
-      ) do
+  def before_send(%{source: :logger, message: %{formatted: "Ranch listener" <> rest}} = event) do
     if String.contains?(rest, "had its request process") do
       false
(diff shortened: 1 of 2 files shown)


lib/sentry_filter.ex
```diff
@@ -23,6 +23,16 @@ defmodule Plausible.SentryFilter do
     %{event | fingerprint: ["ingestion_request"]}
   end
 
+  def before_send(
```
aerosol, line 26: we could add this to SentryFilterTest since it's there

aerosol (approved):

PR #5097, opened 2025-02-19
Hide dashboard import icons when redundant

RobertJoonas:
### Changes

A customer (and Marko) got confused by why we're now showing the imported toggle switch and import related warnings when imported data is not in view. It was actually "in the view", but only used for rendering the small comparison percentages/arrows for Top Stats metrics.

image

The request from @metmarkosaric was to remove both the imported toggle and imported warnings in this scenario. Note that when in comparison mode (and imported data only in comparison range), OR when imports are actually included in the main date range, we still display the icons.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -192,10 +192,29 @@ defmodule PlausibleWeb.Api.StatsController do
 
     params = realtime_period_to_30m(params)
 
-    query =
-      site
-      |> Query.from(params, debug_metadata(conn))
-      |> Query.set_include(:imports_meta, true)
+    {query, original_query_include_imported} =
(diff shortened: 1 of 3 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -192,10 +192,29 @@ defmodule PlausibleWeb.Api.StatsController do
 
     params = realtime_period_to_30m(params)
 
-    query =
-      site
-      |> Query.from(params, debug_metadata(conn))
-      |> Query.set_include(:imports_meta, true)
+    {query, original_query_include_imported} =
```
macobo, line 195: How would you feel about adding a new meta flag instead of this? We'd need to do so eventually anyways and the less logic we add here the better.

Happy to take this over if you'd like.

PR #5098, opened 2025-02-20
Ingestion: Add test demonstrating blank event name is not accepted

macobo:
Kind of a nothing-PR but while poking around in ingestion I thought I found an issue where we could ingest blank event names. This would cause problems in imports queries down the line.

Turns out we do validate event name length. I've added a (previously missing) test to this effect.


aerosol:
redundant and a less efficient version of the test already present in Ingestion.RequestTest :shrug:

aerosol (approved):

PR #5102, opened 2025-02-20
Hide dashboard import icons when redundant #2

macobo:
Alternative solution to https://github.com/plausible/analytics/pull/5097 to not show include/exclude imported data for comparisons in top stats.

What's changing:
- `query.imports_included` only applies to that query and not comparison query
- Query structure: 
    - query.comparison_query is populated in QueryOptimizer
    - `include_meta` -> `dashboard_include_meta` flag rename
    - `imports_exist` -> `completed_imports` list persisted on query

Previous solution was causing unneeded joins with imported tables.

Sadly now QueryResult needs to be somewhat aware of comparisons. I've tried to minimize the footprint though.

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -2,7 +2,6 @@ defmodule Plausible.Stats.Query do
   use Plausible
 
   defstruct utc_time_range: nil,
-            comparison_utc_time_range: nil,
             interval: nil,
             period: nil,
             dimensions: [],
@@ -26,7 +25,9 @@ defmodule Plausible.Stats.Query do
             revenue_warning: nil,
             remove_unavailable_revenue_metrics: false,
             site_id: nil,
-            site_native_stats_start_at: nil
+            site_native_stats_start_at: nil,
+            comparison_utc_time_range: nil,
+            comparison_query: nil
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{DateTimeRange, Filters, Imported, Legacy, Comparisons}
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -7,7 +7,7 @@ defmodule Plausible.Stats.Query do
             dimensions: [],
             filters: [],
             sample_threshold: 20_000_000,
-            imports_exist: false,
+            completed_imports: [],
             imports_in_range: [],
             include_imported: false,
             skip_imported_reason: nil,
@@ -135,17 +135,28 @@ defmodule Plausible.Stats.Query do
   end
 
   def put_imported_opts(query, site) do
-    requested? = query.include.imports
+    query
+    |> struct!(
+      completed_imports:
+        if(site, do: Plausible.Imported.completed_imports(site), else: query.completed_imports)
+    )
+    |> set_imports_in_range()
+    |> set_imports_included()
+  end
 
-    query =
-      if site do
-        struct!(query,
-          imports_exist: Plausible.Imported.any_completed_imports?(site),
-          imports_in_range: get_imports_in_range(site, query)
-        )
-      else
-        query
-      end
+  defp set_imports_in_range(%__MODULE__{period: period} = query)
+       when period in ["realtime", "30m"] do
+    query
+  end
+
+  defp set_imports_in_range(query) do
+    struct!(query,
+      imports_in_range: Plausible.Imported.completed_imports_in_query_range(query)
+    )
+  end
+
+  defp set_imports_included(query) do
+    requested? = query.include.imports
 
     skip_imported_reason = get_skip_imported_reason(query)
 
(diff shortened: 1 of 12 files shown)


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -220,18 +217,19 @@ defmodule PlausibleWeb.Api.StatsController do
...
   defp with_imported_switch_info(%Jason.OrderedObject{} = meta) do
-    case {meta[:imports_included], meta[:imports_skip_reason]} do
-      {true, nil} ->
-        %{visible: true, togglable: true, tooltip_msg: "Click to exclude imported data"}
-
-      {false, nil} ->
-        %{visible: true, togglable: true, tooltip_msg: "Click to include imported data"}
-
-      {false, :unsupported_query} ->
+    case Map.new(meta) do
```
RobertJoonas, line 220: There's one case missing here:

```elixir
%{imports_included_for_main_query: false, imports_included: true} ->
  %{visible: true, togglable: true, tooltip_msg: "Click to exclude imported data"}
```

When imports were included only in comparison range, it still means that imports were included, and the tooltip should say "click to exclude".

test/plausible_web/controllers/api/stats_controller/top_stats_test.exs
```diff
@@ -791,7 +791,7 @@ defmodule PlausibleWeb.Api.StatsController.TopStatsTest do
       assert res["with_imported_switch"] == %{
                "visible" => true,
                "togglable" => true,
-               "tooltip_msg" => "Click to exclude imported data"
+               "tooltip_msg" => "Click to include imported data"
```
RobertJoonas, line 794: Related to my other comment - this should not change.

test/plausible_web/controllers/api/stats_controller/top_stats_test.exs
```diff
@@ -803,9 +803,9 @@ defmodule PlausibleWeb.Api.StatsController.TopStatsTest do
       res = json_response(conn, 200)
 
       assert res["with_imported_switch"] == %{
-               "visible" => true,
-               "togglable" => true,
-               "tooltip_msg" => "Click to include imported data"
+               "visible" => false,
+               "togglable" => false,
+               "tooltip_msg" => nil
```
RobertJoonas, line 808: This shouldn't change either.

I believe instead we should add another test in this describe block that:

* makes a top-stats request without asking for comparisons (but top stats still gives the default comparisons)
* populates data that falls only in that default comparison range
* asserts the same that this test does - `assert res["with_imported_switch"] == %{"visible" => false, "togglable" => false, "tooltip_msg" => nil}`

lib/plausible/stats/query.ex
```diff
@@ -26,7 +25,9 @@ defmodule Plausible.Stats.Query do
             revenue_warning: nil,
             remove_unavailable_revenue_metrics: false,
             site_id: nil,
-            site_native_stats_start_at: nil
+            site_native_stats_start_at: nil,
+            comparison_utc_time_range: nil,
```
RobertJoonas, line 29: How about getting rid of `comparison_utc_time_range`? That information might as well be included under the `comparison_query` itself.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -192,10 +192,7 @@ defmodule PlausibleWeb.Api.StatsController do
 
     params = realtime_period_to_30m(params)
 
-    query =
-      site
-      |> Query.from(params, debug_metadata(conn))
-      |> Query.set_include(:imports_meta, true)
+    query = Query.from(site, params, debug_metadata(conn))
```
RobertJoonas, line 195: One of the changes in https://github.com/plausible/analytics/pull/5097 was to provide the default "previous_period" comparison to the query here (instead of `visitor-graph.js`)

I think that's still required - Top Stats are **always** queried with a comparison option, however,

* By default, comparison mode is `previous_period`, which is set in `visitor-graph.js`. In that "default" case. We **don't want to** display the imported switch / imported warnings.
* When the user is in the comparison mode however (press X on the dashboard, other reports are showing comparison data too), this is when we **do want to** display the imported switch / imported warnings.

macobo: The business logic changes you are describing in this and other comments are pretty complicated and non-intuitive and seem contradictory to what we discussed in the call. Is there a ticket somewhere?

I don't see why the frontend change is at all necessary here to fix the 'bug' that including comparison query in meta.include_imported added but I'm likely missing something

RobertJoonas: > Is there a ticket somewhere with the full context?

There's no ticket at the moment, as it seemed to me like a niche thing to get out of the way with a single PR. Should we create one?

> I don't see why the frontend change is at all necessary here to fix the 'bug' that including comparison query in meta.include_imported added but I'm likely missing something.

The "bug" here is not very straight forward indeed as the desired behaviours are somewhat conflicting.

The Top Stats query currently serves another purpose in addition to simply aggregating the stats. It is to set the state of whether imported data is currently visible on the dashboard or not, AND whether it exists at all in the given query (with comparison) range.

From the API perspective, I wouldn't even say there's a bug - imported data is in the Top Stats comparison range so the request responds with:

* yes - imported data WAS included
* When the imports button (aka "with_imported_switch") is clicked, it will **disable** imported data globally in the dashboard, and hence the tooltip should say  "click to **exclude** imported data"

The issue/weirdness comes from the fact that we're not in the "comparison mode" and yet the backend ends up running a comparison query because Top Stats fetches comparisons even when not in comparison mode - that is to render the small change percentages with arrows. The desired behaviour (as pointed out by Marko) is that when we're not explicitly in comparison mode, the top stats comparison range should "silently" count imported data.

The issue at hand could theoretically be solved by:

1. https://github.com/plausible/analytics/pull/5097 (fixes exactly the problem I'm describing - nothing more, nothing less)
2. not rendering the comparison change arrows UNLESS in comparison mode - I recall we did this at some point and people saw the top stats too "blank" so we put the change arrows back even in the non-comparison mode
3. create two separate queries in the Top Stats endpoint when not in comparison mode

None of these solutions is perfect in my head.

Also, I believe this issue was there already **before** any imports-related work/refactoring that was part of the scroll depth project.

PR #5106, opened 2025-02-23
Scroll Depth Release: ingestion flag check for incremental load increase

RobertJoonas:
### Changes

Implements a straight-forward feature flag check to drop engagements and not process them. The `engagements` flag should already be enabled for all sites that are currently getting engagements (including e.g. plausible.io).

We can use the feature flag to incrementally crank up the percentage of sites for which we **will** process engagements.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -57,15 +57,15 @@ defmodule Plausible.Ingestion.Event do
         for domain <- domains, do: drop(new(domain, request), :spam_referrer)
       else
         Enum.reduce(domains, [], fn domain, acc ->
-          case GateKeeper.check(domain) do
-            {:allow, site} ->
-              processed =
-                domain
-                |> new(site, request)
-                |> process_unless_dropped(pipeline(), context)
+          with {:allow, site} <- GateKeeper.check(domain),
+               :ok <- check_engagements_flag(domain, request) do
+            processed =
+              domain
+              |> new(site, request)
+              |> process_unless_dropped(pipeline(), context)
 
-              [processed | acc]
-
+            [processed | acc]
+          else
             {:deny, reason} ->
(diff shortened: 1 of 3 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -76,6 +76,16 @@ defmodule Plausible.Ingestion.Event do
     {:ok, %{dropped: dropped, buffered: buffered}}
   end
 
+  defp check_engagements_flag(domain, %{event_name: "engagement"}) do
+    if FunWithFlags.enabled?(:engagements, for: %Plausible.Site{domain: domain}) do
+      :ok
+    else
+      {:deny, :engagement_ff_throttle}
```
aerosol, line 83: Please extend `drop_reason()` spec with this value

RobertJoonas: Cheers for noticing! https://github.com/plausible/analytics/pull/5106/commits/a7b731102329aca20f0979f45548be325c7814e7

lib/plausible/ingestion/event.ex
```diff
@@ -76,6 +77,16 @@ defmodule Plausible.Ingestion.Event do
     {:ok, %{dropped: dropped, buffered: buffered}}
   end
 
+  defp check_engagements_flag(domain, %{event_name: "engagement"}) do
+    if FunWithFlags.enabled?(:engagements, for: %Plausible.Site{domain: domain}) do
```
macobo, line 81:
```suggestion
    if FunWithFlags.enabled?(:engagement_ingestion, for: %Plausible.Site{domain: domain}) do
```

This naming should make it clear what the flag is for.

macobo (approved):
Naming suggestion, otherwise lgtm

PR #5111, opened 2025-02-25
Permit :has_not_done filter in segments

apata:
### Changes

- Permits has_not_done filter in segments
- Adds extra tests about having non-existing goals within segments. 
- Slightly changes error message upon having non-existing goals within segments, to bring it in line with other errors about selected filters. 
- Removes JS-style triple equals comparison from backend: it isn't standard

### Tests

### Changelog
- [x] This PR makes an incredibly small user-facing change, which probably doesn't warrant a mention in the changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/segment.ex
+++ b/lib/plausible/segments/segment.ex
@@ -197,15 +197,10 @@ defmodule Plausible.Segments.Segment do
   end
 
   defp dashboard_compatible_filter?(filter) do
-    regular_filter? =
-      is_list(filter) and length(filter) == 3 and
-        is_atom(Enum.at(filter, 0)) and
-        is_binary(Enum.at(filter, 1)) and
-        is_list(Enum.at(filter, 2))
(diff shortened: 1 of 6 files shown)


lib/plausible/segments/segment.ex
```diff
@@ -197,9 +197,15 @@ defmodule Plausible.Segments.Segment do
   end
 
   defp dashboard_compatible_filter?(filter) do
```
macobo, line 199: This could be one big list comprehension:

```
case filter do
  [operation, dimension, _clauses] when is_atom(operation) and is_binary(dimension) -> true
  [:has_not_done, clauses] -> true
  _ -> false
```

This is incidentally more greppable as it uses the same style of pattern matching as the rest of the code and should thus lead to fewer incompatibilities as the format changes.

macobo (approved):

PR #5115, opened 2025-02-25
Flatten :and groups stemming from segment filters on first level

apata:
### Changes

Currently, segment BC (_containing filters `[B, C]`_) is resolved in the following way:
`[A, segment BC] -> [A, [:and, [B, C]]]`. 

That's a problem when B or C is an `event:goal` filter, because that's only allowed at the top level. 

This PR leverages the logical equivalency between `[A, [:and, [B, C]]]` and `[A, B, C]` to circumvent the problem.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/filters.ex
+++ b/lib/plausible/segments/filters.ex
@@ -63,6 +63,13 @@ defmodule Plausible.Segments.Filters do
     end
   end
 
+  defp expand_first_level_and_filters(filter) do
+    case filter do
+      [:and, clauses] -> clauses
+      filter -> [filter]
+    end
+  end
+
   defp replace_segment_with_filter_tree([_, "segment", clauses], preloaded_segments) do
     if length(clauses) == 1 do
       [[:and, Map.get(preloaded_segments, Enum.at(clauses, 0))]]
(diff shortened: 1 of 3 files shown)


lib/plausible/segments/filters.ex
```diff
@@ -98,9 +114,11 @@ defmodule Plausible.Segments.Filters do
   def resolve_segments(original_filters, preloaded_segments) do
     if map_size(preloaded_segments) > 0 do
       {:ok,
-       Filters.transform_filters(original_filters, fn f ->
+       original_filters
+       |> Filters.transform_filters(fn f ->
          replace_segment_with_filter_tree(f, preloaded_segments)
-       end)}
+       end)
+       |> Filters.transform_filters(&expand_first_level_and_filters/1)}
```
macobo, line 121: Nit: Shouldn't this be part of replace_segment_with_filter_tree?

Suggestion: Enum.map would be equivelent here to transform_filters but also would be easier to understand.

macobo (approved):
Minor suggestion, otherwise lgtm.

Edit: Seems like a parser test also needs updating.

PR #5119, opened 2025-02-26
Resolve segment owner names and dates in site timezone in the BE

apata:
### Changes

* Removes workaround with segment owner names and dates. Keeps the date util just in case.
* Indicates clearly the different segments index payloads (one for public dashboards and one for site members).
* Adds tests about claiming ownership of dangling segment. 
* Centralizes segment type label

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/filtering/segments.ts
+++ b/assets/js/dashboard/filtering/segments.ts
@@ -15,9 +15,17 @@ export type SavedSegment = {
   name: string
   type: SegmentType
   owner_id: number
+  owner_name: string
+  /** datetime in site timezone, example 2025-02-26 10:00:00 */
   inserted_at: string
+  /** datetime in site timezone, example 2025-02-26 10:00:00 */
   updated_at: string
 }
+
+export type SavedPublicSiteSegment = Pick<
+  SavedSegment,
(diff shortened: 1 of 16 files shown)


assets/js/dashboard/filtering/segments.ts
```diff
@@ -14,11 +14,24 @@ export type SavedSegment = {
   id: number
   name: string
   type: SegmentType
-  owner_id: number
+  /** null owner_id or owner_name signifies that the owner has been removed from the site */
+  owner_id: number | null
```
ukutaht, line 18: I don't have much experience with typescript types but would something like this be possible?

```
owner: {id: number, name: string} | null
```
?

This would disallow the following invalid state at the type level:

```
{
  owner_id: 123,
  owner_name: null
}
```

ukutaht (approved):

assets/js/dashboard/filtering/segments.ts
apata: True, that state is invalid and I've updated the type definitions to reflect that: https://github.com/plausible/analytics/pull/5119/commits/26b136a1bab90a24af263e865c9fee4a0ed34b63

PR #5123, opened 2025-02-26
use one lock partition per session cache

ruslandoga:
https://github.com/plausible/analytics/pull/5121 created too many new processes [(100 * schedulers)](https://github.com/sasa1977/con_cache/blob/60acf9cd75f8a23f9b9dff45e6881f125e00f703/lib/con_cache.ex#L173). This would bring it down to ~100.

If this ends up being part of the "final" approach, it would probably need a comment with some explanation of what number controls what :)


aerosol (approved):

aerosol:
@ruslandoga anything else you want to include here?

PR #5126, opened 2025-02-27
peep

ruslandoga:
There is some lock contention in `Plausible.PromEx_dist` table under load.
This PR is an attempt to fix it by using Peep and its "ETS per scheduler" tables for storing metrics.

---

Also to check:
- `Plausible.Cache.Stats` https://github.com/plausible/analytics/pull/5127
- `Plausible.RateLimit` (TODO)


aerosol (approved):
Looks good, ship it :ship:

PR #5127, opened 2025-02-27
counters for cache stats

ruslandoga:
`Plausible.Cache.Stats` table sometimes causes lock contention under load.
This is an experiment to see if removing that table makes sense.


lib/plausible/cache/stats.ex
```diff
@@ -35,6 +20,16 @@ defmodule Plausible.Cache.Stats do
...
 
+  def create_counters(cache_name) do
+    :persistent_term.put({__MODULE__, cache_name, @hit}, :counters.new(1, []))
+    :persistent_term.put({__MODULE__, cache_name, @miss}, :counters.new(1, []))
+  end
+
+  defp counter(cache_name, type) do
+    :persistent_term.get({__MODULE__, cache_name, type}, nil) ||
+      raise "counter not found for #{cache_name} #{type}"
+  end
```
ruslandoga, line 31: This is all very quick and dirty. The API can be improved if this works out.

aerosol (approved):
Love it :)

PR #5129, opened 2025-02-28
Implement Site Segments billing feature

apata:
### Changes

Adds Site Segments billing feature and related code.

- /docs PR: https://github.com/plausible/docs/pull/586
- plausible.io/#pricing PR: https://github.com/plausible/website/pull/753

### Tests

### Changelog

### Documentation

### Dark mode


test/plausible_web/plugins/api/controllers/capabilities_test.exs
```diff
@@ -29,7 +29,8 @@ defmodule PlausibleWeb.Plugins.API.Controllers.CapabilitiesTest do
                    "Goals" => false,
                    "Props" => false,
                    "RevenueGoals" => false,
-                   "StatsAPI" => false
+                   "StatsAPI" => false,
+                   "SiteSegments" => false
```
apata, line 33: @ukutaht @aerosol Regarding these changes, even if the current clients can't do anything with the info (there's no Segments API), I think I should leave it in: the reasoning is not to introduce a new "plugin capabilities/features" concept here:
https://github.com/plausible/analytics/blob/master/lib/plausible/plugins/api/capabilities.ex#L14

WDYT?

ukutaht:
@apata I'd propose to have in the growth box: "Segments" and in the business box: "Shared segments".

I agree about the naming - IMO the main difference is whether they can be shared among teammates or not. "Site segments' is not very descriptive.

metmarkosaric:
what Uku said makes sense to me too!

test/plausible_web/plugins/api/controllers/capabilities_test.exs
aerosol: Totally fine to add new keys.

apata: Thanks!

apata:
When downgrading:
image

PR #5130, opened 2025-02-28
peep

ruslandoga:
Same as #5126 but also adapts Peep bucketing to PromEx expectations.


lib/plausible/prom_ex/buckets.ex
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  @impl true
+  def number_of_buckets(config) do
+    config.number_of_buckets
+  end
+
+  @impl true
+  def bucket_for(number, config) when is_integer(number) do
+    elem(:gb_trees.larger(number, config.int_tree), 1)
```
ruslandoga, line 61: Oops, forgot to handle "larger than max" values.

ruslandoga: https://github.com/plausible/analytics/pull/5130/commits/30f6742c8a7e125cf7264cfecc217e573f6c32df

cnkk (approved):

PR #5139, opened 2025-03-04
Fix Oban option, split operations in CSV import export test

apata:
### Changes

This probably doesn't fix the flakiness with this test suite (which I managed to reproduce locally, see Basecamp card), but it fixes Oban options in the test (https://hexdocs.pm/oban/Oban.html#drain_queue/2-options) and clarifies the private unzip testing function to better show what part of the operation failed.

### DIFF
--- a/test/plausible/imported/csv_importer_test.exs
+++ b/test/plausible/imported/csv_importer_test.exs
@@ -351,7 +351,7 @@ defmodule Plausible.Imported.CSVImporterTest do
           storage: on_ee(do: "s3", else: "local")
         )
 
-      assert %{success: 1} = Oban.drain_queue(queue: :analytics_imports, with_safety?: false)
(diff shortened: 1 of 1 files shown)


test/plausible/imported/csv_importer_test.exs
```diff
@@ -1269,8 +1269,10 @@ defmodule Plausible.Imported.CSVImporterTest do
   end
 
   defp unzip_archive(%{tmp_dir: tmp_dir} = context) do
+    {:ok, binary} = :file.read_file(to_charlist(Path.join(tmp_dir, "plausible-export.zip")))
+
     assert {:ok, files} =
-             :zip.unzip(to_charlist(Path.join(tmp_dir, "plausible-export.zip")),
```
ruslandoga, line 1273: Alternative: https://github.com/plausible/analytics/pull/5164

PR #5143, opened 2025-03-05
Bump elixir.yml cache version to fix a test

macobo:
A test in test/plausible_web/plugs/tracker_test.exs was failing due to cached version of tracker script being used in other PRs.

The reason it didn't fail in the original PR adding the test was that changes in tracker script bypassed the cache **for that pr**. The old cached data was still used for other PRs, causing them to fail.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  CACHE_VERSION: v10
+  CACHE_VERSION: v11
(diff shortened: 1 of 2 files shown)


aerosol:
What's the reason for local failures though?

aerosol:
> the first time was because script compilation was still running. 

that must've been it here as well. It does not give any indication of running so I must've assumed it was idle...

apata (approved):

ruslandoga:
Should cache key include the hash of the tracker script?

PR #5144, opened 2025-03-05
Refine team switching and switch from `my_team` to `current_team` 

zoldar:
### Changes

This PR renames "My Team" to "My Personal Sites" when team's setup is not complete yet.

It also enforces team name change during setup, treating "My Personal Sites" as reserved name. There's a default generated on the basis of user's name.

The CTA is also rephrased to "Setup Team" instead of "Create Team" to express the intent better.

The team name is now used in place of "My Sites" heading in sites list view to clearly identify which team the user is switched to. The "My Personal Sites" heading is also the default for a case where user is not a member of any team yet.

The heading change is hidden behind a flag for now.

Other changes done as a part of this PR:

- sites listing behavior is changed depending on the team state
  - if there's no team passed or the passed team is not setup yet, site invitations, sites transfers, guest sites and owned sites (if personal team exists) are listed
  - if the team is setup already, only transfers and sites for that particular team are listed
- when a personal team is setup, it's now possible to create another personal team by add another site under "My Personal Sites"
- once the team is setup if stops being treated as user's personal team
- an embedded team switcher is introduced, showing up to 3 teams the user can switch between directly from navbar; if user is a member of more teams, a "Switch to Another Team" option is presented
- basically all controller actions using team from context are now always using `current_team`
- `current_team` does fall back to `my_team`, if present but otherwise, remains unset if not picked explicitly

### Tests

### DIFF
--- a/lib/plausible/billing/billing.ex
+++ b/lib/plausible/billing/billing.ex
@@ -145,7 +145,7 @@ defmodule Plausible.Billing do
         # this could result in assigning this new subscription to the newly owned team,
         # effectively "shadowing" any old one.
         #
-        # That's why we are always defaulting to creating a new "My Team" team regardless
+        # That's why we are always defaulting to creating a new "My Personal Sites" team regardless
         # if they were owner of one before or not.
         Auth.User
         |> Repo.get!(user_id)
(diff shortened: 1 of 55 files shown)


aerosol:
@zoldar Setting up a team results with an error rendering team general settings + dummy.site disappears from /sites

aerosol (approved):

PR #5145, opened 2025-03-05
Reveal segment filters in v2 Stats API

apata:
### Changes

Reveals segment filter in v2 Stats API. 

docs: https://github.com/plausible/docs/pull/586

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@ All notable changes to this project will be documented in this file.
 - Add text version to emails plausible/analytics#4674
 - Add acquisition channels report
 - Add filter `is not` for goals in dashboard plausible/analytics#4983
-- Add Segments feature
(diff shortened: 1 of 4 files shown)


apata:
@macobo This has an interesting side-effect of allowing `[:has_not_done, a goal filter]` as a filter applied in v2 Stats API. 
Screenshot 2025-03-05 at 14 01 58

macobo (approved):
There's a few tests that could be updated in test/plausible/stats/query_parser_test.exs as well (removing the :internal atom), otherwise good to go.

macobo:
Re has_done/has_not_done: That's okay, releasing those to the public is in plan though not a direct priority.

apata:
@macobo Thanks for the review and responses! I've updated other :internal tests and documented the usage of Segments (see PR description for linky).

PR #5154, opened 2025-03-06
Fix issue with shared link auth with /current-visitors and segments requests

apata:
### Changes

- Fixes issue current-visitors and segments loading on dashboard accessed via shared link.
- Removes feature gate as feature is available for all.
- Removes feature gate plug.

### Tests
Manually tested

### Changelog

### Documentation

### Dark mode


lib/plausible_web/router.ex
```diff
@@ -221,10 +221,6 @@ defmodule PlausibleWeb.Router do
     end
 
     scope "/:domain/segments", PlausibleWeb.Api.Internal do
-      pipeline :segments_endpoints,
-        do: plug(PlausibleWeb.Plugs.FeatureFlagCheckPlug, [:saved_segments])
```
macobo, line 225: As far as I can tell this module is now unused, so we should also remove it.

apata: I think it may be useful for the next feature

assets/js/dashboard/api.ts
```diff
@@ -15,7 +15,7 @@ export class ApiError extends Error {
   }
 }
 
-function serialize(obj: Record<string, string | boolean | number>) {
+export function serialize(obj: Record<string, string | boolean | number>) {
```
macobo, line 18: Given we now export this, the name is kind of hard to grok.

```suggestion
export function serializeUrlParams(params: Record<string, string | boolean | number>): string {
```

macobo (approved):
One suggestion to consider, otherwise LGTM. :shipit:

PR #5160, opened 2025-03-07
Provide top stats graphable metrics from BE

apata:
### Changes

Fixes the issue with filters inside segments not being counted properly for top stats.

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht (approved):

apata:
I started adding some tests to this, but then I realized that perhaps the better way to solve this would be to include whether the top stat is graphable in the top stat item itself, since graphability is a property of that. I will release this version as a hotfix for now.

PR #5161, opened 2025-03-07
Refactor ListReport to TS, share props

apata:
### Changes

This PR refactors ListReport to Typescript and unifies its props with BreakdownModal. No functional changes. 

This is in anticipation of changing how metrics are determined in both ListReport and BreakdownModal. 

Refactoring how metrics are determined is needed because the BE knows better which metrics are possible for each situation. Especially if some of the filters are wrapped inside a segment.

### Tests
- [x] Manually tested to function the same

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/reports/list.tsx
```diff
@@ -89,51 +57,73 @@ function ExternalLink({ item, externalLinkDest }) {
...
+   *    configurable through the `metrics` prop, which also defines the keys under which
+   *    column values are read, and how they're rendered. */
+  fetchData: () => Promise<TResponse>
+  /** A function to be called directly after `fetchData`. Receives the
+   *     raw API response as an argument. The return value is ignored by ListReport. Allows
+   *     hooking into the request lifecycle and doing actions with returned metadata. For
+   *     example, the parent component might want to control what happens when imported data
+   *     is included or not. */
+  afterFetchData?: (response: TResponse) => void
+  afterFetchNextPage?: (response: TResponse) => void
```
RobertJoonas, line 88: This should only be in `BreakdownModalProps`. `ListReport` does not allow fetching the next page.

assets/js/dashboard/stats/reports/list.tsx
```diff
@@ -89,51 +57,73 @@ function ExternalLink({ item, externalLinkDest }) {
...
+export interface SharedReportProps<
+  TListItem extends Record<string, unknown> = Record<string, unknown>,
+  TResponse = { results: TListItem[]; meta: BreakdownResultMeta }
+> {
+  metrics: Metric[]
+  /** A function that takes a list item and returns the
+   *    that should be applied when the list item is clicked. All existing filters matching prefix
+   *    are removed. If a list item is not supposed to be clickable, this function should
+   *    return `null` for that list item. */
+  getFilterInfo: (item: TListItem) => FilterInfo
```
RobertJoonas, line 69: Some words missing from this comment I believe.

assets/js/dashboard/stats/reports/list.tsx
```diff
@@ -89,51 +57,73 @@ function ExternalLink({ item, externalLinkDest }) {
...
+  /** A function to be called directly after `fetchData`. Receives the
+   *     raw API response as an argument. The return value is ignored by ListReport. Allows
+   *     hooking into the request lifecycle and doing actions with returned metadata. For
+   *     example, the parent component might want to control what happens when imported data
+   *     is included or not. */
+  afterFetchData?: (response: TResponse) => void
+  afterFetchNextPage?: (response: TResponse) => void
+}
+
+type ListReportProps = {
```
RobertJoonas, line 91: Nit: Let's define `ListReportProps` here as a union with `SharedReportProps` already (like we're doing with `BreakdownModalProps`)

assets/js/dashboard/stats/modals/breakdown-modal.tsx
```diff
@@ -30,6 +31,17 @@ export type ReportInfo = {
   defaultOrder?: Order
 }
 
+type BreakdownModalProps<TListItem extends Record<string, unknown>> = Omit<
+  SharedReportProps<TListItem>,
+  'fetchData'
```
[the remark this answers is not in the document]
apata: It's making use of the generic TResponse type that I'd have to repeat if I leave it out. 

Ideally, this would not be present at all, and ListReport would also make use of useQuery(): that way, we could finally start making use of FE response caching for certain dashboard views.

assets/js/dashboard/stats/reports/list.tsx
apata: Ah, well spotted that these were different! Actually, I wanted to do both with the props-level intersection so as not to unnecessarily repeat the generic declaration.

RobertJoonas (approved):
Nice! 👍

PR #5163, opened 2025-03-10
Stop counting engagement events into current visitors

macobo:
This was unexpected breakage from releasing new tracker script with engagement tracking to the world. Users are seeing inflated current visitors counts compared to top sources report.

[Helpscout link](https://secure.helpscout.net/conversation/2869053683/23571?viewId=6980900)

### DIFF
--- a/lib/plausible/stats/current_visitors.ex
+++ b/lib/plausible/stats/current_visitors.ex
@@ -13,6 +13,7 @@ defmodule Plausible.Stats.CurrentVisitors do
       from e in "events_v2",
         where: e.site_id == ^site.id,
         where: e.timestamp >= ^first_datetime,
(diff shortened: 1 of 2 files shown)


cnkk (approved):

aerosol (approved):

aerosol:
BTW does "engagement" custom event (goal name) needs to be reserved?

aerosol:
> BTW does "engagement" custom event (goal name) need to be reserved?

:duck: it already is

macobo:
> BTW does "engagement" custom event (goal name) need to be reserved?

Goal name - no. For the custom event in a goal, `engagement` doesn't show up in in suggested events but we don't explicitly disallow it.

PR #5172, opened 2025-03-11
Fix shared links export

macobo:
This fixes a bug with doing csv exports on shared links.

How to reproduce (on master):
1. Open http://localhost:8000/dummy.site/settings/visibility
2. Make sure stats are _not_ publicly accessible
3. Create a shared link
4. Visit shared link in incognito
5. Click to export

Expected result:
1. Export succeeds

Actual result:
1. Export fails with 404

This bug was introduced by a refactor that removed the `auth` search param from link in https://github.com/plausible/analytics/pull/5154

### DIFF
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -103,7 +103,7 @@ export async function get(
 ) {
   const queryString = query
     ? queryToSearchParams(query, [...extraQueryParams])
-    : null
+    : serializeUrlParams(getSharedLinkSearchParams())
 
   const response = await fetch(queryString ? `${url}?${queryString}` : url, {
     signal: abortController.signal,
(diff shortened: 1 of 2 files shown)


apata (approved):

assets/js/dashboard/api.ts
```diff
@@ -98,13 +101,9 @@ export async function get(
   query?: DashboardQuery,
   ...extraQueryParams: unknown[]
 ) {
-  const sharedLinkParams = getSharedLinkSearchParams()
-
   const queryString = query
-    ? serializeUrlParams(
-        queryToSearchParams(query, [...extraQueryParams, sharedLinkParams])
-      )
-    : serializeUrlParams(sharedLinkParams)
```
apata, line 107: Actually, this might break current-visitors, since query is not applied there.

macobo: You are right! https://github.com/plausible/analytics/pull/5172/commits/f39b1840f9e551e19c6badf4c6eb8d2a98858268

apata (approved):

PR #5184, opened 2025-03-12
Account for multiple team ownerships in HS integration

zoldar:
### Changes

This PR fixes HS integration switching it from cookie-based auth of requests after initial callback to a token-based one with the token embedded in the URL.

It also introduces special handling for owners of multiple teams, so that a list of teams is presented with a possibility to switch to a particular one for more details.

### Tests


aerosol (approved):
:1st_place_medal:

PR #5189, opened 2025-03-13
Don't show new time-on-page metric for customers without flag

macobo:
This broke in a refactor - originally api.ts also checked for the feature flag before sending include.legacy_time_on_page_cutoff but this was incorrectly removed due to conflicting changes.

### DIFF
--- a/assets/js/dashboard/query-context.tsx
+++ b/assets/js/dashboard/query-context.tsx
@@ -107,8 +107,9 @@ export default function QueryContextProvider({
         ? postProcessFilters(filters as Filter[])
         : defaultValues.filters,
       labels: (labels as FilterClauseLabels) || defaultValues.labels,
-      legacy_time_on_page_cutoff:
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/query-context.tsx
```diff
@@ -107,8 +107,9 @@ export default function QueryContextProvider({
         ? postProcessFilters(filters as Filter[])
         : defaultValues.filters,
       labels: (labels as FilterClauseLabels) || defaultValues.labels,
-      legacy_time_on_page_cutoff:
-        (legacy_time_on_page_cutoff as string) || site.legacyTimeOnPageCutoff
+      legacy_time_on_page_cutoff: site.flags.new_time_on_page
+        ? (legacy_time_on_page_cutoff as string) || site.legacyTimeOnPageCutoff
+        : undefined
```
apata, line 112: nitpick: we could grab this from defaultValues.legacyTimeOnPageCutoff, as we do with other fallbacks

apata (approved):

PR #5190, opened 2025-03-13
time-on-page: Mark unreasonably high engagement_time values as 0-s

macobo:
Follow-up to https://github.com/plausible/analytics/pull/5181

Users might have the old version of the script cached for a while yet so avoid ingesting invalid data while that's the case


RobertJoonas (approved):
Q: How can we tell when it's the right time to revert this? Perhaps falling back to 1ms for extreme engagement times would make it possible to analyse how many of these are we still receiving? page being visible for only 1ms should be practically impossible to accomplish.

PR #5196, opened 2025-03-13
Implement team deletion and refactor user deletion

zoldar:
### Changes

This PR adds capability to delete current team from a dedicated section under "Team Settings > Danger Zone". 

User deletion procedure was refactored as well to explicitly prevent deletion if associated personal team has an active subscription. The user deletion routine will be further extended in a follow-up with more explicit messaging for cases of setup teams with the user for sole owner. 

The team deletion procedure also accounts for ensuring the subscription is canceled first before allowing to delete the team.

Team deletion can be carried out only by one of the team owners.

### Tests

### DIFF
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -116,8 +116,12 @@ defmodule Plausible.Auth do
   end
 
   defp delete_user!(user) do
-    Plausible.Segments.user_removed(user)
-    Repo.delete!(user)
+    Repo.transaction(fn ->
+      Plausible.Segments.user_removed(user)
+      Repo.delete!(user)
+    end)
+
+    :ok
   end
 
   defp check_can_leave_teams(teams) do
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -87,7 +87,7 @@ defmodule Plausible.Auth do
         teams = Teams.Users.owned_teams(user)
 
         with :ok <- check_can_leave_teams(teams) do
-          personal_team = Enum.find(teams, & &1.setup_complete)
+          personal_team = Enum.find(teams, &(not Teams.setup?(&1)))
           delete_team_and_user(personal_team, user)
         end
 
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -70,52 +70,72 @@ defmodule Plausible.Auth do
     end
   end
 
+  @spec delete_user(Auth.User.t()) ::
+          {:ok, :deleted} | {:error, :is_only_team_owner | :active_subscription}
   def delete_user(user) do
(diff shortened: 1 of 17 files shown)


lib/plausible/auth/auth.ex
```diff
@@ -70,52 +70,72 @@ defmodule Plausible.Auth do
...
-            from ep in Plausible.Billing.EnterprisePlan, where: ep.team_id == ^team.id
-          )
+      {:error, :multiple_teams} ->
+        teams = Teams.Users.owned_teams(user)
 
-          Plausible.Segments.user_removed(user)
-          Repo.delete!(team)
-          Repo.delete!(user)
+        with :ok <- check_can_leave_teams(teams) do
+          personal_team = Enum.find(teams, & &1.setup_complete)
```
aerosol, line 90: This bit is confusing.

Why do we check for a first team set up? Say I'm a secondary owner in two teams. How is the first team different than any other on the list?

lib/plausible/auth/auth.ex
```diff
@@ -70,52 +70,72 @@ defmodule Plausible.Auth do
...
 
-  defp check_can_leave_teams!(user) do
-    user
-    |> Teams.Users.owned_teams()
-    |> Enum.reject(&(&1.setup_complete == false))
-    |> Enum.map(fn team ->
-      check_can_leave_team!(team)
+  defp delete_user!(user) do
+    Plausible.Segments.user_removed(user)
+    Repo.delete!(user)
```
aerosol, line 120: no transaction necessary here?

lib/plausible/auth/auth.ex
zoldar: Agh, crap, that was supposed to be:

```elixir
personal_team = Enum.find(teams, &(not &1.setup_complete))
```

or better yet:

```elixir
personal_team = Enum.find(teams, &(not Teams.setup?(&1)))
```

The should be at most 1 such team per user.

We want to delete the personal team in case the user is a member in multiple teams too.

I'll add a fix and a relevant test, which I've clearly missed. Thanks.

lib/plausible/auth/auth.ex
zoldar: Right, it was on my mind to add it eventually, but I forgot 🤦  Thanks

lib/plausible/auth/auth.ex
zoldar: https://github.com/plausible/analytics/pull/5196/commits/6eca3b687377554c857bb19d7c2e906f690b3b78

lib/plausible/auth/auth.ex
zoldar: https://github.com/plausible/analytics/pull/5196/commits/f86400099cbc9aa9ea193515f4696e76d474e77a

aerosol (approved):

PR #5203, opened 2025-03-17
Store `tracker_script_version` under `ingest_counters`

macobo:
### Changes

This will enable tracking how many events per tracker_script_version to track the rollout for any issues.

Migration PR: https://github.com/plausible/analytics/pull/5202 (to be deployed first)
Tracker PR: https://github.com/plausible/analytics/pull/5198

### DIFF
--- a/lib/plausible/ingestion/counters/record.ex
+++ b/lib/plausible/ingestion/counters/record.ex
@@ -13,5 +13,6 @@ defmodule Plausible.Ingestion.Counters.Record do
     field :domain, Ch, type: "LowCardinality(String)"
     field :metric, Ch, type: "LowCardinality(String)"
     field :value, Ch, type: "UInt64"
+    field :tracker_script_version, Ch, type: "UInt16"
   end
 end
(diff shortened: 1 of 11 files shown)


lib/plausible/ingestion/counters/record.ex
```diff
@@ -13,5 +13,6 @@ defmodule Plausible.Ingestion.Counters.Record do
     field :domain, Ch, type: "LowCardinality(String)"
     field :metric, Ch, type: "LowCardinality(String)"
     field :value, Ch, type: "UInt64"
+    field :tracker_script_version, Ch, type: "UInt16"
```
aerosol, line 16: What's the reasoning behind UInt16 choice here?

Would it be better to use LowCardinatlity(String) so we could annotate script versions in a more meaningful way (such as semver/tags etc.)? Would it make it easier to prevent any unwanted clashes/duplication?

And a follow-up to this - should we have a compile-time allow-list of supported versions instead of accepting anything sent?

lib/plausible/ingestion/event.ex
```diff
@@ -96,7 +96,8 @@ defmodule Plausible.Ingestion.Event do
   def emit_telemetry_buffered(event) do
     :telemetry.execute(telemetry_event_buffered(), %{}, %{
       domain: event.domain,
-      request_timestamp: event.request.timestamp
+      request_timestamp: event.request.timestamp,
+      tracker_script_version: event.request.tracker_script_version || 0
```
aerosol, line 100: WYT about declaring the default value in `Request` struct definition?

aerosol (approved):
@RobertJoonas asked me to review this. Looks like it's working, I left two comments for you to consider @macobo

PR #5206, opened 2025-03-17
Tracker: Pause engagement_time when browser loses focus

macobo:
Consider this scenario:
1. User opens page /blog
2. 3 seconds later, alt-tabs to another window
3. minute later, alt-tabs to the page and closes it

Previously, this would be reported as 1m03s engagement_time. Now this would be reported as 3s engagement_time.

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -151,20 +151,24 @@
     }
   }
 
+  function onVisibilityChange() {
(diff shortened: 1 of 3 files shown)


RobertJoonas (approved):
Looks like this makes us capture engagements on OS task switching too (CMD + Tab; only speaking about Mac here), so I'd expect this to increase the engagement capture rate as well 👍

PR #5209, opened 2025-03-18
Fix pageviews slider

apata:
### Changes

* Fixes the pageviews slider on Firefox by aligning the base of the range input with the base of various texts that appear next to it. 
  * For that, it needs to have same top and bottom margin and vertical-align: middle, which I opted to give using style not TW classes, just to keep it the same as in website #pricing section. 
* Stops slider output text from wrapping around the 1024px screen width mark by changing the font size and removing custom large screen horizontal padding from choose plan page. 

See comment for before / after images.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/choose_plan.ex
+++ b/lib/plausible_web/live/choose_plan.ex
@@ -115,7 +115,7 @@ defmodule PlausibleWeb.Live.ChoosePlan do
 
     ~H"""
     <div class="pt-1 pb-12 sm:pb-16 text-gray-900 dark:text-gray-100">
-      <div class="mx-auto max-w-7xl px-6 lg:px-20">
+      <div class="mx-auto max-w-7xl px-6 lg:px-8">
         <Notice.pending_site_ownerships_notice
           class="pb-6"
           pending_ownership_count={length(@pending_ownership_site_ids)}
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/choose_plan.ex
```diff
@@ -115,7 +115,7 @@ defmodule PlausibleWeb.Live.ChoosePlan do
 
     ~H"""
     <div class="pt-1 pb-12 sm:pb-16 text-gray-900 dark:text-gray-100">
-      <div class="mx-auto max-w-7xl px-6 lg:px-20">
```
apata, line 118: I've got mixed feelings on this fix. I liked the airy padding, but it made the pageview slider output text wrap on screens around 1024px wide. Thoughts?

apata: Before:
Screenshot 2025-03-26 at 11-48-13 Plausible · Simple privacy-friendly alternative to Google Analytics

apata: After:
Screenshot 2025-03-26 at 11-48-50 Plausible · Simple privacy-friendly alternative to Google Analytics

ukutaht (approved):

lib/plausible_web/live/choose_plan.ex
ukutaht: The 'After' version looks better IMO

PR #5216, opened 2025-03-19
Docs links for teams

aerosol:
Ref https://github.com/plausible/docs/pull/589

### DIFF
--- a/lib/plausible_web/live/team_setup.ex
+++ b/lib/plausible_web/live/team_setup.ex
@@ -63,7 +63,7 @@ defmodule PlausibleWeb.Live.TeamSetup do
         <div class="flex justify-between">
           <div>Create a new team</div>
           <div class="ml-auto">
-            <.docs_info slug="user-roles" />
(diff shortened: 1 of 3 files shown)


lib/plausible_web/live/team_setup.ex
```diff
@@ -59,7 +59,14 @@ defmodule PlausibleWeb.Live.TeamSetup do
   def render(assigns) do
     ~H"""
     <.focus_box>
-      <:title>Create a new team</:title>
+      <:title>
+        <div class="flex justify-between">
+          <div>Create a new team</div>
+          <div class="ml-auto">
+            <.docs_info slug="user-roles" />
```
zoldar, line 66: Shouldn't that be `users-roles`, as per https://github.com/plausible/docs/pull/589/files ? (applies to the other occurrences as well).

aerosol: :facepalm: nice!

zoldar (approved):

PR #5220, opened 2025-03-19
Add `GET /api/v1/sites/teams` endpoint and make related adjustments

zoldar:
### Changes

This PR adds a Sites API endpoint for listing teams the user is a member in. The team identifier from that endpoint can be used when either creating sites via `POST /api/v1/sites` to create a site under a particular team or scope the list of sites  to a particular team with `GET /api/v1/sites?team_id=...`.

The team identifier input was removed from Team Settings to avoid confusing users who don't actively use the API, which is the only place where identifier is explicitly used so far.

The endpoint is under scope which is implicitly available to all API keys. The `api_available` field on team entry determines if the team has Stats API feature enabled. If it's disabled, it's not possible to create sites under it (or, for that matter, run other API queries against its sites).

The `GET /api/v1/sites` endpoint was adjusted to scope sites to a team exclusively, if an identifier is provided.

### TODO

- [x] ~make updates to API docs after https://github.com/plausible/docs/pull/591 is in~ extended docs in https://github.com/plausible/docs/pull/591

### Tests

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -231,7 +231,7 @@ defmodule Plausible.Sites do
       where(
         query,
         [team_memberships: tm, guest_memberships: gm, site: s],
-        (tm.role != :guest and tm.team_id == ^team.id) or gm.site_id == s.id
+        tm.role != :guest and tm.team_id == ^team.id
       )
     else
       where(
--- a/test/plausible_web/controllers/api/external_sites_controller_test.exs
+++ b/test/plausible_web/controllers/api/external_sites_controller_test.exs
@@ -14,6 +14,68 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
       {:ok, api_key: api_key, conn: conn}
     end
 
+    describe "GET /api/v1/sites/teams" do
+      test "shows empty list when user is not a member of any team", %{conn: conn} do
+        conn = get(conn, "/api/v1/sites/teams")
+
+        assert json_response(conn, 200) == %{
+                 "teams" => [],
+                 "meta" => %{
+                   "before" => nil,
+                   "after" => nil,
+                   "limit" => 100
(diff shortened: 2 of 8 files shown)


lib/plausible/sites.ex
```diff
@@ -170,7 +170,7 @@ defmodule Plausible.Sites do
       where(
         query,
         [team_memberships: tm, guest_memberships: gm, site: s],
-        (tm.role != :guest and tm.team_id == ^team.id) or gm.site_id == s.id
+        tm.role != :guest and tm.team_id == ^team.id
```
aerosol, line 173: does this not break any existing behaviour for site guests? :thinking:

test/plausible_web/controllers/api/external_sites_controller_test.exs
```diff
@@ -13,6 +13,68 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
...
+                   },
+                   %{
+                     "id" => team_without_stats.identifier,
+                     "name" => "Team Without Stats API",
+                     "api_available" => false
+                   },
+                   %{
+                     "id" => personal_team.identifier,
+                     "name" => "My Personal Sites",
+                     "api_available" => false
```
aerosol, line 66: Wondering if teams with API unavailable should be even exposed.

aerosol (approved):

test/plausible_web/controllers/api/external_sites_controller_test.exs
zoldar: It's a major PITA to exclude them from pagination though.

lib/plausible/sites.ex
zoldar: This function isn't used anywhere else aside from indexing sites in Sites API, so it shouldn't break anything.

PR #5229, opened 2025-03-22
sessions transfer

ruslandoga:
Attempt at transferring in-memory sessions using shared `/persistent` cache.


test/plausible/session/persistence_test.exs
```diff
@@ -0,0 +1,95 @@
...
+    session = struct!(default, overrides)
+    key = {session.site_id, session.user_id}
+    :peer.call(pid, Plausible.Cache.Adapter, :put, [:sessions, key, session, [dirty?: true]])
+  end
+
+  defp get_session(pid, key) do
+    :peer.call(pid, Plausible.Cache.Adapter, :get, [:sessions, key])
+  end
+
+  defp await_transfer(pid, timeout \\ 1000) do
```
ruslandoga, line 76:
```suggestion
  defp await_transfer(pid, timeout \\ :timer.seconds(10)) do
```

lib/plausible/session/transfer.ex
```diff
@@ -0,0 +1,135 @@
...
+  def attempted?(transfer \\ __MODULE__) do
+    not taker_alive?(transfer)
+  end
+
+  defp taker_alive?(sup) do
+    children = Supervisor.which_children(sup)
+    taker = Enum.find_value(children, fn {id, pid, _, _} -> id == :taker && pid end)
+    if is_pid(taker), do: Process.alive?(taker), else: false
+  catch
+    :exit, :noproc -> false
```
ruslandoga, line 27:
```suggestion
    :exit, {:noproc, _} -> false
```

lib/plausible/session/transfer.ex
```diff
@@ -0,0 +1,135 @@
...
+
+    Supervisor.init([taker, giver, alive], strategy: :one_for_one)
+  end
+
+  defp session_version do
+    [
+      Plausible.ClickhouseSessionV2.module_info(:md5),
+      Plausible.Cache.Adapter.module_info(:md5),
+      Plausible.Session.CacheStore.module_info(:md5),
+      __MODULE__.module_info(:md5)
```
ruslandoga, line 81:
```suggestion
      Plausible.Session.Transfer.module_info(:md5)
```

lib/plausible/session/transfer.ex
```diff
@@ -0,0 +1,146 @@
+defmodule Plausible.Session.Transfer do
+  @moduledoc """
+  Cross-deployment transfer for `:sessions` cache.
+
+  It works by establishing a client-server architecture where:
+  - The "taker" one-time task retrieves `:sessions` data from other processes via Unix domain sockets
```
ruslandoga, line 6:
```suggestion
  - The "taker" one-time task retrieves `:sessions` data from other OS processes via Unix domain sockets
```

lib/plausible/session/transfer/tinysock.ex
```diff
@@ -0,0 +1,170 @@
...
+  @moduledoc false
+  use GenServer
+  require Logger
+
+  @tag_data "tinysock"
+  @tag_size byte_size(@tag_data)
+
+  @spec list!(Path.t()) :: [Path.t()]
+  def list!(base_path) do
+    for @tag_data <> _rand = name <- File.ls!(base_path) do
```
aerosol, line 11: Think we should also include `APP_HOST` env var in the tag, so that nodes only transfer "their" chunks even when the storage is shared with multiple machines.

lib/plausible/session/transfer/alive.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Plausible.Session.Transfer.Alive do
+  @moduledoc false
```
aerosol, line 2: It would be beneficial to keep the docs.

lib/plausible/session/transfer/tinysock.ex
aerosol: This is non-standard I guess, but we can't use `HOSTNAME` as it's different with every image.

lib/plausible/session/transfer.ex
```diff
@@ -0,0 +1,149 @@
...
+        {key, %Plausible.ClickhouseSessionV2{} = session} = record
+        Plausible.Cache.Adapter.put(:sessions, key, session)
+      end)
+    end
+  end
+
+  defp file_stat_ctime(path) do
+    case File.stat(path) do
+      {:ok, stat} -> stat.ctime
+      {:error, _} -> nil
```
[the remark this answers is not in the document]
ruslandoga: I haven't thought about much, but the idea was to allow tinysock client to remove stale/broken files in [`sock_connect_or_rm`](https://github.com/plausible/analytics/blob/cac06c428ce9d964d97eabf8babeffd8f7da127a/lib/plausible/session/transfer/tinysock.ex#L120-L134) And since tuple (normal stat.ctime) is > than nil, this ordering is like "asc nulls last" so it's OK to do the cleanup in the end of the transfer.

lib/plausible/session/transfer/alive.ex
ruslandoga: I wanted to have only one place with the docs, `Plausible.Session.Transfer`. It kind of explains the role of the "alive" process.

aerosol (approved):

PR #5232, opened 2025-03-24
Add final usage check before locking sites

RobertJoonas:
### Changes

This PR fixes a bug described below:

Our `CheckUsage` and `SiteLocker` jobs are two independent workers. The first is setting grace periods for users and the other one is acting upon them. The bug is that the `SiteLocker` doesn't look at the usage. It just blindly locks a team when their grace period is over.

Because of that, it's possible that when a user lowers their usage (e.g. deletes or transfers some sites) at the last moment (i.e. after the last CheckUsage job that runs daily at 2PM UTC **AND** before midnight UTC the next day) they might still get locked

This PR introduces a "final check" before we actually proceed with locking the team and sending the "Your stats are now locked" email. If they turn out to have lowered their usage. They'll remain unlocked and their grace period will be silently deleted.

NOTE: This also means that even when a team is locked for a month, the `SiteLocker` job will still automatically check their usage every midnight **IF they still have an active subscription**, and unlock them if their usage has dropped back within their limits.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -147,14 +147,14 @@ defmodule Plausible.Teams.Billing do
         {:needs_to_upgrade, :no_active_trial_or_subscription}
 
       Teams.GracePeriod.expired?(team) ->
-        check_usage_last_chance(team, usage_mod)
+        revise_pageview_usage(team, usage_mod)
 
       true ->
         :no_upgrade_needed
     end
   end
 
-  defp check_usage_last_chance(team, usage_mod) do
+  defp revise_pageview_usage(team, usage_mod) do
     case Plausible.Workers.CheckUsage.check_pageview_usage_two_cycles(team, usage_mod) do
       {:over_limit, _} ->
         {:needs_to_upgrade, :grace_period_ended}
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -128,31 +128,39 @@ defmodule Plausible.Teams.Billing do
     |> Repo.one()
   end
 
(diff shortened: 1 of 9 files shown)


lib/plausible/teams/billing.ex
```diff
@@ -128,35 +128,43 @@ defmodule Plausible.Teams.Billing do
...
       Teams.GracePeriod.expired?(team) ->
-        {:needs_to_upgrade, :grace_period_ended}
+        check_usage_last_chance(team, usage_mod)
 
       true ->
         :no_upgrade_needed
     end
   end
 
+  defp check_usage_last_chance(team, usage_mod) do
```
zoldar, line 157: `revise_usage` ? or `revise_pageview_usage`?

zoldar (approved):
LGTM ✨  A side-note - moving from `sites.locked` to `teams.locked` would be good to do to clean things up a bit, eventually. Maybe something for cooldown 😅

lib/plausible/teams/billing.ex
RobertJoonas: Good catch! this function gets called when rendering the `/sites` page too (by the `upgrade_nag_screen`) so last chance doesn't sound like a good fit indeed.

RobertJoonas:
> A side-note - moving from sites.locked to teams.locked would be good to do to clean things up a bit, eventually. Maybe something for cooldown

Makes sense. Indeed - would be nice to refactor and clean this up separately indeed. A lot of complexity and technical debt here. Another thing that I think should change is that grace periods should belong to subscriptions.

PR #5238, opened 2025-03-25
Change site team

aerosol:
### Changes

In case the user belongs to more than 1 team, in either admin or owner roles, Site's Danger Zone section allows changing the team. Team change is effectively identical to inviting user's own e-mail to site transfer, except bypasses the invitation flow.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/site/membership_controller.ex
+++ b/lib/plausible_web/controllers/site/membership_controller.ex
@@ -15,7 +15,12 @@ defmodule PlausibleWeb.Site.MembershipController do
   use Plausible
   alias Plausible.Site.Memberships
 
-  @only_owner_and_admin_is_allowed_to [:transfer_ownership_form, :transfer_ownership]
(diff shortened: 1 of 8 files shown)


lib/plausible_web/controllers/site/membership_controller.ex
```diff
@@ -137,6 +137,75 @@ defmodule PlausibleWeb.Site.MembershipController do
     end
   end
 
+  def change_team_form(conn, _params) do
```
zoldar, line 140: I think both new actions should be guarded by the plug by adding them to `@only_owner_and_admin_is_allowed_to`?

aerosol: Done via https://github.com/plausible/analytics/pull/5238/commits/524e6c1aa30c340ca3017026177cbdc4002d6aba

zoldar (approved):
✨

PR #5240, opened 2025-03-26
Stop counting today in `7d` and `30d` periods + add `28d` and `90d`

RobertJoonas:
### Changes

Companion docs PRs:

1. https://github.com/plausible/docs/pull/594 (adding the new periods to Stats API docs)
2. https://github.com/plausible/docs/pull/596 (keyboard shortcuts)

* Remove today from `7d` and `30d` periods. Note that:
  * `7d` is currently `[today-6d, today]` (seven days in total) and will change to `[today-7d, today-1d]`, while
  * `30d` is currently `[today-30d, today]` (thirty **one** days in total) but will now change to `[today-30d, today-1d]` with thirty days in total
* Adds two new periods to the dashboard and public API - `28d` and `90d` (shortcuts `F` for `28d` and `N` for `90d`)
* Adds a new keyboard shortcut `P` for the "Last Month" period.
* Removes `30d` period from the dashboard date-picker dropdown (in favour of `28d`)
  * The `T` keybind for it will continue to work though

### Tests

### Changelog

### Documentation
- [x] PRs pending links above

### Dark mode


lib/workers/send_email_report.ex
```diff
@@ -73,8 +73,7 @@ defmodule Plausible.Workers.SendEmailReport do
   end
 
   defp put_last_week_query(%{site: site} = assigns) do
-    today = DateTime.now!(site.timezone) |> DateTime.to_date()
-    date = Date.shift(today, week: -1) |> Timex.end_of_week() |> Date.to_iso8601()
+    date = DateTime.now!(site.timezone) |> Timex.beginning_of_week() |> Date.to_iso8601()
```
RobertJoonas, line 76: I was a bit confused here so leaving a friendly note for the reviewer.

Weekly reports are sent every Monday at 9AM. So we should be able to just pass `today(site.timezone)` as the `date` into the query (or don't pass anything since it's the default anyway). But this shifting to the beginning of the week is going on because of the tests. Otherwise we'd have to pass a `now` into this function.

assets/js/dashboard/query-time-periods.ts
```diff
@@ -341,6 +356,19 @@ export const getDatePeriodGroups = ({
...
         }
+      ],
+      [
+        ['Last 90 Days', 'N'],
+        {
+          search: (s) => ({
+            ...s,
+            ...clearedDateSearch,
+            period: QueryPeriod['90d'],
+            keybindHint: 'T'
```
apata, line 367: tiny issue: keybindHint param doesn't match given keybind.
keybind suggestion: Q for quarter year?

RobertJoonas: Fixed the keybind hint: https://github.com/plausible/analytics/pull/5240/commits/fe483dce04cfde01dfd1f988653b65f0a91bb594

lib/workers/send_email_report.ex
RobertJoonas: Good call! Changed for both monthly and weekly reports in https://github.com/plausible/analytics/pull/5240/commits/5772f8cde21a56bbbea3033f1d9c1c4aa464e0ea.

Also left a comment about what confused me earlier.

assets/js/dashboard/query-time-periods.ts
```diff
@@ -329,9 +332,23 @@ export const getDatePeriodGroups = ({
...
+            keybindHint: 'F'
+          }),
+          isActive: ({ query }) => query.period === QueryPeriod['28d'],
+          onEvent
+        }
+      ],
       [
         ['Last 30 Days', 'T'],
         {
+          hidden: true,
```
apata, line 351: nitpick: check "Last 6 months" item to see how hidden items are currently handled

apata: Also, maybe we want to mention the fact that 30d is now hidden from the menu in the changelog, and that 28d is the new default view?

RobertJoonas: Ah, thanks for pointing that out! Do you have anything against moving the "Last 6 months" item into the full `groups` list with the `hidden: true` option? Would make it easier to discover I think.

ukutaht (approved):

PR #5243, opened 2025-03-27
Supervise user agent parsing

aerosol:
### Changes

Be vigilant about UA parsing congestion by ensuring a maxmium execution time is enforced.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -238,13 +238,13 @@ defmodule Plausible.Ingestion.Event do
 
   defp put_user_agent(%__MODULE__{} = event, _context) do
     case parse_user_agent(event.request) do
-      %UAInspector.Result{client: %UAInspector.Result.Client{name: "Headless Chrome"}} ->
+      {:ok, %UAInspector.Result{client: %UAInspector.Result.Client{name: "Headless Chrome"}}} ->
         drop(event, :bot)
 
-      %UAInspector.Result.Bot{} ->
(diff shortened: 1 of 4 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -430,14 +430,27 @@ defmodule Plausible.Ingestion.Event do
...
 
+  defp parse_user_agent_safe(user_agent) do
+    task =
+      Task.Supervisor.async_nolink(Plausible.UserAgentParseTaskSupervisor, fn ->
+        UAInspector.parse(user_agent)
+      end)
+
+    case Task.yield(task, @parse_user_agent_timeout) || Task.shutdown(task) do
+      {:ok, result} -> {:ok, result}
+      nil -> {:error, :timeout}
```
ruslandoga, line 450: Should we log something to add telemetry to know when it happens? E.g. what if there are normal load conditions when 200ms is not enough to parse a normal user agent.

ruslandoga (approved):

lib/plausible/ingestion/event.ex
aerosol: yes good call

ukutaht (approved):

lib/plausible/ingestion/event.ex
ruslandoga: Yes!

ruslandoga (approved):

PR #5246, opened 2025-03-27
use plain task

ruslandoga:
alternative to #5245


aerosol:
Yeah this is good, merge at will

aerosol (approved):

PR #5248, opened 2025-03-29
Upgrade page fixes

RobertJoonas:
### 1. Do not recommend Growth tier for Business customers

https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8237275669

Stop highlighting the Growth box on the upgrade page for Business subscribers, even if their usage would allow downgrading to Growth. This only creates confusion as the email asks users to upgrade, but on the upgrade page, the Growth plan says "Downgrade to Growth".

Instead we will highlight the Business box as the "Current" tier, and do not "Recommend" any tier at all.

### 2. Stop rendering global billing notices on the upgrade page.

https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7881526960

Notices like "Dashboard locked" or "You've outgrown your subscription tier" were never intended to show up on that page. The only notices that this page should display are:

* Pending ownerships - just to let the user know that these are counted towards their usage on the upgrade page.
* No sites owned - e.g. the user is an invited user and hasn't added any sites for their own team yet. The upgrade page renders in an error state, not allowing the user to click the upgrade buttons.
* Subscription past due or paused - also looks like an error state, the user is not allowed to click the upgrade links. Instead they have to either 1) click the link in the notice to upgrade their billing details or 2) cancel their current subscription and subscribe again, providing new details from scratch

This PR fixes it so that only the intended messages are rendered, and also adds tests to refute any other notices showing up.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/billing/qouta/quota.ex
+++ b/lib/plausible/billing/qouta/quota.ex
@@ -46,15 +46,17 @@ defmodule Plausible.Billing.Quota do
   `:custom` is returned. This means that this kind of usage should get on
   a custom plan.
 
+  To avoid confusion, we do not recommend Growth tiers for customers that
+  are already on a Business tier (even if their usage would fit Growth).
+
   `nil` is returned if the usage is not eligible for upgrade.
   """
-  def suggest_tier(usage, highest_growth_plan, highest_business_plan) do
-    if eligible_for_upgrade?(usage) do
-      cond do
-        usage_fits_plan?(usage, highest_growth_plan) -> :growth
-        usage_fits_plan?(usage, highest_business_plan) -> :business
-        true -> :custom
-      end
+  def suggest_tier(usage, highest_growth, highest_business, owned_tier) do
+    cond do
+      not eligible_for_upgrade?(usage) -> nil
+      usage_fits_plan?(usage, highest_growth) and owned_tier != :business -> :growth
+      usage_fits_plan?(usage, highest_business) -> :business
+      true -> :custom
     end
   end
 
--- a/test/plausible_web/live/choose_plan_test.exs
+++ b/test/plausible_web/live/choose_plan_test.exs
@@ -577,7 +577,7 @@ defmodule PlausibleWeb.Live.ChoosePlanTest do
       assert text_of_element(doc, @slider_value) == "10k"
     end
 
-    test "highlights Business box as the 'Current' tier if it's suitable for their usage", %{
+    test "highlights Business box as the 'Current' tier", %{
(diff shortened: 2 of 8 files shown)


lib/plausible/billing/qouta/quota.ex
```diff
@@ -46,15 +46,17 @@ defmodule Plausible.Billing.Quota do
...
   `nil` is returned if the usage is not eligible for upgrade.
   """
-  def suggest_tier(usage, highest_growth_plan, highest_business_plan) do
-    if eligible_for_upgrade?(usage) do
-      cond do
-        usage_fits_plan?(usage, highest_growth_plan) -> :growth
-        usage_fits_plan?(usage, highest_business_plan) -> :business
-        true -> :custom
-      end
+  def suggest_tier(usage, highest_growth, highest_business, owned_tier) do
```
zoldar, line 54: I'm assuming suggestion does not kick in at all when current owned tier kind is `custom` already, right?

test/plausible_web/live/choose_plan_test.exs
```diff
@@ -1062,23 +1111,50 @@ defmodule PlausibleWeb.Live.ChoosePlanTest do
...
+      Billing.active_grace_period_notice_title(),
+      Billing.subscription_cancelled_notice_title(),
+      Billing.subscription_past_due_notice_title(),
+      Billing.subscription_paused_notice_title(),
+      Billing.upgrade_ineligible_notice_title(),
+      Billing.pending_site_ownerships_notice_title()
+    ]
+    |> Enum.each(fn title ->
+      if title in expected do
+        assert length(String.split(doc, title)) == 2
```
zoldar, line 1151: hmm, why this instead of `assert doc =~ title` or `String.contains?(doc, title)`? Perhaps `String.starts_with?` would be better if you want to reduce the risk of catching substring contained in the other (though capitalisation should prevent that anyway I think)?

zoldar (approved):

lib/plausible/billing/qouta/quota.ex
RobertJoonas: Owned tier being custom would mean that the user will never end up on this page.

test/plausible_web/live/choose_plan_test.exs
RobertJoonas: This makes sure that there's no duplicate warning notices on the screen (i.e. two yellow boxes, one underneath the other).

PR #5252, opened 2025-03-31
Drag-and-Drop statistics

HeadMader:
### Changes

Add DnD(Drag-and-Drop) for statistics and local storage saving for it

### Tests

### Changelog

### Documentation

### Dark mode


HeadMader:
Video preview of feature:

https://github.com/user-attachments/assets/6014b674-7bbe-4ba3-9dce-226ff8a4ed76

ukutaht:
@HeadMader Sorry for the long wait. This is something we'd like to do in the future but it has to be better integrated with the backend. The feature only makes sense if the layout/customization is stored in the database.

PR #5253, opened 2025-03-31
Ingestion: Reject blank engagements with 400

RobertJoonas:
### Changes

Reject "blank engagements" (engagements without any useful information, i.e. no `sd`, nor `e` fields are present in the payload, or both have invalid values) with a 400. 

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/ingestion/request.ex
```diff
@@ -253,20 +252,24 @@ defmodule Plausible.Ingestion.Request do
...
-    end
-  end
-
-  defp put_engagement_time(changeset, %{} = request_body) do
-    if Changeset.get_field(changeset, :event_name) == "engagement" do
       engagement_time = parse_engagement_time(request_body["e"])
 
-      Changeset.put_change(changeset, :engagement_time, engagement_time)
+      case {scroll_depth, engagement_time} do
+        {255, 0} ->
```
aerosol, line 261:
```suggestion
        {@missing_scroll_depth, @missing_engagement_time} ->
```

lib/plausible/ingestion/request.ex
```diff
@@ -253,20 +252,24 @@ defmodule Plausible.Ingestion.Request do
...
+          changeset
+          |> Changeset.add_error(
+            :event_name,
+            "engagement event requires at least one of 'sd' or 'e' fields defined"
```
aerosol, line 265: Do we need to distinguish between `sd` or `e` missing and `sd` or `e` present, but having illegal values - 255/0 ?

lib/plausible/ingestion/request.ex
```diff
@@ -253,20 +252,24 @@ defmodule Plausible.Ingestion.Request do
...
       scroll_depth = parse_scroll_depth(request_body["sd"])
-      Changeset.put_change(changeset, :scroll_depth, scroll_depth)
-    else
-      changeset
-    end
-  end
-
-  defp put_engagement_time(changeset, %{} = request_body) do
-    if Changeset.get_field(changeset, :event_name) == "engagement" do
       engagement_time = parse_engagement_time(request_body["e"])
```
aerosol, line 258: Would it make sense to:

1. Add defaults to schema fields (0 and 255) - if they're really needed. But perhaps nil is enough?
2. Use changeset to validate

```elixir
changeset
|> Changeset.cast(%{scroll_depth: request_body["sd"], engagement_time: request_body["e"]})
|> Changeset.validate_number(:scroll_depth, greater_than: 0, less_than: 100)
|> Changeset.validate_number(:engagement_time, greater_than: 0)
```

This way you don't have to use those custom parsing functions. Unless the "if > 100, then: 100" semantics is important to keep - but this can be still done with a small tweak to the above pipeline. WYT?

lib/plausible/ingestion/request.ex
aerosol: That, assuming my top-most comment is inapplicable.

lib/plausible/ingestion/request.ex
RobertJoonas: I'd say a single error message suffices. Although on second thought, the message could be improved:

```
engagement event requires a valid integer value for at least one of 'sd' or 'e' fields
```

I think 99.9% of cases where this happens is a misconfigured proxy that's not forwarding the `sd` and `e` fields at all. Even if the proxy converts the integer (sent by our script) to string, our ingestion will handle parsing it back to integer.

RobertJoonas: updated message here: https://github.com/plausible/analytics/pull/5253/commits/e4c1eed05bd99581124b45200fba9f3a630f2a0c

lib/plausible/ingestion/request.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5253/commits/e4c1eed05bd99581124b45200fba9f3a630f2a0c

aerosol:
I think first and foremost, this warrants tests in `Plausible.Ingestion.RequestTest`. WYT?

RobertJoonas:
> I think first and foremost, this warrants tests in Plausible.Ingestion.RequestTest. WYT?

@aerosol which functionality are you referring to? `external_controller_test.exs` covers this new 400 case.

aerosol (approved):

PR #5255, opened 2025-03-31
Update headlessui to v1.7.19, refactor site switcher

apata:
### Changes

- ~~Updated headlessui/react to 2 (https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/CHANGELOG.md#changed-2)~~
  - ~~changed how some components there are imported~~
  - this update created a sneaky production-only issue with **all dropmenus** becoming unusable when navigating with browser back on Firefox. Holding off on updating for now.

- Refactored all components that used headlessui Menu to Popover, because the focus management in the new Menu made our UI jump badly, probably due to our sticky top bar
  - https://headlessui.com/react/menu#menu-items:~:text=prop%20is%20set.-,modal,-true setting this to false helped with the jump, but it ruined the keyboard navigability of the menu
  - changed:
    - main graph interval menu 
    - filter operator menu 
    - funnels menu
    - sources report UTM selection menu

- Unified dashboard list report tabs
  - considered headlessui Tabs component, but deemed their keyboard navigability too unexpected, plus hard to marry with funnels and UTM_... dropmenus

- Refactored site switcher to Popover
  - tiny visual change: aligned the button visually with the user account menu
  - tiny visual change: aligned the dropmenu with date period menu
  - medium visual change: changed our placeholder favicon to a svg (looks better IMO, but debatable and optional)

- Refactored filter menu segments section to have scroll to make it usable on mobile (requires the mobile phone to be held vertically)

- Refactored Funnels dropdown to have a title and searching, like segments section in filters menu, because without something like this the updated menu wasn't usable with a lot of funnels.
Screenshot 2025-04-28 at 14 06 32 

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/components/tabs.tsx
```diff
@@ -0,0 +1,128 @@
+import {
```
ukutaht, line 1: Great to have these unified ❤️

We've had an annoying alignment issue with tab links for ages where the active link jumps higher than the other non-active ones. Curiously the source-list one seems to not have that issue on prod. I'm not sure off the top of my head what the difference is but is it possible to use the properly aligned variant here in the unified tabs?

assets/js/dashboard/navigation/routeless-modals-context.tsx
```diff
@@ -24,13 +36,29 @@ export function RoutelessModalsContextProvider({
 }: {
   children: ReactNode
 }) {
+  const [openDropmenus, setOpenDropmenus] = useState<string[]>([])
+
+  const registerDropmenuState = useCallback(
```
ukutaht, line 41: Q: What's the purpose of tracking this state?

ukutaht (commented):
>    tiny visual change: aligned the button visually with the user account menu
>    tiny visual change: aligned the dropmenu with date period menu
>    medium visual change: changed our placeholder favicon to a svg (looks better IMO, but debatable and optional)

Chef's kiss 🤌

assets/js/dashboard/navigation/routeless-modals-context.tsx
apata: Good question. There's an annoying bug that the site switcher currently has with Firefox (more details on Basecamp) and this is an attempt to solve it, but doesn't fully solve it. It's pushed to branch because the bug repros in staging only. Ideally, I'd find some other fix without this particular logic.

assets/js/dashboard/components/tabs.tsx
apata: Sure! I actually thought the jump was the desired behavior (sort of like a pinned folder thingie) and established it for all, but I can easily go for the baseline.

assets/js/dashboard/navigation/routeless-modals-context.tsx
ukutaht: Gotcha. I think it's ok as a workaround but could use a comment to let future readers know why it's there.

assets/js/dashboard/components/tabs.tsx
apata: Baseline is established now.

assets/js/dashboard/navigation/routeless-modals-context.tsx
apata: Downgrading to headlessui-react@1... allowed me to get rid of it.

assets/js/dashboard/components/filter-operator-selector.js
```diff
@@ -7,97 +7,88 @@ import {
...
-                  )}
-                </div>
-              </Menu.Items>
+                  ]
+                ]
+                  .filter(([_operation, supported]) => supported)
+                  .map(([operation]) => (
+                    <button
+                      key={operation}
+                      data-selected={operation === props.selectedType}
```
ukutaht, line 67: Q: is this used anywhere? I'm not seeing it

assets/js/dashboard/components/search-input.tsx
```diff
@@ -1,20 +1,28 @@
...
+  useState,
+  useRef,
+  RefObject
+} from 'react'
 import { isModifierPressed, Keybind } from '../keybinding'
 import { useDebounce } from '../custom-hooks'
 import classNames from 'classnames'
 
 export const SearchInput = ({
+  searchRef,
```
ukutaht, line 13: Q: What's the purpose of this searchRef? I find this component quite a bit harder to understand with two `ref`s in the game

ukutaht (approved):
Approve on the overall PR. If something can be done about the double `ref`s in `SearchInput` then I think it's worth simplifying it but not sure

assets/js/dashboard/components/filter-operator-selector.js
apata: Ah, yes. I'm using the tailwind antipattern of having classnames within variables so it isn't obvious, but `popover.items.classNames.hoverLink` applies styles on `data-[selected=true]`

assets/js/dashboard/components/search-input.tsx
apata: I agree that this should be clearer. Sharing my thought process here:

- This component is an uncontrolled input (from React perspective) to allow flicker-free debouncing of onSearch().
- Because it's uncontrolled, if I want to clear the input from outside this component, I need to pass it a ref and do `searchRef.current.value = ''`
**- But I don't want to mandate a ref be provided from the outside, because this component knows how to focus and blur itself**
- Therefore I fall back to local useRef if it's not provided

I think bolded part was the mistake. It's easy enough to make the ref from the outside be mandatory. Fixed now.

PR #5258, opened 2025-03-31
Fix favicons for domains including a subfolder

RobertJoonas:
### Changes

https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8360831680

Fixes fetching favicons from DuckDuckGo if there's a subfolder provided with the domain name. DuckDuckGo only resolves favicons at a domain level, so we need to strip the pathname before we send the request to DDG.

For example, in our [referer_favicon_domains.json](https://github.com/plausible/analytics/blob/master/priv/referer_favicon_domains.json) we have a few sources that include forward slashes:

```json
{
  "Orange Webmail": "orange.fr/webmail",
  "Bing Images": "bing.com/images/search",
  "Google Product Search": "google.ac/products",
}
```

* https://icons.duckduckgo.com/ip3/orange.fr/webmail.ico -> returns placeholder
* https://icons.duckduckgo.com/ip3/orange.fr.ico -> returns the favicon

Stripping the pathname in the `favicon.ex` plug also fixes fetching favicons on the "/sites" page.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/plugs/favicon.ex
+++ b/lib/plausible_web/plugs/favicon.ex
@@ -94,11 +94,11 @@ defmodule PlausibleWeb.Favicon do
       "/favicon/sources/placeholder" ->
         send_placeholder(conn)
 
-      "/favicon/sources/" <> source ->
-        clean_source = URI.decode_www_form(source)
+      "/favicon/sources/" <> domain ->
+        domain = URI.decode_www_form(domain)
 
         domain =
-          Map.get(favicon_domains, clean_source, clean_source)
+          Map.get(favicon_domains, domain, domain)
           |> String.split("/", parts: 2)
           |> hd()
 
--- a/lib/plausible_web/plugs/favicon.ex
+++ b/lib/plausible_web/plugs/favicon.ex
@@ -96,7 +96,11 @@ defmodule PlausibleWeb.Favicon do
 
       "/favicon/sources/" <> source ->
         clean_source = URI.decode_www_form(source)
-        domain = Map.get(favicon_domains, clean_source, clean_source)
+
+        domain =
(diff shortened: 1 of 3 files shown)


lib/plausible_web/plugs/favicon.ex
```diff
@@ -96,7 +96,11 @@ defmodule PlausibleWeb.Favicon do
 
       "/favicon/sources/" <> source ->
         clean_source = URI.decode_www_form(source)
-        domain = Map.get(favicon_domains, clean_source, clean_source)
+
+        domain =
+          Map.get(favicon_domains, clean_source, clean_source)
+          |> String.split("/", parts: 2)
+          |> hd()
```
apata, line 103: nitpick: Functionally this is good, but the change highlights that this function has a bit of inaccurate naming already. favicon_domains map is implying that there's no further action needed to get the domain, yet as this PR shows, it is. `favicon_sources`? Maybe a good time to rename those to be accurate?

apata (approved):

lib/plausible_web/plugs/favicon.ex
RobertJoonas: Good point. However, I think `domain` is still the best naming there is, because this plug gets called to fetch favicons for our sites too. Our sites have a `domain` field as well and we also allow including the subfolder in that "domain".

Not entirely sure why some "domains" in the `referer_favicon_domains.json` include the pathname, but I don't think rename to `source` just because of that.

I've renamed `source` to `domain` now in the function. https://github.com/plausible/analytics/pull/5258/commits/e296a41db9e26da55818b025ae6ae41537b3d30e

PR #5260, opened 2025-04-01
Update eslint and use eslint flat config

apata:
### Changes

* Updates eslint and plugins
* Migrates to new flat eslint config (which is to be the only way going forward from eslint v10)
* Silences some eslint nags with ignore comments, removes some unnecessary eslint ignore comments

### Tests
- [x] App builds and works the same as ever
- [x] App lints locally and in CI with no complaints

### Changelog

### Documentation

### Dark mode


RobertJoonas (approved):
❤️

PR #5284, opened 2025-04-07
Set current_team when user has team membership in viewed site 

zoldar:
### Changes

This PR enforces setting `current_team` to authorized site's team if it's different but only when user has a team membership in it.

Also, service-level checks for invitations were fixed to exclude editor role from managing invitations (it's currently enforced on plug level anyway but this change makes the service consistent with plug behavior).

### Tests

### DIFF
--- a/lib/plausible/teams/membership.ex
+++ b/lib/plausible/teams/membership.ex
@@ -11,6 +11,8 @@ defmodule Plausible.Teams.Membership do
 
   @type t() :: %__MODULE__{}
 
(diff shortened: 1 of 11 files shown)


lib/plausible/teams/membership.ex
```diff
@@ -11,6 +11,8 @@ defmodule Plausible.Teams.Membership do
 
   @type t() :: %__MODULE__{}
 
+  @type role() :: unquote(Enum.reduce(@roles, &{:|, [], [&1, &2]}))
```
aerosol, line 14: :smiling_imp:

aerosol (approved):

PR #5287, opened 2025-04-07
Remove tables and columns that are unused after migrating to teams

zoldar:
### Changes

Migration cleaning up Postgres schema from tables and columns which are no longer used after migrating to teams.

The constraints are modified to cascade as well to ensure data integrity and no orphaned entries.

The migration will be only run on EE. CE will be handled separately when preparing to cut another release.

### DIFF
--- a/priv/repo/migrations/20250407110434_remove_unused_tables_and_columns.exs
+++ b/priv/repo/migrations/20250407110434_remove_unused_tables_and_columns.exs
@@ -25,7 +25,7 @@ defmodule Plausible.Repo.Migrations.RemoveUnusedTablesAndColumns do
 
       alter table(:subscriptions) do
         remove :user_id
-        modify :team_id, references(:teams, on_delete: :delete_all)
+        modify :team_id, references(:teams, on_delete: :delete_all), null: false
       end
 
       drop constraint(:enterprise_plans, "enterprise_plans_user_id_fkey")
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20250407110434_remove_unused_tables_and_columns.exs
```diff
@@ -0,0 +1,50 @@
...
+        remove :trial_expiry_date
+        remove :grace_period
+        remove :allow_next_upgrade_override
+      end
+
+      drop constraint(:sites, "sites_team_id_fkey")
+
+      alter table(:sites) do
+        remove :accept_traffic_until
+        modify :team_id, references(:teams, on_delete: :delete_all)
```
zoldar, line 19: I'm not 100% sure about that one. We do run a site removal service for each site during deletion and that should be a requirement. Perhaps it's better to leave an orphaned site than to remove it without cleaning up its data completely? \cc @aerosol

aerosol: What are the impliciations of having/accessing an orphaned site data though? :thinking:

aerosol: If we assume we always delete through a dedicated service, then it should be still safe to keep the cascading constraint here, right? 
Another thing is, raw deletion is exposed in CRM. So what should happen if someone clicks it?

aerosol: Ah no, even in CRM it's a service call

zoldar: Yeah, I'll leave it as nullable then.

zoldar: https://github.com/plausible/analytics/pull/5287/commits/c6252d29ebed9b3ec39875a3edb6b98fd2655965

zoldar:
Putting that migration on hold until situation with CE release is sorted.

aerosol (approved):

PR #5289, opened 2025-04-07
Sort by exit rate

RobertJoonas:
### Changes

This PR adds `exit_rate` as an (internal) API v2 metric. This makes it possible to:

* sort by the metric in the dashboard
* allow exit rate comparisons

Note that this PR only does the minimum that's required for API v2 to support the dashboard `exit_rate` metric:

* exit_rate can only be queried with a single dimension - "visit:exit_page". It would also make sense to allow exit rate with a "visit:exit_page" filter or in combination with other session dimensions, but that's out of scope for now.
* Session filters can be used
* Event filters cannot be used (which kind of makes sense and is the current dashboard behaviour)

### Tests

### Changelog

### Documentation

### Dark mode


apata (approved):
Nice work! I think doing the least that was needed is the way to go.

test/plausible/stats/query_parser_test.exs
```diff
@@ -1800,6 +1800,73 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
     end
   end
 
+  describe "exit_rate metric" do
+    test "fails validation on its own", %{site: site} do
```
apata, line 1804: Nitpick: “on its own” could be more specific like “fails when used without visit:exit_page dimension”

test/plausible/stats/query_parser_test.exs
```diff
@@ -1800,6 +1800,73 @@ defmodule Plausible.Stats.Filters.QueryParserTest do
...
+        "date_range" => "all"
+      }
+      |> check_error(
+        site,
+        "Event metric(s) `pageviews` cannot be queried along with session dimension(s) `visit:exit_page`",
+        :internal
+      )
+    end
+
+    test "passing validation", %{site: site} do
```
apata, line 1846: Meganitpick: many other tests use test(“does something”) tense, this one stood out a bit.

lib/plausible/stats/table_decider.ex
```diff
@@ -149,6 +149,7 @@ defmodule Plausible.Stats.TableDecider do
   defp metric_partitioner(_, :time_on_page), do: :event
   defp metric_partitioner(_, :visit_duration), do: :session
   defp metric_partitioner(_, :views_per_visit), do: :session
+  defp metric_partitioner(_, :exit_rate), do: :session
```
macobo, line 152: Couldn't this be `:either` given it's calculated via SpecialMetrics? We'd need to adjust the `__internal_visitors` key though.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- Ability to sort by and compare the `exit_rate` metric in the dashboard Exit Pages > Details report
```
apata, line 7: Nitpick: we could also mention that we changed/fixed the way the metric is rounded on dashboard and in exports

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -934,13 +934,20 @@ defmodule PlausibleWeb.Api.StatsController do
     params = Map.put(params, "property", "visit:exit_page")
     query = Query.from(site, params, debug_metadata(conn))
     {limit, page} = parse_pagination(params)
-    metrics = breakdown_metrics(query, [:visits])
+
+    extra_metrics =
+      if TableDecider.sessions_join_events?(query) do
```
macobo, line 939: Is this check strictly needed now? :thinking: See also the comment on table_decider.

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
-    if :percentage in query.metrics do
-      total_query =
-        query
-        |> remove_filters_ignored_in_totals_query()
-        |> Query.set(
-          dimensions: [],
-          include_imported: query.include_imported,
-          pagination: nil
-        )
+  defp add_exit_rate(q, query, site) do
```
macobo, line 25: Nit: Can we move this function and the above `case` clause to the _bottom_ instead of the top?

Logic: this is the least important special metric.

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
+        include_imported: query.include_imported,
+        dimensions: ["event:page"]
+      )
 
+    joined_q =
       q
-      |> select_merge_as([], total_visitors_subquery(site, total_query, query.include_imported))
-      |> select_merge_as([], %{
-        percentage:
+      |> join(:left, [..., s], p in subquery(SQL.QueryBuilder.build(total_pageviews_query, site)),
```
macobo, line 40:
```suggestion
      |> join(:left, [], p in subquery(SQL.QueryBuilder.build(total_pageviews_query, site)),
```

`s` doesn't seem to be used below.

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
+        on:
+          selected_as(^shortname(query, "visit:exit_page")) ==
+            field(p, ^shortname(total_pageviews_query, "event:page"))
+      )
+
+    if query.include_imported do
+      joined_q
+      |> select_merge_as([_s, _i, p], %{
```
macobo, line 48: How would you feel about

```suggestion
      |> select_merge_as([..., p], %{
```

IMO slightly less fragile to other query building changes outside of this module causing unexpected breakage.

(See also below)

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
   import Plausible.Stats.Util
 
   def add(q, site, query) do
-    q
-    |> maybe_add_percentage_metric(site, query)
-    |> maybe_add_global_conversion_rate(site, query)
-    |> maybe_add_group_conversion_rate(site, query)
-    |> maybe_add_scroll_depth(query)
+    Enum.reduce(query.metrics, q, fn
```
apata, line 15: Question, non-blocking: this has the advantage of brevity, but it seems we lose the definite order these special metrics fns are called: now the order depends on Query.metrics array order. Doesn’t this introduce a new mode of failure, that we don’t currently test for?

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
+          selected_as(^shortname(query, "visit:exit_page")) ==
+            field(p, ^shortname(total_pageviews_query, "event:page"))
+      )
+
+    if query.include_imported do
+      joined_q
+      |> select_merge_as([_s, _i, p], %{
+        exit_rate:
           fragment(
             "if(? > 0, round(? / ? * 100, 1), null)",
```
macobo, line 51: Nit: Capitalize SQL NULL (both here and below)

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -12,73 +12,114 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
-        |> remove_filters_ignored_in_totals_query()
-        |> Query.set(
-          dimensions: [],
-          include_imported: query.include_imported,
-          pagination: nil
-        )
+  defp add_exit_rate(q, query, site) do
+    total_pageviews_query =
+      query
+      |> Query.remove_top_level_filters(["visit:exit_page"])
```
macobo, line 28: Q: Just so I understand the logic correctly, it would be just as correct to transform all `visit:exit_page` filters to `event:page` filters here right?

lib/plausible/stats/sql/special_metrics.ex
apata: Or are they all mutually exclusive, so the order doesn't matter?

lib/plausible/stats/sql/special_metrics.ex
macobo, line 25: Let's add a comment describing the metric above this as well similar to conversion rate, e.g. `Page exit_rate = visits_on_page / pageviews`

macobo (commented):
Great work! Some questions and comments.

test/plausible/stats/query_parser_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/5289/commits/f1135796a7ad6b81b8dd2b8934127bd2aa75804a

test/plausible/stats/query_parser_test.exs
RobertJoonas: https://github.com/plausible/analytics/pull/5289/commits/f1135796a7ad6b81b8dd2b8934127bd2aa75804a

CHANGELOG.md
RobertJoonas: Good catch! https://github.com/plausible/analytics/pull/5289/commits/7dea49d26722512e5d2520a4ee4a4feda386272f

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: Fair point. They *should* be mutually exclusive, but maintaining a clear order is definitely less error-prone. I've refactored adding the metrics via a single function and pattern matching. https://github.com/plausible/analytics/pull/5289/commits/cb8de9b87f6de9e7f264bf5e5b8c52877f363637.

lib/plausible/stats/table_decider.ex
RobertJoonas: No. Exit rate depends directly on the `visit:exit_page` dimension which is a session-only dimension. It might be possible to make it `:either` and then hack it from special metrics, but it wouldn't give a performance benefit in any case - `exit_rate` can only be queried with "visit:exit_page" dimension which already means we'll be querying the sessions table.

lib/plausible_web/controllers/api/stats_controller.ex
RobertJoonas: Yes. Still needed. If this evaluates to true, it means that querying `exit_rate` wouldn't have passed QueryParser metric validations.

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: Agreed! https://github.com/plausible/analytics/pull/5289/commits/a442c38459cc9cd9b18e94a3d5a759be097ed374

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5289/commits/e544d4f2d37c09c6c1a987ea762fa8deab319a12

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: Agreed! https://github.com/plausible/analytics/pull/5289/commits/e544d4f2d37c09c6c1a987ea762fa8deab319a12

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5289/commits/e544d4f2d37c09c6c1a987ea762fa8deab319a12

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: Yes, that's right. Currently the JOIN takes care of it.

lib/plausible/stats/sql/special_metrics.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5289/commits/f7115b06728e9393400a27844f51403253728256

lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -92,118 +94,159 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
...
+          fragment(
+            "if(? > 0, round(? / ? * 100, 1), NULL)",
+            p.pageviews,
+            selected_as(:__internal_visits),
+            p.pageviews
+          )
+      })
+    else
+      joined_q
+      |> select_merge_as([_s, p], %{
```
macobo, line 241:
```suggestion
      |> select_merge_as([..., p], %{
```

macobo (approved):
One suggestion, other than that lgtm. Great work!

lib/plausible/stats/sql/special_metrics.ex
apata: Wow, that's nice!

PR #5295, opened 2025-04-08
Allow marking custom events as non-interactive

macobo:
- Custom events can now be marked as non-interactive in events API and tracker script. Events marked as non-interactive are not counted towards bounce rate.
  - To track custom event in tracker, do `plausible('custom event', { interactive: false })`
  - In events API, set `interactive` parameter to false.
- A session is now marked as a bounce if it has less than 2 pageviews and no interactive custom events.

Note that custom events configured via css can not be marked as non-interactive as they require the user to interact with the site.

Basecamp ref: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/8450017341

Docs: https://github.com/plausible/docs/pull/601

### DIFF
--- a/test/support/test_utils.ex
+++ b/test/support/test_utils.ex
@@ -87,23 +87,6 @@ defmodule Plausible.TestUtils do
     {:ok, conn: conn}
   end
 
-  def create_pageviews(pageviews) do
-    pageviews =
-      Enum.map(pageviews, fn pageview ->
-        pageview =
-          pageview
-          |> Map.delete(:site)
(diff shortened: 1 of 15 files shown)


test/support/test_utils.ex
```diff
@@ -87,23 +87,6 @@ defmodule Plausible.TestUtils do
     {:ok, conn: conn}
   end
 
-  def create_pageviews(pageviews) do
```
ukutaht, line 90: 🙌

ukutaht (approved):

PR #5298, opened 2025-04-08
Change default period to today for brand new sites

RobertJoonas:
### Changes

Default query period to `today` rather than `28d` for brand new dashboards (i.e. created today). When a user lands on the dashboard for the first time, they should see the graph display visitors straight away.

On the next day, when the user returns to the dashboard, the period will default to `28d` *unless* they interact with the datepicker on the first day already which will store the preference in localStorage.

Basecamp ref: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8509441108

**Tricky:**

The `site.native_stats_start_at` field is the best fit for checking this. However, it gets recorded in UTC (`naive_datetime`), and therefore it's value might end up being one day before or after "today" in `site.timezone`.

When deciding whether to default to today, we are comparing two UTC dates - `utc_today` vs `site.native_stats_start_date`. The only issue that surfaces from comparing UTC dates is that the day might "end quicker". E.g. when in site timezone it's 01:00 and in UTC it's 23:00, we would consider the day to be over in an hour already, while for the customer it would be 2AM on the same day.

To overcome this issue, we will also default to today when `site.native_stats_start_date == utc_yesterday`. That means sometimes we will default to `today` for longer than just during the first 24 hours, but it shouldn't be an issue.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/test-utils/app-context-providers.tsx
+++ b/assets/test-utils/app-context-providers.tsx
@@ -22,6 +22,29 @@ type TestContextProvidersProps = {
   preloaded?: { segments?: SavedSegments }
 }
 
+export const defaultSite: PlausibleSite = {
+  domain: 'plausible.io/unit',
+  offset: 0,
+  hasGoals: false,
+  hasProps: false,
+  funnelsAvailable: false,
+  propsAvailable: false,
+  siteSegmentsAvailable: false,
+  conversionsOptedOut: false,
+  funnelsOptedOut: false,
+  propsOptedOut: false,
+  revenueGoals: [],
+  funnels: [],
+  statsBegin: '',
+  nativeStatsBegin: '',
+  embedded: false,
+  background: '',
+  isDbip: false,
+  flags: {},
+  validIntervalsByPeriod: {},
+  shared: false
+}
+
 export const TestContextProviders = ({
   children,
   routerProps,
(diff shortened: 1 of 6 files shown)


apata:
@RobertJoonas Since you mentioned this is somewhat tricky, what if we just redirect the user from site verification flow to the dashboard with ?period=today applied? IIRC this "saves" it as the default as well, for that client (browser).

RobertJoonas:
Good idea @apata! That would indeed be the simplest fix in code. However, opening the dashboard directly in another device or browser (either by the person themselves or a team member), the confusion would still be there so I don't think it would be a sufficient solution.

assets/test-utils/app-context-providers.tsx
```diff
@@ -22,36 +22,36 @@ type TestContextProvidersProps = {
   preloaded?: { segments?: SavedSegments }
 }
 
+export const defaultSite: PlausibleSite = {
```
macobo, line 25: Nitpick: constants generally use uppercase.

macobo (approved):

assets/test-utils/app-context-providers.tsx
RobertJoonas: https://github.com/plausible/analytics/pull/5298/commits/d09f1d1e3ac28f4dd5849c9c6b8b8802efdf8356

PR #5304, opened 2025-04-09
Switch from `Site.locked` to `Team.locked`

zoldar:
### Changes

Move site lock state from individual sites to team.

The changes will be deployed in stages:

- ~first, changes up on updating `team.locked` in parallel with `site.locked`~
- ~next, backfill migration~
- after backfill migration is run and team.locked state confirmed to be consistent, switch to `team.locked` entirely

### TODO

- [x] double check if there are cases of sites with mixed locked state in the same team in the wild
- [x] add backfill
- [x] rename `SiteLocker.update_sites_for`
- [x] see if `SiteLocker` uses can be reduced
- [x] split into multiple PRs, starting from migration

### Tests

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -76,7 +76,7 @@ defmodule Plausible.MixProject do
       {:cors_plug, "~> 3.0"},
       {:credo, "~> 1.5", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
-      {:double, "~> 0.8.0", only: [:test, :ce_test]},
+      {:double, "~> 0.8.0", only: [:dev, :test, :ce_test]},
       {:ecto, "~> 3.12.0"},
       {:ecto_sql, "~> 3.12.0"},
       {:envy, "~> 1.1.1"},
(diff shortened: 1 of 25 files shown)


mix.exs
```diff
@@ -76,7 +76,7 @@ defmodule Plausible.MixProject do
       {:cors_plug, "~> 3.0"},
       {:credo, "~> 1.5", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
-      {:double, "~> 0.8.0", only: [:test, :ce_test]},
+      {:double, "~> 0.8.0", only: [:dev, :test, :ce_test]},
```
aerosol, line 79: necessary?

zoldar: Yes, otherwise `MigrationUtils` fails to compile when running `mix dialyzer`. A case similar to `bypass` I think.

aerosol (approved):

zoldar:
`sites.locked` and `teams.locked` went out of sync after a couple days. Putting this on hold until we figure out why.

PR #5305, opened 2025-04-09
GA4 import: refresh google auth along the way so token doesn't expire during >1hr imports

ewdurbin:
### Changes

During long imports, the `access_token` provided at the start may expire.

Regularly check the token expiration during import and refresh if necessary.

### Tests

### Changelog

### Documentation

### Dark mode


ewdurbin:
Hm, nope this isn't working. I think the correct approach is to add a new case here: https://github.com/plausible/analytics/blob/087b6e77bc8a4ebfb816ab5008a62f1cfab24ffa/lib/plausible/imported/google_analytics4.ex#L100

PR #5327, opened 2025-04-14
Make BackfillTeams completely schemaless

zoldar:
### Changes

As team backfill is now run in (CE-only) migration, its use of Ecto schemas causes issues every time the used schemas change. This PR converts the backfill procedure to use schemaless queries exclusively. The underlying columns and tables must remain for the migration to work but it's not dependent on Ecto-level schemas anymore.


aerosol (approved):
:rocket: awesome

PR #5331, opened 2025-04-15
Change redirects from external to internal where possible

zoldar:
### Changes

We've been using redirect(external: ...) for all paths containing domain because in the past, Phoenix's redirect was overzealous when sanitizing the URL, effectively rejecting URL-encoded UTF8 characters (starting with %). Current version of phoenix has that sanitization relaxed and relying on external redirects is no longer necessary.

### DIFF
--- a/lib/plausible_web/live/verification.ex
+++ b/lib/plausible_web/live/verification.ex
@@ -185,8 +185,8 @@ defmodule PlausibleWeb.Live.Verification do
   end
 
   defp redirect_to_stats(socket) do
-    stats_url = Routes.stats_url(PlausibleWeb.Endpoint, :stats, socket.assigns.domain, [])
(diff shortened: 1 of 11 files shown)


lib/plausible_web/live/verification.ex
```diff
@@ -185,8 +185,8 @@ defmodule PlausibleWeb.Live.Verification do
   end
 
   defp redirect_to_stats(socket) do
-    stats_url = Routes.stats_url(PlausibleWeb.Endpoint, :stats, socket.assigns.domain, [])
```
zoldar, line 188: Was there a particular reason for using _url instead of _path helper here? 🤔 \cc @aerosol

macobo (approved):

PR #5334, opened 2025-04-15
Stats API: Remove business logic from query building

macobo:
### Changes

Small improvement on https://github.com/plausible/analytics/pull/5230 - remove fragile business logic from query building by always doing a full join.


macobo:
This runs head-first into a bunch of import edge cases: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8550177228

Pretty sure this approach right now is a source of bugs but there isn't a clean way to fix this immediately. Given these are quite tricky to solve, closing for now.

PR #5337, opened 2025-04-16
Move invitations related logic under Teams context

zoldar:
### Changes

This PR is basically moving modules and files around so that all invitation related logic which was still under `Site.Memberships` is now under `Teams` context.


aerosol (approved):
Feels a bit inconsistent with function naming still, the `_invitation` suffix looks random? But it's gtg.

zoldar:
@aerosol I don't want to change everything at once. Contents of the modules will be cleaned up in follow-ups later on.

PR #5338, opened 2025-04-16
Change session transfer duration unit

aerosol:
I think we should follow the others, right? @ruslandoga 

Not sure if this warrants a new metric name just in case?


ruslandoga (approved):

lib/plausible/session/transfer.ex
```diff
@@ -18,7 +18,7 @@ defmodule Plausible.Session.Transfer do
   @cmd_dump_cache :get
   @cmd_takeover_done :done
 
-  def telemetry_event, do: [:plausible, :sessions, :transfer]
```
aerosol, line 21: ah, this made the session version change. sigh.

PR #5344, opened 2025-04-16
Only log a pageview for visible pages

### LINKED ISSUE #5343: Do not log pageview for pages that are not viewed (Speculation Rules prerender and new tabs)
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

Currently Plausible tracks pageviews for pages that are potentially not seen, including:

- Prerendered pages. There is support for the older `visibilityState'='prerendered' but not for the new [Speculation Rules API](https://developer.chrome.com/docs/web-platform/prerender-pages).
- Pages opened in new tabs but never visited

These may never be seen by real users, but will log a pageview.

While the second one (new tab) is questionable (it's at least triggered by a user action), the first is becoming increasingly common and is triggered by in the following circumstances:
- When using the URL bar for visiting frequently visited pages in Chrome (and also Safari - but it doesn't use Speculation Rules)
- When the site uses Speculation Rules via manual addition or a plugin
- When the site uses Cloudflare or Wordpress both of which now support this by default (though in prefetch mode, and not prerender... yet).

### Expected behavior

Pageviews are only triggered when pages are viewed.

This would be an easy change to change this line:

https://github.com/plausible/analytics/blob/cb465511bb20dfaf0cfa92343748ad4adea4ec45/tracker/src/plausible.js#L344-L348

to this:

```js
 if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender') { 
```

This would cover both prerendered pages (which are started in hidden stage initially) and also new tab pages.

Note however that this could drop the number of page views for a site if they have a lot of these unseen page views. Or if the analytics code only fired after a page was seen and then hidden, and it was not returned too.

We could handle just the prerendered case by looking at [the `prerenderingchange` event](https://developer.chrome.com/docs/web-platform/prerender-pages#impact-on-analytics) if preferred, but I think it's good to solve both at the same time.

### Screenshots

_No response_

### Environment

```markdown
- OS: All
- Browser: Chrome
- Browser Version: 122 or above
```

tunetheweb:
### Changes

Fixes #5343

### Tests

I couldn't find tests for existing older prerendered check, which this just extends. So not sure if

### Changelog

### Documentation

Can add/update docs if this change is accepted

### Dark mode

### DIFF
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -341,7 +341,7 @@
       }
     }
 
-    if (document.visibilityState === 'prerender') {
+    if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender') {
       document.addEventListener('visibilitychange', handleVisibilityChange);
     } else {
       page()
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -341,7 +341,7 @@
       }
     }
 
-    if (['hidden', 'prerender'].includes(document.visibilityState)) {
+    if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender') {
       document.addEventListener('visibilitychange', handleVisibilityChange);
     } else {
       page()
--- a/tracker/src/plausible.js
+++ b/tracker/src/plausible.js
@@ -341,7 +341,7 @@
       }
     }
 
-    if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender') {
+    if (['hidden', 'prerender'].includes(document.visibilityState)) {
       document.addEventListener('visibilitychange', handleVisibilityChange);
     } else {
       page()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - Make clicking Compare / Disable Comparison in period picker menu close the menu
+- Do not log page views for hidden pages (prerendered pages and new tabs), until pages are viewed
 
 ## v3.0.0 - 2025-04-11
 
--- a/tracker/package.json
+++ b/tracker/package.json
@@ -1,5 +1,5 @@
 {
-  "tracker_script_version": 3,
+  "tracker_script_version": 4,
   "scripts": {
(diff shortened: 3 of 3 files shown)


ukutaht:
Great, thank you @tunetheweb! Will merge once build is green

tunetheweb:
Looks like I'm missing a token - perhaps cause this is a PR from a fork rather than the main repo? If so might be easier to close this and re-open a PR yourselves since it's a small change?

tracker/src/plausible.js
```diff
@@ -341,7 +341,7 @@
       }
     }
 
-    if (document.visibilityState === 'prerender') {
+    if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender') {
```
apata, line 344: I noticed a tiny improvement that would save a few bytes in the compiled script: 
```suggestion
    if (['hidden', 'prerender'].includes(document.visibilityState)) {
```

tracker/src/plausible.js
```diff
@@ -341,7 +341,7 @@
       }
     }
 
-    if (document.visibilityState === 'prerender') {
+    if (['hidden', 'prerender'].includes(document.visibilityState)) {
```
macobo, line 344: Thanks for the contribution.

Could we revert back to the previous implementation. The suggestion by @apata breaks IE11 compat mode. https://caniuse.com/?search=array.includes

tracker/src/plausible.js
macobo: https://github.com/plausible/analytics/pull/5344#discussion_r2048708553 Sorry - we should not use Array.includes in tracker.

tracker/src/plausible.js
macobo:
```suggestion
    if (document.visibilityState === 'hidden' || document.visibilityState === 'prerender')
```

macobo (approved):

macobo:
Taking over and will get this merged.

tracker/src/plausible.js
apata: Good catch!

PR #5349, opened 2025-04-17
Staging subscriptions

RobertJoonas:
### Changes

Make staging plans consistent with prod and streamline that process for the future as well (via a Mix.Task).

### Tests

### Changelog

### Documentation

### Dark mode


lib/mix/tasks/create_paddle_sandbox_plans.ex
```diff
@@ -0,0 +1,203 @@
...
+  end
+
+  defp create_paddle_plan(%{name: name, price: price, type: type, interval_index: interval_index}) do
+    your_unique_token = "abc"
+
+    # Replace this curl command. You might be able to reuse
+    # the request body after replacing your unique token.
+    curl_command = """
+    ... REPLACE ME
+    --data-raw '_token=#{your_unique_token}&plan-id=&default-curr=USD&tmpicon=false&name=#{name}&checkout_custom_message=&taxable_type=standard&interval=#{interval_index}&period=1&type=#{type}&trial_length=&price_USD=#{price}&active_EUR=on&price_EUR=#{price}&active_GBP=on&price_GBP=#{price}'
```
zoldar, line 91: Hmm, maybe we could simplify that part a bit? I've managed to cut down the cURL expression to send only cookie, content-type and raw data and it worked. We could make the cuRL expression predefined here and grab cookie and token from env vars via SANDBOX_PADDLE_COOKIE and SANDBOX_PADDLE_TOKEN (instructing to copy them from the POST request) - WDYT?

RobertJoonas: Yeah, I just figured that since this is not an actual API, it's fragile and not guaranteed to continue working with the same curl request. We're always gonna have to do it in the browser anyway to copy the cookies, so this felt like the most KISS approach.

zoldar (approved):

PR #5357, opened 2025-04-25
Switch to the team on accepting team invitation

zoldar:
### Changes

Previously user remained in their current team after accepting team invitation. Now they are switched to the new team instead.

### Tests


aerosol (approved):
LGTM

PR #5358, opened 2025-04-25
Accept after register via team invite

zoldar:
### Changes

This PR simplifies registration flow when registering from team invitation. Now, the invitation is accepted immediately after registering and the user is switched to the new team at once.

In case of email verification flow only "happy path" with verification code already issued is concerned with implicit team invite approval and redirect to team sites.

### TODO

- [x] Add Auth controller test for login action via register from team invite

### Tests

### DIFF
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -94,12 +94,14 @@ defmodule PlausibleWeb.AuthController do
   def activate_form(conn, params) do
     user = conn.assigns.current_user
     flow = params["flow"] || PlausibleWeb.Flows.register()
+    team_identifier = params["team_identifier"]
 
     render(conn, "activate.html",
       error: nil,
       has_email_code?: Plausible.Users.has_email_code?(user),
       has_any_memberships?: Plausible.Teams.Users.has_sites?(user),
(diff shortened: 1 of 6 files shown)


lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -110,15 +112,21 @@ defmodule PlausibleWeb.AuthController do
...
       :ok ->
         cond do
+          team_identifier not in ["", nil] ->
+            redirect_path = accept_team_invitation(conn, team_identifier, user, flow: flow)
+            redirect(conn, to: redirect_path)
+
           has_any_memberships? ->
             handle_email_updated(conn)
 
           has_any_invitations? ->
```
aerosol, line 127: Can't this clause be joined with the first one?

```
team_identifier not in ["", nil] or has_any_invitations? -> 
```

aerosol (approved):
lgtm

lib/plausible_web/controllers/auth_controller.ex
zoldar: However unlikely, there might be a case where invitation is already accepted at this stage (double-submit for instance). That's why I've put the team identifier clause before membership check.

PR #5362, opened 2025-04-29
Reload dashboard when Stats API version changes

macobo:
Idea from the lovely @apata for making easier breaking changes in the dashboard API: reload the dashboard when the internal API version changes.

This will help us make 'breaking' changes easier.

### DIFF
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -5,6 +5,9 @@ import { serializeApiFilters } from './util/filters'
 let abortController = new AbortController()
 let SHARED_LINK_AUTH: null | string = null
 
+// And keep in sync with lib/plausible_web/controllers/api/stats_controller.ex
+const EXPECTED_API_VERSION = 0
+
 export class ApiError extends Error {
   payload: unknown
   constructor(message: string, payload: unknown) {
@@ -85,6 +88,17 @@ async function handleApiResponse(response: Response) {
   const payload = await response.json()
   if (!response.ok) {
     throw new ApiError(payload.error, payload)
+  }
+
+  if (
+    payload?.api_version !== undefined &&
+    payload?.api_version !== EXPECTED_API_VERSION
+  ) {
+    console.log('Received unexpected version from API. Reloading page...', {
+      payload,
+      EXPECTED_API_VERSION
+    })
+    window.location.reload()
   }
 
   return payload
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -15,6 +15,12 @@ defmodule PlausibleWeb.Api.StatsController do
 
   plug(:date_validation_plug)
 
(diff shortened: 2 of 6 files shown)


assets/js/dashboard/api.ts
```diff
@@ -87,6 +90,17 @@ async function handleApiResponse(response: Response) {
...
 
+  if (
+    payload?.api_version !== undefined &&
+    payload?.api_version !== EXPECTED_API_VERSION
+  ) {
+    console.log('Received unexpected version from API. Reloading page...', {
+      payload,
+      EXPECTED_API_VERSION
+    })
+    window.location.reload()
```
ukutaht, line 101: Say the client has cached a javascript bundle with `EXPECTED_API_VERSION=5`. We deploy new code where it's incremented to 6 but the old javascript stays cached in the users' browser. The cached javascript will expect version 5 but gets version 6. Won't this result in a reload loop? Does window.reload clear browser cache?

ukutaht: Oh my bad, the javascript file will have a new filename which deals with both cdn and browser caches. So basically we're dealing here with the case where someone has an active session on a dashboard with the old JS file as we deploy changes, right?

macobo: > So basically we're dealing here with the case where someone has an active session on a dashboard with the old JS file as we deploy changes, right?

Exactly. As we deploy breaking stats api changes specifically.

ukutaht (approved):

assets/js/dashboard/api.ts
apata, line 101: In my mind, this risks us creating an endless reload loop for all dashboards when we fail to keep things in sync. There's a small chance of that happening, but the damage would be great.

I think we could mitigate it by instead doing `window.location.replace`, adding the `r=...` query parameter with value of EXPECTED_API_VERSION, like `r=5`.

Then, in this function, we can make the reload conditional on query not containing `r=${EXPECTED_API_VERSION}`. 

There's a decent example of something similar in `url-search-params.ts`: perhaps this logic could be colocated there?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -115,7 +121,8 @@ defmodule PlausibleWeb.Api.StatsController do
         comparison_plot: comparison_result && plot_timeseries(comparison_result, metric),
         comparison_labels: comparison_result && label_timeseries(comparison_result, nil),
         present_index: present_index,
-        full_intervals: full_intervals
+        full_intervals: full_intervals,
+        api_version: @api_version
```
apata, line 125: Having it in the JSON payload is good: it catches the eye when refactoring the payloads server side. But, it also misses a few routes that the dashboard is using, like sites, segments, funnels. 

Did you consider providing it in a header? 

It seems to me that if it were a header, we could provide it with a router plug on all those APIs and use it as the definitive trigger for reloading incompatible dashboards.

PR #5363, opened 2025-04-29
ScriptV2: Improved tracker compile.js

macobo:
### Changes

This PR reworks how tracker script compilation works in preparation for the Script V2 work.

**New features**
- Running playwright tests no longer requires running compile.js between changes - the script is dynamically compiled.
- Sped up compilation by utilizing multi-core compilation
- CLI options for compile script - --watch mode, --limit and being able to only compile a specific script variant and `--suffix` to power data analysis
- Compilation + tests now use esm modules
- Progress bar running compile.js to stave off boredom

**Other notes**
- No more handlebars.js for compilation
- variants are now defined in a file (with a generator script if ever needed). We'll add new variants there manually.

**What's left for the future**
- We'll only compile a subset of variants that are needed for local dev
- `plausible-main` and `plausible-npm` variants
- config object used in script.
- Linting and formatter usage in plausible.js

**What about script size?**

Script size increased by at most 0.7% (compressed) due to removal of handlebars and how it acted with inlining certain functions. Doing 2 passes with uglify-js would reduce that to 0.4% but would worsen dev experience.


tracker/compiler/index.js
```diff
@@ -0,0 +1,123 @@
...
+  })
+
+  if (result.code) {
+    return result.code
+  } else {
+    throw result.error
+  }
+}
+
+function equal_lists(a, b) {
```
ukutaht, line 109: verytiny nit: use snakeCase is javascript

ukutaht (approved):

tracker/compiler/index.js
macobo: :man_facepalming:

tracker/compiler/generate-variants.js
```diff
@@ -0,0 +1,21 @@
...
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
+const base_variants = ["hash", "outbound-links", "exclusions", "compat", "local", "manual", "file-downloads", "pageview-props", "tagged-events", "revenue"]
+let legacyVariants = [...g.clone.powerSet(base_variants)]
+  .map(a => a.sort())
+  .map((variant) => ({
+    name: variant.length > 0 ? `plausible.${variant.join('.')}.js` : 'plausible.js',
+    features: variant,
+    globals: Object.fromEntries(variant.map(feature => [featureToCompileKey(feature), true]))
+  }))
```
apata, line 16: I think it's worth thinking about what's constant going forward, `globals` or enabled `features`, and persist just one of these in the variants file. 

Over time, as we work on the script, how these features are enabled in the script will likely change, so I'd say it's enabled features. We can leave it up to the compiling code to derive the correct configuration for enabling the features for that variant.

tracker/compiler/generate-variants.js
```diff
@@ -0,0 +1,21 @@
+import fs from 'fs'
+import path from 'path'
+import { fileURLToPath } from 'url'
+import g from 'generatorics'
+import { featureToCompileKey } from './index.js'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
+const base_variants = ["hash", "outbound-links", "exclusions", "compat", "local", "manual", "file-downloads", "pageview-props", "tagged-events", "revenue"]
```
apata, line 9: I'd like to see these highlighted in the repository, e.g. imported from a `/tracker/constants.js` file. These are really important for current script compilation and Plausible install logic, and we'll need to remain quite mindful of these until all customers have migrated to the next version.

macobo: I don't see the benefit.

It adds another layer of indirection, making the code harder to understand. We have tests in test/plausible_web/plugs/tracker_test.exs covering that the legacy variants are still being served, so any change there needs to go through a review that would explicitly drop tests and code here.

tracker/compiler/generate-variants.js
macobo: Reasons for `generate-variants.js` separating these two:

1. to separate business logic from compilation logic as much as possible
2. make it easily greppable what `COMPILE_FILE_DOWNLOADS` means even without understanding how the compile script works. 

While they're currently a 1:1 mapping `features` and `globals` will serve a different use-case going forward, with `features` being about limiting what script variants are to be built in compile.js during dev flows. So you can compile e.g. `plausible-main` easily as part of `mix phx.server` without compiling >1000 other variants.

tracker/compiler/index.js
```diff
@@ -0,0 +1,123 @@
...
+import { fileURLToPath } from 'url'
+import variantsFile from './variants.json' with { type: 'json' }
+import { canSkipCompile } from './can-skip-compile.js'
+import packageJson from '../package.json' with { type: 'json' }
+import progress from 'cli-progress'
+import { spawn, Worker, Pool } from "threads"
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
+const DEFAULT_GLOBALS = {
```
apata, line 13: There's a bit of duplication / overlap between this and the `featureToCompileKey()` function, but there's no "link" between them, for lack of a better term, which leaves us room for error when moving forward. Could we be more explicit about how boolean feature name (aka "base variant") maps to a global injected at compile time. Supposing we highlight these feature names as I suggest in https://github.com/plausible/analytics/pull/5363/files#r2071246985, I think we should also map them explicitly to the global variables that govern them, in this vein, but could also be simpler:
```js
// tracker/constants.js
const FEATURES = {hash: 'hash'}
const GLOBALS_BY_FEATURE: {
  [FEATURES.hash]: {'COMPILE_HASH': true}, 
  //...
}
```

That way 
1) variants.json file maps script variant name to enabled feature list 
2) the feature list of the variant is mapped to globals that need to be injected in compile script
3) the injected globals result in the appropriate script being built

This PR is very close to this approach, but it's not so clearly connected.

tracker/compiler/generate-variants.js
apata: Good that we have the legacy variants list tested! 

It's just that this particular script, as I understood it, is almost single use, so to me it's not obvious that is the place to look for the base variant names (from which ultimately script names and even global names are derived).

Even if we don't move this, the least we could do is make it uppercase, as most of the other constants.

tracker/compiler/index.js
macobo: In my WIP branch adding `plausible-main` variant I've moved `featureToCompileKey` to be generate-variants specific so this feedback will be tackled in a follow-up.

I don't agree with tying features with globals as mentioned in an above comment - it will complicate rather than simplify work planned.

tracker/compiler/generate-variants.js
apata: To me, storing some of the global values in this file yet still injecting others in the compile script seems weird. The compiler is fully under our control: we can define in the compiler how to arrive at a particular feature combination or other variant. 

It seems to me to support the single, larger main script variant, we could configure it as `{"name": "plausible-main.js", "compiledFeatures": "all"}` and handle the "all" flag in the compiler, injecting all the needed globals in one place. 

Is there some benefit when doing `grep COMPILE_MANUAL` and getting 100s of matches in variants.json, as opposed to a couple of matches where it's linked to feature name `manual` in the compiler script / in constants file?

tracker/compiler/generate-variants.js
macobo: https://github.com/plausible/analytics/pull/5363/commits/14dadb23279e903e92ea1e9f45742450f9a5f13e for uppercasing.

tracker/compiler/generate-variants.js
macobo: I disagree. It seems like we're going in circles on this minute point so let's have a call to unblock this.

Let's discuss how to proceed with this PR on monday's sync.

tracker/compiler/index.js
macobo: https://github.com/plausible/analytics/pull/5363/commits/878d68b903549b2ce8554f4aab9ae6455a3a5d1f - backported the fix onto this branch

tracker/compiler/generate-variants.js
macobo: https://github.com/plausible/analytics/pull/5363/commits/878d68b903549b2ce8554f4aab9ae6455a3a5d1f for renaming features to `compileIds`

PR #5375, opened 2025-05-02
ScriptV2: plausible-main, script size measurements

macobo:
PR depends on https://github.com/plausible/analytics/pull/5363

This PR:
- **1. Creates a new `plausible-main.js` variant.** While not yet served anywhere, this will serve as the "main" script once the project completes. The new variant also is covered with tests.
- **2. Script / CI action that will report on script size changes in a github comment.** Some of the tricks used in `plausible.js` to ensure legacy variants don't increase in size are non-obvious and might be easily broken by subsequent refactors. To avoid this, tracker changes now also report on how tracker script size is changing compared to version on master. The script is also designed so it can be used without going through CI during development.

Not tackled in this PR:
- App-side changes to start serving plausible-main.js

Notes for reviewers:
- The conditional in plausible.js looks quite ugly. Couldn't make uglify output a small script otherwise
- Exact shape of configuration _will_ change in subsequent PRs as we nail down the onboarding flow.


macobo:
Analyzed 1025 tracker script variants for size changes.
The following tables summarize the results, with comparison with the baseline version in parentheses.

Main variants:
|                                 | Brotli | Gzip  | Uncompressed |
| ------------------------------- | ------ | ----- | ------------ |
| plausible-main.js (new variant) | 2180B  | 2494B | 5637B        |

Important legacy variants:
|                                                                                                   | Brotli              | Gzip                | Uncompressed    |
| ------------------------------------------------------------------------------------------------- | ------------------- | ------------------- | --------------- |
| plausible.js                                                                                      | 1093B (+1B / +0.1%) | 1275B (-3B / -0.2%) | 2680B (0B / 0%) |
| plausible.hash.js                                                                                 | 1056B (+2B / +0.2%) | 1236B (-3B / -0.2%) | 2566B (0B / 0%) |
| plausible.pageview-props.tagged-events.js                                                         | 1790B (+1B / +0.1%) | 2092B (-3B / -0.1%) | 4554B (0B / 0%) |
| plausible.file-downloads.hash.pageview-props.revenue.js                                           | 1607B (0B / 0%)     | 1901B (-3B / -0.2%) | 3922B (0B / 0%) |
| plausible.compat.exclusions.file-downloads.outbound-links.pageview-props.revenue.tagged-events.js | 2205B (0B / 0%)     | 2604B (-4B / -0.2%) | 5698B (0B / 0%) |

Summary:
|                                                                                                                     | Brotli               | Gzip                | Uncompressed    |
| ------------------------------------------------------------------------------------------------------------------- | -------------------- | ------------------- | --------------- |
| Largest variant (plausible.compat.exclusions.file-downloads.outbound-links.pageview-props.revenue.tagged-events.js) | 2205B (0B / 0%)      | 2604B (-4B / -0.2%) | 5698B (0B / 0%) |
| Max change (plausible.hash.pageview-props.js)                                                                       | 1129B (+8B / +0.7%)  | 1336B (-4B / -0.3%) | 2749B (0B / 0%) |
| Min change (plausible.file-downloads.outbound-links.pageview-props.tagged-events.js)                                | 2026B (-18B / -0.9%) | 2372B (-4B / -0.2%) | 5118B (0B / 0%) |
| Median change                                                                                                       | 1656B (0B / 0%)      | 1963B (-4B / -0.2%) | 4116B (0B / 0%) |

.github/workflows/tracker-script-update.yml
```diff
@@ -8,27 +8,31 @@ on:
...
-          ref: ${{ github.event.pull_request.head.ref }}
-          token: ${{ secrets.PLAUSIBLE_BOT_GITHUB_TOKEN }}
-          fetch-depth: 0
 
       - name: Checkout master for comparison
         uses: actions/checkout@v4
         with:
-          ref: master
+          # :TODO: Update this to use the master branch once the dependent PR is merged
+          ref: improve-compile
```
macobo, line 24: - [] TODO: Remove this once https://github.com/plausible/analytics/pull/5363 is merged.

.github/workflows/tracker-script-update.yml
```diff
@@ -8,27 +8,31 @@ on:
...
-        run: sudo apt-get install jq -y
+      - name: Install jq and clickhouse-local
+        run: |
+          sudo apt-get install apt-transport-https ca-certificates dirmngr
+          sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 8919F6BD2B48D754
+          echo "deb https://packages.clickhouse.com/deb stable main" | sudo tee \
+              /etc/apt/sources.list.d/clickhouse.list
+          sudo apt-get update
+
+          sudo apt-get install jq clickhouse-server -y
```
apata, line 35: I'm slightly worried about these extra dependencies in the runner container. Can we run the size comparison action in a dedicated docker image that receives just the output from the compile step?

tracker/test/plausible-main.spec.js
```diff
@@ -0,0 +1,169 @@
...
+      ]
+    })
+  })
+
+  test('tracks file downloads (when feature enabled)', async ({ page }) => {
+    await openPage(page, { fileDownloads: true })
+
+    await expectPlausibleInAction(page, {
+      action: () => page.click('#file-download'),
+      expectedRequests: [{ n: 'File Download', p: { url: 'https://awesome.website.com/file.pdf' } }],
```
apata, line 95: If this URL has this file, will this file be actually downloaded? If yes, we might want to mock it with playwright

tracker/test/support/server.js
```diff
@@ -21,7 +21,11 @@ export function runLocalFileServer() {
     const name = req.params.name
     const variant = VARIANTS.find((variant) => variant.name === name)
 
-    const code = compileFile(variant, { returnCode: true })
+    let code = compileFile(variant, { returnCode: true })
```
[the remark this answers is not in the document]
macobo: It isn't the bottleneck on either CI or dev - the browser interactions are. As such optimizing here would be a bad idea.

.github/workflows/tracker-script-update.yml
```diff
@@ -73,3 +77,31 @@ jobs:
...
+        run: |
+          cd tracker
+          OUT=$(node compiler/analyze-sizes.js --baselineSuffix master --currentSuffix pr)
+          # Set multiline output
+          echo "sizes<<EOF" >> $GITHUB_OUTPUT
+          echo "$OUT" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+      - name: Comment script size report on PR
+        uses: thollander/actions-comment-pull-request@v3
```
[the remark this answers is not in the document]
macobo: https://github.com/plausible/analytics/pull/5375/commits/181bac158a23e17f862b0217af4e32596bd6751c

tracker/test/plausible-main.spec.js
macobo: https://github.com/plausible/analytics/pull/5375/commits/e519ee822ed4f03c1801d1d045d4cb75ca295f14

.github/workflows/tracker-script-update.yml
macobo: Can you be more specific what is the concern and of the proposed solution? The action itself seems to perform well enough as-is: https://github.com/plausible/analytics/actions/runs/14900847787/job/41852251784?pr=5375

image

apata: Ah, sorry for vagueness. The concern is increased toolchain vulnerabilities. Perhaps @cnkk can weigh in on whether this concern is valid?

The proposed solution is to declare it as a https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-docker-container-action. 

Another effect of having it dockerized would be that it could be run locally without installing the dependencies on the host machine.

.github/workflows/tracker-script-update.yml
```diff
@@ -8,27 +8,30 @@ on:
       - 'tracker/package-lock.json'
 
 jobs:
-  tracker-increment-reported-version:
+  tracker-script-update:
     runs-on: ubuntu-latest
     permissions:
+      pull-requests: write
       contents: write
```
cnkk, line 15: I guess this can be updated to `contents: read` as the step committing back to the repo uses a custom manually generated token

cnkk, line 15: I guess this can be updated to `contents: read` as the step committing back to the repo uses a custom manually generated token

.github/workflows/tracker-script-update.yml
cnkk: > Ah, sorry for vagueness. The concern is increased toolchain vulnerabilities. Perhaps @cnkk can weigh in on whether this concern is valid?

I cannot follow for this specific case. Sorry. I don't see a decrease in security when running this in a dedicated container.

> Another effect of having it dockerized would be that it could be run locally without installing the dependencies on the host machine.

That's not a big problem as the environment is fresh for each run. But could be an advantage speed wise

apata: Ah, I thought that there are GH related secrets on the runner, and when installing applications directly in the runner these would be exposed to the applications. 

If the GH runner built a docker image with these applications and ran the container, the applications in the container would know nothing more of the host (the GH runner) than they need to. 

It’s also useful locally not to need to install these dependencies directly: with GH runner we are in control of what secrets are exposed in the runner, but on local development machines much more could be compromised.

apata (approved):

PR #5377, opened 2025-05-05
rate limit with atomics

ruslandoga:
might help with rate limiter getting overloaded


aerosol:
Thank you! Will cherry-pick into a collective changeset

PR #5381, opened 2025-05-06
Track form submission events

apata:
### Changes

- Adds config option to track form submission events to tracking script.
- In tests, adds utility to create the page HTML on load, so we'd not need so many fixtures.

### TODO
- [x] Understand why tests are flaky 
  - Issue 1: test runner submits the form before tracker script has fully initialized
  - Issue 2: Playwright Firefox version is less than 133 which implemented fetch with keepalive.

- [x] Fix flakiness in tests

- [x] Figure out if form submission tracking is too eager, catching forms that haven't passed custom validation
  - Fixed with checking `checkValidity` but also respecting `novalidate` attribute on the form element. Note: we don't respect the `formnovalidate` that can be set on the submit input: another limitation.  

- [x] Figure out what to do to catch forms that are submitted with [FormElement.submit()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit), which doesn't go through `onsubmit` event handlers
  - Decided to highlight it as a limitation in tests. The workaround for users will be to use `FormElement.requestSubmit()` or to manually track such submits.

- [x] Figure out if and how to support forms that are not present on script load (a'la form that is inserted into the DOM when user expands a specific page section)
  - Fixed by listening at document level.

### TODO in future PRs

- Figure out if and how to distinguish forms on websites that are using hash routing

- Figure out exact event configuration: what event name to use and what props to include

- Allow setting the option in onboarding, autocreate and autodelete related goal

### Tests

### Changelog
- [x] Entry will be added together with main documentation update

### Documentation
- [x] This change does not need a documentation update: will be released together with new script installation method

### Dark mode
- [x] No UI changes


tracker/src/plausible.js
```diff
@@ -364,6 +364,20 @@
...
+    if (e.target.hasAttribute('novalidate') || e.target.checkValidity()) {
+      plausible('WP Form Completions', { props: { path: location.pathname } });
+    }
+  }
+
+  function attachFormSubmitTracking() {
+    document.addEventListener('submit', trackFormSubmission, true);
+  }
+
+  attachFormSubmitTracking();
```
macobo, line 378:
```suggestion
  document.addEventListener('submit', trackFormSubmission, true);
```

Unneeded indirection.

tracker/src/plausible.js
```diff
@@ -364,6 +364,20 @@
     })
   }
 
+if (COMPILE_CONFIG && config.trackFormSubmissions) {
+  function trackFormSubmission(e) {
+    if (e.target.hasAttribute('novalidate') || e.target.checkValidity()) {
+      plausible('WP Form Completions', { props: { path: location.pathname } });
```
macobo, line 370: The event name is incorrect. Don't worry about wordpress in the current stage, we'll figure out a way to deal with their legacy events/goals later.

```suggestion
      plausible('Form submission', { props: { path: location.pathname } });
```

tracker/test/support/server.js
```diff
@@ -16,6 +16,29 @@ export const LOCAL_SERVER_ADDR = `http://localhost:${LOCAL_SERVER_PORT}`
 
 export function runLocalFileServer() {
   app.use(express.static(FIXTURES_PATH));
+  app.get('/dynamic/*', (req, res) => {
```
macobo, line 19: Please remove this and make into an html template as other tests for consistency. 

If the path needs to be dynamic, a `<script>` tag can read it and append it.

tracker/test/support/initialize-page-dynamically.ts
```diff
@@ -0,0 +1,34 @@
+import { Page } from '@playwright/test'
+import { ScriptConfig } from './types'
+
+interface DynamicPageOptions {
+  scriptConfig: ScriptConfig
+  /** vanilla HTML string, which can contain JS, will be set in the body of the page */
+  bodyContent: string
+}
+
+export function initializePageDynamically(
```
macobo, line 10: This function creates yet another way data is passed to the page. Could we get by creating another version of tracker/test/fixtures/plausible-main.html and dynamically set bodyContent based on a query parameter in an already-set div?

This would keep the approach more consistent across all tests and would remove code from both server.js and yet another helper.

tracker/test/support/test-utils.js
```diff
@@ -231,3 +233,13 @@ function checkEqual(a, b) {
 function delay(ms) {
   return new Promise(resolve => setTimeout(resolve, ms))
 }
+
+export async function openPage(page, config, options = {}) {
```
macobo, line 237: This 'helper' is still plausible-main specific. Let's move it back.

tracker/test/main/form-submissions.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+          n: 'WP Form Completions',
+          p: { path: `/dynamic/${testId}` }
+        }
+      ]
+    })
+  })
+
+  test('limitation: tracks _all_ forms on the same page, but _records them indistinguishably_', async ({
+    page
+  }, { testId }) => {
```
macobo, line 217: Why do we need `testId` in these tests? Looking at the test it is just noise, but I might be missing something.

tracker/test/main/form-submissions.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+      expectedRequests: [
+        {
+          n: 'WP Form Completions',
+          p: { path: `/dynamic/${testId}` }
+        }
+      ]
+    })
+  })
+
+  test('tracks form submissions triggered with submit button', async ({
```
macobo, line 132: Nitpick, not blocking: Move this test first in the describe. Happy path first over 'limitations'

tracker/test/support/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    local: boolean
+    exclusions: boolean
+    manual: boolean
+    revenue: boolean
+    pageviewProps: boolean
+    outboundLinks: boolean
+    fileDownloads: boolean
+    taggedEvents: boolean
+    trackFormSubmissions: boolean
+  }
```
macobo, line 12: Nit: Formatting issues in this file.

tracker/test/main/form-submissions.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+  })
+
+  test('does not track form submissions that do not pass checkValidity', async ({
+    page
+  }, { testId }) => {
+    await initializePageDynamically(page, {
+      scriptConfig: { ...DEFAULT_CONFIG, trackFormSubmissions: true },
+      bodyContent: `
+      <div>
+        <form onsubmit="event.preventDefault(); console.log('Form submitted')">
```
macobo, line 78: Q: Why do we do the event.preventDefault here and in other tests? Given that it's in the test code I assumed it's important for the correctness of the test, but it doesn't seem so.

tracker/test/main/form-submissions.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        </form>
+        <button id="trigger" onclick="document.getElementById('form').submit()">Submit</button>
+      </div>
+      `
+    })
+
+    await expectPlausibleInAction(page, {
+      action: async () => {
+        await page.goto(`/dynamic/${testId}`)
+        await page.click('button#trigger')
```
macobo, line 60:
```suggestion
        await page.click('#submit')
```

Nit: terminology

tracker/test/support/initialize-page-dynamically.ts
apata: The background here is that I found existing tests hard to work on without a direct link between the HTML and actual test. I couldn't easily tell which parts of the fixture were for which test case. 

I guess it wasn't clear that I am proposing this (and the accompanying dynamic route) as an iterative improvement over all existing test setups, including plausible-main and openPage. 

As it's a proposal and WIP, I didn't yet refactor all other tests, nor give it _all_ the features to be able to unify all tests. Consistency is not yet in place, but would be if we go forward with this. 

The benefits of using this is that it brings almost all of the test setup inside the test. I think the diversity in form related tests proves its usefulness. 

A downside is that my IDE doesn't help with this dynamic HTML. I believe it's possible to get most IDEs and [prettier](https://prettier.io/docs/options.html#embedded-language-formatting) to work with it though. 
```js
const content = /* html */`
  <form>...</form>
  `
```

Good call to investigate the way we deliver bodyContent and the configured script. I didn't try passing the DOM over query param yet, but I don't see why it shouldn't work. I tried header first actually, and when that failed, landed on the Pagewright API that worked well enough for the proposal.

apata: Tried passing this over query params and I got it to render the UI. However, the default form submit action replaces query params with fields from the form, which then results in the express server rendering an empty page in response... The Playwright API that I'm using doesn't have that issue. The Playwright API seems good enough to me, pretty direct as well.

tracker/test/main/form-submissions.spec.ts
apata: Do you mean that I should apply this suggestion AND also change the `id` of the button? I am specifically using an `id` that makes it distinguishable from submit clicks in other tests.

tracker/test/main/form-submissions.spec.ts
apata: Valid question! I agree that the test should contain only that which is strictly necessary. This is to eliminate page reload from form submit actions as the cause of test flakiness.

macobo: I think that makes sense, though you still seem to struggle with flakiness. Could we remove it from the inlined html here and move it to the template page? 

It being here makes the reader of the test think it's somehow related to the scenario under test (instead of testing harness).

tracker/test/support/initialize-page-dynamically.ts
macobo: I see, this makes sense. I didn't consider that form submissions navigate :facepalm:

I think integrating with the express server dynamically is quite tricky :sweat_smile: 

Regarding `setContent` idea, perhaps creating a mock route would be simpler route?

E.g something like this.

```
const HTML_TEMPLATE = fs.readFileSync(...)
const ROUTE = "/formSubmissions"
async function openFormPage(page, { path, bodyContent, headContent, config }) {
  const body = HTML_TEMPLATE.replace(...).replace(...)
  page.route(ROUTE, (route) => route.fulfill({ status: 200, contentType: 'text/html', body }))

  await page.goto(ROUTE)
}
```

(I'm not sure whether path is behaviorally important, so made it constant here)

tracker/test/support/server.js
macobo: Resolving this, thread under https://github.com/plausible/analytics/pull/5381/files#r2079004714

tracker/test/main/form-submissions.spec.ts
macobo: Ah, makes sense. In that case:

Nit: perhaps `#javascript-submit` would be a better id?

tracker/test/support/initialize-page-dynamically.ts
ukutaht: Love the idea of increasing cohesion in tests by inlining the HTML ❤️ I also find it hard to work with fixtures especially when serve multiple different tests/assertions.

tracker/src/plausible.js
apata: Went with 'Form Submission' because it's more inline with 'File Download' etc, exact name TBD.

tracker/test/support/initialize-page-dynamically.ts
apata: We decided that `initializePageDynamically` utility will stay in and will be shared between tests that test features in depth. 

I refactored this dynamic HTML not to go through our test server. Instead it's declared as a [Playwright route](https://playwright.dev/docs/api/class-page#page-route). I didn't realize it at first, but `route()` API actually has a few gotchas. AFAICT they don't apply to our situation yet, so it's mergable, just a bit annoying to create potential debt.

It overlaps a bit with `openPage` helper in `plausible-main.spec.js`, but that helper and its accompanying fixture are specifically built to test script initialization, config overrides, and feature toggling, which require much more cleverness than `initializePageDynamically`. There's a general solution that covers both cases, but decided not to seek it yet.

tracker/test/main/form-submissions.spec.ts
apata: Went with `trigger-FormElement-submit`

tracker/test/form-submissions.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+      scriptConfig: { ...DEFAULT_CONFIG, formSubmissions: true },
+      bodyContent: `
+      <div>
+        <form method="GET" ${
+          // this conditional is needed because fetch with keepalive is not implemented in the version of Firefox used by Playwright:
+          // can be removed once the Firefox version is >= v133
+          browserName === "firefox"
+            ? 'onsubmit="event.preventDefault(); setTimeout(() => {this.submit()}, 200)"'
+            : ""
+        }>
```
macobo, line 74: There's a better syntax for skipping tests: https://playwright.dev/docs/test-annotations#conditionally-skip-a-test

```js
test('skip this test', async ({ page, browserName }) => {
  test.skip(browserName === 'firefox', 'Still working on it');
});
```

tracker/test/support/dynamic-page-template.html
```diff
@@ -0,0 +1,29 @@
...
+        }),
+        (window.plausible.init = function (i) {
+          window.plausible.o = i || {};
+        });
+      var script = document.createElement("script");
+      (script.type = "text/javascript"),
+        (script.defer = !0),
+        (script.src = "<%= plausible_script_url %>");
+      var r = document.getElementsByTagName("script")[0];
+      r.parentNode.insertBefore(script, r);
```
macobo, line 23: non-blocking q: Why is this beautified? It'd be better to take this snippet:

```
└> node compile.js --web-snippet

<script>
  window.plausible=window.plausible||function(){(window.plausible.q=window.plausible.q||[]).push(arguments)},window.plausible.init=function(i){window.plausible.o=i||{}};var script=document.createElement("script");script.type="text/javascript",script.defer=!0,script.src="<%= plausible_script_url %>";var r=document.getElementsByTagName("script")[0];r.parentNode.insertBefore(script,r);
  plausible.init()
</script>
```

tracker/test/form-submissions.spec.ts
```diff
@@ -0,0 +1,306 @@
+import { test, Page } from "@playwright/test";
```
macobo, line 1: Nit: Let's remove the semicolons for consistency, most specs don't use them

tracker/test/form-submissions.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  test("limitation: tracks _all_ forms on the same page, but _records them indistinguishably_", async ({
+    page,
+  }, { testId }) => {
+    const { url } = await initializePageDynamically(page, {
+      testId,
+      scriptConfig: { ...DEFAULT_CONFIG, formSubmissions: true },
+      bodyContent: `
+      <div>
+        <form onsubmit="event.preventDefault(); console.log('Form 1 submitted')">
```
macobo, line 265: Q: Is the event.preventDefault needed?

tracker/test/form-submissions.spec.ts
apata: Well spotted, but are semicolon related comments really worth our time and the CI resource use? 

I don't think so, which is why I prepared and shared prettier and linter configurations for the tracker two times in the last 4 weeks, once at the very start of this endeavor. Let's apply them and regain our focus on what matters.

https://github.com/plausible/analytics/compare/script-v2/add-linter-prettier?expand=1

tracker/test/support/dynamic-page-template.html
apata: Ah, I did take that exactly snippet, but when I autoformat the HTML template, it adds whitespace to the script, which makes it look like the snippet source. I didn't see much point in fighting with that, it's just whitespace. The bigger improvement would be to make an actual hard link between the compiled snippet and dynamic page snippet.

apata: I've hard-linked the two in  https://github.com/plausible/analytics/pull/5381/commits/7622b23f315eca2c2eaad96e8c39519ac791808b

It's not exactly fool-proof in my opinion and I could roll the commit back, but at least it takes the manual diffing out.

tracker/test/form-submissions.spec.ts
macobo: Currently this test is two different tests: one on firefox, one elsewhere.

Given that the expected behavior is broken on our playwright firefox version, I think skipping is a better option and communicates intent better. Both options can also be easily cleaned up in a future PR.

tracker/test/form-submissions.spec.ts
macobo: Let's skip the test on firefox instead then and remove the skip once we've upgraded firefox.

apata: What's the benefit of skipping them? 

They run on Firefox and other browsers, and prove that the feature works for forms where the onsubmit handler is custom, which is a commonplace in the web apps that I've worked on. 

Did you disagree with this take: https://github.com/plausible/analytics/pull/5381#discussion_r2100070553?

tracker/test/support/dynamic-page-template.html
macobo: Let's roll it back for now, not important and can make more consistent later if needed.

macobo:
I think this is ready to be merged after these comments are resolved:
- https://github.com/plausible/analytics/pull/5381#discussion_r2101708869
- https://github.com/plausible/analytics/pull/5381#discussion_r2101705958 and https://github.com/plausible/analytics/pull/5381#discussion_r2101710710

tracker/test/support/dynamic-page-template.html
apata: I've rolled it back, but I don't understand your point on making it more consistent later if needed. How can it be any more consistent than using the same function that  `node compile.js --web-snippet` is using?

PR #5383, opened 2025-05-07
Initial UI mock for scriptv2 onboarding

ukutaht:
### Changes

* Adds a new `scriptv2` feature flag
* When `scriptv2` feature flag is enabled, we redirect user from `/:domain/installation` to `/:domain/installationv2` where new installation screen lives for the time being (to be revisited when ready to launch)
* New installation screen separates 'optional measurements' and 'advanced options'. Advanced options is collapsed by default.
* Currently only for manual installation method, will look at switching between installation methods (including new NPM method) in a follow-up PR
* No concern for different flows at the moment, currently only for provisioning flow. Not sure what will happen to "domain changed" flow and "review" flow on this screen
* Missing tests at the moment because the main point is to have something tangible in the UI but it's very much subject to change during the course of the project.
* Doesn't actually do anything - the script config is not saved anywhere.
* Currently the whole thing is very unfinished but the point is having a UI mock to make progress on the project. Needs many more touches before ready to be used for real

It's nowhere near production-ready and needs lots of extra work. Whether to keep this on a branch or keep merging the WIP is up for discussion.

https://github.com/user-attachments/assets/1af8e2ff-2e21-47dc-8e24-3ed17f93383d


macobo (approved):
You asked for a quick review - this seems like a decent starting point!

PR #5389, opened 2025-05-12
ScriptV2: plausible-main application code

macobo:
Follow-up to https://github.com/plausible/analytics/pull/5375, this PR adds support for interpolating script config into plausible-main template and serving it under `/js/s-<id>.js`.

Notes for the reviewer
- We are currently using site.installation_meta['id'], an auto-generated column for the id. We'll separately switch over to a more friendly, shorter id format in an subsequent PR.
- I'm planning for the `PlausibleWeb.Tracker` module to also contain the caching/CDN cache-busting logic down the line
- `site.installation_meta` has subtly different key names from script config. Ideally we'd fix this, but this requires a whole migration cycle, so delaying this.

### DIFF
--- a/lib/plausible_web/plugs/tracker_plug.ex
+++ b/lib/plausible_web/plugs/tracker_plug.ex
@@ -75,6 +75,8 @@ defmodule PlausibleWeb.TrackerPlug do
       |> halt()
     else
       conn
+      |> send_resp(404, "Not found")
+      |> halt()
     end
   end
 
--- a/lib/plausible_web/plugs/tracker_plug.ex
+++ b/lib/plausible_web/plugs/tracker_plug.ex
@@ -1,4 +1,4 @@
-defmodule PlausibleWeb.Tracker do
+defmodule PlausibleWeb.TrackerPlug do
   import Plug.Conn
   use Agent
 
--- a/lib/plausible_web/plugs/tracker_plug.ex
+++ b/lib/plausible_web/plugs/tracker_plug.ex
@@ -1,5 +1,6 @@
 defmodule PlausibleWeb.TrackerPlug do
   import Plug.Conn
+  import Ecto.Query
   use Agent
 
   base_variants = [
@@ -33,18 +34,46 @@ defmodule PlausibleWeb.TrackerPlug do
   end
 
   def call(conn, files_available: files_available) do
-    filename =
-      case conn.request_path do
-        "/js/p.js" ->
-          "p.js"
+    case conn.request_path do
+      "/js/s-" <> path ->
+        if String.ends_with?(path, ".js") do
+          tag = String.replace_trailing(path, ".js", "")
+          request_tracker_script(tag, conn)
+        else
+          conn
+        end
 
-        "/js/" <> requested_filename ->
-          sorted_script_variant(requested_filename)
+      "/js/p.js" ->
+        legacy_request_file("p.js", files_available, conn)
 
-        _ ->
-          nil
-      end
+      "/js/" <> requested_filename ->
+        sorted_script_variant(requested_filename) |> legacy_request_file(files_available, conn)
 
+      _ ->
+        conn
+    end
+  end
+
+  defp request_tracker_script(tag, conn) do
+    site = Plausible.Repo.one(from s in Plausible.Site, where: s.installation_meta["id"] == ^tag)
+
+    if site do
+      script_tag = PlausibleWeb.Tracker.plausible_main_script_tag(site)
+
+      conn
+      |> put_resp_header("content-type", "application/javascript")
+      |> put_resp_header("x-content-type-options", "nosniff")
+      |> put_resp_header("cross-origin-resource-policy", "cross-origin")
+      |> put_resp_header("access-control-allow-origin", "*")
+      |> put_resp_header("cache-control", "public, max-age=60, no-transform")
+      |> send_resp(200, script_tag)
+      |> halt()
+    else
+      conn
+    end
+  end
+
+  defp legacy_request_file(filename, files_available, conn) do
     if filename && MapSet.member?(files_available, filename) do
       location = Application.app_dir(:plausible, "priv/tracker/js/" <> filename)
 
--- a/lib/plausible_web/plugs/tracker_plug.ex
+++ b/lib/plausible_web/plugs/tracker_plug.ex
@@ -18,7 +18,7 @@ defmodule PlausibleWeb.TrackerPlug do
   ]
 
   # Generates Power Set of all variants
-  variants =
+  legacy_variants =
     1..Enum.count(base_variants)
     |> Enum.map(fn x ->
       Combination.combine(base_variants, x)
@@ -26,8 +26,9 @@ defmodule PlausibleWeb.TrackerPlug do
     end)
     |> List.flatten()
 
-  @base_filenames ["plausible", "script", "analytics"]
-  @files_available ["plausible.js", "p.js"] ++ Enum.map(variants, fn v -> "plausible.#{v}.js" end)
+  @base_legacy_filenames ["plausible", "script", "analytics"]
+  @files_available ["plausible.js", "p.js"] ++
+                     Enum.map(legacy_variants, fn v -> "plausible.#{v}.js" end)
 
   def init(opts) do
     Keyword.merge(opts, files_available: MapSet.new(@files_available))
@@ -95,7 +96,7 @@ defmodule PlausibleWeb.TrackerPlug do
 
   defp sorted_script_variant(requested_filename) do
     case String.split(requested_filename, ".") do
-      [base_filename | rest] when base_filename in @base_filenames ->
+      [base_filename | rest] when base_filename in @base_legacy_filenames ->
         sorted_variants =
           rest
           |> Enum.reject(&(&1 in @ignore_variants))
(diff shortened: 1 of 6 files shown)


lib/plausible_web/tracker.ex
```diff
@@ -0,0 +1,43 @@
...
+        # Not enabled values can be omitted
+        {_key, false} -> nil
+      end)
+      |> Enum.reject(&(&1 == nil))
+      |> Enum.sort_by(&String.length/1, :desc)
+      |> Enum.join(",")
+
+    Application.app_dir(:plausible, "priv/tracker/js/plausible-main.js")
+    |> File.read!()
+    |> String.replace("\"<%= @config_json %>\"", "{#{config_js_content}}")
```
ukutaht, line 23: Using the [EEx](https://hexdocs.pm/eex/EEx.htm) has some benefits here:
1. The source file contents are compiled into the release so the source file does not need to be read at runtime
2. EEX uses [iodata](https://hexdocs.pm/elixir/main/io-and-the-file-system.html#iodata-and-chardata) under the hood which is more efficient than running String operations

lib/plausible_web/plugs/tracker_plug.ex
```diff
@@ -0,0 +1,117 @@
...
+      "/js/" <> requested_filename ->
+        sorted_script_variant(requested_filename) |> legacy_request_file(files_available, conn)
+
+      _ ->
+        conn
+    end
+  end
+
+  defp request_tracker_script(tag, conn) do
+    site = Plausible.Repo.one(from s in Plausible.Site, where: s.installation_meta["id"] == ^tag)
```
ukutaht, line 63: Eventually this lookup should be indexed (doesn't need to be done in this PR). I also would like to have a proper schema for `installation_meta` instead of jsonb. Proposing as a future task to migrate `installation_meta` to its own postgres table with the following goals:
1. Btree index on `id` (btree is not available for jsonb fields AFAIK)
2. More constrained schema
3. Align naming between installation_meta and JS compile variables as mentioned in PR description
4. Installation meta is currently an embedded schema on sites which makes Elixir-side CRUD work more complex than if it were a separate schema

lib/plausible_web/tracker.ex
```diff
@@ -0,0 +1,43 @@
...
+      |> plausible_main_config()
+      |> Enum.map(fn
+        {key, value} when is_binary(value) -> "#{key}:#{Jason.encode!(value)}"
+        # :TRICKY: Save bytes by using short-hand for true
+        {key, true} -> "#{key}:!0"
+        # Not enabled values can be omitted
+        {_key, false} -> nil
+      end)
+      |> Enum.reject(&(&1 == nil))
+      |> Enum.sort_by(&String.length/1, :desc)
```
ukutaht, line 18: Q: What's the purpose of sorting by string length?

lib/plausible_web/tracker.ex
```diff
@@ -0,0 +1,43 @@
...
+  @moduledoc """
+  Helper module for building the dynamic tracker script. Used by PlausibleWeb.TrackerPlug.
+  """
+
+  def plausible_main_script_tag(site) do
+    config_js_content =
+      site
+      |> plausible_main_config()
+      |> Enum.map(fn
+        {key, value} when is_binary(value) -> "#{key}:#{Jason.encode!(value)}"
```
ukutaht, line 11: Prefer the recently added [JSON](https://hexdocs.pm/json/readme.html) module in standard library. `Jason` dependency can eventually be removed thanks to the standard library addition.

lib/plausible_web/tracker.ex
```diff
@@ -0,0 +1,43 @@
...
+        # :TRICKY: Save bytes by using short-hand for true
+        {key, true} -> "#{key}:!0"
+        # Not enabled values can be omitted
+        {_key, false} -> nil
+      end)
+      |> Enum.reject(&(&1 == nil))
+      |> Enum.sort_by(&String.length/1, :desc)
+      |> Enum.join(",")
```
ukutaht, line 19: It's possible that some of this hand-crafting can be removed by leveraging the `encoder` function argument to [JSON.encode!](https://hexdocs.pm/elixir/JSON.html#encode!/2). Even if it's 100% correct the idea of joining these lines by `,` and wrapping in `{}` manually feels error-prone. I'm thinking something like:

```elixir
site
|> plausible_main_config()
|> remove_false_values
|> JSON.encode(fn 
  {key, true} -> "#{key}:!0"
  _ -> # fallback to default behaviour somehow
)
```

ukutaht: Also if [JSON.encode_to_iodata](https://hexdocs.pm/elixir/JSON.html#encode_to_iodata!/2) works the same then it's probably better here.

test/plausible_web/plugs/tracker_plug_test.exs
```diff
@@ -0,0 +1,124 @@
...
+        new_site()
+        |> Plausible.Sites.update_installation_meta!(%{script_config: @example_config})
+
+      response = get(conn, "/js/s-#{site.installation_meta.id}.js") |> response(200)
+
+      assert String.contains?(response, "!function(){var")
+      assert String.contains?(response, "domain:\"#{site.domain}\"")
+      assert String.contains?(response, "hash:!0")
+      assert String.contains?(response, "taggedEvents:!0")
+      assert not String.contains?(response, "outboundLinks:!0")
```
ukutaht, line 42:
```suggestion
      refute String.contains?(response, "outboundLinks:!0")
```

test/plausible_web/plugs/tracker_plug_test.exs
```diff
@@ -0,0 +1,124 @@
...
+
+    test "returns the right script extensions no matter the order" do
+      response = get_script("plausible.compat.file-downloads.hash.outbound-links.js")
+
+      assert String.contains?(response, "getElementById(\"plausible\")")
+      assert String.contains?(response, "file-types")
+      assert String.contains?(response, "hashchange")
+      assert String.contains?(response, "Outbound Link: Click")
+
+      assert !String.contains?(response, "data-exclude")
```
ukutaht, line 83:
```suggestion
      refute String.contains?(response, "data-exclude")
```

lib/plausible_web/tracker.ex
macobo: > It's possible that some of this hand-crafting can be removed by leveraging the encoder function argument to [JSON.encode!](https://hexdocs.pm/elixir/JSON.html#encode!/2)

Note that the goal of this function isn't valid json but a valid js object to minimize size.

As such this would increase the resulting script size. [The json encoder](https://github.com/elixir-lang/elixir/blob/v1.18.3/lib/elixir/lib/json.ex#L404) can only encode values and not keys. Keys would end up automatically quoted.

lib/plausible_web/tracker.ex
macobo: Consistent and predictable output for tests in test/plausible_web/tracker_test.exs

lib/plausible_web/tracker.ex
macobo: I've updated config_json to config_js in the next PR to avoid bumping the tracker script version in this one: https://github.com/plausible/analytics/pull/5391/commits/9b1bed1bda7ef1e95af13e04e11abfd521b2a14f

lib/plausible_web/tracker.ex
macobo: 1. File contents are now read at compile-time instead at runtime: https://github.com/plausible/analytics/pull/5389/commits/e4a5db694da92c2d6019b01b23f57dd960d489b2, https://github.com/plausible/analytics/pull/5389/commits/bc3302dc0adf6575474c4e87442af639d7111397
2. I couldn't get EEX to work due to limitations on `plausible.js` side. Namely `config = <%= config_js %>` isn't valid javascript. 

On 2) We could keep config_js quoted, but then we should interpolate json rather than javascript code and do a JSON.parse on the other end. This would increase bundle size which doesn't seem like the greatest trade-off.

I did a microbenchmark and it seems like EEX is actually slower?

```
EEx: 250873µs
String.replace: 8883µs
```

Given this, I'd stick with String.replace, though open to suggestions.

<details><summary>Click to see microbenchmark code</summary>

```
path = "../priv/tracker/js/plausible-main.js"

eex_code = EEx.compile_file(path)
raw_code = File.read!(path)

config_js = ~s({endpoint:"plausible.io/api/event",domain:"plausible.io",taggedEvents:!0,hash:!0})

range = 1..1000

{time_eex, _} = :timer.tc(fn -> Enum.map(range, fn _ -> Code.eval_quoted(eex_code, assigns: %{ config_js: config_js }) end) end)
{time_raw, _} = :timer.tc(fn -> Enum.map(range, fn _ -> String.replace(raw_code, "<%= @config_js %>", config_js) end) end)

IO.puts("EEx: #{time_eex}µs")
IO.puts("Raw: #{time_raw}µs")
```
</details>

lib/plausible_web/plugs/tracker_plug.ex
macobo: This makes sense to me, however I'll tackle this separately as part of creating the plugins API as there's tricky migration work involved which I don't want to get bogged down in yet.

lib/plausible_web/tracker.ex
ukutaht: > Note that the goal of this function isn't valid json but a valid js object to minimize size. (Though I now realize I misnamed the string variable)

Oh I see, I missed that it's actual JS code not a JSON object. Makes sense!

lib/plausible_web/plugs/tracker_plug.ex
ukutaht: No it's not blocking, I was proposing this as a future task (but still as part of this project)

ukutaht (approved):

PR #5391, opened 2025-05-14
ScriptV2: `plausible.init`, `customProperties`, snippet compilation, double initialization protection, `swc`

macobo:
### Changes

Depends on https://github.com/plausible/analytics/pull/5389

This PR adds the following to the script:
1. For `plausible-main`, plausible.init() needs to be called with appropriate configuration values. No behavioral changes to legacy script, but this somewhat increases script size.
2. `plausible-main` no longer has configuration `pageviewProps: true` and instead supports a `customProperties` option in init of type `Record<string, string> | (event: string) => Record<string, string>`. Note that explicitly passed props to `plausible('event', {props})` override these.
3. `node compile.js --web-snippet` now outputs the minified web snippet that can be used in onboarding
4. Script is explicitly protected and tested against cases where `<script />` is included twice or `init()` is called twice.
5. Script compilation is switched from `uglify-js` to `@swc/core`

Along the way:
- New tests were added for (legacy) pageview props
- plausible.js was partially reformatted as code moved around
- does 2 small refactors to reduce size (https://github.com/plausible/analytics/pull/5391/commits/00b9c6e06be837e27e41879faa57efba7b314f6e, https://github.com/plausible/analytics/pull/5391/commits/6cbdecb4168c40ebccdeb9a4ccb995ba9ef222a5)

### SWC switch
**Why** PR ran into a bug in uglify-js (https://github.com/mishoo/UglifyJS/issues/5967). 
**What** With swc and 4 passes, the compilation now finishes in 2.5s instead of 30 on my machine. However the, output is slightly larger than uglify-js, generally adding ~5-10B to brotli output in comparison.

Note this could have been a separate PR but this demonstrates size gain from this better than a PR on master.

------

For reviewers: Due to the whitespace changes and code moving around, changes to plausible.js should be easier to follow commit-by-commit.


tracker/src/plausible.js
```diff
@@ -1,300 +1,159 @@
...
-    }
+  }
 
-    if (COMPILE_HASH && (!COMPILE_CONFIG || config.hash)) {
-    payload.h = 1
-    }
+  // Track custom properties for pageviews and other events
+  // Note that engagement events track custom properties differently, using `currentEngagementProps`
+  if (COMPILE_CUSTOM_PROPERTIES && config.customProperties && eventName !== 'engagement') {
+    var props = (typeof config.customProperties === 'object') ? config.customProperties : config.customProperties(eventName)
```
ukutaht, line 118: On the following lines we would error out in many cases like:

```js
customProperties: ["foo", "bar"]
```

```js
customProperties: 1
```

```js
customProperties: function() { "randomness" }
```

etc.

If the client makes a mistake like this I think we should still be able to record the event.

tracker/src/plausible.js
```diff
@@ -303,45 +162,216 @@
...
+  }
+}
+
+function init(overrides) {
+  if (COMPILE_CONFIG && window.plausible && window.plausible.l) {
+    console.warn('Plausible analytics script was already initialized, skipping init')
+    return
+  }
+
+  if (COMPILE_CONFIG && overrides) {
```
ukutaht, line 310: Did you consider

```javascript
Object.assign(config, overrides)
```

?

tracker/src/plausible.js
```diff
@@ -356,243 +386,266 @@
...
-    }
+    var MIDDLE_MOUSE_BUTTON = 1
 
-    if (COMPILE_FILE_DOWNLOADS && (!COMPILE_CONFIG || config.fileDownloads)) {
-    if (isDownloadToTrack(hrefWithoutQuery)) {
-      return sendLinkClickEvent(event, link, { name: 'File Download', props: { url: hrefWithoutQuery } })
-    }
-    }
-  }
+    function handleLinkClickEvent(event) {
```
ukutaht, line 420: Code organization: the `init()` function is pretty messy. I understand it just does everything we used to do at the top level but putting it into a function like this highlights the problems with structure.

The diff here is already quite big. I'm proposing as a followup PR to improve code organization in this file. Ideally the init function would only read the config, set up global variables and event listeners. Function definitions like `handleLinkClickEvent` should be extracted out of `init()` IMO.

tracker/test/fixtures/plausible-main.html
```diff
@@ -15,26 +15,43 @@
...
+
+      if (!params.has('skipPlausibleInit')) {
+        plausible.init()
+      }
+
+      // Load the script with the passed config
+      const config = params.get('script_config')
+      const script = document.createElement('script')
+      script.src = `/tracker/js/plausible-main.js?script_config=${encodeURIComponent(config)}`
+      script.setAttribute('event-some-prop', '456')
```
ukutaht, line 48: This looks like leftover from old fixtures

```suggestion
```

tracker/src/plausible.js
ukutaht: Support should be there since it's a ES6 feature. Using `fetch` means we're using promises which is ES6 as well.

tracker/test/fixtures/plausible-main.html
macobo: https://github.com/plausible/analytics/pull/5391/commits/bebc99be2115eda25af027103326a71e3e12d7bf

tracker/src/plausible.js
```diff
@@ -1,300 +1,159 @@
...
     payload.p = props
-    }
+  }
 
-    if (COMPILE_HASH && (!COMPILE_CONFIG || config.hash)) {
-    payload.h = 1
-    }
+  // Track custom properties for pageviews and other events
+  // Note that engagement events track custom properties differently, using `currentEngagementProps`
+  if (COMPILE_CUSTOM_PROPERTIES && config.customProperties && eventName !== 'engagement') {
```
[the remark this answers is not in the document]
macobo: Bit of a :man_facepalming: moment. Fixed in https://github.com/plausible/analytics/pull/5391/commits/04d06d46cc550ae1b86697c877fb63de98f66571

tracker/src/plausible.js
macobo: https://github.com/plausible/analytics/pull/5391/commits/e8db45b012212c7aa8e24aff981e0f60f80636cd

tracker/src/plausible.js
macobo: Thanks, this makes things better. https://github.com/plausible/analytics/pull/5391/commits/22fb702964fea5f98ec8c5166ebfff2c7b4b278b 


I read caniuse wrong and got lost in that we don't want `domain` to be overridable in here but using Object.assign works well regardless. Added some tests as well for the implicit behavior.

tracker/src/plausible.js
macobo: Agreed, will do so. It's also better to do this in a separate PR as I'm slightly paranoid about if the minifier would correctly remove dead functions or not. The script size analysis comment should make this obvious if not.

ukutaht (approved):

PR #5394, opened 2025-05-14
Starter tier: Upgrade page remodelling

RobertJoonas:
### Feature flag

This PR introduces a new upgrade page under a `starter_tier` feature flag. To make the separation cleaner, all existing upgrade page logic will continue to live in modules/functions with a `legacy` prefix (all of which can safely be deleted once the feature flag is live):

* `choose_plan.ex` -> `legacy_choose_plan.ex`
* `plan_box.ex` -> `legacy_plan_box.ex`
* `plan_benefits.ex` -> `legacy_plan_benefits.ex`
* `choose_plan_test.exs` -> `legacy_choose_plan_test.exs`
* `Plausible.Billing.Quota.suggest_tier` -> `Plausible.Billing.Quota.legacy_suggest_tier`

### Changes

Adds Starter plans on the upgrade page along with some other changes:

* Removes header (Plausible Logo and account dropdown) - replaces it with a simple "<- back to settings" link
* Yearly price is now selected by default (unless the user is already on a monthly plan)
* Yearly plans now also display the monthly price and monthly discount.
* moves the "current cycle / last 30 days usage" and the "what happens if I go over my pageview limit?" paragraph into a FAQ style accordion menu

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/billing_controller.ex
```diff
@@ -19,10 +19,19 @@ defmodule PlausibleWeb.BillingController do
   def choose_plan(conn, _params) do
     team = conn.assigns.current_team
 
+    {live_module, hide_header?} =
```
ukutaht, line 22: This route is currently rendered as a 'dead' controller route which renders a liveview within it. It's useful for this case where we need to render a different liveview based on feature flag but there are other ways of doing it. At the end of this project when `Legacy` liveviews can be deleted, I think it would make sense to refactor this route to render the `ChoosePlan` liveview directly from the router.

lib/plausible_web/live/choose_plan.ex
```diff
@@ -112,21 +139,51 @@ defmodule PlausibleWeb.Live.ChoosePlan do
...
-              else: "Upgrade your account"}
-          </p>
+
+        <div class="mt-6 w-full md:flex">
+          <a
+            href={Routes.settings_path(PlausibleWeb.Endpoint, :subscription)}
+            class="hidden md:flex md:w-1/6 h-max text-indigo-600 hover:text-indigo-700 dark:text-indigo-500 dark:hover:text-indigo-600 text-sm font-bold gap-1 items-center"
+          >
+            <span>←</span>
+            <p>Back to Settings</p>
```
ukutaht, line 149: This isn't vertically aligned with the 'Traffic based plans that match your growth header'. I think it would look better if they were aligned.

Screenshot 2025-05-19 at 12 42 31

test/plausible/billing/plans_test.exs
```diff
@@ -36,7 +36,7 @@ defmodule Plausible.Billing.PlansTest do
       |> assert_generation(2)
     end
 
-    test "growth_plans_for/1 returns v4 plans for expired legacy subscriptions" do
+    test "growth_plans_for/1 returns latest plans for expired legacy subscriptions" do
```
ukutaht, line 39: I'm confused, the `test_helper` does enable the `started_tier` feature flag but these tests still assert v4 as the latest generation... why don't these `assert_generation(5)`?

RobertJoonas: Good catch! Since in tests we're enabling `starter_tier` globally, it makes sense for the default/latest generation to be 5, not 4. Changed that in: https://github.com/plausible/analytics/pull/5394/commits/6d07f8db849087c4431b131d1ac4893f58156a89.

This reminded me of another thing I wanted to refactor (apologies if this PR is getting too big already!): https://github.com/plausible/analytics/pull/5394/commits/e8d803e63ec02e73aa9acb36433ebfd2d7c32e90. In emails, we suggest only the pageview volume (e.g. 10k, 100k, etc) but not the actual plan, however the current code is recommending a specific plan which is pretty much redundant complexity. The "volume tiers" are the same for all plans, regardless of the generation. This commit starts suggesting only the pageview volume.

lib/plausible_web/controllers/billing_controller.ex
RobertJoonas: Agreed. Will make a note in the card table to come back to this when removing the flag!

ukutaht (approved):

PR #5397, opened 2025-05-15
Replace uses of `Teams.owned_sites/1` with better functions where applicable

zoldar:
### Changes

Small refactor of `Teams.owned_sites/1` and its uses.


aerosol (approved):

zoldar:
@aerosol Sure.

PR #5407, opened 2025-05-21
tracker_script_configuration: schema, double-writes (step 2)

macobo:
PR in a series:
1. `tracker_script_configuration` migration
2. `tracker_script_configuration` schema, double-writes
3. `tracker_script_configuration` backfill
4. `tracker_script_configuration`: drop writes and reads to `installation_meta`
5. Plugins API for `tracker_script_configuration`

### DIFF
--- a/lib/plausible/ecto/types/nanoid.ex
+++ b/lib/plausible/ecto/types/nanoid.ex
@@ -1,0 +1,20 @@
+defmodule Plausible.Ecto.Types.Nanoid do
+  @moduledoc """
+  Custom column type for nanoid strings
+  """
+
+  use Ecto.Type
+
+  def type(), do: :string
+
+  def cast(value) when is_binary(value), do: {:ok, value}
+  def cast(_), do: :error
+
+  def load(value), do: {:ok, value}
+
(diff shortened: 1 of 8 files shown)


lib/plausible/ecto/types/nanoid.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Plausible.Ecto.Types.Nanoid do
```
aerosol, line 1: Why is this necessary if `Ecto.UUID` exists?

aerosol: ah, just saw the details dropdown, you want it compact for aesthetics

macobo: Schema notes:
> `id` uses Nanoid library to generate ids for nice and readable ids as these will be used for serving scripts. Example: qyhkWtOWaTN0YPkhrcJgy

We'll be serving the new script from under `plausible.io/js/s-<id>.js`. UUIDs look ugly and this is prettier. :)

ukutaht (approved):

PR #5409, opened 2025-05-21
tracker_script_configuration: drop writes and reads to `installation_meta` (step 4)

macobo:
PR in a series:
1. `tracker_script_configuration` migration (https://github.com/plausible/analytics/pull/5406)
2. `tracker_script_configuration` schema, double-writes (https://github.com/plausible/analytics/pull/5407)
3. `tracker_script_configuration` backfill (https://github.com/plausible/analytics/pull/5408)
4. **`tracker_script_configuration`: drop writes and reads to `installation_meta` (this PR)**
5. Plugins API for `tracker_script_configuration` (https://github.com/plausible/analytics/pull/5410)

See sections below for more details:

<details><summary><b>On schema</b></summary>

- `id` uses Nanoid library to generate ids for nice and readable ids as these will be used for serving scripts. Example: `qyhkWtOWaTN0YPkhrcJgy`
- `installation_type` is marked as nullable since old installation flow could cause inserts where `installation_type` would not be set. This is for backwards compatibility to avoid messing with the schema too much.
- `revenue_tracking`, `tagged_events`, `pageview_props` and `track_404_pages` columns are for old schema compatibility and will not be used once new onboarding is fully live
- Old embedded schema was not removed as new installation flow (which I did not modify to avoid conflicts) relies on it.
- Backfill created a new DataMigration to make it possible to use (snapshot) of schema + app code in migration

</details>

<details><summary><b>On plugins API</b></summary>

This will be used by wordpress plugin to integrate into the app.

Currently it also creates goals as part of it's code but this will be changed as the new onboarding goes live.
</details>

### DIFF
--- a/lib/plausible/site/tracker_script_configuration.ex
+++ b/lib/plausible/site/tracker_script_configuration.ex
@@ -4,6 +4,7 @@ defmodule Plausible.Site.TrackerScriptConfiguration do
   """
   use Ecto.Schema
   import Ecto.Changeset
+  import Ecto.Query
 
   @type t() :: %__MODULE__{}
 
@@ -51,6 +52,20 @@ defmodule Plausible.Site.TrackerScriptConfiguration do
       conflict_target: [:site_id],
       returning: true
     )
+  end
+
+  def get_or_create!(site_id) do
+    existing_configuration =
+      Plausible.Repo.one(
+        from(c in Plausible.Site.TrackerScriptConfiguration, where: c.site_id == ^site_id)
+      )
+
+    if existing_configuration do
+      existing_configuration
+    else
+      {:ok, new_configuration} = upsert(%{site_id: site_id})
+      new_configuration
+    end
   end
 
   defp fields_to_update(configuration_map) do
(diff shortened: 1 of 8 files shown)


lib/plausible/site/tracker_script_configuration.ex
```diff
@@ -53,6 +54,20 @@ defmodule Plausible.Site.TrackerScriptConfiguration do
     )
   end
 
+  def get_or_create!(site_id) do
```
ukutaht, line 57: Convention: do not put `Repo` calls in schema definitions. Functions like `get_or_create!` and `upsert` belong to separate service modules.

lib/plausible_web/plugs/tracker_plug.ex
```diff
@@ -60,17 +60,23 @@ defmodule PlausibleWeb.TrackerPlug do
...
 
       conn
       |> put_resp_header("content-type", "application/javascript")
       |> put_resp_header("x-content-type-options", "nosniff")
       |> put_resp_header("cross-origin-resource-policy", "cross-origin")
       |> put_resp_header("access-control-allow-origin", "*")
       |> put_resp_header("cache-control", "public, max-age=60, no-transform")
+      # CDN-Tag is used by BunnyCDN to tag cached resources. This allows us to purge
+      # specific tracker scripts from the CDN cache.
+      |> put_resp_header("cdn-tag", "tracker_script::#{tracker_script_configuration.id}")
```
ukutaht, line 79: nice 👍

ukutaht (approved):

lib/plausible/site/tracker_script_configuration.ex
macobo: Does this also apply to `upsert` function above?

macobo: As this refactor is a bit involved - fixing things in https://github.com/plausible/analytics/pull/5410

PR #5410, opened 2025-05-21
tracker_script_configuration: plugins API (step 5)

macobo:
PR in a series:
1. `tracker_script_configuration` migration
2. `tracker_script_configuration` schema, double-writes
3. `tracker_script_configuration` backfill
4. `tracker_script_configuration`: drop writes and reads to `installation_meta`
5. Plugins API for `tracker_script_configuration`


lib/plausible_web/plugins/api/controllers/tracker_script_configuration.ex
```diff
@@ -0,0 +1,70 @@
+defmodule PlausibleWeb.Plugins.API.Controllers.TrackerScriptConfiguration do
+  @moduledoc """
+  Controller for the Shared Link resource under Plugins API
```
ukutaht, line 3:
```suggestion
  Controller for the Tracker Script Configuration resource under Plugins API
```

lib/plausible_web/plugins/api/controllers/tracker_script_configuration.ex
```diff
@@ -0,0 +1,70 @@
...
+          private: %{
+            open_api_spex: %{body_params: %{tracker_script_configuration: update_params}}
+          }
+        } = conn,
+        _params
+      ) do
+    site = conn.assigns.authorized_site
+
+    update_params =
+      update_params
```
ukutaht, line 54: Limiting to relevant params is usually done by calling `Ecto.Schema.cast` in the schema. Example: https://github.com/plausible/analytics/blob/0b7bccd6a5c138a1be01f7d5af99191bd77203ee/lib/plausible/site.ex#L90

PR #5411, opened 2025-05-21
Introduce migration and schemas for SSO

zoldar:
### Changes

Migration will be extracted and released separately after the PR is approved.


extra/lib/plausible/auth/sso/saml_config.ex
```diff
@@ -0,0 +1,32 @@
...
+  import Ecto.Changeset
+
+  @type t() :: %__MODULE__{}
+
+  @fields [:idp_signin_url, :idp_entity_id, :idp_cert_pem, :idp_metadata]
+
+  embedded_schema do
+    field :idp_signin_url, :string
+    field :idp_entity_id, :string
+    field :idp_cert_pem, :string
```
zoldar, line 24: PEM certificate validation might be extended later on to reject certificates that fail to parse using x509 lib for instance.

aerosol (approved):

PR #5421, opened 2025-05-24
Revert "Log salts hash"

ruslandoga:
Reverts plausible/analytics#5401

I think it can be merged on Tuesday / Wednesday. It would've probably logged enough salt rotations for initial debugging by then.


aerosol (approved):
:+1:

aerosol:
@ruslandoga we'll have to try again, log level on prod is warning...

PR #5390, opened 2025-05-13
Customer support

aerosol:
### Changes

This PR introduces new Customer Support interface (EE only), that aims to replace Kaffy.
Tests will be added once more feedback is collected.

### Tests
- [ ] Automated tests have been added
- [x] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [x] The UI has been tested both in dark and light mode
- [ ] This PR does not change the UI

PR #5423, opened 2025-05-26
Remove kaffy

aerosol:
We no longer need it, all remaining CRM features will be developed under `/cs` route.

Ref #5390


zoldar (approved):
🔥

PR #5424, opened 2025-05-26
Implement `SSO.provision_user/1` and extend `UserAuth.log_in_user/3`

zoldar:
### Changes

This PR adds logic for provisioning SSO user during during login, including conversion of existing user and creation of new user on the fly.

### TODO

- [ ] ~expire session according to identity.expires_at~
- [ ] ~do _not_ touch/prolong the session if it's for SSO user~

NOTE: Session expiration will be handled in a follow-up PR

### Tests

### DIFF
--- a/lib/plausible_web/user_auth.ex
+++ b/lib/plausible_web/user_auth.ex
@@ -21,7 +21,7 @@ defmodule PlausibleWeb.UserAuth do
     @type login_subject() :: Auth.User.t()
   end
 
-  @spec log_in_user(Plug.Conn.t(), Auth.User.t() | Auth.SSO.Identity.t(), String.t() | nil) ::
+  @spec log_in_user(Plug.Conn.t(), login_subject(), String.t() | nil) ::
           Plug.Conn.t()
   def log_in_user(conn, subject, redirect_path \\ nil)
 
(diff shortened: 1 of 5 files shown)


lib/plausible_web/user_auth.ex
```diff
@@ -28,6 +39,37 @@ defmodule PlausibleWeb.UserAuth do
...
+
+          conn
+          |> set_user_session(user)
+          |> set_logged_in_cookie()
+          |> Phoenix.Controller.redirect(to: redirect_to)
+
+        {:error, :integration_not_found} ->
+          conn
+          |> log_out_user()
+          |> Phoenix.Controller.redirect(to: "/")
```
zoldar, line 65: Eventually this will redirect to SSO login form with "User not found" error.

lib/plausible_web/user_auth.ex
```diff
@@ -28,6 +39,37 @@ defmodule PlausibleWeb.UserAuth do
...
+
+        {:error, :multiple_memberships, user} ->
+          log_in_user(conn, user, redirect_path)
```
zoldar, line 68: We let the user pass under the assumption that they will be blocked from accessing SSO team until they leave other teams - most likely via a dedicated plug.

PR #5432, opened 2025-05-27
Script v2/update playwright

apata:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/test/plausible-web.spec.js
+++ b/tracker/test/plausible-web.spec.js
@@ -10,7 +10,8 @@ import {
   hideAndShowCurrentTab,
   metaKey,
   mockRequest,
-  e as expecting
+  e as expecting,
+  ignoreEngagementRequests
 } from './support/test-utils'
 import { test, expect } from '@playwright/test'
 import { LOCAL_SERVER_ADDR } from './support/server'
@@ -52,7 +53,7 @@ test.describe('plausible-web.js', () => {
     })
   })
 
-  test('does not trigger any events when `local` config is disabled', async ({ page }) => {
+  test('does not trigger any events with localhost domain when `local` config is disabled', async ({ page }) => {
     await expectPlausibleInAction(page, {
       action: () => openPage(page, { captureOnLocalhost: false }),
       expectedRequests: [],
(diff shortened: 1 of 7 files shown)


tracker/test/support/mock-many-requests.ts
```diff
@@ -0,0 +1,56 @@
...
+      await delay(responseDelay)
+    }
+    await route.fulfill({
+      status: 202,
+      contentType: 'text/plain',
+      body: 'ok'
+    })
+  })
+
+  const getWaitForRequests = () =>
```
macobo, line 34: Q: Why this change from resolving the promise from within `page.route`?

tracker/test/plausible-web.spec.js
```diff
@@ -52,25 +53,47 @@ test.describe('plausible-web.js', () => {
...
+
+    test('does not track outbound links without the feature being enabled', async ({ page, browserName }) => {
+    await expectPlausibleInAction(page, {
+      action: async () => {
+        await openPage(page, {})
+        await page.click('#outbound-link')
+      },
+      expectedRequests: [
+        { n: 'pageview', p: expecting.toBeUndefined() },
+        ...(browserName === 'webkit' ? [] : [{ n: 'engagement' }]) // WebKit does not send engagement events for outbound links
```
macobo: Let's add `shouldIgnoreRequest: ignoreEngagementRequests` over this as it's not _that_ relevant for the test.

This seems like something to investigate - I'll add a task to the backlog to add and fix this problem separately.

tracker/test/plausible-web.spec.js
```diff
@@ -52,25 +53,47 @@ test.describe('plausible-web.js', () => {
     })
   })
 
-  test('does not trigger any events when `local` config is disabled', async ({ page }) => {
+  test('does not trigger any events with localhost domain when `local` config is disabled', async ({ page }) => {
```
macobo: The new wording seems off - events are triggered _from_ domains not with. 

`the does not trigger any events from localhost domain when `local` config is disabled` seems also off to me because that's applicable all tests.

I suggest we revert to the previous test name.

tracker/test/file-downloads.spec.js
```diff
@@ -44,11 +44,19 @@ test.describe('legacy file-downloads extension', () => {
   test('starts download only once', async ({ page }) => {
     await page.goto('/file-download.html')
     const downloadURL = LOCAL_SERVER_ADDR + '/' + await page.locator('#local-download').getAttribute('href')
-
-    const downloadRequestMockList = mockManyRequests({ page, path: downloadURL, numberOfRequests: 2 })
+    const timeToWaitMs = 3000
+    let requestCount = 0
+    await page.route(downloadURL, async (route) => {
```
macobo, line 49: Q: Why is the change needed here?

tracker/test/support/mock-many-requests.ts
apata: Good question! The poll does look a bit weird.

The context for refactor is that the previous implementation didn't `await` the async `page.route` API, so sometimes the action started before the route was setup. This implementation fixes that as easily as I could think of without refactoring other logic. This implementation allows us to do this exactly: 

1. setup route (wait till finished)
2. run action (wait till finished) // in the caller
3. set time limit and request count limit for requests (wait till finished)

I don't have resolve available in the route handler. To have it available, I'd need the Promise to be declared above the route handler, which I don't think I can do without breaking 1.

I can get rid of it, but only way I can think of is if I make the action happen here as well.

tracker/test/file-downloads.spec.js
apata: I refactored `mockManyRequests` `numberOfRequests` option name to `expectedRequestCount`. At the time I wasn't aware of this test case. When I realized that this test case also used that function and I updated it, the test looked even weirder than before: `mockManyRequests` is called with `"expectedRequestCount": 2` and the actual expectation is that request count is 1. 

Because the test previously wasn't all that clear, and because I'm not sold that we actually want one utility that does it all, I didn't think on it further and used Playwright and Node APIs directly.

tracker/test/plausible-web.spec.js
apata: I thought I had clarified the test for myself and others who don't know why and which events are disabled. There was no indication in the test itself or the spec that the events are being sent from localhost domain.

tracker/test/plausible-web.spec.js
apata: Can we leave it in? Then the task can reference this line of the code

macobo: I'd strongly prefer not to. Tests get confusing when they test _different_ things across different platforms.

tracker/test/file-downloads.spec.js
macobo: Makes sense, this was previously a weird one-off case as well.

tracker/test/support/mock-many-requests.ts
macobo: Ah, makes sense. Let's keep this as-is, though it might change further in the future.

PR #5434, opened 2025-05-27
Implement SSO login against fake SAML endpoints

zoldar:
### Changes

This PR implements provisional SSO login views along with (for now) fake SAML flow implementation.

- Newly provisioned SSO user is added as a member to the team (that was missed originally when implementing `SSO.provision_user/1` 🙈 )
- Account for and handle a case of logging in SSO user exceeding SSO team's team members limit
- Always switch to SSO team when logging in as SSO user
- Fix a bug where SSO user logging in via standard login does not have their `timeout_at` set (that will be needed for Owner login fallback)
- Implement SSO login controller actions necessary for SAML flow, faked by posting directly to consume endpoint instead of going via IdP for now

### Tests

### Dark mode


extra/lib/plausible/auth/sso.ex
```diff
@@ -162,13 +178,37 @@ defmodule Plausible.Auth.SSO do
...
+    result =
+      Repo.transaction(fn ->
+        with {:ok, user} <- Repo.insert(changeset),
+             :ok <- Teams.Invitations.check_team_member_limit(team, role, user.email),
+             {:ok, team_membership} <-
+               Teams.Invitations.create_team_membership(team, role, user, now) do
+          if team_membership.role != :guest do
+            {:identity, team, user}
+          else
+            Repo.rollback(:integration_not_found)
```
aerosol, line 194: Is this error accurate here? :thinking: 
Looks uncovered by tests.

lib/plausible_web.ex
```diff
@@ -66,6 +68,25 @@ defmodule PlausibleWeb do
     end
   end
 
+  on_ee do
+    def extra_view do
+      quote do
+        use Phoenix.View,
+          root: "extra/lib/plausible_web/templates",
```
aerosol, line 75: :heart:

lib/plausible_web/templates/auth/login_form.html.heex
```diff
@@ -49,6 +49,17 @@
         </.styled_link>
         instead.
       </:item>
+      <:item :if={ee?() and Plausible.sso_enabled?()}>
+        <%= on_ee do %>
```
aerosol, line 53: is this actually compile-time as opposed to `:if={ee?()}` ? looks redundant though since `item` is already rendering conditionally

zoldar: The `Routes.sso_path` is compiled conditionally so it won't be available on CE. At the same time, I can't wrap <:item></:iteam> with `on_ee` because Phoenix complains it's not a valid syntax. Hence I had resorted to using both.

extra/lib/plausible/auth/sso.ex
zoldar: `create_team_membership` being an upsert might end up with reusing existing guest membership. I'm treating it on equal footing with user with identical email but not an explicit member on the team (the email uniqueness clause below).

Even though initial "find_user" lookup excludes guest memberships so this branch is not reachable under normal circumstances, given unlucky timing, the user could become guest right at the time of provisioning. So that's just being extra cautious.

PR #5440, opened 2025-05-28
CRM: team (un)lock regardless of grace period

aerosol:
### Changes

On top of #5438 this PR implements Team (un)locking capability in CRM.
If grace period is present it'll be ended (on lock) or removed (on unlock). 
Otherwise, the manually applied lock will hold and prevent users from accessing dashboards, regardless of any background quota/usage checks. This means, the only way to remove CRM-applied lock, is to manually remove it via CRM.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible_web/live/customer_support/live/site.ex
+++ b/extra/lib/plausible_web/live/customer_support/live/site.ex
@@ -79,6 +79,13 @@ defmodule PlausibleWeb.CustomerSupport.Live.Site do
             <.styled_link patch={"/cs/teams/team/#{@site.team.id}"}>
               {@site.team.name}
             </.styled_link>
+
+            <span
+              :if={Plausible.Teams.locked?(@site.team)}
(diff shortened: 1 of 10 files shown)


ukutaht (commented):
I think we should also avoid sending email reports to teams locked by admin

here: https://github.com/plausible/analytics/blob/ca7804cf86fd364d1d299eead068c14a4f8ff5d8/lib/workers/schedule_email_reports.ex#L38

and here: https://github.com/plausible/analytics/blob/ca7804cf86fd364d1d299eead068c14a4f8ff5d8/lib/workers/schedule_email_reports.ex#L79

aerosol:
@ukutaht thanks, pushed 279d4ddcf6 and also included traffic change notifications

ukutaht (commented):
Looks good! Missing tests though

ukutaht (approved):

aerosol:
@ukutaht tests added via https://github.com/plausible/analytics/pull/5440/commits/588d0a1630595e26749ad01667dea6185358b8b1

PR #5449, opened 2025-06-02
Proposal: ScriptV2 bundler

macobo:
### Changes

This PR proposes using rollup as a bundler for tracker script.

Benefit:
- Splits the code across multiple files

Drawbacks:
- More dependencies
- It increases the script size due to needing to use `config.domain`/`config.endpoint`/`config.autoCapturePageviews` over vars - this doesn't minify as nicely. See comment below for stats.

Best reviewed commit-by-commit.


apata:
Fleshing out the benefits and drawbacks more, I see the following 
* probably increased development ease from better code organization
* maybe increased risk of invalid code from needing to consider module to module interactions
  * (can be mitigated with linter, maybe typescript)
* increased supply chain attack risk from additional binary
* script size increase

@macobo Did you already try organizing the code in the single file? If you did and found it still hard to comprehend, then I think the drawbacks, small as they are, are acceptable. Even if you didn't try that, I'd be ok with us proceeding with rollup for sake of development velocity.

ukutaht (approved):

tracker/compiler/index.js
```diff
@@ -56,11 +57,11 @@ export async function compileAll(options = {}) {
   console.log(`Completed compilation of ${variants.length} variants in ${((Date.now() - startTime) / 1000).toFixed(2)}s`);
 }
 
-export function compileFile(variant, options) {
-  const baseCode = options.baseCode || getCode()
+export async function compileFile(variant, options) {
+  const wrappedCode = wrapInstantlyEvaluatingFunction(options.bundledCode || await bundleCode())
```
apata, line 61: nitpick, non-blocking: if we always get the bundledCode from options, there's less conditionals to go through when trying to understand this function

macobo: It'd require exposing `bundleCode` to server.js, keeping it. It's also a pretty standard memoization strategy.

PR #5442, opened 2025-05-29
CRM: internal TCP endpoint 

aerosol:
### Changes

This PR introduces new endpoint for superadmin access, so we can expose it internally via tailscale.

### Tests
- [x] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #5454, opened 2025-06-02
CRM: list API Keys

aerosol:
On top of #5442

### DIFF
--- a/extra/lib/plausible_web/live/customer_support/live/user.ex
+++ b/extra/lib/plausible_web/live/customer_support/live/user.ex
@@ -200,10 +200,8 @@ defmodule PlausibleWeb.CustomerSupport.Live.User do
   end
 
   defp keys_query(user) do
-    team_ids = user |> Plausible.Teams.Users.teams() |> Enum.map(& &1.id)
-
     from api_key in Plausible.Auth.ApiKey,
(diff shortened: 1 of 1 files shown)


extra/lib/plausible_web/live/customer_support/live/user.ex
```diff
@@ -148,4 +198,24 @@ defmodule PlausibleWeb.CustomerSupport.Live.User do
         {:noreply, assign(socket, form: to_form(changeset))}
     end
   end
+
+  defp keys_query(user) do
+    team_ids = user |> Plausible.Teams.Users.teams() |> Enum.map(& &1.id)
+
+    from api_key in Plausible.Auth.ApiKey,
+      where: api_key.user_id == ^user.id or api_key.team_id in ^team_ids,
```
zoldar: this is going to list all teams' keys as well I think, not just for that user. Just filtering by user should be enough, because keys can have either:

- user _and_ team set (new keys)
- user set and team set as nil (legacy keys)

```suggestion
      where: api_key.user_id == ^user.id,
```

aerosol: ah, right, OK. thanks 88ad833f6f

zoldar (approved):

PR #5459, opened 2025-06-03
ScriptV2: NPM library codebase

macobo:
Depends on https://github.com/plausible/analytics/pull/5449. If that gets rejected, I will rebase accordingly.

This PR is the start of work for creating a new NPM library for our tracker script off of the same tracker codebase.

```
import { init, track } from '@plausible/tracker' # or some other library name

init({ 
  domain: "plausible.io"
})
```

It:
- Has typescript types (via a type declaration file)
- Shares majority of tests with plausible-web (needs a rebase after https://github.com/plausible/analytics/pull/5432 is merged)
- Has protections against double-init/track-before-init, throwing in that case

Work left for the future:
- Release process via CI
- Renaming the library, MIT license, CHANGELOG.md and README.md
- Commonjs variant. Work on this was in this PR but was hard to test for.


tracker/compiler/analyze-sizes.js
```diff
@@ -229,21 +230,34 @@ function addSign(value) {
 }
 
 function readPlausibleScriptSizes() {
-  const files = fs.readdirSync(TRACKER_FILES_DIR).filter((filename) =>
-    !['.gitkeep', 'p.js'].includes(filename) && (filename.includes(currentSuffix) || filename.includes(baselineSuffix))
-  )
+  const trackerFileSizes = fs.readdirSync(TRACKER_FILES_DIR)
```
apata, line 233: question: Why don't we read the files by their names as declared in variants json? The suffixes are known in this script. At the moment, it seems that this script would succeed even if some of the variants weren't generated and there's extra work involved with isRelevantFile.

tracker/test/shared-configuration-tests.js
```diff
@@ -0,0 +1,243 @@
+/*
+Tests for plausible-web.js script variant
```
apata: nitpick: leftover header

tracker/test/plausible-npm.spec.js
```diff
@@ -0,0 +1,72 @@
+/*
+Tests for plausible-web.js script variant
+
+Unlike in production, we're manually interpolating the script config in this file to
+better test the script in isolation of the plausible codebase.
```
apata:
```suggestion
Tests for plausible-npm.js variant

Config is set at init(), as we expect consumers to do in production.
```

tracker/test/shared-configuration-tests.js
```diff
@@ -0,0 +1,243 @@
+/*
```
apata: Less repetition in tests, very nice!

apata (approved):
I like this! I appreciate the shared tests and that the npm module is tested at all. Lots of removed lines and small nuisances fixed.

tracker/test/plausible-npm.spec.js
macobo: :man_facepalming:

tracker/compiler/analyze-sizes.js
macobo: No particular reason - that approach would work as well but would have its own gotchas (e.g. what if a file doesn't exist with previous suffix as it was added in this compile run?). Ultimately given this is pretty isolated code :man_shrugging:

Daniel15:
If you search Google for "Plausible npm", the old package is still returned, and the new one isn't in the results at all. Most people are still downloading the old script rather than the new one, based on npm download stats. Do you have plans to mark the old one as deprecated at some point?

PR #5461, opened 2025-06-03
Implement `SSO.Domains.remove/1,2` and `SSO.Domains.check_can_remove/1`

zoldar:
### Changes

This PR extends SSO domains API with ability to remove SSO domains.

### Tests


extra/lib/plausible/auth/sso/domains.ex
```diff
@@ -50,6 +51,83 @@ defmodule Plausible.Auth.SSO.Domains do
...
+    |> users_by_domain_query()
+    |> Repo.all()
+  end
+
+  defp users_by_domain_query(sso_domain) do
+    from(
+      u in Auth.User,
+      where: u.sso_integration_id == ^sso_domain.sso_integration_id,
+      where: u.type == :sso,
+      where: ilike(u.email, ^"%@#{sso_domain.domain}")
```
zoldar: Another option is maintaining a separate explicit FK on users (users.sso_domain_id), in addition to integration ID. By making this constraint restrict on domain deletion, we could plug a hole where an SSO user can be left behind without domain (something that I'll address for users.sso_integration_id as it's currently setting null on delete implicitly). This requires some _very_ unlucky timing though (but it's not impossible). WDYT?

zoldar: I'll add a separate FK. The migration: https://github.com/plausible/analytics/pull/5464

PR #5462, opened 2025-06-03
Fix invalid onIgnoredEvent call

apata:
### Changes

The arguments of an onIgnoredEvent call are in the wrong order.

### Tests

### Changelog

### Documentation

### Dark mode


macobo:
Good catch, thanks! This runs into conflict with https://github.com/plausible/analytics/pull/5449, bundled the change in there and will merge it as part of that branch.

PR #5465, opened 2025-06-03
Script v2: Remove delays except when using .compat extension

apata:
#### Functional
Removes delays and navigation related overrides from tracker script, except when using .compat extension.
This is possible because fetch with keepalive: true reliably sends events that happen during navigating away.

#### Support
- Adds tests for v2 targets for tagged events, outbound links, file downloads
- Adds .compat extension specific tests for tagged events, outbound links, file downloads
- Improves initializePageDynamically to allow testing with legacy scripts
- Removes mockRequest, refactors tests where it was used
  - pageviews spec
  - hash exclusions

Introduces the following testing structure:
```md
There's a spec file for every feature (ex: file downloads). The spec file contains
-- describe block for parity cases between legacy and v2 (tests generated with a loop, with just the script install method differing between iterations)
  -- test case
  -- ...
-- describe block for cases specific to v2 targets (if any)
   -- test case
   -- ...
-- describe block for cases specific to legacy (if any)
   -- test case
   -- ...
-- describe block for cases specific to legacy compat mode (if any)
   -- test case
   -- ...
```

### Tests

### Changelog

### Documentation

### Dark mode


tracker/test/support/types.ts
```diff
@@ -1,13 +1,13 @@
...
-  hashBasedRouting: boolean;
-  outboundLinks: boolean;
-  fileDownloads: boolean;
-  formSubmissions: boolean;
-  captureOnLocalhost: boolean;
-  autoCapturePageviews: boolean;
-};
+  hashBasedRouting: boolean
+  outboundLinks: boolean
+  fileDownloads: boolean | string[]
```
RobertJoonas, line 4: This also calls for a change in `tracker/npm_package/README.md`.

tracker/test/file-downloads.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+        expectedRequests: [{ n: 'File Download', p: { url: isoFileURL } }],
+        shouldIgnoreRequest: [isPageviewEvent, isEngagementEvent]
+      })
+      await expect(isoMock.getRequestList()).resolves.toHaveLength(1)
+    })
+  })
+}
+
+test.describe('file downloads feature edge cases (esm)', () => {
+  test('malformed `fileDownloads: "iso"` option enables the feature with default file types', async ({
```
RobertJoonas: Q: if `boolean | string[]` is the allowed type, why not throw an "invalid config" error when this happens?

tracker/test/file-downloads.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+      action: () => page.click(`a[href="${isoFileURL}"]`),
+      refutedRequests: [{ n: 'File Download' }],
+      shouldIgnoreRequest: [isPageviewEvent, isEngagementEvent]
+    })
+    await expect(isoMock.getRequestList()).resolves.toHaveLength(1)
+  })
+})
+
+test.describe('file downloads feature edge cases (legacy .compat extension)', () => {
+  for (const { caseName, fulfill } of [
```
RobertJoonas: A lot of delay in these tests which makes them pretty slow (> 10s).

I believe the same coverage could be achieved with the three following test cases:

* "sends only one `File Download` event"
  * should run ~3s
  * makes sure that exactly one event request gets sent (wait for 3000ms)
  * does not care about download requests
* "if event sending is slow, starts exactly one download"
  * should run ~6s
  * waits for a single event request (delay the event response 3000ms, timeout 4000ms)
  * makes sure exactly 1 download started (wait for 6000ms)
* "starts download even when event request fails"
  * should run < 1s
  * wait for 1 event request (default delay 3000ms but should happen instantly if it works)
  * wait for 1 download request (should happen instantly if it works)
  
PS: Why not use `expectPlausibleInAction` utility to expect the event requests?

tracker/test/file-downloads.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+      testId
+    }) => {
+      const filePath = '/file.csv'
+      const { getRequestList } = await mockManyRequests({
+        page,
+        path: `${LOCAL_SERVER_ADDR}${filePath}`,
+        fulfill: {
+          contentType: 'text/csv'
+        },
+        countOfRequestsToAwait: 2
```
RobertJoonas: Sounds like an overkill to test the "no more than one download gets started" (i.e. wait for 2 requests but expect only one) in four different tests. I don't see how this behaviour could differ between a regular click, link child click, `target=_blank`, and `controlOrMeta` modifier cases. Let's keep it only here to speed up the test suite?

apata: Good shout about overkill! Moved it to legacy .compat tests only, since we only play around with default behaviour in .compat mode now.

tracker/test/file-downloads.spec.ts
apata: Thanks for thinking through this! I took this into account as best I could and refactored all the affected features to have a similar suite of .compat tests. The reason not to use expectPlausibleInAction is because it makes it harder to run with a Promise.all (because the actual action is wrapped into one of the promises). It also doesn't have the fulfill API yet.

tracker/test/file-downloads.spec.ts
apata: This was script config API decision, and it may be still up for debate if you'd like. I think throwing on unexpected configuration has advantages, or warning and defaulting to false. Prevents unintentional events. The current behaviour seems more connected with JS truthy values than customer experience.

tracker/playwright.config.js
```diff
@@ -6,7 +6,8 @@ import { defineConfig, devices  } from '@playwright/test'
  */
 export default defineConfig({
   testDir: './test',
-  timeout: 60 * 1000,
+  /* Can be overridden in specific tests with test('a longer running test', async () => { test.setTimeout(<longer timeout>); // test content... }) */
+  timeout: 10 * 1000,
```
RobertJoonas, line 10: Nice! 👌

tracker/test/file-downloads.spec.js
```diff
@@ -1,129 +0,0 @@
-import { mockRequest, metaKey, expectPlausibleInAction } from './support/test-utils'
-import { expect, test } from '@playwright/test'
-import { LOCAL_SERVER_ADDR } from './support/server'
-
-test.describe('legacy file-downloads extension', () => {
-  test('sends event and does not start download when link opens in new tab', async ({ page }) => {
```
RobertJoonas, line 6: A link click with a `controlOrMeta` modifier, or a regular click on a `target="_blank"` link should not start the download in the current page context.

This behaviour is not relevant in v2, nor legacy without compat, but crucial to test for the legacy `.compat` mode where our script is interrupts browser behaviour. Tests should make sure our tracking doesn't change anything that would happen naturally in the browser.

Currently, the whole test suite passes, if I remove the following logic:

image
image

_NOTE: Also relevant for outbound links and tagged events._

tracker/test/custom-event-edge-cases.spec.js
```diff
@@ -1,45 +0,0 @@
...
-
-    await expectPlausibleInAction(page, {
-      action: () => page.click('#tagged-outbound-download-link'),
-      expectedRequests: [{ n: 'Foo', p: { url: 'https://awesome.website.com/file.pdf' } }],
-      awaitedRequestCount: 3,
-      expectedRequestCount: 1
-    })
-  })
-
-  test('sends manual non-interactive custom event', async ({ page }) => {
```
RobertJoonas, line 37: This was removed. Accidental?

tracker/test/outbound-links.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+      testId,
+      scriptConfig:
+        '<script id="plausible" defer src="/tracker/js/plausible.compat.local.manual.outbound-links.js"></script>',
+      bodyContent: `<a href="${outboundUrl}">➡️</a>`
+    })
+    await page.goto(url)
+    const navigationPromise = page.waitForRequest(outboundUrl, {
+      timeout: 7000
+    })
+    const trackingPromise = page.waitForResponse('**/api/event', {
```
RobertJoonas: Feels a bit weird that we're mocking a response to the event request. In reality, it wouldn't even be initiated if navigation gets in the way. How about simplifying this test using `expectPlausibleInAction` with a responseDelay of 5100? It could just refute an `Outbound Link: Click` event while expecting a navigation request to happen after 5s.

tracker/test/outbound-links.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+        const outboundMock = await mockManyRequests({
+          page,
+          path: outboundUrl,
+          fulfill: {
+            status: 200,
+            contentType: 'text/html',
+            body: '<!DOCTYPE html><html><head><title>other page</title></head><body>other page</body></html>'
+          },
+          awaitedRequestCount: 1
+        })
```
RobertJoonas, line 54: This code block could perhaps be extracted into a helper function in this file and be used in 6 places. E.g. `mockNavigationRequest(outboundUrl, awaitedRequestCount = 1)`.

tracker/test/tagged-events.spec.ts
```diff
@@ -0,0 +1,523 @@
...
+      testId,
+      scriptConfig:
+        '<script id="plausible" defer src="/tracker/js/plausible.compat.local.manual.tagged-events.js"></script>',
+      bodyContent: `<a class="plausible-event-name=Subscribe" href="${targetPage.url}">Click to subscribe</a>`
+    })
+    await page.goto(url)
+    const navigationPromise = page.waitForRequest(targetPage.url, {
+      timeout: 7000
+    })
+    const trackingPromise = page.waitForResponse('**/api/event', {
```
RobertJoonas: Ditto my other comment for outbound links. The event request should not even be made because navigation happens before.

tracker/test/file-downloads.spec.ts
RobertJoonas: >  I think throwing on unexpected configuration has advantages, or warning and defaulting to false. Prevents unintentional events.

Agreed. Could you link to the discussion where this was decided?

tracker/test/outbound-links.spec.ts
apata: > Feels a bit weird that we're mocking a response to the event request. In reality, it wouldn't even be initiated if navigation gets in the way.

Ah, but the tracking request _is_ initiated, it just isn't usually finished in these scenarios. This behavior is the substance of the test. I think I'm checking that rather cleverly with `trackingPromise.then(Date.now).catch(Date.now)`. 

If we remove the mock for event API, tracking request finishes with 404 immediately and navigation happens. There are other cases for that behavior.  

If I try to simplify the test to expectPlausibleInAction, internally it _also_ mocks api/event, right? Seems like the level of weirdness would stay constant if I did that.

tracker/test/outbound-links.spec.ts
apata: True! I actually considered that, and a bunch of other such opportunities in the files that I've worked on. It is a annoying to see repetitive code, but it's also a pain to investigate spec-specific APIs in every spec. Shared constants between test cases is not exactly great either.

The only nice way I've seen it done is chainable test data functions for the primary things that appear in tests, shared across specs. Sort of like we do with new_site() etc in Elixir tests. I'd prefer not to introduce that new paradigm in this PR though.

tracker/test/file-downloads.spec.js
apata: Ah, damn! I hadn't checked if they fail without that block. The previous tests didn't verify that the browser actually requested the file in new tab, that's why I made these changes. In this fresh commit, I check both now: https://github.com/plausible/analytics/pull/5465/commits/68cc81bb8c218b654d4367318e8bdb8b09e3f20e

tracker/test/custom-event-edge-cases.spec.js
apata: Thanks for spotting it! It was really out of place next to the overlap cases, so I removed it. I've put it under manual.spec.js now. https://github.com/plausible/analytics/pull/5465/commits/68cc81bb8c218b654d4367318e8bdb8b09e3f20e

RobertJoonas (approved):

PR #5467, opened 2025-06-04
Add resolved script features to legacy script endpoint metrics

apata:
### Changes

Adds resolved script features (aka base variant names) to legacy script serving counter metric. May be useful to gauge popularity or lack thereof of particular base variants. Ideally would not be needed, but can't find the breakdown by file in the CDN.

### Tests
Manually testable. Checkout the PR, run the app locally, visit http://localhost:8000/js/plausible.tagged-events.compat.js, then go to http://localhost:8000/metrics and look for the following
```
# HELP plausible_prom_ex_plausible_tracker_script_request_legacy 
# TYPE plausible_prom_ex_plausible_tracker_script_request_legacy counter
plausible_prom_ex_plausible_tracker_script_request_legacy{compat="true",exclusions="false",file_downloads="false",hash="false",local="false",manual="false",outbound_links="false",pageview_props="false",revenue="false",status="200",tagged_events="true"} 1
```

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/telemetry/plausible_metrics.ex
+++ b/lib/plausible/telemetry/plausible_metrics.ex
@@ -119,8 +119,16 @@ defmodule Plausible.PromEx.Plugins.PlausibleMetrics do
         counter(
           metric_prefix ++ [:tracker_script, :request, :legacy],
           event_name: PlausibleWeb.TrackerPlug.telemetry_event(:legacy),
-          tags: [:status],
-          tag_values: &%{status: &1.status}
+          tags: [:status] ++ PlausibleWeb.TrackerPlug.telemetry_legacy_variant_labels(),
+          tag_values: fn payload ->
+            Map.merge(
+              Map.new(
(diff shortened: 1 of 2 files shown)


aerosol:
Please be careful multiplying tags in grafana. Low cardinality is recommended. Is this a finite set of names? If we can normalize `variant.a.b` and `variant.b.a` into something uniform, we probably should. Disregard if inapplicable.

apata:
> Please be careful multiplying tags in grafana. Low cardinality is recommended. Is this a finite set of names? If we can normalize `variant.a.b` and `variant.b.a` into something uniform, we probably should. Disregard if inapplicable.

Good point, @aerosol. There's ~1024 unique legacy script names, which is probably too many values for one label. I could instead add one label (with boolean value) for every legacy option:

`plausible_prom_ex_plausible_tracker_script_request_legacy{status: <status>, compat: <boolean>, manual: <boolean>, ...}`

WDYT?

apata:
Closing, not needed right now

PR #5470, opened 2025-06-04
Add unique index on `users.sso_identity_id`

zoldar:
### Changes

This index was missing. Given we currently expect SSO user emails being unique across the system, the same expectation should be set for NameID/identity ID (though it's yet to be proved whether we can truly rely on it).

### DIFF
--- a/priv/repo/migrations/20250604094230_add_unique_index_on_users_sso_identity_id.exs
+++ b/priv/repo/migrations/20250604094230_add_unique_index_on_users_sso_identity_id.exs
@@ -1,0 +1,11 @@
+defmodule Plausible.Repo.Migrations.AddUniqueIndexOnUsersSsoIdentityId do
+  use Ecto.Migration
+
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20250604094230_add_unique_index_on_users_sso_identity_id.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Plausible.Repo.Migrations.AddUniqueIndexOnUsersSsoIdentityId do
+  use Ecto.Migration
+
+  import Plausible.MigrationUtils
+
+  def change do
+    if enterprise_edition?() do
```
aerosol, line 7: interesting. will it generate the `down` migration with that conditional as well?

zoldar: Why not? 🤔  The conditional wraps a reversible operation.

PR #5474, opened 2025-06-05
Starter Tier: Shared Link Feature Gates

RobertJoonas:
### Changes

This PR implements API feature gates for shared links, while granting special access per `shared_link.name` to links created by the plugin API (i.e. by our WP plugin).

Here's the breakdown of changes in this PR:

* With https://github.com/plausible/analytics/pull/5472 migrated, every enterprise plan has access to the SharedLinks feature according to the data. This PR stops granting `SharedLinks` access to enterprise plans in code.
  * Also, removes SharedLinks access from expired trials. Note that this will take effect immediately - it is not controlled by the `starter_tier` feature flag. The dashboards are locked for them anyway, so this should be OK.
* Shared Links API feature gates 
  * Disable shared link creation without access to SharedLinks (exception: allowed via Plugins API)
  * Declare `WordPress - Shared Dashboard` as the reserved "special" shared link name and prohibit creating new links with that name (except via Plugins API) and renaming existing links to that name.
  * The `GET /share/:domain` (i.e. the shared link itself) route is now restricting access (see screenshot below) according to `SharedLinks` feature availability, UNLESS the shared link has the special `WordPress - Shared Dashboard` name
  * The special WP shared link is not displayed under Site Settings > Visibility > Shared Links 
* Warn about losing access to the shared links feature when the team has any shared links created with a name other than `WordPress - Shared Dashboard` and wants to subscribe to the Starter tier.

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/segments.ex
+++ b/lib/plausible/segments/segments.ex
@@ -369,13 +369,4 @@ defmodule Plausible.Segments do
 
     "#{field} #{formatted_message}"
   end
-
-  @spec get_site_segments_usage_query(list(pos_integer())) :: Ecto.Query.t()
-  def get_site_segments_usage_query(site_ids) do
-    from(segment in Segment,
-      as: :segment,
-      where: segment.type == :site,
-      where: segment.site_id in ^site_ids
-    )
-  end
 end
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -338,17 +338,27 @@ defmodule Plausible.Sites do
     !!stats_start_date(site)
   end
 
-  def create_shared_link(site, name, password \\ nil) do
-    changes =
-      SharedLink.changeset(
-        %SharedLink{
-          site_id: site.id,
-          slug: Nanoid.generate()
-        },
-        %{name: name, password: password}
-      )
+  def create_shared_link(site, name, opts \\ []) do
+    password = Keyword.get(opts, :password, nil)
 
-    Repo.insert(changes)
+    site = Plausible.Repo.preload(site, :team)
+    skip_feature_check? = Keyword.get(opts, :skip_feature_check?, false)
+    feature_access? = Plausible.Billing.Feature.SharedLinks.check_availability(site.team) == :ok
+
+    if skip_feature_check? or feature_access? do
+      changes =
+        SharedLink.changeset(
+          %SharedLink{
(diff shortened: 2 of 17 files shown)


lib/plausible/segments/segments.ex
```diff
@@ -369,13 +369,4 @@ defmodule Plausible.Segments do
 
     "#{field} #{formatted_message}"
   end
-
-  @spec get_site_segments_usage_query(list(pos_integer())) :: Ecto.Query.t()
```
aerosol, line 373: :sparkles: :nail_care:

lib/plausible/sites.ex
```diff
@@ -338,17 +354,39 @@ defmodule Plausible.Sites do
...
-    changes =
-      SharedLink.changeset(
-        %SharedLink{
-          site_id: site.id,
-          slug: Nanoid.generate()
-        },
-        %{name: name, password: password}
-      )
+  def create_shared_link(site, name, opts \\ []) do
+    password = Keyword.get(opts, :password, nil)
```
aerosol:
```suggestion
    password = Keyword.get(opts, :password)
```

lib/plausible_web/controllers/site_controller.ex
```diff
@@ -608,17 +625,23 @@ defmodule PlausibleWeb.SiteController do
...
-        redirect(conn, to: Routes.site_path(conn, :settings_visibility, site.domain))
-
-      {:error, changeset} ->
-        conn
-        |> assign(:skip_plausible_tracking, true)
-        |> render("edit_shared_link.html",
-          site: site,
-          changeset: changeset
-        )
+    if params["name"] in Plausible.Sites.shared_link_special_names() do
```
aerosol: Feels like this could be centralized at `SharedLink.changeset` level? It can of course accept options to differentiate between caller contexts. WYT?

aerosol (approved):
Looks good! Minor remarks inline

lib/plausible_web/controllers/site_controller.ex
RobertJoonas: Good call. https://github.com/plausible/analytics/pull/5474/commits/21a7be3fe0a8157ab38acfe7eaf37754c10df8dc

PR #5477, opened 2025-06-06
ScriptV2: Tracker release process, documentation

macobo:
This PR:
1. Adds a new CI-based release process for tracker npm code
2. Adds actions which validate correct labels and CHANGELOG.md has been updated.
3. Adds documentation for the tracker codebase (README.md, ARCHITECTURE.md), npm code.

The new release process is as follows:

1. When creating PRs with tracker changes, a `tracker release: ` label must be set and CHANGELOG.md updated under Unreleased section.
2. On merge:
    - package.json and CHANGELOG.md are automatically updated and committed
    - new release is pushed to NPM 
    - the team is notified

Notes:
- The release currently releases `macobo-test-tracker` library instead of an official one. We'll switch over to the "official" library once we're ready to release.
- Weird diffs for LICENSE/CHANGELOG files might be due to symlinks and how github portrays them. In practice, CHANGELOG and LICENSE files are identical under tracker/ and tracker/npm_library.
- We'll change how `tracker_script_version` works at a later date.

### DIFF
--- a/.github/workflows/tracker-script-npm-release.yml
+++ b/.github/workflows/tracker-script-npm-release.yml
@@ -31,15 +31,15 @@ jobs:
       - name: Install dependencies
         run: npm install --prefix tracker
 
-      - name: Bump the patch version
+      - name: Bump the patch version and update changelog
         if: "${{ contains(github.event.pull_request.labels.*.name, 'tracker-release: patch') }}"
         run: npm run npm:prepare_release:patch --prefix tracker
 
-      - name: Bump the minor version
+      - name: Bump the minor version and update changelog
         if: "${{ contains(github.event.pull_request.labels.*.name, 'tracker-release: minor') }}"
         run: npm run npm:prepare_release:minor --prefix tracker
 
-      - name: Bump the major version
+      - name: Bump the major version and update changelog
         if: "${{ contains(github.event.pull_request.labels.*.name, 'tracker-release: major') }}"
         run: npm run npm:prepare_release:major --prefix tracker
 


tracker/npm_package/README.md
```diff
@@ -0,0 +1,106 @@
...
+import { init } from 'macobo-test-tracker'
+
+init({
+  domain: 'my-app.com'
+})
+```
+
+### Configuration options
+
+See also [plausible.d.ts](https://github.com/plausible/analytics/blob/master/tracker/npm_package/plausible.d.ts) for typescript types.
```
apata, line 40: nitpick, non-blocking: This section and types duplicate each other. Since the consumers of this document are probably web developers, can we just keep the link to the types? Also, it seems we can use relative file links that also resolve properly on npm website: https://stackoverflow.com/a/68738752

tracker/README.md
```diff
@@ -0,0 +1,49 @@
...
+More instructions can be found in [yalc repo](https://github.com/wclr/yalc).
+
+## Cloud deployment
+
+Handled via PRs. When making tracker changes, it's required to:
+- Tag your PR with a `tracker release:` label
+- Update `tracker/CHANGELOG.md`
+
+After merge github actions automatically:
+- include the updated tracker scripts in the next cloud deploy
```
apata: meganitpick: 
```suggestion
- includes the updated tracker scripts in the next cloud deploy
```

tracker/ARCHITECTURE.md
```diff
@@ -0,0 +1,86 @@
...
+
+Smaller scripts lead to better user experience and reduce the amount of data we miss due to script download and runtime.
+
+We minify our scripts using the speedy `@swc/core` library and track script size changes on every pull request.
+
+### 2. Many targets, single codebase
+
+Plausible provides a web 'snippet' users can include on their site, tooling for wordpress and other plugins as well as an npm package.
+
+As Plausible doesn't have the workforce to maintain multiple code bases, everything is built from the same underlying source code.
```
apata, line 22: non-blocking: I think somewhere I'd like to see a list of the targets / variants with their differences listed. 

Also, I feel we have a bit of uncertainty between "base variant", "variant", "target terms" encoded in the repository. 

The following suggestion (which BTW can definitely be improved upon) goes with "variant" & "feature", omitting target entirely. 
```suggestion
### 2. Variants
Defined in variants.json (must be regenerated whenever globals related logic changes)

- tracker script for web
  - tracking features are toggled from Plausible site settings, without needing to update the tracked webpage
  - possible to provide some additional options on tracked webpage, as well as override the ones resolved from Plausible site settings
  
- tracker module published to npm
  - same as above, except all tracking features and options must be toggled within javascript

- legacy tracker script variants 
  - tracking features are toggled by including a specific variant of the script on the tracked webpage
```

Finally, there's one more little suggestion: since workforce isn't the only reason to depend build the variants from the same source code, maybe it can be omitted entirely. There's technical benefits as well, such as higher parity between different installs, probably more.

tracker/ARCHITECTURE.md
```diff
@@ -0,0 +1,86 @@
+# Architecture
+
+This document describes the design goals informing the architecture of the Plausible tracker script codebase as well as a map
+to how the code is laid out.
+
+## Design goals
+
+We want to provide users a javascript suite to capture page views and activities done on the web page.
+
+Over time, the following key design goals have emerged:
```
apata, line 10: suggestion, non-blocking: I noticed that this chapter includes goals as well as implementation details. This means that tooling needed to support each goal is introduced in semi-random order (swc is mentioned first, rather than variants and what they are).

I think we could avoid mixing the two by listing the goals as bullet points first, then describing the script compile and support logic separately step-by-step.

tracker/package.json
```diff
@@ -1,11 +1,16 @@
...
   "scripts": {
     "deploy": "node compile.js",
     "test": "npx playwright test",
     "test:local": "npx playwright test",
-    "start": "node test/support/server.js"
+    "start": "node test/support/server.js",
+    "npm:prepare_release:patch": "npm run npm:bump_version patch && npm run npm:update_code_and_changelog",
+    "npm:prepare_release:minor": "npm run npm:bump_version minor && npm run npm:update_code_and_changelog",
+    "npm:prepare_release:major": "npm run npm:bump_version major && npm run npm:update_code_and_changelog",
+    "npm:bump_version": "npm --prefix npm_package version --no-git-tag-version --",
```
apata, line 12: suggestion, non-blocking: I checked that `npm --prefix tracker/npm_package version --no-git-tag-version -- patch` works, so we should be able to get rid of the duplicative package script names (one for every update level) and instead run have the && in the workflow file, where each update level anyway has a separate job
```yaml
      - name: Bump the minor version and create git tag on release
        if: "${{ contains(github.event.pull_request.labels.*.name, 'tracker-release: minor') }}"
        run: npm run npm:bump_version --prefix tracker/npm_package && npm:update_code_and_changelog --prefix tracker
```

.github/workflows/tracker-script-npm-release.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Publish tracker script to NPM
+        run: npm publish
+        working-directory: tracker/npm_package
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.TRACKER_RELEASE_NPM_TOKEN }}
+
+      - name: Commit and Push changes
+        uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5
+        with:
```
apata, line 59: suggestion, non-blocking: for peace of mind, let's be specific with the files and folders we expect to have changed and specify `with.add`: https://github.com/EndBug/add-and-commit?tab=readme-ov-file#inputs

tracker/npm_package/README.md
macobo: This is intentional - I wanted to ensure that README.md contains all the examples and contents to get started. If this becomes a hassle to maintain we _can_ start linking, but this is more user-friendly for now.

tracker/package.json
macobo: This was intentional - I want to keep the yaml as lean as possible for this situation.

PR #5478, opened 2025-06-06
Log warning when m/meta parameter used in ingestion

macobo:
This parameter was added in October 2020 and then renamed 2 days later: https://github.com/plausible/analytics/commit/40900c7653684d36135b4955b0a0ccfa6b7a304c and https://github.com/plausible/analytics/commit/0b6e645b44b40cc69a01a4fed5445a00277b5a4b

We are hoping we may be able to remove it which this logging will help with.

Planning to revert this once we have a couple of days worth of data.

### DIFF
--- a/lib/plausible/ingestion/request.ex
+++ b/lib/plausible/ingestion/request.ex
@@ -233,7 +233,9 @@ defmodule Plausible.Ingestion.Request do
 
     # Temporary instrumentation to see if `m`/`meta` field was used.
     if request_body["m"] || request_body["meta"] do
-      Logger.warning("Ingestion: request body contains meta field. #{inspect(request_body)}")
+      Logger.warning(
+        "Ingestion: request body contains meta field. domain=#{request_body["d"] || request_body["domain"]}"
+      )
     end
 
     changeset


lib/plausible/ingestion/request.ex
```diff
@@ -230,6 +231,11 @@ defmodule Plausible.Ingestion.Request do
       |> Enum.take(@max_props)
       |> Map.new()
 
+    # Temporary instrumentation to see if `m`/`meta` field was used.
+    if request_body["m"] || request_body["meta"] do
+      Logger.warning("Ingestion: request body contains meta field. #{inspect(request_body)}")
```
apata: question: Do we absolutely need to log the request_body? It increases the risk of data leak if data from the visitors of our customers is in server logs, however briefly. If the question is whether m or meta is backed up by p or props in these requests, maybe log the map keys of the request body?

macobo: The reason to log request_body is to ease debugging: we will know what sites and events trigger the field. `request_body` should only contain information that ends up in ClickHouse and we don't retain logs for long so I don't see a problem with this approach.

We're expecting the number of times this to be hit to be zero, but we'll see.

ukutaht: Yeah I'd be more careful with logging IP addresses (personal data) but the request body doesn't contain anything that doesn't end up in the database anyways.

apata: At this point in the ingestion pipeline, do we know that it doesn't, @ukutaht? Also, there's a difference in access levels and access protocol between logs and database. Since we can answer the question without logging the full request body, I think we should avoid it.

apata: What I mean is that there's no guarantee that our customers are refraining from including personal identifiers of their customers in the request payload, within `meta`, `m`, `p`, `props`, or any other random key that the server otherwise ignores. 

```sh
curl -X POST http://localhost:8000/api/event -H 'content-type:application/json' -d '{"d": "dummy.site", "n": "pageview", "u": "/purchase", "m": {"client": "...personal data"}, "customer name": "...personal data"}'
```

```
14:36:02.526 request_id=GEdcwJjxa1RqKw0AAUhB [warning] Ingestion: request body contains meta field. %{"customer name" => "...personal data", "d" => "dummy.site", "m" => %{"client" => "...personal data"}, "n" => "pageview", "u" => "/purchase"}
```

My argument is that we'd rather not know, and store it only where absolutely necessary.

Additionally, wouldn't merging this open us up for being spammed with warnings maliciously for the duration this is deployed? 

In my mind, the question that we need to answer to make the decision about meta is whether m or meta is used, and if they are used, are they equal to p or props. If it's used and not equal, we want to log the domain and domain only. By the count of unique domains we can tell whether we can kindly ask each of them (if any) to migrate or whether we need to keep supporting it.

macobo: This is overcomplicating things, but for the sake of getting this merged I've pushed a change to only log domain. All the other information will be stored in clickhouse and I can trade off my time to look things up if needed.

apata: > This is overcomplicating things, but for the sake of getting this merged I've pushed a change to only log domain.

You needn't have pushed a change if you didn't find my argument convincing, @macobo! One approval was sufficient to deploy. Hopefully there's nothing to look up and the reduced risk from logging the bare minimum is without time tradeoffs.

PR #5479, opened 2025-06-06
ScriptV2: remove installation_meta mentions from schema

macobo:
Nothing is reading or writing this anymore. We can this way (hopefully) eventually drop the column as well.


ukutaht:
> We can this way (hopefully) eventually drop the column as well.

Is there anything blocking/preventing us from doing it immediately?

macobo:
We can't before this is deployed. 

After this PR, I'm worried about the table becoming locked during column drop. Before postgres 11/12 dropping a column caused the whole table to be rewritten and this would for sure could cause an incident. In newer versions it grabs a lock that blocks all SELECTs which could also cause trouble.

PR #5480, opened 2025-06-09
CRM: Refund lock

aerosol:
### Changes

This PR adds a "Refund lock" link to cancelled subscriptions. 
This is different than a grace period lock (it does not depend on it). This lock cannot be undone manually, the user will have to resubscribe in order to unlock their dashboards.

image

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/teams.ex
+++ b/lib/plausible/teams.ex
@@ -59,8 +59,7 @@ defmodule Plausible.Teams do
     def on_trial?(team) do
       team = with_subscription(team)
 
-      not Plausible.Billing.Subscriptions.active?(team.subscription) &&
-        trial_days_left(team) >= 0
+      is_nil(team.subscription) and trial_days_left(team) >= 0
     end
   else
     def on_trial?(_), do: always(true)
--- a/lib/plausible_web/components/billing/notice.ex
+++ b/lib/plausible_web/components/billing/notice.ex
@@ -114,14 +114,28 @@ defmodule PlausibleWeb.Components.Billing.Notice do
         } = assigns
       ) do
     ~H"""
-    <aside id="global-subscription-cancelled-notice" class="container">
+    <aside
+      :if={not Subscriptions.expired?(@subscription)}
+      id="global-subscription-cancelled-notice"
+      class="container"
+    >
       <.notice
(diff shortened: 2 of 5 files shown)


lib/plausible/teams.ex
```diff
@@ -59,8 +59,7 @@ defmodule Plausible.Teams do
     def on_trial?(team) do
       team = with_subscription(team)
 
-      not Plausible.Billing.Subscriptions.active?(team.subscription) &&
-        trial_days_left(team) >= 0
+      is_nil(team.subscription) and trial_days_left(team) >= 0
```
RobertJoonas, line 62: ✨

I've also confirmed that there are no active trials with expired subscriptions.

lib/plausible_web/components/billing/notice.ex
```diff
@@ -114,14 +114,28 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+          You have access to your stats until <span class="font-semibold inline"><%= Calendar.strftime(@subscription.next_bill_date, "%b %-d, %Y") %></span>.
+          <.link
+            class="underline inline-block"
+            href={Routes.billing_path(PlausibleWeb.Endpoint, :choose_plan)}
+          >
+            Upgrade your subscription
+          </.link>
+          to make sure you don't lose access.
+        </p>
+        <.lose_grandfathering_warning subscription={@subscription} />
```
RobertJoonas: Perhaps let's revert the change that inlined this and keep this HTML defined in the `subscription_cancelled_notice_body` function like before? My bad that I didn't realise the duplication before.

I tried refactoring `dismissable` into a single function, but hiding the notice with the help of localstorage is pretty hacky. The intent is that even if the user hides the global notice, they should still see it in the account settings. And on top of that, since we don't want to display it double, we *do not* show the notice in account settings unless it's globally hidden 🙈

RobertJoonas (approved):

lib/plausible_web/components/billing/notice.ex
aerosol: https://github.com/plausible/analytics/pull/5480/commits/f0fa89786700426396c615ba240c50415054c25b

PR #5488, opened 2025-06-10
ScriptV2: `transformRequest`

macobo:
### Changes

This PR makes a few changes to the tracker script:

#### 1. support for config.transformRequest

transformRequest allows users to either manipulate or ignore requests to plausible
with a minimal footprint.

Some use-cases we're aware of and are planning for:
1. Ignoring requests similar to previous exclusion rules in WordPress
2. Normalizing urls for requests

#### 2. Allow passing options.url for overriding url when calling track

Previously we allowed option `u` which is unintuitively named. That stays supported but is removed from types file.

#### 3. Release process: Commit as plausible bot

We'll commit the package.json/CHANGELOG.md change as Plausible-Bot which _should_ be allowed to push on master.

### DIFF
--- a/tracker/npm_package/README.md
+++ b/tracker/npm_package/README.md
@@ -50,7 +50,7 @@ See also [plausible.d.ts](https://github.com/plausible/analytics/blob/master/tra
 | `formSubmissions` | Whether to track form submissions. | `false` |
 | `captureOnLocalhost` | Whether to capture events on localhost. | `false` |
 | `customProperties` | Object or function that returns custom properties for a given event. | `{}` |
-| `transformRequest` | Function that allows transforming or ignoring requests | `null` |
+| `transformRequest` | Function that allows transforming or ignoring requests | |
 
 #### Using `customProperties`
 


tracker/npm_package/plausible.d.ts
```diff
@@ -67,3 +71,20 @@ export type PlausibleEventRevenue = {
...
+  d: string,
+  // Referrer
+  r?: string | null,
+  // Custom properties
+  p?: CustomProperties,
+  // Revenue information
+  $?: PlausibleEventRevenue,
+  // Whether the event is interactive
+  i?: boolean,
+} & Record<string, any>
```
apata: Nice to see this type laid out plain and simple!

nitpick: According to TS docs, using `any` is an anti pattern (https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any). For values that we're blindly passing through, `Record<string, unknown>` works just as well.

tracker/npm_package/plausible.d.ts
```diff
@@ -34,6 +34,10 @@ export interface PlausibleConfig {
   // Custom properties to add to all events tracked.
   // If passed as a function, it will be called when `track` is called.
   customProperties?: CustomProperties | ((eventName: string) => CustomProperties)
+
+  // A function that can be used to transform the payload before it is sent to the API.
+  // If the function returns null, the event will be ignored.
```
apata: suggestion, non-blocking: A more verbose variant

If transformRequest is defined, runs on all tracking events before the actual API request is made. It applies to all events: those sent automatically by the script and those sent manually on the page using the provided track function. If the function returns any **falsy** value for a particular payload, that event will not be sent. This can be used to avoid sending certain types of events, or modifying any event parameters, e.g. to clean URLs of values that should not be recorded.

tracker/npm_package/plausible.d.ts
```diff
@@ -56,7 +60,7 @@ export interface PlausibleEventOptions {
 
   // Overrides the URL of the page that the event is being tracked on.
   // If not provided, `location.href` will be used.
-  u?: string
+  url?: string
```
apata, line 66: nitpick, non-blocking: You're probably aware, but JIC want to mention that there's a JSDoc `@deprecated` tag that some IDEs helpfully show to the user. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated
```js
  /** @deprecated, use `url` instead */
  u?: string
  /** Overrides the URL of the page that the event is being tracked on. If not provided, `location.href` will be used. */
  url?: string
```
This would allow people that start using these type definitions to decide for themselves when they want to refactor their existing track calls.

tracker/npm_package/plausible.d.ts
apata: Also, it might be worth mentioning that if this function uses APIs that aren't available in the browsers their visitors use, events from such visitors will be missed. Same goes for `customProperties` function form.

tracker/src/track.js
```diff
@@ -55,7 +55,7 @@ export function track(eventName, options) {
   payload.v = COMPILE_TRACKER_SCRIPT_VERSION
 
   if (COMPILE_MANUAL) {
-    var customURL = options && options.u
+    var customURL = options && (options.u || options.url)
```
apata, line 58: question: are `0`, `""` possibly valid `u` values that we don't want to fall back on?

apata: Ah I see that it wouldn't matter, because they'd be excluded further down anyway with `payload.u = customURL ? customURL : location.href`. Non-empty string is the actual definition for `url` type then.

apata (approved):
It's a powerful utility! Really cleanly implemented as well, great job.

I see that there's a few gotchas with it and users will have to know what they are doing or they're going to mess up their stats (e.g. if they normalize `pageview` event urls but forget to normalize `engagement` urls). 

I hope it's not going to bite us in support: I mean needing to debug what went wrong with the users custom `transformRequest` implementations that may be different page-by-page on their website.

It hooks in at a time when some of the context is lost (e.g. what the download link element was that triggered File Download). Also, hash-related logic has run already as well. That limits its uses a bit. Although it has access to the current state of window, that might be stale by the time it runs. Food for thought.

It's also global (same function runs for all events on the same page). That may be a pain to manage for larger sites. 

Anyway, because of its very small footprint, it could easily live on in the script even if we decide to bring in some more specific config options in the future.

tracker/npm_package/plausible.d.ts
macobo: The NPM package does not have any existing users and the `u` property is undocumented. As such I'm intentionally not exposing it to new users to keep the types clean.

PR #5491, opened 2025-06-11
Onboarding review flow

ukutaht:
### Changes

Makes the 'review' flow functional in scriptv2 onboarding so that tracker can be viewed/updated after initial installation. Changes are:
* Render the 'review' flow for installationv2 screen
* Adds toggles back into the installation screen
* In installationv2, create tracker script config with `form_submissions`, `outbound_links` and `file_downloads` enabled by default
* Also sync goals immediately when tracker script is created with these options enabled

Remaining work for onboarding:
* Verification tool has to handle scriptv2
* Domain change flow
* Upgrade notices for sites with old script installed
* Tests for new installation screen

### DIFF
--- a/extra/lib/plausible_web/live/customer_support/live/team.ex
+++ b/extra/lib/plausible_web/live/customer_support/live/team.ex
@@ -358,10 +358,9 @@ defmodule PlausibleWeb.CustomerSupport.Live.Team do
               :if={not mod.free?()}
               x-on:change="featureChangeCallback(event)"
               type="checkbox"
-              name={"#{f.name}[features[]][]"}
-              value={mod.name()}
+              value={mod in (f.source.changes[:features] || [])}
+              name={"#{f.name}[features[]][#{mod.name()}]"}
               label={mod.display_name()}
-              checked={mod in (f.source.changes[:features] || [])}
             />
 
             <div class="mt-8 flex align-center gap-x-4">
@@ -878,7 +877,12 @@ defmodule PlausibleWeb.CustomerSupport.Live.Team do
   end
 
   defp update_features_to_list(params) do
-    Map.put(params, "features", Enum.reject(params["features[]"], &(&1 == "false" or &1 == "")))
+    features =
+      params["features[]"]
+      |> Enum.reject(fn {_key, value} -> value == "false" or value == "" end)
+      |> Enum.map(fn {key, _value} -> key end)
+
+    Map.put(params, "features", features)
   end
 
   defp preview_number(n) do
--- a/test/plausible_web/live/customer_support/teams_test.exs
+++ b/test/plausible_web/live/customer_support/teams_test.exs
@@ -130,19 +130,19 @@ defmodule PlausibleWeb.Live.CustomerSupport.TeamsTest do
             "site_limit" => "1000",
             "team_member_limit" => "30",
             "hourly_api_request_limit" => "1000",
-            "features[]" => [
-              "false",
-              "false",
-              "false",
-              "teams",
-              "false",
-              "shared_links",
-              "false",
-              "false",
-              "false",
-              "false",
-              "sites_api"
-            ]
+            "features[]" => %{
+              "stats_api" => "false",
+              "outbound_links" => "false",
+              "form_submissions" => "false",
+              "file_downloads" => "false",
+              "funnels" => "false",
+              "props" => "false",
+              "shared_segments" => "false",
+              "revenue_goals" => "false",
+              "site_segments" => "false",
+              "shared_links" => "true",
+              "sites_api" => "true"
+            }
           }
         })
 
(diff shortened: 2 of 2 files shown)


lib/plausible_web/live/installation.ex
```diff
@@ -40,50 +40,47 @@ defmodule PlausibleWeb.Live.Installation do
         :viewer
       ])
 
-    socket =
-      if FunWithFlags.enabled?(:scriptv2, for: site) do
-        redirect(socket, to: "/#{domain}/installationv2?flow=#{params["flow"]}")
-      else
-        socket
+    if FunWithFlags.enabled?(:scriptv2, for: site) do
```
ukutaht, line 43: The only change here is putting the initialization code in the `else` block of this conditional. Why? Because the initialization code will create the tracker_script_config but we want to create it differently when `scriptv2` is enabled.

lib/plausible_web/live/installationv2.ex
```diff
@@ -19,19 +19,34 @@ defmodule PlausibleWeb.Live.InstallationV2 do
       ])
 
     tracker_script_configuration =
-      PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(site.id)
+      PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(site, %{
```
ukutaht, line 22: Eventually these defaults can be set at the schema level. For now I didn't do it since we're also keeping the old onboarding flow functional and it shouldn't use these defaults.

lib/plausible_web/tracker.ex
```diff
@@ -71,21 +71,36 @@ defmodule PlausibleWeb.Tracker do
     updated_config
   end
 
-  def get_or_create_tracker_script_configuration!(site_id) do
-    configuration = Repo.get_by(TrackerScriptConfiguration, site_id: site_id)
+  def get_or_create_tracker_script_configuration!(site, params \\ %{}) do
```
macobo, line 74: Note: This breaks plugins tests: test/plausible_web/plugins/api/controllers/tracker_script_configuration_test

lib/plausible_web/tracker.ex
```diff
@@ -71,21 +71,36 @@ defmodule PlausibleWeb.Tracker do
...
+      {:ok, created_config} =
+        Repo.transaction(fn ->
+          created_config =
+            TrackerScriptConfiguration.installation_changeset(
+              %TrackerScriptConfiguration{site_id: site.id},
+              params
+            )
+            |> Repo.insert!()
+
+          sync_goals(site, %{}, created_config)
```
macobo, line 89: I don't think we want to do this here.

If you remember, we discussed _always_ creating the goals as part of the new onboarding flow. Otherwise NPM users would not see e.g. outbound links. (Ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8689743935)

Doing it in onboarding also keeps this function leaner.

macobo (commented):
Approach looks solid. Two comments so far.

lib/plausible_web/tracker.ex
ukutaht: > If you remember, we discussed always creating the goals as part of the new onboarding flow. Otherwise NPM users would not see e.g. outbound links. (Ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8689743935)

All the goals are created already in the onboarding flow as the tracker script config is created.

> Doing it in onboarding also keeps this function leaner.

To move goal creation to onboarding, onboarding needs to be aware of whether the tracker script config is already created or not. So in addition to the `Tracker.get_or_create` I would need separate `get` and `create` functions. Is this expansion of the `Tracker` module API preferable than the implementation in this PR?

macobo: I think there's more discussion to be had here but I think it's wise to do this outside of this PR. Will do so under https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8689743935 later.

macobo (approved):
PR looks good pending tests succeed.

ukutaht:
@aerosol fyi I had some issue with checkboxes. Our checkbox implementation allows values other than `true / false` which is different from the phoenix generated form helpers. I think this was done for the custom plan form in CS. 

The problem it caused is when a field is `true` and the checkbox is re-submitted with no changes, it would flip the field to `false`. Ran into this in the course of creating the PR but this also currently affects the `public?` checkbox on site form in CS. When you try to make a site public=true and then submit the form again, it will fail with a validation error.

This could be worked around with more carefully managing the `value` and `checked` properties on the inputs. But I thought it would be more straight-forward to revert to the standard checkbox implementation from the generator which only allows boolean values: https://github.com/plausible/analytics/pull/5491/commits/cad79ea44cc28b09ff16d994fde392298dca986b#diff-8c349e39a07f88c3cb1d60bbb5069d0b98a6a059aaab5618d7271fda420f8e53R109

And refactor the place where non-boolean values were used: https://github.com/plausible/analytics/pull/5491/commits/501f3576f1e7dbbb757e4387da25d070361bdf15

WDYT?

aerosol:
@ukutaht that looks way better, thank you. I was raging my way through it not really understanding what I was doing.

PR #5493, opened 2025-06-11
ScriptV2: Rework WordPress plugin token adding flow

macobo:
### Changes

This PR aims to make plugin token creation flow less error-prone for WordPress users. Basecamp ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8720166828

<details>
<summary>Currently, when users are configuring the WordPress plugin users are presented the token before it's actually created.  This has led to issues where users copy the token but don't hit "Add" and run into issues. (click to see screenshot)</summary>
image
</details>

This PR splits the flow into two: Users first need to add a description and hit create. Only once they've done that is the token created and displayed.

### Recordings of the new flow

Wordpress flow: [wordpress.webm](https://github.com/user-attachments/assets/5ebbf41f-45a6-4f58-9305-fc9be116b032)

Changing via settings screen: [settings.webm](https://github.com/user-attachments/assets/3444e718-a2fa-40fe-b6fe-562e69875ebf)

### Notes for reviewers

Feel free to roast the code - I have little LiveView experience and some feedback on 'proper' ways of doing things would not hurt.

### DIFF
--- a/lib/plausible_web/live/plugins/api/token_form.ex
+++ b/lib/plausible_web/live/plugins/api/token_form.ex
@@ -36,8 +36,7 @@ defmodule PlausibleWeb.Live.Plugins.API.TokenForm do
        token: token,
        form: form,
        domain: domain,
-       rendered_by: pid,
-       tabs: %{custom_events: true, pageviews: false}
+       rendered_by: pid
      )}
   end
 
--- a/lib/plausible_web/live/plugins/api/token_form.ex
+++ b/lib/plausible_web/live/plugins/api/token_form.ex
@@ -36,7 +36,8 @@ defmodule PlausibleWeb.Live.Plugins.API.TokenForm do
        token: token,
        form: form,
        domain: domain,
-       rendered_by: pid
+       rendered_by: pid,
+       token_generated: false
      )}
   end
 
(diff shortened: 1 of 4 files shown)


lib/plausible_web/live/plugins/api/token_form.ex
```diff
@@ -37,81 +37,91 @@ defmodule PlausibleWeb.Live.Plugins.API.TokenForm do
...
-            Add Plugin Token for {@domain}
-          </.title>
+          <%= if @token_generated do %>
+            <div :if={@token_generated} class="mt-4">
+              <.input_with_clipboard
+                id="token-clipboard"
+                name="token_clipboard"
+                label="Plugin Token"
+                value={@token.raw}
+                onfocus="this.value = this.value;"
```
zoldar: I know this is carried over from existing code (and there are a couple other occurrences across the project as well) but I'm wondering if this actually does anything? 🤔

lib/plausible_web/live/plugins/api/token_form.ex
```diff
@@ -37,81 +37,91 @@ defmodule PlausibleWeb.Live.Plugins.API.TokenForm do
...
           class="max-w-md w-full mx-auto bg-white dark:bg-gray-800 shadow-md rounded px-8 pt-6 pb-8 mb-4 mt-8"
-          phx-submit="save-token"
-          phx-click-away="cancel-add-token"
+          phx-click-away="close-token-modal"
         >
-          <.title>
-            Add Plugin Token for {@domain}
-          </.title>
+          <%= if @token_generated do %>
+            <div :if={@token_generated} class="mt-4">
```
zoldar: Hmm, I think we can skip that `:if` given the surrounding one?

macobo: Great catch :star2:. I went through a couple different iterations of this trying to get a feel what would be best and this accidentally got left in after one of these refactors :facepalm:

lib/plausible_web/live/plugins/api/token_form.ex
macobo: I believe the intent was that when someone clicks on the input, it would automatically select all. However this does not work on my browser so removing it.

PR #5495, opened 2025-06-12
ScriptV2: Add note that NPM libary only works in browser environments

macobo:
Ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8709676872

This not being clear caused multiple users to file issues against https://github.com/plausible/plausible-tracker


RobertJoonas (approved):
👍

PR #5496, opened 2025-06-12
Periodically refresh salts so that non-rotating nodes catch up

aerosol:
### Changes

Salts are rotated via Oban cron-type job that is executed once at midnight on a single node exclusively. This means, other nodes won't update their salts state until they're restarted. Before we have a cluster over which we could broadcast updates to secondary nodes, each will periodically refresh salts from db, resulting in - at most - a minute-ish out of sync window, which should be acceptable. This should help with session tracking and session hand-off.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/session/salts.ex
+++ b/lib/plausible/session/salts.ex
@@ -9,8 +9,6 @@ defmodule Plausible.Session.Salts do
 
   @impl true
   def init(opts) do
-    Process.flag(:trap_exit, true)
-
     name = opts[:name] || __MODULE__
     now = opts[:now] || DateTime.utc_now()
     clean_old_salts(now)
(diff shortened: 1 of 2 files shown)


lib/plausible/session/salts.ex
```diff
@@ -23,8 +21,13 @@ defmodule Plausible.Session.Salts do
         {:read_concurrency, true}
       ])
 
-    salts =
-      Repo.all(from s in "salts", select: s.salt, order_by: [desc: s.inserted_at], limit: 2)
+    refresh(name, now)
```
ruslandoga, line 23: Maybe we can calculate time left till midnight and schedule once (per day)?

ruslandoga: Ah, I misunderstood the idea behind this PR.

lib/plausible/session/salts.ex
```diff
@@ -39,15 +42,20 @@ defmodule Plausible.Session.Salts do
           %{previous: nil, current: new}
       end
 
-    log_state("init", state)
```
ruslandoga, line 42: We can also remove `require Logger` now, I think.

ruslandoga (approved):

PR #5502, opened 2025-06-16
ScriptV2: Self hosted tracker script cache

macobo:
This PR leverages Plausible.Cache to provide a cached script delivery mechanism on self-hosted.


RobertJoonas (approved):
LGTM 👍

PR #5504, opened 2025-06-16
Dogfood script v2

RobertJoonas:
#### Companion PRs

* https://github.com/plausible/docs/pull/612
* https://github.com/plausible/website/pull/806

### Changes

* Updates dogfood tracking to use script v2
* Adds a few dogfooding tests
* Removes api_endpoint override that is no longer needed for testing purposes

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/dogfood.ex
```diff
@@ -3,29 +3,51 @@ defmodule PlausibleWeb.Dogfood do
...
+          "Qo3A7Ksnbn-wYQWMijuR3"
+
+        env == "prod" ->
+          "6_srOGVV9SLMWJ1ZpUAbG"
+
+        env == "staging" ->
+          "egYOCIzzYzPL9v6GHLc-7"
+
+        env in ["dev", "ce_dev"] ->
+          PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(1).id
```
apata: suggestion, non-blocking: If dummy.site is also dogfood, it creates the situation where the stats are not stable during the development of other features. I suggest we default to empty script url in these cases

assets/js/liveview/live_socket.js
```diff
@@ -18,13 +18,8 @@ if (csrfToken && websocketUrl) {
...
       this.handleEvent('send-metrics', ({ event_name }) => {
-        const afterMetrics = () => {
-          this.pushEvent('send-metrics-after', { event_name })
-        }
-        setTimeout(afterMetrics, 5000)
-        if (window.trackCustomEvent) {
-          window.trackCustomEvent(event_name, { callback: afterMetrics })
-        }
+        window.plausible(event_name)
+        this.pushEvent('send-metrics-after', { event_name })
```
apata, line 22: question: In the changed implementation, send-metrics-after is started at the same time as the fetch promise. Shouldn't we still use the callback API so it would really be "send-metrics-after"?

RobertJoonas: Here we can assume that `fetch()` is always called with `keepalive: true` (we're not using `compat` mode), so even if submitting the form would happen right after the fetch promise is started (in reality it takes a roundtrip to the server), we could still expect the request to be delivered.

extra/lib/plausible_web/dogfood.ex
RobertJoonas: Fair point but I personally feel like it's convenient if `dummy.site` receives stats for today out of the box. If you want the ClickHouse data to be static, you can just select a historical time range or switch to another site. Do you have any specific scenario in mind where this could become disturbing or confusing?

apata: Example: I'm working on segments BE with a segment applied and refreshing the dummy.site. Did stats change because one of my refreshes was counted as a visitor, or because I messed up the code that applies the filters? 

Also, is there actually any guarantee that site ID 1 exists in dev, ce_dev envs? 

If we want the convenience, how about we comment this line out, and give instructions to uncomment with desired site_id?

lib/plausible_web/templates/layout/_tracking.html.heex
```diff
@@ -1,44 +1,28 @@
...
-      <%= if @conn.assigns[:dogfood_page_path] do %>
-        const pageUrl = '<%= Path.join(plausible_url(), @conn.assigns[:dogfood_page_path]) %>'
-      <% else %>
-        const pageUrl = window.location.href
-      <% end %>
+    window.plausible=window.plausible||function(){(window.plausible.q=window.plausible.q||[]).push(arguments)},window.plausible.init=function(i){window.plausible.o=i||{}};var script=document.createElement("script");script.type="text/javascript",script.defer=!0,script.src=scriptParams.script_url;var r=document.getElementsByTagName("script")[0];r.parentNode.insertBefore(script,r);
 
-      const props = {
-        logged_in: <%= is_map(@conn.assigns[:current_user]) %>,
+    plausible.init({
```
apata, line 11: Can we override in init with `formSubmissions: false`, just in case?

extra/lib/plausible_web/dogfood.ex
RobertJoonas: Sure, nothing wrong with keeping it as is - not tracking itself by default. https://github.com/plausible/analytics/pull/5504/commits/4692f7d52136589127ff49c50c8d5c36bbda7c9f

lib/plausible_web/templates/layout/_tracking.html.heex
RobertJoonas: What should this protect us against? I believe we should indeed have tests in place making sure that form submissions, outbound links, and file downloads are *not tracked* unless explicitly enabled in `init`.

extra/lib/plausible_web/dogfood.ex
```diff
@@ -3,29 +3,57 @@ defmodule PlausibleWeb.Dogfood do
...
-    else
-      "#{PlausibleWeb.Endpoint.url()}/js/script.local.manual.pageview-props.tagged-events.pageleave.js"
-    end
+  defp script_url(assigns) do
+    env = Application.get_env(:plausible, :environment)
+    selfhost? = Application.get_env(:plausible, :is_selfhost)
+
+    tracker_script_config_id =
+      cond do
+        env == "prod" and selfhost? ->
```
ukutaht, line 22: What instance of Plausible is this case for?

lib/plausible_web/templates/layout/_tracking.html.heex
```diff
@@ -1,44 +1,28 @@
...
         browser_language: navigator.language || navigator.userLanguage
-      }
-
-      <%= if @conn.assigns[:current_user] do %>
-        props.theme = '<%= @conn.assigns[:current_user].theme %>'
-      <% end %>
-
-      const defaultEventOptions = {u: pageUrl, props: props}
+      },
+      transformRequest: function (payload) {
```
ukutaht, line 17: As discussed in the call we need a solution for normalizing URLs for dashboard routes (now that we are tracking these routes since we're not using `manual` mode anymore)

apata: @ukutaht I tested this on review and navigating around on the dashboard I saw no other events except pageview and engagement events with normalised URLs. Adding tracking to the deeper paths on dashboard with URL normalisation could maybe be a separate PR? 
Screenshot 2025-06-18 at 07 47 24

lib/plausible_web/templates/layout/_tracking.html.heex
apata: You're right! There are tests in place that the script config works as intended. Those don't protect us from accidentally enabling it as part of some manual testing routine: maybe I worry unnecessarily, though, feel free to resolve with no action.

lib/plausible_web/templates/layout/_tracking.html.heex
ukutaht: Oh you're right @apata. I'm confused why navigating to details views does not send a pageview but you're right it doesn't.

extra/lib/plausible_web/dogfood.ex
RobertJoonas: Good question! I assumed it's an instance created by someone in the team to test CE, but I don't know tbh. Perhaps @aerosol knows - what instance of plausible are we tracking at ee.plausible.io?

lib/plausible_web/templates/layout/_tracking.html.heex
RobertJoonas: > I'm confused why navigating to details views does not send a pageview but you're right it doesn't.

It was because the plausible.io script configuration had `hash_based_routing` set to `true` on staging for some reason. I've changed it to `false` so you can now see pageviews getting triggered on SPA navigation as well. https://github.com/plausible/analytics/pull/5504/commits/0662ea94d896d079d24133039ef2ac9714013773 also starts capturing those SPA routes (instead of overriding with a static `/:dashboard` path).

PR #5505, opened 2025-06-16
SSO: implement domain validation background service

aerosol:
### Changes

This PR implements background worker for SSO domain validation. The job gets exponential snooze on validation failure, leading up to 14 retries over roughly ~34 hours. Instead of deleting already running jobs, a retry is always called on `enqueue`, effectively scheduling the job for immediate execution (to provide the ability to quickly restart the process via UI).

New status indicators per `SSO.Domain` have been introduced: `in_progress` and `invalid` to mirror job's status. Each unique enqueue resets the status to `in_progress`. For now, nothing actually enqueues the job - a follow-up PR will be submitted with end to end integration.

On domain validation success/failure e-mail notifications will be sent to the relevant team's owners. Copy is subject to change, but paging @metmarkosaric to have a look at:

- https://github.com/plausible/analytics/blob/domain-verification-worker/lib/plausible_web/templates/email/sso_domain_validation_failure.html.heex
- https://github.com/plausible/analytics/blob/domain-verification-worker/lib/plausible_web/templates/email/sso_domain_validation_success.html.heex

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/workers/sso_domain_validation_worker_test.exs
+++ b/test/workers/sso_domain_validation_worker_test.exs
@@ -1,100 +1,109 @@
 defmodule Plausible.Auth.SSO.Domain.Validation.WorkerTest do
   use Plausible.DataCase
-  use Oban.Testing, repo: Plausible.Repo
+  use Plausible
 
-  use Bamboo.Test, shared: true
+  on_ee do
+    use Oban.Testing, repo: Plausible.Repo
 
-  alias Plausible.Auth.SSO.Domain.Validation.Worker
+    use Bamboo.Test, shared: true
 
-  alias Plausible.Auth.SSO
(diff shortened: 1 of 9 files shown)


test/workers/sso_domain_validation_worker_test.exs
```diff
@@ -0,0 +1,109 @@
+defmodule Plausible.Auth.SSO.Domain.Validation.WorkerTest do
```
aerosol, line 1: I should move that under `extra/` right?

zoldar: 🤔 I don't think we have moved any tests under `extra` yet, right? I recall looking into this for a short bit but gave up.

zoldar (approved):
✨

PR #5506, opened 2025-06-16
Revise system behaviour for SSO users

zoldar:
### Changes

This PR makes the logic account for SSO user being treated differently to standard user, ensuring the state integrity is not violated.

- standard login and password reset exclude SSO users from search, with exception of SSO team owners
- SSO user can't delete their account, disable 2FA, update their name, email or password
- SSO user can't accept team invitations
- SSO user can't add websites under "My Personal Sites"; they are also restricted from adding sites via API to a personal team
- Implicit personal team creation is restricted for SSO users, with hard/crashing restriction on forcible creation in billing webhook (should happen very rarely, given very unlucky circumstances and timing)
- SSO user is removed when they team membership is removed
- SSO user's role can't be changed to Owner when team have Force SSO policy enabled and user does not have 2FA setup
- Standard user can't be provisioned as SSO user if they have a personal team with any sites or active subscription

### Tests


lib/plausible_web/plugs/restrict_type.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Plausible.Plugs.RestrictType do
```
aerosol: I don't like the vagueness of "type" here. Should be "login type" or "auth type" or "user type" even.

lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -17,6 +17,9 @@ defmodule PlausibleWeb.SettingsController do
   plug Plausible.Plugs.AuthorizeTeamAccess,
        [:owner] when action in [:team_danger_zone, :delete_team]
 
+  plug Plausible.Plugs.RestrictType,
+       :sso when action in [:update_name, :update_email, :update_password]
```
aerosol: Shouldn't this be standard? :thinking:

lib/plausible/users.ex
```diff
@@ -16,7 +16,16 @@ defmodule Plausible.Users do
     end
   else
     @spec type(Auth.User.t()) :: :standard
-    def type(_user), do: :standard
+    def type(_user) do
```
aerosol, line 19: We have `always` when you `use Plausible` that should work here.

lib/plausible/teams/memberships/update_role.ex
```diff
@@ -46,6 +48,29 @@ defmodule Plausible.Teams.Memberships.UpdateRole do
...
+      if team.policy.force_sso == :all_but_owners and not Plausible.Auth.TOTP.enabled?(user) do
+        {:error, :mfa_disabled}
+      else
+        :ok
+      end
+    end
+
+    defp check_can_promote_to_owner(_team, _user, _new_role), do: :ok
+  else
+    defp check_can_promote_to_owner(_team, _user, _new_role) do
```
aerosol, line 62: ditto `always`?

lib/plausible_web/controllers/settings_controller.ex
zoldar: Yeah, that's ambiguous as well. Maybe that should be an option, like 

```elixir
plug Plausible.Plugs.RestictUserType,
       deny: :sso when action in [:update_name, :update_email, :update_password]
```

We could only support `deny` for a single type. _IF_ ever needed, support for `allow` option could be included meaning "allow only this type but deny all others".

lib/plausible/users.ex
zoldar: Nope, just checked `always` does not cut it as the smartass compiler checks the actual types returned https://github.com/plausible/analytics/actions/runs/15758442078/job/44419090993?pr=5506

lib/plausible_web/plugs/restrict_type.ex
zoldar: https://github.com/plausible/analytics/pull/5506/commits/e148969b3ff3a5e579157fa822d0f64a52eec243

lib/plausible_web/controllers/settings_controller.ex
zoldar: https://github.com/plausible/analytics/pull/5506/commits/2d9093fdf4149dbebb98311a319d9791a6a8243b

PR #5513, opened 2025-06-17
End-to-end sso domain verification

aerosol:
### Changes

This PR integrates SSO domain verification process into SSO management UI.
Best effort is made to keep background job and domain status in sync.

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible/auth/sso/domains.ex
```diff
@@ -100,6 +114,7 @@ defmodule Plausible.Auth.SSO.Domains do
           Repo.transaction(fn ->
             Enum.each(domain_users, &SSO.deprovision_user!/1)
             Repo.delete!(sso_domain)
+            :ok = SSO.Domain.Verification.Worker.cancel(sso_domain.domain)
```
zoldar: I think the same should be done for `SSO.remove_integration` as it removes all associated domains by cascade?

aerosol: https://github.com/plausible/analytics/pull/5513/commits/9fe16adf7f46f9f5e2221ba5a4bae2571113f4b2

aerosol:
@zoldar made things a little bit more concise with https://github.com/plausible/analytics/pull/5513/commits/92b91c301149a75e5002b1da4c6860cc0fdbed7b - apart from cosmetic changes, all API is provided through the `Domains` context, removing circular dependency.

extra/lib/plausible/auth/sso/domains.ex
```diff
@@ -17,6 +20,34 @@ defmodule Plausible.Auth.SSO.Domains do
...
+          {:ok, _} = Verification.Worker.enqueue(domain)
+          {:ok, sso_domain}
+        end
+      end)
+
+    result
+  end
+
+  @spec cancel_verification(String.t()) :: :ok
+  def cancel_verification(domain) when is_binary(domain) do
```
zoldar, line 38: `cancel_verification!`?

extra/lib/plausible_web/live/sso_management.ex
```diff
@@ -458,8 +486,10 @@ defmodule PlausibleWeb.Live.SSOManagement do
     {:noreply, socket}
   end
 
-  def handle_event("show-manage", _params, socket) do
-    {:noreply, route_mode(socket, :manage)}
+  def handle_event("verify-domain-submit", %{"domain" => domain}, socket) do
+    SSO.Domains.start_verification(domain)
```
zoldar: I'm wondering if it wouldn't make sense to introduce an env var like `SSO_DOMAIN_SKIP_CHECKS` and make `Verification.Worker.enqueue` pass `skip_checks` option on the basis of it? It would make testing stuff locally in dev env easier IMHO.

aerosol: Will https://github.com/plausible/analytics/pull/5513/commits/fff27fce79464b3e8b54277549a02d3bfac44318 work? Or would you like to still be able to skip through env var?

zoldar: Nice! That should work fine. Testing against more than one domain would require some additional fiddling still but that's going to be very rare anwyay I think.

extra/lib/plausible/auth/sso/domains.ex
aerosol: Hmm sure about that?

> A trailing bang (exclamation mark) signifies a function or macro where failure cases raise an exception.

Other than postgres pool unavailable, I don't think there's an exception ever thrown on failure. But in that sense any function with db side effects raises.

extra/lib/plausible_web/live/sso_management.ex
aerosol: Yeah, verification is one-off process most of the time, restarting the dns mock with different integration id shouldn't cause too much fiddling?

zoldar: I meant more than one domain at a time. Again, NBD, it's fine as is 👍

extra/lib/plausible/auth/sso/domains.ex
zoldar: Ah, ok, makes sense 👍

PR #5514, opened 2025-06-17
ScriptV2: `callback`, file downloads and `logging` changes

macobo:
### Changes

This PR resolves some open discussions from ScriptV2:
- Adds `logging` option
- Improve `callback` option, add tests
- Improve `fileDownloads` typing, export `DEFAULT_FILE_TYPES`

### DIFF
--- a/tracker/test/callbacks.spec.js
+++ b/tracker/test/callbacks.spec.js
@@ -1,0 +1,60 @@
+import { test, expect } from '@playwright/test'
+import { LOCAL_SERVER_ADDR } from './support/server'
+
+async function openPage(page, src, endpoint = `${LOCAL_SERVER_ADDR}/api/event`) {
+  await page.goto(`/callbacks.html?src=${src}&endpoint=${endpoint}`)
+}
+
+function trackWithCallback(page) {
+  return page.evaluate(() => window.callPlausible())
+}
+
+function testCallbacks(trackerScript, expectedNetworkError) {
+  const trackerScriptwithoutLocal = trackerScript.replace('.local', '')
+
+  test("on successful request", async ({ page }) => {
+    await openPage(page, trackerScript)
+
+    const callbackResult = await trackWithCallback(page)
+    expect(callbackResult).toEqual({ status: 202 })
+  })
+
+  test('on ignored request', async ({ page }) => {
+    await openPage(page, trackerScriptwithoutLocal)
+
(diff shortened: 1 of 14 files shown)


tracker/test/callbacks.spec.js
```diff
@@ -0,0 +1,60 @@
...
+async function openPage(page, src, endpoint = `${LOCAL_SERVER_ADDR}/api/event`) {
+  await page.goto(`/callbacks.html?src=${src}&endpoint=${endpoint}`)
+}
+
+function trackWithCallback(page) {
+  return page.evaluate(() => window.callPlausible())
+}
+
+function testCallbacks(trackerScript) {
+  const trackerScriptwithoutLocal = trackerScript.replace('.local', '')
```
RobertJoonas: Nit: since this is only used in one test. WDYT about moving it into that test, or even inline it:

`await openPage(page, trackerScript.replace('.local', ''))`.

Declaring it up here feels like it's supposed to be used by every test.

tracker/test/callbacks.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+async function openPage(page, src, endpoint = `${LOCAL_SERVER_ADDR}/api/event`) {
+  await page.goto(`/callbacks.html?src=${src}&endpoint=${endpoint}`)
+}
+
+function trackWithCallback(page) {
+  return page.evaluate(() => window.callPlausible())
+}
+
+function testCallbacks(trackerScript) {
```
RobertJoonas: Nit: renaming the argument `trackerScript` -> `trackerScriptSrc` would make the intent clear without having to jump to `callbacks.html` or scrolling down this file.

tracker/test/shared-configuration-tests.js
```diff
@@ -49,13 +50,36 @@ export function testPlausibleConfiguration({ openPage, initPlausible, fixtureNam
       })
     })
 
-    test('does not trigger any events when `local` config is disabled', async ({ page }) => {
+    test('does not trigger any events when `local` config is disabled but console.warns', async ({ page }) => {
```
RobertJoonas:
```suggestion
    test('does not trigger any events and console.warns when `local` config is disabled', async ({ page }) => {
```

RobertJoonas (approved):
A few nits inline. Other than that looks good to me 👍

PR #5520, opened 2025-06-23
Persist login type preference (SSO/standard)

aerosol:
### Changes

This PR makes "login form type" sticky upon first successful authentication.

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/controllers/sso_controller.ex
```diff
@@ -4,14 +4,23 @@ defmodule PlausibleWeb.SSOController do
...
   alias Plausible.Auth.SSO
+  alias PlausibleWeb.LoginPreference
 
   alias PlausibleWeb.Router.Helpers, as: Routes
 
   plug Plausible.Plugs.AuthorizeTeamAccess,
        [:owner] when action in [:sso_settings]
 
   def login_form(conn, params) do
-    render(conn, "login_form.html", error: params["error"])
```
aerosol, line 14: Not sure what was the `error: params["error"]` for here, but it allowed some (harmless?) injection. @zoldar

zoldar: It's for passing error on redirect to login form, like in https://github.com/plausible/analytics/blob/master/lib/plausible_web/user_auth.ex#L37-L72. Not the best idea though - I'll rewrite it to use flash messages mechanism instead.

lib/plausible_web/login_preference.ex
```diff
@@ -0,0 +1,35 @@
...
+  This module handles storing and retrieving the user's preferred login method
+  (standard or SSO) to provide a better user experience by showing their
+  preferred option first.
+  """
+
+  @cookie_name "login_preference"
+  @cookie_max_age 60 * 60 * 24 * 365
+
+  @spec set_preference(Plug.Conn.t(), String.t()) :: Plug.Conn.t()
+  def set_preference(conn, preference) when preference in ["standard", "sso"] do
```
zoldar: Perhaps, to avoid setting more cookies than absolutely necessary, we could assume standard is the default and delete it if preference is switched back to standard login?

aerosol: https://github.com/plausible/analytics/pull/5520/commits/6a6e14dbbf354ca9af48494548c8e8c3c861f86e

extra/lib/plausible_web/controllers/sso_controller.ex
zoldar: Login forms now use flash messages for error passing https://github.com/plausible/analytics/pull/5520/commits/e41d67acc03471b734b0535dc6d949c1a2a7210c

zoldar:
I have changed mock-dns port to 5354 and renamed the parameter from `integration_id` to `domain_id` in https://github.com/plausible/analytics/pull/5520/commits/a0329e70355f5ee2b46e7b0efa5e188eb33f8d6c

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -224,8 +226,27 @@ defmodule PlausibleWeb.AuthController do
...
   end
 
-  def login_form(conn, _params) do
-    render(conn, "login_form.html")
+  on_ee do
+    def login_form(conn, params) do
+      login_preference = LoginPreference.get(conn)
+      error = Phoenix.Flash.get(conn.assigns.flash, :login_error)
+
+      case {login_preference, params["prefer"], error} do
```
zoldar, line 234: Presence of error in flash implies the prior redirect to this particular view was intentional, so we skip redirect here. \cc @aerosol

zoldar (approved):

PR #5522, opened 2025-06-23
Implement plug for checking SSO team access

zoldar:
### Changes

This PR implements mechanism for blocking standard users from access to SSO team with "all_but_owners" policy set. Changes are described in more detail below:

- Provisioned team members are now explicitly labeled in team management view:

  image
- Once "Force SSO" is enabled, all non-SSO members are locked out of accessing the team until they log out and log in via SSO:

  image
- If non-SSO member does not meet criteria for provisioning, they are presented with appropriate issue description instead:

  image
  image
- When non-SSO member attempts to log in via SSO but does not meet the provisioning criteria, they are also presented with issue description tailored to that context:

  image
  image
- User enabling "Force SSO" must be logged in via SSO:

  image

### TODO

- [x] improve copy of provision notice and issue screens
- [x] consider adding flash error/warning when user logs in via SSO but provisioning fails
- [x] add warning or even block enabling "Force SSO" if the enabling owner does not meet criteria at the time

### Tests

### DIFF
--- a/extra/lib/plausible_web/templates/sso/provision_issue.html.heex
+++ b/extra/lib/plausible_web/templates/sso/provision_issue.html.heex
@@ -2,16 +2,23 @@
   <:title>
     Single Sign-On enforcement
   </:title>
+  <:subtitle>
+    The owner of <span :if={is_nil(@conn.assigns[:current_team])}>the team</span>
+    <span :if={@conn.assigns[:current_team]}>"{@conn.assigns[:current_team].name}"</span>
+    has turned off regular email and password logins.
+    To keep things secure and simple, you can only sign in using your organization's
+    Single Sign-On (SSO) system.
+  </:subtitle>
 
-  <p :if={@issue == :multiple_memberships}>
+  <p :if={@issue == :multiple_memberships} class="text-sm">
     To access this team, you must first leave all other teams.
   </p>
 
-  <p :if={@issue == :multiple_memberships_noforce}>
+  <p :if={@issue == :multiple_memberships_noforce} class="text-sm">
     To log in as an SSO user, you must first leave all other teams.
   </p>
 
-  <div :if={@issue == :active_personal_team} class="flex flex-col space-y-2">
+  <div :if={@issue == :active_personal_team} class="text-sm flex flex-col space-y-2">
     <p>
       To access this team, you must either remove or transfer all sites you own under "My Personal Sites".
     </p>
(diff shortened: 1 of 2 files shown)


lib/plausible_web/router.ex
```diff
@@ -12,9 +12,23 @@ defmodule PlausibleWeb.Router do
     plug PlausibleWeb.Plugs.NoRobots
     on_ee(do: nil, else: plug(PlausibleWeb.FirstLaunchPlug, redirect_to: "/register"))
     plug PlausibleWeb.AuthPlug
+    on_ee(do: plug(Plausible.Plugs.SSOTeamAccess))
     plug PlausibleWeb.Plugs.UserSessionTouch
   end
 
+  on_ee do
+    pipeline :browser_sso_notice do
```
zoldar, line 20: It's the same as `:browser` except `SSOTeamAccess` plug. Meant to using for "browser" routes which should remain accessible even if user is locked out of the team.

extra/lib/plausible_web/templates/sso/provision_issue.html.heex
```diff
@@ -0,0 +1,44 @@
+<.focus_box>
+  <:title>
+    Single Sign-On enforcement
+  </:title>
+
```
aerosol: Let's add a `:subtitle` section with something a bit more descriptive:

> The owner of this team has turned off regular email and password logins. To keep things secure and simple, you can only sign in using your organization's Single Sign-On (SSO) system.

We could consider including the team's name as well for clarity if that's not too difficult?

lib/plausible_web/live/team_management.ex
```diff
@@ -333,7 +333,12 @@ defmodule PlausibleWeb.Live.TeamManagement do
   defp entry_label(%Layout.Entry{role: :guest, type: :membership}, _), do: nil
   defp entry_label(%Layout.Entry{type: :invitation_pending}, _), do: "Invitation Pending"
   defp entry_label(%Layout.Entry{type: :invitation_sent}, _), do: "Invitation Sent"
+
+  defp entry_label(%Layout.Entry{meta: %{user: %{id: id, type: :sso}}}, %{id: id}),
+    do: "You (SSO)"
+
   defp entry_label(%Layout.Entry{meta: %{user: %{id: id}}}, %{id: id}), do: "You"
+  defp entry_label(%Layout.Entry{meta: %{user: %{type: :sso}}}, _), do: "SSO Member"
```
aerosol, line 341: :heart:

extra/lib/plausible_web/templates/sso/provision_issue.html.heex
zoldar: How about this:

image

and for logged out user, this:

image

https://github.com/plausible/analytics/pull/5522/commits/62fe5e8e1215346071551ae139be95bb327fa575

aerosol: Nice!

PR #5525, opened 2025-06-24
Implement automatic re-login on SSO session expiration

zoldar:
### Changes

This PR introduces a mechanism for automatically re-triggering login when SSO user's session expires. Given session timeout for SSO accounts is a lot shorter (on the order of hours) and is not prolonged with user activity, it should substantially reduce inconvenience to manually re-entering email every couple hours.

It's not bulletproof as it does not account for XHR internal API requests which will start returning errors, making stats dashboard stale until explicitly reloaded. Getting this aspect right is trickier though as we don't want to end up in unintentional redirect loop, which would be worse. Settling for tracking that for full page requests should be enough for now.

### TODO

- [x] tests
- [x] set minimum session length to 30 mintues to avoid abuse

### Tests

### DIFF
--- a/test/plausible_web/plugs/handle_expired_session_test.exs
+++ b/test/plausible_web/plugs/handle_expired_session_test.exs
@@ -1,0 +1,78 @@
+defmodule Plausible.Plugs.HandleExpiredSessionTest do
+  use PlausibleWeb.ConnCase, async: true
+  use Plausible
+
+  on_ee do
+    use Plausible.Teams.Test
+
+    alias Plausible.Auth.UserSessions
+    alias Plausible.Plugs.HandleExpiredSession
+    alias Plausible.Repo
+
+    alias PlausibleWeb.Router.Helpers, as: Routes
+
+    test "passes through when there's no expired_session" do
+      conn = HandleExpiredSession.call(build_conn(), [])
+      refute conn.halted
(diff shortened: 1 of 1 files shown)


extra/lib/plausible_web/plugs/handle_expired_session.ex
```diff
@@ -0,0 +1,64 @@
...
+
+  def call(conn, []) do
+    maybe_trigger_login(conn, conn.assigns[:expired_session])
+  end
+
+  defp maybe_trigger_login(conn, nil), do: conn
+
+  defp maybe_trigger_login(conn, user_session) do
+    if Plausible.Users.type(user_session.user) == :sso do
+      Plausible.Auth.UserSessions.revoke_by_id(user_session.user, user_session.id)
```
zoldar, line 24: Originally I left session behind but that was causing redirect loops due to this plug (and AuthPlug) getting plugged, well, everywhere, including login screens. We could make an exemption for SSO login form path to avoid that. 

At the same time though, that's going to leave a very noisy track record of expired sessions in the DB. The expired session cleanup is currently removing sessions which timed out +7 days ago.

When we eventually add audit logging and also track those SSO session expiration events, lack of retention shouldn't be as much of an issue? 🤔

PR #5532, opened 2025-06-26
ScriptV2: Remove hashBasedRouting from server-interpolated config

macobo:
New onboarding _does not_ have a checkbox for hash based routing, so remove it from server-interpolated config


ukutaht:
Yes, this will be docs-only

ukutaht (approved):

PR #5534, opened 2025-06-26
Recommend installation type

ukutaht:
Uses `Verification.Checks` in `InstallationV2` to determine the recommended installation type. Display recommended type in the `provisioning` flow.

Next steps:
* Domain change flow
* Tests


macobo (approved):
Makes sense to me, but perhaps @RobertJoonas has additional thoughts?

PR #5538, opened 2025-06-30
WIP Script v2: Handle links within svgs correctly

apata:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/custom-events.js
+++ b/tracker/src/custom-events.js
@@ -32,7 +32,7 @@ function handleLinkClickEvent(event) {
   if (event.type === 'auxclick' && event.button !== MIDDLE_MOUSE_BUTTON) { return }
 
   var link = getLinkEl(event.target)
-  var hrefWithoutQuery = link && link.href && link.href.split('?')[0]
+  var hrefWithoutQuery = link && link.href && typeof link.href === 'string' && link.href.split('?')[0]
 
   if (COMPILE_TAGGED_EVENTS) {
     if (isElementOrParentTagged(link, 0)) {
--- a/tracker/src/custom-events.js
+++ b/tracker/src/custom-events.js
@@ -22,17 +22,34 @@ function isLink(element) {
 function shouldFollowLink(event, link) {
   // If default has been prevented by an external script, Plausible should not intercept navigation.
   if (event.defaultPrevented) { return false }
-
-  var targetsCurrentWindow = !link.target || link.target.match(/^_(self|parent|top)$/i)
+  var target = link.getAttribute('target')
+  var targetsCurrentWindow = !target || target.match(/^_(self|parent|top)$/i)
   var isRegularClick = !(event.ctrlKey || event.metaKey || event.shiftKey) && event.type === 'click'
   return targetsCurrentWindow && isRegularClick
+}
+
+function getUnifiedUrl(link) {
+  var href
+  if (link && link.href) {
+    if (typeof link.href.animVal === 'string') {
+      href = link.href.animVal
+    } else if (typeof link.href === 'string') {
+      href = link.href
+    }
+  }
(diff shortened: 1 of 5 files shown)


tracker/src/custom-events.js
```diff
@@ -22,17 +22,34 @@ function isLink(element) {
...
+  var href
+  if (link && link.href) {
+    if (typeof link.href.animVal === 'string') {
+      href = link.href.animVal
+    } else if (typeof link.href === 'string') {
+      href = link.href
+    }
+  }
+  if (href) {
+    const u = new URL(l, document.baseURI)
```
macobo: This does not work on IE11/compat: https://caniuse.com/url

Why do we need `new URL` here in the first place?

Could this function be:

```javascript
function getHref(link) {
  if (link && link.href) {
    return link.href.animVal || link.href
  }
}
```

And we keep the rest as it was?

apata: Thanks for the comment! I tried that, but .host attribute is not available on <a> elements within svgs, so we'd need to figure that out as well. I went for new URL API for it, checking if there's an IE compatible way without manual splitting.

macobo: Why do we need the host for the svgs? The previous code was not using it either, it worked off purely the href.

apata: It's used in isOutboundLink function on main branch https://github.com/plausible/analytics/blob/528693889d1b9781ba5912ec54356a72dacce5c1/tracker/src/custom-events.js#L96

tracker/src/custom-events.js
```diff
@@ -115,6 +115,21 @@ function isTagged(element) {
...
 }
 
+function normalizeHref(link) {
+  if (link && link.href) {
+    /** If the link is an SVG element, href is an object @see SVGAnimatedString. */
+    if (link.href.animVal) {
+      // Create a temporary anchor element to normalize reading the href attribute
+      var a = document.createElement('a')
+      a.setAttribute('href', link.href.animVal)
+      return a.href
```
apata: This hack is needed because
Screenshot 2025-06-30 at 16 22 51
automatic `href` transform does not work when the `a` element is within an svg.

apata: not needed any more because not tackling invalid url value in this PR

tracker/src/custom-events.js
```diff
@@ -19,20 +19,23 @@ function isLink(element) {
...
+
+  return true;
 }
 
 function handleLinkClickEvent(event) {
   if (event.type === 'auxclick' && event.button !== MIDDLE_MOUSE_BUTTON) { return }
 
   var link = getLinkEl(event.target)
-  var hrefWithoutQuery = link && link.href && link.href.split('?')[0]
+  var hrefWithoutQuery = link && typeof link.href === 'string' && link.href.split('?')[0]
```
ukutaht, line 38: I would check if extracting a helper like

```javascript
function isLinkWithHref(link) {
  return link && typeof link.href === 'string'
}
}
```

would reduce the final bundle size by a few bytes because this check is used in 2 places.

apata: Thanks! I checked and it actually increases bundle size. The reason seems to be that this necessitates the declaration `function k(e) return e&&"string"==typeof e.href` in the bundle, which contains inevitable JS words `function` and `return`, and these tip the scales in favor of inline. 

For this helper, we'd probably start to see size benefits if it were used 4 places. 

With helper:
Screenshot 2025-07-29 at 19 20 00

Without helper: 
Screenshot 2025-07-29 at 19 14 58

apata:
Included rewrite of some flaky tests. Moved feature specific v2 tests to feature specs. Apologies for the huge diff.

PR #5541, opened 2025-07-01
Stop segments tooltip cutoff

apata:
### Changes

Renders tooltips in searchable segment menu outside the scrollable container using createPortal to stop them from being cut off.

### Tests
- [x] Manual tests Chrome, Firefox, Safari, Edge, mobile instances

### Changelog

### Documentation

### Dark mode


apata:
> Approach looks good to me.
> 
> I'm approving it but just offering an alternative for consideration: For tooltips it may be appropriate to use `document.body` as the portal target for all tooltips. This way:
> 
> 1. The Tooltip component API stays a bit simpler (no `containerRef`)
> 2. Tooltip logic is more straightforward with less conditionals
> 
> Happy for you to ignore the suggestion if there's some downsides I haven't thought about.

Thanks! I mostly wanted to patch the ugly visual issue quickly without checking / regression in the other use cases. Document.body definitely works, but such tooltips would then be outside headless ui dropmenu tree (at least in this case), so it might have not played well with headless ui click and focus logic.

PR #5542, opened 2025-07-01
Update ignored routes in OpenTelemetry sampler to include additional variants

cnkk:


aerosol:
But why and why not `/api//event/` etc.?

cnkk:
In Honeycomb, I saw that we have requests for `/api/event`, even though these should be ignored.
After further analysis, I found that other HTTP targets were being used. [See honeycomb](https://ui.honeycomb.io/plausible/datasets/plausible-prod?query=%7B%22time_range%22%3A2419200%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%22http.route%22%2C%22http.target%22%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22COUNT%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22http.route%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%22%2Fapi%2Fevent%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%7B%22op%22%3A%22COUNT%22%2C%22order%22%3A%22descending%22%7D%5D%2C%22havings%22%3A%5B%5D%2C%22trace_joins%22%3A%5B%5D%2C%22limit%22%3A1000%2C%22derived%22%3A%5B%5D%7D&vs=hideCompare).

`/api//event/` has not been used in the last 60 days; only the added targets have been used.

PR #5546, opened 2025-07-02
Script v2: Do not send Form: Submission events for tagged forms

apata:
### Changes

- Stops duplicate events if formSubmissions tracking is enabled and the form is tagged.

- Also removes some noise from formSubmissions tests

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/test/features-hierarchy-on-overlap.spec.ts
+++ b/tracker/test/features-hierarchy-on-overlap.spec.ts
@@ -10,6 +10,7 @@ import { LOCAL_SERVER_ADDR } from './support/server'
 import { initializePageDynamically } from './support/initialize-page-dynamically'
 import { mockManyRequests } from './support/mock-many-requests'
 import { ScriptConfig } from './support/types'
+import { customSubmitHandlerStub } from './support/html-fixtures'
 
 const DEFAULT_CONFIG: ScriptConfig = {
   domain: 'example.com',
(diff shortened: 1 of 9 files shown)


tracker/test/features-hierarchy-on-overlap.spec.ts
```diff
@@ -149,3 +155,100 @@ for (const mode of ['legacy', 'web']) {
...
+              init(${JSON.stringify(config)})
+              window.plausible = { l: true }
+            </script>
+          `
+          },
+          mode
+        ),
+        bodyContent: `
+          <div class="plausible-event-name--A+Tagged+Form">
+            <form onsubmit="event.preventDefault(); console.log('Form submitted');">
```
RobertJoonas: Leftover log?

```suggestion
            <form>
```

RobertJoonas (approved):
One comment inline about a dangling log, but otherwise looks good to me! 👍

tracker/test/features-hierarchy-on-overlap.spec.ts
apata: It's deliberate, but debatable if it's good. 

It's supposed to represent a custom form onsubmit handler (using fetch etc): https://github.com/plausible/analytics/blob/master/tracker/test/form-submissions.spec.ts#L309. 

I've duplicated the code and thus the context from the function doc is lost. I should export it and use it in both test suites.

PR #5549, opened 2025-07-02
Verification v2

RobertJoonas:
### Changes

WIP - objectives:

- [ ] `verifier.js` is written in Vanilla JS (called by Puppeteer wrapper via `page.evaluate()`) and can be tested with a thin Playwright wrapper as well
- [ ] `verifier.js` verifies `plausibleInstalled` (= `window.plausible` exists) and returns `callBackStatus` (response to the test event) + tests
- [ ] `verifier.js` detects `dataDomainMismatch` + tests (*EXPERIMENTAL)
- [ ] `verifier.js` detects WP, WP plugin, GTM, and known cookiebots (*EXPERIMENTAL)
- [ ] `verifier.js` detects `proxyLikely` + tests (*EXPERIMENTAL)
- [ ] `verifier.js` detects `manualExtension` + tests (*EXPERIMENTAL)
- [ ] `verifier.js` detects `snippetUnknownAttributes` + tests (*EXPERIMENTAL)

_*EXPERIMENTAL - we won't use it yet, but will collect telemetry, trying to prove that JS is able to capture the same information. A follow-up PR will clean up Checks.Snippet._

### Tests

### Changelog

### Documentation

### Dark mode


tracker/test/verifier/verifier-v1.spec.js
```diff
@@ -0,0 +1,474 @@
...
+    await mockEventResponseSuccess(page)
+
+    const { url } = await initializePageDynamically(page, {
+      testId,
+      scriptConfig: `<script defer data-domain="www.right.com" src="/tracker/js/plausible.local.js"></script>`
+    })
+
+    const result = await verify(page, {url: url, expectedDataDomain: 'right.com'})
+
+    expect(result.data.dataDomainMismatch).toBe(false)
```
ukutaht, line 236: Hmm, I would think that including `www` in the `data-domain` is a user error and we need to notify them of that. Am I missing something?

ukutaht (approved):
Looks good overall. Interested in what @apata thinks as well

tracker/test/verifier/verifier-v1.spec.js
RobertJoonas: Ah, failed to mention this so I understand the confusion. Normally we'd detect a `dataDomainMismatch` with simply `snippet[data-domain] != expectedDataDomain`, but our ingestion actually strips any trailing `www.` from the `data-domain` attribute value. Code ref:
https://github.com/plausible/analytics/blob/73c46c6f4c08123a1a45cd0c6fd3642f74eec1e3/lib/plausible/ingestion/request.ex#L186

So even if the snippet `data-domain` attribute has `www.` in it while `site.domain` doesn't, we don't consider that a mismatch.

The current `Checks.Snippet` behaviour covers this case too, a bit indirectly though - `data-domain` is checked against a "final_domain" from the `FetchBody` check. The `final_domain` is the site hostname after following redirects. So basically, currently we're only allowing `www.` in data-domain ONLY IF it's in the actual hostname of the site.

tracker/test/verifier/verifier-v1.spec.js
```diff
@@ -0,0 +1,474 @@
...
+      `
+    })
+
+    const result = await verify(page, {url: url, expectedDataDomain: SOME_DOMAIN})
+
+    expect(result.data.unknownAttributes).toBe(true)
+  })
+})
+
+test.describe('v1 verifier (logging)', () => {
```
apata, line 475: suggestion, non-blocking: We could be more confident in the correctness of the verifier logging logic if we state exactly what we expect console contents to be, including error, debug, warning messages.
Example 
```ts
const messages = []
page.context().on('console', logs.push({type: msg.type(), text: msg.text()}))
// ... action under test
expect(consoleMessages).toEqual([
  {type: "log", message: expect.stringContaining("Plausible Verification] Starting snippet detection"))},
  // ...other messages
])

```

tracker/verifier/verifier-v1.js
```diff
@@ -0,0 +1,70 @@
...
+    cookieBannerLikely = checkCookieBanner()
+  }
+
+  log(`cookieBannerLikely: ${cookieBannerLikely}`)
+
+  return {
+    data: {
+      completed: true,
+      plausibleInstalled: plausibleFunctionDiagnostics.plausibleInstalled,
+      callbackStatus: plausibleFunctionDiagnostics.callbackStatus || 0,
```
apata, line 57: nitpick, non-blocking: it seems we could use `callbackStatus` constant here, it already has been set to `|| 0`

tracker/verifier/check-cookie-banner.js
```diff
@@ -0,0 +1,35 @@
+const SELECTORS = {
+  // https://github.com/cavi-au/Consent-O-Matic/blob/master/rules/cookiebot.json
```
apata, line 2: nitpick, non-blocking: I would like the comment to explain what values we're syncing from there and why, so when that repo structure changes we could still figure it out.

tracker/test/check-unknown-attributes.test.js
```diff
@@ -0,0 +1,114 @@
+import { test, expect } from '@playwright/test'
```
apata: q: Why is this suite outside /tracker/test/verifier folder? Do we need to run it for all browsers?

tracker/playwright.config.js
```diff
@@ -25,11 +25,13 @@ export default defineConfig({
     {
       name: 'firefox',
       use: { ...devices['Desktop Firefox'] },
+      testIgnore: 'test/verifier/**',
```
apata, line 28: suggestion, non-blocking: can we comment here why we're not running verifier tests with this browser?

tracker/test/support/initialize-page-dynamically.ts
```diff
@@ -12,6 +12,13 @@ interface DynamicPageOptions {
   path?: string
 }
 
+interface DynamicPageOptionsFullPage {
```
apata: nitpick, non-blocking: by repeating part of the interface, we're losing the JSDoc comment on the `path?: string` when the type is inferred to be DynamicPageOptionsFullPage. This makes the IDE and probably AI code agents less helpful. There's various ways to share these, for example: 

```ts
interface SharedOptions {
  /** unique ID that becomes part of the dynamic page URL */
  testId: string
  /** optional path to append to the dynamic page URL */
  path?: string
}

interface TemplatedResponse {
  /** string like `<script defer id="plausible" src="/plausible.compat.local.js"></script>` or ScriptConfig to be set to web snippet */
  scriptConfig: ScriptConfig | string
  /** vanilla HTML string, which can contain JS, will be set in the body of the page */
  bodyContent: string
}

interface FullResponse {
  response: string
}

export async function initializePageDynamically(
  page: Page,
  {
    testId,
    path = '',
    ...options
  }: SharedOptions & (TemplatedResponse | FullResponse)
): Promise<DynamicPageInfo> {
    const url = `/dynamic/${testId}${path}`
    await page.context().route(url, async (route) => {
      let responseBody: string
      if ('response' in options) {
        // ...
}
```

tracker/verifier/snippet-checks.js
```diff
@@ -0,0 +1,64 @@
+import { runThrottledCheck } from "./run-check"
+
+export async function waitForSnippetsV1(log) {
+  log('Starting snippet detection...')
+
+  let snippetCounts = await waitForFirstSnippet()
+
+  if (snippetCounts.all > 0) {
+    log(`Found snippets: head=${snippetCounts.head}; body=${snippetCounts.body}`)
```
apata, line 9: ultranitpick, non-blocking: there's a semicolon between the two values but in the next logs, there's a comma

lib/plausible/verification/checks/installation.ex
```diff
@@ -1,34 +1,81 @@
...
+    }
+  }
+  """
 
   @moduledoc """
   Calls the browserless.io service (local instance can be spawned with `make browserless`)
-  and runs #{@verification_script_filename} via the [function API](https://docs.browserless.io/HTTP-APIs/function).
+  and runs verifier script via the [function API](https://docs.browserless.io/HTTP-APIs/function).
+
+  The verification uses a vanilla JS script that runs in the browser context,
```
apata, line 36: Really helpful extended comment! 💯

lib/plausible/verification/checks/installation.ex
```diff
@@ -1,34 +1,81 @@
...
+
+  - `data.plausibleInstalled` - whether or not the `plausible()` window function was found
+
+  - `data.callbackStatus` - integer. 202 indicates that the server acknowledged the test event.
+                            The test event ingestion is discarded based on user-agent, see:
+                            `Plausible.Verification.user_agent/0`
+
+  - `data.dataDomainMismatch` - whether or not script[data-domain] mismatched with site.domain
+
+  - `data.proxyLikely` - whether the script[src] is not a plausible.io URL
```
apata, line 52: nitpick, non-blocking:  I think at this level, we can keep it opaque how proxyLikely is exactly determined. Also, this comment does not explain the meaning of `proxyLikely` when there's multiple snippets (the actual condition is `snippets.some`). How's this?
```
  - `data.proxyLikely` - whether for some of the snippets, the script might be proxied
```

tracker/test/verifier/verifier-v1.spec.js
```diff
@@ -0,0 +1,474 @@
...
+
+    await route.fulfill({
+      status: 202,
+      contentType: 'text/plain',
+      body: 'ok'
+    })
+  })
+}
+
+test.describe('v1 verifier (basic diagnostics)', () => {
```
apata, line 23: suggestion, non-blocking: Can we add a test for opening a page where the user has copied the snippet from installation code box to document head? The one that dynamically adds the script

RobertJoonas:
> We should probably mention it in tracker architecture.md as well, but that's totally non-blocking.

Good call @apata! Adding a reminder to do so in the next PR since even the JS verifier code structure is about to change.

tracker/test/verifier/verifier-v1.spec.js
RobertJoonas: The logic is meant more for internal debugging. Changing a log message or two doesn't affect anything that works in prod, so I think it's fine to change one of the console logs and see these tests still passing.

tracker/verifier/verifier-v1.js
RobertJoonas: I agree that with the recent script updates this should work, but I decided to leave the behaviour as is for now. There's some tiny risk of missing something, and we'l get rid of this code anyway with the release of script v2.

tracker/playwright.config.js
RobertJoonas: Noted. Coming up in the next refactoring PR as well. The Browserless function API calls the function on Chrome, so there's no need to test this in multiple envs. Only Chrome is completely enough.

lib/plausible/verification/checks/installation.ex
RobertJoonas: Let's not bother 😀 As I mentioned, this is expected to go away soon.

tracker/verifier/check-cookie-banner.js
RobertJoonas: https://github.com/plausible/analytics/pull/5549/commits/850c8d63ca31f6ba2f89394cd03c0e011ce34142

tracker/test/check-unknown-attributes.test.js
RobertJoonas: Good catch, thanks! Shouldn't be. https://github.com/plausible/analytics/pull/5549/commits/35d983f730b26b8a575ffefe81e7bf07a2fccb3e

tracker/test/support/initialize-page-dynamically.ts
RobertJoonas: https://github.com/plausible/analytics/pull/5549/commits/f42852e951cb706948363fe45572a0db6a108a64

tracker/verifier/snippet-checks.js
RobertJoonas: https://github.com/plausible/analytics/pull/5549/commits/1bbf760896d934df01549cf09635a8b2415a6b9d

tracker/test/verifier/verifier-v1.spec.js
RobertJoonas: Can't believe I missed this. Kind of the point we started this 😀 https://github.com/plausible/analytics/pull/5549/commits/e429edf09f71bf8fe272dd390432c2f37c8537c2

PR #5555, opened 2025-07-07
Implement ability to leave team

zoldar:
### Changes

This PR adds ability to leave current team via Team Settings > Leave Team.

image

image

image

image

image

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/settings/danger_zone.html.heex
+++ b/lib/plausible_web/templates/settings/danger_zone.html.heex
@@ -32,7 +32,7 @@
           method="delete"
           theme="danger"
         >
-          Delete my account
+          Delete My Account
         </.button_link>
     <% end %>
   </.tile>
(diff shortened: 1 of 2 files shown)


ukutaht:
Do I understand correctly that viewers and editors will not be able to leave team because they don't have access to the relevant settings are?

zoldar:
@ukutaht No, Team Settings are visible to all members but team management part is basically "read-only" for members without privileges.

ukutaht:
@zoldar Gotcha, thanks :)

aerosol (approved):
Was wondering if we should place in Danger Zone instead maybe? Other than that, LGTM.

zoldar:
@aerosol I thought about it but decided against it as it's not something that _endangers_/affects the team, only the current user. It's also somewhat part of team membership management. Does that make sense?

PR #5559, opened 2025-07-08
Script v2: Sync event.props.path for special path-based events from event.pathname

apata:
### Changes
If this PR is merged, then special path-based system goals (new term / module, name up for debate) won't need to have event.props.path defined during tracking requests. 

These goals/events are "WP Form Completions", "Form: Submission", "404".

It will instead be set on receiving the event to be the same as event.pathname, unless event.props.path is explicitly defined. 

We don't override it when it is explicitly defined to remain backwards compatible with users that are already sending these events with a particular value. 

From our side, we can't distinguish intentional values and unintentional values for event.props.path, so the benefits of the sync will mostly apply to new customers, or customers that begin using these features after this PR is merged.

This approach stores the same pathname twice on our side for special path-based events, but the rate of these special events is many magnitudes less than rate of other events, so the duplicated data is insignificant in the grand scheme of things. 

### Todo after merge
- [ ] Make sure 404 snippet in docs is accurate https://plausible.io/docs/error-pages-tracking-404#step-3-paste-this-piece-of-code-to-your-404-page-template
- [ ] Remove unnecessary event.props.path from form submissions logic in current Wordpress Plausible plugin

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/goals/system_goals.ex
+++ b/lib/plausible/goals/system_goals.ex
@@ -19,14 +19,17 @@ defmodule Plausible.Goals.SystemGoals do
     @goals_with_path
   end
 
+  @spec special_goals_for(String.t()) :: [String.t()]
   def special_goals_for("event:props:url"), do: goals_with_url()
   def special_goals_for("event:props:path"), do: goals_with_path()
 
+  @spec maybe_sync_props_path_with_pathname(String.t(), map()) ::
+          {:ok, map()} | {:error, String.t()}
   def maybe_sync_props_path_with_pathname(_pathname, %{"path" => _} = _props) do
     {:error, "Path has been already set in props, won't override"}
   end
 
-  def maybe_sync_props_path_with_pathname(pathname, props) do
+  def maybe_sync_props_path_with_pathname(pathname, %{} = props) do
     {:ok, Map.merge(props, %{"path" => pathname})}
   end
 end
--- a/lib/plausible/goals/system_goals.ex
+++ b/lib/plausible/goals/system_goals.ex
@@ -1,0 +1,32 @@
+defmodule Plausible.Goals.SystemGoals do
+  @moduledoc """
+  This module contains logic for special goals
+  """
+
+  # Special system goals which can be filtered by 'url' custom property
+  @goals_with_url ["Outbound Link: Click", "Cloaked Link: Click", "File Download"]
+
+  # Special system goals which can be filtered by 'path' custom property
+  @goals_with_path ["404", "WP Form Completions", "Form: Submission"]
+
+  @spec goals_with_url() :: [String.t()]
+  def goals_with_url() do
+    @goals_with_url
+  end
+
+  @spec goals_with_path() :: [String.t()]
+  def goals_with_path() do
+    @goals_with_path
+  end
+
+  def special_goals_for("event:props:url"), do: goals_with_url()
+  def special_goals_for("event:props:path"), do: goals_with_path()
+
+  def maybe_sync_props_path_with_pathname(_pathname, %{"path" => _} = _props) do
+    {:error, "Path has been already set in props, won't override"}
+  end
+
+  def maybe_sync_props_path_with_pathname(pathname, props) do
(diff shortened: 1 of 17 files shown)


lib/plausible/ingestion/request.ex
```diff
@@ -171,6 +172,20 @@ defmodule Plausible.Ingestion.Request do
...
 
+  defp maybe_put_props_path(changeset) do
+    with true <-
+           Changeset.get_field(changeset, :event_name) in Plausible.Goals.SystemGoals.goals_with_path(),
+         {:ok, props} <-
+           Plausible.Goals.SystemGoals.maybe_sync_props_path_with_pathname(
+             Changeset.get_field(changeset, :pathname),
+             Changeset.get_field(changeset, :props)
+           ) do
+      Changeset.put_change(changeset, :props, props)
```
apata: This technically allows storing a 31st prop for these special events, maybe need to validate again? 
```ex
%{
  #...up to 30 other props,
  "path" => pathname #  Note: max length of this prop is not enforced except indirectly through max url length
}
```

lib/plausible/goals/system_goals.ex
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  @spec special_goals_for(String.t()) :: [String.t()]
+  def special_goals_for("event:props:url"), do: goals_with_url()
+  def special_goals_for("event:props:path"), do: goals_with_path()
+
+  @spec maybe_sync_props_path_with_pathname(String.t(), map()) ::
+          {:ok, map()} | {:error, String.t()}
+  def maybe_sync_props_path_with_pathname(_pathname, %{"path" => _} = _props) do
+    {:error, "Path has been already set in props, won't override"}
```
RobertJoonas: Nit: using atom as the reason would be more idiomatic.

```suggestion
    {:error, :props_path_already_set}
```

lib/plausible/ingestion/request.ex
RobertJoonas: Yes, we should validate that there are 30 props max. Perhaps let's move `put_pathname` above `put_props`, and then handle the special path prop case in `put_props` itself?

test/support/factory.ex
```diff
@@ -149,6 +157,21 @@ defmodule Plausible.Factory do
     |> evaluate_lazy_attributes()
   end
 
+  defp maybe_sync_pathname_to_meta_for_special_goal(attrs, pathname) do
```
RobertJoonas: I think `event_factory` should be better off without business logic - the less surprises the better. If in a specific test, the event requires `props.path`, we would give it the `props.path` directly. If it needs both `pathname` and `props.path` we can also give it both.

tracker/test/form-submissions.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { test } from '@playwright/test'
 import { LOCAL_SERVER_ADDR } from './support/server'
 import {
+  e,
```
RobertJoonas, line 4: First time seeing this. Could we rename this to something like `expectEventAttr`?

test/support/factory.ex
apata: Thanks! I was on the fence about it and will remove! https://github.com/plausible/analytics/pull/5559/commits/750cdec107306c407de9f4f0b19a156d9cde0ce5 I just thought it might be inconvenient to set both with the current factory: we'd have to declare pathname twice.
```ex
build(
  :event, 
  
  pathname: "/blog/some-long-name...", 
  "meta.key": ["path"], 
  "meta.value": ["/blog/some-long-name..."]
)
```

lib/plausible/goals/system_goals.ex
apata: Thanks! I refactored this, but roger that on the :error, atom pattern!

lib/plausible/ingestion/request.ex
apata: Good shout, did so in  https://github.com/plausible/analytics/pull/5559/commits/69e5af5daace8b4c28844f01645ac353785b9aa9!

PR #5560, opened 2025-07-08
Audit trail for SSO 

aerosol:
### Changes

This PR adds basic audit trail for SSO-related changes. It _does not_ instrument any interactions with IDPs for instance.

Since audit trail is SSO-scoped for now, EE scope is also implied.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/audit.ex
```diff
@@ -0,0 +1,217 @@
...
+defimpl Plausible.Audit.Encoder, for: Map do
+  def encode(x) do
+    {allow_not_loaded, data} = Map.pop(x, :__allow_not_loaded__)
+
+    Enum.reduce(data, %{}, fn
+      {k, %Ecto.Association.NotLoaded{}}, acc ->
+        if k in allow_not_loaded do
+          acc
+        else
+          raise "#{k} association not loaded. Either preload, exclude or mark it as :optional in #{__MODULE__} options"
```
aerosol:
```suggestion
          raise "#{k} association not loaded. Either preload, exclude or mark it as :allow_not_loaded in #{__MODULE__} options"
```

test/plausible/audit_test.exs
```diff
@@ -0,0 +1,274 @@
...
+               %Plausible.Audit.Entry{
+                 name: "user_update",
+                 change: %{
+                   "after" => %{"name" => "John Doe"},
+                   "before" => %{
+                     "name" => "Jane Smith"
+                   }
+                 }
+               }
+             ] = Audit.list_entries(entity: "Plausible.Auth.User", entity_id: "#{user.id}")
```
ukutaht, line 221: Not a fan of coupling the audit log to module names like `Plausible.Auth.User`. I get that it's convenient but `entity: "user"` would IMO be more stable (=less likely to change) and useful.

extra/lib/plausible/auth/sso/domains.ex
```diff
@@ -175,14 +175,14 @@ defmodule Plausible.Auth.SSO.Domains do
   def mark_verified!(sso_domain, method, now \\ NaiveDateTime.utc_now(:second)) do
     sso_domain
     |> SSO.Domain.verified_changeset(method, now)
-    |> Repo.update!()
+    |> Repo.update_with_audit!("sso_domain_verified")
```
ukutaht: Hmm would it make sense to separate user vs system actions for domain verification?

Something like:
1. User action - `start_domain_verification` or similar
2. System action - `domain_verification_succeeded` or `domain_verification_failed`

test/plausible/audit_test.exs
```diff
@@ -0,0 +1,274 @@
...
+        password_confirmation: "very-secret-and-very-long-123"
+      })
+
+    {:ok, %Plausible.Auth.User{name: "Jane Doe"} = user} =
+      Repo.insert_with_audit(changeset, "user_insert")
+
+    assert [
+             %Plausible.Audit.Entry{
+               name: "user_insert",
+               change: %{}
```
ukutaht: Are the struct fields not encoded in the log entry for inserts?

I'm asking because if true, it feels a bit inconsistent since in the updates we do log the actual fields. If we log the before and after `name` field in a user update, why not log the `name` in the insert log as well for consistency?

test/plausible/audit_test.exs
aerosol: And where would you pass that entity name for initial collection? Another argument to `Repo.insert_with_audit`? It's specifically database entity coupled with database ID, always.

extra/lib/plausible/auth/sso/domains.ex
aerosol: I would like to avoid instrumenting UI with audit calls. I think we could utilize the `Entry.meta` field for that - and have context set specifically for cron-jobs for example. WYT? As for different event names, that's a though one if we like to stick instrumentation at context modules (such as `Plausible.Auth.SSO.Domains`).

test/plausible/audit_test.exs
aerosol: They are encoded, the test is sloppy

test/plausible/audit_test.exs
aerosol: In other initial attempts I was doing something like `Module.split(schema) |> List.last()` but that's not great for resources like `SSO.Integration` that would then become vaguely named `Integration`.

aerosol: I could infer raw table name if you like - e.g. `users` in this case.

zoldar: Perhaps `entity_name` could be part of `Encoder` protocol? 🤔

extra/lib/plausible/auth/sso/domains.ex
zoldar: Wouldn't presence (or lack of) `user_id` in event context be enough of a hint to tell apart user actions from system events? 🤔

extra/lib/plausible/auth/sso/saml_config.ex
```diff
@@ -21,6 +21,8 @@ defmodule Plausible.Auth.SSO.SAMLConfig do
   @fields [:idp_signin_url, :idp_entity_id, :idp_cert_pem, :idp_metadata]
   @required_fields @fields -- [:idp_metadata]
 
+  @derive {Plausible.Audit.Encoder, only: [:id, :idp_signin_url, :idp_entity_id, :idp_metadata]}
```
zoldar:
```suggestion
  @derive {Plausible.Audit.Encoder, only: [:id, :idp_signin_url, :idp_entity_id, :idp_cert_pem, :idp_metadata]}
```
?

lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,119 @@
...
+        data
+
+      {0, n} when n > 0 ->
+        changes
+
+      {0, 0} ->
+        %{}
+
+      _ ->
+        %{before: data, after: changes}
```
zoldar: Would it make sense to limit "data" only to keys which occur in "changes"? Something like 

```elixir
Map.take(data, Map.keys(changes))
```

?

extra/lib/plausible/auth/sso/domains.ex
aerosol: Yes, we could leverage the lack of user_id, but then.. what is this for @ukutaht? I'm not sure what's the semantic difference between verification started and then verification concluded. Do we need this separation for audit purposes?

lib/plausible/audit/encoder.ex
aerosol: Nice, good idea!

extra/lib/plausible/auth/sso/saml_config.ex
aerosol: Yeah I guess it won't hurt.

extra/lib/plausible/auth/sso/domains.ex
ukutaht: I'll take an actual user flow as an example why I made these suggestions.

Let's say owner 1 adds a domain, then owner 2 kicks off the verification, and after `@max_snoozes` we find the domain could not be verified.

Running through this locally I get this audit trail (I've lowered `@max_snoozes` to reach the end state faster):

```
plausible_dev=# select name, user_id, datetime from audit_entries order by datetime;
         name          | user_id |          datetime
-----------------------+---------+----------------------------
 sso_domain_added      |       1 | 2025-07-16 11:54:20.264949
 sso_domain_unverified |       2 | 2025-07-16 11:54:37.246405
 sso_domain_unverified |         | 2025-07-16 11:54:37.832768
 sso_domain_unverified |         | 2025-07-16 11:54:53.736405
 sso_domain_unverified |         | 2025-07-16 11:55:25.207941
```

IMO it's quite hard to look at this audit trail and reconstruct what was done by whom at what point. You'd need to understand that the first `sso_domain_unverified` has the id of the user who kicked off the verification process. Each retry is logged which adds unnecessary noise. It's also unclear whether the background job is still ongoing or finished.

I'd find something like this easier to make sense of for both CS and customers:

```
plausible_dev=# select name, user_id, datetime from audit_entries order by datetime;
         name          | user_id |          datetime
-----------------------+---------+----------------------------
 sso_domain_added      |       1 | 2025-07-16 11:54:20.264949
 sso_domain_verification_started |       2 | 2025-07-16 11:54:37.246405
 sso_domain_verification_failed |     {system}    | 2025-07-16 11:55:25.207941
```

Since getting the domain verification result can take 34h in the background, I think it needs to be a separate  action where the actor is `system` not a user. Otherwise let's say we filter the audit log by user id - we would get the false impression that the user was messing around with SSO settings up to 34h after they have ended their actual session. I think when an action has `user_id` it should mean that the user was actively doing something, hence background jobs should not emit logs with user_id.

I'm suggesting to add sso_domain_verification_started for completeness, since it can be done independently from sso_domain_added by a different user at a different time. Having timestamped the start of the verification process also gives us an idea how long verification result took.

WDYT, does this make sense?

test/plausible/audit_test.exs
aerosol: @zoldar what do you mean by that?

lib/plausible/audit/encoder.ex
aerosol: ed8a36ec58

extra/lib/plausible/auth/sso/domains.ex
aerosol: > Each retry is logged which adds unnecessary noise.

@ukutaht By that, do you mean  we should only log either success or last job attempt before giving up?

> WDYT, does this make sense?

Thanks for elaborating. I see your point and will try to adapt that specific case so it's more compact and not too coupled with implementation details.

aerosol: @ukutaht here's my attempt at addressing your comments: 23f9771ca8 - let me know if that's any better, in the meantime I'll improve coverage and rest of this PR. Will rebase later, not to make anyone lose their review context.

test/plausible/audit_test.exs
aerosol: @ukutaht would you mind addressing my ideas here? I've hardened the tests with `entity_id` check at https://github.com/plausible/analytics/pull/5560/commits/de111c5c2bbb3dd5ede9dee31ecb667f80abc6a2 - would like to also include `entity` assertion once we settle on what we like to store there.

test/plausible/audit_test.exs
aerosol: Double checked, they were not encoded, so a good call. On it.

lib/plausible/audit/encoder.ex
aerosol: This breaks encoding's derive, so that fields like password hash get logged. Looking to revert unless I find a nicer solution

test/plausible/audit_test.exs
aerosol: 0068546c6f the cleanest I can do

lib/plausible/audit/encoder.ex
aerosol: Had to revert it - 0068546c6f

extra/lib/plausible/auth/sso/domains.ex
ukutaht: @aerosol very nice 👌

test/plausible/audit_test.exs
ukutaht: Hmm yeah I suppose passing it manually every time is annoying and error-prone.

I think I prefer the DB table name over Elixir module name but I don't have strong feelings about it.

aerosol: Not having to maintain any mappings makes any further migrations/trail processing easier since we can do `String.to_existing_atom(schema)` and be done with it. Of course that won't work if any of the module names changes in the meantime, but what are the odds?

@zoldar any preference here?

aerosol: I prefer to leave it. @cnkk included some payloads in https://github.com/plausible/knowledge_base/blob/7ec369a2bc527de03b8c4af4b37b079d764ccd63/src/planning/rfcs/0012-audit-logging.md and it looks like "target.type" is something closer to actual schema module rather than anything else. Not sure how closely do we need to align with that at this point.

aerosol:
~~Welp. Don't understand https://github.com/plausible/analytics/actions/runs/16437080836/job/46449083748?pr=5560~~

extra/lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,120 @@
...
+
+defimpl Plausible.Audit.Encoder, for: [Integer, BitString, Float] do
+  def encode(x, _opts), do: x
+end
+
+defimpl Plausible.Audit.Encoder, for: [DateTime, Date, NaiveDateTime, Time] do
+  def encode(x, _opts), do: to_string(x)
+end
+
+defimpl Plausible.Audit.Encoder, for: [Atom] do
```
zoldar:
```suggestion
defimpl Plausible.Audit.Encoder, for: Atom do
```

extra/lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,120 @@
...
+        if k in allow_not_loaded or not raise_on_not_loaded? do
+          acc
+        else
+          raise Plausible.Audit.EncoderError,
+            message:
+              "#{k} association not loaded. Either preload, exclude or mark it as :allow_not_loaded in #{__MODULE__} options"
+        end
+
+      {k, v}, acc ->
+        Map.put(acc, k, Plausible.Audit.Encoder.encode(v))
```
zoldar: Would it make sense to carry opts over to `encode` here, for nested structs?

extra/lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,120 @@
...
+
+defimpl Plausible.Audit.Encoder, for: [Atom] do
+  def encode(nil, _opts), do: nil
+  def encode(true, _opts), do: true
+  def encode(false, _opts), do: false
+  def encode(x, _opts), do: Atom.to_string(x)
+end
+
+defimpl Plausible.Audit.Encoder, for: List do
+  def encode(x, _opts), do: Enum.map(x, &Plausible.Audit.Encoder.encode/1)
```
zoldar: Ditto about passing opts for nested structs in a list

extra/lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,120 @@
...
+            do:
+              struct
+              |> Map.drop(
+                unquote(except)
+                |> Map.put(:__allow_not_loaded__, unquote(allow_not_loaded))
+              )
+          )
+
+        true ->
+          quote(do: :maps.remove(:__struct__, struct))
```
zoldar: Shouldn't that put `__allow_not_loaded__` in the map as well?

extra/lib/plausible/audit/entry.ex
```diff
@@ -0,0 +1,94 @@
...
+    field :name, :string
+    field :entity, :string
+    field :entity_id, :string
+    field :meta, :map
+    field :change, :map, default: %{}
+    # default 0 is still useful in tests?
+    field :user_id, :integer, default: 0
+    field :team_id, :integer, default: 0
+    field :datetime, :naive_datetime_usec
+    field :actor_type, :string, default: "system"
```
zoldar: NBD, but could that be an enum?

extra/lib/plausible/audit/entry.ex
```diff
@@ -0,0 +1,94 @@
...
+        params
+      )
+
+    %__MODULE__{name: name}
+    |> cast(params, [:entity, :entity_id, :meta, :user_id, :team_id, :actor_type])
+    |> validate_required([:name, :entity, :entity_id, :actor_type])
+    |> put_change(:datetime, NaiveDateTime.utc_now())
+  end
+
+  def new(name, %{__struct__: struct, id: id}, params \\ %{}) do
```
zoldar, line 53:
```suggestion
  def for_struct(name, %{__struct__: struct, id: id}, params \\ %{}) do
```
?

extra/lib/plausible/audit/entry.ex
```diff
@@ -0,0 +1,94 @@
...
+    changeset(name, Map.merge(%{entity: to_str(struct), entity_id: to_str(id)}, params))
+  end
+
+  def name(name, entity, entity_id, params \\ %{})
```
zoldar:
```suggestion
  def for_entity(name, entity, entity_id, params \\ %{})
```
?

extra/lib/plausible/audit/encoder.ex
```diff
@@ -0,0 +1,120 @@
...
+        data
+
+      {0, n} when n > 0 ->
+        changes
+
+      {0, 0} ->
+        %{}
+
+      _ ->
+        %{before: data, after: changes}
```
zoldar, line 29: So that scoping trick didn't work like intended in the end? 🤔

extra/lib/plausible/auth/sso.ex
```diff
@@ -143,7 +148,7 @@ defmodule Plausible.Auth.SSO do
       team
       |> Ecto.Changeset.change()
       |> Ecto.Changeset.put_embed(:policy, policy_changeset)
-      |> Repo.update()
+      |> Repo.update_with_audit("sso_forced", %{team_id: team.id})
```
zoldar: This call is also used to disable Force SSO by passing `:none` for `mode`. So maybe the name should be `sso_force_mode_changed`?

test/plausible/audit_test.exs
zoldar: > what do you mean by that?

Sorry, that was a misguided idea, as encoder protocol would have to derive the entity name from struct/map/changeset anyway.

Using stringified module seems fine to me. I'm not sure if casting back to atom it going to be necessary at any point - even if - we can implement some sensible fallback strategy

extra/lib/plausible/audit/encoder.ex
aerosol: https://github.com/plausible/analytics/pull/5560#discussion_r2218891037

extra/lib/plausible/auth/sso.ex
aerosol: 76d6c306a7

extra/lib/plausible/audit/entry.ex
aerosol: I don't see the point of this given we have pattern matching, although your comment made me realise that second function was supposed to be another `new` clause that, turns out, is unused. So I'll just remove it.  4ba96e8252

extra/lib/plausible/audit/entry.ex
aerosol: 4ba96e8252

extra/lib/plausible/audit/entry.ex
aerosol: a36d35dad0

extra/lib/plausible/audit/encoder.ex
aerosol: 7aed7c4a95

extra/lib/plausible/audit/encoder.ex
aerosol: 4eb374cab0

extra/lib/plausible/audit/encoder.ex
aerosol: 4eb374cab0

zoldar (approved):
✨

PR #5561, opened 2025-07-08
Add tests for installationv2

ukutaht:
Adds tests for the new InstallationV2 liveview, largely inspired by existing InstallationTest module.

Also makes a small change to ensure that installation type detection is not run when `?type` URL parameter is supplied to installation view.

### DIFF
--- a/lib/plausible_web/live/installationv2.ex
+++ b/lib/plausible_web/live/installationv2.ex
@@ -30,7 +30,10 @@ defmodule PlausibleWeb.Live.InstallationV2 do
 
     flow = params["flow"] || Flows.provisioning()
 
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/installationv2.ex
```diff
@@ -30,7 +30,10 @@ defmodule PlausibleWeb.Live.InstallationV2 do
 
     flow = params["flow"] || Flows.provisioning()
 
-    if connected?(socket) and flow == Flows.provisioning() do
+    detect_installation_type? =
+      connected?(socket) and flow == Flows.provisioning() and !params["type"]
```
apata, line 34: Nice to see this made explicit, makes the UI easier to read as well!

PR #5565, opened 2025-07-10
Scriptv2 - change domain

ukutaht:
### Changes

* Creates a new change domain v2 flow which is entered when the `scriptv2` flag is enabled for the site in question
* New flow is implemented as a liveview
* New flow bypasses the installation and verification screens, redirects directly back to settings/general
* A note is left for cases where manual action still must be taken (WP, NPM and events API)


apata:
@ukutaht This looks good to me as is, but I couldn't resist trying to reorder the blocks of info. For your consideration 
Screenshot 2025-07-15 at 11 09 46

ukutaht:
I like it, could you push that visual change to this branch? :)

PR #5572, opened 2025-07-15
Script v2: Allow verifying that tracker installed correctly (1st iteration)

apata:
### Changes
Introduces v2 verification

### TODO in future PRs
More diagnostics cases
Diagnostics unit tests
Unify Browserless checks shared logic

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
```diff
@@ -0,0 +1,32 @@
...
+      url_that_maybe_busts_cache =
+        Plausible.InstallationSupport.URL.bust_url(url)
+
+      state2 =
+        Plausible.InstallationSupport.Checks.InstallationV2.perform(%{
+          state
+          | url: url_that_maybe_busts_cache
+        })
+
+      put_diagnostics(state2, diagnostics_are_from_cache_bust: true)
```
RobertJoonas: I understand the interpretation part itself is not implemented yet, but to me a diagnostic called `plausible_is_initialized_after_cache_bust` would make more sense.

We should only arrive to the "clear your site cache" recommendation if the verifier is able to detect an initialised plausible function only after cache bust. In other words, I don't think we want to recommend clearing cache unless we're certain that it will make the Plausible script work on the page.

FYI: I've also seen a case where the cache bust URL crashes the page and 500 is returned. In that case, it would be a lot more useful to keep relying on the existing diagnostics from the previous check.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -0,0 +1,162 @@
...
+          domain_is_expected? ->
+        success()
+
+      (tracker_is_version_2? or tracker_is_maybe_legacy?) and
+        (request_targets_plausible_api? or request_targets_local_proxy?) and
+          not domain_is_expected? ->
+        error(@error_unexpected_domain)
+
+      true ->
+        interpret(diagnostics, data_domain, url)
```
RobertJoonas: Looks like if we end up in this `cond` branch, we just keep calling the same function clause infinitely, until exhaustion.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -0,0 +1,162 @@
...
+
+    tracker_is_version_2? = plausible_variant in @supported_variants
+    tracker_is_maybe_legacy? = is_nil(plausible_variant)
+
+    # in all envs, we want to succeed with verification of live websites
+    request_targets_plausible_api? =
+      String.starts_with?(request_url, "#{PlausibleWeb.Endpoint.url()}/api/event") or
+        String.starts_with?(request_url, "https://plausible.io/api/event")
+
+    request_targets_local_proxy? = String.starts_with?(request_url, "/")
```
RobertJoonas: I think we might as well omit these checks. There are many other possible ways that a proxy might be configured, and we cannot determine the success based on the request URL that the browser sees. 

For example, the request url might be proxied under:

* `/subdirectory/api/event`
* `https://third.party/api/event`
* `/custom-endpoint`
* etc...

and still be forwarded to `https://plausible.io/api/event` correctly.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -0,0 +1,162 @@
...
+
+    cond do
+      (tracker_is_version_2? or tracker_is_maybe_legacy?) and
```
RobertJoonas: In what case would `tracker_is_version_2? or tracker_is_maybe_legacy?` evaluate to false? I'm a bit confused about why these are being used to determine the installation success?

As we agreed on the flow chart:

image

We would label both cases as success.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  @unknown_error Error.new!(%{
+                   message: "Your Plausible integration is not working",
+                   recommendation:
+                     "Please manually check your integration to make sure that the Plausible snippet has been inserted correctly",
+                   url:
+                     "https://plausible.io/docs/troubleshoot-integration#how-to-manually-check-your-integration"
+                 })
+  def interpret(%__MODULE__{service_error: _service_error} = diagnostics, _data_domain, url) do
```
RobertJoonas: Looks like matching on `%__MODULE__{service_error: _service_error}` doesn't have any effect here (the struct is always guaranteed to have the `service_error` key).

```suggestion
  def interpret(%__MODULE__{} = diagnostics, _data_domain, url) do
```

tracker/installation_support/check-disallowed-by-csp.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Checks if the CSP policy disallows the given hosts/domains or subdomains.
+ * @param {Record<string, string>} responseHeaders - Response headers with keys normalized to lowercase like { "x-foo": "bar" }
+ * @param {string[]} hostsToCheck - Array of domains/hosts to check. Must be provided.
+ * @returns {boolean}
+ */
+export function checkDisallowedByCSP(responseHeaders, hostsToCheck) {
```
RobertJoonas: Q: Why does `hostsToCheck` need to be an array? When would we need to check multiple hosts?

tracker/installation_support/verifier-v2.js
```diff
@@ -0,0 +1,214 @@
...
+      }
+    }
+
+    while (!plausibleIsInitialized) {
+      if (isPlausibleInitialized()) {
+        plausibleIsInitialized = true
+        plausibleVersion = getPlausibleVersion()
+        plausibleVariant = getPlausibleVariant()
+      }
+    }
```
RobertJoonas: This looks CPU intense. I'm not sure how Browserless would react to this but I would consider adding a small delay (e.g. 10ms) in every loop to make it more CPU friendly. A tiny delay would be a negligible effect on the UX.

```javascript
while (!plausibleIsOnWindow) {
  if (isPlausibleOnWindow()) {
    plausibleIsOnWindow = true
  }
  await new Promise(resolve => setTimeout(resolve, 10))
}
```

lib/plausible/installation_support/verification/diagnostics.ex
apata: Thanks! Agreed that it looks superfluous, but it's at least it's explicitly following the diagram. It would evaluate to false when plausible.s is set to 'custom' or 'john's plugin' for example. I considered not having this check, but then there's room for confusion, as if source doesn't matter.

lib/plausible/installation_support/verification/diagnostics.ex
apata: Hmm, but how many ways are there exactly? I think four. 

1. A known Plausible ingestion endpoint: `https://plausible.io/api/event`
2. Local URL that is defined from root `/custom`, `/api/event` 
3. Local URL that is defined relative to current location `custom/event`
4. Other full URL `https://example.com/api/custom`

I think I made it accept 1 & 2 only because these resemble the suggested installation methods (`web`, `wordpress`). 

But it does seem 3 and 4 are totally valid if we're talking about our NPM. 

Changing it would mean that recording a 200/202 status fetch with expected payload against any URL is a success.

tracker/installation_support/check-disallowed-by-csp.js
apata: Good question! It's JIC we want to check plausible.io AND specific subdomains. I will change it to one for now.

lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
apata: Thanks! This looks super wonky, so your concern is very valid.

My intent is that a success after cache bust is actually a fail, with helpful advice on how to clear site cache. If the user then clears site cache, the next time they verify, it will succeed with the first run: there'll be no more additional errors like domain mismatch etc. 

> only arrive to the "clear your site cache" recommendation if the verifier is able to **detect an initialised plausible function**

The bolded part is not a guarantee that cache bust is the only thing needed. For example, the function may be initialised, but the request might fail, or might be for an invalid domain. What we're really looking for is interpret success.

RobertJoonas: > My intent is that a success after cache bust is actually a fail

Agreed 100%

> with helpful advice on how to clear site cache. If the user then clears site cache, the next time they verify, it will succeed with the first run: there'll be no more additional errors like domain mismatch etc.

Ah, so from what I understand you're thinking to provide 2 in 1 recommendations like "Please clear your website cache AND check that the snippet on your site matches the installation instructions exactly"?

I'm not against it, but would prefer to recommend one "fix" at a time. This would leave the interpretation code in Elixir more lean, and seeing a single recommendation at a time might feel less overwhelming for the customer as well.

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: > But it does seem 3 and 4 are totally valid if we're talking about our NPM.

Another example - [CloudFlare standalone proxy guide](https://plausible.io/docs/proxy/guides/cloudflare). It's talking about v1, but it's totally valid to use the same instructions for a v2 snippet as well. The API path might look something like:

```
https://your-worker-name.your-cloudflare-zone.workers.dev/your-folder-name/event
```

> Changing it would mean that recording a 200/202 status fetch with expected payload against any URL is a success.

This is how the current v1 verification works as well, and I don't think there's anything wrong with that. In order for us to detect it as a false positive, the client would need to set up their own route that returns 200 or 202 for the event request. If this ever becomes a problem, I think we should consider setting a specific response header to check.

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: > It would evaluate to false when plausible.s is set to 'custom' or 'john's plugin' for example

In the rare case of such a custom integration, the client most likely knows what they're doing and it's fine to just show success as long as the event goes through.

lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
RobertJoonas: Sorry, I have come to realise that the way you've implemented it makes perfect sense.

* if normal (w/o cache bust) verification is a success, we do not attempt cache bust at all. I.e. `diagnostics_are_from_cache_bust` is false.
* if normal (w/o cache bust) verification is NOT a success, we verify again with the cache bust URL and set `diagnostics_are_from_cache_bust` to `true`
  * if after the cache bust verification, the final diagnostics get interpreted as a success, we know for a fact that cache was the issue. And therefore, purging the website cache is guaranteed to make the next verification a success
  * otherwise, if after the cache bust verification does not turn out to be a success, cache is not the issue.

apata: Cheers, thanks for clarifying it even better than I could! 

I've added the diagnostic case for this now.

lib/plausible/installation_support/verification/diagnostics.ex
apata: In that rare case, with the client knowing what they're doing, will they be waiting for our verification to succeed? 

I'm using this explicit approach and the diagnostic struct that diverges from v1 because I want the Elixir interpreter to be based as little as possible on the inner workings of the verifier and more on the inner workings of the tracker (what we attach to window, how callback works etc). That way, the verifier code can be treated as a black box most of the time. 

What if I refactor to check that source is nil or binary? In that way, clients that have crafted plausible objects with numeric, boolean, or object-shaped `s` values will see an unknown error because it's not following our verification API contract. Or I can just remove these checks, up to you.

lib/plausible/installation_support/checks/installation_v2.ex
```diff
@@ -0,0 +1,127 @@
...
+          "Browserless function returned with completed: false, error.message: #{inspect(data["error"]["message"])}",
+          state
+        )
+      )
+
+      put_diagnostics(state, service_error: data["error"]["message"])
+    end
+  end
+
+  defp handle_browserless_response(state, _body, status) do
```
RobertJoonas, line 117: Let's log the response body here too?

lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
```diff
@@ -0,0 +1,39 @@
...
+      url_that_maybe_busts_cache =
+        Plausible.InstallationSupport.URL.bust_url(url)
+
+      state_after_cache_bust =
+        Plausible.InstallationSupport.Checks.InstallationV2.perform(%{
+          state
+          | url: url_that_maybe_busts_cache
+        })
+
+      put_diagnostics(state_after_cache_bust, diagnostics_are_from_cache_bust: true)
```
RobertJoonas, line 36: I’ve seen a case where the “cache bust url” crashes the webpage. If the first run (w/o cache bust) returns diagnostics hinting to a known installation issue, I believe this information could get masked by a service error with cache bust diagnostics.

That said, there's no need to action this in this PR, but I just wanted to point it out.

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: I do believe it's completely okay to just ignore `window.plausible.s` in the verification process, and I would personally remove these checks for simplicity. A numeric, boolean or object-shaped plausible variant should be super rare, and I just think it's not something to be concerned about. If the test event responds with a success, that's enough to label the verification as a success. If the test event fails however, `plausible_variant` could be a useful debug diagnostic.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -0,0 +1,162 @@
...
+
+  @unknown_error Error.new!(%{
+                   message: "Your Plausible integration is not working",
+                   recommendation:
+                     "Please manually check your integration to make sure that the Plausible snippet has been inserted correctly",
+                   url:
+                     "https://plausible.io/docs/troubleshoot-integration#how-to-manually-check-your-integration"
+                 })
+  defp unknown_error(diagnostics, url) do
+    Sentry.capture_message("Unhandled case for site verification",
```
RobertJoonas: This error message seems to match exactly with the `LegacyVerification` one. Let's change it to `Unhandled case for site verification (v2)` to keep these separately in Sentry.

RobertJoonas (approved):
A couple of new comments inline but other than that, as agreed, we can merge and iterate on this. Nice work! 👍

lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
apata: Thanks! That one case that you mentioned was in my mind when I chose this approach. I decided that it's more beneficial to provide info to many users about their current installations (not cached previous installations).

lib/plausible/installation_support/checks/installation_v2.ex
apata: Thanks! I considered it but I don't want to accidentally log things like API keys etc.

PR #5580, opened 2025-07-22
Add `window.plausible` to NPM + `window.plausible.v2`

RobertJoonas:
### Changes

* Set `window.plausible` in the NPM tracker, with an option to disable binding it to the window.
* Set a `v2` flag on both NPM and Plausible Web scripts, to be used for v1 detection in the review installation and domain change flows.

Unrelated small fix: Fix a date-related billing test by removing starter tier assertions. It's sufficient to test the allowance margin logic for Growth and Business only for the time being.

### Tests

### Changelog
- [x] Entry has been added to tracker changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/test/plausible-npm.spec.js
+++ b/tracker/test/plausible-npm.spec.js
@@ -6,7 +6,6 @@ Config is set at init(), as we expect consumers to do in production.
 import { test, expect } from '@playwright/test'
 import { LOCAL_SERVER_ADDR } from './support/server'
 import { testPlausibleConfiguration, callInit } from './shared-configuration-tests'
-import { delay } from './support/test-utils'
 
 const DEFAULT_CONFIG = {
   domain: 'example.com',
@@ -21,15 +20,6 @@ async function openPage(page, config, options = {}) {
   if (!options.skipPlausibleInit) {
     await callInit(page, { ...DEFAULT_CONFIG, ...config }, 'window')
   }
-}
-
-function doesNotResolveWithin(promise, timeout) {
-  return Promise.race([
-    promise.then(() => {
-      throw new Error('Unexpected promise resolution')
-    }),
-    delay(timeout)
-  ])
 }
 
 test.describe('NPM package', () => {
--- a/tracker/npm_package/README.md
+++ b/tracker/npm_package/README.md
@@ -54,7 +54,7 @@ See also [plausible.d.ts](https://github.com/plausible/analytics/blob/master/tra
 | `logging` | Whether to log on ignored events. | `true` |
 | `customProperties` | Object or function that returns custom properties for a given event. | `{}` |
 | `transformRequest` | Function that allows transforming or ignoring requests | |
-| `bindToWindow` | Binds `track` to `window.plausible` after init. This is used by verification tool to detect whether plausible is installed | `true` |
+| `bindToWindow` | Binds `track` to `window.plausible` which is used by Plausible installation verification tool to detect whether Plausible has been installed correctly. If `bindToWindow` is set to false, the installation verification tool won't be able to automatically detect it on your site.  | `true` |
 
 #### Using `customProperties`
 


tracker/npm_package/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.3.3] - 2025-07-22
+
+- Bind the `track` function into `window.plausible`. This makes it possible for the Plausible verification agent to verify a successful installation. Can be disabled setting the `bindToWindow` config option to `false`.
+
```
apata, line 11: issue: the change should be under ## Unreleased heading: there's a script that creates the correct version title automatically when the package is released

tracker/src/plausible.js
```diff
@@ -35,6 +35,14 @@ function init(overrides) {
...
     window.plausible.init = init
     window.plausible.l = true
+    window.plausible.v2 = COMPILE_PLAUSIBLE_WEB
+  }
+
+  // Bind to window to be detectable by the verifier tool
+  // This is done in a 'safe' way to avoid breaking the page if window is frozen or running without window
+  if (COMPILE_PLAUSIBLE_NPM && config.bindToWindow && typeof window !== 'undefined') {
+    window.plausible = track
+    window.plausible.v2 = true
```
apata: question: shouldn't we also set the plausible.l = true here?

tracker/test/plausible-npm.spec.js
```diff
@@ -60,4 +70,19 @@ test.describe('NPM package', () => {
...
+  })
+
+  test('does not bind to window if bindToWindow is false', async ({ page }) => {
+    await openPage(page, {}, { skipPlausibleInit: true })
+    await callInit(page, { ...DEFAULT_CONFIG, bindToWindow: false }, 'window')
+    await doesNotResolveWithin(
+      page.waitForFunction('window.plausible !== undefined'),
+      1000
+    )
+  })
```
apata: nitpick, non-blocking: I think we can assert the same thing directly with testing APIs without extra utils
```suggestion
    await openPage(page, {}, { skipPlausibleInit: true })
    await callInit(page, { ...DEFAULT_CONFIG, bindToWindow: false }, 'window')
    await expect(
      page.waitForFunction(() => window.plausible !== undefined, undefined, { timeout: 1000 }),
    ).rejects.toThrow('page.waitForFunction: Timeout 1000ms exceeded.')
  })
```

tracker/src/plausible.js
```diff
@@ -35,6 +35,14 @@ function init(overrides) {
     window.plausible = track
     window.plausible.init = init
     window.plausible.l = true
+    window.plausible.v2 = COMPILE_PLAUSIBLE_WEB
```
apata: question, non-blocking: Is it too much future proofing if we set it to a value that allows distinguishing between web and npm, and between the current tracker_script_version and previous ones? Like `web:${tracker_script_version}` and `npm:${tracker_script_version}`

apata: Alternatively, if we go for middle ground, we could do `window.plausible.v=${tracker_script_version}` if `COMPILE_PLAUSIBLE_WEB` or `COMPILE_PLAUSIBLE_NPM`. Legacy installs would be detectable if it's not set.

ukutaht: I like squeezing more information in the variable. Seems like there's no extra cost but could become useful in the future.

apata:
@ukutaht Regarding extra information in variables, me and @RobertJoonas discussed it a little bit and came to the conclusion that it's better to have v as a separate field from installation source (s). That way, it doesn't need to be string parsed to compare with known tracker version integer.

tracker/src/plausible.js
RobertJoonas: https://github.com/plausible/analytics/pull/5580/commits/d4785b20f70175c1406e97b2156357ce77a99986

Issue #5518, opened 2025-06-19
Can't add a pageview goal with a / in the path


koenedaele:
### Past Issues Searched

- [x] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [x] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Plausible Cloud from plausible.io

### Describe the bug

I tried to add a goal. Went to site settings, clicked on Goals, cliked on Add Goal. A popup appears. Chose Goal Trigger "pageview". I tried to add a URL like /pages/** (fictional example). As soon as I type the /, my cursor jumps to the Search box for the Goals in the underlying page, away from the popup form. I can pick a path from the dropdown, but only a small number of possible paths are showing there.

### Expected behavior

To be able to pick the exact path I want.

### Environment

```markdown
- OS: Windows
- Browser: Google Chrome
- Browser Version: 136.0.7103.114
```

aerosol:
Thanks for the report @koenedaele - this is annoying indeed. Must be a regression from https://github.com/plausible/analytics/commit/ec2a56001604148f43394ee348747a8a95b7977b - I'll look for a fix.

aerosol:
@koenedaele the fix is up, let me know if that helps. Thanks!

koenedaele:
Yes, everything working as expected now. Thanks!

PR #5583, opened 2025-07-22
Fix shifting focus to filter bar when typing / in a legit input

aerosol:
Fixes https://github.com/plausible/analytics/issues/5518 - @apata @zoldar @RobertJoonas do you know if there's a better way to do it?

### DIFF
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -878,7 +878,7 @@ defmodule PlausibleWeb.Components.Generic do
               x-ref="filter_text"
               phx-debounce={200}
               autocomoplete="off"
-              x-on:keydown.prevent.slash.window="$refs.filter_text.focus(); $refs.filter_text.select();"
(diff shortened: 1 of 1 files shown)


zoldar (approved):

apata:
@aerosol I think if this fix works in at least some of the cases, we should merge to unblock users.

Unexpected keybind activations were a problem on the React dashboard as well. Currently it's mostly mitigated 
1) by looking at the event target (rather than document)
https://github.com/plausible/analytics/blob/master/assets/js/dashboard/keybinding.tsx#L8-L34

2) making the event handler registration conditional on navigation state (like whether popup is open or not) 
https://github.com/plausible/analytics/blob/master/assets/js/dashboard/dashboard-keybinds.tsx#L23
https://github.com/plausible/analytics/blob/master/assets/js/dashboard/router.tsx#L205

3) scoping event handlers to a particular element, which can trigger and handle the event before the more general event handlers kick in
https://github.com/plausible/analytics/blob/master/assets/js/dashboard/components/search-input.tsx#L48-L60

Out of these, the scenario we seem to be looking at resembles 2. The modal has grabbed full focus: only its own keybinds and keybinds of elements within it should be active. Not sure how to do it with generic components and alpine though.

apata (approved):

aerosol:
@apata thank you!

PR #5591, opened 2025-07-24
Detector.js - detect v1 and technologies used on the website

RobertJoonas:
### Changes

This PR introduces `detector.js` as well as its Elixir wrapper - `Plausible.InstallationSupport.Detection`.

This is meant to be integrated into the `InstallationV2` LiveView module (cc @ukutaht)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/installation_support/detection.ex
+++ b/lib/plausible/installation_support/detection.ex
@@ -1,0 +1,86 @@
+defmodule Plausible.InstallationSupport.Detection do
+  @moduledoc """
+  Exposes a perform function which visits the given URL via a Browserless
+  /function API call, and in a returns the following diagnostics:
+
+  * v1_detected (optional - detection can take up to 3s)
+  * gtm_likely
+  * wordpress_likely
+  * wordpress_plugin
+
+  These diagnostics are used to determine what installation type to recommend,
+  and whether to provide a notice for upgrading an existing v1 integration to v2.
+  """
+  require Logger
+  alias Plausible.InstallationSupport
+
+  path = Application.app_dir(:plausible, "priv/tracker/installation_support/detector.js")
+  # On CI, the file might not be present for static checks so we create an empty one
+  File.touch!(path)
+
+  @detector_code File.read!(path)
+  @external_resource "priv/tracker/installation_support/detector.js"
+
+  # Puppeteer wrapper function that executes the vanilla JS verifier code
+  @puppeteer_wrapper_code """
+  export default async function({ page, context }) {
+    try {
+      await page.setUserAgent(context.userAgent);
+      await page.goto(context.url);
+
+      await page.evaluate(() => {
+        #{@detector_code}
+      });
+
+      return await page.evaluate(async (detectV1, debug) => {
+        return await window.scanPageBeforePlausibleInstallation(detectV1, debug);
+      }, context.detectV1, context.debug);
+    } catch (error) {
+      const msg = error.message ? error.message : JSON.stringify(error)
+      return {data: {completed: false, error: msg}}
+    }
+  }
+  """
+
+  def perform(url, opts \\ []) do
+    req_opts =
+      [
+        headers: %{content_type: "application/json"},
+        body:
+          Jason.encode!(%{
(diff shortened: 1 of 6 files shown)


tracker/installation_support/detector.js
```diff
@@ -7,22 +7,23 @@ window.scanPageBeforePlausibleInstallation = async function(detectV1, debug) {
...
   }
 
+  let v1Detected = null
+
+  if (detectV1) {
+    log('Waiting for Plausible function...')
+    const plausibleFound = await waitForPlausibleFunction(3000)
+    log(`plausibleFound: ${plausibleFound}`)
+    v1Detected = plausibleFound && typeof window.plausible.s === 'undefined'
+    log(`v1Detected: ${v1Detected}`)
```
apata, line 17: issue, non-blocking: I'm not so sure of this being the right condition. As I read it, it would evaluate to true for v2 bootstrapper (window.plausible will be defined and since init hasn't been called yet, window.plausible.s will be undefined). 

It's definitely mergable for the time being, but maybe the better way would be to rule out v2 (by trying to and failing to detect v2).

ukutaht: > As I read it, it would evaluate to true for v2 bootstrapper

The [test](https://github.com/plausible/analytics/blob/b098ec3b889d0311cf28c34e2b35e6592a66c095/tracker/test/installation_support/detector.spec.js#L96) seems to work but there's probably a race condition. I'm guessing it works because the tracking script is executed before the puppeteer script but relying on this behaviour feels brittle.

Perhaps moving the `typeof window.plausible.s === 'undefined'` check into `waitForPlausibleFunction` would be more reliable since it will keep waiting when `window.plausible` is truthy and `window.plausible.s` is undefined.

lib/plausible/installation_support/detection.ex
```diff
@@ -0,0 +1,86 @@
...
+  * wordpress_likely
+  * wordpress_plugin
+
+  These diagnostics are used to determine what installation type to recommend,
+  and whether to provide a notice for upgrading an existing v1 integration to v2.
+  """
+  require Logger
+  alias Plausible.InstallationSupport
+
+  path = Application.app_dir(:plausible, "priv/tracker/installation_support/detector.js")
```
ukutaht: Is there a reason why `path` is not a module attribute?

lib/plausible/installation_support/detection.ex
```diff
@@ -0,0 +1,86 @@
...
+  # On CI, the file might not be present for static checks so we create an empty one
+  File.touch!(path)
```
ukutaht: This seems questionable to me. In general I think it's best to avoid touching source files during compilation. It's kinda of like a function mutating it's inputs. Yes, it can work when handled carefully but it adds a bunch of ways things can go wrong.

I think a better way would be to default to empty string when

```elixir
@detector_code File.read!(path)
```

Cannot find the file. This way it stays more pure without mutating the state of the filesystem.

tracker/installation_support/detector.js
RobertJoonas: Good catch! The simplest, most straight-forward fix to this seems to be waiting for `window.plausible.l` to evaluate to `true`. If it doesn't (e.g. when `init()` has not been called), it would keep waiting until the timeout. Made the change in https://github.com/plausible/analytics/pull/5591/commits/20ce2da133ecefb94f43ccc96738600456f1a06a.

lib/plausible/installation_support/detection.ex
RobertJoonas: Not really. Fixed in https://github.com/plausible/analytics/pull/5591/commits/90967d191a034cbe0b36bb1a6511039400ecbab4.

lib/plausible/installation_support/detection.ex
RobertJoonas: Makes sense. In https://github.com/plausible/analytics/pull/5591/commits/90967d191a034cbe0b36bb1a6511039400ecbab4 we avoid calling `File.touch!/1` and defaulting to an empty string when the file cannot be read.

PR #5599, opened 2025-07-28
Verifier v1: Retry verification on JS navigation

RobertJoonas:
### Changes

Retry verifier-v1 execution when client JS navigates and destroys current page context.

I've tested this with a local Browserless instance against a few sites that end up with the following warning log on current master:

```
[warning] [VERIFICATION] Browserless JS error: "Protocol error (Runtime.callFunctionOn): Execution context was destroyed."
```

With the changes introduced in this PR, the local Browserless call returns successful diagnostics. Therefore, merging this, I would expect the volume of these specific error cases to drop significantly.

Unfortunately, as stated in a code comment, this behaviour is currently not testable. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/installation_support/checks/installation.ex
+++ b/lib/plausible/installation_support/checks/installation.ex
@@ -8,12 +8,35 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
   @verifier_code File.read!(path)
   @external_resource "priv/tracker/installation_support/verifier-v1.js"
 
-  # Puppeteer wrapper function that executes the vanilla JS verifier code
+  # Puppeteer wrapper function that executes the vanilla JS verifier code.
+
+  # ### TRICKY (NOT TESTED): Handling client side JS navigation.
+
+  # We've seen numerous cases where client JS navigates or refreshes the
+  # page after load. Any such JS behaviour destroys the Puppeteer page
+  # context, meaning that our verifier execution gets interrupted and we
+  # end up in the `catch` clause.
+
+  # To make our best effort verifying these sites, we retry (up to twice)
+  # running the verifier again if we encounter this specific error.
+
+  # Important: On retries, we work with the client-modified page context
+  # instead of calling `page.goto(context.url)` again (which would most
+  # likely result in another interruptive navigation).
+
+  # Unfortunately, as things stand today, we cannot test this without
+  # spinning up a real Browserless instance or bringing in a bunch of
+  # test deps for Puppeteer.
   @puppeteer_wrapper_code """
   export default async function({ page, context }) {
-    try {
+    const MAX_RETRIES = 2
+
+    async function attemptVerification(isFirstAttempt) {
       await page.setUserAgent(context.userAgent);
-      await page.goto(context.url);
+
+      if (isFirstAttempt) {
+        await page.goto(context.url);
+      }
 
       await page.evaluate(() => {
         #{@verifier_code}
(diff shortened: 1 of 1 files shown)


lib/plausible/installation_support/checks/installation.ex
```diff
@@ -8,12 +8,35 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
...
+  # To make our best effort verifying these sites, we retry (up to twice)
+  # running the verifier again if we encounter this specific error.
+
+  # Important: On retries, we work with the client-modified page context
+  # instead of calling `page.goto(context.url)` again (which would most
+  # likely result in another interruptive navigation).
+
+  # Unfortunately, as things stand today, we cannot test this without
+  # spinning up a real Browserless instance or bringing in a bunch of
+  # test deps for Puppeteer.
```
apata: nitpick, non-blocking: I'd personally prefer if we said that we test this manually AND that we have playwright tests that test the retry.

lib/plausible/installation_support/checks/installation.ex
```diff
@@ -8,12 +8,35 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
...
-    try {
+    const MAX_RETRIES = 2
+
+    async function attemptVerification(isFirstAttempt) {
       await page.setUserAgent(context.userAgent);
-      await page.goto(context.url);
+
+      if (isFirstAttempt) {
+        await page.goto(context.url);
+      }
```
apata: nitpick, non-blocking: Since we're using the existing page context, we shouldn't need to set the user agent again. If we bring these out of `attemptVerification`, we won't need `isFirstAttempt` argument. That simplifies things a bit I think.

lib/plausible/installation_support/checks/installation.ex
```diff
@@ -22,9 +45,24 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
...
+        const shouldRetry = typeof error.message === 'string' && error.message.toLowerCase().includes('execution context')
+
+        if (shouldRetry && attempt <= MAX_RETRIES) {
+          // Brief delay before retry
+          await new Promise(resolve => setTimeout(resolve, 500))
+          continue
+        }
+
+        const msg = error.message ? error.message : JSON.stringify(error)
+        return {data: {completed: false, error: msg}}
```
apata: issue: At the moment, any errors in this block aren't caught (for example, accessing .message on an undefined error would throw and `throw undefined` is valid JS). I think we should have an outer catch block around the whole thing as well.

apata: If we don't want to have the error parsing twice, we can rethrow the error in the inner catch block. 
```
          await new Promise(resolve => setTimeout(resolve, 500))
          continue
        }

       throw error
    }
```

lib/plausible/installation_support/checks/installation.ex
```diff
@@ -22,9 +45,24 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
...
-      const msg = error.message ? error.message : JSON.stringify(error)
-      return {data: {completed: false, error: msg}}
+    }
+
+    for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
+      try {
+        const isFirstAttempt = attempt === 0
+        return await attemptVerification(isFirstAttempt)
+      } catch (error) {
+        const shouldRetry = typeof error.message === 'string' && error.message.toLowerCase().includes('execution context')
```
apata: suggestion, non-blocking: since error might be undefined in JS, we should access its fields defensively
```suggestion
        const shouldRetry = typeof error?.message === 'string' && error.message.toLowerCase().includes('execution context')
```

lib/plausible/installation_support/checks/installation.ex
RobertJoonas: As we'd essentially only be able to test the replicated Playwright code, it would feel more like fake confidence to me. I've mentioned manual testing and taking extra care when changing this code in https://github.com/plausible/analytics/pull/5599/commits/105e7184699812cf4c22a2dc0bd7fa611f36c1f4

lib/plausible/installation_support/checks/installation.ex
RobertJoonas: Indeed. I've now introduced a second `try` block as per your other comment, so this can just be at the start of the first, top-level `try` block. https://github.com/plausible/analytics/pull/5599/commits/181d2e41dc393efc054cc0989eaaacd2bfc89a23

lib/plausible/installation_support/checks/installation.ex
RobertJoonas: Makes sense! https://github.com/plausible/analytics/pull/5599/commits/181d2e41dc393efc054cc0989eaaacd2bfc89a23

PR #5607, opened 2025-08-04
Script v2: Allow changing tracker config over Sites API (WIP)

apata:
### Changes

Propose API structure

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -115,50 +115,86 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
...
 
-    case Sites.create(user, params, team) do
-      {:ok, %{site: site}} ->
-        json(conn, site)
+    case Repo.transact(fn ->
+           case Sites.create(user, params, team) do
+             {:ok, %{site: site}} ->
+               try do
+                 tracker_script_configuration =
+                   PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(
```
ukutaht: It would be cleaner to create a non-raising variant of `get_or_create_tracker_script_configuration!` in order to avoid using exceptions for control flow here.

The common pattern in Elixir, if both variants are needed, is as follows:

```elixir
def get_or_create_tracker_script_configuration(site, params) do
  # do the work
  if success do
    {:ok, result}
  else
    {:error, details}
  end
end

def get_or_create_tracker_script_configuration!(site, params) do
  {:ok, result} = get_or_create_tracker_script_configuration(site, params)
  result
end
```

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -115,50 +115,86 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
...
+                     site,
+                     if(params["tracker_script_configuration"],
```
ukutaht: This fallback should be unnecessary. If the key is missing, `get_or_create_tracker_script_configuration` will default to empty map

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
ukutaht: Once we have a non-raising variant that conforms to the standard `{:ok, result} | {:error, details}` pattern, the whole transaction can be refactored to the more idiomatic form:

```elixir
Repo.transact(fn -> 
  with {:ok, site} <- Sites.create(user, params, team)
       {:ok, config} <- Tracker.get_or_create_tracker_script_configuration(site, params["tracker_script_configuration"]) do
  {:ok, struct(site, tracker_script_configuration: tracker_script_configuration)
 else
   # Handle errors
end)
```

ukutaht (approved):

PR #5608, opened 2025-08-04
Clear sessions cache on transfer

zoldar:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/ingestion/write_buffer.ex
```diff
@@ -99,6 +129,34 @@ defmodule Plausible.Ingestion.WriteBuffer do
...
+
+      case :ets.lookup(name, :state) do
+        [state: %{locker: pid}] when is_pid(pid) ->
+          lock_loop(name, start, lock_timeout, lock_interval)
+
+        _ ->
+          :pass
+      end
+    else
+      unlock(name)
```
zoldar: That's  a bit of a weak point here. The `lock_timeout` ensures the buildup of queues in write buffers is not too big. However, if transfer gets stuck for whatever reason or takes longer than the timeout, the writebuffers will resume processing events.

One way to go about this could be wiping cache here before unlock. I'm not sure, however, whether it won't break consistency of transferred cache if the wipe happens right in the middle of transfer. Likely not, as `:ets.delete_all_objects` is supposedly atomic, but I'm not 100% certain about how table dumping works with deletion in the middle of `:ets.tab2list`.

lib/plausible/ingestion/write_buffer.ex
```diff
@@ -19,27 +22,54 @@ defmodule Plausible.Ingestion.WriteBuffer do
...
     buffer = opts[:buffer] || []
     max_buffer_size = opts[:max_buffer_size] || default_max_buffer_size()
     flush_interval_ms = opts[:flush_interval_ms] || default_flush_interval_ms()
+    lock_timeout_ms = opts[:lock_timeouts_ms] || @lock_timeout
+    lock_interval_ms = opts[:lock_interval_ms] || @lock_interval
 
     Process.flag(:trap_exit, true)
     timer = Process.send_after(self(), :tick, flush_interval_ms)
 
+    ^name = :ets.new(name, [:named_table, :set, :public])
```
aerosol: do we need an ets table to hold the lock? would it be possible to keep it in the process state? No loop necessary either. Or am I reading this wrong?

lib/plausible/session/transfer.ex
```diff
@@ -108,20 +112,43 @@ defmodule Plausible.Session.Transfer do
...
-        end
+        list_cache_names(session_version, parent)
 
       {@cmd_dump_cache, cache} ->
         Cache.Adapter.cache2list(cache)
 
       @cmd_takeover_done ->
+        # Wipe the cache after transfer is complete to avoid making the transferred cache stale
+        Cache.Adapter.wipe(:sessions)
+        # Unblock ingest after trasnsfer is done
```
aerosol:
```suggestion
        # Unblock ingest after transfer is done
```

lib/plausible/ingestion/write_buffer.ex
zoldar: I have changed locking approach - we only need to lock Session.WriteBuffer for the duration of the transfer, no need to block both queues.

Also the session cache is wiped when the lock times out.

lib/plausible/ingestion/write_buffer.ex
zoldar: The problem is, process state can only be accessed from within the process. While locking could be done with message passing, unlocking requires doing that outside the process queue which holds all the pending events to process stuck there for the duration of the lock. Unless you know some other way? 😅

PR #5611, opened 2025-08-04
Crm clean

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/live/customer_support/live.ex
```diff
@@ -0,0 +1,117 @@
...
+
+    update_params = [
+      id: "#{resource_key}-#{resource.id}-#{tab}",
+      tab: tab,
+      tab_params: tab_params
+    ]
+
+    update_params = Keyword.put(update_params, resource_key, resource)
+
+    send_update(tab_component_fn.(tab), update_params)
```
zoldar: Maybe call that `tab_routing_fn` instead? It gave me a good pause and some looking at one of call sites to understand what it does.

extra/lib/plausible_web/live/customer_support/live.ex
```diff
@@ -0,0 +1,117 @@
...
+          aria-label="Tabs"
+        >
+          {render_slot(@tabs)}
+        </nav>
+      </div>
+    </div>
+    """
+  end
+
+  def handle_tab_change_for(socket, tab, params, resource_key, tab_component_fn) do
```
zoldar: How about calling this function `redirect_to_tab` or `go_to_tab`? Also, passing higher order function only to resolve the tab name to the component when caller has already all the context feels a bit too roundabout - maybe it'd be clearer to pass `component` instead?

extra/lib/plausible_web/live/customer_support/live.ex
```diff
@@ -0,0 +1,117 @@
...
+      end
+
+      def handle_info({:error, msg}, socket) do
+        {:noreply, put_flash(socket, :error, msg)}
+      end
+
+      def handle_info({:navigate, path, success_msg}, socket) do
+        socket = if success_msg, do: put_flash(socket, :success, success_msg), else: socket
+        {:noreply, push_navigate(socket, to: path)}
+      end
```
zoldar, line 37: That could use some convenience API functions doing `send(self(), ...)` underneath, wouldn't it? 🤔

extra/lib/plausible_web/live/customer_support.ex
```diff
@@ -1,162 +1,48 @@
 defmodule PlausibleWeb.Live.CustomerSupport do
   @moduledoc """
-  Customer Support UI
+  Customer Support UI Router
```
zoldar: It would be good to give a short comment on where to look for the routing part IMHO, as it's not contained of referred to in this module.

extra/lib/plausible_web/live/customer_support/components/search.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  defp maybe_focus_search("user:" <> rest) do
+    {[Resource.User], rest, limit: 90}
+  end
+
+  defp maybe_focus_search(input) do
+    {@resources, input, limit: 30}
+  end
+
+  defp resource_path(%{type: "team", id: id}) do
```
zoldar: There's resource.type() and then there's the struct field `type`. I'm wondering now - perhaps `defstruct` should be declared as part of __using__ macro. That would remove the necessity to declare `type` as an explicit field and open up the possibility to define `resource_path` (and perhaps `render_result`) as a part of a protocol, whose implementation could be kept close to respective live component like `PlausibleWeb.Live.CustomerSupport.Site`?

extra/lib/plausible_web/live/customer_support/live.ex
aerosol: b9280933d4

extra/lib/plausible_web/live/customer_support/live.ex
aerosol: b9280933d4

extra/lib/plausible_web/live/customer_support.ex
aerosol: It's a dead comment really, thanks for pointing it out. Fixed in e8777ba - routes are now available in the router without any magic.

extra/lib/plausible_web/live/customer_support/live.ex
aerosol: Sort of, just not sure if here - components (not only live views) are sending those messages.

aerosol: 0aca102389

extra/lib/plausible_web/live/customer_support/components/search.ex
aerosol: LMK if you like that: 046a9b8020

zoldar: That works too 👍

aerosol: Were you thinking something else? I had hard time figuring it out

zoldar: I meant turning particular resource modules into structs themselves by literally putting `defstruct` definition inside the macro, so that you can match on `%Resource.Site{}` and not `%{type: xxx}`. With this in, it wouldn't be necessary to throw every bit of customization into Resource.Site (like, path, or LV component) but define a protocol (or protocols) suitable for every context.

```elixir
defimpl Plausible.CustomerSupport.ResourceComponent, for: Resource.Site do
  def search_result(assigns)
    ...
  end
end
```

same for routing/paths etc. But, I don't know, at the current stage, the approach you took is fine as well.

aerosol: Ah neat, perhaps an overkill for three resources though. Let's see how this pans out.

PR #5615, opened 2025-08-05
update ecto_ch and ch

ruslandoga:
👋 

I noticed https://github.com/plausible/analytics/pull/5609 and thought it's probably a good opportunity to update ecto_ch/ch as well. Especially because I made a mistake in ecto_ch 0.6.0 which allowed, but not fully supported, Ecto v3.13 (https://github.com/plausible/ecto_ch/pull/231). ecto_ch now uses stricter versioning (~> 3.13.0 vs previous ~> 3.12) which prevents this problem from reappearing in the future :)


apata:
> 👋
> 
> I noticed #5609 and thought it's probably a good opportunity to update ecto_ch/ch as well. Especially because I made a mistake in ecto_ch 0.6.0 which allowed, but not fully supported, Ecto v3.13 ([plausible/ecto_ch#231](https://github.com/plausible/ecto_ch/pull/231)). ecto_ch now uses stricter versioning (~> 3.13.0 vs previous ~> 3.12) which prevents this problem from reappearing in the future :)

Thanks for spotting it, @ruslandoga! Do you recommend 0.71 over 0.8?

ruslandoga:
ecto_ch [v0.8.0](https://github.com/plausible/ecto_ch/blob/master/CHANGELOG.md#080-2025-07-17) uses a newer version of Ch ([v0.5.0+](https://github.com/plausible/ch/blob/master/CHANGELOG.md#050-2025-07-17)) which adds some "preliminary" support for some new ClickHouse types (Variant, JSON, Time, Dynamic). If these new types are not necessary then it's probably "safer" to stay on v0.7.0 since it's pretty much v0.6.0 (from which we are upgrading) but with the Ecto 3.13 "fix". v0.8.0 hasn't been reported to be problematic yet, but I am also not sure if there are many people using it.

apata:
> ecto_ch [v0.8.0](https://github.com/plausible/ecto_ch/blob/master/CHANGELOG.md#080-2025-07-17) uses a newer version of Ch ([v0.5.0+](https://github.com/plausible/ch/blob/master/CHANGELOG.md#050-2025-07-17)) which adds some "preliminary" support for some new ClickHouse types (Variant, JSON, Time, Dynamic). If these new types are not necessary then it's probably "safer" to stay on v0.7.0 since it's pretty much v0.6.0 (from which we are upgrading) but with the Ecto 3.13 "fix". v0.8.0 hasn't been reported to be problematic yet, but I am also not sure if there are many people using it.

Thanks for explaining! 

I think I might have accidentally blocked merging this PR by trying to create a preview environment for it, which doesn't work for forks. Now "all checks pass" check won't succeed.

I have opened another PR with these changes at https://github.com/plausible/analytics/pull/5616, and I'll close this PR in favor of that, thanks!

PR #5619, opened 2025-08-06
Script v2: Handle verifying sites where there's an immediate JS redirect

apata:
### Changes

Like legacy verifier, v2 verifier should handle verifying sites where there's an immediate JS redirect. 

To do so, this PR adds logic to catch puppeteer errors about 'execution context' being canceled (which is the signal that navigation happened in the browser tab, canceling the function that was being evaluated), and to reattempt verification in the new location.

It also fixes the installation v2 module verifier loading logic to match the improved v1 module logic. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/installation_support/checks/installation_v2.ex
+++ b/lib/plausible/installation_support/checks/installation_v2.ex
@@ -1,40 +1,73 @@
 defmodule Plausible.InstallationSupport.Checks.InstallationV2 do
   require Logger
 
-  path = Application.app_dir(:plausible, "priv/tracker/installation_support/verifier-v2.js")
-  # On CI, the file might not be present for static checks so we create an empty one
-  File.touch!(path)
+  @verifier_code_path "priv/tracker/installation_support/verifier-v2.js"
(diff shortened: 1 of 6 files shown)


lib/plausible/installation_support/checks/installation_v2.ex
```diff
@@ -1,40 +1,73 @@
...
-  path = Application.app_dir(:plausible, "priv/tracker/installation_support/verifier-v2.js")
-  # On CI, the file might not be present for static checks so we create an empty one
-  File.touch!(path)
+  @verifier_code_path "priv/tracker/installation_support/verifier-v2.js"
+  @external_resource @verifier_code_path
 
-  @verifier_code File.read!(path)
-  @external_resource "priv/tracker/installation_support/verifier-v2.js"
+  # On CI, the file might not be present for static checks so we default to empty string
+  @verifier_code (case File.read(Application.app_dir(:plausible, @verifier_code_path)) do
```
ukutaht, line 8: nice 👍

PR #5621, opened 2025-08-07
Script v2/include prefix in tracker config

apata:
### Changes

After this PR, tracker script IDs are saved to Postgres with a prefix like `pa-V5OUguy5m04s95q12fg24` (currently just `V5OUguy5m04s95q12fg24`). 

For the sake of compatibility we still support requesting scripts like `s-V5OUguy5m04s95q12fg24`. 

In that case, it first looks for the config by `V5OUguy5m04s95q12fg24` and if it doesn't find it, tries `pa-V5OUguy5m04s95q12fg24` as well (maybe that config has been migrated already). 

There's a data migration script that prefixes all existing tracker script IDs with `pa-`. 

There's a separate PR for the migrations/2025.... file.  

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/data_migration/prefix_tracker_script_configuration_id.ex
+++ b/lib/plausible/data_migration/prefix_tracker_script_configuration_id.ex
@@ -1,0 +1,83 @@
+defmodule Plausible.DataMigration.PrefixTrackerScriptConfigurationId do
+  @moduledoc """
+  Migration to update tracker script configuration IDs to use the new prefixed format.
+
+  This migration:
+  1. Processes tracker configurations in batches to avoid memory issues
+  2. Uses batch updates for better performance
+  3. Handles the case where tracker configurations might not exist for all sites
+  4. Provides progress logging
+  """
+
+  import Ecto.Query
+  alias Plausible.Repo
+  alias Plausible.Site.TrackerScriptConfiguration
+
+  @batch_size 100
+
+  def run do
+    total_configs = count_total_configurations()
+    IO.puts("Found #{total_configs} total tracker configurations to process")
+
+    process_batch(0, total_configs)
+  end
+
+  defp count_total_configurations do
+    Repo.aggregate(
+      from(config in TrackerScriptConfiguration,
+        where: not like(config.id, "pa-%")
+      ),
+      :count,
+      :id
+    )
+  end
+
+  defp process_batch(offset, total_configs) do
+    configs = get_configurations_batch(offset)
+
+    if length(configs) > 0 do
+      IO.puts(
+        "Processing batch #{div(offset, @batch_size) + 1} (#{offset + 1}-#{offset + length(configs)} of #{total_configs})"
+      )
+
+      process_configurations_batch(configs)
+      process_batch(offset + @batch_size, total_configs)
+    else
+      IO.puts("Migration completed!")
+    end
+  end
+
+  defp get_configurations_batch(offset) do
+    Repo.all(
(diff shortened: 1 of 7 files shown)


lib/plausible/ecto/types/nanoid.ex
```diff
@@ -1,6 +1,6 @@
-defmodule Plausible.Ecto.Types.Nanoid do
+defmodule Plausible.Ecto.Types.NanoidBase do
```
ukutaht, line 1: Perhaps a [parameterized type](https://hexdocs.pm/ecto/Ecto.ParameterizedType.html) would make more sense here?

So we would have one type with optional `prefix` parameter.

lib/plausible_web/live/installationv2/instructions.ex
```diff
@@ -245,6 +245,10 @@ defmodule PlausibleWeb.Live.InstallationV2.Instructions do
   end
 
   defp tracker_url(tracker_script_configuration) do
+    if String.starts_with?(tracker_script_configuration.id, "pa-") do
```
ukutaht, line 248: It seems to me like the result of this conditional is unused. Elixir does not have early return.

lib/plausible/ecto/types/nanoid.ex
apata: I was looking for something like that initially but didn't find it, thanks! 

However, I'm not seeing yet where we'd benefit from being able to unpack or comprehend the prefix parameter. 

Prefixes are really easy handled in code without any additional Ecto magic. Even if we end up having tens of these, it seems easy to whip up a 

```ex
case unknown_id do 
  "pa-" <> rest when length(rest) == 21 -> handle_script_id(unknown_id)
  "api-" <> _rest -> handle_api_key...
end
```

To me the extra complexity of that feature makes it more risky, but maybe it's just because I haven't built up enough trust for the library yet.

lib/plausible_web/live/installationv2/instructions.ex
apata: Thanks! I added a test for good measure.

lib/plausible/data_migration/prefix_tracker_script_configuration_id.ex
```diff
@@ -0,0 +1,96 @@
...
+      {:ok, %{num_rows: updated_count}} ->
+        if updated_count != length(configs) do
+          IO.puts(
+            "Warning: Batch #{batch_num} - Expected to update #{length(configs)} configurations, but updated #{updated_count}"
+          )
+        end
+
+      {:error, error} ->
+        IO.puts("Error updating batch #{batch_num}: #{inspect(error)}")
+        # Continue with next batch instead of crashing
```
macobo, line 93: Q: Why is this preferrable to crashing?

macobo (approved):
Looks good to me, solid approach wrt batch migration.

Requires a few follow-up PRs:
- Upgrading live demo tracking code (and markos site)
- Removing the s- fallback code.

I'm not sure it's best to run this via a migration vs doing it via console - let me know if you want to discuss.

lib/plausible/data_migration/prefix_tracker_script_configuration_id.ex
apata: Valid question! Let's say the error is related to timing (can't get a connection within timeout etc), I still want it to try the next batches. I'll have to run it again to handle the missed batch, but at that time, there's less batches to run, so it's less likely to fail a second time.

PR #5634, opened 2025-08-13
Only display "Invoices" on sidebar when subscription is present

aerosol:
### Changes

In the settings sidebar, we'll only display invoices when there's (any) subscription present.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/views/layout_view.ex
+++ b/lib/plausible_web/views/layout_view.ex
@@ -85,6 +85,8 @@ defmodule PlausibleWeb.LayoutView do
     current_team = conn.assigns[:current_team]
     current_team_role = conn.assigns[:current_team_role]
 
+    subscription? = !!(conn.assigns[:current_team] && conn.assigns.current_team.subscription)
+
     options = %{
       "Account Settings" =>
         [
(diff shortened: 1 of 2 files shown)


lib/plausible_web/views/layout_view.ex
```diff
@@ -93,7 +95,7 @@ defmodule PlausibleWeb.LayoutView do
           if(not Teams.setup?(current_team),
             do: %{key: "Subscription", value: "billing/subscription", icon: :circle_stack}
           ),
-          if(not Teams.setup?(current_team),
+          if(not Teams.setup?(current_team) and subscription?,
```
macobo, line 99: Q: Is there a scenario where user was previously subscribed but now isn't or has a past invoice for a different reason? If that were the case, I think we should still display invoices.

aerosol: Thanks! The subscription always exists, even when cancelled/expired, so this should be fine still.

lib/plausible_web/views/layout_view.ex
```diff
@@ -85,6 +85,8 @@ defmodule PlausibleWeb.LayoutView do
     current_team = conn.assigns[:current_team]
     current_team_role = conn.assigns[:current_team_role]
 
+    subscription? = !!(conn.assigns[:current_team] && conn.assigns.current_team.subscription)
```
macobo, line 89:
```suggestion
    # NOTE: Subscription will still exist if it has expired or cancelled
    subscription? = !!(conn.assigns[:current_team] && conn.assigns.current_team.subscription)
```

lib/plausible_web/views/layout_view.ex
macobo: Thanks! I added a small suggestion above to make this more obvious, but other than that lgtm. :+1:

lib/plausible_web/views/layout_view.ex
aerosol: ok

PR #5635, opened 2025-08-13
Script v2: Make detection take less time

apata:
### Functional changes

* Make v1 detection timeout shorter 3 -> 1.5
* Set less lenient browserless.io /function API retry policy
* Decrease retries of browserless.io /function API calls
* Fail immediately if site definitely not reachable

#### CheckRunner supports one check declaring that no further checks are necessary
This is so checks don't need to be aware of each other. 
Used when Url check fails: continuing with the checks pipeline even if further checks were no-ops would cause pointless message cycling.

#### Fast DNS check module added Checks.Url
This is to fail faster with verification and detection when the site is definitely not reachable. Runs with either specific URL or site domain. The specific URL path is not yet used in the app, even with `scriptv2` flag. The idea is to support user specifying URL during verification (to be implemented in next PR).

#### Detection is now a checks pipeline with diagnostics module
This is to reuse Checks.Url. 

#### DNS lookup wrapper module added
This is to be able to mock lookups in tests.

#### InstallationSupport.Result has "data" key
This is to pass something along with `ok?: true` results.

### Tests

### Changelog
- [x] This PR does not make a user-facing change, except if `scriptv2` flag set

### Documentation

### Dark mode


apata:
I noticed after making this PR that we have three different ways to handle DNS queries during tests now. 

- `Plausible.Test.Support.DNSServer`
- mox with `Plausible.DNSInterface`
- code branches in prod to make testing easier

```ex
defmodule Plausible.Auth.SSO.Domain.Verification.Worker do
  @moduledoc """
  Background service validating SSO domains ownership

  `bypass_checks` and `skip_checks` job args are for testing purposes only
  (former will fail the verification, latter will succeed with `dns_txt` method)
  """
```

I personally prefer mox and would suggest refactoring the other two.

lib/plausible/installation_support/checks/detection.ex
```diff
@@ -0,0 +1,146 @@
+defmodule Plausible.InstallationSupport.Checks.Detection do
+  @moduledoc """
+  Calls the browserless.io service (local instance can be spawned with `make browserless`)
+  and runs verifier script via the [function API](https://docs.browserless.io/HTTP-APIs/function).
```
RobertJoonas:
```suggestion
  and runs detector script via the [function API](https://docs.browserless.io/HTTP-APIs/function).
```

lib/plausible/installation_support/checks/detection.ex
```diff
@@ -0,0 +1,146 @@
...
+  @detector_code_path "priv/tracker/installation_support/detector.js"
+  @external_resource @detector_code_path
+
+  # On CI, the file might not be present for static checks so we default to empty string
+  @detector_code (case File.read(Application.app_dir(:plausible, @detector_code_path)) do
+                    {:ok, content} -> content
+                    {:error, _} -> ""
+                  end)
+
+  # Puppeteer wrapper function that executes the vanilla JS verifier code
```
RobertJoonas:
```suggestion
  # Puppeteer wrapper function that executes the vanilla JS detector code
```

lib/plausible/installation_support/detection/checks.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Plausible.InstallationSupport.Detection.Checks do
+  @moduledoc """
+  Checks that are performed during tracker script installation verification.
```
RobertJoonas:
```suggestion
  Checks that are performed pre-installation, providing recommended installation method diagnostics and whether v1 is used on the site.
```

lib/plausible_web/live/change_domain_v2.ex
```diff
@@ -155,10 +155,34 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
...
   end
 
-  defp run_detection(url) do
-    case Detection.perform(url, detect_v1?: true) do
-      {:ok, result} -> {:ok, %{detection_result: result}}
-      e -> e
+  defp run_detection(domain) do
+    url_to_detect = nil
+
+    case url_to_detect
```
RobertJoonas: Let's extract detection_result into a variable and then do `case detection_result do`

lib/plausible_web/live/installationv2.ex
```diff
@@ -144,19 +144,22 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
 
-        type =
-          case result do
-            %{gtm_likely: true} -> "gtm"
-            %{wordpress_likely: true} -> "wordpress"
-            _ -> "manual"
-          end
-
-        {type, result.v1_detected}
+    case url_to_verify
```
RobertJoonas: Same comment here - let's extract a `detection_result` var for readability.

RobertJoonas:
> I noticed after making this PR that we have three different ways to handle DNS queries during tests now.
> 
> * `Plausible.Test.Support.DNSServer`
> * mox with `Plausible.DNSInterface`
> * code branches in prod to make testing easier

Nice observation! Although I'm not sure what production code branches you're referring to?

I personally think having these two test utils (DNSServer vs Mox) is fine since they're different levels of abstraction. I feel like having DRY tests is more important than sticking to a single test util here.

PR #5636, opened 2025-08-13
Include "Direct / None" in source filter suggestions

aerosol:
### Changes

We support "Direct / None" source filtering via dashboard click, but it doesn't show up as a filter suggestion when typing. This PR fixes that.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/filter_suggestions.ex
+++ b/lib/plausible/stats/filter_suggestions.ex
@@ -218,23 +218,33 @@ defmodule Plausible.Stats.FilterSuggestions do
         query_sessions(query)
       end
 
-    from(e in base_q,
-      where: fragment("? ilike ?", field(e, ^filter_name), ^filter_query),
-      select: field(e, ^filter_name),
-      group_by: ^filter_name,
-      order_by: [desc: fragment("count(*)")]
-    )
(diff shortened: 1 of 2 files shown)


lib/plausible/stats/filter_suggestions.ex
```diff
@@ -218,23 +218,33 @@ defmodule Plausible.Stats.FilterSuggestions do
...
+        site,
+        query,
+        filter_name,
+        filter_query
+      )
+      |> limit(25)
+      |> ClickhouseRepo.all()
+      |> Enum.filter(fn suggestion -> suggestion != "" end)
+
+    suggestions =
```
macobo: This is functional, but IMO would be cleaner as another chained function. How about:

```
defp add_direct_suggestion(suggestions, filter_name, filter_search) do
      if filter_name == :referrer_source and
           String.contains?(String.downcase("Direct / None"), String.downcase(filter_search)) do
        ["Direct / None" | suggestions]
      else
        suggestions
      end
end
```

And keeping everything chained as it was?

aerosol: cf9fda8fcc

macobo (approved):

PR #5640, opened 2025-08-14
Shared links

aerosol:
### Changes

Upon successful shared link authentication we should keep the query parameters intact, so that filters are applied and can be modified via the shared link itself.
https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8662254339

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -326,12 +327,20 @@ defmodule PlausibleWeb.StatsController do
 
         conn
         |> put_resp_cookie(shared_link_cookie_name(slug), token)
-        |> redirect(to: "/share/#{URI.encode_www_form(shared_link.site.domain)}?auth=#{slug}")
```
aerosol, line 329: fixing this omission in a bit

PR #5648, opened 2025-08-18
ScriptV2: TrackerScriptCache on ee

macobo:
### Changes

This PR adds TrackerScriptCache on ee and reduces its refresh interval. The main goal is to reduce number of database queries made for new unique paths.

On ee, the cache will contain `{id, true}` pairs such that the instance knows what ids are valid.

I tested the cache with ~1M tracker script configs, with its size being reported at 12MB.

Future improvements:
- Refreshing cache across the elixir cluster in lib/plausible_web/tracker.ex on create/update.

Basecamp ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8936344275

### DIFF
--- a/lib/plausible_web/tracker_script_cache.ex
+++ b/lib/plausible_web/tracker_script_cache.ex
@@ -34,14 +34,14 @@ defmodule PlausibleWeb.TrackerScriptCache do
   end
 
   @impl true
-  def get_from_source(id, opts \\ []) do
+  def get_from_source(id) do
     query =
       base_db_query()
       |> where([t], t.id == ^id)
 
     case Plausible.Repo.one(query) do
       %TrackerScriptConfiguration{} = tracker_script_configuration ->
-        cache_content(tracker_script_configuration, opts)
+        cache_content(tracker_script_configuration)
 
       _ ->
         nil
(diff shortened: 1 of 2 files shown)


macobo:
Good catch. Yes we should - I'll add a case to lib/plausible_web/controllers/api/system_controller.ex similar to Site Cache.

lib/plausible_web/tracker_script_cache.ex
```diff
@@ -30,14 +34,14 @@ defmodule PlausibleWeb.TrackerScriptCache do
   end
 
   @impl true
-  def get_from_source(id) do
+  def get_from_source(id, opts \\ []) do
```
aerosol: The idea behind `get_from_source` was to be never called in production; it's only a fallback for tests to never use cache unless explicitly instructed to (for when cache is disabled).

I think it's best to use `get_or_store/2` for prod call sites, if possible.

aerosol: (the whole dance is primarily to avoid shared memory clashes when running tests in parallel)

macobo: Yeah, I was being too clever here to be DRY. Removed.

aerosol:
Just double checking: if this means it might take up to 30 seconds for an integration to become functional, is the small refresh interval short enough from new site provisioning/UX perspective?

macobo:
> Just double checking: if this means it might take up to 30 seconds for an integration to become functional, is the small refresh interval short enough from new site provisioning/UX perspective?

We're hoping so. As I mentioned in basecamp, the next step would be to do some elixir cluster broadcasting to "elixir cache invalidate" when the config gets updated/created. If this is something you're interested in picking up, it would be great.

aerosol:
image

I think it's going to be mainly @cnkk and it highly depends on his availability

PR #5649, opened 2025-08-18
Don't render external links for WP search_query property

aerosol:
### Changes

Hovering on WP Search Queries breakdown in Custom Properties panel should not generate external links. 

Second commit makes external link icons render in Firefox too, when they're actually meant to be displayed. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/reports/list.tsx
+++ b/assets/js/dashboard/stats/reports/list.tsx
@@ -38,7 +38,7 @@ function ExternalLink<T>({
         target="_blank"
         rel="noreferrer"
         href={dest}
-          className="w-4 h-4 invisible group-hover:visible"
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/reports/list.tsx
```diff
@@ -38,7 +38,7 @@ function ExternalLink<T>({
         target="_blank"
         rel="noreferrer"
         href={dest}
-        className="w-4 h-4 hidden group-hover:block"
```
zoldar, line 41: Maybe visibility would be a more fitting property here? https://tailwindcss.com/docs/visibility

aerosol: Thanks, pushed.

PR #5653, opened 2025-08-19
Implement dedicated persistence service

zoldar:
### Changes

Implements ability to switch backends for persisting events and sessions in ClickHouse. 3 backends are available:

- `Embedded` - works like the existing combination of `register_session` and `write_to_buffer` steps, doing all the processing, buffering and actual database inserts inside `analytics` app
- `Remote` - sends the event payload to a remote instance which, after processing and buffering, returns updated event back; managing buffering, session tracking and writing to ClickHouse is delegated to the remote app when using this backend
- `EmbeddedWithRelay` - combines `Embedded` with `Remote` treating `Embedded` as the primary, authoritative backend for persisting sessions and events; `Remote` is run fully asynchronously in that mode; this backend is needed only during the transitional period for proper load testing of remote persistence service; it assumes a separate ClickHouse database is setup for the service to avoid double writes

New env vars:
- `PERSISTOR_BACKEND` - determines which backend is used. Defaults to `embedded`.
- `PERSISTOR_URL` - URL of remote persistence endpoint. Defaults to `http://localhost:8001/event`. Used by `Remote` and `EmbeddedWithRelay`
- `PERSISTOR_TIMEOUT_MS` - timeout in ms for waiting for response from persistence service before giving up. Defaults to 10_000. Used by `Remote` and `EmbeddedWithRelay`
- `PERSISTOR_COUNT` - number of HTTP2 connections (pools, acutally) dedicated to communicating with remote persistence service. Defaults to 200. Used by `Remote` and `EmbeddedWithRelay`.

Nothing should change from operational perspective when using `Embedded` backend except `write_to_buffer` step no longer being recorded in metrics as it's merged into `register_session`. `EmbeddedWithRelay` records extra `register_remote_session` ingestion step. `Remote` backend can drop events with 3 new drop reasons: `persist_timeout` (timeout calling persistor endpoint), `persist_decode_error` (failure to decode persistor response) and `persist_error` (any other unclassified error). In addition to new errors, more detailed failure reasons are logged as warnings.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -148,7 +148,8 @@ persistor_backend =
     "remote" -> Plausible.Ingestion.Persistor.Remote
   end
 
-persistor_url = get_var_from_path_or_env(config_dir, "PERSISTOR_URL", "http://localhost:8001/event")
+persistor_url =
+  get_var_from_path_or_env(config_dir, "PERSISTOR_URL", "http://localhost:8001/event")
 
 persistor_count = get_int_from_path_or_env(config_dir, "PERSISTOR_COUNT", 200)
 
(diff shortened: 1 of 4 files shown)


aerosol (approved):
solid work 💪

PR #5655, opened 2025-08-19
ScriptV2: Use async instead of defer, new snippet

macobo:
This PR switches us over to the proposal in https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8935985563.

Depends on https://github.com/plausible/analytics/pull/5662


lib/plausible_web/templates/layout/_tracking.html.heex
```diff
@@ -6,7 +6,7 @@
   >
     var scriptParams = JSON.parse(document.currentScript.dataset.scriptParams)
 
-    window.plausible=window.plausible||function(){(window.plausible.q=window.plausible.q||[]).push(arguments)},window.plausible.init=function(i){window.plausible.o=i||{}};var script=document.createElement("script");script.type="text/javascript",script.defer=!0,script.src=scriptParams.script_url;var r=document.getElementsByTagName("script")[0];r.parentNode.insertBefore(script,r);
+    window.plausible=window.plausible||function(){(window.plausible.q=window.plausible.q||[]).push(arguments)},window.plausible.init=function(i){window.plausible.o=i||{}};var script=document.createElement("script");script.type="text/javascript",script.async=!0,script.src=scriptParams.script_url;var r=document.getElementsByTagName("script")[0];r.parentNode.insertBefore(script,r);
```
RobertJoonas, line 9: Let's switch to the cleaner/shorter snippet format here too?

RobertJoonas (approved):
Not very confident in our test suite covering all pitfalls, but consulting with AI tools, the whole tracker code should be async safe. Let's see how it goes 🤞(PS: remider to change the base branch to master)

lib/plausible_web/templates/layout/_tracking.html.heex
macobo: Will update separately along with the others once we have confirmed that async functions as expected.

PR #5656, opened 2025-08-20
Keep google auth deletions idempotent

aerosol:
### Changes

Rapid double-click and similar scenarios likely weren't handled well when deleting google auth.

### DIFF
--- a/lib/plausible_web/controllers/site_controller.ex
+++ b/lib/plausible_web/controllers/site_controller.ex
@@ -300,10 +300,11 @@ defmodule PlausibleWeb.SiteController do
       conn.assigns[:site]
       |> Repo.preload(:google_auth)
 
-    Repo.delete!(site.google_auth)
+    if site.google_auth do
+      Repo.delete!(site.google_auth)
+    end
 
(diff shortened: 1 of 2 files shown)


lib/plausible_web/controllers/site_controller.ex
```diff
@@ -300,10 +300,11 @@ defmodule PlausibleWeb.SiteController do
       conn.assigns[:site]
       |> Repo.preload(:google_auth)
 
-    Repo.delete!(site.google_auth)
-
-    conn = put_flash(conn, :success, "Google account unlinked from Plausible")
+    if site.google_auth do
+      Repo.delete!(site.google_auth)
+    end
```
zoldar, line 305: Or maybe even something like:

```elixir
Repo.delete_all(from a in Site.GoogleAuth, where: a.id == ^site.google_auth.id)
```

?

zoldar (approved):

lib/plausible_web/controllers/site_controller.ex
aerosol: @zoldar but I'd still have to check if site.google_auth is non-nil

PR #5657, opened 2025-08-20
ScriptV2: Domain change refinement

macobo:
### Changes

This PR refines the scriptv2 domain change flow by:
1. Improving the "Back to settings button"
2. Tweaking dark mode support
3. Purging CDN cache on domain change

The major change is also detecting whether npm library was used and showing the "warning" to update their configuration in 72 hours.

Along the way we also support storing/detecting npm module within the onboarding module.

Basecamp ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8949869924

### DIFF
--- a/lib/plausible/installation_support/checks/detection.ex
+++ b/lib/plausible/installation_support/checks/detection.ex
@@ -139,7 +139,7 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
     do: [
       v1_detected: data["v1Detected"],
       gtm_likely: data["gtmLikely"],
-      npm_likely: data["npmLikely"],
+      npm: data["npm"],
       wordpress_likely: data["wordpressLikely"],
       wordpress_plugin: data["wordpressPlugin"],
       service_error: nil
--- a/lib/plausible/installation_support/checks/detection.ex
+++ b/lib/plausible/installation_support/checks/detection.ex
@@ -139,6 +139,7 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
     do: [
       v1_detected: data["v1Detected"],
       gtm_likely: data["gtmLikely"],
+      npm_likely: data["npmLikely"],
       wordpress_likely: data["wordpressLikely"],
       wordpress_plugin: data["wordpressPlugin"],
       service_error: nil
--- a/lib/plausible/installation_support/checks/installation.ex
+++ b/lib/plausible/installation_support/checks/installation.ex
@@ -106,8 +106,6 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
 
   - `data.gtmLikely` - whether or not the site uses GTM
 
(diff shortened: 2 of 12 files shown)


lib/plausible/installation_support/checks/installation.ex
```diff
@@ -106,6 +106,8 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
 
   - `data.gtmLikely` - whether or not the site uses GTM
 
+  - `data.npmLikely` - whether or not the site uses the NPM package
```
RobertJoonas: This module is used for v1 verification. The intent behind all of these diagnostics was to gain confidence in Browserless from the beginning.

I don't think we need to touch this file anymore during the scriptv2 project, other than getting rid of it at the end.

tracker/installation_support/check-npm.js
```diff
@@ -0,0 +1,8 @@
+export function checkNPM(document) {
+  console.log('checkNPM', document, window.plausible)
```
RobertJoonas: Dangling log

```suggestion
```

lib/plausible/installation_support/checks/detection.ex
```diff
@@ -139,6 +139,7 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
     do: [
       v1_detected: data["v1Detected"],
       gtm_likely: data["gtmLikely"],
+      npm_likely: data["npmLikely"],
```
RobertJoonas: The word "likely" sounds a bit misleading here. I think we should rather name it just `npm` (similar to how `wordpressPlugin` is not `wordpressPluginLikely`).

The diagnostics containing the word "likely" are meant for trying to guess the most appropriate installation instructions for this site. It has nothing to do with how Plausible might already be installed.

macobo: Cleaned it up. 

Aside: the fact changing this takes so much effort is IMO a sign of too many abstractions are involved during installation that each remap things to similar other things.

lib/plausible/installation_support/checks/installation.ex
```diff
@@ -180,6 +180,7 @@ defmodule Plausible.InstallationSupport.Checks.Installation do
             {:wordpress_plugin_diff, :wordpress_plugin?, "wordpressPlugin"},
             {:wordpress_likely_diff, :wordpress_likely?, "wordpressLikely"},
             {:gtm_likely_diff, :gtm_likely?, "gtmLikely"},
+            {:npm_diff, :npm?, "npm"},
```
RobertJoonas:
```suggestion
```

lib/plausible/installation_support/checks/detection.ex
RobertJoonas: Agreed. I think we could get rid of the interpretation layer in detection.

RobertJoonas (approved):

PR #5660, opened 2025-08-20
ScriptV2: Show warning when using Legacy GTM template, update links

macobo:
### Changes

Shows a warning linking to (not yet written) GTM template upgrade guide.

image

### DIFF
--- a/lib/plausible_web/live/installationv2.ex
+++ b/lib/plausible_web/live/installationv2.ex
@@ -102,8 +102,23 @@ defmodule PlausibleWeb.Live.InstallationV2 do
           }>
             <.notice class="mt-4" theme={:yellow}>
               Your website is running an outdated version of the tracking script. Please
-              <.styled_link new_tab href="https://plausible.io/docs">update</.styled_link>
+              <.styled_link new_tab href="https://plausible.io/docs/script-update-guide.md">
+                update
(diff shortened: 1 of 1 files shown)


lib/plausible_web/live/installationv2.ex
```diff
@@ -102,11 +102,26 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
+                update
+              </.styled_link>
               your tracking script before continuing
             </.notice>
           </div>
 
+          <div :if={
+            @v1_detected.result == true and @recommended_installation_type.result == "gtm" and
+              @installation_type.result == "gtm"
+          }>
```
RobertJoonas, line 115: Hmm, I think it's a bit trickier than that. `gtm` being the recommended installation type does not mean that plausible is installed via gtm - it just means that the site is using GTM for something.

However, considering that we're certain that there's a v1 script installed, I do agree that it makes sense to show that notice in the GTM tab as well. Perhaps just with the same general message linking to a docs page that will list instructions for manual and gtm separately?

macobo: The documents will end up linking to each other. One document trying to do both will be too involved.

I've updated the phrasing to indicate that it's a possibility that gtm template was used.

PR #5661, opened 2025-08-20
fix CE warning in tests

RobertJoonas:
### Changes

Fix console warnings on `ce_test`. Run InstallationSupport related test suites only on Enterprise Edition.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/live/installationv2_test.exs
+++ b/test/plausible_web/live/installationv2_test.exs
@@ -1,540 +1,546 @@
 defmodule PlausibleWeb.Live.InstallationV2Test do
   use PlausibleWeb.ConnCase, async: true
+  use Plausible
 
-  import Phoenix.LiveViewTest
-  import Plausible.Test.Support.HTML
-  import Plausible.Teams.Test
-  import Mox
+  @moduletag :ee_only
 
-  alias Plausible.Site.TrackerScriptConfiguration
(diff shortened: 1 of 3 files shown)


test/plausible_web/live/installationv2_test.exs
```diff
@@ -1,540 +1,546 @@
 defmodule PlausibleWeb.Live.InstallationV2Test do
   use PlausibleWeb.ConnCase, async: true
+  use Plausible
 
-  import Phoenix.LiveViewTest
-  import Plausible.Test.Support.HTML
-  import Plausible.Teams.Test
-  import Mox
+  @moduletag :ee_only
```
aerosol, line 5: I guess the tag when everything's wrapped with `on_ee` is redundant

aerosol (approved):

test/plausible_web/live/installationv2_test.exs
RobertJoonas: Yeah, I was seeing the same pattern across the whole codebase (that both `@moduletag :ee_only` and `on_ee` are used in combination). This is why I wrote it like that. Doesn't hurt to keep it either, I guess? Makes it clear already at the top of the file, that everything is `ee_only`?.

PR #5668, opened 2025-08-26
Trim `month`, `year`, `day` periods to local now on main graph

aerosol:
### Changes

Ref: https://feedback.plausible.io/410

Before:
image

After:
image

Before:
image

After:
image

Before:
image

After:
image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/time.ex
+++ b/lib/plausible/stats/time.ex
@@ -50,7 +50,7 @@ defmodule Plausible.Stats.Time do
   end
 
   defp time_labels_for_dimension("time:month", query) do
-    date_range = Query.date_range(query)
+    date_range = Query.date_range(query, trim_trailing: should_trim_future_dates?(query))
 
     n_buckets =
       Timex.diff(
@@ -68,7 +68,7 @@ defmodule Plausible.Stats.Time do
   end
 
   defp time_labels_for_dimension("time:week", query) do
-    date_range = Query.date_range(query)
+    date_range = Query.date_range(query, trim_trailing: should_trim_future_dates?(query))
 
     n_buckets =
       Timex.diff(
@@ -86,7 +86,8 @@ defmodule Plausible.Stats.Time do
   end
 
   defp time_labels_for_dimension("time:day", query) do
-    Query.date_range(query)
+    query
+    |> Query.date_range(trim_trailing: should_trim_future_dates?(query))
     |> Enum.into([])
     |> Enum.map(&format_datetime/1)
   end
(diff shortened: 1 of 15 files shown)


test/plausible/stats/time_test.exs
```diff
@@ -232,4 +233,226 @@ defmodule Plausible.Stats.TimeTest do
              ]
     end
   end
+
+  describe "time_labels/1 with current period trimming" do
```
macobo: Note: This definetly deserves a changelog entry.

lib/plausible/stats/time.ex
```diff
@@ -50,7 +50,7 @@ defmodule Plausible.Stats.Time do
   end
 
   defp time_labels_for_dimension("time:month", query) do
-    date_range = Query.date_range(query)
+    date_range = Query.date_range(query, trim_trailing: should_trim_future_dates?(query))
```
macobo: (Random spot to start a thread) Problem: This causes problems in Stats APIv2.

1. If I query APIv2 for period `year`, the response query will report the date_range to the end of the year but the time labels won't reflect that:

image

2. This might break existing integrations in unexpected ways.

The way I'd tackle these two problems would be to:
1. Add a new `query.include` flag, e.g. `trim_trailing_dates`. This can be set in priv/json-schemas/query-api-schema.json and lib/plausible/stats/filters/query_parser.ex
2. Update the query time range in lib/plausible/stats/query_optimizer.ex as a separate clause when `query.include.trim_trailing_dates` is set.
3. Update timeseries endpoint, calling `Query.set_include(query, :trim_trailing, true)`

aerosol:
@macobo reverted original attempt at https://github.com/plausible/analytics/pull/5668/commits/ad4410686da811f0d53c42e946ad2a8a57b8ed82 and re-implemented it according to your pointers at https://github.com/plausible/analytics/pull/5668/commits/1af839b505af1929aa1323cacd4f708cc5224cff

You originally asked to alter the "timeseries" endpoint, but I'm not sure if doing it globally like that makes sense, since we only want/need this change for the main graph presentation purely, thus no APIv2 changes/tests are required. Maybe no point in even exposing that flag to the users? What do you think?

macobo:
> You originally asked to alter the "timeseries" endpoint

You did in lib/plausible_web/controllers/api/stats_controller.ex

> since we only want/need this change for the main graph presentation purely, thus no APIv2 changes/tests are required. Maybe no point in even exposing that flag to the users? What do you think?

We'd like to keep parity between what the stats api and dashboard provides. As such I'd default to providing new functionality in stats API as well.

---

Functionality looks good, great work! Only thing I'd add is a 1-2 tests under e.g. test/plausible_web/controllers/api/external_stats_controller/query_test.exs checking `month`/`year` functionality e2e. query-internal-test endpoint provides a `date` parameter you can set that makes this reasonable to test without needing to add new 'relative to current date' functionality for tests.

aerosol:
Technically the api accepts date_time_range if I'm not mistaken so the parity could be still achieved by putting users in control of how they like to clamp the range, instead of making them request a range that doesn't exist and then having us clamp it _only if_ the user supplies a vaguely named flag :sweat_smile:

aerosol:
> query-internal-test endpoint provides a date parameter you can set that makes this reasonable to test without needing to add new 'relative to current date' functionality for tests.

what does that mean? how would that work? do you mean something like https://github.com/plausible/analytics/pull/5668/commits/3288b8bf7b3881044ff6da51792200343cd05f83 ?

macobo (approved):
Sorry for this getting confusing! I think the main issue were loose ends from APIv2 migration.

I've pushed some commits that: 
- replace `period` with `input_date_range` and set it in APIv2
- Populate query.now based on `date` param
- Added a changelog + tweaked the test and where include was added.

I think this is ready to go live.

PR #5669, opened 2025-08-26
Script v2: Verifier accepts cookies

apata:
### Changes

Makes use of https://www.npmjs.com/package/@duckduckgo/autoconsent to accept cookies during installation verification process. 

I picked this approach in favor of hand-rolling a cookie consenter because I don't think we'll be able to maintain a hand-rolled solution at a good-enough level as consent management platforms (CMPs) wax and wane in popularity and our user base becomes more international.

The autoconsent library and rules to detect CMPs are bundled into the verifier script. This increases its size: the compiled verifier-v2.js file is now ~840kb (up from ~3kb on master) , of which 800kb are rulesets for 100s of different cookie banner setups. Since the extra size takes longer to deliver to Browserless.io, I think we should trim it down. 

Edit: It's now trimmed down to be ~100kb.

Todos
- [x] Fix unexpected CI-only test failure with "installed plausible web variant › there are more than maxAttempts JS navigations"
- [x] Expose in Verification diagnostics
- [x] Remove cookieBannerLikely related diagnostics case
- [x] Replace test on real pages with a couple of tests on fake pages
- [x] Stop waiting for cookie consent if plausible function test successful
- [x] Trim rules to recognize and accept cookies of the following CMPs
  - Cookiebot
  - Quantcast
  - Iubenda
  - Usercentrics (button-based)
  - didomi.io
  - oil.js <-- a likely defunct CMP that's compiled in
  - optanon
  - quantcast2 
  - springer <--  a likely defunct CMP that's compiled in
  - wordpressgdpr <-- a likely defunct CMP that's compiled in
  

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/test/installation_support/verifier-v2.spec.ts
+++ b/tracker/test/installation_support/verifier-v2.spec.ts
@@ -263,13 +263,21 @@ test.describe('installed plausible web variant', () => {
     const [result, _] = await Promise.all([
       executeVerifyV2(page, {
         ...DEFAULT_VERIFICATION_OPTIONS,
+        timeoutMs: 1500,
         responseHeaders
       }),
-      page.evaluate(
-        ({ targetUrl }) =>
-          setTimeout(() => (window.location.href = targetUrl), 250),
-        { targetUrl: urlBeta }
-      )
+      (async () => {
+        // start navigation timer only when the verifier code has been added to the page
+        await page.waitForFunction(`() => !!window.verifyPlausibleInstallation`)
+        await page.evaluate(
+          ({ targetUrl }) => {
+            setTimeout(() => {
+              window.location.href = targetUrl
+            }, 500)
+          },
+          { targetUrl: urlBeta }
+        )
+      })()
     ])
     await expect(page.getByText('beta')).toBeVisible()
 
@@ -301,7 +309,6 @@ test.describe('installed plausible web variant', () => {
   test('there are more than maxAttempts JS navigations', async ({ page }, {
     testId
   }) => {
-    const timeoutBetweenAttemptsMs = 100
     const maxAttempts = 2
 
     const { url: urlGamma } = await initializePageDynamically(page, {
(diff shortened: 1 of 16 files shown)


tracker/test/installation_support/verifier-v2.spec.ts
```diff
@@ -334,14 +342,35 @@ test.describe('installed plausible web variant', () => {
...
         maxAttempts,
         responseHeaders
       }),
-      page.evaluate(
-        (url) => setTimeout(() => (window.location.href = url), 250),
-        urlBeta
-      )
+      (async () => {
+        // start navigation timer only when the verifier code has been added to the page
+        await page.waitForFunction(`() => !!window.verifyPlausibleInstallation`)
```
RobertJoonas, line 352: I feel like the tests for the retry logic, i.e.:

* "there is a JS navigation immediately on page load to a page with the provided snippet"
* "there are more than maxAttempts JS navigations"

are adding unnecessary complexity without bringing any real value (if not even raising false confidence). They are testing our test harness, not production behavior which is implemented in Puppeteer.

Since the "Fix unexpected CI-only test failure with JS navigations" todo in this PR hints these tests are already problematic, I would suggest simply getting rid of them.

RobertJoonas (approved):
Other than one comment in line about the retry tests, LGTM 👍

tracker/test/installation_support/verifier-v2.spec.ts
apata: Thanks, @RobertJoonas! You are right that those tests are relying on similarity between puppeteer and playwright verifier wrappers that is not kept in sync automatically. 

I hear you on false confidence, but my first thought is not to leave it untested, rather make the tests give us real confidence.

apata: I personally found them valuable in developing the puppeteer wrapper, that's why I kept them in: not so much for confidence as for development ease, especially when doing test-driven development.

PR #5673, opened 2025-08-26
Verification v2: Better Browserless error handling + verify custom URL option

RobertJoonas:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -198,30 +198,38 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
-    %Result{
-      ok?: false,
-      errors: [error.message],
-      recommendations: [%{text: error.recommendation, url: error.url}]
+  def interpret(%__MODULE__{service_error: "net::" <> _}, expected_domain, url) do
+    attempted_url =
+      if url do
+        String.split(url, "?") |> List.first()
+      else
+        "https://#{expected_domain}"
```
apata: nitpick, non-blocking: This else branch should never be hit, because DNS check should set a flag that prevents further checks for nil url. I'd prefer we added an `is_binary(url)` guard and get an internal server error if nil url is passed, because that would be unexpected.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -198,30 +198,38 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
 
-  defp error(%Error{} = error) do
-    %Result{
-      ok?: false,
-      errors: [error.message],
-      recommendations: [%{text: error.recommendation, url: error.url}]
+  def interpret(%__MODULE__{service_error: "net::" <> _}, expected_domain, url) do
+    attempted_url =
+      if url do
+        String.split(url, "?") |> List.first()
```
apata: nitpick, non-blocking: Here we hide the true url that we accessed. Didn't we find that that our cache busting query parameter can sometimes even be the cause of some verification failures? If the question is between visually appealing and showing the customer the truth, I'd choose truth, but that might be just me.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -198,30 +198,38 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
-                          })
-  def interpret(%__MODULE__{service_error: service_error}, _expected_domain, _url)
-      when service_error in [:domain_not_found, :invalid_url],
-      do: error(@error_domain_not_found)
-
-  def interpret(%__MODULE__{} = diagnostics, _expected_domain, url),
-    do: unknown_error(diagnostics, url)
+  def interpret(%__MODULE__{service_error: service_error}, expected_domain, url)
+      when service_error in [:domain_not_found, :invalid_url] do
+    attempted_url = if url, do: url, else: "https://#{expected_domain}"
```
apata, line 211: nitpick, non-blocking: regarding the else branch, same as https://github.com/plausible/analytics/pull/5673#discussion_r2303394145

lib/plausible/installation_support/verification/diagnostics.ex
apata: The reason I'm pointing it out is because I don't think this module should duplicate how the URL is built from expected domain: that responsibility is on the URL check (once we can remove legacy code branches).

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -198,30 +198,38 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
 
-  defp success() do
-    %Result{ok?: true}
+    %Error{
```
apata: suggestion, non-blocking: Legacy verification errors have string templating logic. I thought that was pretty cleverly done and really lightweight, but I didn't port it yet, because we didn't need it before this error. What do you think of porting that, so these new error could be created compile time with Error.new!? That would give us the "best" of both: compile-time protection against typos etc, while still supporting flexible messages.

test/plausible_web/live/verification_test.exs
```diff
@@ -58,6 +58,28 @@ defmodule PlausibleWeb.Live.VerificationTest do
...
 
+    @tag :ee_only
+    test "from custom URL input form to verification", %{conn: conn, site: site} do
+      # Get liveview with ?custom_url=true query param
+      {:ok, lv, html} =
+        conn |> no_slowdown() |> live("/#{site.domain}/verification?custom_url=true")
+
+      verifying_installation_text = "Verifying your installation"
+
+      # Assert form is rendered instead of kicking of verification automatically
```
apata:
```suggestion
      # Assert form is rendered instead of kicking off verification automatically
```

lib/plausible_web/live/verification.ex
```diff
@@ -112,12 +117,17 @@ defmodule PlausibleWeb.Live.Verification do
...
-  def handle_event("retry", params, socket) do
+  def handle_event("retry", _, socket) do
     launch_delayed(socket)
-    {:noreply, reset_component(socket, params["url_to_verify"])}
+    {:noreply, reset_component(socket)}
+  end
+
+  def handle_event("verify-custom-url", %{"custom_url" => custom_url}, socket) do
+    socket = launch_delayed(socket, custom_url)
+    {:noreply, reset_component(socket)}
```
apata: question: why do we need to do `socket = launch_delayed(...` on L129 but not on L125 or L119?

apata (approved):
I think this is a good approach and it seems to work well!

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: Considering it was only a single case, I don't think it's worth rendering the URL with the long query parameter for everyone. Even for the cases where the query param is what breaks the site, it's not terrible UX if we don't show the real URL in the error message.

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: Hmm, I think it's possible to hit the else branch here.

image

URL is `nil` and service error is `:domain_not_found`. Exactly what's handled by this function clause.

As per the other comment, it's indeed impossible there, because errors like `net::.*` can surface only from Browserless which is impossible if we skip the Browserless check.

lib/plausible_web/live/verification.ex
RobertJoonas: Launch delayed always returns the socket, but modifies it (assigning `url_to_verify` and `custom_url_input?`) only when `custom_url` is passed as argument. Actually, thinking of it now, it would be cleaner to extract those socket assigns out of this function and do it right above the `launch_delayed` call.

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: Indeed, the URL can never be nil here. Changed that in https://github.com/plausible/analytics/pull/5673/commits/6f22d75c18cf71dfa720a680cb57f46856bbb0de

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: Good call. I've done that now 👍

lib/plausible_web/live/verification.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5673/commits/909eb0cc38cecdedcda2a39a888abbd116fdec36

apata: Good idea!

PR #5674, opened 2025-08-27
Avoid slowdowns in tests due to code server waits

aerosol:
This is supposed to be taken care of
by https://github.com/elixir-lang/elixir/commit/d05731194e0c9ef6fa00dedd1803f1e03623b976 but not quite it seems.

The first test with a fairly involved exunit template (ConnCase) is usually 700ms vs 60ms whichever comes second. (48-core Threadripper, but similar results on macs AFAIU). This makes --trace / --slowest flags produce misleading results.

h/t @zoldar

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -23,8 +23,12 @@ end
 
 default_exclude = [:slow, :minio, :migrations]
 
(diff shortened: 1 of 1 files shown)


zoldar (approved):

aerosol:
Ref https://github.com/elixir-lang/elixir/pull/14811

PR #5679, opened 2025-08-27
Don't cache app and dependency specific dialyzer PLTs on CI

zoldar:
### Changes

This change reduces risk of dialyzer going wonky due to stale PLTs, which [happened in the past](https://github.com/plausible/analytics/pull/4920#issuecomment-2550394046). The PLTs for app and dependencies are kept separate from core PLTs. While core PLTs are kept in the CI cache, the app and dep PLTs are kept outside of it.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -147,7 +147,7 @@ jobs:
           path: |
             deps
             _build
-            priv/plts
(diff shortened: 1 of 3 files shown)


zoldar:
Given the timings, it does not seem to be worth it really. The problems occur very rarely. Closing.

PR #5683, opened 2025-08-27
Verification: new interpretation case for non-200 page responses

RobertJoonas:
### Changes

Basecamp ref: https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/8925426346#__recording_9007056623 (screen recording)

In addition to Browserless running into a network error with `page.goto(context.url)`, the response status returned by that function can indicate that the returned HTML page is not what we expected. Note that this response status does not matter if window.plausible is detected on the page.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/installation_support/verification/diagnostics_test.exs
+++ b/test/plausible/installation_support/verification/diagnostics_test.exs
@@ -258,7 +258,7 @@ defmodule Plausible.InstallationSupport.Verification.DiagnosticsTest do
                        errors: [
                          ^any(
                            :string,
-                           ~r/.*could not load your website.*/
+                           ~r/.*couldn't verify your website at https:\/\/#{expected_domain}.*/
                          )
                        ],
                        recommendations: [
@@ -266,7 +266,7 @@ defmodule Plausible.InstallationSupport.Verification.DiagnosticsTest do
                            text:
                              ^any(
                                :string,
-                               ~r/https:\/\/example.com.*403.*verify your installation manually/
+                               ~r/403 error.*firewall.*authentication.*CDN.*verify your integration manually/
                              ),
                            url:
                              "https://plausible.io/docs/troubleshoot-integration#how-to-manually-check-your-integration"


lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -233,13 +234,41 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
     |> struct!(data: %{offer_custom_url_input: true})
   end
 
+  @error_non_200_page_response Error.new!(%{
+                                 message:
+                                   "We could not load your website to verify the installation",
+                                 recommendation:
+                                   "Your site URL (<%= @attempted_url %>) responded with status <%= @page_response_status %>. Please make sure there are no authentication requirements, firewall/CDN rules blocking our request, or errors on your server. Alternatively, you can follow our documentation and verify your installation manually",
```
apata: suggestion, non-blocking: I tweaked this a little bit to make it similar to the other error
```suggestion
                                   "We couldn't verify your website at <%= @attempted_url %>",
                                 recommendation:
                                   "Our verification tool encountered a <%= @page_response_status %> error. Please check for anything that might be blocking it from reaching your site, like a firewall, authentication requirements, or CDN rules. If you'd prefer, you can skip this and verify your integration manually.",
```

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -233,13 +234,41 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
+          plausible_is_on_window: plausible_is_on_window,
+          plausible_is_initialized: plausible_is_initialized,
+          response_status: page_response_status
+        },
+        _expected_domain,
+        url
+      )
+      when is_binary(url) and page_response_status != 200 and plausible_is_on_window != true and
+             plausible_is_initialized != true do
+    attempted_url = String.split(url, "?") |> List.first()
```
apata: nitpick, non-blocking: This is used twice now, maybe we can extract it to `defp shorten_url`?

apata (approved):

lib/plausible/installation_support/verification/diagnostics.ex
RobertJoonas: This looks good, thanks! Commited.

PR #5687, opened 2025-09-01
Script v2: Update tests relating to tagged a elements as well as examples from docs

apata:
### Changes

There's a problem tracking tagged a elements if they're within svg tags. `url` property isn't added automatically. It's hard to fix using ES5-only logic without increasing script size or script complexity (if polyfilling only for .compat variants)

Including the URL automatically with tagged links is convenience functionality that I propose we document with this edge case like this: https://github.com/plausible/docs/pull/635.

This PR 
1) updates the test case about this
2) adds tests for transformRequest examples from docs

### Tests

### Changelog

### Documentation
- [ ] https://github.com/plausible/docs/pull/635

### Dark mode

### DIFF
--- a/tracker/test/tagged-events.spec.ts
+++ b/tracker/test/tagged-events.spec.ts
@@ -26,7 +26,15 @@ test.beforeEach(async ({ page }) => {
       await route.fulfill({
         status: 200,
         contentType: 'text/html',
-        body: '<!DOCTYPE html><html><head><title>mocked page</title></head><body>mocked page</body></html>'
+        body: /* HTML */ `<!DOCTYPE html>
+          <html>
+            <head>
+              <title>mocked page</title>
+            </head>
+            <body>
+              mocked page
+            </body>
+          </html>`
       })
     })
 })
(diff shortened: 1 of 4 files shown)


tracker/test/transform-request.spec.ts
```diff
@@ -253,3 +254,163 @@ for (const mode of ['web', 'esm']) {
     })
   })
 }
+
+test.describe(`transformRequest examples from /docs work`, () => {
```
apata, line 269: These should help verify that what we recommend to users actually works as intended

tracker/test/tagged-events.spec.ts
```diff
@@ -434,8 +434,12 @@ for (const mode of ['legacy', 'web']) {
         action: () => page.click('circle'),
         expectedRequests: [
           {
-            n: 'link click'
-            // bug with p.url, can't assert
+            n: 'link click',
+            p: {
+              expected: { url: {} },
```
RobertJoonas, line 470: Feels a bit weird to expect `event.props.url` be an empty object. It doesn't make sense for any custom prop value be of type object, since it's stored as string. Could we make a change in the tracker script to fall back to `event.props.url = null` when this happens? That would allow us to also get rid of this hacky assertion below.

apata: Thanks! It is definitely weird, but I thought that instituting pre-processing for some or all props client-side was out of scope for this PR. It would add a bit of weight to the script (not sure how much). I tested that object shaped prop values are ignored by the server, so it would be a second layer of protection.

RobertJoonas: Ok, makes sense! 👍

PR #5691, opened 2025-09-02
Script v2: Adds linter and prettier, satisfies linter

apata:
### Changes
No functional changes. 

- Adds linter to /tracker folder
- Configures linter for various environments we have in the repo
  - node environment for compiling
  - es5 tracker script browser environment
  - modern JS verification script browser environment
  - node / browser playwright testing env
- Satisfies linter or adds ignore for special cases, erring towards not overriding recommended configuration without good justification

- Adds prettier for /tracker folder
- Configures prettier to match the configuration in assets
- Does not satisfy prettier yet: it will be a separate PR that only changes the format of ~70 files

### Tests

### Changelog

### Documentation

### Dark mode


tracker/src/config.js
```diff
@@ -1,9 +1,12 @@
+// eslint-disable-next-line no-redeclare
 var location = window.location
+// eslint-disable-next-line no-redeclare
 var document = window.document
```
apata: These two variables redeclare `document` and `location` globals, which even in ES5 environments can be accessed without referencing window (open browser console, execute `console.log(document)`). 

I tried refactoring these variable names (and the export/import names), but that caused some tests to start failing. 

I didn't get too in-depth, but it looks like our script, even in modules where it's importing document from config, is sometimes actually relying on the original declaration, not the redeclared value. I suspect this might become the source of sneaky bugs in the future. 

Also, the module engagement.js is specifically not importing document: it's using the original declaration. 

As far as I can see, the reason to redeclare and export is to save some bytes, which I don't think is worth the risks. I'm proposing to refactor to always access the original declaration in all tracker files.

apata: > As far as I can see, the reason to redeclare and export is to save some bytes, which I don't think is worth the risks. I'm proposing to refactor to always access the original declaration in all tracker files.

Hmm, apparently these exports don't actually save bytes: https://github.com/plausible/analytics/pull/5691/commits/7cf9e38fd0978610e04f5ce006074a2ecd9696ad

RobertJoonas (approved):
LGTM 👍

PR #5692, opened 2025-09-02
CE: Fix v2 onboarding, review installation, and domain change flows

RobertJoonas:
### Changes

This PR improves the UX on Community Edition, where there's no Browserless service available to carry out pre-installation scans or verification checks.

- [x] Turn off the pre-installation scan /  v1 script detection on CE. Make sure there's no loading spinner showing up and the user lands straight to the installation (or domain change) screen. 
- [x] Get rid of the "Tag Manager" tab on the InstallationV2 screen. There are no existing instructions for v2 script installation on GTM without the new template, and adding those instructions is out of scope for now.
- [x] In the Review Installation flow, add a link in the footer (only CE; only manual installation) linking to the v1 -> v2 migration guide:
image

- [x] In the Domain Change flow, change the copy of the yellow warning message (about maybe having to take additional action to change the site domain of the installation) and display it always on CE: 
image

- [x] Tests: enable existing `:ee_only` tests for CE and add new ones explicit to CE where necessary
- [ ] Update changelog about script v2

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/change_domain_v2.ex
+++ b/lib/plausible_web/live/change_domain_v2.ex
@@ -2,12 +2,16 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
   @moduledoc """
   LiveView for the change domain v2 flow.
   """
+  use Plausible
   use PlausibleWeb, :live_view
 
   alias PlausibleWeb.Router.Helpers, as: Routes
   alias PlausibleWeb.Live.ChangeDomainV2.Form
-  alias Plausible.InstallationSupport.{Detection, Result}
   alias Phoenix.LiveView.AsyncResult
+
+  on_ee do
+    alias Plausible.InstallationSupport.{Detection, Result}
+  end
 
   def mount(
         %{"domain" => domain},
@@ -28,19 +32,25 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
      )}
   end
 
-  def handle_params(_params, _url, socket) do
-    socket =
-      if socket.assigns.live_action == :success and connected?(socket) do
-        site_domain = socket.assigns.site.domain
+  on_ee do
+    def handle_params(_params, _url, socket) do
+      socket =
+        if socket.assigns.live_action == :success and connected?(socket) do
+          site_domain = socket.assigns.site.domain
 
-        assign_async(socket, :detection_result, fn ->
-          run_detection(site_domain)
-        end)
-      else
-        socket
-      end
+          assign_async(socket, :detection_result, fn ->
+            run_detection(site_domain)
+          end)
+        else
+          socket
+        end
 
-    {:noreply, socket}
+      {:noreply, socket}
+    end
+  else
+    def handle_params(_params, _url, socket) do
+      {:noreply, socket}
+    end
   end
 
   def render(%{live_action: :change_domain_v2} = assigns) do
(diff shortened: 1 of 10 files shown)


lib/plausible_web/live/verification.ex
```diff
@@ -190,7 +187,7 @@ defmodule PlausibleWeb.Live.Verification do
 
   def handle_info({:all_checks_done, %State{} = state}, socket) do
     interpretation =
-      if(FunWithFlags.enabled?(:scriptv2, for: socket.assigns.site),
+      if(PlausibleWeb.Tracker.scriptv2?(socket.assigns.site),
```
apata: question: Can we provide the second argument with socket.assigns.current_user here as well? It might be a bit weird to see the legacy interpretation

lib/plausible_web/live/change_domain_v2.ex
```diff
@@ -90,62 +104,115 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
...
+    defp ce_generic_notice(assigns) do
+      assigns = assign(assigns, docs_link: @change_domain_docs_link)
+
+      ~H"""
+      <.notice data-test-id="ce-generic-notice" class="mt-4" title="Additional steps may be required">
+        If you're using our legacy script (i.e. your Plausible snippet includes the
+        <code>data-domain</code>
+        attribute), OR if you've installed Plausible using
+        our NPM package, you <i>must</i>
+        also update the site <code>domain</code>
```
apata, line 206:
```suggestion
        our NPM package, to guarantee continuous tracking, you <i>must</i>
        also update the site <code>domain</code>
```

apata: It's a mega nitpick, feel free to ignore. The justification is that it softens the blow from "must <do something>".

lib/plausible_web/live/change_domain_v2.ex
```diff
@@ -2,13 +2,21 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
...
   alias PlausibleWeb.Router.Helpers, as: Routes
   alias PlausibleWeb.Live.ChangeDomainV2.Form
-  alias Plausible.InstallationSupport.{Detection, Result}
   alias Phoenix.LiveView.AsyncResult
 
+  on_ee do
+    alias Plausible.InstallationSupport.{Detection, Result}
+  end
+
+  @change_domain_docs_link "https://plausible.io/docs/change-domain-name/"
```
apata: meganitpick, non-blocking: to remove the trailing slash 
```suggestion
  @change_domain_docs_link "https://plausible.io/docs/change-domain-name"
```

lib/plausible_web/live/change_domain_v2.ex
```diff
@@ -90,62 +104,115 @@ defmodule PlausibleWeb.Live.ChangeDomainV2 do
...
+        %Result{ok?: false, errors: errors} ->
+          {:error, List.first(errors, :unknown_reason)}
+      end
+    end
+  else
+    defp ce_generic_notice(assigns) do
+      assigns = assign(assigns, docs_link: @change_domain_docs_link)
+
+      ~H"""
+      <.notice data-test-id="ce-generic-notice" class="mt-4" title="Additional steps may be required">
```
apata: nitpick, non-blocking: at least in web, the attribute `data-testid` seems to be much more common than `data-test-id` https://github.com/search?q=%22data-testid%22&type=code

lib/plausible_web/live/installationv2.ex
```diff
@@ -94,40 +136,23 @@ defmodule PlausibleWeb.Live.InstallationV2 do
             >
               <Icons.wordpress_icon /> WordPress
             </.tab>
-            <.tab patch={"?type=gtm&flow=#{@flow}"} selected={@installation_type.result == "gtm"}>
-              <Icons.tag_manager_icon /> Tag Manager
-            </.tab>
+            <%= on_ee do %>
```
apata, line 139: question, non-blocking: This definitely works, but it's a second source of truth for what's already conditional. What if we check for all .tab elements that `<method> in @installation_methods` instead?

lib/plausible_web/live/installationv2.ex
```diff
@@ -145,18 +170,30 @@ defmodule PlausibleWeb.Live.InstallationV2 do
               flow={@flow}
               recommended_installation_type={recommended_installation_type}
             />
-            <Instructions.gtm_instructions
-              :if={@installation_type.result == "gtm"}
-              recommended_installation_type={recommended_installation_type}
-              tracker_script_configuration_form={@tracker_script_configuration_form.result}
-            />
+            <%= on_ee do %>
+              <Instructions.gtm_instructions
```
apata, line 174: question, non-blocking: This definitely works, but it's a second source of truth for what's already conditional. What if we check for all `Instructions.` elements that `<method> in @installation_methods` instead?

lib/plausible_web/live/installationv2.ex
```diff
@@ -167,25 +204,71 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
-    case detection_result do
-      %Result{
-        ok?: true,
-        data: %{suggested_technology: suggested_technology, v1_detected: v1_detected}
-      } ->
-        {suggested_technology, v1_detected}
+        _ ->
+          {"manual", false}
+      end
+    end
```
apata: suggestion, non-blocking: at least for me, I'd prefer to see the CE version of this function quite close to the EE version. I think it might be clearer if we have two on_ee blocks, the first that just has the two variants of detect_recommended_installation_type and the second that has just the EE-only private functions.

lib/plausible_web/live/installationv2.ex
```diff
@@ -27,19 +38,49 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
+          assign_loading_states(socket)
+        end
       else
-        assign_loading_states(socket)
+        # On Community Edition, there's no v1 detection, nor pre-installation
+        # site scan - we just default the pre-selected tab to "manual".
+
+        # Although it's functionally unnecessary, we stick to using `%AsyncResult{}`
+        # for these assigns to minimize branching out the CE code and maintain only
+        # a single `render` function.
```
apata, line 62: I agree with this choice, nice!

lib/plausible_web/live/installationv2.ex
```diff
@@ -198,12 +281,12 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
         assign(assigns,
           class:
-            "bg-white dark:bg-gray-800 rounded-md px-3.5 py-2.5 text-sm font-medium flex items-center"
+            "bg-white dark:bg-gray-800 rounded-md px-3.5 py-2.5 text-sm font-medium flex items-center flex-1 justify-center whitespace-nowrap"
         )
       else
         assign(assigns,
           class:
-            "bg-gray-100 dark:bg-gray-700 rounded-md px-3.5 py-2.5 text-sm font-medium flex items-center cursor-pointer"
+            "bg-gray-100 dark:bg-gray-700 rounded-md px-3.5 py-2.5 text-sm font-medium flex items-center cursor-pointer flex-1 justify-center whitespace-nowrap"
```
apata, line 291: Good solution for the wrapping and scaling issues!

lib/plausible_web/live/change_domain_v2.ex
RobertJoonas: I don't think it sounds that aggressive 😀 I'll leave it as is. Already a lot of text in this yellow notice...

lib/plausible_web/live/installationv2.ex
RobertJoonas: I considered that but decided not to go through with that because:

1) the current way allows us to not compile the GTM code on CE at all (not just conditionally render)
2) IF we would do this, then it would become inconsistent with the instructions rendering, where every installation type has its own function, and we would still wrap it with `on_ee`.

So I'd personally prefer to keep this small source of truth duplication.

lib/plausible_web/live/verification.ex
RobertJoonas: Good catch! When kicking off the verification, we're indeed looking at the current user's flag status too, so definitely makes sense to do it here as well. https://github.com/plausible/analytics/pull/5692/commits/0da14701565eac7d19b93f8b5876da6f0c79a6b6

lib/plausible_web/live/change_domain_v2.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5692/commits/977255a7ad51f5358d86abf195a555ffdd9f20c7

lib/plausible_web/live/change_domain_v2.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5692/commits/977255a7ad51f5358d86abf195a555ffdd9f20c7

lib/plausible_web/live/installationv2.ex
RobertJoonas: Answered the same question above. Resolving this.

lib/plausible_web/live/installationv2.ex
RobertJoonas: Good call! https://github.com/plausible/analytics/pull/5692/commits/51944418a426614c35e2e70cdd0766b752b7eeac

PR #5693, opened 2025-09-02
Script v2: Show "Plausible not detected" errors with installation type specific recommendations

apata:
### Changes
The problems this PR attempts to solve are
1) weird message and recommendation in the case of unknown error 
2) too many unknown errors logged in sentry

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/installation_support/verification/diagnostics.ex
+++ b/lib/plausible/installation_support/verification/diagnostics.ex
@@ -188,7 +188,7 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
                                message:
                                  "We couldn't verify your website at <%= @attempted_url %>",
                                recommendation:
-                                 "Our verification tool encountered a network error while trying to verify your website. Please verify your integration manually",
+                                 "Accessing the website resulted in a network error. Please verify your installation manually",
                                url: @verify_manually_url
                              })
 
@@ -201,13 +201,13 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
     |> struct!(data: %{offer_custom_url_input: true})
   end
 
-  @error_non_200_page_response Error.new!(%{
-                                 message:
-                                   "We couldn't verify your website at <%= @attempted_url %>",
-                                 recommendation:
-                                   "Our verification tool encountered a <%= @page_response_status %> error. Please check for anything that might be blocking it from reaching your site, like a firewall, authentication requirements, or CDN rules. If you'd prefer, you can skip this and verify your integration manually",
-                                 url: @verify_manually_url
-                               })
+  @error_unexpected_page_response Error.new!(%{
+                                    message:
+                                      "We couldn't verify your website at <%= @attempted_url %>",
+                                    recommendation:
+                                      "Accessing the website resulted in an unexpected status code <%= @page_response_status %>. Please check for anything that might be blocking us from reaching your site, like a firewall, authentication requirements, or CDN rules. If you'd prefer, you can skip this and verify your installation manually",
+                                    url: @verify_manually_url
+                                  })
 
   def interpret(
         %__MODULE__{
@@ -218,12 +218,13 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
         _expected_domain,
         url
       )
-      when is_binary(url) and page_response_status not in [200, nil] and
+      when is_binary(url) and not is_nil(page_response_status) and
+             (page_response_status < 200 or page_response_status >= 300) and
              plausible_is_on_window != true and
              plausible_is_initialized != true do
     attempted_url = shorten_url(url)
 
-    @error_non_200_page_response
+    @error_unexpected_page_response
     |> error(attempted_url: attempted_url, page_response_status: page_response_status)
     |> struct!(data: %{offer_custom_url_input: true})
   end
--- a/lib/plausible/installation_support/verification/diagnostics.ex
+++ b/lib/plausible/installation_support/verification/diagnostics.ex
@@ -64,8 +64,6 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
   @spec interpret(t(), String.t(), String.t()) :: Result.t()
   def interpret(
         %__MODULE__{
-          plausible_is_on_window: true,
-          plausible_is_initialized: true,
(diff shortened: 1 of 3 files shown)


lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -234,7 +218,8 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
         _expected_domain,
         url
       )
-      when is_binary(url) and page_response_status != 200 and plausible_is_on_window != true and
+      when is_binary(url) and page_response_status not in [200, nil] and
```
RobertJoonas: This reminded me - Karl proposed to consider 202 a successful response here. Otherwise, the recommendation to the user would say we encountered a 202 error, which is not really an error. wdyt?

test/plausible/installation_support/verification/diagnostics_test.exs
```diff
@@ -252,28 +278,159 @@ defmodule Plausible.InstallationSupport.Verification.DiagnosticsTest do
...
+                               "https://plausible.io/docs/troubleshoot-integration#how-to-manually-check-your-integration"
+                           }
+                         ]
+                       } = Diagnostics.interpret(diagnostics, expected_domain, url_to_verify)
+      end
     end
   end
+
+  # creates a randomized diagnostics struct, and overwrites only the fields defined in opts
+  defp randomized_diagnostics(opts) do
```
RobertJoonas: I think randomizing the irrelevant diagnostics makes the tests harder to read and reason about, also introducing nondeterminism. I suggest we stick to using the `%Diagnostics{}` struct where the unspecified fields fall back to `nil`.

RobertJoonas (approved):

test/plausible/installation_support/verification/diagnostics_test.exs
apata: Thanks! 

The problem randomized_diagnostics() tries to solve is detecting what's actually relevant for the particular interpretation to occur. As I discovered, the diagnostics objects in the tests were insufficiently defined: these majority null-values structs would never occur in the real world. 

Regarding readability, is it the length of the private function? The cases themselves didn't change much. 

What came to mind though is your earlier suggestion to mark the debug-only (not checked in interpretation) values, something like

```ex
  # In this struct
  # 1) nil value means the result is indeterminate (e.g. we didn't even get to the part where response_status is set)
  # 2) _ prefix means the variable is not used in interpretation: it's for debugging purposes
  defstruct selected_installation_type: nil,
            disallowed_by_csp: nil,
            tracker_is_in_html: nil,
            plausible_is_on_window: nil,
            plausible_is_initialized: nil,
            diagnostics_are_from_cache_bust: nil,
            test_event: nil,
            response_status: nil,
            service_error: nil,
            _plausible_version: nil,
            _plausible_variant: nil,
            _cookies_consent_result: nil,
            _attempts: nil
```

RobertJoonas: I like your suggestion to prefix debug fields with `_`! Let's do that 👍

> Regarding readability, is it the length of the private function?

Not really. It's more the randomness aspect in unit tests that I dislike, and I can see it becoming a maintenance burden.

> As I discovered, the diagnostics objects in the tests were insufficiently defined: these majority null-values structs would never occur in the real world.

As long as these unit tests go through all the code branches in `diagnostics.ex`, I think it's not critical to mimic real world cases (i.e. full Diagnosics struct that can happen in the real world). But I don't mind if they are.

lib/plausible/installation_support/verification/diagnostics.ex
apata: https://github.com/plausible/analytics/pull/5693/commits/b98c639b632f1c4be1a51f4b49081e8bd285bf09

test/plausible/installation_support/verification/diagnostics_test.exs
apata: I removed the randomizer logic: https://github.com/plausible/analytics/pull/5693/commits/ccd3d73de1ac84419458183c87be4e28ce963660

I will mark the debug values in another pure refactor PR.

However, I feel the need to point out that random inputs in tests are already the norm in unit tests. As an example, any time new_site() is called, the site has a random domain and ID. 

All the features are built to handle various input, and the tests are already built to assert in such a way as not to depend on these exact values. 

Is this a maintenance burden? 

Yes, probably it is, but it's a burden that's worth it to help discover issues with unexpected inputs. Would you be more or less confident in the service if you knew that every single time the tests ran, new random input combinations were used for all the values that the test case is not specifically depending on? 

By accepting this burden, we can actually also improve unit test readability. There'd be no need to craft, read, and validate input values that do not matter for a particular test. 

For example, I doubt it's necessary for `example.com` to appear ~1400 times in tests. 
The real statement these tests are trying to make, I think, is that given `any_domain()` _and some other specific input_, the feature works. 

Even if any_domain() were to be the standard (and I'm not saying it should be the standard, just to complete the example), there'd still be some unit tests where explicit domain values are given, to test the validators, serializers, or where it's just inconvenient to assert without using an exact domain. 

I hope this clears up my position a little bit, thanks!

RobertJoonas: > For example, I doubt it's necessary for example.com to appear ~1400 times in tests.
The real statement these tests are trying to make, I think, is that given any_domain() and some other specific input, the feature works.

@apata that's not quite the case. Site domains are not randomized in tests - they're auto-incremented by every test:

* é-0.example.com
* é-1.example.com
* é-2.example.com

That's because async tests are using a shared ClickHouse database, so every time we're building ClickHouse events/sessions in a test and want to assert on those, we can just use the site specific to that test.

Should every test have their own CH, using `example.com` would be completely fine. The domain names in the tests are not trying to increase test coverage. They're just a smart way to make the test suite run faster and more performant.

Coming back to the randomness in this PR though - I get it that we're randomizing the fields that "should not matter", and the tests should always pass regardless. But now imagine that some code change made one of the "irrelevant fields" make a difference, and out of 40 different random combinations there are 10 (due to that specific field) that make the test fail. Suddenly we have a flaky test that fails 75% of the time. This is what I'm against.

PR #5694, opened 2025-09-02
Stats: Rebuild session smearing for timeseries

macobo:
When graphing timeseries, **Session smearing** refers to counting sessions in _each_ of the time buckets a session was active in instead of only when it ended. This works well for visitors/visit graphs, but for visit duration metric it was causing severe over-counting under each value.

image

This PR:
1. Refactors session smearing to only work for visits/visitors metrics and no longer with visit_duration
2. Fixes a bug with visits metric when smearing

To facilitate this, internals of the query building are revamped to allow joining more than 1 event query with 1 session query. Some business logic was also moved to QueryOptimizer.

Basecamp ref: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8546432041#__recording_9025673227

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -28,7 +28,8 @@ defmodule Plausible.Stats.Query do
             site_id: nil,
             site_native_stats_start_at: nil,
             # Contains information to determine how to combine legacy and new time on page metrics
-            time_on_page_data: %{}
+            time_on_page_data: %{},
+            sql_join_type: :left
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{DateTimeRange, Filters, Imported, Legacy, Comparisons}
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -29,7 +29,8 @@ defmodule Plausible.Stats.Query do
             site_native_stats_start_at: nil,
             # Contains information to determine how to combine legacy and new time on page metrics
             time_on_page_data: %{},
-            sql_join_type: :left
+            sql_join_type: :left,
+            smear_session_metrics: false
 
   require OpenTelemetry.Tracer, as: Tracer
   alias Plausible.Stats.{DateTimeRange, Filters, Imported, Legacy, Comparisons}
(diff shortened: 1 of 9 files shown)


lib/plausible/stats/query.ex
```diff
@@ -28,7 +28,9 @@ defmodule Plausible.Stats.Query do
             site_id: nil,
             site_native_stats_start_at: nil,
             # Contains information to determine how to combine legacy and new time on page metrics
-            time_on_page_data: %{}
+            time_on_page_data: %{},
+            sql_join_type: :left,
```
ukutaht, line 32: It's become a pattern to default a query field to something and then override it in the query building/optimizing phase. Both of these fields are an example where the initial value is tentative and later on it might get changed. I think it would be more solid to make them kind of enums where they start out in an `:tbd` state by default. Then, at some point in the pipeline it would go from that value to the value we determine. WDYT?

lib/plausible/stats/query_optimizer.ex
```diff
@@ -303,4 +318,19 @@ defmodule Plausible.Stats.QueryOptimizer do
       |> DateTimeRange.to_timezone("Etc/UTC")
     end
   end
+
+  # Normally we can always LEFT JOIN as this is more performant and tables
+  # are expected to contain the same dimensions.
+
+  # The only exception is using the "time:minute" dimension where the sessions
```
ukutaht, line 325: Why doesn't the same logic apply to `time:hour` dimension? In other words, we smear in both `time:minute` and `time:hour` cases but only use FULL JOIN for `time:minute`, and I don't get the reason behind it.

macobo: This logic isn't new - it was just moved from TableDecider to a better place to remove business logic from within Ecto Query building. Moving it was a due to the join refactor that needed to happen.

It's indeed likely that this was subtly broken when this behavior was introduced in https://github.com/plausible/analytics/pull/5230

I'm slightly wary of changing this in _this_ PR due to potential untested issues creeping in. I can tackle this in a follow-up though.

lib/plausible/stats/query.ex
macobo: Both `sql_join_type` and `smear_session_metrics` cases have clear defaults (no/left) and we mostly leave them alone except in exceptional cases. Adding a third TBD state would just increase the amount of code involved but not clear things up IMO.

There's a case for using a map, naming it `internals` or something and in the few places these values are used doing a `Map.get(query.internals, :key, default)` but this would lose the benefit of having everything listed out in one place.

All this to say: I think this is fine as-is. :thinking:

lib/plausible/stats/query_optimizer.ex
ukutaht: makes sense as a follow-up

PR #5696, opened 2025-09-03
Add top goal conversions to weekly/monthly e-mail digest

aerosol:
### Changes

Includes top goal conversions (unique visitors) in weekly/monthly e-mail report

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/email_report.ex
+++ b/lib/plausible/stats/email_report.ex
@@ -1,13 +1,14 @@
 defmodule Plausible.Stats.EmailReport do
   @moduledoc """
-  This module exposes a `get/2` function that returns a map
-  of stats needed for email reports. These stats include:
+  This module exposes functions that return a map of stats needed for email reports.
+  These stats include:
 
   * Total pageviews
   * Unique visitors
   * Bounce rate
   * A list of Top 5 sources (excluding "Direct / None")
   * A list of Top 5 pages
+  * A list of Top 5 goals (when goals exist)
 
   where total pageviews, unique visitors, and bounce rate
   also include the change compared to previous period.
@@ -18,10 +19,26 @@ defmodule Plausible.Stats.EmailReport do
 
   @aggregate_metrics [:pageviews, :visitors, :bounce_rate]
 
-  def get(site, query) do
-    aggregate_and_compare(site, query)
+  def get_for_period(site, period, date) when period in ["7d", "month"] do
+    {:ok, query} =
+      Query.build(
+        site,
+        :internal,
+        %{
+          "site_id" => "#{site.id}",
+          # metrics will be overridden in the pipeline that follows, and the build interface forces us to pass something
+          "metrics" => ["visitors"],
+          "date_range" => period,
+          "date" => date
+        },
+        %{}
+      )
+
+    site
+    |> aggregate_and_compare(query)
     |> put_top_5_pages(site, query)
     |> put_top_5_sources(site, query)
+    |> put_top_5_goals(site, query)
   end
 
   defp aggregate_and_compare(site, query) do
@@ -35,13 +52,12 @@ defmodule Plausible.Stats.EmailReport do
 
     @aggregate_metrics
     |> Enum.with_index()
-    |> Enum.map(fn {metric, idx} ->
+    |> Enum.into(%{}, fn {metric, idx} ->
       value = Enum.at(result.metrics, idx)
       change = Enum.at(result.comparison.change, idx)
 
       {metric, %{value: value, change: change}}
     end)
-    |> Enum.into(%{})
   end
 
   defp put_top_5_pages(stats, site, query) do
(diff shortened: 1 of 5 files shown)


lib/plausible/stats/email_report.ex
```diff
@@ -60,4 +76,37 @@ defmodule Plausible.Stats.EmailReport do
...
+  defp put_top_5_goals(stats, site, query) do
+    date_str = Date.to_iso8601(query.now)
+    period_str = if query.period == "month", do: "month", else: "7d"
+
+    {:ok, q} =
+      Query.build(
+        site,
+        :internal,
+        %{
+          "site_id" => "#{site.id}",
```
macobo:
```suggestion
          "site_id" => site.domain,
```

The "site_id" we expose in the API is actually the domain (https://plausible.io/docs/stats-api#site_id-). Misnomer :sweat_smile:

lib/plausible/stats/email_report.ex
```diff
@@ -60,4 +76,37 @@ defmodule Plausible.Stats.EmailReport do
...
+          "metrics" => ["visitors"],
+          "dimensions" => ["event:goal"],
+          "date_range" => period_str,
+          "date" => date_str,
```
macobo: `date` so far has only been used in tests and isn't part of the public API.

Instead of passing "month" and "7d", could we pass in an array under date_range? E.g. `["2025-01-01", "2025-01-31"]`

lib/plausible/stats/email_report.ex
```diff
@@ -18,10 +19,26 @@ defmodule Plausible.Stats.EmailReport do
...
 
-  def get(site, query) do
-    aggregate_and_compare(site, query)
+  def get_for_period(site, period, date) when period in ["7d", "month"] do
+    {:ok, query} =
+      Query.build(
+        site,
+        :internal,
+        %{
+          "site_id" => "#{site.id}",
```
macobo:
```suggestion
          "site_id" => site.domain,
```

lib/plausible/stats/email_report.ex
```diff
@@ -60,4 +76,37 @@ defmodule Plausible.Stats.EmailReport do
...
+          "pagination" => %{"limit" => 5}
+        },
+        %{}
+      )
+
+    goals =
+      site
+      |> Plausible.Stats.query(q)
+      |> Map.fetch!(:results)
+      |> Enum.map(fn %{metrics: [visitors], dimensions: [goal_name]} ->
```
macobo, line 103: Q: Could we simplify the templates by removing these maps and using the response directly in the templates? Fewer intermediary abstractions.

lib/plausible/stats/email_report.ex
aerosol: it works no matter what we pass :shrug:

aerosol: https://github.com/plausible/analytics/pull/5696/commits/5b7eb0165c1748529c0a505f0925c32257f065dc

lib/plausible/stats/email_report.ex
aerosol: https://github.com/plausible/analytics/pull/5696/commits/5b7eb0165c1748529c0a505f0925c32257f065dc

lib/plausible/stats/email_report.ex
macobo: Yup, it's not validated in this flow but required in the schema. Still, it being correct should help avoid future mistakes.

lib/plausible/stats/email_report.ex
aerosol: > Instead of passing "month" and "7d", could we pass in an array under date_range?

I wanted to make sure this is the same what the dashboard is showing. `month` and `7d` sounded more reassuring in that regard. Reason being, people will click e-mail report link and end-up on the dashboard - any discrepancy is going to raise questions. WYT?

lib/plausible/stats/email_report.ex
aerosol: idk, in templates I don't see anything wrong with calling fields by name instead of matching arbitrary list positions

macobo: My point was that this could both remove code and allow to remove an unneeded abstraction from the code. See https://github.com/plausible/analytics/commit/91e2400207ff1bf31a71c8901c63457445e224cd

That said, it's not a blocker.

lib/plausible/stats/email_report.ex
```diff
@@ -18,10 +19,27 @@ defmodule Plausible.Stats.EmailReport do
...
+      Query.build(
+        site,
+        :internal,
+        %{
+          # site_id parameter is required, but it doesn't matter what we pass here since the query is executed against a specific site later on
+          "site_id" => site.domain,
+          # metrics will be overridden in the pipeline that follows, and the build interface forces us to pass something
+          "metrics" => ["visitors"],
+          "date_range" => period,
+          "date" => date
```
macobo, line 33: "date" also used here. Prefer to set `date_range` instead.

lib/plausible/stats/email_report.ex
aerosol: My point was exactly to not do that because you can't be sure what's under that `[first_element_match]` as it's completely separated from its implicit context

lib/plausible/stats/email_report.ex
aerosol: Ah, existing test breaks without it, can you tell why? :thinking:

macobo: It's due to `put_last_week_date` which populates the "date" parameter. I think the fix is to make put_last_month_date/put_last_week_date set a `date_range` instead of date and use that over relative time ranges when constructing the query.

aerosol:
closing in favor of https://github.com/plausible/analytics/pull/5711

PR #5701, opened 2025-09-04
Script v2: Rate limit tech detection

apata:
### Changes

Prevents excessive browserless calls to detect technology.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/installation_support/detection/checks.ex
+++ b/lib/plausible/installation_support/detection/checks.ex
@@ -48,7 +48,7 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
   @first_slowdown_ms 1000
   def run_with_rate_limit(url, data_domain, opts \\ []) do
     case Plausible.RateLimit.check_rate(
-           "site_detection_#{data_domain}",
+           "site_detection:#{data_domain}",
            :timer.minutes(60),
            10
          ) do
--- a/lib/plausible/installation_support/detection/checks.ex
+++ b/lib/plausible/installation_support/detection/checks.ex
@@ -43,4 +43,26 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
       state.url
     )
   end
+
+  @unthrottled_checks 3
+  @first_slowdown_ms 1000
+  def run_with_rate_limit(url, data_domain, opts \\ []) do
+    case Plausible.RateLimit.check_rate(
+           "site_detection_#{data_domain}",
+           :timer.minutes(60),
+           10
+         ) do
+      {:allow, count} when count <= @unthrottled_checks ->
+        {:ok, run(url, data_domain, opts)}
+
+      {:allow, count} when count > @unthrottled_checks ->
+        # slowdown steps 1x, 4x, 9x, 16x, ...
+        slowdown_ms = @first_slowdown_ms * (count - @unthrottled_checks) ** 2
+        :timer.sleep(slowdown_ms)
+        {:ok, run(url, data_domain, opts)}
+
+      {:deny, limit} ->
+        {:error, {:rate_limit_exceeded, limit}}
+    end
+  end
 end
--- a/test/plausible_web/live/change_domain_v2_test.exs
+++ b/test/plausible_web/live/change_domain_v2_test.exs
@@ -290,7 +290,7 @@ defmodule PlausibleWeb.Live.ChangeDomainV2Test do
         # exceed the rate limit for site detection
         Plausible.RateLimit.check_rate(
           Plausible.RateLimit,
-          "site_detection_#{new_domain}",
+          "site_detection:#{new_domain}",
           :timer.minutes(60),
           1,
           100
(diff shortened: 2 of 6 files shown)


test/plausible_web/live/change_domain_v2_test.exs
```diff
@@ -24,6 +24,9 @@ defmodule PlausibleWeb.Live.ChangeDomainV2Test do
           [{192, 168, 1, 2}]
         end)
 
+        # prevent rate limit from kicking in in tests
+        :ets.delete_all_objects(Plausible.RateLimit)
```
apata: This is problematic, because it seems to impact other concurrently running tests.

apata: However, I can't be specific to the object key because there's a random element in the mix in creating the full_key (see Plausible.RateLimit check_rate implementation). 

```
  defp clear_rate_limit(domain), do:
    :ets.delete(Plausible.RateLimit, {"site_detection_#{domain}", :timer.minutes(60)})
```

lib/plausible/installation_support/detection/checks.ex
```diff
@@ -43,4 +43,26 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
       state.url
     )
   end
+
+  @unthrottled_checks 3
+  @first_slowdown_ms 1000
+  def run_with_rate_limit(url, data_domain, opts \\ []) do
+    case Plausible.RateLimit.check_rate(
+           "site_detection_#{data_domain}",
```
RobertJoonas: Nit: Let's use `:` for separating the dynamic part of the key (consistency with other rate limiters)

```suggestion
           "site_detection:#{data_domain}",
```

test/plausible_web/live/change_domain_v2_test.exs
RobertJoonas: I think we don't need to disable it because every test uses a site with a unique domain. The issue in this test file specifically is that there are 7 tests renaming the domain to `new-example.com`, and we end up rate-limiting this specific domain. What we can do is simply make sure it's always unique. And since we know that `site.domain` is unique, we can just use that.

```
new_domain = "abc.#{site.domain}"
```

I just tried doing that, and there's a failure that follows, in `assert_patch`:

```
** (ArgumentError) expected PlausibleWeb.Live.ChangeDomainV2 to patch to
"/abc.é-16.example.com/change-domain-v2/success", but got a patch to
"/abc.%C3%A9-16.example.com/change-domain-v2/success"
```

To fix, we can call `URI.encode_www_form` on `new_domain` in the assertion.

test/plausible_web/live/change_domain_v2_test.exs
```diff
@@ -279,6 +282,47 @@ defmodule PlausibleWeb.Live.ChangeDomainV2Test do
              )
     end
 
+    @tag :ee_only
+    test "ratelimit is respected: browserless request isn't made and the notice is generic", %{
+      conn: conn,
+      site: site
+    } do
+      capture_log(fn ->
```
RobertJoonas: Nit: Since we're not asserting on log output, we could also move capture log into the test tag.

```
@tag ee_only: true, capture_log: true
```

There's another test in this file where we could do the same.

test/plausible_web/live/installationv2_test.exs
```diff
@@ -18,6 +18,13 @@ defmodule PlausibleWeb.Live.InstallationV2Test do
     :ok
   end
 
+  on_ee do
+    setup do
+      :ets.delete_all_objects(Plausible.RateLimit)
+      :ok
+    end
+  end
+
```
RobertJoonas: In light of my other comment - this is not needed I think?

lib/plausible/installation_support/detection/checks.ex
apata: Thanks, I refactored site_verification_ now as well.

test/plausible_web/live/change_domain_v2_test.exs
apata: Thanks! This gets formatted to `@tag(ee_only: true, capture_log: true)`, which doesn’t compile

RobertJoonas: Hmm doesn't happen to me 🤨 Also, existing example here: https://github.com/plausible/analytics/blob/db448d740470441d1cc30f99009d54c135a9c3ec/test/plausible_web/controllers/site_controller_test.exs#L1026

Note that it's a keyword list, with `true` values for both keys. The formatter does put parentheses around it If you do `@tag :ee_only, :capture_log` (two separate arguments).

RobertJoonas (approved):

test/plausible_web/live/change_domain_v2_test.exs
apata: Ah, thanks! I had tried `@tag :ee_only, capture_log: true` and missed the existing examples.

PR #5705, opened 2025-09-08
Fix: time series trimming

macobo:
https://github.com/plausible/analytics/pull/5699 accidentally broke time series trimming on main graph

It tried to remove a seemingly redundant Query.optimize call, but this was implicitly needed for time labels functionality. :facepalm: 

Introduces tooling lifted from https://github.com/plausible/analytics/pull/5680 to test this so it doesn't regress anymore.

### DIFF
--- a/lib/plausible/stats/filters/query_parser.ex
+++ b/lib/plausible/stats/filters/query_parser.ex
@@ -31,7 +31,8 @@ defmodule Plausible.Stats.Filters.QueryParser do
       if now do
         {now, DateTime.shift_zone!(now, site.timezone) |> DateTime.to_date()}
       else
-        {DateTime.utc_now(:second), today(site)}
+        now = Plausible.Stats.Query.Test.get_fixed_now()
+        {now, today(now, site)}
       end
 
     with :ok <- JSONSchema.validate(schema_type, params),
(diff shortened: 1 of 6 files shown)


lib/plausible/stats/filters/query_parser.ex
```diff
@@ -289,7 +290,7 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
-  defp today(site), do: DateTime.now!(site.timezone) |> DateTime.to_date()
+  defp today(now, site), do: now |> DateTime.shift_zone!(site.timezone) |> DateTime.to_date()
```
RobertJoonas: This function body now represents exactly what happens when `now` is given as an argument to `Query.parse/4`. We could leave the conditional only on `now` (not `date`), and get rid of the `today` function as well:

```elixir
now =  if now, do: now, else: Plausible.Stats.Query.Test.get_fixed_now()
date = now |> DateTime.shift_zone!(site.timezone) |> DateTime.to_date()
```

RobertJoonas (approved):

lib/plausible/stats/filters/query_parser.ex
macobo: https://github.com/plausible/analytics/pull/5705/commits/2d13d1cdddec3c4346acfe3ef7dfc8f71895b5ed

PR #5707, opened 2025-09-08
Script v2: Change priority of CSP related error for verifying manual installations

apata:
### Changes
If the user is verifying a manual installation, CSP error must only be shown after they've actually tried to add the snippet to their site. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/installation_support/verifier-v2.js
+++ b/tracker/installation_support/verifier-v2.js
@@ -32,7 +32,6 @@ async function verifyPlausibleInstallation(options) {
   const { stopRecording, getInterceptedFetch } = startRecordingEventFetchCalls()
 
   const {
-    trackerIsInHtml,
     plausibleIsInitialized,
     plausibleIsOnWindow,
     plausibleVersion,
@@ -41,10 +40,10 @@ async function verifyPlausibleInstallation(options) {
     cookiesConsentResult,
     error: testPlausibleFunctionError
   } = await testPlausibleFunction({
-    isTrackerInHtml: () => isInHtml(trackerScriptSelector),
     timeoutMs,
     debug
   })
+  const trackerIsInHtml = isInHtml(trackerScriptSelector)
 
   if (testPlausibleFunctionError) {
     log(
@@ -176,9 +175,8 @@ function getPlausibleVariant() {
   return window.plausible?.s
 }
 
-async function testPlausibleFunction({ timeoutMs, debug, isTrackerInHtml }) {
+async function testPlausibleFunction({ timeoutMs, debug }) {
   return new Promise((_resolve) => {
-    let trackerIsInHtml = isTrackerInHtml()
     let plausibleIsOnWindow = isPlausibleOnWindow()
     let plausibleIsInitialized = isPlausibleInitialized()
     let plausibleVersion = getPlausibleVersion()
@@ -189,7 +187,6 @@ async function testPlausibleFunction({ timeoutMs, debug, isTrackerInHtml }) {
       engineLifecycle: 'not-started'
     }
     let timeout = null
-    let trackerScriptPollInterval = null
     let plausibleOnWindowPollInterval = null
     let plausibleInitializedPollInterval = null
     let testEventPollInterval = null
@@ -198,7 +195,6 @@ async function testPlausibleFunction({ timeoutMs, debug, isTrackerInHtml }) {
 
     const resolve = (overrides) => {
       clearTimeout(timeout)
-      clearInterval(trackerScriptPollInterval)
       clearInterval(plausibleOnWindowPollInterval)
       clearInterval(plausibleInitializedPollInterval)
       clearInterval(testEventPollInterval)
@@ -208,7 +204,6 @@ async function testPlausibleFunction({ timeoutMs, debug, isTrackerInHtml }) {
 
       resolved = true
       _resolve({
-        trackerIsInHtml,
         plausibleIsOnWindow,
         plausibleIsInitialized,
         plausibleVersion,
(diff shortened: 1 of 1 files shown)


lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -4,9 +4,11 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
   """
   require Logger
 
-  # in this struct, nil means indeterminate
+  # In this struct
```
aerosol, line 7: You can define struct fields as atoms, e.g. `:selected_installation_type` which is a shorter equivalent. Not that you should, just saying.

lib/plausible/installation_support/verification/diagnostics.ex
```diff
@@ -150,6 +152,21 @@ defmodule Plausible.InstallationSupport.Verification.Diagnostics do
...
+          tracker_is_in_html: false,
+          selected_installation_type: selected_installation_type,
+          plausible_is_on_window: plausible_is_on_window,
+          plausible_is_initialized: plausible_is_initialized,
+          service_error: nil
+        },
+        _expected_domain,
+        _url
+      )
+      when selected_installation_type in ["manual", nil] and plausible_is_on_window != true and
```
aerosol, line 168: cosmetics but `!plausible_is_on_window` should suffice here

test/plausible/installation_support/verification/diagnostics_test.exs
```diff
@@ -158,6 +158,33 @@ defmodule Plausible.InstallationSupport.Verification.DiagnosticsTest do
                      } = Diagnostics.interpret(diagnostics, expected_domain, url_to_verify)
     end
 
+    test "returns error when the snippet is not found for manual installation method, it has priority over CSP-related error" do
+      expected_domain = "example.com"
+      url_to_verify = "https://#{expected_domain}"
+
+      diagnostics =
```
aerosol: Ah so you went with tests against pure structs. Is there a integration test confirming that this set of field values is possible to achieve?

aerosol (approved):

test/plausible/installation_support/verification/diagnostics_test.exs
```diff
@@ -158,6 +158,33 @@ defmodule Plausible.InstallationSupport.Verification.DiagnosticsTest do
                      } = Diagnostics.interpret(diagnostics, expected_domain, url_to_verify)
     end
 
+    test "returns error when the snippet is not found for manual installation method, it has priority over CSP-related error" do
+      expected_domain = "example.com"
+      url_to_verify = "https://#{expected_domain}"
```
aerosol, line 163: Those two don't make any difference, so maybe skip in test?

test/plausible/installation_support/verification/diagnostics_test.exs
apata: Thanks! There's not. You'd be right to point out that this creates false confidence. I tried to manage that by having the rest of the struct randomised , to make sure that _exactly_ these fields determine the result, but reverted it (https://github.com/plausible/analytics/pull/5693/commits/ccd3d73de1ac84419458183c87be4e28ce963660, why: https://github.com/plausible/analytics/pull/5693#discussion_r2318587398).

Now that I think about, it wouldn't be hard to refactor all these cases to live-view tests, where the verifier request output is mocked... I think we should do that.

aerosol: Maybe not as high as live view itself though, but the function doing verification that the LV calls perhaps?

lib/plausible/installation_support/verification/diagnostics.ex
apata: Thanks! I'm getting `** (ArgumentError) invalid expression in guard, ! is not allowed in guards. To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html` if I do that though. Agree that it stands out as weird though.

aerosol: ah you're right, fine

test/plausible/installation_support/verification/diagnostics_test.exs
RobertJoonas: I like the idea by @aerosol! Perhaps along with other refactoring changes we've talked about, @apata? We could follow the example of `LegacyVerification.ChecksTest`

lib/plausible_web/live/installationv2.ex
```diff
@@ -300,17 +300,19 @@ defmodule PlausibleWeb.Live.InstallationV2 do
...
+        :installation
+      )
 
     {:noreply,
      push_navigate(socket,
        to:
          Routes.site_path(socket, :verification, socket.assigns.site.domain,
-           flow: socket.assigns.flow
+           flow: socket.assigns.flow,
+           installation_type: config.installation_type
```
RobertJoonas, line 309: Nothing wrong with passing the query param, but since the `tracker_script_config` object gets updated here we can also read the `installation_type` from that object during the `mount` phase of the Verification LiveView.

Why? That would fix it for the case where the customer is redirected directly into the verification screen (e.g. from [stats_controller.ex](https://github.com/plausible/analytics/blob/f3950b21aa3735a528fb94d410f7f34eae61b82f/lib/plausible_web/controllers/stats_controller.ex#L92)) and we don't know the installation method yet.

tracker/installation_support/verifier-v2.js
```diff
@@ -32,6 +41,7 @@ async function verifyPlausibleInstallation({
     cookiesConsentResult,
     error: testPlausibleFunctionError
   } = await testPlausibleFunction({
+    isTrackerInHtml: () => isInHtml(trackerScriptSelector),
```
RobertJoonas: How about checking this directly in this function without polling, after `await testPlausibleFunction` is finished? Doesn't have to be in this PR.

apata: Thanks! I hadn't even considered that, updated in https://github.com/plausible/analytics/pull/5707/commits/eda2186584637d87876283aabd4de553dc4f89a7.

PR #5710, opened 2025-09-09
APIv2: Make site_id not required when calling `Query.build` with internal schema

macobo:
Feedback from https://3.basecamp.com/5308029/buckets/26383192/messages/9035043082#__recording_9050481657 indicated this causes some confusion. Let's add an exception so it's not required in this scenario.

### DIFF
--- a/extra/lib/plausible_web/live/funnel_settings/form.ex
+++ b/extra/lib/plausible_web/live/funnel_settings/form.ex
@@ -352,7 +352,6 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
         site,
         :internal,
         %{
-          "site_id" => site.domain,
           "date_range" => "month",
           "metrics" => ["pageviews"]
         },
--- a/lib/plausible/segments/segment.ex
+++ b/lib/plausible/segments/segment.ex
@@ -135,7 +135,6 @@ defmodule Plausible.Segments.Segment do
         site,
         :internal,
         %{
-          "site_id" => site.domain,
           "metrics" => ["visitors"],
           "date_range" => "7d",
           "filters" => filters
--- a/lib/plausible/stats/filters/query_parser.ex
+++ b/lib/plausible/stats/filters/query_parser.ex
@@ -29,6 +29,13 @@ defmodule Plausible.Stats.Filters.QueryParser do
   def parse(site, schema_type, params, now \\ nil) when is_map(params) do
     now = now || Plausible.Stats.Query.Test.get_fixed_now()
     date = now |> DateTime.shift_zone!(site.timezone) |> DateTime.to_date()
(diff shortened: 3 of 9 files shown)


aerosol:
Maybe enforce it matches the site's domain provided to Stats.query instead? It's going to get even more confusing otherwise.

macobo:
Not sure how and why that would help?

We have two cases:
- Public API, where we get the site via plug in the first place using the same site_id param in PlausibleWeb.Plugs.AuthorizePublicAPI
- Internal usecases where as you correctly pointed out the `site` and `site_id` param duplicate each other and you need to pass `site` to Query.build(...)

aerosol:
Yes but elsewhere you told me I have to pass `site_id` that's an actual domain to avoid future mistakes (https://github.com/plausible/analytics/pull/5696#discussion_r2317932930). So, in the spirit of future mistake awareness, instead of diverging those contracts we could strengthen it - that way I would've gotten an error passing a numeric `site_id`. You seemed fine with it being redundant, that wasn't my problem either.

macobo:
Numeric values do not pass validation already and get the error `"#/site_id: Type mismatch. Expected String but got Integer."`. You were casting the id to be a string (`"site_id" => "#{site.id}"`).

aerosol:
I mean the confusing issue was no matter what string I would pass there, it wouldn't have mattered. So I wasn't sure if it's used _for anything, at all_. lol

macobo:
1. We do use it in the external/public API - PlausibleWeb.Plugs.AuthorizePublicAPI uses it to figure out what the site to put on `conn` is.
2. It's marked mandatory in the schema because we want the validations and [playground](https://plausible.io/docs/stats-api-playground) to yell if it's missing

You're correct that it's unused for :internal schema and that was confusing. This PR fixes that by removing the redundancy.

It seems you'd prefer to take another approach. Shall we close this?

aerosol:
Yeah, IMHO this would suffice if we like to keep an uniform interface:

```diff
diff --git a/lib/plausible/stats/query_runner.ex b/lib/plausible/stats/query_runner.ex
index d6db51bb6f..151a5f4a32 100644
--- a/lib/plausible/stats/query_runner.ex
+++ b/lib/plausible/stats/query_runner.ex
@@ -33,7 +33,7 @@ defmodule Plausible.Stats.QueryRunner do
     :results
   ]
 
-  def run(site, query) do
+  def run(%Plausible.Site{id: site_id} = site, %Plausible.Stats.Query{site_id: site_id} = query) do
     optimized_query = QueryOptimizer.optimize(query)
 
     %__MODULE__{main_query: optimized_query, site: site}
```

aerosol:
whoops I meant domain lol

aerosol:
Maybe something like https://github.com/plausible/analytics/pull/5713 ?

PR #5712, opened 2025-09-09
Reduce reliance on `Timex` and use native time API where feasible

zoldar:
### Changes

This PR replaces uses of `Timex` with native date and time APIs and moves all the remaining `Timex` references behind `Plausible.Times` and `Plausible.Timezones`. All cases where either `:gap` or `:ambiguous` time computation result can occur are handled gracefully (the wrapped `Timex` `end_of` and `beginning_of` functions handle that internally already).

### TODO

- [x] intentionally break each new function in `Times` and `Timezones` to see if tests cover them at least

### Tests

### DIFF
--- a/lib/plausible/billing/subscriptions.ex
+++ b/lib/plausible/billing/subscriptions.ex
@@ -27,7 +27,7 @@ defmodule Plausible.Billing.Subscriptions do
 
   def expired?(%Subscription{next_bill_date: next_bill_date} = subscription) do
     deleted? = Subscription.Status.deleted?(subscription)
-    expired? = Date.compare(next_bill_date, Date.utc_today()) == :lt
+    expired? = Date.before?(next_bill_date, Date.utc_today())
 
     deleted? && expired?
   end


lib/plausible/times.ex
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  @spec end_of_week(DateTime.t()) :: DateTime.t()
+  def end_of_week(dt) do
+    Timex.end_of_week(dt)
+  end
+
+  @spec end_of_month(DateTime.t()) :: DateTime.t()
+  def end_of_month(dt) do
+    Timex.end_of_week(dt)
```
macobo: Heads up: Typo/bug here!

Q: Would it make sense to use `defdelegate` here until we swap out the implementations?

zoldar: Nice catch, thanks! The focus for now is gracefully handling ambiguities which I haven't got down to yet. 

The first step is ensuring that we either use native API or wrap Timex where possible to keep the logic in one place. 

It's likely I'll deploy this wrapping change in smaller batches to review impact more carefully, I'm not sure yet.

Next, I'll add explicit handling of ambiguous and gap error results where relevant - this also applies those beginning and end functions which won't be simple delegations for long.

Eventually, we might swap out some of those functions for native implementations but I don't think we'd manage that for all. We'll evaluate whether it makes sense on case by case basis later on.

At least that's my idea 😅

macobo: Sounds solid, just thought to shout out the typo as I noticed it. :muscle:

lib/plausible/google/ga4/http.ex
```diff
@@ -269,7 +269,7 @@ defmodule Plausible.Google.GA4.HTTP do
         date =
           case report["rows"] do
             [%{"dimensionValues" => [%{"value" => date_str}]}] ->
-              Timex.parse!(date_str, "%Y%m%d", :strftime) |> NaiveDateTime.to_date()
+              Plausible.Times.parse!(date_str, "%Y%m%d", :strftime) |> NaiveDateTime.to_date()
```
aerosol, line 272: Image

The news outlet we need

zoldar: image

lib/plausible/billing/subscriptions.ex
```diff
@@ -27,7 +27,7 @@ defmodule Plausible.Billing.Subscriptions do
 
   def expired?(%Subscription{next_bill_date: next_bill_date} = subscription) do
     deleted? = Subscription.Status.deleted?(subscription)
-    expired? = Timex.compare(next_bill_date, Date.utc_today()) < 0
+    expired? = Date.compare(next_bill_date, Date.utc_today()) == :lt
```
aerosol: is that an equivalent of `Date.before?` ?

lib/plausible/stats/interval.ex
```diff
@@ -75,15 +75,15 @@ defmodule Plausible.Stats.Interval do
     table =
       with %Date{} = from <- Keyword.get(opts, :from),
            %Date{} = to <- Keyword.get(opts, :to),
-           true <- abs(Timex.diff(from, to, :months)) > 12 do
+           true <- abs(Plausible.Times.diff(from, to, :month)) > 12 do
```
aerosol, line 78: wtf, `DateTime.diff` has no `:month`? :(

aerosol (approved):

lib/plausible/stats/interval.ex
zoldar: Yeah, unfortunately `day` is the highest possible unit when diffing.

lib/plausible/billing/subscriptions.ex
zoldar: Oh good one, I'll change it 👍

zoldar: https://github.com/plausible/analytics/pull/5712/commits/4ac9ae2bb09e8c2b03b373ebe0d46aba6a023cab

PR #5716, opened 2025-09-10
Use capture (instead of bubble) for document.EventListener in custom-events.js

RobertJoonas:
### Changes

This PR tackles an issue from the [tracker script problem log card on Basecamp](https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7911166300) - "Tagged events issue with stopPropagation".

Our custom event tracking listens for events on the document element, but there might be something in between the document and event target in the HTML tree that stops the event from bubbling up to document, i.e. calling `event.stopPropagation()` This PR solves this by using the [capture phase](https://javascript.info/bubbling-and-capturing) instead, basically making sure that our event listener is run before anything else.

Note: Using the capture phase for legacy compat mode is not possible (also explained this in a code comment). This surfaced from failing tests where we assert that our script does not intercept navigation if default is externally prevented. https://github.com/plausible/analytics/blob/8d6d828d1d9e55e95e0e3fb232214e6182d381ba/tracker/test/tagged-events.spec.ts#L527 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/custom-events.js
+++ b/tracker/src/custom-events.js
@@ -9,6 +9,17 @@ var MIDDLE_MOUSE_BUTTON = 1
 var PARENTS_TO_SEARCH_LIMIT = 3
 var fileTypesToTrack = DEFAULT_FILE_TYPES
 
+// We track custom events with eventListeners attached to the `document` in order
+// to capture interactions with elements that get added to the DOM dynamically.
+//
+// Normally we set the eventListeners in the capture phase to make sure our event
+// handlers run before anything else on the page that may call `stopPropagation`.
+// We can safely let our handler be the first one, since it does not intercept the
+// event in any way. However that's not true for compat mode where we call
+// `event.preventDefault()` to delay navigation. Therefore, in compat mode we use
+// eventListeners in the bubble phase.
+var useCapturePhase = !COMPILE_COMPAT
+
 function getLinkEl(link) {
   while (link && (typeof link.tagName === 'undefined' || !isLink(link) || !link.href)) {
     link = link.parentNode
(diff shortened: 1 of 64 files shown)


tracker/test/tagged-events.spec.ts
```diff
@@ -112,6 +112,39 @@ for (const mode of ['legacy', 'web']) {
...
+          {
+            web: { ...DEFAULT_CONFIG},
+            legacy:
+              '<script async src="/tracker/js/plausible.local.tagged-events.js"></script>'
+          },
+          mode
+        ),
+        bodyContent: `
+          <div id="stop-click-propagation">
+            <button class="plausible-event-name=ButtonClick">📥</a>
```
apata: issue: The end tag </a> doesn't match

tracker/src/custom-events.js
```diff
@@ -283,8 +294,8 @@ export function init() {
     }
     
 
-    document.addEventListener('submit', handleTaggedFormSubmitEvent)
-    document.addEventListener('click', handleTaggedElementClickEvent)
-    document.addEventListener('auxclick', handleTaggedElementClickEvent)
+    document.addEventListener('submit', handleTaggedFormSubmitEvent, useCapturePhase)
+    document.addEventListener('click', handleTaggedElementClickEvent, useCapturePhase)
+    document.addEventListener('auxclick', handleTaggedElementClickEvent, useCapturePhase)
```
apata, line 299: I'm a bit worried about unforeseen impact with this change. 

- Let's say the user has tagged a `<a href="https://plausible.io" class="...">Click here</a>`. 
- Let's say their app logic requires them to block the link from navigating and they do so with a listener higher up in the tree that calls stopPropagation() and preventDefault(). 
- Let's say it's not their intent to track these clicks, because from the perspective of the website visitor, nothing happened. 
- If we set the true value, this event will be tracked regardless of preventDefault and stopPropagation, against the intent of the user.

Also, wouldn't setting this to true cause unexpected tracking events to appear? This would potentially mess up stats for those who relied on Plausible document level listener logic.

apata: I think it would be safer to document that for tagged events to work, the event must reach event listeners at the document level - it's an important part of the current Plausible tracker API. 

We can suggest various workarounds if people are having trouble with this. 

Alternatively, we could read `taggedEvents: {useCapturePhase: <boolean>}` from the config.

RobertJoonas: Fair points. I picked up this task because it seemed like an easy fix at first sight, but you're right that it might have unexpected impact. I guess this requires some more investigation and possibly reports from customers to act on. As far as I'm aware, we only have one report of this (HelpScout link in the basecamp card).

The `useCapturePhase` init setting is potentially a nice middle ground, but I wouldn't implement it until there's actual need for it. Let's shelve this for now.

PR #5717, opened 2025-09-10
Script v2: Maybe load installation type from DB

apata:
### Changes

- Fixes the issue when StatsController redirects to :verification without installation_type not known at that point.
- Fixes typo in error case
- Fixes custom_url screen being allowed for non v2 verification
- Adds v2 verification tests
- Adds utility to assert on html that contains single quotes without manually re-encoding chars

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/installation.ex
+++ b/lib/plausible_web/live/installation.ex
@@ -43,7 +43,14 @@ defmodule PlausibleWeb.Live.Installation do
     if PlausibleWeb.Tracker.scriptv2?(site) do
       {:ok,
        redirect(socket,
-         to: Routes.site_path(socket, :installation_v2, site.domain, flow: params["flow"])
+         to:
+           Routes.site_path(
+             socket,
+             :installation_v2,
+             site.domain,
+             [flow: params["flow"], type: params["installation_type"]]
+             |> Keyword.filter(fn {_k, v} -> not is_nil(v) and v != "" end)
+           )
        )}
     else
       flow = params["flow"]
--- a/lib/plausible_web/live/verification.ex
+++ b/lib/plausible_web/live/verification.ex
@@ -34,7 +34,9 @@ defmodule PlausibleWeb.Live.Verification do
 
     super_admin? = Plausible.Auth.is_super_admin?(current_user)
     has_pageviews? = has_pageviews?(site)
-    custom_url_input? = params["custom_url"] == "true"
+
+    custom_url_input? =
+      PlausibleWeb.Tracker.scriptv2?(site, current_user) and params["custom_url"] == "true"
 
     socket =
       assign(socket,
--- a/lib/plausible_web/live/verification.ex
+++ b/lib/plausible_web/live/verification.ex
@@ -233,8 +233,7 @@ defmodule PlausibleWeb.Live.Verification do
         params["installation_type"] in PlausibleWeb.Tracker.supported_installation_types() ->
           params["installation_type"]
 
-        saved_installation_type =
-            PlausibleWeb.Tracker.get_tracker_script_configuration(site)[:installation_type] in @supported_installation_types_atoms ->
+        (saved_installation_type = get_saved_installation_type(site)) in @supported_installation_types_atoms ->
           Atom.to_string(saved_installation_type)
 
         true ->
@@ -242,6 +241,16 @@ defmodule PlausibleWeb.Live.Verification do
       end
     else
       params["installation_type"]
+    end
+  end
+
+  defp get_saved_installation_type(site) do
+    case PlausibleWeb.Tracker.get_tracker_script_configuration(site) do
+      %{installation_type: installation_type} ->
+        installation_type
+
+      _ ->
+        nil
     end
   end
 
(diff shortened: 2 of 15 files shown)


lib/plausible_web/live/installation.ex
```diff
@@ -43,7 +43,14 @@ defmodule PlausibleWeb.Live.Installation do
...
       {:ok,
        redirect(socket,
-         to: Routes.site_path(socket, :installation_v2, site.domain, flow: params["flow"])
+         to:
+           Routes.site_path(
+             socket,
+             :installation_v2,
+             site.domain,
+             [flow: params["flow"], type: params["installation_type"]]
+             |> Keyword.filter(fn {_k, v} -> not is_nil(v) and v != "" end)
```
apata, line 52: note: needed, because otherwise URL has pointless empty params like `?flow=`

lib/plausible_web/live/verification.ex
```diff
@@ -220,6 +222,26 @@ defmodule PlausibleWeb.Live.Verification do
...
+
+        saved_installation_type =
+            PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(site).installation_type in @supported_installation_types_atoms ->
+          Atom.to_string(saved_installation_type)
+
+        true ->
+          PlausibleWeb.Tracker.fallback_installation_type()
+      end
+    else
+      params["installation_type"]
```
apata, line 245: note: This LV is shared with v1 verification as well, and this code path is the original one, no point in changing when it's about to be deleted

lib/plausible_web/live/verification.ex
```diff
@@ -220,6 +222,26 @@ defmodule PlausibleWeb.Live.Verification do
...
+  @supported_installation_types_atoms PlausibleWeb.Tracker.supported_installation_types()
+                                      |> Enum.map(&String.to_atom/1)
+  defp get_installation_type(params, site, current_user) do
+    if PlausibleWeb.Tracker.scriptv2?(site, current_user) do
+      cond do
+        params["installation_type"] in PlausibleWeb.Tracker.supported_installation_types() ->
+          params["installation_type"]
+
+        saved_installation_type =
+            PlausibleWeb.Tracker.get_or_create_tracker_script_configuration!(site).installation_type in @supported_installation_types_atoms ->
```
RobertJoonas: Following up about that Basecamp question.

Although this is an edge case and the customer should always have a `tracker_script_config` object here, it's possible that they don't. E.g. they navigate from their dashboard directly to `plausible.io/example.com/verification` by modifying the URL.

From the fact that we don't have a `recommended_installation_type` here, I agree that creating the `tracker_script_config` object would be wrong. Otherwise, going through the installation -> verification flow the next time, the `tracker_script_config` would exist already and **not** be created with the right defaults.

So I think here we should indeed only **get** the tracker_script_config - not create it. However, that could also be a separate (new) function. We don't necessarily have to split up the existing `get_or_create`.

lib/plausible_web/live/verification.ex
```diff
@@ -42,7 +42,7 @@ defmodule PlausibleWeb.Live.Verification do
         domain: domain,
         has_pageviews?: has_pageviews?,
         component: @component,
-        installation_type: params["installation_type"],
+        installation_type: get_installation_type(params, site, socket.assigns.current_user),
```
RobertJoonas: Nit: Could create a variable `current_user = socket.assigns.current_user` and replace other two occurrences too.

lib/plausible_web/live/verification.ex
RobertJoonas: Aside from the other comment on the same code line:

I don't think we need this extra validation here. There's an Enum validation in the schema field already in `Plausible.Site.TrackerScriptConfiguration`. Would suffice to just check if it's not `nil`.

lib/plausible_web/live/verification.ex
```diff
@@ -149,11 +149,13 @@ defmodule PlausibleWeb.Live.Verification do
...
       domain = socket.assigns.domain
-      installation_type = socket.assigns.installation_type
 
       {:ok, pid} =
         if PlausibleWeb.Tracker.scriptv2?(socket.assigns.site, socket.assigns.current_user) do
-          Verification.Checks.run(socket.assigns.url_to_verify, domain, installation_type,
+          Verification.Checks.run(
+            socket.assigns.url_to_verify,
+            domain,
+            socket.assigns.installation_type,
```
RobertJoonas, line 163: It looks like this will never be passed as `nil` (at least for v2), so we can also clean up this in `Plausible.InstallationSupport.Verification.Diagnostics`:

image

to just `selected_installation_type == "manual"`.

lib/plausible_web/live/verification.ex
apata: Thanks! Supported installation types actually differ by env, but the DB enum doesn't: the current implementation protects from this mismatch causing unexpected bugs.

apata: Thanks! Updated and added tests
https://github.com/plausible/analytics/pull/5717/commits/c2ff828a4f18f7d286d8a4e73edfd9cd1caf7a7b

RobertJoonas: Ah, makes sense!

RobertJoonas (approved):
LGTM! 👍

PR #5725, opened 2025-09-15
Clean up number formatting util

sanne-san:
### Changes

- Deduplicate number_format util
- Add tests for number_format util
- Add number formatting to visitor change percentage in site overview

Spotted on X:
CleanShot 2025-09-15 at 16 42 28@2x

### Tests

### DIFF
--- a/extra/lib/plausible_web/live/customer_support/team.ex
+++ b/extra/lib/plausible_web/live/customer_support/team.ex
@@ -2,7 +2,7 @@ defmodule PlausibleWeb.Live.CustomerSupport.Team do
   @moduledoc """
   Team coordinator LiveView for Customer Support interface.
 
-  Manages tab-based navigation and delegates rendering to specialized 
+  Manages tab-based navigation and delegates rendering to specialized
   components: Overview, Members, Sites, Billing, SSO, and Audit.
   """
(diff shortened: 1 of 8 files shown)


sanne-san:
@ukutaht I put the util in stats_view as we discussed, but now that it's being used in other files, there might be a better place. Any thoughts?

ukutaht:
I've found [PlausibleWeb.TextHelpers](https://github.com/plausible/analytics/blob/1531386b767c6827216ab63d87fa447c46d37bf7/lib/plausible_web/views/text_helpers.ex#L1) which seems to be the place for various formatting utils. I'd suggest moving it there.

One thing though - this function variant:
```elixir
 def number_format(unlimited) when unlimited in [-1, "unlimited", :unlimited] do
    "unlimited"
  end
```

is specific to the billing context and should not be extracted to the new helper function in my opinion. It's a quirk where for account limits like how many team members are allowed, etc we use -1 in the database to represent 'unlimited'. But since this only has meaning in the billing components I think this logic could be unexpected when this helper is used in some other context.

sanne-san:
@ukutaht thanks for the input! I've updated it accordingly.

ukutaht (approved):

PR #5729, opened 2025-09-16
Use dictionary lookup to fetch rollup site IDs

aerosol:
The exact way of dictionary provisioning and maintenance is TBD.

This solves two severe performance bottlenecks with previous implementation: building queries that are likely to hit raw payload and AST depth limits.

We can run this on prod by providing some dictionary ad-hoc first.


aerosol:
Closing in favour of https://github.com/plausible/analytics/pull/5730

PR #5733, opened 2025-09-17
Add endpoints for managing custom props in Sites API

zoldar:
### Changes

This PR adds support for listing custom props from the Sites API as well as adding and removing them from Sites Provisioning API.

Respective documentation PR is at https://github.com/plausible/docs/pull/641

### Tests

### Documentation

### DIFF
--- a/lib/plausible_web/router.ex
+++ b/lib/plausible_web/router.ex
@@ -355,6 +355,7 @@ defmodule PlausibleWeb.Router do
           delete "/goals/:goal_id", ExternalSitesController, :delete_goal
 
           put "/custom-props", ExternalSitesController, :add_custom_prop
+          # Property name can contain forward slashes, hence we match on wildcard here
           delete "/custom-props/*property", ExternalSitesController, :delete_custom_prop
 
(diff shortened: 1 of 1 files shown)


lib/plausible_web/router.ex
```diff
@@ -353,6 +354,9 @@ defmodule PlausibleWeb.Router do
           put "/goals", ExternalSitesController, :find_or_create_goal
           delete "/goals/:goal_id", ExternalSitesController, :delete_goal
 
+          put "/custom-props", ExternalSitesController, :add_custom_prop
+          delete "/custom-props/*property", ExternalSitesController, :delete_custom_prop
```
aerosol, line 359: Is the wildcard used so the property in URL can contain forward slashes?

zoldar: @aerosol correct

zoldar: Added a clarifying comment in https://github.com/plausible/analytics/pull/5733/commits/3fcdc37fe7a95b00b8a87845ae926702431891ba

PR #5735, opened 2025-09-18
Migration: Consolidated views

RobertJoonas:
### Changes

Add migration to introduce consolidated views as special sites with `consolidated: true`.

### DIFF
--- a/priv/repo/migrations/20250916154337_add_consolidated_field_to_sites.exs
+++ b/priv/repo/migrations/20250916154337_add_consolidated_field_to_sites.exs
@@ -8,8 +8,6 @@ defmodule Plausible.Repo.Migrations.AddConsolidatedFieldToSites do
       alter table(:sites) do
         add :consolidated, :boolean, null: false, default: false
       end
-
-      create index(:sites, [:id], where: "consolidated = true")
     end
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20250916154337_add_consolidated_field_to_sites.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.Repo.Migrations.AddConsolidatedFieldToSites do
+  use Ecto.Migration
+
+  def change do
+    alter table(:sites) do
```
aerosol: Probably good to have an index here. Are we going to make consolidated views available in CE? If not, let's put this behind `MigrationUtils.community_edition?`.

aerosol: I retract the index remark, I think postgres will deal with it just fine. We'll maybe add a composite index later on. @zoldar WYT?

aerosol: After some thinking, I'm not sure if index is really necessary at this point. Let's skip for now.

PR #5737, opened 2025-09-18
Change Site schema: Add consolidated (boolean) field

RobertJoonas:
### Changes

Also fixes the data migration referencing the Site.Schema, which it shouldn't do.


lib/plausible/site.ex
```diff
@@ -12,6 +12,7 @@ defmodule Plausible.Site do
   @derive {Jason.Encoder, only: [:domain, :timezone]}
   schema "sites" do
     field :domain, :string
+    field :consolidated, :boolean
```
aerosol: you'll have to wrap it with `on_ee`

PR #5739, opened 2025-09-18
Bootstrap the interface for consolidated sites

RobertJoonas:
### Changes

* Introduces two predicate functions `Sites.regular?/1` and `Sites.consolidated?/1`
* Introduces a `ConsolidatedView` context module (`ee_only`; `/extra` folder) with public functions (and tests):
  * `site_ids/1` - returns `{:ok, site_ids}` for this team's consolidated view if it exists or `{:error, :not_found}` if not. Will start to use cache at a later stage.
  * `enable/1` - (idempotent) create a new site with `consolidated: true` for a given team
  * `disable/1` - (idempotent) delete the consolidated site of the given team
  * `eligible?/1` - just a placeholder for now but should check whether the team can access this feature 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -1,8 +1,8 @@
 defmodule Plausible.ConsolidatedView do
   @moduledoc """
-  Contextual interface for consolidated views, 
+  Contextual interface for consolidated views,
   each implemented as Site object serving as
-  pointers to team's regular sites. 
+  pointers to team's regular sites.
   """
 
   use Plausible
@@ -11,11 +11,6 @@ defmodule Plausible.ConsolidatedView do
 
   alias Plausible.Teams.Team
   alias Plausible.{Repo, Site}
-
-  @spec cv_domain(Team.t()) :: String.t()
-  def cv_domain(%Team{} = team) do
-    "cv-#{team.identifier}"
-  end
 
   @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :upgrade_required}
   def enable(%Team{} = team) do
@@ -28,7 +23,7 @@ defmodule Plausible.ConsolidatedView do
 
   @spec disable(Team.t()) :: :ok
   def disable(%Team{} = team) do
-    from(s in Site, where: s.consolidated and s.domain == ^cv_domain(team))
+    from(s in Site, where: s.consolidated and s.domain == ^make_id(team))
     |> Plausible.Repo.delete_all()
 
     :ok
@@ -36,23 +31,36 @@ defmodule Plausible.ConsolidatedView do
 
   @spec site_ids(Team.t()) :: [pos_integer()] | {:error, :not_found}
   def site_ids(%Team{} = team) do
-    case Repo.get_by(Site, domain: cv_domain(team)) do
+    case get(team) do
       nil -> {:error, :not_found}
       _found -> {:ok, owned_site_ids(team)}
     end
   end
 
-  defp do_enable(%Team{} = team) do
-    cv_domain = cv_domain(team)
+  @spec get(Team.t() | String.t()) :: Site.t() | nil
+  def get(team_or_id)
 
-    case Repo.get_by(Site, domain: cv_domain) do
+  def get(%Team{} = team) do
+    team |> make_id() |> get()
+  end
+
+  def get(id) when is_binary(id) do
+    Repo.get_by(Site, domain: id, consolidated: true)
+  end
+
+  defp do_enable(%Team{} = team) do
+    case get(team) do
       nil ->
-        Site.new_for_team(team, %{consolidated: true, domain: cv_domain})
+        Site.new_for_team(team, %{consolidated: true, domain: make_id(team)})
         |> Repo.insert()
 
       cv ->
         {:ok, cv}
     end
+  end
+
+  defp make_id(%Team{} = team) do
+    team.identifier
   end
 
   # TODO: Only active trials and business subscriptions should be eligible.
--- a/test/plausible/consolidated_view_test.exs
+++ b/test/plausible/consolidated_view_test.exs
@@ -11,7 +11,7 @@ defmodule Plausible.ConsolidatedViewTest do
 
       test "creates and persists a new consolidated site instance", %{team: team} do
         assert {:ok, %Plausible.Site{consolidated: true}} = ConsolidatedView.enable(team)
-        assert Plausible.Repo.get_by(Plausible.Site, domain: ConsolidatedView.cv_domain(team))
+        assert ConsolidatedView.get(team)
       end
 
       test "is idempotent", %{team: team} do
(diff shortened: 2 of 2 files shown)


extra/lib/plausible/consolidated_view.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  use Plausible
+
+  import Ecto.Query
+
+  alias Plausible.Teams.Team
+  alias Plausible.{Repo, Site}
+
+  @spec cv_domain(Team.t()) :: String.t()
+  def cv_domain(%Team{} = team) do
```
aerosol: Let's make this function private and push to the bottom - it's an internal detail. Not sure if we need the `cv_` prefix really, the context is known. It just happens that this will be put into `domain` field for now - that's accidental, because we have to put something in there, but in general it serves as a consolidated view id going forward. So maybe consider something like `defp make_id(team)`?

extra/lib/plausible/consolidated_view.ex
```diff
@@ -0,0 +1,71 @@
...
+  def disable(%Team{} = team) do
+    from(s in Site, where: s.consolidated and s.domain == ^cv_domain(team))
+    |> Plausible.Repo.delete_all()
+
+    :ok
+  end
+
+  @spec site_ids(Team.t()) :: [pos_integer()] | {:error, :not_found}
+  def site_ids(%Team{} = team) do
+    case Repo.get_by(Site, domain: cv_domain(team)) do
```
aerosol: neat. I think we can also make sure the site we're looking for has `consolidated: true` for completeness.

extra/lib/plausible/consolidated_view.ex
```diff
@@ -0,0 +1,71 @@
...
+      cv ->
+        {:ok, cv}
+    end
+  end
+
+  # TODO: Only active trials and business subscriptions should be eligible.
+  # This function should call a new underlying feature module.
+  defp eligible?(%Team{}), do: always(true)
+
+  # TEMPORARY: Will be replaced with `Teams.owned_site_ids` once it starts
```
aerosol, line 70: 👍

test/plausible/consolidated_view_test.exs
```diff
@@ -0,0 +1,72 @@
...
+    use Plausible.DataCase, async: true
+    import Ecto.Query
+    alias Plausible.ConsolidatedView
+
+    describe "enable/1" do
+      setup [:create_user, :create_team]
+
+      test "creates and persists a new consolidated site instance", %{team: team} do
+        assert {:ok, %Plausible.Site{consolidated: true}} = ConsolidatedView.enable(team)
+        assert Plausible.Repo.get_by(Plausible.Site, domain: ConsolidatedView.cv_domain(team))
```
aerosol: To make `cv_domain` opaque we could implement this `ConsolidatedView.get(id)` or `ConsolidatedView.get_site(id)` - whichever name you find more suitable,  mentioned earlier.

So something like   

```elixir
assert {:ok, %Plausible.Site{consolidated: true, domain: consolidated_view_id}}
assert ConsolidatedView.get(consolidated_view_id)
```

This helps us deal with multiple views in the future.

We could also have `get(team)` that would delegate to `get(team.identifier)`

```elixir
def get(%Team{identifier: consolidated_view_id}) do
  get(consolidated_view_id)
end
def get(id) when is_binary(id) do
  Repo.get_by(Site, domain: id, consolidated: true) 
end
```

The `cv-` prefix is unnecessary. We have `sites.consolidated` if we ever query. What's in domain should not have any significance other than satisfying the unique constraint. And that whole confusing mapping between identifier and domain will be encapsulated in a single place. WYT?

test/plausible/consolidated_view_test.exs
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      test "is idempotent", %{team: team} do
+        assert {:ok, s1} = ConsolidatedView.enable(team)
+        assert {:ok, s2} = ConsolidatedView.enable(team)
```
aerosol, line 19: will `{:ok, ^s1}` work here? We could skip what follows then

extra/lib/plausible/consolidated_view.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Plausible.ConsolidatedView do
+  @moduledoc false
```
aerosol:
```suggestion
  @moduledoc """
  Contextual interface for consolidated views, 
  each implemented as Site object serving as
  a pointer to team's regular sites. 
  """
```

aerosol (commented):
nice! comments inline

test/plausible/consolidated_view_test.exs
RobertJoonas: > The cv- prefix is unnecessary. We have sites.consolidated if we ever query. What's in domain should not have any significance other than satisfying the unique constraint. And that whole confusing mapping between identifier and domain will be encapsulated in a single place. WYT?

You're right that the `cv` prefix does not add any functional value. I just figured that since `domain` is (at least currently) a "key value" being used as `site_id` in the Stats API and surfaced on any dashboard URL route, it should make it a bit more evident that we're dealing with a consolidated view, rather than a normal dashboard.

I'm thinking of this more like a UX improvement. I can't think of any downsides to this either, other than the fact that we have to create some internal mapping in code when getting and creating consolidated sites. Sounds like a small price to pay though?

aerosol: Yeah, I see your point, but we can address API level UX at... API level, when the time comes? That may help with overall clarity.

API can even accept something like a literal "rollup" instead of domain, and we'll translate that to current_team.identifier if necessary (for UX sake). 

Keeping UUID as is for now, has the benefit of maybe becoming actual primary key one day (ConslidatedView.id) when we'll introduce multiple views per team.

RobertJoonas: > Keeping UUID as is for now, has the benefit of maybe becoming actual primary key one day (ConslidatedView.id) when we'll introduce multiple views per team

Right. That makes sense!

aerosol: And if `cv-` in front of an UUID is to serve as a reminder that you're not querying a real domain,  UUID alone has the same characteristics - it's just a sequence of weird characters. Expecting users to decipher it, when they already have the identifier in use is maybe a little too hopeful?

test/plausible/consolidated_view_test.exs
RobertJoonas: Doesn't work because `s1` has `team` preloaded, but `s2` doesn't.

extra/lib/plausible/consolidated_view.ex
RobertJoonas: Done. Pushed to the bottom and renamed to `make_id` (the function also does not add te `cv-` prefix anymore). https://github.com/plausible/analytics/pull/5739/commits/3b7feaa2c7f06e778912bd403acc1cb5a655458a

extra/lib/plausible/consolidated_view.ex
RobertJoonas: Done in https://github.com/plausible/analytics/pull/5739/commits/3b7feaa2c7f06e778912bd403acc1cb5a655458a. Using the  `get/1` function now which queries with the `consolidated: true` filter.

aerosol (approved):

PR #5742, opened 2025-09-22
Provide safeguards for regular/consolidated sites distinction

aerosol:
### Changes

This PR attempts to filter out consolidated views (`Sites` with `consolidated=true`) by default. We'll be enabling access as we go, depending on the feature scope required. No existing functionality should be affected at this point. Some missing tests were added BTW. Best to review commit by commit.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/auth/auth.ex
+++ b/lib/plausible/auth/auth.ex
@@ -299,7 +299,7 @@ defmodule Plausible.Auth do
 
   defp find_team_by_site(domain) do
     Repo.one(
-      from s in Plausible.Site,
+      from s in Plausible.Site.regular(),
         inner_join: t in assoc(s, :team),
         where: s.domain == ^domain or s.domain_changed_from == ^domain,
         select: t
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -581,7 +581,7 @@ defmodule Plausible.Teams.Billing do
 
   defp query_team_member_emails(team, pending_ownership_site_ids, exclude_emails) do
     pending_owner_memberships_q =
-      from s in Plausible.Site,
+      from s in Plausible.Site.regular(),
         inner_join: t in assoc(s, :team),
         inner_join: tm in assoc(t, :team_memberships),
         inner_join: u in assoc(tm, :user),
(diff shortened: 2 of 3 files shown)


lib/plausible/sites.ex
```diff
@@ -45,12 +45,24 @@ defmodule Plausible.Sites do
   """
   def shared_link_special_names(), do: @shared_link_special_names
 
-  def get_by_domain(domain) do
-    Repo.get_by(Site, domain: domain)
+  def get_by_domain(domain, opts \\ []) do
```
RobertJoonas, line 48: This function is the same as the bang one below, with the only difference being `Repo.get_by!` (bang) vs `Repo.get_by`. To get rid of the duplication, perhaps we could make a pattern matching assertion in the bang function like:

```elixir
def get_by_domain!(domain, opts \\ []) do
  {:ok, site} = get_by_domain(domain, opts)
  site
end
```

I guess the intent is to crash if the site is not found, which should work nicely?

test/plausible/teams_test.exs
```diff
@@ -449,7 +449,7 @@ defmodule Plausible.TeamsTest do
       assert [%{id: ^site_id1}, %{id: ^site_id2}] = Teams.owned_sites(team, 2)
     end
 
-    test "rerutns empty list for no team" do
```
aerosol: :bow:

lib/plausible/sites.ex
```diff
@@ -406,41 +419,64 @@ defmodule Plausible.Sites do
     )
   end
 
-  def get_for_user!(user, domain, roles \\ [:owner, :admin, :editor, :viewer]) do
+  def get_for_user!(user, domain, opts \\ []) do
```
RobertJoonas, line 421: Very similar suggestion here as for the `get_by_domain` function. The function bodies are almost identical, and duplication could be avoided. Small difference in this one though - the bang one here also preloads the team. But this can also be done after the pattern matching assertion. WDYT?

test/plausible/teams_test.exs
RobertJoonas: 🧑‍🤝‍🧑

lib/plausible/sites.ex
aerosol: Yeah, the intent is not only to crash, but also to raise `Ecto.NoResultsError` (that oftentimes is handled specifically, e.g. translated to 404 in phoenix). This exception also needs a `queryable` field that is probably useful in Sentry and such. That said, I think the duplication is OK to keep. Did 19dff7c068 while at it

lib/plausible/sites.ex
aerosol: I think we can tackle it outside of this PR, don't want to break anything as per https://github.com/plausible/analytics/pull/5742#discussion_r2371674104

PR #5744, opened 2025-09-22
Make `sites.consolidated` column available in CE

aerosol:
### Changes

Keeping it EE-only causes more headache than it's worth


aerosol:
Ref https://github.com/plausible/analytics/pull/5742

PR #5749, opened 2025-09-23
Remove virtual rollups in favour of persisted consolidated sites

RobertJoonas:
### Changes

Get rid of the current virtual rollups and replace them with consolidated views. Note that there's no functionality for creating any consolidated views yet, other than calling `Plausible.ConsolidatedView.enable(team)` manually from `iex -S mix`. The next PR will make it possible to create consolidated views from the CRM.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -35,7 +35,7 @@ defmodule Plausible.ConsolidatedView do
     :ok
   end
 
-  @spec site_ids(Team.t()) :: [pos_integer()] | {:error, :not_found}
+  @spec site_ids(Team.t()) :: {:ok, [pos_integer()]} | {:error, :not_found}
   def site_ids(%Team{} = team) do
     case get(team) do
       nil -> {:error, :not_found}
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -43,6 +43,7 @@ defmodule Plausible.ConsolidatedView do
     end
   end
 
+  @spec native_stats_start_at(Team.t()) :: NaiveDateTime.t() | nil
   def native_stats_start_at(%Team{} = team) do
     q =
       from(sr in Site.regular(),
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -71,8 +71,11 @@ defmodule Plausible.ConsolidatedView do
   defp do_enable(%Team{} = team) do
     case get(team) do
       nil ->
+        native_stats_start_at = native_stats_start_at(team)
+
         team
         |> Site.new_for_team(%{consolidated: true, domain: make_id(team)})
+        |> Site.set_native_stats_start_at(native_stats_start_at)
         |> Repo.insert()
 
       consolidated_view ->
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -20,7 +20,7 @@ defmodule Plausible.ConsolidatedView do
     from s in q, where: s.consolidated == true
   end
 
-  @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :no_sites | :upgrade_required}
+  @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :no_sites}
   def enable(%Team{} = team) do
     with :ok <- ensure_eligible(team), do: do_enable(team)
   end
@@ -90,10 +90,10 @@ defmodule Plausible.ConsolidatedView do
   # TODO: Only active trials and business subscriptions should be eligible.
   # This function should also call a new underlying feature module.
   defp ensure_eligible(%Team{} = team) do
-    cond do
-      always(false) -> {:error, :upgrade_required}
-      Plausible.Teams.owned_sites_count(team) == 0 -> {:error, :no_sites}
-      true -> :ok
+    if Plausible.Teams.owned_sites_count(team) == 0 do
+      {:error, :no_sites}
+    else
+      :ok
     end
   end
 end
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -43,6 +43,17 @@ defmodule Plausible.ConsolidatedView do
     end
   end
 
+  def native_stats_start_at(%Team{} = team) do
+    q =
+      from(sr in Site.regular(),
+        group_by: sr.team_id,
+        where: sr.team_id == ^team.id,
+        select: min(sr.native_stats_start_at)
+      )
+
+    Repo.one(q)
+  end
+
   @spec get(Team.t() | String.t()) :: Site.t() | nil
   def get(team_or_id)
 
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -20,13 +20,9 @@ defmodule Plausible.ConsolidatedView do
     from s in q, where: s.consolidated == true
   end
 
-  @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :upgrade_required}
+  @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :no_sites | :upgrade_required}
   def enable(%Team{} = team) do
-    if eligible?(team) do
-      do_enable(team)
-    else
-      {:error, :upgrade_required}
-    end
+    with :ok <- ensure_eligible(team), do: do_enable(team)
   end
 
   @spec disable(Team.t()) :: :ok
@@ -89,6 +85,12 @@ defmodule Plausible.ConsolidatedView do
   end
 
   # TODO: Only active trials and business subscriptions should be eligible.
-  # This function should call a new underlying feature module.
-  defp eligible?(%Team{}), do: always(true)
+  # This function should also call a new underlying feature module.
+  defp ensure_eligible(%Team{} = team) do
+    cond do
+      always(false) -> {:error, :upgrade_required}
+      Plausible.Teams.owned_sites_count(team) == 0 -> {:error, :no_sites}
+      true -> :ok
+    end
+  end
 end
--- a/lib/plausible/stats/filters/query_parser.ex
+++ b/lib/plausible/stats/filters/query_parser.ex
@@ -77,13 +77,17 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
-  def get_consolidated_site_ids(%Plausible.Site{} = site) do
-    if Plausible.Sites.consolidated?(site) do
-      {:ok, site_ids} = Plausible.ConsolidatedView.site_ids(site.team)
-      site_ids
-    else
-      nil
+  on_ee do
+    def get_consolidated_site_ids(%Plausible.Site{} = site) do
+      if Plausible.Sites.consolidated?(site) do
+        {:ok, site_ids} = Plausible.ConsolidatedView.site_ids(site.team)
+        site_ids
+      else
+        nil
+      end
     end
+  else
+    def get_consolidated_site_ids(_site), do: nil
   end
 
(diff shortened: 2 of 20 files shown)


lib/plausible/stats/sql/where_builder.ex
```diff
@@ -41,46 +41,23 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
-  defp filter_site_time_range(:events, query) do
-    {first_datetime, last_datetime} = utc_boundaries(query)
-
-    dynamic(
-      [e],
-      e.site_id == ^query.site_id and e.timestamp >= ^first_datetime and
-        e.timestamp <= ^last_datetime
-    )
+  defp filter_site_id(query) do
+    case query.consolidated_site_ids do
```
RobertJoonas, line 49: @aerosol we discussed taking `consolidated_site_ids` out of the query object and I still think it's a good idea, but perhaps can be done after the cache changes are in? The reason is that the current implementation of `Plausible.ConsolidatedView.site_ids/1` only takes `team` as argument, which we don't have here at this step of query building.

extra/lib/plausible/consolidated_view.ex
```diff
@@ -35,14 +35,25 @@ defmodule Plausible.ConsolidatedView do
...
-  @spec site_ids(Team.t()) :: [pos_integer()] | {:error, :not_found}
+  @spec site_ids(Team.t()) :: {:ok, [pos_integer()]} | {:error, :not_found}
   def site_ids(%Team{} = team) do
     case get(team) do
       nil -> {:error, :not_found}
       _found -> {:ok, Teams.owned_sites_ids(team)}
     end
   end
 
+  def native_stats_start_at(%Team{} = team) do
```
aerosol: It could use typespecs and a test

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -77,14 +77,19 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
   end
 
-  def get_rollup_site_ids(%Plausible.Site{rollup: true} = site) do
-    Plausible.Cache.Adapter.get(:site_ids, site.team_id, fn ->
-      Plausible.Teams.owned_sites_ids(site.team)
-    end)
+  on_ee do
+    def get_consolidated_site_ids(%Plausible.Site{} = site) do
+      if Plausible.Sites.consolidated?(site) do
+        {:ok, site_ids} = Plausible.ConsolidatedView.site_ids(site.team)
```
aerosol: what if for some reason it doesn't match the :ok tuple? should we crash here or is some fallback more preferable?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -77,14 +77,19 @@ defmodule Plausible.Stats.Filters.QueryParser do
...
+        site_ids
+      else
+        nil
```
aerosol: nil is implicit, you can skip the else clause here

lib/plausible/stats/legacy/legacy_query_builder.ex
```diff
@@ -42,17 +42,19 @@ defmodule Plausible.Stats.Legacy.QueryBuilder do
...
-    site_ids =
-      Plausible.Cache.Adapter.get(:site_ids, site.team_id, fn ->
-        Plausible.Teams.owned_sites_ids(site.team)
-      end)
-
-    struct!(query, rollup_site_ids: site_ids)
+  on_ee do
+    defp put_consolidated_site_ids(query, %Plausible.Site{} = site) do
+      if Plausible.Sites.consolidated?(site) do
+        {:ok, site_ids} = Plausible.ConsolidatedView.site_ids(site.team)
```
aerosol: again, is crashing here OK? ditto: making it a single `ConsolidatedView` function accepting `Site.t()`

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -60,7 +60,9 @@ defmodule PlausibleWeb.StatsController do
     can_see_stats? = not Teams.locked?(site.team) or site_role == :super_admin
     demo = site.domain == "plausible.io"
     dogfood_page_path = if demo, do: "/#{site.domain}", else: "/:dashboard"
-    skip_to_dashboard? = conn.params["skip_to_dashboard"] == "true"
+
+    skip_to_dashboard? =
+      conn.params["skip_to_dashboard"] == "true" or Plausible.Sites.consolidated?(site)
```
aerosol, line 65: nice, WYT about adding a test?

test/support/teams/test.ex
```diff
@@ -23,6 +23,10 @@ defmodule Plausible.Teams.Test do
     Plug.Conn.put_session(conn, :current_team_id, team.identifier)
   end
 
+  def new_consolidated_view(team) do
+    new_site(team: team, domain: team.identifier, consolidated: true)
```
aerosol: cool, WYT about calling ConsolidatedView.enable(team) here? It should help us enforce any provisioning conditions moving forward.

lib/plausible/stats/filters/query_parser.ex
aerosol: that said, maybe we could make it `ConsolidatedView.site_ids(Site.t())` - or `site_ids!` where we'd preload the team (which will be no-op when called from here) and do the matching in a single place?

And have it type-annotated and tested separately?

lib/plausible/stats/sql/where_builder.ex
```diff
@@ -41,46 +41,23 @@ defmodule Plausible.Stats.SQL.WhereBuilder do
...
+      nil -> dynamic([x], x.site_id == ^query.site_id)
+      ids -> dynamic([x], fragment("? in ?", x.site_id, ^ids))
```
aerosol:
```suggestion
      [_ | _] = ids -> dynamic([x], fragment("? in ?", x.site_id, ^ids))
```

This makes it defensive enough against an empty list, WYT?

lib/plausible/stats/filters/query_parser.ex
```diff
@@ -77,14 +77,19 @@ defmodule Plausible.Stats.Filters.QueryParser do
     end
   end
 
-  def get_rollup_site_ids(%Plausible.Site{rollup: true} = site) do
-    Plausible.Cache.Adapter.get(:site_ids, site.team_id, fn ->
```
aerosol, line 81: Feel free to remove it from application supervisor child specs as well: https://github.com/plausible/analytics/blob/query-consolidated/lib/plausible/application.ex#L61-L66

lib/plausible/stats/sql/where_builder.ex
aerosol: Sure, let's have a closer look later on

lib/plausible_web/controllers/stats_controller.ex
RobertJoonas: I added a test in the same commit where this change was made: https://github.com/plausible/analytics/pull/5749/commits/113dab276f2cd9aac4ecdcb0cd7f9a6ca277b28d

aerosol: Oh I'm sorry, my bad

test/plausible/consolidated_view_test.exs
```diff
@@ -84,5 +85,28 @@ defmodule Plausible.ConsolidatedViewTest do
...
+
+        datetimes = [
+          ~N[2024-01-01 12:00:00],
+          ~N[2024-01-01 11:00:00],
+          ~N[2024-02-01 12:00:00]
+        ]
+
+        for dt <- datetimes, do: new_site(team: team, native_stats_start_at: dt)
+
+        assert ConsolidatedView.native_stats_start_at(team) == Enum.min(datetimes)
```
aerosol, line 114: nice!

extra/lib/plausible/consolidated_view.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5749/commits/ad813a511f8f7ce393729e210e7abd4aa47c1302

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: With https://github.com/plausible/analytics/pull/5749/commits/b6027a8beed901e3f4e1d7f934db99010d282a40 we'll start fetching the site_ids from cache. Whether or not we should crash here is a good question though. If the site is consolidated, it should also be in the cache, right?

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5749/commits/d7e4817d18b6585f6df8489ddb6836ab4497c580

lib/plausible/stats/legacy/legacy_query_builder.ex
RobertJoonas: Ditto my response to your other comment. Let's continue the conversation above and resolve this one? Clearly applies for both cases the same way.

test/support/teams/test.ex
RobertJoonas: Did that in https://github.com/plausible/analytics/pull/5749/commits/85fcd37f2792479ca8b3c83839f3d5455619529e. Note that this resulted in some stats query test breaking because of native_stats_start_date. By calling the `new_site` test util, we're populating the site with a native_stats_start_at of `~N[2020-01-01 00:00:00]`. With `ConsolidatedView.enable` it's set to the current moment.

For now, I've simply adjusted the `native_stats_start_at` in both tests, but not the best pattern to continue. Any other ideas?

lib/plausible/stats/sql/where_builder.ex
RobertJoonas: Makes sense! https://github.com/plausible/analytics/pull/5749/commits/3c127a93dd352ecb7ccde50130f038d01b4d6de1

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Thanks! https://github.com/plausible/analytics/pull/5749/commits/71b8b02fd6656a340a9cbda7b25cde24feb3e5b3

test/support/teams/test.ex
aerosol: Great. The reason I thought this change will be useful is it should uncover any flaws in thinking thus far. Which it did. So now we can consider making enable/1 set the `native_stats_start_at` at the moment of creation. Why? Regular sites have to exists (at least two -- we'll probably want to make this a well defined constraint later on), and this consolidated one is not awaiting any stats. So compute this date right away. We'll also have to think about refreshing it, when the regular site set changes. 
So WYT about not slapping dates on tests and instead make the computation part of `enable/1` in this PR?

lib/plausible/stats/filters/query_parser.ex
aerosol: You raise a good point, I'll make sure we don't cache consolidated sites when there's not enough regular sites. Though, we need symmetry in how we enforce this -- regular site set changes should be reflected whether the consolidated view is enabled or not. Let's take site deletion for example -- we'll have to make sure an orphaned consolidated view has no right of staying afloat.

test/support/teams/test.ex
RobertJoonas: Makes sense!

> So WYT about not slapping dates on tests and instead make the computation part of enable/1 in this PR?

Sounds good! I'll do that.

lib/plausible/stats/filters/query_parser.ex
RobertJoonas: Yep. Some thinking material for the next PRs I suppose. In the scope of this one, should be OK to assume that a consolidated site always has its cache entry.

test/support/teams/test.ex
RobertJoonas: Done now with https://github.com/plausible/analytics/pull/5749/commits/2f9fd11ae7b6f616e798ba355623d8fc53171d1e & https://github.com/plausible/analytics/pull/5749/commits/14dcbdd7653e19c082adff47a23e90296e96af6d

PR #5753, opened 2025-09-24
Check for Sites API feature against respective team when using API key

zoldar:
### Changes

This PR refactors plug authorizing public Stats and Sites API requests with a focus on error handling revamp.

### Tests

### DIFF
--- a/test/plausible_web/controllers/api/external_sites_controller_sites_crud_api_test.exs
+++ b/test/plausible_web/controllers/api/external_sites_controller_sites_crud_api_test.exs
@@ -21,6 +21,17 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerSitesCrudApiTest do
     end
 
     describe "POST /api/v1/sites" do
+      setup %{user: user} do
+        subscribe_to_enterprise_plan(user,
+          features: [
+            Plausible.Billing.Feature.StatsAPI,
+            Plausible.Billing.Feature.SitesAPI
+          ]
+        )
+
+        :ok
+      end
+
       test "can create a site", %{conn: conn} do
         conn =
           post(conn, "/api/v1/sites", %{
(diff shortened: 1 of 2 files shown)


extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -483,6 +483,12 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
       {:missing, param} ->
         H.bad_request(conn, "Parameter `#{param}` is required to create a custom property")
 
+      {:error, :upgrade_required} ->
```
zoldar, line 486: Missing support for this case surfaced after the refactor 😅

lib/plausible_web/plugs/authorize_public_api.ex
```diff
@@ -122,13 +122,30 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPI do
...
       site = Plausible.Repo.preload(site, :completed_imports)
       {:ok, assign(conn, :site, site)}
     end
   end
 
+  defp verify_by_scope(conn, api_key, "sites:" <> scope_suffix = scope) do
+    feature =
+      case scope_suffix do
+        "read:" <> _ ->
+          Plausible.Billing.Feature.StatsAPI
```
zoldar, line 136: Read-only endpoints of Sites API are available to teams which have Stats API feature enabled.

lib/plausible_web/plugs/authorize_public_api.ex
```diff
@@ -260,17 +322,32 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPI do
...
   end
 
-  defp send_error(conn, _, {:error, :site_locked}) do
+  defp feature_payment_error(conn, feature) do
+    feature_name = feature.display_name()
+
     H.payment_required(
       conn,
-      "This Plausible site is locked due to missing active subscription. In order to access it, the site owner should subscribe to a suitable plan"
+      "The account that owns this API key does not have access to #{feature_name}. Please make sure you're using the API key of a subscriber account and that the subscription plan includes #{feature_name}"
```
zoldar, line 350: The messaging may not reflect the reality where we check the feature availability on a team. It holds for team-scoped keys but not for legacy keys where the key is not associated with the team but only the user. At the same time, "account" sounds a bit awkward here. We could change it to something like "The team/site does not have access to ..." but then it's going to be completely off for legacy keys. I'm leaving it as it was for now, unless somebody has a better idea.

test/plausible_web/controllers/api/external_sites_controller_sites_crud_api_test.exs
```diff
@@ -331,21 +352,6 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerSitesCrudApiTest do
                        }) = response
       end
 
-      test "does not allow creating more sites than the limit", %{conn: conn, user: user} do
```
zoldar, line 334: This test wasn't for a realistic scenario:

- only teams on enterprise plans can have Sites API enabled
- enterprise plans' site limits are not enforced - a usage alert notification is sent instead

, hence it was removed

ukutaht (approved):

PR #5756, opened 2025-09-25
Tailwind v4

ukutaht:
Upgrades TailwindCSS from v3 to v4.

This is in general required to upgrade other dependencies:
phoenix 18 & phoenix liveview 1.1
-> requires upgrading storybook to 0.9.0
-> requires upgrading tailwind to v4

With this PR the storybook will be left in a pretty broken state because we're still on 0.8 which expects tailwind 3. In order to upgrade storybook we also have to upgrade phoenix and liveview. It's a bit of a dependency hell but I think tailwind 4 was the worst of it. Now we can do a follow-up with Phoenix 18, Liveview 1.1 and Storybook 0.9. Should work in theory.

I was able to mostly replicate the existing style we have. But tailwind 3->4 has many many breaking changes so I'm sure I missed something. I'll put this on a preview env for more manual exploratory testing and comparisons with prod. Even so I'm sure some styles will be missed but IMO we should take the plunge and march forward with this.


ukutaht:
@sanne-san @RobertJoonas All reported issues fixed, plus one conflict I found between `dark` and `hover` modifiers in site/account settings sidebar links.

PR #5758, opened 2025-09-26
Move InstallationSupport code to `/extra` (EE only) folder

RobertJoonas:
### Changes

Moves all InstallationSupport submodules into `/extra` as these will not be needed on CE going forward from the Script v2 release. Reviewing commit by commit highly recommended due to a bunch of renamings.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -21,10 +21,6 @@ defmodule Plausible.Ingestion.Event do
             salts: nil,
             changeset: nil
 
-  on_ee do
-    @verification_user_agent Plausible.InstallationSupport.user_agent()
-  end
-
   @type drop_reason() ::
           :bot
           | :spam_referrer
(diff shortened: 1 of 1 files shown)


lib/plausible/ingestion/event.ex
```diff
@@ -21,7 +21,9 @@ defmodule Plausible.Ingestion.Event do
             salts: nil,
             changeset: nil
 
-  @verification_user_agent Plausible.InstallationSupport.user_agent()
+  on_ee do
```
ukutaht: Suggestion: move the module attribute just above the the function where it's used and nest them both under a single `on_ee` block.

ukutaht (approved):

lib/plausible/ingestion/event.ex
RobertJoonas: Good call! https://github.com/plausible/analytics/pull/5758/commits/01ad962107e7b4dec2c30126af6adde3b775faeb

PR #5760, opened 2025-09-29
Consolidated lifecycle

aerosol:
### Changes

With this change, regular site removal disables consolidated view, if there's no sites to consolidate any more.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -25,10 +25,20 @@ defmodule Plausible.ConsolidatedView do
     with :ok <- ensure_eligible(team), do: do_enable(team)
   end
 
+  @spec enabled?(Team.t()) :: boolean
+  def enabled?(%Team{} = team) do
+    not is_nil(get(team))
+  end
+
   @spec disable(Team.t()) :: :ok
   def disable(%Team{} = team) do
     Plausible.Repo.delete_all(from(s in sites(), where: s.domain == ^make_id(team)))
     :ok
+  end
+
+  @spec has_sites_to_consolidate?(Team.t()) :: boolean()
+  def has_sites_to_consolidate?(%Team{} = team) do
+    Teams.owned_sites_count(team) > 0
   end
 
   @spec site_ids(Team.t() | String.t()) :: {:ok, [pos_integer()]} | {:error, :not_found}
@@ -90,10 +100,6 @@ defmodule Plausible.ConsolidatedView do
   # TODO: Only active trials and business subscriptions should be eligible.
   # This function should also call a new underlying feature module.
   defp ensure_eligible(%Team{} = team) do
-    if Plausible.Teams.owned_sites_count(team) == 0 do
-      {:error, :no_sites}
-    else
-      :ok
-    end
+    if has_sites_to_consolidate?(team), do: :ok, else: {:error, :no_sites}
   end
 end
--- a/test/plausible/consolidated_view_test.exs
+++ b/test/plausible/consolidated_view_test.exs
@@ -7,13 +7,13 @@ defmodule Plausible.ConsolidatedViewTest do
     alias Plausible.ConsolidatedView
     import Plausible.Teams.Test
 
-    describe "enable/1" do
+    describe "enable/1 and enabled?/1" do
       setup [:create_user, :create_team]
 
       test "creates and persists a new consolidated site instance", %{team: team} do
         new_site(team: team)
         assert {:ok, %Plausible.Site{consolidated: true}} = ConsolidatedView.enable(team)
-        assert ConsolidatedView.get(team)
+        assert ConsolidatedView.enabled?(team)
       end
 
       test "is idempotent", %{team: team} do
@@ -30,6 +30,7 @@ defmodule Plausible.ConsolidatedViewTest do
 
       test "returns {:error, :no_sites} when the team does not have any sites", %{team: team} do
         assert {:error, :no_sites} = ConsolidatedView.enable(team)
+        refute ConsolidatedView.enabled?(team)
       end
 
       @tag :skip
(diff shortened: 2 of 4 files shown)


RobertJoonas (approved):
Looks good to me! We should probably do the same on site ownership transfers.

aerosol:
@RobertJoonas please re-review, updated with changes we've discussed

extra/lib/plausible/consolidated_view.ex
```diff
@@ -15,6 +15,21 @@ defmodule Plausible.ConsolidatedView do
 
   import Ecto.Query
 
+  @spec reset_if_enabled(Team.t()) :: :ok
+  def reset_if_enabled(%Team{} = team) do
+    {:ok, :ok} =
+      Repo.transaction(fn ->
+        if enabled?(team) do
+          disable(team)
+          enable(team)
```
RobertJoonas: It looks a bit strange to delete the consolidated site and then add it back instead of simply updating it. Also, thinking a few steps ahead, a deletion doesn't sound great with goals involved (since they'd have to be deleted as well).

Couldn't we just do something like this?

```elixir
get(team)
|> change_stats_dates()
|> maybe_bump_updated_at()
|> Repo.update()
```

where `maybe_bump_updated_at` would make sure we're bumping the `updated_at` field even if there's no change in the stats dates. That's required for the cache to refresh recently updated on site deletion, I think?

test/plausible/consolidated_view_test.exs
```diff
@@ -92,17 +93,10 @@ defmodule Plausible.ConsolidatedViewTest do
       end
     end
 
-    describe "native_stats_start_at/1" do
+    describe "stats start" do
```
RobertJoonas: Nitpick: Maybe the one test in this describe block could be moved under `enable/1` instead? E.g.: "creates a consolidated view with stats_start_date and native_stats_start_at of the oldest included site"

extra/lib/plausible/consolidated_view.ex
```diff
@@ -25,12 +40,22 @@ defmodule Plausible.ConsolidatedView do
...
+  end
+
   @spec disable(Team.t()) :: :ok
   def disable(%Team{} = team) do
     Plausible.Repo.delete_all(from(s in sites(), where: s.domain == ^make_id(team)))
     :ok
   end
 
+  @spec has_sites_to_consolidate?(Team.t()) :: boolean()
+  def has_sites_to_consolidate?(%Team{} = team) do
```
RobertJoonas: Could be a private function, I think?

lib/plausible/sites.ex
```diff
@@ -348,27 +348,23 @@ defmodule Plausible.Sites do
...
-
-        start_date
-      end
+      site.stats_start_date
     end
   end
 
+  def stats_start_date(%Site{stats_start_date: %Date{} = date}) do
+    date
+  end
```
RobertJoonas, line 365: Looks like when moving this function clause back above the `def stats_start_date(%Site{consolidated: true} = site)` one, then the tests still pass. Let's add a test that would catch this.

Perhaps a direct unit test for the `stats_start_date` function in `sites_test.exs`? E.g. "resets consolidated sites' stats dates even when stats_start_date exists already"

extra/lib/plausible/consolidated_view.ex
aerosol: oof you're right, great point with goals

aerosol: apart from the code you wrote, we'll still have to handle situation in which it's not eligible to stay, that is, if there's not enough sites to consolidate - in which case, a full delete will be made.

aerosol: WYT? 0daebd174f

extra/lib/plausible/consolidated_view.ex
aerosol: 80c09d1d35

test/plausible/consolidated_view_test.exs
aerosol: 23721ff8dd

lib/plausible/sites.ex
aerosol: 95e639ea34

extra/lib/plausible/consolidated_view.ex
aerosol: And https://github.com/plausible/analytics/pull/5760/commits/84844f65524e5f53ee21cf067272f11bd7b7889a so that credo doesn't cry

extra/lib/plausible/consolidated_view.ex
```diff
@@ -56,26 +82,32 @@ defmodule Plausible.ConsolidatedView do
...
+
+      site_or_changeset
+      |> Site.set_native_stats_start_at(native_stats_start_at)
+      |> Site.set_stats_start_date(start_date)
+    else
+      site_or_changeset
+    end
+  end
+
+  defp bump_updateed_at(struct_or_changeset) do
```
RobertJoonas: Spotted a double "e" in the function name

extra/lib/plausible/consolidated_view.ex
```diff
@@ -15,6 +15,26 @@ defmodule Plausible.ConsolidatedView do
 
   import Ecto.Query
 
+  @spec reset_if_enabled(Team.t()) :: :ok
+  def reset_if_enabled(%Team{} = team) do
+    case get(team) do
+      nil ->
+        :skip
```
RobertJoonas, line 22: JIC: Was your intent to return this from the function? Currently we're always returning `:ok`, even if no consolidated_view is found.

aerosol: yep, "_if_enabled" - we want it to be no-op in some cases as indicated by the test included

extra/lib/plausible/consolidated_view.ex
aerosol: thx 4a01ae391b

cnkk:
@cubic-dev-ai review this pull request

PR #5766, opened 2025-09-30
Upgrade phoenix & liveview

ukutaht:
### Changes

Upgrades to Phoenix 1.18 and Liveview 1.1.

Some changes are due to new `mix format` rules coming from updated deps.

`bcyrpt` was also updated due to `elixir_make` dependency for NIFs. Needed in order to bring in the required `lazy_html` for the new liveview version.

`phoenix_storybook` updated to 0.9.0. Even though we want some sort of styleguide, I'm considering removing this dependency in the future and hand-rolling a replacement or something, IDK. It's not easy to work with it.

TODO:
- [x] Remove Floki completely in favor of LazyHTML
- [x] Review whether `:browser_sso_notice` pipeline requires put_root_layout

Future possibilities:
* Set up colocated hooks

### DIFF
--- a/extra/lib/plausible/installation_support/checks/detection.ex
+++ b/extra/lib/plausible/installation_support/checks/detection.ex
@@ -54,7 +54,7 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
   """
 
   # We define a timeout for the browserless endpoint call to avoid waiting too long for a response
-  @endpoint_timeout_ms 2_000
+  @endpoint_timeout_ms 3_000
 
   # This timeout determines how long we wait for window.plausible to be initialized on the page, used for detecting whether v1 installed
   @plausible_window_check_timeout_ms 1_500
@@ -121,9 +121,11 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
     end
   end
 
-  defp handle_browserless_response(state, _body, status) do
+  defp handle_browserless_response(state, body, status) do
     error = "Unhandled browserless response with status: #{status}"
-    Logger.warning(warning_message(error, state))
+
+    warning_message("#{error}; body: #{inspect(body)}", state)
+    |> Logger.warning()
 
     put_diagnostics(state, service_error: error)
   end
--- a/extra/lib/plausible/installation_support/checks/installation_v2.ex
+++ b/extra/lib/plausible/installation_support/checks/installation_v2.ex
@@ -146,9 +146,11 @@ defmodule Plausible.InstallationSupport.Checks.InstallationV2 do
     end
   end
 
-  defp handle_browserless_response(state, _body, status) do
+  defp handle_browserless_response(state, body, status) do
     error = "Unhandled browserless response with status: #{status}"
-    Logger.warning(warning_message(error, state))
+
+    warning_message("#{error}; body: #{inspect(body)}", state)
+    |> Logger.warning()
 
     put_diagnostics(state, service_error: error)
   end
--- a/extra/lib/plausible/stats/sampling.ex
+++ b/extra/lib/plausible/stats/sampling.ex
@@ -97,7 +97,9 @@ defmodule Plausible.Stats.Sampling do
     estimate_by_filters(duration_adjusted_traffic, query.filters)
   end
 
-  @filter_traffic_multiplier 1 / 20.0
+  @filter_traffic_multiplier 1 / 4.0
+  @max_filters 2
+
   defp estimate_by_filters(estimation, filters),
-    do: estimation * @filter_traffic_multiplier ** length(filters)
+    do: estimation * @filter_traffic_multiplier ** min(length(filters), @max_filters)
 end
--- a/lib/plausible/ingestion/event.ex
+++ b/lib/plausible/ingestion/event.ex
@@ -27,6 +27,7 @@ defmodule Plausible.Ingestion.Event do
           | GateKeeper.policy()
           | :invalid
           | :dc_ip
+          | :threat_ip
           | :site_ip_blocklist
           | :site_country_blocklist
           | :site_page_blocklist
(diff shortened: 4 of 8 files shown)


extra/lib/plausible/auth/sso/domain/verification.ex
```diff
@@ -60,8 +60,10 @@ defmodule Plausible.Auth.SSO.Domain.Verification do
     with %Req.Response{body: body} = response when is_binary(body) <-
            run_request(url_override || "https://#{sso_domain}"),
          true <- html?(response),
-         {:ok, html} <- Floki.parse_document(body),
-         [_ | _] <- Floki.find(html, ~s|meta[name="#{@prefix}"][content="#{domain_identifier}"]|) do
+         html <- LazyHTML.from_document(body),
```
ukutaht, line 63: This passes tests. It's the only production code where I've migrated from Floki -> LazyHTML in this PR.

ukutaht:
I've completely refactored our `Floki` usage over to the new `LazyHTML`. However, the one place I didn't do it is in the installation/verification checks code.

@RobertJoonas @apata once we're on scriptV2, can we remove the old checks that use Floki? In that case I would wait for next week to completely remove Floki. If we still need HTML parsing in Elixir, I will refactor over to LazyHTML in this PR. Do we need these backend checks going forward?

RobertJoonas:
> once we're on scriptV2, can we remove the old checks that use Floki? In that case I would wait for next week to completely remove Floki. If we still need HTML parsing in Elixir, I will refactor over to LazyHTML in this PR. Do we need these backend checks going forward?

@ukutaht yes, all the `LegacyVerification` code will be removed. That includes:

* Plausible.InstallationSupport.LegacyVerification.Checks
* Plausible.InstallationSupport.LegacyVerification.Diagnostics
* Plausible.InstallationSupport.LegacyVerification.Errors
* Plausible.InstallationSupport.Checks.CSP
* Plausible.InstallationSupport.Checks.FetchBody
* Plausible.InstallationSupport.Checks.Installation
* Plausible.InstallationSupport.Checks.ScanBody
* Plausible.InstallationSupport.Checks.SnippetCacheBust
* Plausible.InstallationSupport.Checks.Snippet

ukutaht:
OK good, then I won't put any effort into refactoring those

test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -1180,7 +1179,7 @@ defmodule PlausibleWeb.AuthControllerTest do
 
       assert element_exists?(
                html,
-               ~s|a[data-method="post"][data-to="#{Routes.auth_path(conn, :initiate_2fa_setup)}"|
```
aerosol, line 1183: hah, how did that work before

test/plausible_web/live/team_setup_test.exs
```diff
@@ -263,8 +263,8 @@ defmodule PlausibleWeb.Live.TeamSetupTest do
 
       html = lv |> render()
 
-      assert [_ | _] = find(html, "#{member_el()}:nth-of-type(1) a")
-      assert find(html, "#{member_el()}:nth-of-type(2) a") == []
+      assert Enum.count(find(html, "#{member_el()}:nth-of-type(1) a")) >= 2
```
aerosol: `[_ | _]` is non-empty, so maybe > 1 to preserve the original intention

test/plausible_web/live/team_management_test.exs
```diff
@@ -166,14 +166,14 @@ defmodule PlausibleWeb.Live.TeamMangementTest do
...
-      assert length(find(html, member_el())) == 1
+      assert Enum.count(find(html, member_el())) == 1
 
       assert text_of_element(html, "#{guest_el()}:first-of-type button") == "Guest"
 
       change_role(lv, 1, "viewer", guest_el())
       html = render(lv)
 
-      assert length(find(html, member_el())) == 2
+      assert Enum.count(find(html, member_el())) == 2
```
aerosol: depending on how you feel about it, this seems to be so common, we could consider a soup helper for it:

```elixir
assert elem_count(html, member_el()) == 2
```

```elixir
def elem_count(html, selector) do
  html |> find(selector) |> Enum.count()
end
```

test/plausible_web/live/register_form_test.exs
```diff
@@ -272,7 +272,7 @@ defmodule PlausibleWeb.Live.RegisterFormTest do
                _csrf_input,
                _action_input,
                email_input | _
-             ] = find(html, "input")
+             ] = find(html, "input") |> Enum.into([])
```
aerosol, line 275: Sorry, what's the reason for not including `Enum.into([])` in html.ex? Feels like the changeset would be smaller then?

zoldar: AFAICT it's probably because LazyHTML.query returns a struct for multiple entities too, which is necessary if the output is going to be processed by LazyHTML any further - which it is, internally in the HTML helper module at least. Potentially the one exposing raw query output could be made private and the public could always expose a list instead (BTW, `List.wrap()` works here too.

zoldar (approved):

test/plausible_web/controllers/auth_controller_test.exs
ukutaht: Floki seems to be very lenient with selectors 🤷‍♂️

test/plausible_web/live/team_setup_test.exs
ukutaht: Ah good point.

If it's just asserting that it's non-empty then it seems logically the same as `element_exists?`. I've refactored it as such, let me know if I missed something.

test/plausible_web/live/team_management_test.exs
ukutaht: I like it! https://github.com/plausible/analytics/pull/5766/commits/44aa3c0e5f78ed121a018583996b8e3161a663fd

PR #5772, opened 2025-10-01
Fix parsing of percentage setting for persistor from env in runtime config

zoldar:
Welp, turns out it was technically set to 100% 😂


aerosol:
image

cnkk (approved):

PR #5775, opened 2025-10-06
Fix wrapping in .tooltip

apata:
### Changes

Tries to fix issue from Tailwind 4 upgrade that makes tooltips overflow. Apparently in the new version "whitespace-nowrap" is stronger than it used to be.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -543,7 +543,8 @@ defmodule PlausibleWeb.Components.Generic do
       "-translate-y-full",
       "z-[1000]",
       "sm:max-w-72",
-      "whitespace-nowrap"
+      "whitespace-normal",
+      "break-words"
     ]
 
     tooltip_position_classes =


apata:
Damn, looks like we need to fix the width as well, at the moment it renders too thin
Screenshot 2025-10-06 at 13 27 01

apata:
Fixed now 
Screenshot 2025-10-06 at 15 13 59

ukutaht:
On screens smaller than `sm` the tooltips do not display at all. Is this intended?

aerosol:
We need to get rid of those long term, it's really hard to make sure no other use breaks with any change

apata:
> On screens smaller than `sm` the tooltips do not display at all. Is this intended?

@ukutaht I think the idea is that smaller screens are likely to be touchscreens, where hover state is dubious. Since it's a generic component, I'd rather not experiment in this quick fix PR with anything more advanced. 

> We need to get rid of those long term, it's really hard to make sure no other use breaks with any change

@aerosol Agreed that we need to get rid of these. They are bad for accessibility and for mobile screens. Regarding breaking things, the shortest tooltip that I found was `[ Learn more ]`. It would now start to render as `[ Learn more       ]`, but I think that's good enough for a quick fix.

PR #5777, opened 2025-10-06
Swap out GTM template link in installation instructions

ukutaht:


apata:
@ukutaht There's an alternative proposed https://github.com/plausible/analytics/pull/5779

PR #5778, opened 2025-10-06
Increase detection call timeout

apata:
### Changes

Increases detection call timeout. The idea is to prefer to get a timeout from the inner function that's run on the site, rather than destroy the whole function context. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/installation_support/checks/detection.ex
+++ b/extra/lib/plausible/installation_support/checks/detection.ex
@@ -54,10 +54,10 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
   """
 
   # We define a timeout for the browserless endpoint call to avoid waiting too long for a response
-  @endpoint_timeout_ms 2_000
+  @endpoint_timeout_ms 3_000
 
   # This timeout determines how long we wait for window.plausible to be initialized on the page, used for detecting whether v1 installed
-  @plausible_window_check_timeout_ms 1_500
+  @plausible_window_check_timeout_ms 2_000
 
   # To support browserless API being unavailable or overloaded, we retry the endpoint call if it doesn't return a successful response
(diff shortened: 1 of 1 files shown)


extra/lib/plausible/installation_support/checks/detection.ex
```diff
@@ -54,10 +54,10 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
   """
 
   # We define a timeout for the browserless endpoint call to avoid waiting too long for a response
-  @endpoint_timeout_ms 2_000
+  @endpoint_timeout_ms 3_000
 
   # This timeout determines how long we wait for window.plausible to be initialized on the page, used for detecting whether v1 installed
-  @plausible_window_check_timeout_ms 1_500
+  @plausible_window_check_timeout_ms 2_000
```
RobertJoonas: I think we should keep this at 1500ms. The way I understand it, the intent is to increase the difference between these two timeout vars. `plausible_window_check_timeout_ms` determines how long our script works on the page, but the `endpoint_timeout_ms` also includes the time to get a successful response from the website in question, which might easily be over 500ms (with redirects, etc).

That makes me wonder if we should perhaps even increase the `endpoint_timeout_ms` to something like 5s? If reaching the website is super slow, this check would be slow too. Otherwise, it would finish in a reasonable time.

RobertJoonas: Let's try with 3000ms for now and see how it goes.

apata: @RobertJoonas Thanks! I think we can try to sneak up on the most acceptable time bit by bit. I'm resetting the window check to 1.5s: was on the fence about it as well.

PR #5779, opened 2025-10-06
Tweaks to the GTM onboarding copy 

metmarkosaric:


apata:
@metmarkosaric The template says "Script ID", so maybe we can add a space app-side as well?

metmarkosaric:
@apata yes, good point!

lib/plausible_web/live/installationv2/instructions.ex
```diff
@@ -147,13 +147,13 @@ defmodule PlausibleWeb.Live.InstallationV2.Instructions do
...
-            <.styled_link href="https://plausible.io/docs/google-tag-manager" new_tab={true}>
-              Read our Tag Manager guide
+            <.styled_link href="https://plausible.io/gtm-template" new_tab={true}>
+              Install the Plausible template in GTM
             </.styled_link>
           </:item>
 
           <:item>
-            Once done, click the button below to verify your installation.
+            Paste your Script ID into the template and click the button below to verify your installation.
```
apata, line 156: This is better than before definitely, but I think maybe it can be improved further?

The message that I'd like to get across (not necessarily in these words, just structurally and link-wise).

If you're familiar with how GTM templates work, just go ahead and add our [Plausible template](https://tagmanager.google.com/gallery/#/owners/plausible/templates/plausible-gtm-template) to your workspace and copy this ID ([ID]) into the input form on the template. Apply the template GTM-side and come back here to click Verify.

If you don't know about GTM templates, [here's a doc to help.](https://plausible.io/gtm-template)

Verify

metmarkosaric: hmm I see. i've tried to make it look and feel more like the WP page which is very short and works all fine so far. we can add more copy or change the design of this page later if people struggle with the current instructions. next up we'll improve the GTM guide

PR #5784, opened 2025-10-07
Fine tune sampling logic: filter estimate multiplier and cap filter exponent

zoldar:
### Changes

This change reduces the risk of dropping sampling when querying very large quantities of data due to too aggressive relaxation of sampling threshold when applying filters. The cap on the exponent ensures there's a clear upper boundary. The multiplier got adjusted as well so that sampling is not reduced as sharply with a second (and further) filter.

### Tests

### DIFF
--- a/extra/lib/plausible/stats/sampling.ex
+++ b/extra/lib/plausible/stats/sampling.ex
@@ -97,7 +97,9 @@ defmodule Plausible.Stats.Sampling do
     estimate_by_filters(duration_adjusted_traffic, query.filters)
   end
 
-  @filter_traffic_multiplier 1 / 20.0
+  @filter_traffic_multiplier 1 / 4.0
+  @max_filters 2
+
   defp estimate_by_filters(estimation, filters),
-    do: estimation * @filter_traffic_multiplier ** length(filters)
+    do: estimation * @filter_traffic_multiplier ** min(length(filters), @max_filters)
 end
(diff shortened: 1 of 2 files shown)


aerosol (approved):

extra/lib/plausible/stats/sampling.ex
```diff
@@ -97,7 +97,9 @@ defmodule Plausible.Stats.Sampling do
     estimate_by_filters(duration_adjusted_traffic, query.filters)
   end
 
-  @filter_traffic_multiplier 1 / 20.0
+  @filter_traffic_multiplier 1 / 4.0
+  @max_filters 2
+
   defp estimate_by_filters(estimation, filters),
-    do: estimation * @filter_traffic_multiplier ** length(filters)
+    do: estimation * @filter_traffic_multiplier ** min(length(filters), @max_filters)
```
aerosol, line 104: do we still want to use power instead of multiplication here?

zoldar: There's a cap so, that's still a max of 16 * 10_000_000 = 160M which should still be acceptable. Exponential is more fitting to how filters in general tend to narrow the dataset down. We could make the slope "softer" by allowing more filters but then, well, it's very rare filtering goes over a 4-5 filters, so 16 "steps" would be too much and too flat (the relaxation would be too small to avoid issues with sampling when using very specific filters). Exponential makes more sense for that purpose IMHO.

PR #5790, opened 2025-10-09
Installation support telemetry

RobertJoonas:
### Changes

* Emit telemetry events for unhandled / handled `Detection.Checks` runs
* Emit telemetry events for unhandled / handled `Verification.Checks` runs
* In `Verification.Checks`, skip `InstallationV2CacheBust` when `InstallationV2` fails with the default/unhandled case
    * it makes telemetry more accurate - a flaky failure could otherwise go unnoticed, being overwritten by "we encountered an issue with your site's cache"
    * it improves the UX. From what we've seen so far, the majority of unhandled cases are caused by Browserless timeouts, random 400s, and flaky behaviour (either uncovered bugs in the verification script or flaky behaviour on customers' sites). Retrying with cache bust and getting a success will only add confusion since the issue is most likely not the site cache.
* Bug fix: Silence Sentry errors in the intermediate interpretation step that is used to determine whether the cache bust check should or should not run.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/installation_support/detection/checks.ex
+++ b/extra/lib/plausible/installation_support/detection/checks.ex
@@ -37,7 +37,8 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
     )
   end
 
-  def telemetry_event(name), do: [:plausible, :detection, name]
+  def telemetry_event_handled(), do: [:plausible, :detection, :handled]
+  def telemetry_event_unhandled(), do: [:plausible, :detection, :unhandled]
 
   def interpret_diagnostics(%State{} = state) do
     result = Detection.Diagnostics.interpret(state.diagnostics, state.url)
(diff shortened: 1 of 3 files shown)


lib/plausible/telemetry/plausible_metrics.ex
```diff
@@ -177,6 +181,38 @@ defmodule Plausible.PromEx.Plugins.PlausibleMetrics do
...
             ),
           else: nil
+        ),
+        on_ee(
+          do:
+            counter(
+              metric_prefix ++ [:detection, :handled],
+              event_name: InstallationSupport.Detection.Checks.telemetry_event(:handled)
+            ),
+          else: nil
```
aerosol: nil is implicit

extra/lib/plausible/installation_support/detection/checks.ex
```diff
@@ -37,11 +37,28 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
     )
   end
 
+  def telemetry_event(name), do: [:plausible, :detection, name]
```
aerosol: explicit function names are better - typo prone, and you won't have to find out that the new metric is not declared elsewhere. So `telemetry_event_handled/0` and `telemetry_event_unhandled/0` are just safer to use at compile-time

RobertJoonas: Makes sense! Changed in https://github.com/plausible/analytics/pull/5790/commits/ed4f301e459af3e628ac7ff78040e9d39a56bf15. Thanks for the review @aerosol!

PR #5792, opened 2025-10-09
Log unhandled verification and detection cases

RobertJoonas:
### Changes

* Stop passing arguments around (comment inline)
* Log the `data_domain` and `diagnostics` whenever there's an unhandled verification or detection case

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/application.ex
+++ b/lib/plausible/application.ex
@@ -196,6 +196,7 @@ defmodule Plausible.Application do
     setup_request_logging()
     setup_sentry()
     setup_opentelemetry()
+    Plausible.Ingestion.Persistor.TelemetryHandler.install()
 
     setup_geolocation()
     Location.load_all()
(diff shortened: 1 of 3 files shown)


extra/lib/plausible/installation_support/detection/checks.ex
```diff
@@ -40,11 +40,15 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
...
-    result = Detection.Diagnostics.interpret(state.diagnostics, state.url)
+  def interpret_diagnostics(%State{
+        diagnostics: diagnostics,
+        data_domain: data_domain,
+        url: url
+      }) do
+    result = Detection.Diagnostics.interpret(diagnostics, url)
 
     case result.data do
-      %{unhandled: true, diagnostics: diagnostics, url: url} ->
```
RobertJoonas, line 47: introduced this code in https://github.com/plausible/analytics/pull/5790, but this is unnecessary -- we can just use the same values from the `state` directly. The diagnostics, url, nor data_domain do not change when they're getting interpretted.

PR #5795, opened 2025-10-10
Consolidated view settings

RobertJoonas:
### Changes

image

* Filter out the irrelevant site settings from consolidated view settings. It now looks like the screenshot above, with the following sections removed:
  * General > Domain Change
  * General > Review Installation
  * People (the whole section) - everyone in the team can access the consolidated view. No guests allowed.
  * Visibility (the whole section) - we decided to go with a very minimal configuration. See [this basecamp comment](https://3.basecamp.com/5308029/buckets/43891605/card_tables/cards/9131317744#__recording_9131372126) for the related discussion.
  * Funnels - removed as per Marko's comment [here](https://3.basecamp.com/5308029/buckets/43891605/messages/9052698223#__recording_9060538821)
  * Integrations (the whole section) - Google Search Console can only be connected to a single domain. Looker studio is currently considered out of scope.
  * Imports/Exports - cannot import stats because there's no ingestion to a consolidated view, and the actual data belongs to children sites. The same should apply for imported data. For the same reason, a full CSV export is not available - it's only meant to be imported into another Plausible instance.
  * Shields (the whole section) removed. No ingestion means no need for shields
  * Danger Zone (the whole section) - Ownership cannot be transferred. There are no stats to reset. Deleting the site will not be possible -- only hiding it.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/goal_settings/form.ex
+++ b/lib/plausible_web/live/goal_settings/form.ex
@@ -112,7 +112,7 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
       phx-target={@myself}
     >
       <.title>
-        Add goal for {if(Plausible.Sites.consolidated?(@site), do: "consolidated view", else: @domain)}
+        Add goal for {Plausible.Sites.display_name(@site)}
       </.title>
 
       <.tabs current_user={@current_user} site={@site} selected_tab={@selected_tab} myself={@myself} />
--- a/lib/plausible_web/live/props_settings/form.ex
+++ b/lib/plausible_web/live/props_settings/form.ex
@@ -58,10 +58,7 @@ defmodule PlausibleWeb.Live.PropsSettings.Form do
           phx-click-away="cancel-allow-prop"
         >
           <.title>
-            Add property for {if(Plausible.Sites.consolidated?(@site),
-              do: "consolidated view",
-              else: @domain
-            )}
+            Add property for {Plausible.Sites.display_name(@site)}
           </.title>
 
           <div class="mt-6">
--- a/lib/plausible_web/templates/layout/site_settings.html.heex
+++ b/lib/plausible_web/templates/layout/site_settings.html.heex
@@ -8,10 +8,7 @@
   </.styled_link>
   <div class="pb-5 border-b border-gray-200 dark:border-gray-750">
     <h2 class="text-2xl font-bold leading-7 text-gray-900 dark:text-gray-100 sm:text-3xl sm:leading-9 sm:truncate">
-      Settings for {if(Plausible.Sites.consolidated?(@site),
-        do: "the consolidated view",
-        else: @site.domain
-      )}
+      Settings for {Plausible.Sites.display_name(@site)}
     </h2>
   </div>
   <div class="lg:grid lg:grid-cols-12 lg:gap-x-5 lg:mt-4">
--- a/test/plausible_web/controllers/site_controller_test.exs
+++ b/test/plausible_web/controllers/site_controller_test.exs
@@ -593,7 +593,7 @@ defmodule PlausibleWeb.SiteControllerTest do
           |> find("[data-testid=site_settings_sidebar] a")
           |> Enum.map(fn a -> {text(a), text_of_attr(a, "href")} end)
 
-        assert resp =~ "Settings for the consolidated view"
+        assert resp =~ "Settings for consolidated view"
 
         assert items == [
                  {"General", "/#{site.domain}/settings/general"},
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -26,6 +26,10 @@ defmodule Plausible.Sites do
     def consolidated?(%Site{}), do: always(false)
   end
 
+  def display_name(%Site{} = site) do
+    if consolidated?(site), do: "consolidated view", else: site.domain
+  end
+
   @shared_link_special_names ["WordPress - Shared Dashboard"]
   @doc """
   Special shared link names are used to distinguish between those


lib/plausible_web/templates/site/settings_goals.html.heex
```diff
@@ -13,7 +13,7 @@
       <p>
         Define actions that you want your users to take, like visiting a certain page, submitting a form, etc.
       </p>
-      <p :if={ee?()}>
+      <p :if={ee?() and Plausible.Sites.regular?(@site)} data-test-id="setup-funnels-cta">
```
aerosol, line 16: nice catch

lib/plausible_web/templates/layout/site_settings.html.heex
```diff
@@ -8,7 +8,10 @@
   </.styled_link>
   <div class="pb-5 border-b border-gray-200 dark:border-gray-750">
     <h2 class="text-2xl font-bold leading-7 text-gray-900 dark:text-gray-100 sm:text-3xl sm:leading-9 sm:truncate">
-      Settings for {@site.domain}
+      Settings for {if(Plausible.Sites.regular?(@site),
```
aerosol: It's minor and totally a nit-pick, but how about we check for `consolidated?` here and fall back to a domain instead? Just in case a third option ever emerges :)

test/plausible_web/controllers/site_controller_test.exs
```diff
@@ -591,6 +592,8 @@ defmodule PlausibleWeb.SiteControllerTest do
         |> find("[data-testid=site_settings_sidebar] a")
         |> Enum.map(fn a -> {text(a), text_of_attr(a, "href")} end)
 
+      assert resp =~ "Settings for the consolidated view"
```
aerosol: :ok_hand:

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -621,4 +618,7 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
     </.tooltip>
     """
   end
+
+  defp editing_non_revenue_goal?(%{goal: %{currency: nil}} = _assigns), do: true
```
aerosol: There's `Plausible.Goal.Revenue.revenue?` even used in this module already.

lib/plausible/goals/goals.ex
```diff
@@ -353,6 +354,15 @@ defmodule Plausible.Goals do
     end
   end
 
+  defp check_no_currency_if_consolidated(site, changeset) do
+    if not is_nil(Ecto.Changeset.get_field(changeset, :currency)) and
```
aerosol: The `not is_nil` seems to be unnecessary? Also flip the order of those predicates, so that we circuit break sooner.

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -111,7 +111,9 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
       phx-submit="save-goal"
       phx-target={@myself}
     >
-      <.title>Add goal for {@domain}</.title>
+      <.title>
+        Add goal for {if(Plausible.Sites.regular?(@site), do: @domain, else: "consolidated view")}
```
aerosol: ditto, maybe explicitly check for `consolidated?` instead?

lib/plausible/props.ex
```diff
@@ -186,6 +186,21 @@ defmodule Plausible.Props do
     )
   end
 
+  on_ee do
+    defp site_id_filter(%Plausible.Site{} = site) do
+      if Plausible.Sites.consolidated?(site) do
+        site_ids = Plausible.ConsolidatedView.Cache.get(site.team.identifier)
```
aerosol: Wondering if `site.domain`, while technically the same, would be more appropriate at this level. I'm thinking we could also change other call sites by the way. The reasoning being, we have access to the view-site itself, and the cache is built on per-site basis:

```elixir
      select: %{
        consolidated_view_id: sc.domain,
```

Another thing is, are we going to lose the db connection if the query is sent over the wire with `@max_sites_per_view`? If so, this demands some clever approach or a compromise.

Try something like:

```
for i <- 1..15000, do: new_site(owner: u, domain: "crash-#{i}.example.com")
```

And refresh/restart the cache to see what happens.

lib/plausible_web/live/props_settings/form.ex
```diff
@@ -56,7 +57,12 @@ defmodule PlausibleWeb.Live.PropsSettings.Form do
           phx-submit="allow-prop"
           phx-click-away="cancel-allow-prop"
         >
-          <.title>Add property for {@domain}</.title>
+          <.title>
+            Add property for {if(Plausible.Sites.regular?(@site),
```
aerosol: ditto 🙉

aerosol (commented):
Love it, @RobertJoonas, great work. Some suggestions inline. Once we follow up with route-level authentication enforcement this should be complete.

lib/plausible_web/templates/layout/site_settings.html.heex
RobertJoonas: Good call. Updated in all three places you mentioned. https://github.com/plausible/analytics/pull/5795/commits/bbc945ad3cf9d2edcc89975b8c7d8cc85de980ca

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: Agreed. https://github.com/plausible/analytics/pull/5795/commits/51db662512eabc2339feaa5995dad33fb74b41a5

lib/plausible/goals/goals.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5795/commits/a0b50a58f3f9ada8f0e77f550230a79059e3d673

lib/plausible_web/live/goal_settings/form.ex
RobertJoonas: Had to fix this up twice so ignore that commit. See the file in full diff instead: https://github.com/plausible/analytics/pull/5795/files#diff-5d8ab8130a88f328ae953c117f37965290817fd3280f173ee9dcaac46f2d37ec

lib/plausible/props.ex
RobertJoonas: Make sense to lookup site_ids by domain instead of site.team.identifier. https://github.com/plausible/analytics/pull/5795/commits/c675ec43aba19addd5b22cfb52b634acfc699a59

> Another thing is, are we going to lose the db connection if the query is sent over the wire with @max_sites_per_view? If so, this demands some clever approach or a compromise.

Didn't quite catch what you're saying here? But I did try with 15k sites and it worked well. Even confirmed that if I send a pageview with `props: %{author: "john"}` to `crash10000.com` then it shows up nicely in the prop suggestions for the consolidated view.

lib/plausible_web/live/props_settings/form.ex
```diff
@@ -56,7 +57,12 @@ defmodule PlausibleWeb.Live.PropsSettings.Form do
           phx-submit="allow-prop"
           phx-click-away="cancel-allow-prop"
         >
-          <.title>Add property for {@domain}</.title>
+          <.title>
+            Add property for {if(Plausible.Sites.consolidated?(@site),
```
ukutaht: Suggestion: Extract this as a function in `Plausible.Site`:

```elixir
  def display_name(site) do
    if consolidated?(site), do: "consolidated view", else: site.domain
  end
```

Then use it in goal settings, props settings, and also as the general header (I think it's OK to omit "the" from the title)

lib/plausible/props.ex
aerosol: Yeah I meant to check if the "props query" won't be too big with a huge list of site ids. But I understand you're confirming it's fine.

aerosol (approved):

PR #5800, opened 2025-10-13
Retry on 408 timeout from Browserless.io

apata:
### Changes

Starts retrying browserless.io 408 responses again. Experiments show that a request made in quick succession after a 408 timeout can succeed quickly. If it does not, there's now a hard cap on the check itself. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/installation_support/verifier-v2.js
+++ b/tracker/installation_support/verifier-v2.js
@@ -29,7 +29,7 @@ async function verifyPlausibleInstallation(options) {
 
   const disallowedByCsp = checkDisallowedByCSP(responseHeaders, cspHostToCheck)
 
-  stopAutoIgnore()
+  forceIgnoreWebdriverCondition()
   const { stopRecording, getInterceptedFetch } = startRecordingEventFetchCalls()
 
   const {
@@ -279,7 +279,7 @@ async function testPlausibleFunction({ timeoutMs, debug }) {
   })
 }
 
-function stopAutoIgnore() {
+function forceIgnoreWebdriverCondition() {
   window.__plausible = true
 }
 


extra/lib/plausible/installation_support/browserless_config.ex
```diff
@@ -8,6 +8,8 @@ defmodule Plausible.InstallationSupport.BrowserlessConfig do
     case status do
       # rate limit
       429 -> {:delay, 1000}
+      # timeout
+      408 -> {:delay, 500}
```
RobertJoonas: In detection, given that the endpoint timeout is 5s, and we also delay the retry request by 500ms, I don't think it's realistic that the retry will finish in 500ms. I don't think I ever saw a successful request that fast -- even with a minimal Puppeteer function.

In verification, it could help though.

apata: You're right, I had deluded myself into thinking maybe. I'll adjust it so 408 is retried in verification.

extra/lib/plausible/installation_support/browserless_config.ex
```diff
@@ -4,14 +4,39 @@ defmodule Plausible.InstallationSupport.BrowserlessConfig do
...
+    {:delay, :timer.seconds(1)}
+
+    iex> retry_browserless_request([429, 400]).(nil, %{status: 400})
+    {:delay, :timer.seconds(0.5)}
+
+    iex> retry_browserless_request([429, 400]).(nil, %{status: 408})
+    nil
+
+    iex> retry_browserless_request([429, 400]).(nil, :some_error)
+    false
```
RobertJoonas, line 29: Nit: let's change `nil` to `false` so that the return type is consistent. A typespec wouldn't hurt either.

apata: Thanks! Overall I agree with both thoughts and I had actually started optimising and typespeccing this utility function, but it's called only two times and it's probably not going to expand to tens of calls. Even now I'm actually wondering if we'd be better off with non-doctested individual anonymous retry functions at the two call sites.

PR #5806, opened 2025-10-14
Broadcast tracker script config updates

apata:
### Changes

Broadcasts tracker script config creates / updates to other app nodes, reducing the time between creating / updating a script and being able to reliably receive it. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/tracker.ex
+++ b/lib/plausible_web/tracker.ex
@@ -72,38 +72,11 @@ defmodule PlausibleWeb.Tracker do
 
   def build_script(nil), do: nil
 
-  on_ee do
-    defp broadcast_script_upsert(tracker_script_configuration, _site) do
-      PlausibleWeb.TrackerScriptCache.broadcast_put(
-        tracker_script_configuration.id,
-        PlausibleWeb.TrackerScriptCache.cache_content(tracker_script_configuration)
-      )
-    end
-  else
-    defp broadcast_script_upsert(tracker_script_configuration, site) do
-      tracker_script_configuration =
-        enrich_tracker_script_configuration_with_site(tracker_script_configuration, site)
-
-      PlausibleWeb.TrackerScriptCache.broadcast_put(
-        tracker_script_configuration.id,
-        PlausibleWeb.TrackerScriptCache.cache_content(tracker_script_configuration)
-      )
-    end
-
-    # This function enriches tracker script configuration with site domain if the site association has not been loaded.
-    # This can happen when the tracker script configuration is the output of update or insert queries.
-    defp enrich_tracker_script_configuration_with_site(tracker_script_configuration, site) do
-      case tracker_script_configuration do
-        %TrackerScriptConfiguration{site: %{domain: _domain}} ->
-          tracker_script_configuration
-
-        %TrackerScriptConfiguration{site_id: site_id} when site_id == site.id ->
-          struct(tracker_script_configuration, site: Map.take(site, [:domain]))
-
-        _ ->
-          tracker_script_configuration
-      end
-    end
+  defp broadcast_script_upsert(tracker_script_configuration) do
+    PlausibleWeb.TrackerScriptCache.broadcast_put(
+      tracker_script_configuration.id,
+      PlausibleWeb.TrackerScriptCache.cache_content(tracker_script_configuration)
+    )
   end
 
   def update_script_configuration(site, config_update, changeset_type) do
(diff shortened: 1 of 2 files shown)


lib/plausible/cache.ex
```diff
@@ -64,7 +64,7 @@ defmodule Plausible.Cache do
       alias Plausible.Cache.Adapter
       require Logger
 
-      @spec broadcast_put(any(), Keyword.t()) :: :ok
+      @spec broadcast_put(any(), any(), Keyword.t()) :: :ok
```
aerosol, line 67: 👍

lib/plausible_web/tracker.ex
```diff
@@ -67,6 +67,13 @@ defmodule PlausibleWeb.Tracker do
 
   def build_script(nil), do: nil
 
+  def broadcast_script_update(tracker_script_configuration),
+    do:
+      PlausibleWeb.TrackerScriptCache.broadcast_put(
```
aerosol: minor, just noticed this - by convention, `*Web` namespace if for HTTP things. Cache is completely agnostic in that sense, even if Web things are its clients.

lib/plausible_web/tracker_script_cache.ex
```diff
@@ -59,7 +59,7 @@ defmodule PlausibleWeb.TrackerScriptCache do
     end)
   end
 
-  defp cache_content(tracker_script_configuration) do
+  def cache_content(tracker_script_configuration) do
```
aerosol: a bit unclear what is this function for. returns either always `true` or some string for CE? And `build_script` for CE does eex-like string replace but not using [EEx](https://hexdocs.pm/eex/1.12.3/)? Very confusing. 🤔

apata: Thanks! Agreed that this could use a comment. Will put something like the following:

On EE, we cache that this particular script exists to prevent Postgres stress from lookups of non-existent script IDs. The heavy lifting of caching is left to the CDN. If the ID exists, we always generate a fresh version of the script. 

On CE, since we don't anticipate them using a CDN, we cache the rendered script. 

Regarding the CE solution, it means that every refresh interval, every script is re-rendered by the server. This doesn't seem ideal actually. 

Regarding building the script, we planned to use EEx but there were issues. There's a discussion on the PR that introduced it.

lib/plausible_web/tracker.ex
apata: Thanks, I hadn't considered the namespace question. At the moment we have the script at Plausible.Site.TrackerScriptConfiguration. By following existing naming principles, its Cache should be at Plausible.Site.TrackerScriptConfiguration.Cache, correct?

apata: Ah damn, it's not so straightforward. It caches the script, not the configuration. The script's module is PlausibleWeb.Tracker. So PlausibleWeb.Tracker.Cache or refactor also PlausibleWeb.Tracker to Plausible.Tracker?

aerosol: > refactor also PlausibleWeb.Tracker to Plausible.Tracker

Probably this, but we can live with not doing it

lib/plausible_web/tracker_script_cache.ex
apata: > Regarding the CE solution, it means that every refresh interval, every script is re-rendered by the server. This doesn't seem ideal actually.

Just realised that only every updated script is re-rendered by the server on CE. That's not a problem.

lib/plausible_web/tracker.ex
```diff
@@ -141,7 +144,11 @@ defmodule PlausibleWeb.Tracker do
                    params
                  )
                ) do
+          created_config =
+            Repo.preload(created_config, :site)
```
aerosol: Do we need the two extra db queries is `site` is already available?

apata: Thanks! 

I think we don't. In these functions, site is actually available as an argument, so theoretically we don't need additional DB queries at all. 

Refresh all updated scripts -> load tracker script config with site association, run generate_script/1
Get or create config -> (site is argument) receive tracker script config without site association, run generate_script/2
Update config -> (site is argument) receive tracker script config without site association, run generate_script/2

I started to refactor it but it got messy so I pushed the interim state.

aerosol: What I meant was 

```elixir
%{created_config | site: site}
```

if this doesn't carry any risk of outdated reads

apata: > if this doesn't carry any risk of outdated reads

Good question. 

What we need from the site association is site.domain for building the script for the cache

```
ON CE
T1 This conn: Starts process to update tracker script config
T2 Other conn: Starts process to update domain name
T3 Other conn: Finishes process to update domain name
T4 This conn: Expands stale site domain to tracker script
T5 This conn: Caches the stale tracker script and broadcasts this to all nodes
T6 Application: Checks for tracker_script_config entities that have been updated in the last 15 minutes, doesn't find anything
Result: stale tracker script cached until next refresh_all, which may be in 180mins

ON EE
T1 This conn: Starts process to update tracker script config
T2 Other conn: Starts process to update domain name
T3 Other conn: Finishes process to update domain name and starts a job that clears the cached tracker script on CDN in 10s
T4 This conn: Caches that a tracker script with this ID exists and broadcasts this to all nodes
T5 Application: Checks for tracker_script_config entities that have been updated in the last 15 minutes, doesn't find anything
Result: local cache is correct, CDN cache is purged
```

So there is some risk with outdated reads on CE. I don't think it's a huge risk, because we keep handling events with the stale domain for a while. 

But I also see other risks doing it like this. The site from the function arguments has a bunch more associations loaded. By setting it to the struct we can accidentally expose those associations down the line.

apata: @aerosol I know you already approved the version with the preloads, but I didn't think it was good enough, so I had another look at this. 

First I tried injecting the site from args only when it doesn't exist and is needed, accepting the small risk on CE from stale domain reads. 

Then I realized we don't need to accept that small risk: on CE we don't need to optimize postgres queries so much. 

I'm now proposing we reload the config with necessary associations after insert and update operations, but only on CE. I also changed the base query a little bit, so as not to select the whole site when only one field is needed.

PR #5808, opened 2025-10-15
Fix cache bust verification + function clause service error

RobertJoonas:
### Changes

* Make sure cache bust URL is only used in the cache bust check, not the initial installation check
* Make sure cache bust installation check fully replaces the diagnostics rather than adding them on top of the already existing diagnostics reported by `Checks.InstallationV2`
* Fix `service_error: :function_clause` showing up in diagnostics and make the retries work again. This was caused by passing a float delay to Req retries while they should be integers.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/types/query-api.d.ts
+++ b/assets/js/types/query-api.d.ts
@@ -213,11 +213,8 @@ export interface QueryApiSchema {
           match_day_of_week?: boolean;
           /**
            * If custom period. A list of two ISO8601 dates or timestamps to compare against.
-           *
-           * @minItems 2
-           * @maxItems 2
            */
-          date_range: [string, string];
+          date_range: DateTimeRange | DateRange;
         };
   };
   pagination?: {
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -128,7 +128,7 @@ defmodule Plausible.Stats.Comparisons do
   end
 
   defp get_comparison_date_range(source_query, %{mode: "custom"} = options) do
-    DateTimeRange.to_date_range(options.date_range, source_query.timezone)
+    DateTimeRange.new!(options.date_range.first, options.date_range.last, source_query.timezone)
   end
 
   defp maybe_match_day_of_week(comparison_date_range, source_date_range, options) do
--- a/lib/plausible/stats/datetime_range.ex
+++ b/lib/plausible/stats/datetime_range.ex
@@ -38,6 +38,10 @@ defmodule Plausible.Stats.DateTimeRange do
     new!(first, last)
   end
 
+  def new!(%DateTime{} = first, %DateTime{} = last, _timezone) do
+    new!(first, last)
+  end
+
   def new!(%DateTime{} = first, %DateTime{} = last) do
     first = DateTime.truncate(first, :second)
     last = DateTime.truncate(last, :second)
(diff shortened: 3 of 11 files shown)


extra/lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
```diff
@@ -31,16 +31,15 @@ defmodule Plausible.InstallationSupport.Checks.InstallationV2CacheBust do
         state
 
       _known_installation_failure ->
-        url_that_maybe_busts_cache =
-          Plausible.InstallationSupport.URL.bust_url(url)
+        reset_diagnostics = %InstallationSupport.Verification.Diagnostics{
+          selected_installation_type: state.diagnostics.selected_installation_type
```
RobertJoonas, line 35: This feels like a bit of a misfit in the diagnostics. Should we perhaps move it directly into the `state` struct rather than keeping it under `state.diagnostics`? (cc @apata)

apata: When I brought it in, I had the same thought, but we decided that since it is used in interpretation, it should be in diagnostics.

test/plausible/installation_support/verification/checks_test.exs
```diff
@@ -99,7 +113,54 @@ defmodule Plausible.InstallationSupport.Verification.ChecksTest do
...
+        end)
+
+        state =
+          Checks.run(url_to_verify, expected_domain, "manual",
+            report_to: nil,
+            async?: false,
+            slowdown: 0
+          )
+
+        assert is_nil(state.diagnostics.tracker_is_in_html)
```
apata: Nitpick, non-blocking: The other tests in this module use assert_matches rather than field by field. It’s common in the repo to do it field by field though, so your call. 

assert_matches has the advantage that it says this and only this, and on mismatch it outputs all the issues at once rather than only the first mismatching field.

test/plausible/installation_support/verification/checks_test.exs
```diff
@@ -99,7 +113,54 @@ defmodule Plausible.InstallationSupport.Verification.ChecksTest do
                          )
                          |> Checks.interpret_diagnostics()
 
-        assert 2 == :atomics.get(verification_counter, 1)
+        assert 2 == :atomics.get(counter, 1)
+      end
+
+      test "cache bust diagnostics fully replace the initial installation check diagnostics" do
```
apata, line 119: Typo: -> fully replaces

extra/lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
RobertJoonas: Alright, looks like this would become the fourth argument to the `Diagnostics.interpret` function, so we should probably make it `interpret/2` taking `diagnostics` and `state`... Maybe at a later stage. If this looks good now, I'd merge it.

test/plausible/installation_support/verification/checks_test.exs
RobertJoonas: IDK, to me sounds OK because "diagnostics" is plural? 😀

extra/lib/plausible/installation_support/checks/installation_v2_cache_bust.ex
apata: Though that move would also impact logging, since we currently log diagnostics and data_domain.

test/plausible/installation_support/verification/checks_test.exs
RobertJoonas: Good call! https://github.com/plausible/analytics/pull/5808/commits/5530d1d1072b355dc8d6f06f9683aedccd742386

test/plausible/installation_support/verification/checks_test.exs
apata: Actually my bad, it's not a typo

PR #5809, opened 2025-10-16
Add controller-level tests for timezone handling in Query

ukutaht:
From conversations with @aerosol we probably want to refactor Query internals to use NaiveDateTime so we don't carry around useless and confusing timezone information in the `Query` datetimes.

This PR adds some tests that give guardrails and increase confidence in timezone behaviour that is observable to clients:
* Relative date handling: `today` should be the current day in the site's timezone, not utc. Previously untested.
* Ambigous and gap handling in timezone conversions: Currently tested via unit tests internally which also use DateTime over Naive. Doesn't give much confidence for refactoring. New controller-level tests only test observable behaviour, not internals, helping much more with refactoring.
* Legacy time on page cutoff: the cutoff is shifted to site timezone with ambigous/gap handling. I don't know if this is really necessary, a couple hours here or there with the cutoff should not have noticeable effects on data. We could move this to Naive and ignore any timezones in cutoff handling. But I wanted to cover this behaviour with tests so we can be intentional about it in refactoring.

### DIFF
--- a/test/plausible_web/controllers/api/external_stats_controller/query_timezone_test.exs
+++ b/test/plausible_web/controllers/api/external_stats_controller/query_timezone_test.exs
@@ -164,6 +164,46 @@ defmodule PlausibleWeb.Api.ExternalStatsController.QueryTimezoneTest do
       context
     end
 
+    test "uses site timezone for legacy_time_on_page_cutoff conversion", %{
+      conn: conn,
+      site: site
+    } do
+      # Set site to EST timezone (UTC-5)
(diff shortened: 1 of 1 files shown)


ukutaht:
> Nice one! I admit I haven't checked if the comments are true though 😅

Fair, but I've verified that each test has a corresponding piece of production code that will cause the test to fail if the implementation changes. So we know that the tests are actually testing something :)

aerosol:
I'll merge since it fell out of the queue

PR #5812, opened 2025-10-16
Start passing official installs that are using legacy .compat script

apata:
### Changes

Makes verifier pass official installs that are not using fetch. 
This change is needed to pass current Plausible WP plugin installs if they are using the enhanced measurements option.
It can be removed once the newer version of Plausible WP is prevalent.
These installations will still receive an upgrade notification before verification, so the success is not misleading.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/tracker/installation_support/verifier-v2.js
+++ b/tracker/installation_support/verifier-v2.js
@@ -72,14 +72,15 @@ async function verifyPlausibleInstallation(options) {
       'script[data-domain][src]'
     )
     if (firstLegacySnippet) {
+      // legacy installations may list multiple domains in a comma-separated list
       const domainString = firstLegacySnippet.getAttribute('data-domain')
-      const domain = domainString && domainString.split(',').shift()
+      const firstDomain = domainString && domainString.split(',').shift()
 
       interceptedTestEvent = {
         request: {
           normalizedBody: {
             __legacyCompatInstallation: true,
-            domain
+            domain: firstDomain
           }
         },
         response: { status: testEvent.callbackResult.status }


tracker/installation_support/verifier-v2.js
```diff
@@ -54,12 +54,40 @@ async function verifyPlausibleInstallation(options) {
...
+            __legacyCompatInstallation: true,
+            domain
+          }
+        },
+        response: { status: testEvent.callbackResult.status }
+      }
+    }
+  }
+
+  log(interceptedTestEvent)
```
RobertJoonas: I don't think this is intentional?

tracker/installation_support/verifier-v2.js
```diff
@@ -54,12 +54,40 @@ async function verifyPlausibleInstallation(options) {
...
+    [200, 202].includes(testEvent.callbackResult?.status)
+  ) {
+    log(
+      `The callback result indicates a successful request, assuming legacy .compat installation that uses XMLHttpRequest`
+    )
+    const firstLegacySnippet = document.querySelector(
+      'script[data-domain][src]'
+    )
+    if (firstLegacySnippet) {
+      const domain = firstLegacySnippet.getAttribute('data-domain')
```
RobertJoonas: This needs some good old `String.split` magic because there can be multiple comma separated domains here.

apata: I see, I didn't even consider that case. Passing just one of the domains from the split seems about the same to the customer as not splitting at all (the verification would pass only on one of the domains and fail on the others). Should I maybe split diagnostics side? Multiple domains aren't supported on v2

apata: If we follow these possibilities further, we'd also check for multiple snippets, which starts to look like a bit of a rabbit hole. Maybe `domain && String.split(domain, ',').shift()` is good enough?

RobertJoonas: Actually, yes. I agree that it's a rabbit hole. Maybe we shouldn't even try to split it... As long as there are multiple domains involved, the user could see a domain mismatch error. Let's just take the stance that multiple comma-separated domains is a v1 feature that v2 verification is not supposed to cover.

RobertJoonas: > Maybe domain && String.split(domain, ',').shift() is good enough?

Or well, we can also do this as our best effort and stop there.

RobertJoonas: Perhaps a code comment would be useful though.

RobertJoonas (approved):

PR #5818, opened 2025-10-21
Script v2: Identify detection attempt

apata:
### Changes

Needed because our Wordpress plugin renders plugin meta HTML only when the query parameter is present.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/installation_support/detection/checks_test.exs
+++ b/test/plausible/installation_support/detection/checks_test.exs
@@ -13,7 +13,7 @@ defmodule Plausible.InstallationSupport.Detection.ChecksTest do
     @moduletag :capture_log
 
     describe "running detection" do
-      test "handles wordpress detection, retrying once one 429" do
+      test "handles wordpress detection, retrying on 429" do
         expected_domain = "example.com"
         url_to_verify = nil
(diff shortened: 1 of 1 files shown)


extra/lib/plausible/installation_support/checks/detection.ex
```diff
@@ -77,7 +77,7 @@ defmodule Plausible.InstallationSupport.Checks.Detection do
           Jason.encode!(%{
             code: @puppeteer_wrapper_code,
             context: %{
-              url: url,
+              url: Plausible.InstallationSupport.URL.bust_url(url),
```
aerosol, line 80: won't that render the whole cache busting step redundant?

apata: Thanks! There's two Checks pipelines now and an attempt to diagnose cache issues is made only in Verification.Checks, not in Detection.Checks. 

In the detection pipeline, we 1) verify the URL 2) (if URL is legit) have Browserless.io visit the website to sniff for legacy installation, wordpress, gtm signals.

RobertJoonas (approved):

PR #5819, opened 2025-10-22
Improve dark mode

sanne-san:
### Changes

- Switch from `slate` to `zinc` for the gray color palette
- Darken overall dark mode UI
- Switch from `green` to `emerald` for the green color palette
- Update a few previously missed instances of title case to sentence case
- Consolidate button styles and change naming from `bright` to `secondary`
- Update button disabled styles
- Fix tooltip not adjusting to content width
- Update graph tooltip layout and typography
- Add transition effects to hover states
- Reduce footer logo size

### Tests

### Changelog

### Dark mode

### DIFF
--- a/test/plausible_web/plugs/favicon_test.exs
+++ b/test/plausible_web/plugs/favicon_test.exs
@@ -6,8 +6,6 @@ defmodule PlausibleWeb.FaviconTest do
   import Mox
   setup :verify_on_exit!
 
-  @placeholder_icon File.read!("priv/link_favicon.svg")
-
   setup_all do
     opts = PlausibleWeb.Favicon.init(nil)
 
@@ -147,6 +145,8 @@ defmodule PlausibleWeb.FaviconTest do
   end
 
   describe "Fallback to placeholder icon" do
+    @placeholder_icon File.read!("priv/link_favicon.svg")
+
     test "falls back to placeholder when DDG returns a non-2xx response", %{plug_opts: plug_opts} do
       expect(
         Plausible.HTTPClient.Mock,


ukutaht:
Some comments from generally clicking around on the preview deployment:

**1. Chartjs tooltips**

Love the updates to the tooltips. But I think you missed that we also use chartjs tooltips for funnels and those are broken on this branch (I'm hovering over a funnel step in the screenshot, it should display a tooltip):
Screenshot 2025-10-22 at 17 38 55

**2. Main graph axis lines**

The horizontal lines for y-axis are very dark and almost invisible:

Screenshot 2025-10-22 at 17 40 44

This was the case before this PR but with everything getting even darker, I think the lines are even harder to see than before.

**3. Focus rings**

Again, not something that was introduced in this PR but since we're improving dark mode I wanted to highlight that the focus rings have a white and indigo part to them which looks weird:

Screenshot 2025-10-22 at 17 42 31

**4. Disabled inputs**

Disabled inputs are visually identical to regular inputs. This one for example:
Screenshot 2025-10-22 at 17 44 23

Looks the same as inputs that can be used.

assets/js/dashboard/components/popover.tsx
```diff
@@ -32,14 +32,14 @@ const panel = {
 
 const toggleButton = {
```
apata, line 33: nitpick, non-blocking: Tooltips within popovers in dark mode blend into the popover. I remember struggling to get it "close enough" with the older theme, so maybe it's unsolvable without refactoring the menu. I never liked hover tooltips: not good for responsive UI.

Image

Image

apata: BTW the commented line is random

sanne-san: Thanks, missed this one! Agree that maybe tooltip isn't ideal here but to avoid more scope creep I've updated the color for now :)

assets/js/dashboard/stats/modals/referrer-drilldown.js
```diff
@@ -81,11 +81,14 @@ function ReferrerDrilldownModal() {
   }
 
   const renderIcon = useCallback((listItem) => {
+    const sourceName = listItem.name.toLowerCase()
```
ukutaht: I agree it's quite hacky but I can't think of a better way.

I do think we should extract a `<Favicon>` component that handles the white background logic in order to reduce duplicating this logic.

sanne-san: Yes good point – done!

apata (approved):

apata:
The new dark mode looks very good! Approved. 

@sanne-san There's a few things I spotted though that are small enough to be tackled in additional PRs

1) Focusable area of one of the top chart icon buttons goes out of whack (left: this PR, right: staging).
Screenshot 2025-10-28 at 09 19 48
Screenshot 2025-10-28 at 09 20 30

2) Horizontal lines of top chart remain in the previous theme style (so they're extra prominent when swapping from dark to light).

PR #5822, opened 2025-10-23
Verification Fixes

RobertJoonas:
### Changes

NOTE: Reviewing commit by commit recommended

#### The only customer facing functional change

When the Browserless request responds with an unexpected status code (400 | 429), or the request times out, we'll show this to the customer:

image

#### Observability changes

1. Add the check module name into the `service_error` when our own Elixir check timeout kicks in. This makes it clearer in the logged diagnostics, whether it was cache bust or the initial installation_v2 check that timed out.

2. Set the Req `receive_timeout` option to the current `endpoint_timeout` which gets passed to the /function API request. With this we should stop getting 408s from Browserless. See this commit (description) for more info: https://github.com/plausible/analytics/commit/557f18f34f387d6d5223be26530a48d3393d884f

3. In case we arrive to the case displayed in the above screenshot, a warning log + telemetry `:unhandled` event are emitted (current behaviour). With this PR, we'll start notifying Sentry with a new error message that will be separate from the current `"Unhandled case for site verification (v2)"`. These new sentry errors, called `"Browserless failure in verification (v2)"` will be turned into [escalating issues](https://docs.sentry.io/product/issues/states-triage/escalating-issues/) in Sentry, because a certain volume of those events is expected.

#### Refactorings

* Make `verification/checks_test.exs` more DRY, also organise them into describe blocks
* Turn service error into a map with `code` and `extra` keys, where `code` should always be an atom.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/ingestion/persistor/remote.ex
+++ b/lib/plausible/ingestion/persistor/remote.ex
@@ -53,6 +53,10 @@ defmodule Plausible.Ingestion.Persistor.Remote do
     [:plausible, :remote_ingest, :request, :duration]
   end
 
+  def telemetry_decode_duration() do
+    [:plausible, :remote_ingest, :decode, :duration]
+  end
+
   defp request(override_url, event, session_attrs, headers) do
     Plausible.PromEx.Plugins.PlausibleMetrics.measure_duration(
       telemetry_request_duration(),
@@ -90,16 +94,21 @@ defmodule Plausible.Ingestion.Persistor.Remote do
   end
 
   defp decode_payload(payload) do
-    case Base.decode64(payload, padding: false) do
-      {:ok, data} ->
-        event_data = :erlang.binary_to_term(data)
-        event = struct(Plausible.ClickhouseEventV2, event_data)
+    Plausible.PromEx.Plugins.PlausibleMetrics.measure_duration(
+      telemetry_decode_duration(),
+      fn ->
+        case Base.decode64(payload, padding: false) do
+          {:ok, data} ->
+            event_data = :erlang.binary_to_term(data)
+            event = struct(Plausible.ClickhouseEventV2, event_data)
 
-        {:ok, event}
+            {:ok, event}
 
-      _ ->
-        {:error, :invalid_web_encoding}
-    end
+          _ ->
+            {:error, :invalid_web_encoding}
+        end
+      end
+    )
   catch
     _, _ ->
       {:error, :invalid_payload}
--- a/lib/plausible/telemetry/plausible_metrics.ex
+++ b/lib/plausible/telemetry/plausible_metrics.ex
@@ -277,6 +277,15 @@ defmodule Plausible.PromEx.Plugins.PlausibleMetrics do
         measurement: :duration
       ),
       distribution(
+        metric_prefix ++ [:persistor, :remote, :decode, :duration, :millisecond],
+        event_name: Persistor.Remote.telemetry_decode_duration(),
+        reporter_options: [
+          buckets: [1, 10, 25, 50, 75, 100, 250, 350, 500, 750, 1_000, 5_000, 10_000, 30_000]
+        ],
+        unit: {:native, :millisecond},
+        measurement: :duration
+      ),
+      distribution(
         metric_prefix ++ [:persistor, :remote, :request, :duration, :millisecond],
         event_name: Persistor.TelemetryHandler.request_event(),
         reporter_options: [
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -489,7 +489,7 @@ defmodule PlausibleWeb.Components.Generic do
         <.title>
           {render_slot(@title)}
 
-          <.docs_info :if={@docs} slug={@docs} class="absolute top-4 right-4" />
+          <.docs_info :if={@docs} slug={@docs} class="absolute top-4 right-4 z-1" />
         </.title>
         <div :if={@subtitle != []} class="text-sm mt-px text-gray-500 dark:text-gray-400 leading-5">
           {render_slot(@subtitle)}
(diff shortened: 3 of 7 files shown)


extra/lib/plausible/installation_support/checks/installation_v2.ex
```diff
@@ -74,8 +74,8 @@ defmodule Plausible.InstallationSupport.Checks.InstallationV2 do
   # To support browserless API being unavailable or overloaded, we retry the endpoint call if it doesn't return a successful response
   @max_retries 1
 
-  # We define a timeout for the browserless endpoint call to avoid waiting too long for a response
-  @endpoint_timeout_ms 15_000
+  # We define a timeout for the Browserless endpoint call to avoid waiting too long for a response
+  @req_timeout 15_000
```
apata, line 79: Does Browserless.io tear down the environment if we timeout on our side? Or does it keep running (and therefore wasting resources)? If it keeps running, we still may want both to be in place.

test/plausible/installation_support/verification/checks_test.exs
```diff
@@ -10,159 +10,88 @@ defmodule Plausible.InstallationSupport.Verification.ChecksTest do
...
-          assert %Result{
-                   ok?: true
-                 } ==
-                   Checks.run(url_to_verify, expected_domain, "manual",
-                     report_to: nil,
-                     async?: false,
-                     slowdown: 0
-                   )
-                   |> Checks.interpret_diagnostics()
+    describe "URL check" do
```
apata, line 19: nitpick, non-blocking: This test module is testing Checks.run (and interpret_diagnostics). I think it shouldn't be split with `describe` based on what individual checks are in the pipeline. 

If we want this type of isolation, we could test each of those checks individually.

test/plausible/installation_support/verification/checks_test.exs
```diff
@@ -575,30 +370,299 @@ defmodule Plausible.InstallationSupport.Verification.ChecksTest do
...
+                             }
+                           ]
+                         } = Checks.interpret_diagnostics(state)
         end
       end
     end
 
-    defp stub_verification_result(js_data) when is_map(js_data) do
-      counter = :atomics.new(1, [])
+    describe "observability" do
```
apata, line 532: nitpick, non-blocking: I know this is a controversial topic, but I'd prefer if a test presented the situation (what the domain is, what the A records are, what the Browserless.io responses are), then asserted on all the things we expect to happen in that scenario: result, telemetry, logs, request counts.

It makes for longer individual tests, but then we don't have to wonder whether it's not asserted because it doesn't happen or whether it's because there's another test somewhere that sets up the same situation to assert only on some of the expected output.

extra/lib/plausible/installation_support/verification/checks.ex
apata, line 83: issue: Why do we need the different Sentry message actually? Both "Browserless failure in verification (v2)" and "Unhandled case for site verification (v2)" will have the same criteria for reaction from our side: if there's unusually high rate of these. I think investigating rare "Unhandled case for site verification (v2)" isn't worth it.

extra/lib/plausible/installation_support/checks/url.ex
```diff
@@ -22,12 +22,12 @@ defmodule Plausible.InstallationSupport.Checks.Url do
     else
       {:error, :no_a_record} ->
         put_diagnostics(%State{state | skip_further_checks?: true},
-          service_error: :domain_not_found
+          service_error: %{code: :domain_not_found}
```
apata, line 25: Nice! The random type of service_error bothered me as well

extra/lib/plausible/installation_support/checks/installation_v2.ex
RobertJoonas: Good point. Let's give the timeout for both then and let it be the same?

extra/lib/plausible/installation_support/verification/checks.ex
RobertJoonas: A failure related to Browserless is not directly actionable from our side but stumbling onto a diagnostics struct with something that we haven't accounted for, is indeed actionable. That's why I think they should be separate.

apata (approved):

extra/lib/plausible/installation_support/verification/checks.ex
apata: We didn't agree whether "Unhandled case for site verification" should be set to escalating, but it's no harm have two different Sentry messages. Can always unify back to one.

extra/lib/plausible/installation_support/checks/installation_v2.ex
apata: We agreed to set endpoint timeout param as well JIC

test/plausible/installation_support/verification/checks_test.exs
RobertJoonas: Discussed this in a call. The intent of the describe blocks is to highlight the check in focus, while all the pipeline behaviour is actually relevant.

test/plausible/installation_support/verification/checks_test.exs
RobertJoonas: Discussed this as well. The `expected_domain` and `url_to_verify` are just dummy values in this test file, so it's unnecessary to declare them again and again, or even to think about them when reading the tests. An exception is the first describe block (URL check) where the DNS lookup is mocked per domain. There, the domain is also declared in individual tests.

extra/lib/plausible/installation_support/checks/installation_v2.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5822/commits/afd57df11d0ae648429c5a35c5f35261b567760d

PR #5832, opened 2025-10-27
CRM: 24h charts + minor extensions

aerosol:
### Changes

This PR introduces several CRM extensions:
  - adds 24h charts to sites and consolidated view entries
  - moves "new custom plan" button to the top of the list
  - introduces custom plan deletion (local only), if plan doesn't belong to the subscription
  - adds managed proxy price modifier to custom plan estimation

Context:
 - https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/9113635206
 - https://3.basecamp.com/5308029/buckets/43891605/card_tables/cards/9198444751

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible/customer_support/enterprise_plan.ex
```diff
@@ -2,53 +2,54 @@ defmodule Plausible.CustomerSupport.EnterprisePlan do
...
-        pageviews_per_month,
-        sites_limit,
-        team_members_limit,
-        api_calls_limit,
-        features
-      ) do
+  @spec estimate(Keyword.t()) :: Decimal.t()
+  def estimate(basis) do
+    basis =
+      Keyword.validate!(basis, [
```
zoldar, line 8: TIL, nice

lib/plausible/billing/enterprise_plan.ex
```diff
@@ -23,6 +23,7 @@ defmodule Plausible.Billing.EnterprisePlan do
     field :team_member_limit, Plausible.Billing.Ecto.Limit
     field :features, {:array, Plausible.Billing.Ecto.Feature}, default: []
     field :hourly_api_request_limit, :integer
+    field :managed_proxy_price_modifier, :boolean, default: false, virtual: true
```
zoldar, line 26: So we don't preserve the managed proxy setting in plan definition and only use it for estimate? Is this going to become part of the schema later on?

aerosol: No, @metmarkosaric said it's just price estimation modifier, I wasn't sure if there's a better way to handle it?

zoldar: No it's fine. In theory you could set `name` and `value` (using `@socket.params`) on the checkbox instead of fields, but it's not worth the extra mess.

PR #5833, opened 2025-10-27
Clean up detection sentry events + tests

RobertJoonas:
### Changes

This PR cleans up Sentry reporting for detection. Doing so, we're also changing the way we observe detection failures. Instead of "unhandled vs handled" (which makes sense in verification), a detection result is either a success or a failure. The telemetry events are also renamed accordingly in this PR.

A detection failure can be two things - an issue with the customer website, or an issue on our side. The issues on our side are further split into Browserless issues vs our own. Hence the telemetry bit is branched out quite a lot, but I think it should help us prioritize any future issues better.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/installation_support/detection/diagnostics.ex
+++ b/extra/lib/plausible/installation_support/detection/diagnostics.ex
@@ -61,7 +61,7 @@ defmodule Plausible.InstallationSupport.Detection.Diagnostics do
 
   def interpret(%__MODULE__{service_error: %{code: code}}, _url)
       when code in [:domain_not_found, :invalid_url] do
-    failure(:client_issue)
+    failure(:customer_website_issue)
   end
 
   def interpret(%__MODULE__{service_error: %{code: code}}, _url)
@@ -74,9 +74,14 @@ defmodule Plausible.InstallationSupport.Detection.Diagnostics do
         _url
       ) do
     cond do
-      String.contains?(extra, "net::") -> failure(:client_issue)
-      String.contains?(String.downcase(extra), "execution context") -> failure(:client_issue)
-      true -> failure(:unknown_issue)
+      String.contains?(extra, "net::") ->
+        failure(:customer_website_issue)
+
+      String.contains?(String.downcase(extra), "execution context") ->
+        failure(:customer_website_issue)
+
+      true ->
+        failure(:unknown_issue)
     end
   end
 
--- a/extra/lib/plausible/installation_support/detection/checks.ex
+++ b/extra/lib/plausible/installation_support/detection/checks.ex
@@ -52,8 +52,8 @@ defmodule Plausible.InstallationSupport.Detection.Checks do
         %{ok?: true} ->
           {false, false, nil}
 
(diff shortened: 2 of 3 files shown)


extra/lib/plausible/installation_support/detection/diagnostics.ex
```diff
@@ -56,29 +56,41 @@ defmodule Plausible.InstallationSupport.Detection.Diagnostics do
...
-        },
+        %__MODULE__{service_error: %{code: :browserless_client_error, extra: extra}},
         _url
-      )
-      when service_error in [:domain_not_found, :invalid_url],
-      do: %Result{ok?: false, data: nil, errors: [Atom.to_string(service_error)]}
+      ) do
+    cond do
+      String.contains?(extra, "net::") -> failure(:client_issue)
+      String.contains?(String.downcase(extra), "execution context") -> failure(:client_issue)
```
apata: nitpick, non-blocking: This remap from :browserless_client_error to :client_issue and :unknown_issue loses some information in the process. At this point, extra is known, but once we map it to :unknown_issue, it gets thrown away and in the checks module we go to the case 

```
        _unknown_failure ->
          {true, true, "Unknown failure"}
```

It seems there's actually two codes that we want to emit from the Detection check, :browserless_client_error and :browserless_client_error_silenced

extra/lib/plausible/installation_support/detection/diagnostics.ex
```diff
@@ -56,29 +56,41 @@ defmodule Plausible.InstallationSupport.Detection.Diagnostics do
         } = diagnostics,
         _url
       ) do
-    get_result(PlausibleWeb.Tracker.fallback_installation_type(), diagnostics)
+    success(PlausibleWeb.Tracker.fallback_installation_type(), diagnostics)
+  end
+
+  def interpret(%__MODULE__{service_error: %{code: code}}, _url)
+      when code in [:domain_not_found, :invalid_url] do
+    failure(:client_issue)
```
apata: nitpick, non-blocking: this does not seem like a client issue (browser driven by Browserless.io) issue, it's invalid input issue: we don't even start the browser in these scenarios

extra/lib/plausible/installation_support/detection/diagnostics.ex
RobertJoonas: The information doesn't get lost. These error-grouping atoms are just for determining which kind of telemetry we want. In any case, if it's an interesting case, the whole diagnostics struct (e.g. `%{..., service_error: %{code: :browserless_client_error, extra: "whatever message"}}, ...`) is logged and captured by Sentry as well.

extra/lib/plausible/installation_support/detection/diagnostics.ex
RobertJoonas: Fair point, thanks! I've renamed it to `:customer_website_issue` which should make it more clear. https://github.com/plausible/analytics/pull/5833/commits/04942e686431432bf0990136a98431fea7c4c24f

PR #5836, opened 2025-10-27
Verification / Detection follow-ups

RobertJoonas:
### Changes

1. Consider internal check timeouts to be Browserless issues in both detection and verification
2. Assert on captured sentry messages

Note: This PR makes the `timeout_ms` (that is currently fixed for each InstallationSupport.Check module) configurable on the CheckRunner level. This will also be useful for increasing detection timeout for the domain change flow. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/consolidated_view.ex
+++ b/extra/lib/plausible/consolidated_view.ex
@@ -26,6 +26,7 @@ defmodule Plausible.ConsolidatedView do
         if has_sites_to_consolidate?(team) do
           consolidated_view
           |> change_stats_dates(team)
+          |> change_timezone(majority_sites_timezone(team))
           |> bump_updated_at()
           |> Repo.update!()
         else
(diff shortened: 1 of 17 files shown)


test/plausible/installation_support/verification/checks_observability_test.exs
```diff
@@ -0,0 +1,188 @@
+defmodule Plausible.InstallationSupport.Verification.ChecksObservabilityTest do
```
apata, line 1: suggestion, non-blocking: checks_test, checks_observability_test for both Verification and Detection have quite a bit of similar code, like stubbing, test sentry conf, run_checks, test scenarios. 

I think we could cut down on this duplication a lot by doing the following:
In the main checks_test modules, we make sure that expected %Result's are achieved. 
In o11y-related unit tests, we make sure that given certain %Result's, handle_result triggers certain logs, metrics etc

RobertJoonas: Fair suggestion by all means, but I personally like how these tests do not worry about any intermediate results and focus solely on how a certain Browserless response gets handled. They're still fast, cheap and easy to follow.

PR #5838, opened 2025-10-28
Update site switcher UI to accommodate for consolidated view

sanne-san:
### Changes

- `Back to sites` and `Site settings` are styled as buttons at the top, replacing the `View all` link at the bottom. This is to avoid confusion with the new `All sites` consolidated view.
- As a first item in the lists of sites, `All sites` will show with a `0` shortcut.
- The popover styling has been updated, rendering the `roundedStart`, `roundedEnd` and `groupRoundedEnd` unnecessary.

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/stats_controller.ex
+++ b/lib/plausible_web/controllers/stats_controller.ex
@@ -1,5 +1,4 @@
 defmodule PlausibleWeb.StatsController do
-  sit
   use Plausible
 
   @moduledoc """


assets/js/dashboard/site-switcher.tsx
```diff
@@ -140,18 +202,24 @@ export const SiteSwitcher = () => {
...
-                    <span className="mr-auto">Site settings</span>
+                    <Cog8ToothIcon className="size-4 mr-1.5" />
+                    Site settings
                   </a>
-                  <MenuSeparator />
-                </>
-              )}
+                )}
+              </div>
+              {(canSeeSiteSettings || canSeeViewAllSites) && <MenuSeparator />}
```
RobertJoonas, line 222: Not related to the changes in this PR but I think when both `canSeeSiteSettings` and `canSeeViewAllSites` are false (E.g.: see the site picker in [this shared link](https://plausible.io/share/plausible.io?auth=IQ_8cgFbRr2EjLAGkMBGv)), then we should not render the site picker as a dropdown at all? There's nothing useful to render there, other than the current site.

ukutaht (approved):

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -392,6 +402,17 @@ defmodule PlausibleWeb.StatsController do
...
         embedded? = conn.params["embed"] == "true"
 
+        consolidated_view? =
+          on_ee(do: Plausible.Sites.consolidated?(shared_link.site), else: false)
+
+        team_has_consolidated_view? =
+          on_ee(
+            do: Plausible.ConsolidatedView.ok_to_display?(shared_link.site.team, current_user),
+            else: false
+          )
```
RobertJoonas: @aerosol with `Plausible.Sites.consolidated?` we took the approach to always make it return false on CE, so the call sites don't have to worry about it. How about doing the same with `ok_to_display?`? Doing so, we could avoid the (extra) `on_ee` calls here, and also the private `consolidated_view_ok_to_display?` wrapper function `live/sites.ex`. WDYT?

RobertJoonas (approved):
Looks good to me! Nice work with the `popover.tsx` styling! 👌

zoldar (approved):

lib/plausible_web/controllers/stats_controller.ex
aerosol: Do you mean move `ok_to_display?` to `Plausible.Sites`? Or move `consolidated_view.ex` out of `extra/`?

aerosol: There's no super clean way forward with all the CE/EE branching unfortunately, I'm going to merge as is and we can get back to it.
Good point though, that `Plausible.Sites.consolidated?` is available at all times, so I'll skip the `on_ee` macro in there. 2139aeec7f

assets/js/dashboard/site-switcher.tsx
aerosol: good call, we can file a separate card for this

PR #5843, opened 2025-10-28
Use Prima for invitation modal

ukutaht:
### Changes

Use [Prima](https://github.com/plausible/prima) for /sites page invitation modal.

### Tests

### Changelog

### Dark mode

### DIFF
--- a/assets/js/liveview/live_socket.js
+++ b/assets/js/liveview/live_socket.js
@@ -1,5 +1,5 @@
 /**
-  These 3 modules are resolved from '../deps' folder,
+  The modules below this comment block are resolved from '../deps' folder,
   which does not exist when running the lint command in Github CI
 */
 /* eslint-disable import/no-unresolved */


mix.lock
```diff
@@ -135,13 +136,15 @@
...
   "referrer_blocklist": {:git, "https://github.com/plausible/referrer-blocklist.git", "d6f52c225cccb4f04b80e3a5d588868ec234139d", []},
   "req": {:hex, :req, "0.5.6", "8fe1eead4a085510fe3d51ad854ca8f20a622aae46e97b302f499dfb84f726ac", [:mix], [{:brotli, "~> 0.3.1", [hex: :brotli, repo: "hexpm", optional: true]}, {:ezstd, "~> 1.0", [hex: :ezstd, repo: "hexpm", optional: true]}, {:finch, "~> 0.17", [hex: :finch, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}, {:mime, "~> 2.0.6 or ~> 2.1", [hex: :mime, repo: "hexpm", optional: false]}, {:nimble_csv, "~> 1.0", [hex: :nimble_csv, repo: "hexpm", optional: true]}, {:plug, "~> 1.0", [hex: :plug, repo: "hexpm", optional: true]}], "hexpm", "cfaa8e720945d46654853de39d368f40362c2641c4b2153c886418914b372185"},
+  "rustler": {:hex, :rustler, "0.37.1", "721434020c7f6f8e1cdc57f44f75c490435b01de96384f8ccb96043f12e8a7e0", [:mix], [{:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}], "hexpm", "24547e9b8640cf00e6a2071acb710f3e12ce0346692e45098d84d45cdb54fd79"},
```
aerosol: Can we avoid that dependency by making it test/dev only in upstream repo?

aerosol: Ah I see it's used for the code-block component but the syntax highlighter dependency also marked as optional, hm

ukutaht: Yes I agree. Autumn and rustler should not be included in the library.

There's also another related problem. This compilation warning when used as a dependency:

```
==> prima
Compiling 5 files (.ex)
    warning: PrimaWeb.Endpoint.config_change/2 is undefined (module PrimaWeb.Endpoint is not available or is yet to be defined)
    │
 27 │       PrimaWeb.Endpoint.config_change(changed, removed)
    │                         ~
    │
    └─ lib/prima/application.ex:27:25: Prima.Application.config_change/3
```

Basically I haven't figured out a clean separation between packaging `prima` for library vs the demo application.

aerosol: I think it'd be OK to put the demo application under `example/` or `demo/` with its own root.

zoldar:
image

The buttons don't seem to align properly in mobile view.

mix.lock
ukutaht: @aerosol Can you elaborate what you mean?

zoldar: Sorry for butting in - but what I _think_ he means is that the demo application can be put in a subdirectory in project's root, like "<project_root>/example" and be a completely separate app created with "mix new ...". In the demo app's mix.exs, "prima" can be added as a dependency via relative path, like:

```elixir
  defp deps do
    [
      {:prima, path: "..", override: true},
      # ....
    ]
```

with all other demo-specific depedencies, completely separate from the library.

Ecto (among others) uses a pattern like this:

https://github.com/elixir-ecto/ecto/tree/master/examples/friends

aerosol: Yeah exactly, sorry for not expressing it clearly enough

ukutaht: Thanks @aerosol @zoldar! I've implemented the separation as suggested. Now the library is at the root and the isolated demo app at [prima/demo](https://github.com/plausible/prima/tree/main/demo) with its own mix.exs.

This has solved both problems:
1. Autumn and rustler at no longer included as transitive deps
2. Warning about `Prima.Endpoint.config_change` is gone

ukutaht:
> The buttons don't seem to align properly in mobile view.

@zoldar fixed in https://github.com/plausible/analytics/pull/5843/commits/7e75fecf5d1465265a4e761c044ae299c0a5a468

The PR is now ready for final review from my perspective

assets/js/liveview/live_socket.js
```diff
@@ -1,11 +1,12 @@
 /**
-  These 3 modules are resolved from '../deps' folder, 
-  which does not exist when running the lint command in Github CI 
+  These 3 modules are resolved from '../deps' folder,
```
apata:
```suggestion
  The modules below this comment block are resolved from '../deps' folder,
```

lib/plausible_web/live/components/prima_modal.ex
```diff
@@ -0,0 +1,49 @@
...
+  use PlausibleWeb, :component
+  alias Prima.Modal
+
+  attr :id, :string, required: true
+  slot :inner_block, required: true
+
+  def modal(assigns) do
+    ~H"""
+    <Modal.modal id={@id}>
+      <Modal.modal_overlay
```
apata, line 12: This is totally non-blocking since it's not a Prima-integration issue, but on my machine, the account menu opener button and Plausible icon/button in the top nav appear above the overlay, which I think is unusual UX.

ukutaht: Yes this is because the navbar has a z-index of 20 and I'm not sure why. This behaviour did not change, the previous alpine.js modal backdrop had the same issue.

In a future PR I will replace the navbar dropdown with a Prima one which uses a portal for the menu, so z-index should be unnecessary.

PR #5844, opened 2025-10-28
Fix migration for CE 3.0.1 to current

apata:
### Changes

In the process of figuring out the new CE release, I discovered that the migration `BackfillTrackerScriptConfiguration` failed on an entry in the database created with CE version 3.0.1. This fixes it.

### Tests
- [x] Tested manually by 
  - booting a fresh 3.0.1 CE
  - creating a website
  - locally building a new CE image from this branch
  - replacing the 3.0.1 image with new CE image in CE compose file

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/data_migration/backfill_tracker_script_configuration.ex
+++ b/lib/plausible/data_migration/backfill_tracker_script_configuration.ex
@@ -74,18 +74,14 @@ defmodule Plausible.DataMigration.BackfillTrackerScriptConfiguration do
     installation_meta = site.installation_meta
 
     installation_type =
-      if(installation_meta,
-        do:
-          Map.get(installation_meta, "installation_type", false) ||
-            installation_meta.installation_type,
-        else: nil
-      )
+      if installation_meta do
+        Map.get(installation_meta, "installation_type")
+      end
 
     script_config =
-      if(installation_meta,
-        do: Map.get(installation_meta, "script_config", false) || installation_meta.script_config,
-        else: %{}
-      )
+      if installation_meta do
+        Map.get(installation_meta, "script_config", %{})
+      end
 
     %{
       id: Nanoid.generate(),


lib/plausible/data_migration/backfill_tracker_script_configuration.ex
```diff
@@ -72,8 +72,20 @@ defmodule Plausible.DataMigration.BackfillTrackerScriptConfiguration do
...
+          Map.get(installation_meta, "installation_type", false) ||
+            installation_meta.installation_type,
+        else: nil
+      )
+
+    script_config =
+      if(installation_meta,
+        do: Map.get(installation_meta, "script_config", false) || installation_meta.script_config,
+        else: %{}
+      )
```
zoldar: AFAIU, site.installation_meta originally was an embed?

If it's not anymore then "dot" access version is basically dead code and both statements can be simplified to something like:

```suggestion

    installation_type =
      if installation_meta do
        Map.get(installation_meta, "installation_type", false)
      end

    script_config =
      if installation_meta do
        Map.get(installation_meta, "script_config", %{})
      else
        %{}
      end
```

wdyt?

apata: Thanks! I will apply your suggestion with a little edit: we want to default installation_type to nil, not false. https://github.com/plausible/analytics/commit/00a1d807958430478e6f3a392f91aedd39a72dec

This backfill originally used `from(s in Plausible.Site ...` https://github.com/plausible/analytics/commit/93dfb21669d0232ad2544664a67221234b48720c which caused the field installation_meta to be returned as struct. When this migration was run on prod, the dot syntax was the correct way to access the values. 

However, that query didn't stand the test of time apparently https://github.com/plausible/analytics/commit/bf587b05fd364c341cb72c06886a46b77071d83c#diff-3e51a2c082c1f655749fda812b9ebf59bb13fd73aa038dae18c05bef4a95286f, and it was changed to `from(s in "sites"` in which case installation_meta is returned as map with string keys. Since that time, I suppose the migration has been buggy even on EE and it's only lucky that it didn't fail in our dev envs.

PR #5849, opened 2025-10-29
Move shared link form to modal

sanne-san:
### Changes

- As a preliminary step for enabling adding a pinned segment to a shared link, the functionality to add or edit a shared link is being moved to a modal.
- Minor UI improvements, such as update the style of the edit and delete buttons.

### Tests

### Dark mode

### DIFF
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -879,7 +879,7 @@ defmodule PlausibleWeb.Components.Generic do
       case {assigns.inner_block, assigns[:icon_name]} do
         {[], nil} ->
           raise ArgumentError,
-                "Either `icon_name` attributre or icon inner block must be provided"
+                "Either `icon_name` attribute or icon inner block must be provided"
 
         {[_ | _], icon_name} when not is_nil(icon_name) ->
           raise ArgumentError, "Only one of `icon_name` and icon inner block must be provided"


sanne-san:
Relied heavily on the pattern used for adding a goal, please review thoroughly

lib/plausible_web/live/shared_link_settings.ex
```diff
@@ -0,0 +1,182 @@
...
+      ) do
+    socket =
+      socket
+      |> assign_new(:site, fn %{current_user: current_user} ->
+        current_user
+        |> Plausible.Sites.get_for_user!(domain, roles: [:owner, :admin, :editor, :super_admin])
+      end)
+      |> assign_new(:shared_links, fn %{site: site} ->
+        Plausible.Repo.all(
+          from(l in Plausible.Site.SharedLink,
```
zoldar, line 23: It would be good to apply a consistent sorting to that list, even if it's merely by shared link's ID like in https://github.com/plausible/analytics/blob/shared-link-modal-refactor-proposal/lib/plausible/plugins/api/shared_links.ex#L16-L16. Though, perhaps, sorting by names would make more sense in that context?

On that occasion, we could also consider if the query logic shouldn't be moved outside live component. Shared links miss their own "context" currently. Maybe introducing a module like "Plausible.Site.SharedLinks" and extracting the whole `Repo.all` call there under something like "list(site)" would be a good idea? That of course goes beyond review of this particular PR and we could tackle that in a completely separate follow-up. This would also let us address deduplicating common logic in other places where shared links are manipulated, like the API.

lib/plausible_web/live/shared_link_settings.ex
```diff
@@ -0,0 +1,182 @@
...
+    socket =
+      socket |> assign(form_shared_link: shared_link) |> Modal.open("shared-links-form-modal")
+
+    {:noreply, socket}
+  end
+
+  def handle_event("delete-shared-link", %{"slug" => slug}, socket) do
+    site_id = socket.assigns.site.id
+
+    case Plausible.Repo.delete_all(
```
zoldar, line 136: Ditto about moving deletion logic to a dedicated context module, like `Plausible.Site.SharedLinks`. Yet again, please feel free to ignore this entirely - I'm marking this more as a mental note for now. We definitely don't have to tackle this in the context of this PR, to avoid scope creep.

lib/plausible_web/live/shared_link_settings.ex
```diff
@@ -0,0 +1,182 @@
...
+          socket
+          |> put_live_flash(:error, "Could not find Shared Link")
+
+        {:noreply, socket}
+    end
+  end
+
+  def handle_info({:shared_link_added, _shared_link}, socket) do
+    shared_links =
+      Plausible.Repo.all(
```
zoldar, line 161: Note to self: ditto about extraction.

lib/plausible_web/live/shared_link_settings/form.ex
```diff
@@ -0,0 +1,118 @@
...
+    """
+  end
+
+  def edit_form(assigns) do
+    ~H"""
+    <.form :let={f} for={@form} phx-submit="save-shared-link" phx-target={@myself}>
+      <.title>Edit shared link</.title>
+
+      <.input field={f[:name]} label="Name" required="required" autocomplete="off" />
+      <.input
```
zoldar: The original implementation does not allow setting password for a shared link which was initially created without one. Was that a deliberate decision? 🤔 

If it is, there are 2 more things to consider here:

* The edited shared link shows this input as empty, regardless if the password was already set before or not. At least some visual indication ("placeholder" mask of big dots for instance) would help reducing confusion
* There's no way of turning password-protected shared link into a password-less one, only the other way around. Again, maybe that's deliberate, but it might be confusing to the user.

lib/plausible_web/live/shared_link_settings/form.ex
```diff
@@ -0,0 +1,118 @@
...
+    </.form>
+    """
+  end
+
+  def create_form(assigns) do
+    ~H"""
+    <.form :let={f} for={@form} phx-submit="save-shared-link" phx-target={@myself}>
+      <.title>New shared link</.title>
+      <.input field={f[:name]} label="Name" required="required" autocomplete="off" />
+      <.input
```
zoldar, line 58: From an UX point of view - would it make sense to hide password prompt behind a checkbox? It wouldn't influence the logic, just give a stronger visual cue that password is optional. Though it also adds to the friction with an extra step.

lib/plausible_web/components/generic.ex
```diff
@@ -366,7 +366,7 @@ defmodule PlausibleWeb.Components.Generic do
 
   attr(:href, :string, required: true)
   attr(:new_tab, :boolean, default: false)
-  attr(:class, :string, default: nil)
+  attr(:class, :any, default: nil)
```
zoldar: An alternative approach here could be:

```suggestion
  attr(:class, :string, default: "")
```

IMHO it's slightly better as it preserves the type validation.

lib/plausible_web/components/generic.ex
```diff
@@ -865,26 +865,59 @@ defmodule PlausibleWeb.Components.Generic do
   end
 
   attr :href, :string, default: nil
-  attr :icon, :atom, default: :pencil_square
+  attr :class, :any, default: nil
```
zoldar: Ditto about the alternative:

```suggestion
  attr :class, :string, default: ""
```

lib/plausible_web/components/generic.ex
```diff
@@ -865,26 +865,59 @@ defmodule PlausibleWeb.Components.Generic do
...
   attr :rest, :global, include: ~w(method disabled)
 
   def edit_button(assigns) do
```
zoldar: I've started wondering if the icon buttons could benefit from some shared logic enforcing baseline consistency and setting some constraints. I have explored this for a bit and came up with https://github.com/plausible/analytics/compare/shared-link-modal-refactor-proposal?expand=1#diff-b9ed778c43107a72f2570c95e1e5e4d50ac7e0cf36b71e7ae87fb13423506342. 

It's my attempt at trying to define an underlying `icon_button` component which adds enough flexibility to allow to define an arbitrary icon (via slots mechanism) but with a baseline of styling and theming shared among all icon buttons. I'm not sure if I have exposed all the right knobs so to speak, though 😅 

It's absolutely optional though and just an idea. Please feel free to dismiss for now.

lib/plausible_web/live/shared_link_settings/form.ex
sanne-san: Definitely, great suggestion! I limited this PR to just moving everything over to a modal, but that's one of the next steps as part of [adding a segment to a shared link](https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/8492485011#__recording_9124363073).

lib/plausible_web/live/shared_link_settings.ex
sanne-san: Thanks! I'll look at sorting – moving the logic outside the live component is something I'd need help with so I'll leave that for now and we can pick that up as a separate item as you suggested :)

lib/plausible_web/live/shared_link_settings/form.ex
sanne-san: Great catch, it wasn't my intention to change that so will fix!

lib/plausible_web/components/generic.ex
zoldar: Ugh, sorry, I have missed the fact that the class attribute is not "flattened" into a single string and passed down to a component 🙈  Using any here and at the other spot is the right choice then. Again, sorry for trouble.

lib/plausible_web/components/generic.ex
zoldar: Ditto about `any` for class being the right call.

lib/plausible_web/components/generic.ex
sanne-san: Yeah this is great, thank you 🙏

PR #5850, opened 2025-10-29
Change domain flow improvements

RobertJoonas:
### Changes

Improve the domain change flow. Companion PR: https://github.com/plausible/docs/pull/644

Some screen recordings and screenshots can be found in https://3.basecamp.com/5308029/buckets/42034199/card_tables/cards/9149604457

* Styling improvements:
  * Do not change the footer style between the form and success views
  * Remove the yellow background notice and replace with just a paragraph
  * Always replace the loading spinner with an icon and a message:
    * Detection failure -> ⚠️ We could not reach your new domain
    * Detection success: v1 | NPM | WP -> ⚠️ Additional action required
    * Detection success: no v1 detected -> ✅ Your new domain should be tracking nicely
* Always render a link in the footer to the domain change documentation
* When detection fails, render a more generic notice (additional action required -> additonal action *may be* required)
* For a successful detection where no further action is required, link to a new section introduced in the docs PR - the domain change checklist

### Tests

### Changelog

### Documentation
- [x] https://github.com/plausible/docs/pull/644

### Dark mode

### DIFF
--- a/lib/plausible_web/live/change_domain.ex
+++ b/lib/plausible_web/live/change_domain.ex
@@ -143,7 +143,7 @@ defmodule PlausibleWeb.Live.ChangeDomain do
 
           <:failed>
             <div class="flex items-center">
-              <Heroicons.x_mark class="w-4 h-4 mr-2 font-bold text-red-500" />
+              <Heroicons.exclamation_triangle class="w-4 h-4 mr-2 text-yellow-500" />
               <span class="text-sm font-bold">
                 We could not reach your new domain
               </span>


lib/plausible_web/live/change_domain.ex
```diff
@@ -118,10 +142,26 @@ defmodule PlausibleWeb.Live.ChangeDomain do
           </:loading>
 
           <:failed>
-            <.generic_notice />
+            <div class="flex items-center">
```
apata, line 145: nitpick, non-blocking: 
Image 
This dialog seems confusing to me because it's both declaring a domain change success and issuing a red alert. I think if we subtitle the detection part with something like "Detecting tracker on your site", it'd be more clear.

lib/plausible_web/live/change_domain.ex
```diff
@@ -131,52 +171,113 @@ defmodule PlausibleWeb.Live.ChangeDomain do
...
+      </p>
       """
     end
 
-    defp generic_notice(assigns) do
-      assigns = assign(assigns, docs_link: @change_domain_docs_link)
+    defp npm_notice(assigns) do
+      ~H"""
+      <p class="mt-4 text-sm">
+        We've detected you're using our npm package. This means that you'll also need
```
apata: nitpick, non-blocking: This isn't an issue with this line, but "npm package" is inconsistently capitalised in the repo. The lowercase format should be appropriate https://github.com/npm/cli#faq-on-branding

apata: Alternatively, we could just say "We've detected you're using the module @plausible-analytics/tracker. This means that ...". 

Technically, one doesn't need NPM to use our ESM module.

lib/plausible_web/live/change_domain.ex
RobertJoonas: Let's use the 🤷 emoji instead of the red X? 🤣 Jokes aside, how about using the same yellow warning-like exclamation-triangle, as we do when further action is required?

RobertJoonas: https://github.com/plausible/analytics/pull/5850/commits/3c7f11c79b6d30909b2da8ea86df862fb33d7072

RobertJoonas: image

lib/plausible_web/live/change_domain.ex
RobertJoonas: > This isn't an issue with this line, but "npm package" is inconsistently capitalised in the repo. The lowercase format should be appropriate https://github.com/npm/cli#faq-on-branding

Today I learned!

> Alternatively, we could just say "We've detected you're using the module @plausible-analytics/tracker. This means that ...".

https://github.com/plausible/analytics/pull/5850/commits/fe35f879638b6f670f81f14d4caa773c7eed1932

PR #5855, opened 2025-10-30
Add ability to enforce 2FA for all members of the team

zoldar:
### Changes

~Depends on https://github.com/plausible/analytics/pull/5868~ _merged_

This PR introduces ability to enforce 2FA for members of the team. It can only be toggled by team owners.

There's a new section under "Team Settings" > "General" called "Force Two-Factor Authentication"

image

Enforcement can be toggled on and off at any time, but only by one of team owners:

image

Additionally, disabling enforcement of 2FA requires password confirmation for added security:

image

Members trying to access team with "Enforce 2FA" enabled but without 2FA setup yet are forcibly redirected to the first step of 2FA setup:

image

User in such situation can either finish the setup or switch to another team. However, whenever they switch to the team with enforcement enabled they will be redirected to the setup until they complete it.

An e-mail notification is sent to all team members (except the owner user triggering the change) to help clarify why they got locked out of their dashboard:

image

### Technicalities

The flag is stored in team's policy embed as an additional `force_2fa` property which defaults to false.

As this feature is available from both CE and EE, the `policy` embed is getting expose in CE as well. The column is added to CE in a CE-only migration which will be extracted from this PR once it's reviewed.

The gating is done by `RequireAccount` plug as the guarded routes perfectly overlap for both. The 2FA setup routes and team switch route are excluded from enforcement check, similarly like email verification routes are excluded for unverified user enforcement.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/settings_controller_test.exs
+++ b/test/plausible_web/controllers/settings_controller_test.exs
@@ -1706,7 +1706,7 @@ defmodule PlausibleWeb.SettingsControllerTest do
   describe "POST /team/force_2fa/enable" do
     setup [:create_user, :log_in, :create_team, :setup_team]
 
-    test "enables enforcing 2FA", %{conn: conn, team: team, user: user} do
+    test "enables enforcing 2FA", %{conn: conn, team: team} do
       refute team.policy.force_2fa
 
       conn = post(conn, Routes.settings_path(conn, :enable_team_force_2fa))
@@ -1717,23 +1717,31 @@ defmodule PlausibleWeb.SettingsControllerTest do
                "2FA is now required for all team members"
 
       assert Repo.reload!(team).policy.force_2fa
+    end
 
-      assert_matches [
-                       %{
-                         name: "force_2fa_enabled",
-                         user_id: ^user.id,
-                         team_id: ^team.id,
-                         actor_type: :user,
-                         change: %{
-                           "before" => %{"policy" => %{"force_2fa" => false}},
-                           "after" => %{"policy" => %{"after" => %{"force_2fa" => true}}}
+    on_ee do
+      test "adds entry to audit log", %{conn: conn, team: team, user: user} do
+        conn = post(conn, Routes.settings_path(conn, :enable_team_force_2fa))
+
+        assert redirected_to(conn, 302) == Routes.settings_path(conn, :team_general)
+
+        assert_matches [
+                         %{
+                           name: "force_2fa_enabled",
+                           user_id: ^user.id,
+                           team_id: ^team.id,
+                           actor_type: :user,
+                           change: %{
+                             "before" => %{"policy" => %{"force_2fa" => false}},
+                             "after" => %{"policy" => %{"after" => %{"force_2fa" => true}}}
+                           }
                          }
-                       }
-                     ] =
-                       Plausible.Audit.list_entries(
-                         entity: "Plausible.Teams.Team",
-                         entity_id: "#{team.id}"
-                       )
+                       ] =
+                         Plausible.Audit.list_entries(
+                           entity: "Plausible.Teams.Team",
+                           entity_id: "#{team.id}"
+                         )
+      end
     end
 
     test "sends e-mail to all other team members", %{conn: conn, team: team, user: user} do
@@ -1807,7 +1815,7 @@ defmodule PlausibleWeb.SettingsControllerTest do
       {:ok, user: user}
     end
 
-    test "disables enforcing 2FA", %{conn: conn, user: user, team: team} do
+    test "disables enforcing 2FA", %{conn: conn, team: team, user: user} do
       {:ok, team} = Plausible.Teams.enable_force_2fa(team, user)
 
       conn =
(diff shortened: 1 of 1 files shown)


lib/plausible/teams.ex
```diff
@@ -320,6 +320,52 @@ defmodule Plausible.Teams do
     )
   end
 
+  @spec force_2fa_enabled?(Teams.Team.t() | nil) :: :boolean
```
ukutaht: I believe the correct return value spec would be `boolean()` instead of `:boolean`.

lib/plausible/teams.ex
```diff
@@ -320,6 +320,52 @@ defmodule Plausible.Teams do
...
+          {:ok, Teams.Team.t()} | {:error, Ecto.Changeset.t()}
+  def enable_force_2fa(team, user) do
+    with {:ok, team} <- set_force_2fa(team, true) do
+      team
+      |> Teams.Memberships.all(exclude_guests?: true)
+      |> Enum.each(fn membership ->
+        if membership.user.id != user.id do
+          team
+          |> PlausibleWeb.Email.force_2fa_enabled(membership.user, user)
+          |> Plausible.Mailer.send()
```
ukutaht: If we don't care whether it succeeds or fails, `Plausible.Mailer.deliver_later/1` would make more sense here to send the emails asynchronously.

lib/plausible_web/templates/settings/team_general.html.heex
```diff
@@ -43,6 +43,86 @@
...
+      <a id="force-2fa">Force Two-Factor Authentication (2FA)</a>
+    </:title>
+    <:subtitle>
+      Increase account security by requiring all team members to enable 2FA.
+    </:subtitle>
+
+    <div
+      :if={@force_2fa_enabled? and @current_team_role == :owner}
+      id="disable-force-2fa"
+      x-data="{dsiableTeamForce2FAOpen: false}"
```
ukutaht: typo: `dsiableTeamForce2FAOpen` -> `disableTeamForce2FAOpen`

lib/plausible/teams.ex
```diff
@@ -320,6 +320,52 @@ defmodule Plausible.Teams do
...
+
+  defp set_force_2fa(team, enabled?) do
+    params = %{policy: %{force_2fa: enabled?}}
+
+    team
+    |> Ecto.Changeset.cast(params, [])
+    |> Ecto.Changeset.cast_embed(:policy,
+      with: &Teams.Policy.force_2fa_changeset(&1, &2.force_2fa)
+    )
+    |> Repo.update_with_audit("force_2fa_changed", %{team_id: team.id})
```
ukutaht: Suggestion: Include the `enabled?` key in audit logs. Currently it looks like it won't be able to tell from audit logs whether force_2fa was enabled or disabled.

ukutaht: It also looks like the audit entry is not tested currently. I wanted to make sure the audit entry has the correct `user_id` and `actor_type` inferred from the audit context.

lib/plausible_web/templates/email/force_2fa_enabled.html.heex
```diff
@@ -0,0 +1,8 @@
+{@enabling_user.email} has enabled a 2FA requirement for the "{@team.name}" team.<br /><br />
+Please
+<a href={Routes.auth_url(PlausibleWeb.Endpoint, :force_initiate_2fa_setup) <> "?__team=" <> @team.identifier}>
```
ukutaht: Less error-prone and guaranteed to be URL-safe:

```suggestion
<a href={Routes.auth_url(PlausibleWeb.Endpoint, :force_initiate_2fa_setup, %{"__team" => @team.identifier})}>
```

lib/plausible_web/templates/auth/force_initiate_2fa_setup.html.heex
```diff
@@ -0,0 +1,28 @@
+<script>
```
ukutaht, line 1: What's the need for this intermediate screen that redirects automatically? Why not take the user to 2FA setup screen directly?

test/plausible_web/controllers/settings_controller_test.exs
```diff
@@ -1703,6 +1703,229 @@ defmodule PlausibleWeb.SettingsControllerTest do
...
+        subject: "Your team now requires 2FA",
+        to: [nil: member1.email]
+      )
+
+      assert_email_delivered_with(
+        subject: "Your team now requires 2FA",
+        to: [nil: member2.email]
+      )
+
+      # guests are not notified becuause they are not affected
```
ukutaht: Surprised codespell doesn't catch these

```suggestion
      # guests are not notified because they are not affected
```

lib/plausible/teams.ex
zoldar: Ah, okay so that's why dialyzer was acting wonky ALOL \cc @aerosol

lib/plausible_web/templates/email/force_2fa_enabled.html.heex
zoldar: Yeah keyword version _usually_ works as well, however, some path helpers _don't_ accept parameters, hence I was a bit apprehensive here. I'll change it and see if it works, thanks.

lib/plausible_web/templates/auth/force_initiate_2fa_setup.html.heex
zoldar: I wanted to preserve the setup initiation being POST as it does generate the authentication key for instance.

lib/plausible_web/templates/settings/team_general.html.heex
zoldar: JFC 🤡 , thanks

lib/plausible/teams.ex
zoldar: https://github.com/plausible/analytics/pull/5855/commits/e31873ba920744ef62b4959e0667224198641f50

lib/plausible/teams.ex
zoldar: https://github.com/plausible/analytics/pull/5855/commits/d1fb649a676cfcb8e4514bf78634a3193b297073

lib/plausible_web/templates/settings/team_general.html.heex
zoldar: https://github.com/plausible/analytics/pull/5855/commits/7c3e75696bdc4f7e7952ea91a7e7349011d9418d

lib/plausible_web/templates/auth/force_initiate_2fa_setup.html.heex
ukutaht: I see, makes sense.

Currently when landing on this page, it redirects very fast which results in kind of a jarring flash on the screen.

UX suggestion: Add a 3-5 second delay before the form is submitted via JS

lib/plausible_web/templates/email/force_2fa_enabled.html.heex
zoldar: https://github.com/plausible/analytics/pull/5855/commits/9ed172fe7def0c4c5a75f8e96a970013b8777194

lib/plausible/teams.ex
zoldar: https://github.com/plausible/analytics/pull/5855/commits/bf024fff404a5acbd69b3af3a352c2c9b5701dbc adds 2 distinct audit log entries: "force_2fa_enabled" and "force_2fa_disabled". The missing policy field is now added to audit entry encoder too.

image

lib/plausible_web/templates/auth/force_initiate_2fa_setup.html.heex
zoldar: Introducing 3 second delay before submitting form in https://github.com/plausible/analytics/pull/5855/commits/efef45f69bdfd122be8116e49c4ead19b33c1ff1

PR #5858, opened 2025-11-03
Improve top bar design

sanne-san:
### Changes

- Improve button styles and consistency in top bar
- Update filter pill style
- Update filter menu style
- Change realtime view to show current visitors and last updated in top bar rather than in the graph tabs
- Update tooltip style and add delay to show tooltip
- Update main graph tabs style

### Tests

### Dark mode

### DIFF
--- a/assets/js/dashboard/nav-menu/filters-bar.tsx
+++ b/assets/js/dashboard/nav-menu/filters-bar.tsx
@@ -9,14 +9,14 @@ import { popover, BlurMenuButtonOnEscape } from '../components/popover'
 import { isSegmentFilter } from '../filtering/segments'
 import { useRoutelessModalsContext } from '../navigation/routeless-modals-context'
 import { DashboardQuery } from '../query'
+import { MenuSeparator } from './nav-menu-components'
 
 // Component structure is
 // `..[ filter (x) ]..[ filter (x) ]..[ three dot menu ]..`
 // where `..` represents an ideally equal length.
 // The following calculations guarantee that.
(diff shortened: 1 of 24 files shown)


assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -160,7 +160,6 @@ export const FiltersBar = ({ accessors }: FiltersBarProps) => {
             : null,
         seeMoreWidth:
           SEE_MORE_LEFT_MARGIN_PX +
-          SEE_MORE_WIDTH_PX +
```
apata, line 163: I know this is a draft, so feel free to ignore, but for the math to work out, this value needed to be known and used in the calculations.

If you open https://pr-5858.review.plausible.io/plausible.io?f=is,country,US&f=is,page,/open-source-website-analytics&f=is,region,US-TX&l=US,United%20States&l=US-TX,Texas at **550px width** (just random example), we expect the query period menu not to overflow the bar, but it does, even though it would be possible to avoid overflow if the Country is United States filter wrapped into the vertical menu. 
Left: PR, right: staging
Image

I know it's weird to hard code the width of  the button in PX: I used to dynamically measure it, but it created a loop where it first needed to be drawn (let's say with size-5), then measured, then maybe thrown away.

sanne-san: I see, thank you! That's good to know.

Closing this PR for now as we concluded that the top area will need a bit of a bigger rethink, probably after the migration.

PR #5866, opened 2025-11-04
Consolidated View life cycle + billing integration

aerosol:
### Changes

This PR integrates consolidated view life cycle with billing properties.
It relies on two migrations (~~included here still~~), extracted to separate PRs:

- https://github.com/plausible/analytics/pull/5878
- https://github.com/plausible/analytics/pull/5877
- https://github.com/plausible/analytics/pull/5881

Main changes:

- new feature flag: `:consolidated_view` - we no longer display anything based on `super_admin` role; flag is not setup on production yet, meaning existing consolidated views will disappear (and no CTAs will be shown).
- new billing feature: `ConsolidatedView`
- all existing plans are updated, so that business tier includes the newly added features - either through static definitions or database migration
- whenever `/sites` is visited an attempt is made to create consolidated view. If the team is eligible, and the feature flag is raised, the consolidated view card is shown. Otherwise a CTA card is displayed (courtesy of @sanne-san)
  - CTA card has several variants, depending on the context, but eventually leads to upgrade. Once the upgrade is performed, `/sites` will include the consolidated view card.
  - CTA card can be permanently dismissed per user and team; because local storage and cookies turned out to be very difficult to work with in this setup, the decision was made to introduce a new postgres table capable of storing user preferences in a team context (it's almost a 1:1 copy of the existing site user preferences table)
  - When CTA is dismissed, the "New Site" button becomes a prima drop-down (courtesy of @ukutaht) from which the user may restore it (by selecting `"+ New consolidated view"` dropdown item)
  - CRM now presents a more elaborate alert on deleting consolidated view, as well as the availability status (whether upgrade is required)
  - some minor improvements BTW: 
    - shared links enforce regular sites, 
    - accessing a consolidated view via pre-existing URL, while ineligible, redirects to `/sites`
    - at least 2 sites are now required to create (or even suggest) a consolidated view
    - `ConsolidatedView.enabled?/1` has been removed, since enabling doesn't mean availability 
     

**TODO:**
 - staging preview is currently unavailable due to pending migrations 
 - ~~there are visual issues with card plots and overlapping dropdowns~~

https://github.com/user-attachments/assets/d3d287ce-9773-483e-a336-e9b0e90b095a

 - ~~the dropdown doesn't respect dark mode~~

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/site-context.tsx
+++ b/assets/js/dashboard/site-context.tsx
@@ -29,7 +29,7 @@ export function parseSiteFromDataset(dataset: DOMStringMap): PlausibleSite {
 // Update this object when new feature flags are added to the frontend.
 type FeatureFlags = Record<never, boolean>
 
-const siteContextDefaultValue = {
+export const siteContextDefaultValue = {
   domain: '',
   /** offset in seconds from UTC at site load time, @example 7200 */
   offset: 0,
--- a/assets/js/dashboard/stats/behaviours/goal-conversions.js
+++ b/assets/js/dashboard/stats/behaviours/goal-conversions.js
@@ -51,7 +51,7 @@ function SpecialPropBreakdown({ prop, afterFetchData }) {
 
   function getExternalLinkUrlFactory() {
     if (prop === 'path') {
-      return (listItem) => url.externalLinkForPage(site.domain, listItem.name)
+      return (listItem) => url.externalLinkForPage(site, listItem.name)
     } else if (prop === 'search_query') {
       return null // WP Search Queries should not become external links
     } else {
--- a/assets/js/dashboard/stats/pages/index.js
+++ b/assets/js/dashboard/stats/pages/index.js
@@ -20,7 +20,7 @@ function EntryPages({ afterFetchData }) {
   }
 
   function getExternalLinkUrl(page) {
-    return url.externalLinkForPage(site.domain, page.name)
+    return url.externalLinkForPage(site, page.name)
   }
 
   function getFilterInfo(listItem) {
@@ -66,7 +66,7 @@ function ExitPages({ afterFetchData }) {
   }
 
   function getExternalLinkUrl(page) {
-    return url.externalLinkForPage(site.domain, page.name)
+    return url.externalLinkForPage(site, page.name)
   }
 
   function getFilterInfo(listItem) {
@@ -112,7 +112,7 @@ function TopPages({ afterFetchData }) {
   }
 
   function getExternalLinkUrl(page) {
-    return url.externalLinkForPage(site.domain, page.name)
+    return url.externalLinkForPage(site, page.name)
   }
 
   function getFilterInfo(listItem) {
--- a/assets/js/dashboard/util/url.test.ts
+++ b/assets/js/dashboard/util/url.test.ts
@@ -1,4 +1,5 @@
 import { apiPath, externalLinkForPage, isValidHttpUrl, trimURL } from './url'
+import { siteContextDefaultValue } from '../site-context'
 
 describe('apiPath', () => {
   it.each([
@@ -32,10 +33,19 @@ describe('externalLinkForPage', () => {
   ])(
     'when domain is %s and page is %s, it should return %s',
     (domain, page, expected) => {
-      const result = externalLinkForPage(domain, page)
+      const site = { ...siteContextDefaultValue, domain: domain }
+      const result = externalLinkForPage(site, page)
       expect(result).toBe(expected)
     }
   )
+
+  it('returns null for consolidated view', () => {
+    const consolidatedView = {
+      ...siteContextDefaultValue,
+      isConsolidatedView: true
+    }
+    expect(externalLinkForPage(consolidatedView, '/some-page')).toBe(null)
+  })
 })
 
 describe('isValidHttpUrl', () => {
--- a/assets/js/dashboard/util/url.ts
+++ b/assets/js/dashboard/util/url.ts
@@ -8,11 +8,15 @@ export function apiPath(
 }
 
 export function externalLinkForPage(
-  domain: PlausibleSite['domain'],
+  site: PlausibleSite,
   page: string
 ): string | null {
+  if (site.isConsolidatedView) {
+    return null
+  }
+
   try {
-    const domainURL = new URL(`https://${domain}`)
+    const domainURL = new URL(`https://${site.domain}`)
     return `https://${domainURL.host}${page}`
   } catch (_error) {
     return null


extra/lib/plausible/consolidated_view.ex
```diff
@@ -16,13 +16,33 @@ defmodule Plausible.ConsolidatedView do
...
+  end
+
   @spec ok_to_display?(Team.t() | nil, User.t() | nil) :: boolean()
   def ok_to_display?(team, user) do
     with %Team{} <- team,
          %User{} <- user,
-         true <- Plausible.Auth.is_super_admin?(user),
-         true <- enabled?(team),
-         true <- has_sites_to_consolidate?(team) do
+         true <- flag_enabled?(team),
```
RobertJoonas: The `user` argument here is unnecessary I believe. I would like to use this function when checking whether a stats report or traffic spike email should be sent for a consolidated view. There's no user in that context.

extra/lib/plausible_web/live/customer_support/team/components/consolidated_views.ex
```diff
@@ -74,7 +76,7 @@ defmodule PlausibleWeb.CustomerSupport.Team.Components.ConsolidatedViews do
               <.delete_button
                 phx-click="delete-consolidated-view"
                 phx-target={@myself}
-                data-confirm="Are you sure you want to delete this consolidated view?"
+                data-confirm="Are you sure you want to delete this consolidated view? It will be recreated whenever eligible subscription/trial accesses /sites for that team."
```
RobertJoonas: Should we also mention that the configuration gets lost with this action?

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -402,9 +405,9 @@ defmodule PlausibleWeb.StatsController do
...
         embedded? = conn.params["embed"] == "true"
 
-        consolidated_view? = Plausible.Sites.consolidated?(shared_link.site)
+        true = Plausible.Sites.regular?(shared_link.site)
 
-        team_has_consolidated_view? =
+        consolidated_view_available? =
           on_ee(
             do: Plausible.ConsolidatedView.ok_to_display?(shared_link.site.team, current_user),
             else: false
```
RobertJoonas: I think it's fine to just say that this is `false` for shared links too? Even if I'm logged in and seeing my own shared link, I will not be able to switch to the consolidated view in the site picker, or in fact, even see any of my other sites there. There's a login link at the top right though.

lib/plausible_web/live/sites.ex
```diff
@@ -89,11 +91,38 @@ defmodule PlausibleWeb.Live.Sites do
...
         </p>
-        <div class="mt-4 flex sm:ml-4 sm:mt-0">
-          <a href={"/sites/new?flow=#{PlausibleWeb.Flows.provisioning()}"} class="button">
-            + Add website
-          </a>
-        </div>
+        <!-- The `z-49` class is to make the dropdown appear above the site cards and (TODO) below the top-right drop down. -->
+          <!-- The proper solution is for Prima to render the dropdown menu within a <.portal> element to avoid -->
+          <!-- any stacking context issues. TODO  -->
+        <PrimaDropdown.dropdown
```
RobertJoonas, line 95: The dropdown needs a darkmode update:

Image

lib/plausible_web/live/sites.ex
```diff
@@ -190,29 +228,84 @@ defmodule PlausibleWeb.Live.Sites do
     """
   end
 
-  def upgrade_card(assigns) do
+  def consolidated_view_card_cta(assigns) do
```
RobertJoonas, line 234: There's a slight UI annoyance I'm seeing locally. Try to hide the CTA card, then navigate to another page, e.g. click on one of the site cards to see a dashboard. Now, hit back in the browser history. The card flickers for a second and then disappears. Same happens when you've hidden it before and then restore it.

We should make sure that hiding/restoring the card also updates the entry in browser history stack.

test/plausible_web/live/sites_test.exs
RobertJoonas: This is no longer necessar I think? Looks like a duplicate of "single site won't show neither CTA or view - team setup"

lib/plausible_web/controllers/stats_controller.ex
aerosol: Yeah I think you're right

extra/lib/plausible_web/live/customer_support/team/components/consolidated_views.ex
aerosol: Not sure. There has to be a reason we like to ever delete it. 
Also speaking of life cycle: currently if you transfer out a site, and you've got not enough sites to consolidate, we'll hard delete your view. Same will be triggered upon deleting regular sites. Do you think the users will expect the configuration to persist? Is it a significant loss?

lib/plausible_web/controllers/stats_controller.ex
aerosol: https://github.com/plausible/analytics/pull/5866/commits/00556c8b3ffbff311c6d45b8a6e949ce5f6dd5e5

extra/lib/plausible/consolidated_view.ex
aerosol: https://github.com/plausible/analytics/pull/5866/commits/93412ec3ba6e83525539973f44eec9a5f05b41f5 thanks!

lib/plausible_web/live/sites.ex
aerosol: Ah, it needs appropriate classes in `PlausibleWeb.Components.PrimaDropdown`

test/plausible_web/live/sites_test.exs
aerosol: Good shout, I think we can turn it into a "feature-flag down" test.

aerosol: c116447e9e

lib/plausible_web/live/sites.ex
```diff
@@ -115,9 +144,18 @@ defmodule PlausibleWeb.Live.Sites do
 
       <div :if={@has_sites?}>
         <ul class="my-6 grid grid-cols-1 gap-6 sm:grid-cols-2 lg:grid-cols-3">
-          <!-- Insert upgrade_card here -->
+          <.consolidated_view_card_cta
```
aerosol, line 146: I'll see if I can extract this to a separate component and maybe even extract the state management out of here. It's getting really noisy.

extra/lib/plausible_web/live/customer_support/team/components/consolidated_views.ex
RobertJoonas: > Also speaking of life cycle: currently if you transfer out a site, and you've got not enough sites to consolidate, we'll hard delete your view. Same will be triggered upon deleting regular sites. Do you think the users will expect the configuration to persist? Is it a significant loss?

I don't think that's a huge concern and in any case it can be tackled separately if it ever becomes a problem.

aerosol: https://github.com/plausible/analytics/pull/5866/commits/386c1735680a296902655eb6c29c1f835122eaec

lib/plausible_web/live/sites.ex
sanne-san: ✅

extra/lib/plausible/consolidated_view.ex
```diff
@@ -16,13 +16,32 @@ defmodule Plausible.ConsolidatedView do
...
+  def ok_to_display?(team) do
     with %Team{} <- team,
-         %User{} <- user,
-         true <- Plausible.Auth.is_super_admin?(user),
-         true <- enabled?(team),
-         true <- has_sites_to_consolidate?(team) do
+         true <- flag_enabled?(team),
+         true <- view_enabled?(team),
+         true <- has_sites_to_consolidate?(team),
+         :ok <- Plausible.Billing.Feature.ConsolidatedView.check_availability(team) do
```
zoldar: That's clever, but perhaps using a pure logical expression like:

```elixir
is_struct(team, Team) and 
  flag_enabled?(team) and
  view_enabled?(team) and
  has_sites_to_consolidate?(team) and
  Plausible.Billing.Feature.ConsolidatedView.check_availability(team) == :ok
```

would convey the intent better?

extra/lib/plausible/consolidated_view.ex
```diff
@@ -56,16 +75,12 @@ defmodule Plausible.ConsolidatedView do
     from(s in q, where: s.consolidated == true)
   end
 
-  @spec enable(Team.t()) :: {:ok, Site.t()} | {:error, :no_sites | :team_not_setup}
+  @spec enable(Team.t()) ::
+          {:ok, Site.t()} | {:error, :no_sites | :team_not_setup | :upgrade_required}
   def enable(%Team{} = team) do
     with :ok <- ensure_eligible(team), do: do_enable(team)
```
zoldar: As that's the only spot `ensure_eligible` is used so far - maybe let's inline it?

lib/plausible/teams/user_preference.ex
```diff
@@ -0,0 +1,30 @@
...
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @type t() :: %__MODULE__{}
+
+  @options [:consolidated_view_cta_dismissed]
+
+  schema "team_user_preferences" do
```
zoldar: While this way user retains their preference even if they leave the team and later join it, it has some cons:

- preferences linger for as long as user and team exists in the system, in case of user/site preferences, the user is much more likely to keep "following" the site even after it's transferred somewhere else; usually, when a user leaves a team, they leave it for good, I think
- the relation is more complex

So, like you suggested elsewhere, making it associated with membership (member_user_preferences?) might make more sense.

lib/plausible/teams/users.ex
```diff
@@ -7,6 +7,37 @@ defmodule Plausible.Teams.Users do
...
+  @spec set_preference(Plausible.Auth.User.t(), Teams.Team.t(), atom(), any()) ::
+          Teams.UserPreference.t()
+  def set_preference(user, team, option, value) when option in Teams.UserPreference.options() do
+    user
+    |> Teams.UserPreference.changeset(team, %{option => value})
+    |> Repo.insert!(
+      conflict_target: [:user_id, :team_id],
+      # This way of conflict handling enables doing upserts of options leaving
+      # existing, unrelated values intact.
+      on_conflict: from(p in Teams.UserPreference, update: [set: [{^option, ^value}]]),
```
zoldar: Huh, full query is needed for that? Doesn't `on_conflict: [set: [...]]` work the same in practice?

lib/plausible/teams/users.ex
```diff
@@ -7,6 +7,37 @@ defmodule Plausible.Teams.Users do
...
+
+    query =
+      from(
+        tup in Teams.UserPreference,
+        where: tup.user_id == ^user.id,
+        where: tup.team_id == ^team.id,
+        select: field(tup, ^option)
+      )
+
+    Repo.one(query) || Map.get(defaults, option)
```
zoldar:
```suggestion
    Repo.one(query) || Map.fetch!(defaults, option)
```

just in case? 😅

lib/plausible_web/live/sites.ex
```diff
@@ -69,31 +71,57 @@ defmodule PlausibleWeb.Live.Sites do
...
           {Teams.name(@current_team)}
-          <.unstyled_link
-            :if={Teams.setup?(@current_team)}
-            data-test-id="team-settings-link"
-            href={Routes.settings_path(@socket, :team_general)}
-          >
-            <Heroicons.cog_6_tooth class="hidden group-hover:inline size-5 dark:text-gray-100 text-gray-900" />
-          </.unstyled_link>
         </h2>
+        <.unstyled_link
```
zoldar, line 78: Maybe a good opportunity to use `icon_button` component, which was recently introduced? 😄 Though it might require introducing one more theme.

lib/plausible_web/live/sites.ex
zoldar: Moving conditional expressions to the start of `render` and assigning them to something like `show_consolidated_view_cta?` and `show_consolidated_view?` might help clean it up as well.

lib/plausible_web/live/sites.ex
zoldar: https://github.com/phoenixframework/phoenix_live_view/issues/3508 looks like a similar issue and from what I see it was tricky to address and I'm not event sure it got fully resolved?

zoldar (approved):
God's work 🙏  Left some feedback, most of it just nice-to-haves. Keeping preferences on membership instead of team/user combo makes more sense to me.

EDIT: Oh, and one more thing I've noted and then forgot about. Perhaps `ConslidatedView.reset_if_enabled` could be turned into an upsert? That would pair well with `disable` which is resilient to the view struct getting stale already.

lib/plausible/teams/user_preference.ex
aerosol: https://github.com/plausible/analytics/pull/5866/commits/5108d40842023a48e36aff405e8f2677e58a424e

extra/lib/plausible/consolidated_view.ex
aerosol: ba695c4d58

extra/lib/plausible/consolidated_view.ex
aerosol: 9a1dff7c02

lib/plausible/teams/users.ex
aerosol: 47e30e3d57

lib/plausible/teams/users.ex
aerosol: ah I've blindly copied that from `Site.UserPreference`. Not sure what was the intention there, but I assumed you knew what you were doing :) https://github.com/plausible/analytics/blob/consolidated-view-billing/lib/plausible/sites.ex#L113-L115

lib/plausible_web/live/sites.ex
aerosol: hrm @sanne-san do you think it's something we should do in this context?

lib/plausible/teams/users.ex
zoldar: Ah, hahaha, okay, let's leave it as is, though I can't remember why I did it as a query and not just a keyword dit :D

aerosol: "trust myself bro" moment

PR #5867, opened 2025-11-04
Limit consolidated site access to specific site routes

RobertJoonas:
### Changes

No visible changes in this PR, but just protecting unnecessary routes being accessed for consolidated views. A few examples:

* `GET /:domain/settings/people`
* `GET api/stats/:domain/funnels/:id`
* `POST /sites/:domain/memberships/invite`

The idea is to update the `AuthorizeSiteAccess` plug to respond with a 404 if the site in question is a consolidated view AND there's no `assign` explicitly allowing consolidated view access.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/plugs/authorize_public_api.ex
```diff
@@ -193,7 +193,7 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPI do
   defp maybe_verify_site_access(conn, api_key, feature) do
     case find_site(conn.params["site_id"]) do
       {:ok, site} ->
-        verify_site_access(api_key, site, feature)
+        verify_site_access(conn, api_key, site, feature)
```
aerosol: Maybe pass assigns only?

lib/plausible_web/plugs/authorize_site_access.ex
```diff
@@ -188,6 +189,14 @@ defmodule PlausibleWeb.Plugs.AuthorizeSiteAccess do
     end
   end
 
+  defp ensure_consolidated_view_access(conn, site) do
+    if Plausible.Sites.consolidated?(site) && !conn.assigns[:allow_consolidated_views] do
```
aerosol: Would it make sense to use private assigns? cc @zoldar WYT?

- https://hexdocs.pm/phoenix/Phoenix.Router.html#scope/2-options
- https://hexdocs.pm/plug/Plug.Conn.html#module-private-fields

lib/plausible_web/router.ex
```diff
@@ -268,38 +268,41 @@ defmodule PlausibleWeb.Router do
...
+        get "/:domain/suggestions/:filter_name", StatsController, :filter_suggestions
+
+        get "/:domain/suggestions/custom-prop-values/:prop_key",
+            StatsController,
+            :custom_prop_value_filter_suggestions
+      end
+    end
+
+    scope "/:domain/segments", PlausibleWeb.Api.Internal,
+      assigns: %{allow_consolidated_views: true} do
```
aerosol: have we tried working with segments on a consolidated view? I'm completely unfamiliar with the feature

lib/plausible_web/router.ex
```diff
@@ -317,7 +320,12 @@ defmodule PlausibleWeb.Router do
   end
 
   scope "/api/v2", PlausibleWeb.Api,
-    assigns: %{api_scope: "stats:read:*", api_context: :site, schema_type: :public} do
+    assigns: %{
+      api_scope: "stats:read:*",
+      api_context: :site,
+      schema_type: :public,
+      allow_consolidated_views: true
```
aerosol: We still don't list consolidated view in Sites API, do we?

lib/plausible_web/plugs/authorize_public_api.ex
zoldar: 👍 I'd go even further - pass `allow_consolidated_views?` directly to the function. As it's past 3 arguments, I'd also consider turning it into `opts` with each argument named, maybe?

lib/plausible_web/plugs/authorize_site_access.ex
zoldar: I have not hard preferences here. Though if that state is something that will not be used in controller actions or templates, `private` makes sense 👍

lib/plausible_web/router.ex
RobertJoonas: I just think of segments as groups of filters that you can save for your own user, or for the site. I can't think of anything related to segments that would cause conflict within consolidated views. I've tried playing around with them on prod, and everything works nicely as expected.

lib/plausible_web/router.ex
RobertJoonas: No, we don't.

lib/plausible_web/plugs/authorize_public_api.ex
RobertJoonas: 👌 Followed @zoldar's suggestion in https://github.com/plausible/analytics/pull/5867/commits/2a2ef613877fb1fc260d4a71a8fcb928d6d6425f

lib/plausible_web/plugs/authorize_site_access.ex
RobertJoonas: > Though if that state is something that will not be used in controller actions or templates, private makes sense

Yeah, it's only used for restricting access as soon as possible. Changed from `conn.assigns` to `conn.private` in
https://github.com/plausible/analytics/pull/5867/commits/0694662fc94fcda24e1f4b7964612406e0b30634

RobertJoonas:
Thanks for the review @zoldar and @aerosol! I've addressed your comments and also added a tests for both plugs.

PR #5869, opened 2025-11-04
Reduce noise in 2FA enforce notifications and update docs link

zoldar:
### Changes

Follow-up to #5855 

### Tests

### DIFF
--- a/test/plausible_web/controllers/settings_controller_test.exs
+++ b/test/plausible_web/controllers/settings_controller_test.exs
@@ -1766,13 +1766,15 @@ defmodule PlausibleWeb.SettingsControllerTest do
       # The email come in order in which they are sent.
       # As the logic sending them does not force any order,
       # we have to match them in order-independent way.
-      Enum.each(1..2, fn _ ->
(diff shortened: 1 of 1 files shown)


test/plausible_web/controllers/settings_controller_test.exs
```diff
@@ -1749,20 +1749,36 @@ defmodule PlausibleWeb.SettingsControllerTest do
...
+      # The email come in order in which they are sent.
+      # As the logic sending them does not force any order,
+      # we have to match them in order-independent way.
+      Enum.each(1..2, fn _ ->
+        assert assert_delivered_email_matches(%{
+                 subject: "Your team now requires 2FA",
+                 to: [{_, email}]
+               })
+
+        assert email in [member1.email, member2.email]
```
apata: nitpick, non-blocking: This assert doesn't fail if the two emails go to one of the members. Can we pop the used email from the list after the assert?

zoldar: https://github.com/plausible/analytics/pull/5869/commits/ad1fbcb9a5b15c21b9bbeb6631e16376a85de435

PR #5870, opened 2025-11-04
Fix shared link hostname

apata:
### Changes

Adds back the baseurl to shared links. It's needed for the links to be one-click shareable. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/sites_test.exs
+++ b/test/plausible/sites_test.exs
@@ -319,7 +319,7 @@ defmodule Plausible.SitesTest do
 
       Plausible.Teams.complete_setup(site1.team)
 
-      # excluded 
+      # excluded
       new_site(owner: user1, domain: "consolidated.example.com", consolidated: true)
 
       # guest site access
@@ -373,7 +373,7 @@ defmodule Plausible.SitesTest do
       site5 = new_site(domain: "team.example.com", owner: user4)
       add_member(site5.team, user: user1, role: :editor)
 
-      # excluded 
+      # excluded
       new_site(owner: user1, domain: "consolidated.example.com", consolidated: true)
 
       assert %{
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/shared_link_settings.ex
```diff
@@ -177,8 +177,4 @@ defmodule PlausibleWeb.Live.SharedLinkSettings do
 
     {:noreply, socket}
   end
-
-  defp shared_link_dest(site, link) do
-    Routes.stats_path(PlausibleWeb.Endpoint, :shared_link, site.domain, auth: link.slug)
```
apata, line 182: This has the benefit of ensuring proper encoding even for auth query parameter, which Plausible.Sites.shared_link_url/2 doesn't actually do. We could make use of this same logic in Plausible.Sites.shared_link_url/2, but it didn't make sense to me when I considered the relation between the modules Routes / Plausible.Sites.

test/plausible/sites_test.exs
```diff
@@ -1021,4 +1021,28 @@ defmodule Plausible.SitesTest do
...
+    end
+
+    test "doesn't share the same domain formatting with public dashboard links" do
+      site = new_site(domain: "a-café.fr")
+      link = insert(:shared_link, site: site)
+
+      assert "http://localhost:8000/a-café.fr" = PlausibleWeb.StatsView.pretty_stats_url(site)
+
+      assert "http://localhost:8000/share/a-caf%C3%A9.fr?" <> _q =
+               Sites.shared_link_url(site, link)
```
apata, line 1045: This codifies current non-ideal behavior that I noticed. It's worth thinking about how to unify the two and all our other route utils as well.

zoldar (approved):

PR #5874, opened 2025-11-05
Updated empty states across settings

sanne-san:
### Changes

- All empty states have been enhanced by center-aligned copy, link to docs, and CTA. 
- Whenever empty state is shown, the setting's subtitle is hidden.

### Tests

### Dark mode

### DIFF
--- a/test/plausible_web/live/sites_test.exs
+++ b/test/plausible_web/live/sites_test.exs
@@ -15,7 +15,7 @@ defmodule PlausibleWeb.Live.SitesTest do
 
       text = text(html)
 
-      assert text =~ "My Personal Sites"
+      assert text =~ "My personal sites"
       assert text =~ "Add your first personal site"
       refute text =~ "Go to team sites"
     end


lib/plausible_web/controllers/site_controller.ex
```diff
@@ -163,36 +163,75 @@ defmodule PlausibleWeb.SiteController do
...
 
+    has_shared_links? =
+      Repo.exists?(
+        from(l in Plausible.Site.SharedLink,
+          where:
+            l.site_id == ^site.id and
+              l.name not in ^Plausible.Sites.shared_link_special_names()
+        )
+      )
+
```
sanne-san: I added this so that I can hide the setting subtitle when the empty state is shown. I'm not sure this is the right place to put it, and it works but requires a page refresh for it to take effect.

extra/lib/plausible_web/live/funnel_settings/list.ex
```diff
@@ -12,11 +12,13 @@ defmodule PlausibleWeb.Live.FunnelSettings.List do
   def render(assigns) do
     ~H"""
     <div>
-      <.filter_bar filter_text={@filter_text} placeholder="Search Funnels">
-        <.button id="add-funnel-button" phx-click="add-funnel" mt?={false}>
-          Add funnel
-        </.button>
-      </.filter_bar>
+      <%= if String.trim(@filter_text) != "" || Enum.count(@funnels) > 0 do %>
```
RobertJoonas: I think the `String.trim(@filter_text) != ""` condition is redundant here as it's always true when `Enum.count(@funnels) > 0` is true. The search input isn't rendered without any funnels, so there cannot be any text in there either, right? Or is there a specific case that I'm missing here?

extra/lib/plausible_web/live/funnel_settings/list.ex
```diff
@@ -42,17 +44,33 @@ defmodule PlausibleWeb.Live.FunnelSettings.List do
           </:tbody>
         </.table>
       <% else %>
-        <p class="mt-12 mb-8 text-sm text-center">
-          <span :if={String.trim(@filter_text) != ""}>
+        <%= if String.trim(@filter_text) != "" do %>
```
RobertJoonas: This render function is getting a bit hard to read with the nested if/else conditionals. How about we extract some functions? I'm thinking:

```elixir
def render(assigns) do
  assigns = assign(assigns, :searching?, String.trim(assigns.filter_text) != "")

  ~H"""
  <%= if Enum.count(@funnels) > 0 do %>
    # ... keep the filter bar, add button, table
  <% else %>
    <.no_search_results :if={@searching?}/>
    <.create_first_funnel :if={not @searching?}
  <% end %>
  """
end

defp no_search_results(assigns) do
  ~H"""
  <p class="mt-12 mb-8 text-sm text-center">
    No funnels found for this site. Please refine or
    <.styled_link phx-click="reset-filter-text" id="reset-filter-hint">
       reset your search.
    </.styled_link>
  </p>
  """
end
```

extra/lib/plausible_web/live/funnel_settings.ex
```diff
@@ -70,15 +70,20 @@ defmodule PlausibleWeb.Live.FunnelSettings do
...
+        class="flex flex-col items-center justify-center pt-5 pb-6 max-w-md mx-auto"
+      >
+        <h3 class="text-center text-base font-medium text-gray-900 dark:text-gray-100 leading-7">
           Ready to dig into user flows?
-        </h1>
-        <p class="mt-4 mb-6 max-w-lg text-sm text-gray-500 dark:text-gray-400 leading-5 text-center">
-          Set up a few goals first (e.g. <b>"Signup"</b>, <b>"Visit /"</b>, or <b>"Scroll 50% on /blog/*"</b>) and return here to build your first funnel!
+        </h3>
+        <p class="text-center text-sm mt-1 text-gray-500 dark:text-gray-400 leading-5 text-pretty">
+          Set up a few goals like <span class="font-medium text-indigo-600 dark:text-gray-100">Signup</span>, <span class="font-medium text-indigo-600 dark:text-gray-100">Visit /</span>, or
```
RobertJoonas: Could we extract a small inline text component (perhaps `<.highlighted>` in `PlausibleWeb.Components.Generic`?) to not have to repeat these classnames for every element? Or if it's something that will definitely not be reused, it could also be a private function in this file. WDYT?

extra/lib/plausible_web/live/funnel_settings/list.ex
aerosol: `or` is perfectly valid here since both left and right hand side expressions return a boolean

lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -16,17 +16,19 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
...
-          id="add-goal-button"
-          phx-click="add-goal"
-          mt?={false}
-          x-data
-          x-on:click={Modal.JS.preopen("goals-form-modal")}
-        >
-          Add goal
-        </.button>
-      </.filter_bar>
+      <%= if String.trim(@filter_text) != "" || Enum.count(@goals) > 0 do %>
```
RobertJoonas: Ditto: `String.trim(@filter_text) != "" ||` irrelevant?

lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -89,17 +91,35 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
           </:tbody>
         </.table>
       <% else %>
-        <p class="mt-12 mb-8 text-center text-sm">
-          <span :if={String.trim(@filter_text) != ""}>
+        <%= if String.trim(@filter_text) != "" do %>
```
RobertJoonas: Maybe follow the same logic here too (as I suggested for funnels), extracting the components?

lib/plausible_web/live/props_settings/list.ex
```diff
@@ -11,11 +11,14 @@ defmodule PlausibleWeb.Live.PropsSettings.List do
   def render(assigns) do
     ~H"""
     <div>
-      <.filter_bar filter_text={@filter_text} placeholder="Search Properties">
-        <.button phx-click="add-prop" mt?={false}>
-          Add property
-        </.button>
-      </.filter_bar>
+      <%= if String.trim(@filter_text) != "" || (is_list(@props) && length(@props) > 0) do %>
```
RobertJoonas: Ditto: `String.trim(@filter_text) != "" ||` irrelevant?

lib/plausible_web/live/props_settings/list.ex
```diff
@@ -32,17 +35,33 @@ defmodule PlausibleWeb.Live.PropsSettings.List do
           </:tbody>
         </.table>
       <% else %>
-        <p class="mt-12 mb-8 text-center text-sm">
-          <span :if={String.trim(@filter_text) != ""}>
+        <%= if String.trim(@filter_text) != "" do %>
```
RobertJoonas: Ditto: I think extracting functions (like in goals and funnels) could make the render function more readable.

lib/plausible_web/controllers/site_controller.ex
aerosol: Hey @sanne-san - this is unfortunately, not the most lucky problem to have. Looks like we have a lot to unpack with regards to how we've been randomly migrating settings sections to live views. 

A smol background story: settings were originally all static, standard HTML forms etc. With need of having many dynamic forms over the time (and us getting a safe-ish playground to learn LiveView the hard way), we've started turning them into live views. Because rewriting everything was risky we've started embedding live views (modules operating on `@socket` / websocket) in static templates (as in "rendered once", a.k.a "dead" - using `@conn` / HTTP) bit by bit, where needed. Live views are where react-like diffs are applied. That's why you're struggling with refresh - the part that renders the tile and its slots is static :skull:.

Normally it'd best to move the `<.tile></.tile>` element to the live view that's embedded underneath - `PlausibleWeb.Live.SharedLinkSettings` in this case. But there are cans of worms ahead: you'll get the compiler complain about `site_role` missing. You can of course provide it in the live view's `mount()` function (either execute the query or pass it down [via the session params](https://github.com/plausible/analytics/blob/sanne-empty-states/lib/plausible_web/templates/site/settings_visibility.html.heex#L47) in the static template and assign it to the socket). 

But it also turns out that the `tile` component is a bit complicated - in an attempt of generic handling of toggling features and/or upgrade CTAs in case of insufficient subscriptions, it is why it wants the `site_role` assign provided. It also needs `@conn` - which we don't have access to in live views.. And `SharedLinks` is indeed a feature that's not available to everyone, so we need some handling of billing availability still.

With Goals, Funnels, CustomProps it might get even weirder, since we have live views embedding live views there and they're separate processes so they'd need to exchange messages with each other :see_no_evil: 

I wouldn't recommend getting there now; sounds like a massive undertaking to hide the subtitle. :grimacing:

aerosol: Maybe it's possible to remove the `:subtitle` slot from static template and render its markup in live view, verbatim? Just so we don't have to deal with all the `tile` problems.

aerosol: @sanne-san proposal to solve this: https://github.com/plausible/analytics/pull/5891. Should also work with goals and funnels now. I haven't looked into any existing test failures though. cc @RobertJoonas

extra/lib/plausible_web/live/funnel_settings/list.ex
sanne-san: I think it's needed when filtering returns no matches: when `@funnels` becomes empty, but `filter_text` still has the search text, right?

extra/lib/plausible_web/live/funnel_settings.ex
sanne-san: Yeah, good idea!

extra/lib/plausible_web/live/funnel_settings/list.ex
sanne-san: Yeah this is much better, thanks 👌

lib/plausible_web/components/site/feature.ex
```diff
@@ -53,7 +53,7 @@ defmodule PlausibleWeb.Components.Site.Feature do
     """
   end
 
-  defp target(site, setting, conn, set_to) when is_boolean(set_to) do
+  def target(site, setting, conn, set_to) when is_boolean(set_to) do
```
sanne-san: Needed to change this back to being public in order for the tests to be able to call it.

lib/plausible_web/components/site/toggle_live.ex
```diff
@@ -78,6 +78,8 @@ defmodule PlausibleWeb.Components.Site.Feature.ToggleLive do
             "#{feature_mod.display_name()} are now hidden from your dashboard"
           end
 
+        send(self(), {:site_updated, updated_site})
```
sanne-san: This was needed to make the content appear/disappear instantly upon toggling the feature toggle

lib/plausible_web/live/props_settings/list.ex
```diff
@@ -9,13 +9,18 @@ defmodule PlausibleWeb.Live.PropsSettings.List do
...
-        </.button>
-      </.filter_bar>
+      <%= if @searching? or (is_list(@props) && length(@props) > 0) do %>
+        <.filter_bar filter_text={@filter_text} placeholder="Search Properties">
+          <.button phx-click="add-prop" mt?={false}>
+            Add property
+          </.button>
+        </.filter_bar>
+      <% end %>
+
```
aerosol, line 23:
```suggestion
      <.filter_bar :if={@searching? or length(@props) > 0} filter_text={@filter_text} placeholder="Search Properties">
        <.button phx-click="add-prop" mt?={false}>
          Add property
        </.button>
      </.filter_bar>  
```

when can `props` be anything else than a list here? 🤔

lib/plausible_web/live/sites.ex
```diff
@@ -88,7 +88,14 @@ defmodule PlausibleWeb.Live.Sites do
       <PlausibleWeb.Team.Notice.team_invitations team_invitations={@team_invitations} />
 
       <div class="relative z-10 pt-4 flex flex-col sm:flex-row sm:items-center sm:justify-between gap-y-2">
-        <.search_form :if={@has_sites?} filter_text={@filter_text} uri={@uri} />
+        <.search_form
+          :if={
+            @has_sites? and
+              (Teams.setup?(@current_team) or @sites.entries != [] or @filter_text != "")
```
aerosol: Why is `Teams.setup?(@current_team)` here? 

If a team is setup and all its sites have been deleted, then perhaps it's pointless to show the search bar? The +Add button becomes broken then though.

With several new conditions here and there, we might want to test those cases to guard it from regression - removing that `:if` breaks zero tests.

test/plausible_web/live/funnel_settings_test.exs
```diff
@@ -33,12 +33,12 @@ defmodule PlausibleWeb.Live.FunnelSettingsTest do
...
         assert resp =~ "From signup to blog"
         refute resp =~ "Your account does not have access"
-        refute resp =~ "upgrade your subscription"
+        refute resp =~ "please upgrade your subscription"
         assert element_exists?(resp, "a[href=\"https://plausible.io/docs/funnel-analysis\"]")
       end
 
       test "search funnels input is rendered", %{conn: conn, site: site} do
-        setup_goals(site)
+        {:ok, _} = setup_funnels(site)
```
aerosol, line 41: 👌 ✨

lib/plausible_web/live/props_settings/list.ex
aerosol: The suggestion above displays in a weird way, apologies, probably not a great idea to commit it directly :sweat:

lib/plausible_web/live/sites.ex
sanne-san: The reason is that I improved the personal sites empty state, so that it's hidden there when personal sites is empty and team sites isn't. 
CleanShot 2025-11-19 at 09 36 03@2x

But I realise now that I should also fix up the empty state for team sites, and perhaps that can simplify the whole thing a bit (although we still want a different call to action on the personal sites empty state).

extra/lib/plausible_web/live/funnel_settings/list.ex
RobertJoonas: You're absolutely right! My bad 🙈

lib/plausible_web/live/props_settings/list.ex
RobertJoonas: Agreed with Adam, even though the condition existed already before this PR, it's safe to remove. I think this is some kind of leftover from when `site.allowed_event_props` (which can be `nil`, meaning that all props are allowed to display) was passed into this component directly. But I just checked that that's not the case anymore and `@props` gets the value of `site.allowed_event_props || []`.

sanne-san: Ok cool, yeah I kept it as I wasn't sure if I was missing any context on that. Updated it to be the same as goals and funnels.

lib/plausible_web/components/site/feature.ex
aerosol: it's gone now

lib/plausible_web/components/generic.ex
```diff
@@ -502,21 +501,24 @@ defmodule PlausibleWeb.Components.Generic do
...
           site={@site}
-          conn={@conn}
+          feature_mod={@feature_mod}
           current_user={@current_user}
         />
       </header>
-      <div :if={@show_content?} class="border-b dark:border-gray-700 mx-6"></div>
-      <div :if={@show_content?} class="relative">
+      <div class={["border-b dark:border-gray-700 mx-6", if(not @show_content?, do: "hidden")]}></div>
+      <div class={["relative", if(not @show_content?, do: "hidden")]}>
```
RobertJoonas, line 517: Nice! 👌

lib/plausible_web/components/generic.ex
```diff
@@ -502,21 +501,24 @@ defmodule PlausibleWeb.Components.Generic do
         <div :if={@subtitle != []} class="text-sm mt-px text-gray-500 dark:text-gray-400 leading-5">
           {render_slot(@subtitle)}
         </div>
-        <PlausibleWeb.Components.Site.Feature.toggle
+
+        <.live_component
           :if={@feature_toggle?}
-          feature_mod={@feature_mod}
+          module={PlausibleWeb.Components.Site.Feature.ToggleLive}
+          id={"feature-toggle-#{@site.id}-#{@feature_mod}"}
```
RobertJoonas, line 510: What's this `id` used for? I'm seeing another id attribute defined by the `ToggleLive` component itself (the wrapper div):

```
<div class="mt-4" id={"feature-#{@feature_mod.name()}-toggle"}>
```

Should we stick to a single one?

sanne-san:
> Noticed this small UI thing btw (not related to this PR but should be a very easy fix):
> 
> image

Thank you!  I'll look into this little bug separately 🙏

lib/plausible_web/components/generic.ex
aerosol: Tests. I guess we prefer data-test-id?

PR #5877, opened 2025-11-11
Migration: Add user preferences per team table

aerosol:
Migration extracted from #5866 5866

Superseded by https://github.com/plausible/analytics/pull/5881


aerosol:
Holding up. @zoldar (when you're back) do you think per team membership would be wiser to do?

aerosol:
I have a patch ready against https://github.com/plausible/analytics/pull/5866 with "team membership preferences", just in case

PR #5886, opened 2025-11-14
Update goal settings design

sanne-san:
### Changes

- Replace the `Add goal` button in goal settings with a dropdown button to directly select the goal type. This way, a modal opens with the correct form for the selected goal type. The tabs in the modal have been removed.
- Add a new `pill` component to show the goal type in the table in a more distinct way. The `settings_badge` component is replaced with the `pill` component. The `pill` component that was used in `plan_box.ex` is renamed to `highlight_pill`.
- Replaced `Belongs to funnel` text with a funnel icon in the goal settings list.
- Some small tweaks like increasing the search bar width, the padding of the table cells, and adding a header to the goal settings list.

### Tests

### Dark mode

### DIFF
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -522,7 +522,7 @@ defmodule PlausibleWeb.Components.Generic do
             current_team={@current_team}
             site={@site}
           >
-            <div class="p-6 pb-14">
+            <div class="p-6">
               {render_slot(@inner_block)}
             </div>
           </PlausibleWeb.Components.Billing.feature_gate>


sanne-san:
@aerosol I used the Prima dropdown component that you used on `/sites` too. Seeing a weird quirk and I wonder if this is related to the same issue we've been having with page refresh in the [empty states work](https://github.com/plausible/analytics/pull/5874), or it's a problem with the component: after adding a goal, the dropdown button doesn't respond unless I do a page refresh – see video. 

https://github.com/user-attachments/assets/1749ec07-c443-4add-b7e2-7697fda7c064

aerosol:
> @aerosol I used the Prima dropdown component that you used on `/sites` too. Seeing a weird quirk and I wonder if this is related to the same issue we've been having with page refresh in the [empty states work](https://github.com/plausible/analytics/pull/5874), or it's a problem with the component: after adding a goal, the dropdown button doesn't respond unless I do a page refresh – see video.
> CleanShot.2025-11-14.at.13.02.12.mp4

I'm not sure what's at fault yet, but the prima button element is losing its event listener on successful goal submission. Could be related either to pre-existing modal complexity or some hooks lifecycle, e.g. prima dropdown missing `updated()` hook perhaps? (https://hexdocs.pm/phoenix_live_view/js-interop.html#client-hooks-via-phx-hook)

aerosol:
https://github.com/plausible/prima/pull/15

ukutaht:
@aerosol agree with your analysis and solution. I've merged https://github.com/plausible/prima/pull/17 and published a new version of prima `0.1.9` on hex.

lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -15,23 +16,50 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
...
+            >
+              <Heroicons.plus class={PrimaDropdown.dropdown_item_icon_class()} /> Custom event
+            </PrimaDropdown.dropdown_item>
+            <PrimaDropdown.dropdown_item
+              phx-click="add-goal"
+              phx-value-goal-type="scroll"
+              x-data
+              x-on:click={Modal.JS.preopen("goals-form-modal")}
+            >
+              <Heroicons.plus class={PrimaDropdown.dropdown_item_icon_class()} /> Scroll depth
```
ukutaht: The `dropdown_item_icon_class` is new to me. I understand it was introduced in a previous PR but some feedback on it: it includes classes with `group-hover:` modifier which should not be necessary. The `data-focus` attribute from prima library is applied on hover so we should not have to duplicate `group-hover:{class}` and `group-data-focus:{class}` - they are equivalent.

The `data-focus` attribute is also added when an item is selected via keyboard shortcut (arrow up/down) and should be the single source of truth for whether a dropdown item should be styled as highlighted.

sanne-san: Thanks for the feedback, I'll update that! Reason I added the `dropdown_item_icon_class` is to avoid it being styled slightly differently everywhere an icon is added to a dropdown, but if there is a better way lmk!

ukutaht: Yeah it makes sense, I was just trying to say that the hover classes should be unnecessary which isn't a big deal

lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -160,15 +160,38 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
...
-        id="add-goal-button"
-        phx-click="add-goal"
-        x-data
-        x-on:click={Modal.JS.preopen("goals-form-modal")}
-        class="mt-4"
-      >
-        Add goal
-      </.button>
+      <PrimaDropdown.dropdown id="add-goal-dropdown-empty" class="mt-4">
+        <PrimaDropdown.dropdown_trigger as={&button/1} mt?={false}>
```
aerosol, line 164: This function has no `mt?` attribute, we can dispose of it

PR #5891, opened 2025-11-18
Patch to Sanne's empty states: Make `tile` component lv-embeddable

aerosol:
PR against https://github.com/plausible/analytics/pull/5874 

This makes `tile` component play well in LV context where more dynamic approach is necessary. @RobertJoonas I've tried various things (including making a "live tile" - which seemed like an overkill) and this is the smallest friction I could come up with.

### DIFF
--- a/lib/plausible_web/components/site/feature.ex
+++ b/lib/plausible_web/components/site/feature.ex
@@ -7,11 +7,12 @@ defmodule PlausibleWeb.Components.Site.Feature do
 
   attr(:site, Plausible.Site, required: true)
   attr(:feature_mod, :atom, required: true, values: Plausible.Billing.Feature.list())
-  attr(:conn, Plug.Conn, required: true)
+  attr(:conn, :any, default: nil)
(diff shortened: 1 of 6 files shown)


lib/plausible_web/components/site/feature.ex
```diff
@@ -7,11 +7,12 @@ defmodule PlausibleWeb.Components.Site.Feature do
...
   attr(:site, Plausible.Site, required: true)
   attr(:feature_mod, :atom, required: true, values: Plausible.Billing.Feature.list())
-  attr(:conn, Plug.Conn, required: true)
+  attr(:conn, :any, default: nil)
+  attr(:current_user, :any, default: nil)
   attr(:class, :any, default: nil)
   slot(:inner_block)
 
-  def toggle(assigns) do
+  def toggle(%{conn: %Plug.Conn{}} = assigns) do
```
RobertJoonas, line 15: Nice!💡

PR #5893, opened 2025-11-19
Refactor building the Query struct

RobertJoonas:
### Changes

_This PR is a refactor and doesn't introduce any new behaviour._

In the upcoming dashboard migration from React to LiveView, we want to be able to create `Query` structs from native data structures. E.g.:

* pass the site struct itself into the `Query.build` function, not the `site_id`
* pass `DateTime` structs instead of iso8601 strings
* pass metrics as atoms instead of strings
* etc...

The biggest refactor in this PR is changing the current `Query.build` function into `Query.parse_and_build` which calls two separate modules (`QueryParser` and `QueryBuilder` (new) for the respective actions).

It also introduces an intermediate data structure that represents the parsed state between those two phases - `ParsedQueryParams`.

There was a lot of query "*building*" logic and validations incorporated into `QueryParser`. The test file (with more than 3k lines of code) got turned into `QueryParseAndBuildTest`, now asserting on a Query struct outcome rather than an arbitrary map.

This PR doesn't tackle the `date_range` input aspect yet. For now, `QueryBuilder` will expect a clean `utc_time_range` input (via the `ParsedQueryParams` struct). It shouldn't block this PR though. I can tackle it later. E.g.: we'll likely want to build queries like:

```elixir
query1 = QueryBuilder.build(site, %{metrics: [:visitors], date_range: :month, date: ~D[2025-01-01]})
query2 = QueryBuilder.build(site, %{metrics: [:visitors], date_range: [~D[2025-01-01], ~D[2025-01-31]]})
```

With that, we can also migrate any existing ad-hoc `Query` construction (e.g. in email reports or `query_24h_stats`) to use this new version of `Query.build`.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/stats/query/query_from_test.exs
+++ b/test/plausible/stats/query/query_from_test.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.Stats.QueryTest do
+defmodule Plausible.Stats.Query.QueryFromTest do
   use Plausible.DataCase, async: true
   use Plausible.Teams.Test
   alias Plausible.Stats.Query
--- a/test/plausible/stats/query/query_optimizer_test.exs
+++ b/test/plausible/stats/query/query_optimizer_test.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.Stats.QueryOptimizerTest do
+defmodule Plausible.Stats.Query.QueryOptimizerTest do
   use Plausible.DataCase, async: true
 
   alias Plausible.Stats.{Query, QueryOptimizer, DateTimeRange, ParsedQueryParams}
--- a/test/plausible/stats/query/query_parse_and_build_test.exs
+++ b/test/plausible/stats/query/query_parse_and_build_test.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.Stats.QueryParseAndBuildTest do
+defmodule Plausible.Stats.Query.QueryParseAndBuildTest do
   use Plausible.DataCase
   use Plausible.Teams.Test
   import Plausible.AssertMatches
--- a/test/plausible/stats/query/query_parser_test.exs
+++ b/test/plausible/stats/query/query_parser_test.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.Stats.QueryParserTest do
+defmodule Plausible.Stats.Query.QueryParserTest do
   use Plausible.DataCase
   import Plausible.Stats.QueryParser
 
--- a/test/plausible/stats/query/query_result_test.exs
+++ b/test/plausible/stats/query/query_result_test.exs
@@ -1,4 +1,4 @@
-defmodule Plausible.Stats.QueryResultTest do
+defmodule Plausible.Stats.Query.QueryResultTest do
   use Plausible.DataCase, async: true
   use Plausible.Teams.Test
   alias Plausible.Stats.{Query, QueryRunner, QueryResult, QueryOptimizer}


lib/plausible/stats/query_builder.ex
```diff
@@ -0,0 +1,363 @@
...
+  on_ee do
+    alias Plausible.Stats.Goal.Revenue
+
+    def preload_revenue(site, preloaded_goals, metrics, dimensions) do
+      Revenue.preload(site, preloaded_goals, metrics, dimensions)
+    end
+
+    defp validate_revenue_metrics_access(site, query) do
+      if Revenue.requested?(query.metrics) and not Revenue.available?(site) do
+        {:error, "The owner of this site does not have access to the revenue metrics feature."}
```
zoldar, line 124: For future consideration: make the validation errors more friendly for internal consumption (well formed tuples with atoms/values) and build the human-formatted versions on top of it in the relevant contexts?

lib/plausible/stats/query_builder.ex
```diff
@@ -0,0 +1,363 @@
...
+         :ok <- validate_metrics(query),
+         :ok <- validate_include(query) do
+      query =
+        query
+        |> set_time_on_page_data(site)
+        |> put_comparison_utc_time_range()
+        |> Query.put_imported_opts(site)
+
+      on_ee do
+        # NOTE: The Query API schema does not allow the sample_threshold param
```
zoldar, line 29: That seems to be only used in a couple tests against `Query.from`, but yeah, not here.

lib/plausible/stats/query_parser.ex
```diff
@@ -110,7 +57,7 @@ defmodule Plausible.Stats.Filters.QueryParser do
     parse_list(filters, &parse_filter/1)
   end
 
-  def parse_filters(_invalid_metrics), do: {:error, "Invalid filters passed."}
+  def parse_filters(nil), do: {:ok, nil}
```
zoldar, line 60: Hm, I know we went over it but, do we want to let it pass silently? While APIv2 schema might ensure this is a list, passing invalid value internally getting silently ignored might cause confusion. Maybe we should even let it crash then?

RobertJoonas: Hmm, actually, that's another thing I wanted to touch base with you on...

> passing invalid value internally getting silently ignored might cause confusion.

The change we're currently looking at is in `QueryParser`, which will not be used for internal query building. Now, while QueryParser "parses" the filters, it's also validating the format. Kind of tricky to separate parsing and validating filters but yeah, I agree that even the internal `QueryBuilder.build` should return `{:error, :invalid_filters}` when you try to construct a query with `filters: [%{"event:page" => "/"}, %{"visit:source" => "Google"}]` (which is obviously not the correct representation of filters).

zoldar: Ah, yeah, indeed, filter parsing is still part of builder. Okay, I guess let's leave it be for now and try to address it in the future follow-ups?

RobertJoonas:
> Just for the record - one potentially necessary thing the we have touched on when going over this PR though it's outside of scope - we'll also need a way to turn a Query back into query string.

Hmm I think there are different "query objects" we need to consider. The one that gets serialized into a URL query string will be what's currently defined in React:

```javascript
export const queryDefaultValue: DashboardQuery = {
  period: '28d' as QueryPeriod,
  comparison: null,
  match_day_of_week: true,
  date: null,
  from: null,
  to: null,
  compare_from: null,
  compare_to: null,
  filters: [],
  resolvedFilters: [],
  labels: {},
  with_imported: true
}
```

I think the `%Query{}` struct as we know it will not need to be serializable.

However, we'll still need to be able to construct a `%Query{}` from the `DashboardQuery` params (+ some extra report-dependent params e.g. `dimensions`). Which is pretty much what `Query.from{}` is currently doing... I think it makes sense to get to a state where:

* There are two query parsers - DashboardQueryParser & StatsAPIQueryParser
* Both these QueryParsers spit out `%ParsedQueryParams{}`
* From this point, building and validating the query will be the same for API queries and internal queries

PR #5910, opened 2025-11-25
Use prima modal in ip_rules settings

ukutaht:
### Changes

Refactors the IP rule settings to use Prima modal over alpine modal.

Some yak shaving:
I had to add facilities to Prima to be able to close a modal from the backend by pushing an event. I decided to change the namespace for the existing open/close API as well. Since we are the only users of Prima I think it's ok. The breaking change is from `0.1.9 -> 0.2.0`.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/components/prima_modal.ex
+++ b/lib/plausible_web/live/components/prima_modal.ex
@@ -4,11 +4,12 @@ defmodule PlausibleWeb.Live.Components.PrimaModal do
   alias Prima.Modal
 
   attr :id, :string, required: true
+  attr :use_portal?, :boolean, default: Mix.env() not in [:test, :ce_test]
   slot :inner_block, required: true
 
   def modal(assigns) do
     ~H"""
-    <Modal.modal portal={Mix.env() not in [:test, :ce_test]} id={@id}>
+    <Modal.modal portal={@use_portal?} id={@id}>
       <Modal.modal_overlay
         transition_enter={{"ease-out duration-300", "opacity-0", "opacity-100"}}
         transition_leave={{"ease-in duration-200", "opacity-100", "opacity-0"}}


lib/plausible_web/live/components/prima_modal.ex
```diff
@@ -8,7 +8,7 @@ defmodule PlausibleWeb.Live.Components.PrimaModal do
 
   def modal(assigns) do
     ~H"""
-    <Modal.modal id={@id}>
+    <Modal.modal portal={Mix.env() !== :test} id={@id}>
```
ukutaht: Portals do not work well with LazyHTML: https://github.com/phoenixframework/phoenix_live_view/issues/3948

Solution: render modals without portals in test env

lib/plausible_web/live/shields/ip_rules.ex
```diff
@@ -139,50 +136,63 @@ defmodule PlausibleWeb.Live.Shields.IPRules do
...
-              class="max-w-md w-full mx-auto"
-            >
-              <.title>Add IP to block list</.title>
-
-              <div class="mt-4">
-                <p
-                  :if={not ip_rule_present?(@ip_rules, @remote_ip)}
-                  class="text-sm text-gray-500 dark:text-gray-400 mb-4"
+          <PrimaModal.modal id="ip-rule-form-modal">
+            <div class="p-5 sm:p-6 max-w-md">
```
zoldar, line 140: Given the close button will be reused in subsequent implementations, maybe it's worth it to create a custom Prima modal wrapper tailored for these use cases? Or do you want to defer that until the next implementation (which is fine!)?

lib/plausible_web/live/components/prima_modal.ex
```diff
@@ -8,7 +8,7 @@ defmodule PlausibleWeb.Live.Components.PrimaModal do
 
   def modal(assigns) do
     ~H"""
-    <Modal.modal id={@id}>
+    <Modal.modal portal={Mix.env() not in [:test, :ce_test]} id={@id}>
```
zoldar: `Mix.env` is not available in releases at runtime and this causes a crash in staging: https://sentry.plausible.io/organizations/plausible/issues/1556/?environment=staging&project=2&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=0&utc=true

Probably best to move that to compile time with something like:

```elixir

@portal_modal? Mix.env() in [:test, :ce_test]

def modal(assigns) do
  ~H"""
  <Modal.modal portal={@portal_modal?} id={@id}>
```

maybe?

aerosol: `attr` is probably better here:

```elixir
  attr(:portal_modal?, :boolean, default: Mix.env() not in [:test, :ce_test])

```

lib/plausible_web/live/shields/ip_rules.ex
```diff
@@ -139,50 +136,63 @@ defmodule PlausibleWeb.Live.Shields.IPRules do
...
+                      to block your own traffic, or enter a custom address below.
+                    </p>
 
-              <.input
-                field={f[:description]}
-                label="Description (optional)"
-                placeholder="e.g. The Office"
-              />
+                    <.input
+                      data-autofocus
```
aerosol, line 173: what is the reason behind that change?

lib/plausible_web/live/components/prima_modal.ex
ukutaht: https://github.com/plausible/analytics/pull/5910/commits/52ce8457bc50da4b17a10324d1d1df174136748d

lib/plausible_web/live/shields/ip_rules.ex
ukutaht: When opening modals with embedded forms, it seems to be good UX practice to automatically focus the first input.

`data-autofocus` is a prima feature - the element will receive focus automatically after modal is opened: https://github.com/plausible/prima/blob/7471f7987b0b276cb7796a8baa29629fcc59cef1/lib/prima/modal.ex#L28-L37

lib/plausible_web/live/shields/ip_rules.ex
ukutaht: Yes, there's definitely some styling to DRY out with prima modals. My thinking was to leave it for a future PR.

PR #5916, opened 2025-11-27
Limit preloading goals

aerosol:
https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/9318933498#__recording_9325003591


aerosol:
Part of #5917

PR #5917, opened 2025-11-27
Enforce max limit for goals per site

aerosol:
Ref https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/9318933498

### DIFF
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -128,8 +128,8 @@ defmodule Plausible.Goals do
 
   def find_or_create(_, %{"goal_type" => "event"}, _), do: {:missing, "event_name"}
 
-  def find_or_create(site, %{"goal_type" => "page", "page_path" => _} = params, _) do
-    create(site, params, do_upsert())
+  def find_or_create(site, %{"goal_type" => "page", "page_path" => _} = params, opts) do
+    create(site, params, do_upsert(opts))
   end
 
   def find_or_create(_, %{"goal_type" => "page"}, _), do: {:missing, "page_path"}


lib/plausible/goals/goals.ex
```diff
@@ -105,23 +121,25 @@ defmodule Plausible.Goals do
...
+    create(site, params, do_upsert(opts))
   end
 
-  def find_or_create(_, %{"goal_type" => "event"}), do: {:missing, "event_name"}
+  def find_or_create(_, %{"goal_type" => "event"}, _), do: {:missing, "event_name"}
 
-  def find_or_create(site, %{"goal_type" => "page", "page_path" => _} = params) do
-    create(site, params, upsert?: true)
+  def find_or_create(site, %{"goal_type" => "page", "page_path" => _} = params, _) do
+    create(site, params, do_upsert())
```
zoldar: Just to double-check - is skipping `opts` here intentional?

lib/plausible/goals/goals.ex
```diff
@@ -346,16 +373,31 @@ defmodule Plausible.Goals do
...
-    if Ecto.Changeset.get_field(changeset, :currency) do
+    if Changeset.get_field(changeset, :currency) do
       site = Plausible.Repo.preload(site, :team)
       Plausible.Billing.Feature.RevenueGoals.check_availability(site.team)
     else
       :ok
     end
   end
 
+  defp check_goals_limit(site, changeset, opts) do
```
zoldar, line 384: An alternative approach could be wrapping goal insertion in a transaction, doing the `count` check afterwards (provided insert actually inserts and does not do an upsert) and rolling it back if count turns out above the limit. It would spare the extra goal_exists check. But it's fine as is though too.

lib/plausible/goals/goals.ex
aerosol: yep, nothing ever is able to provide them

lib/plausible/goals/goals.ex
aerosol: > An alternative approach could be wrapping goal insertion in a transaction

it's wrapped in transaction, except for bulk "find or create"

> (provided insert actually inserts and does not do an upsert)

we do both and need to check limit for both

> It would spare the extra goal_exists check. 

the goal_exists check is precisely for when upsert is requested and performed

aerosol:
@ukutaht could you have a look please?

lib/plausible/goals/goals.ex
aerosol: oh wait

aerosol: good call, thx

ukutaht (approved):

PR #5922, opened 2025-12-01
Upgrade dependencies

aerosol:
Upgrade all the dependencies that seem relatively safe


mix.exs
```diff
@@ -97,33 +97,33 @@ defmodule Plausible.MixProject do
       {:mox, "~> 1.0", only: [:test, :ce_test]},
       {:nanoid, "~> 2.1.0"},
       {:nimble_totp, "~> 1.0"},
-      {:oban, "~> 2.19.1"},
+      {:oban, "~> 2.20.1"},
```
aerosol, line 100: there's supposed to be migration for new discarded states indexing, trying to locate any details

aerosol: migration adds indexes only, so due to our CI limitations, we'll run it after this is merged

aerosol: https://hexdocs.pm/oban/Oban.Migration.html#module-usage

PR #5923, opened 2025-12-01
Add report percentages to dashboard and details view

sanne-san:
### Changes

- Add percentage values to the reports that are currently lacking this.
- Change the way percentage is displayed to show only on hover (and tap on mobile). On the dashboard, all report percentages show when hovering the report. In the detailed breakdown, hovering on a row shows the percentage.
- Remove tooltip for all values except the ones that are abbreviated. The tooltip is still shown for all metrics in comparison mode.
- In the detailed breakdown, abbreviated numbers are displayed in full when hovering its corresponding table row.
- Render tooltip via portal as it doesn't play nice with the overflow:hidden class needed to make the sliding hover effect work.
- Update tooltip style, including removing the arrow for a more modern feel.
- Remove the logic that hides the `Details` button when all items are visible on the dashboard, so the detailed breakdown can be viewed at all times.
- Improved detailed breakdown modal's responsiveness; the inner table is now scrollable, rather than the entire modal.

https://github.com/user-attachments/assets/292004e6-6ffe-439f-9c78-5e08fc2ffd4c

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/pages_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/pages_test.exs
@@ -967,7 +967,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
       populate_stats(
         site,
         site_import.id,
-        for(_ <- 1..25, do: build(:imported_visitors, date: ~D[2020-01-01]))
+        for(_ <- 1..24, do: build(:imported_visitors, date: ~D[2020-01-01]))
       )
 
       conn =
@@ -984,7 +984,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
                  "bounce_rate" => 0,
                  "time_on_page" => 48,
                  "scroll_depth" => 50,
-                 "percentage" => 19.23
+                 "percentage" => 20.0
                },
                %{
                  "name" => "/native-only",
@@ -993,7 +993,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
                  "bounce_rate" => 0,
                  "time_on_page" => 60,
                  "scroll_depth" => 40,
-                 "percentage" => 3.85
+                 "percentage" => 4.0
                },
                %{
                  "name" => "/imported-only",
@@ -1002,7 +1002,7 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
                  "bounce_rate" => 0,
                  "time_on_page" => 30,
                  "scroll_depth" => 10,
-                 "percentage" => 76.92
+                 "percentage" => 80.0
                }
              ]
     end


lib/plausible/stats/sql/special_metrics.ex
```diff
@@ -44,7 +44,7 @@ defmodule Plausible.Stats.SQL.SpecialMetrics do
     |> select_merge_as([], %{
       percentage:
         fragment(
-          "if(? > 0, round(? / ? * 100, 1), null)",
+          "if(? > 0, round(? / ? * 100, 2), null)",
```
apata, line 47: question: If we include another digit, it implies that it's a meaningful digit. Is our tracking accuracy really at this level? Let's say we show the value of 50.04%. Is the true value (assuming perfect tracking) in the range 50.03% ... 50.05% or more like 50.0% ... 50.1%? 

I'm also concerned that historical stats API requests would start yielding slightly different values, which may cause frustration when compared with previously requested data.

test/plausible_web/controllers/api/stats_controller/imported_test.exs
```diff
@@ -877,20 +884,22 @@ defmodule PlausibleWeb.Api.StatsController.ImportedTest do
...
                  %{
-                   "bounce_rate" => 0,
+                   "bounce_rate" => 0.0,
                    "time_on_page" => 60,
                    "visitors" => 3,
                    "pageviews" => 4,
                    "scroll_depth" => nil,
-                   "name" => "/some-other-page"
+                   "name" => "/some-other-page",
+                   "percentage" => 300.0
```
apata: issue, blocking: something seems to be off when we consider imported data in our calculations. Perhaps its a separate bug and a  matter for another time, but I think it may be surprising to customers to see this sort of values.

lib/plausible/stats/sql/special_metrics.ex
sanne-san: The reason this was introduced is because there's often a long tail of <0.1% items. Rather than showing 0% for all of these, we'd like to show 0.05%, 0.06%, etc. So in the majority of cases we don't care about the second decimal, only for numbers below 0.1.

apata: Roger, thanks for explaining!

test/plausible_web/controllers/api/stats_controller/imported_test.exs
apata: The issue was with the test scenario not being representative of a real import. In a real import, imported_visitors table would also be populated: it would have at least as many imported visitors as the biggest source in imported_sources. So if there's Source A with 2 visitors and Source B with 1 visitor, there should be at least 2 imported visitors in imported_visitors table.

test/plausible_web/controllers/api/stats_controller/pages_test.exs
```diff
@@ -942,6 +964,12 @@ defmodule PlausibleWeb.Api.StatsController.PagesTest do
         )
       ])
 
+      populate_stats(
+        site,
+        site_import.id,
+        for(_ <- 1..25, do: build(:imported_visitors, date: ~D[2020-01-01]))
```
apata: question: Why did we have to insert 25 visitors here rather than 20 visitors from imported source /native-and-imported + 4 from /imported-only? At the moment, it's still an invalid import because one of the imported_visitors doesn't have a source. Is the percentage assert maybe wrong

apata (approved):

PR #5924, opened 2025-12-01
Add :segment_id field to shared_links schema

apata:
### Changes

Adds field to shared_links schema that allows limiting that shared link to this particular segment only.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/priv/repo/migrations/20251201154500_add_limited_to_segment_to_shared_links.exs
+++ b/priv/repo/migrations/20251201154500_add_limited_to_segment_to_shared_links.exs
@@ -5,6 +5,7 @@ defmodule Plausible.Repo.Migrations.AddLimitedToSegmentToSharedLinks do
     alter table(:shared_links) do
       add :segment_id, references(:segments, on_delete: :delete_all)
     end
+
     create index(:shared_links, [:segment_id])
   end
 end


priv/repo/migrations/20251201154500_add_limited_to_segment_to_shared_links.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Plausible.Repo.Migrations.AddLimitedToSegmentToSharedLinks do
+  use Ecto.Migration
+
+  def change do
+    alter table(:shared_links) do
+      add :limited_to_segment_id, :integer, null: true
```
aerosol: Are we consciously rejecting foreign key constraint here?

apata: That's a good question! With this semi-conscious decision, I prioritized segment delete convenience over shared link consistency, but now that I think about it, it's not obviously the right way to go. Asking for more input from CS.

Supposing we add the foreign key constraint here. We can't null the field automatically on segment delete because `null` value `:limited_to_segment_id` has the meaning that the view is not limited. 
We'd have to define a value for the field representing a "deleted segment" to automatically update all the references on delete, making the schema more complex. 
OR we'd have to let the user manually solve all those references in shared links before they can delete the segment, which seems inconvenient.

aerosol: Are you saying that limited link shouldn't become unlimited when the segment is deleted? I think it makes perfect sense, but maybe I misunderstood?

apata: We decided to have the FK constraint and cascade delete, and to prevent unintended deletes, we will show a confirmation dialog that lists the shared links that will be deleted.

priv/repo/migrations/20251201154500_add_limited_to_segment_to_shared_links.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Plausible.Repo.Migrations.AddLimitedToSegmentToSharedLinks do
+  use Ecto.Migration
+
+  def change do
+    alter table(:shared_links) do
+      add :segment_id, references(:segments, on_delete: :delete_all)
+    end
+
+    create index(:shared_links, [:segment_id])
```
aerosol, line 9: this feels unnecessary, foreign keys are indexes on their own?

apata: Postgres docs advise creating an index on the referencing columns: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK:~:text=Since%20a%20DELETE,the%20referencing%20columns.

aerosol: :facepalm: /me goes on a walk of shame

apata: Ain't no shame, thanks for raising the question!

PR #5925, opened 2025-12-02
Allow creating and editing limited views

apata:
### Changes

- [x] Merge shared_links schema change https://github.com/plausible/analytics/pull/5924 and migrate
- [x] Allow creating and editing limited views
- [x] Indicate limited view on shared link dashboard
- [x] Ensure queries are made with the limited view

### TODO
- [x] Fix regression from unexpected horizontal overflow of table cells

### Tests

### Changelog

### Documentation
- [ ] https://github.com/plausible/docs/pull/655

### Dark mode

### DIFF
--- a/lib/plausible_web/live/components/form.ex
+++ b/lib/plausible_web/live/components/form.ex
@@ -54,7 +54,6 @@ defmodule PlausibleWeb.Live.Components.Form do
   attr(:max_one_error, :boolean, default: false)
   slot(:help_content)
   slot(:inner_block)
-  slot(:link)
 
   def input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
     errors = if Phoenix.Component.used_input?(field), do: field.errors, else: []
@@ -91,9 +90,6 @@ defmodule PlausibleWeb.Live.Components.Form do
         <option :if={@prompt} value="">{@prompt}</option>
         {Phoenix.HTML.Form.options_for_select(@options, @value)}
       </select>
-      <div :if={@link != [] && Enum.empty?(@errors)} class="mt-1">
-        {render_slot(@link)}
-      </div>
       <.error :for={msg <- @errors}>{msg}</.error>
     </div>
     """
@@ -206,9 +202,6 @@ defmodule PlausibleWeb.Live.Components.Form do
         {@rest}
       />
       {render_slot(@inner_block)}
-      <div :if={@link != [] && Enum.empty?(@errors)} class="mt-0.5">
-        {render_slot(@link)}
-      </div>
       <.error :for={msg <- @errors}>
         {msg}
       </.error>
--- a/lib/plausible_web/live/shared_link_settings/form.ex
+++ b/lib/plausible_web/live/shared_link_settings/form.ex
@@ -73,6 +73,7 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.Form do
                 {@shared_link.segment.id, @shared_link.segment.name}
             }
           />
+          <.learn_more />
         </div>
       </div>
 
@@ -147,6 +148,7 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.Form do
             suggest_fun={fn input, _ -> get_segment_suggestions(assigns.site, input) end}
             selected={nil}
           />
+          <.learn_more />
         </div>
       </div>
       <.button type="submit" class="w-full">
@@ -155,6 +157,19 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.Form do
     </.form>
     """
   end
+
+  defp learn_more(assigns),
+    do: ~H"""
+    <div class="mt-1">
+      <PlausibleWeb.Components.Generic.unstyled_link
+        href="https://plausible.io/docs/filters-segments#how-to-save-a-segment"
+        new_tab
+        class="text-xs text-indigo-600 dark:text-indigo-400"
+      >
+        Learn about segments
+      </PlausibleWeb.Components.Generic.unstyled_link>
+    </div>
+    """
 
   defp get_segment_suggestions(site, input) do
     {:ok, segments} = Plausible.Segments.search_by_name(site, input)
--- a/test/plausible_web/live/shared_link_settings/form_test.exs
+++ b/test/plausible_web/live/shared_link_settings/form_test.exs
@@ -20,9 +20,19 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.FormTest do
       html = render(lv)
 
       assert html =~ "New shared link"
+      assert html =~ "Password protect"
+      assert html =~ "Limit to segment"
+      assert element_exists?(html, ~s|button[role="switch"]#password-protect-|)
+      assert element_exists?(html, ~s|button[role="switch"]#limit-view-|)
       assert element_exists?(html, ~s|input[name="shared_link[name]"]|)
       assert element_exists?(html, ~s|input[name="shared_link[password]"][type="password"]|)
+      assert element_exists?(html, ~s|input[name="shared_link[segment_id]"]|)
       assert element_exists?(html, ~s|button[type="submit"]|)
+
+      assert element_exists?(
+               html,
+               ~s|a[href="https://plausible.io/docs/filters-segments#how-to-save-a-segment"]|
+             )
     end
 
     test "renders error on empty submission", %{conn: conn, session: session} do
@@ -56,6 +66,31 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.FormTest do
       refute shared_link.password_hash
     end
 
+    test "creates a shared link limited to segment", %{conn: conn, site: site, session: session} do
+      segment = insert(:segment, type: :site, site: site, name: "Scandinavia")
+      lv = get_liveview(conn, session)
+
+      lv |> element("button#add-shared-link-button") |> render_click()
+
+      lv
+      |> find_form()
+      |> render_submit(%{shared_link: %{name: "Limited to Scandinavia", segment_id: segment.id}})
+
+      html = render(lv)
+      assert html =~ "Limited to Scandinavia"
+      assert html =~ "Shared link saved"
+
+      shared_link =
+        Plausible.Repo.get_by(Plausible.Site.SharedLink,
+          name: "Limited to Scandinavia",
+          site_id: site.id,
+          segment_id: segment.id
+        )
+
+      assert shared_link
+      refute shared_link.password_hash
+    end
+
     test "creates a shared link with password", %{conn: conn, site: site, session: session} do
       lv = get_liveview(conn, session)
       refute render(lv) =~ "Protected Link"
@@ -77,11 +112,12 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.FormTest do
       assert shared_link.password_hash
     end
 
-    test "renders form fields for editing a shared link", %{
-      conn: conn,
-      site: site,
-      session: session
-    } do
+    test "renders form fields for editing a shared link (not limited to segment, not protected)",
+         %{
+           conn: conn,
+           site: site,
+           session: session
+         } do
       shared_link = insert(:shared_link, site: site, name: "Existing Link")
 
       lv = get_liveview(conn, session)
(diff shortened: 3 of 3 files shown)


aerosol:
There's some conflicts that need to be resolved

lib/plausible/segments/segments.ex
```diff
@@ -70,6 +70,63 @@ defmodule Plausible.Segments do
     {:ok, Repo.all(query)}
   end
 
+  def search_by_name(%Plausible.Site{} = site, name) do
+    type = :site
+    fields = [:id, :name]
+
+    name_empty? = is_nil(name) or (is_binary(name) and String.trim(name) == "")
```
aerosol: When do we expect this to be `nil`? How about call it `input` - following the call site - or even `user_input`and assume binary in the function clause?

lib/plausible/segments/segments.ex
```diff
@@ -70,6 +70,63 @@ defmodule Plausible.Segments do
...
+      else
+        from([segment] in base_query,
+          select: %{
+            id: segment.id,
+            name: segment.name,
+            match_rank:
+              fragment(
+                "CASE
+                  WHEN lower(?) = lower(?) THEN 0                    -- exact match
+                  WHEN lower(?) LIKE lower(?) || '%' THEN 1          -- starts with
```
aerosol, line 101: This is nice but a bit of an overkill, since there's only 4 customers with more than 20 site segments. I'd consider simplifying that query and using `Enum.sort_by/3` combined with `String.jaro_distance/2` to achieve similar effectiveness and arguably simpler code. Ignore if you wish.

lib/plausible_web/controllers/api/internal/segments_controller.ex
```diff
@@ -100,6 +100,33 @@ defmodule PlausibleWeb.Api.Internal.SegmentsController do
...
+
+      {:error, :segment_not_found} ->
+        segment_not_found(conn, params["segment_id"])
+
+      {:ok, shared_links} ->
+        json(conn, shared_links |> Enum.map(& &1.name))
+    end
+  end
+
+  def get_related_shared_links(%Plug.Conn{} = conn, _params), do: invalid_request(conn)
```
aerosol, line 128: Isn't that redundant? If no controller clause matched, phoenix should respond with a bad request, no?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1599,6 +1600,42 @@ defmodule PlausibleWeb.Api.StatsController do
     end
   end
 
+  defp validate_required_filters_plug(conn, _opts) do
+    if conn.assigns.shared_link && conn.assigns.shared_link.segment_id do
+      segment_id = conn.assigns.shared_link.segment_id
```
aerosol: Could match in the function clause instead. I feel the plug name could be better here, since it's more specific to segments than filters?

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1599,6 +1600,42 @@ defmodule PlausibleWeb.Api.StatsController do
...
+            "The first filter must be for the segment with id #{segment_id}"
+          )
+      end
+    else
+      conn
+    end
+  end
+
+  defp ensure_expected_segment_filter_present(
+         %{"filters" => filters} = _params,
```
aerosol, line 1640: will users ever run into an error like that? Unclear why do we need to enforce a filter to be "first".

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -405,12 +406,28 @@ defmodule PlausibleWeb.StatsController do
...
         current_user = conn.assigns[:current_user]
         site_role = get_fallback_site_role(conn)
-        shared_link = Plausible.Repo.preload(shared_link, site: :owners)
+        shared_link = Plausible.Repo.preload(shared_link, site: [:owners], segment: [])
         stats_start_date = Plausible.Sites.stats_start_date(shared_link.site)
 
         flags = get_flags(current_user, shared_link.site)
-
-        {:ok, segments} = Plausible.Segments.get_all_for_site(shared_link.site, site_role)
+        limited_to_segment_id = shared_link.segment && shared_link.segment.id
```
aerosol: I'd make this `SharedLink.limited_to_segment?/1`

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -405,12 +406,28 @@ defmodule PlausibleWeb.StatsController do
...
+
+        segments =
+          if is_nil(limited_to_segment_id) do
+            {:ok, segments} = Plausible.Segments.get_all_for_site(shared_link.site, site_role)
+            segments
+          else
+            shared_link.segment
```
aerosol, line 426: This doesn't seem to be tested anywhere?

lib/plausible_web/live/components/form.ex
```diff
@@ -74,8 +75,14 @@ defmodule PlausibleWeb.Live.Components.Form do
 
   def input(%{type: "select"} = assigns) do
     ~H"""
-    <div class={@mt? && "mt-4"}>
-      <.label for={@id} class={if @help_text, do: "mb-0.5", else: "mb-1.5"}>{@label}</.label>
+    <div class={@mt? && "mt-6"}>
+      <.label
+        :if={@label != nil and @label != ""}
```
aerosol, line 79: also known as `:if={@label}`? 
When is it ever `""`?

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -405,12 +406,28 @@ defmodule PlausibleWeb.StatsController do
       not Teams.locked?(shared_link.site.team) ->
         current_user = conn.assigns[:current_user]
         site_role = get_fallback_site_role(conn)
-        shared_link = Plausible.Repo.preload(shared_link, site: :owners)
+        shared_link = Plausible.Repo.preload(shared_link, site: [:owners], segment: [])
```
aerosol: The intention behind preloading an empty list is unclear here. Don't understand what's going on. Did you mean `:segment`? Oh and you've got `Repo` aliased here.

lib/plausible_web/live/shared_link_settings/form.ex
```diff
@@ -43,6 +44,39 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.Form do
...
+            if (!limitViewEnabled) {
+              input.value = '';
+            }
+          }
+        "
+        class="flex flex-col gap-y-2"
+      >
+        <PlausibleWeb.Components.Generic.toggle_field
+          id="limit-view"
+          id_suffix=""
```
aerosol: That's the default, no?

lib/plausible_web/live/shared_link_settings/form.ex
```diff
@@ -55,27 +89,89 @@ defmodule PlausibleWeb.Live.SharedLinkSettings.Form do
...
+            } else {
+              input.value = '';
+            }
+          }
+        "}
+        class="flex flex-col gap-y-2"
+      >
+        <PlausibleWeb.Components.Generic.toggle_field
+          id="password-protect"
+          id_suffix=""
```
aerosol: ditto, default, can we omit?

lib/plausible_web/components/generic.ex
```diff
@@ -471,6 +471,51 @@ defmodule PlausibleWeb.Components.Generic do
...
+  attr :disabled, :boolean, default: false
+  attr :label, :string, required: true
+  attr :help_text, :string, default: nil
+  attr :help_text_conditional, :boolean, default: false
+  attr :mt?, :boolean, default: true
+
+  attr(:rest, :global)
+
+  def toggle_field(assigns) do
+    help_text_conditional = assigns[:help_text_conditional] || false
```
aerosol: why do we need this if `attr` macro defines a default?

Wondering if we could simplify this. As a component user I'd have very hard time figuring out what do I need to pass to `help_text_conditional`, what does it do even, not to mention `js_active_var`. Not a fan of this interface.

lib/plausible_web/templates/stats/stats.html.heex
```diff
@@ -52,6 +52,7 @@
     data-is-consolidated-view={Jason.encode!(@consolidated_view?)}
     data-consolidated-view-available={Jason.encode!(@consolidated_view_available?)}
     data-team-identifier={@team_identifier}
+    data-limited-to-segment-id={Jason.encode!(@limited_to_segment_id)}
```
aerosol, line 55: Will injecting some random integer here make react request someone else's segment?

lib/plausible_web/controllers/stats_controller.ex
apata: Thanks! The intent is to preload both owners and segment relation, and that's the syntax I got from Ecto docs. It works, but I'll check whether it can be simpler.

lib/plausible_web/controllers/api/internal/segments_controller.ex
apata: Checking, thought it responds with internal server error.

lib/plausible_web/templates/stats/stats.html.heex
apata: Good question, the answer is no.

There's actually no GET /segments/:id or GET /segments endpoints for the FE to call. 

There used to be at one point, but there aren't any more. 

The way it works now is: We stuff all the site's segments into data-segments, parse it in React, and update that FE segments list whenever the user changes or deletes one. Or the list gets updated whenever the dashboard is refreshed. It's not a great system, but does the job. LV dashboard should allow us to make it right.

If this `data-limited-to-segment-id` is set, it's the ID of the single segment stuffed into data-segments. If there's no such segment in data-segments, it's treated as if it was null. Because it was a shared link and the stats BE still expects segment filter to be present, all the reports get errors, which visually looks like the endless loader.

lib/plausible_web/controllers/api/stats_controller.ex
apata: Good question! They won't if they play by the rules. If they try to request stats with the wrong filters though, they'll get a bad request error. The alternative to having it first is allow it to be anywhere, which we wouldn't be able to match on. Limited view FE forces it to be first, so I thought we might as well make use of that. Will add a test for this error.

Matching on raw filters params before they go through Query parsing, as I do here, is not ideal though. If I were to match on filters after Query.from, I'd have to add this little function to every internal stats endpoint, so I went this way.

lib/plausible_web/controllers/api/stats_controller.ex
apata: Roger that on the match! Regarding the name, I specifically wanted it to be more general, so readers wouldn't need to be aware of the details of segments and limited views, just that some filters may be required.

lib/plausible/segments/segments.ex
apata: Thanks for looking those stats up! It is indeed overkill then for this particular combobox. I liked how it came out though. I suggest we keep it and introduce it to other comboboxes as well. Server side sorting means we should query for bigger pages - it'd be good to be able to avoid it.

lib/plausible_web/controllers/stats_controller.ex
apata: Well spotted, will add test!

lib/plausible_web/controllers/api/internal/segments_controller.ex
apata: Checked just now. If the endpoint handler doesn't have any matching function clauses, it returns an internal server error. These fallbacks (all of the segments endpoints have them) haven't been triggered lately, maybe never (checked some traces). Looks like our authorization plugs are pretty robust. Nothing in the FE depends on 400 over 500, so I could remove these clauses. It's just that all of these endpoints will be superfluous after LV dashboard anyway, so I don't see the point in perfecting them.

lib/plausible_web/live/components/form.ex
apata: True, it does look superfluous! However, I just checked that if the rendered component is `<.input label="" ...>`, with just the check `:if={@label}`, the label box gets rendered. Since this exact check is present on another component, maybe it can stay for the sake of consistency?

lib/plausible_web/live/shared_link_settings/form.ex
apata: https://github.com/plausible/analytics/pull/5925/changes/288ca80fcd25e0d9a1f647ffac29846dcc20d1c1

lib/plausible_web/controllers/stats_controller.ex
apata: It can be simpler, just that the keyword list within a keyword list needs to be last elem: https://github.com/plausible/analytics/pull/5925/changes/ea458fafb17023795ca255febb59161b19515e04

lib/plausible_web/controllers/api/stats_controller.ex
apata: Added test: https://github.com/plausible/analytics/pull/5925/changes/88a80a1ba78f1db742b89f821f6ce6666ef05bde

lib/plausible_web/controllers/api/stats_controller.ex
apata: Refactored with function clauses, kept the name: https://github.com/plausible/analytics/commit/88a80a1ba78f1db742b89f821f6ce6666ef05bde

lib/plausible_web/controllers/stats_controller.ex
apata: https://github.com/plausible/analytics/pull/5925/changes/a8ff7f69a95770cf7338316ccd88cb08033be606

lib/plausible_web/live/shared_link_settings/form.ex
apata: https://github.com/plausible/analytics/commit/288ca80fcd25e0d9a1f647ffac29846dcc20d1c1

lib/plausible_web/components/generic.ex
apata: Agreed that the interface was a bit clunky and mysterious. How's this? https://github.com/plausible/analytics/pull/5925/commits/84e6a071b0e699de757bd742300fbdbce4220cb0

lib/plausible_web/live/components/form.ex
```diff
@@ -84,6 +91,9 @@ defmodule PlausibleWeb.Live.Components.Form do
         <option :if={@prompt} value="">{@prompt}</option>
         {Phoenix.HTML.Form.options_for_select(@options, @value)}
       </select>
+      <div :if={@link != [] && Enum.empty?(@errors)} class="mt-1">
```
ukutaht: Where is this used?

apata: Good catch! It's not at the moment, but it used to be needed to link to segments doc from create shared link. Not sure why I removed it. Will add it back in

Image

apata: https://github.com/plausible/analytics/pull/5925/commits/85b587f3dfb9dd4f3567dbb5b0bcfa4f82032683

PR #5930, opened 2025-12-03
Add necessary scaffolding for enabling LV on dashboard

zoldar:
### Changes

This PR introduces basics for supporting gradual deployment of  LiveView site dashboard elements.

- Add a forked `phoenix_live_view` dependency with added switch allowing disabling pushState calls (the branch is based off v1.1.18 tag)
- Configure liveSocket to disable pushState specifically for dashboard view and load user prefs from localStorage using meta tags; meta tags themselves are behind a newly introduced `live_dashboard` feature flag
- Implement LiveDashboard hook allowing to install various "widgets". The main widget, "dashboard-root" implements delegation of navigation to and from `reactor-dom-router`; respective event dispatching and handlers are added on React end in `navigation/use-app-navigate.tsx` and in `dashboard/index.tsx`
- A `Live.Dashboard` LV is now embedded in dashboard stats view, behind a feature flag
- A scaffolding of pages breakdown is implemented, swapped with `<Pages />` react component via `LiveViewPortal` component, also behind a flag
- Implement basic tile and tab components with optimistic loading of tab selections and persisting tab choice client-side using localStorage
- Add a very rudimentary LV test case

### Tests

### DIFF
--- a/assets/js/liveview/live_socket.js
+++ b/assets/js/liveview/live_socket.js
@@ -10,7 +10,7 @@ import { LiveSocket } from 'phoenix_live_view'
 import { Modal, Dropdown } from 'prima'
 import LiveDashboard from './live_dashboard'
 import topbar from 'topbar'
-/* eslint-disable import/no-unresolved */
+/* eslint-enable import/no-unresolved */
 
 import Alpine from 'alpinejs'
 


assets/js/liveview/live_dashboard.js
```diff
@@ -0,0 +1,156 @@
+/**
+ * Hook used by LiveView dashboard.
+ *
+ * Defines various widgets to use by various dashboard specific components.
+ */
+
+const WIDGETS = {
```
ukutaht, line 7: Personally I'm not a fan of introducing this new concept (widget). It took a bit of jumping around and reading the code to understand it's just a liveview hook. Why not call them what they are - hooks - and register them directly instead of via `data-widget`? This way anyone who's familiar with liveview (and the concept of hooks) understands immediately what's going on without having to learn and get used to an intermediary layer.

assets/js/liveview/live_dashboard.js
```diff
@@ -0,0 +1,156 @@
...
+        const type = e.target.dataset.type || null
+
+        if (type === 'dashboard-link') {
+          this.url = e.target.href
+          const uri = new URL(this.url)
+          // Domain is dropped from URL prefix, because that's what react-dom-router
+          // expects.
+          const path = '/' + uri.pathname.split('/').slice(2).join('/')
+          this.el.dispatchEvent(
+            new CustomEvent('dashboard:live-navigate', {
```
ukutaht, line 25: Really like the solution for frontend navigation and getting react <> liveview to talk to each other. Impressive work 👏

assets/js/liveview/live_dashboard.js
ukutaht: By the way - I do like the shared structure for adding/removing listeners and cleaning them up. In Prima I have the same problem where all the components duplicate some amount of this structure.

I wonder if this can be solved via some kind of mixin or inheritance instead. I haven't really experimented with sharing behaviour between hooks yet.

lib/plausible_web/live/components/dashboard/base.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  def dashboard_link(assigns) do
+    url = "/" <> assigns.site.domain <> assigns.href
+
+    assigns = assign(assigns, :url, url)
+
+    ~H"""
+    <.link
+      data-type="dashboard-link"
+      href={@url}
```
ukutaht, line 22: We'll probably want to change this to `patch` over `href` eventually to get pushstate routing instead of http-based routing. Is there something preventing us from using `patch` from the get-go?

lib/plausible_web/live/components/dashboard/tile.ex
```diff
@@ -0,0 +1,92 @@
...
+      <div data-tile class="w-full flex justify-between h-full">
+        <div id={@id <> "-title"} class="flex gap-x-1" phx-update={@update_mode}>
+          <h3 data-title class="font-bold dark:text-gray-100">{@title}</h3>
+        </div>
+
+        <div
+          :if={@tabs != []}
+          id={@id <> "-tabs"}
+          phx-update={@update_mode}
+          phx-hook="LiveDashboard"
```
ukutaht, line 32: The optimistic loading is currently implemented via custom JS in the hook. I wonder if this could be achieved via `Phoenix.LiveView.JS{}` commands instead and whether that would be any cleaner in the end (requires trying it out).

lib/plausible_web/live/components/dashboard/tile.ex
```diff
@@ -0,0 +1,92 @@
...
+  # Optimistic rendering requires preventing LV patching of
+  # title and tabs. The update of those is handled by `tab` 
+  # widget hook.
+  attr :connected?, :boolean, required: true
+
+  slot :tabs
+  slot :inner_block, required: true
+
+  def tile(assigns) do
+    assigns = assign(assigns, :update_mode, if(assigns.connected?, do: "ignore", else: "replace"))
```
ukutaht, line 19: I understand why we ignore LV patches but what's the purpose of using `phx-update="replace"` when socket is not connected?

lib/plausible_web/live/dashboard.ex
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def handle_params_internal(_params, _url, socket) do
+    {:noreply, socket}
+  end
+
+  def render(assigns) do
+    ~H"""
+    <div id="live-dashboard-container" phx-hook="LiveDashboard" data-widget="dashboard-root">
+      <.portal_wrapper id="pages-breakdown-live-container" target="#pages-breakdown-live">
```
ukutaht, line 51: Clever!

lib/plausible_web/live/components/dashboard/tile.ex
```diff
@@ -0,0 +1,92 @@
...
+
+    ~H"""
+    <div data-tile id={@id}>
+      <div data-tile class="w-full flex justify-between h-full">
```
ukutaht, line 23: Why the duplicated and nested `data-tile` attributes?

lib/plausible_web/live/components/dashboard/tile.ex
```diff
@@ -0,0 +1,92 @@
...
+
+      {render_slot(@inner_block)}
+    </div>
+    """
+  end
+
+  attr :id, :string, required: true
+  slot :inner_block, required: true
+
+  def tabs(assigns) do
```
ukutaht, line 48: Is this used?

lib/plausible_web/live/components/dashboard/tile.ex
ukutaht: I'll play around with this a bit

ukutaht: I have some ideas but I don't think it's important to focus on this right now. I'll implement these ideas in an upcoming `Prima.Tabs` component separately and plan to use it here as well.

zoldar:
> Please add me as a reviewer in the future to these foundational PRs that establish liveview patterns for the dashboard.

I think I have added you as a reviewer - I haven't waited for your review though - sorry about that. Either way, things here are going to undergo ongoing improvements as we go anyway.

assets/js/liveview/live_dashboard.js
zoldar: My worry was about having to add multitude of hooks, majority of them concerned with dashboard specific stuff. I'm not saying it's rational 😅 And you are right, it's not worth introducing cognitive load of a new concept I came up with "on the go". I'll see what can we do about structure sharing.

lib/plausible_web/live/components/dashboard/base.ex
zoldar: Hmm, I got the impression patch is heavily tied to pushState - I'll double check that assumption though, maybe I have misinterpreted something. Either way, keeping things link dashboard links abstracted should let as replace `href` with `patch` where needed during the final migration phase.

lib/plausible_web/live/components/dashboard/tile.ex
zoldar: `replace` is the default (there's no empty value). Though, now that I think of it, dead view most likely does not care about `phx-update` attribute at all, so it can be always set to `ignore`. I'll change that, thanks.

lib/plausible_web/live/components/dashboard/tile.ex
zoldar: That's a brainfart, I'll fix it.

lib/plausible_web/live/components/dashboard/tile.ex
zoldar: Ah, it's dead code from earlier iteration, I'll remove it, thanks.

ukutaht:
> I think I have added you as a reviewer - I haven't waited for your review though - sorry about that. Either way, things here are going to undergo ongoing improvements as we go anyway.

Oh you did, I just didn't review in time :) No worries

lib/plausible_web/live/components/dashboard/tile.ex
ukutaht: Yeah that's exactly what I was thinking 👍

lib/plausible_web/live/components/dashboard/base.ex
ukutaht: > Either way, keeping things link dashboard links abstracted should let as replace href with patch where needed during the final migration phase.

True, feel free to ignore this for now. I don't have much experience with live navigation but my impression is that:
1. `patch` calls `handle_params`, keeps liveview mounted and socket connected
2. `navigate` mounts another liveview while keeping socket connected
3. `href` does a full http cycle so socket is re-connected and any liveviews have to remount

As I understand, `href` should only be used when navigating to a dead view or external resource.

But as you say, this can be easily changed when we switch navigation over to liveview completely at the end.

assets/js/liveview/live_dashboard.js
ukutaht: > My worry was about having to add multitude of hooks, majority of them concerned with dashboard specific stuff. I'm not saying it's rational

Haha yeah I don't think there's a problem with having many hooks. Surely `hooks` is just a map in liveview that looks them up in the same way as `WIDGETS[this.widget]`.

It is a bit annoying to have to manually import and register them all when initializing a liveSocket. But I think there's also value in being explicit rather than magical. For namespacing and organization perhaps we can make something like this:

```js
// live_dashboard.js

import HookA from "./hooks/hook_a.js"
import HookB from "./hooks/hook_b.js"
import HookC from "./hooks/hook_c.js"

export default {
  HookA,
  HookB,
  HookC
}

// live_socket.js

import LiveDashboard from './live_dashboard.js

let Hooks = { Modal, Dropdown, LiveDashboard } // these are given to LiveSocket during initialization

// dashboard_component.ex

<div phx-hook="LiveDashboard.HookA">
```


There's also a new [ColocatedHook](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.ColocatedHook.html) feature in Phoenix. I haven't tried it, not sure if/how it would work if you want to import some shared helpers or dependencies in these co-located hooks for example.

lib/plausible_web/live/components/dashboard/tile.ex
ukutaht: One idea might be prudent to talk about now.

As a general principle I've started preferring to style different states via html attribute rather than shuffling classes around. Quite common with tailwind. Here's what I mean:

Currently when a tab active state changes, the css classes are changed as well via:

```elixir
   def tab(assigns) do
    assigns =
      assign(assigns,
        active_classes:
          "text-indigo-600 dark:text-indigo-500 font-bold underline decoration-2 decoration-indigo-600 dark:decoration-indigo-500",
        inactive_classes: "hover:text-indigo-700 dark:hover:text-indigo-400 cursor-pointer"
      )

    ~H"""
    <button
      class="rounded-sm truncate text-left transition-colors duration-150"
      data-tab={@value}
      data-label={@label}
      data-storage-key="pageTab"
      data-active-classes={@active_classes}
      data-inactive-classes={@inactive_classes}
      phx-click="set-tab"
      phx-value-tab={@value}
      phx-target={@target}
    >
      <span class={if(@value == @active, do: @active_classes, else: @inactive_classes)}>
        {@label}
      </span>
    </button>
    """
  end
```

vs my preferred approach where the class list is static and styles the component via data-attribute:

```elixir
 def tab(assigns) do
    ~H"""
    <button
      class="rounded-sm truncate text-left transition-colors duration-150"
      data-active={@active} // <- this is used for styling 
      data-tab={@value}
      data-label={@label}
      data-storage-key="pageTab"
      phx-click="set-tab"
      phx-value-tab={@value}
      phx-target={@target}
    >
      <-- active classes are prefixed with `data-active:` -->
      <span class="hover:text-indigo-700 dark:hover:text-indigo-400 cursor-pointer
  data-active:text-indigo-600 data-active:dark:text-indigo-500 data-active:font-bold 
  data-active:underline data-active:decoration-2 data-active:decoration-indigo-600 
  data-active:dark:decoration-indigo-500"> 
        {@label}
      </span>
    </button>
    """
  end
```

It doesn't work exactly the same because the base/inactive classes are not removed when state changes. So for example you might need to also override `data-active:cursor-normal` to get the original cursor styling for active state.

For me when working with styling I do prefer to have a static list like this to work with. It's especially helpful when debugging styling with devtools in the browser. Also using the data-attribute to announce different states is quite useful for testing and debugging.

@sanne-san WDYT?

assets/js/liveview/live_dashboard.js
zoldar: WDYT about the approach taken in https://github.com/plausible/analytics/pull/5937 ?

lib/plausible_web/live/components/dashboard/base.ex
zoldar: I have changed it to `patch` (https://github.com/plausible/analytics/pull/5937) as it currently does not really matter - `href` is rendered in the markup either way and the click is "hijacked" by the navigation hooks, so it will still work.

assets/js/liveview/live_dashboard.js
ukutaht: Looks good!

PR #5932, opened 2025-12-04
Validate password cookie for password-protected shared links on internal stats API requests

apata:
### Changes

- To make internal stats API requests for password-protected shared links, shared link auth cookie must be set in the requests.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/site/shared_link.ex
+++ b/lib/plausible/site/shared_link.ex
@@ -46,6 +46,6 @@ defmodule Plausible.Site.SharedLink do
     end
   end
 
-  def get_type(%__MODULE__{password_hash: hash}) when not is_nil(hash), do: :password_protected
-  def get_type(%__MODULE__{}), do: :unlisted
+  def password_protected?(%__MODULE__{password_hash: hash}) when not is_nil(hash), do: true
+  def password_protected?(%__MODULE__{}), do: false
 end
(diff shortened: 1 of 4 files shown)


lib/plausible/site/shared_link.ex
```diff
@@ -45,4 +45,7 @@ defmodule Plausible.Site.SharedLink do
         change(link, password_hash: hash)
     end
   end
+
+  def get_type(%__MODULE__{password_hash: hash}) when not is_nil(hash), do: :password_protected
+  def get_type(%__MODULE__{}), do: :unlisted
```
ukutaht: naming: `:unlisted` doesn't feel descriptive to me, not sure if it's some kind of terminology I'm not familiar with? How about `:public`?

ukutaht (approved):
One comment about naming, otherwise LGTM 👍

lib/plausible/site/shared_link.ex
apata: Thanks! I agree that it's a bit mystical, but that name was in the codebase already. I'm wary of :public because it might be confused with the fully public dashboard. 

I can go simple with `:with_password`, `:without_password` or just `def password_protected?(link)` `true / false`? If we go with the boolean, matching on it isn't as pretty in `with ... <- {function name}`

apata: In plugins API, shared_link schema has a field :password_protected. I've decided to unify the checks with that in https://github.com/plausible/analytics/pull/5932/commits/b8bb107d9118c1318998db2e7a138a546f85990a

PR #5935, opened 2025-12-08
Display segment filters to anyone that can see the dashboard being filtered by the segment

apata:
### Changes

- Segment filters are visible to anyone who can view the dashboard with that segment applied, including personal segments on public dashboards

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-- Segment filters are displayed to anyone that can see the dashboard filtered by the segment, including for a personal segment applied on a public dashboard
+- Segment filters are visible to anyone who can view the dashboard with that segment applied, including personal segments on public dashboards
 
 ### Fixed
 


apata:
To see what's changed, click on the segment pill on these dashboards
https://pr-5935.review.plausible.io/example-5935.site?f=is,segment,61&l=segment-61,artur%27s%20segment
https://pr-5935.review.plausible.io/example-5935.site?f=is,segment,62&l=segment-62,site%20segment
vs
https://staging.plausible.io/example-5935.site/filter/segment?f=is,segment,61&l=segment-61,artur%27s%20segment
https://staging.plausible.io/example-5935.site?f=is,segment,62&l=segment-62,site%20segment

assets/js/dashboard/filtering/segments.ts
```diff
@@ -148,6 +158,36 @@ export function resolveFilters(
   })
 }
 
+export function canExpandSegment({
```
ukutaht, line 161: Sorry I don't have the full context here but in what case would the user be able to see the dashboard but not expand the segment at all?

In other words, why isn't `canExpandSegment` always `true`?

ukutaht: From @metmarkosaric's comment [here](https://3.basecamp.com/5308029/buckets/44692882/card_tables/cards/9256703917#__recording_9349447719) it seems like whenever a segment is shown to the user they should be able to expand it. So I'm questioning the need to have any conditional logic for expanding at all.

sanne-san (approved):

assets/js/dashboard/filtering/segments.ts
apata: Thanks! From my perspective, seeing the contents of a segment and expanding it to the filters bar are different actions. 

Expanding puts the constituent filters on the filters bar (and in URL) which also exposes the Update | Save a copy | Delete | Close segment menu. 

Of that menu, many user roles (think :public or :viewer looking at a site segment) wouldn't be able to take particular actions. We'd have to bring the messy conditions in Plausible.Segments to the FE to conditionally enable these buttons, or risk disappointing users (makes changes, clicks Save, gets 403). 

Also, in Limited Views, the way I've implemented the unremovable segment in the FE wouldn't work if we allow :public role to expand the segment. 

I do think it'd be good to allow anyone to expand and save a copy for example. It will be easier to do with the LV dashboard though.

aerosol:
So is that change about not displaying the red warning?

apata:
> So is that change about not displaying the red warning?

@aerosol Yep! Before this, personal and site segment details weren't displayed to specific site roles, even though those roles were allowed to see the data filtered by the segment. The opacity had some benefits, since filters themselves can contain private info, but this PR removes that opacity.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Segment filters are displayed to anyone that can see the dashboard filtered by the segment, including for a personal segment applied on a public dashboard
```
aerosol:
```suggestion
- Segment filters are visible to anyone who can view the dashboard with that segment applied, including personal segments on public dashboards
```

assets/js/dashboard/filtering/segments.ts
ukutaht: Ah thanks, I conflated the idea of 'view contents' with 'expand'. My bad!

CHANGELOG.md
apata: Nice!

PR #5936, opened 2025-12-08
Goals with custom props (Stats API queries, funnels)

aerosol:
### Changes

To ease the review pain, it's probably best to exclude whitespace and read commit by commit :skull:

I am mainly interested in review focused on test cases, though any comments are of course appreciated

----

This PR implements the foundation for custom props attached to goals.

While it's a weird mix of two billing features going forward, we should try to avoid confusing "goal's custom props" with "custom props" as a dashboard/breakdown feature. They may or may not share the same key set, but from user's point of view this might be two separate areas of interest.

Discussion threads:
  - origin: https://3.basecamp.com/5308029/buckets/44692882/card_tables/cards/9256699412
  - flesh out: https://3.basecamp.com/5308029/buckets/44692882/card_tables/cards/9358934181

_not implemented_ here:
- any billing constraints
- any UI wiring
- Sites/Plugins API integration

scope:
- migration
  - `custom_props` map column is added to `goals` table
  - revisited unique constraints: display name is unique at all times, and the general "goal configuration" (e.g. page path, scroll threshold, custom props) is enforced unique
    - though we still don't allow the same event name with two different currencies (to avoid falling into revenue goals impl details)
  - migration will be extracted to a separate PR once review is concluded
- an arbitrary map of custom props can be now added to the goal definition (via low level interface only), we'll start of with max 3 items and maybe bump that limit in the future
  - if custom props are attached, a special variant of the goal `JOIN` query is generated
  - custom props conditions are joined using the AND operator
  - if the goal definition does not include custom props, all events matching the goal name will be found, regardless if any custom props were attached during ingestion
  - in case of multi-dimensional breakdowns (e.g. by `event:goal` AND by `event:props:plan`) only events conforming to the goal definition will be matched, so including any custom props conditions
  - funnels can also consist of goals with custom props attached 

### Tests

### Changelog
  **not yet**

### Documentation
  **not yet**

### Dark mode


lib/plausible/stats/goals.ex
```diff
@@ -77,33 +79,50 @@ defmodule Plausible.Stats.Goals do
   def goal_join_data(query) do
     goals = query.preloaded_goals.matching_toplevel_filters
 
-    %{
-      indices: Enum.with_index(goals, 1) |> Enum.map(fn {_goal, idx} -> idx end),
```
aerosol, line 81: each map key has been iterating over goals here; new implementation does it once and then reverses to keep the order

priv/repo/migrations/20251209120138_goals_custom_props.exs
```diff
@@ -0,0 +1,41 @@
...
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  # adds custom_props to goals, but also updates unique constraints to guard for either display name 
+  # or broadly speaking "goal configuration"
+
+  def change do
+    drop(unique_index(:goals, [:user_id, :event_name], name: :goals_event_name_unique))
```
ukutaht: `:user_id` looks suspect here. Shouldn't it be `:site_id`?

`\d goals` in postgres gives this line

```
    "goals_event_name_unique" UNIQUE, btree (site_id, event_name) WHERE event_name IS NOT NULL
```

lib/plausible/stats/goals.ex
ukutaht: Nice!

lib/plausible/stats/sql/query_builder.ex
```diff
@@ -171,15 +171,29 @@ defmodule Plausible.Stats.SQL.QueryBuilder do
...
+        join: goal in Expression.event_goal_join(goal_join_data),
+        hints: "ARRAY",
+        on: true,
+        select_merge: %{
+          ^shortname(query, dimension) => fragment("?", goal)
+        },
+        group_by: goal
+      )
+    else
+      # in the spirit of making it easier to work with goals having custom propos attached,
```
ukutaht: Unifying these code paths looks pretty good to me as well: https://github.com/plausible/analytics/commit/df7e6174b254a9b43e75924a94fc3c3a65fc9726

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
```diff
@@ -0,0 +1,403 @@
...
+          "event_name" => "Purchase",
+          "custom_props" => %{"variant" => "A", "plan" => "premium"}
+        })
+
+      populate_stats(site, [
+        build(:event,
+          name: "Purchase",
+          "meta.key": ["variant", "plan"],
+          "meta.value": ["A", "premium"],
+          user_id: @user_id
```
ukutaht: User ID could be omitted from tests like this where the same user never makes more than 1 event. Or, if we want to be being explicit, at least it should be done consistently so the previous test should also include explicit incrementing user IDs.

In the rest of the test suite the pattern is to omit when not needed, and only add explicit user_ids if we're testing the same user triggering more than 1 event.

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
```diff
@@ -0,0 +1,403 @@
...
+        })
+
+      results = json_response(conn, 200)["results"]
+      assert length(results) == 2
+      assert Enum.member?(results, %{"dimensions" => ["Purchase", "free"], "metrics" => [1, 1]})
+
+      assert Enum.member?(results, %{"dimensions" => ["Purchase", "premium"], "metrics" => [1, 1]})
+    end
+
+    test "empty custom_props map does not filter events", %{conn: conn, site: site} do
```
ukutaht: Seems like a duplicate of test on line 105

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
```diff
@@ -0,0 +1,403 @@
...
+          "filters" => [["is", "event:goal", ["Click"]]]
+        })
+
+      assert json_response(conn, 200)["results"] == [
+               %{"dimensions" => ["2021-01-01"], "metrics" => [1]},
+               %{"dimensions" => ["2021-01-02"], "metrics" => [1]}
+             ]
+    end
+
+    test "different goals with same event name but different custom props are distinguished", %{
```
ukutaht, line 309: The description doesn't seem to match the test setup. Description says 'same event name' but the goals created in the test setup have different event names (`Button Click A` and `Button Click B`).

I like the test itself, I think if the goals both used `Button Click` as the event name, and different display names (`Red Button`, `Blue Button`), I think it would exemplify an important real-world use-case where the same event name with different props can be created as distinct goals with different display names.

ukutaht (commented):
Looks good!

I wonder if props would be easier to work with if we migrated them to a `Map(String, String)` data type in Clickhouse. Those array ops are hard to follow. Also, as far as I can tell, the array ops take `n*m` time for each row where

n = number of props on goal
m = number of props on clickhouse event

so having a hard limit for both is definitely needed. With a map it could be linear time based on the number of props on the goal.

I'm not suggesting to migrate now, just a general note that this could be a good thing to tackle in the future. For now it is what it is. I think the implementation in this PR is perfectly good given the schema.

Had some comments about tests and unifying the macros. Let me know what you think.

lib/plausible/stats/sql/query_builder.ex
aerosol: More or less what I initially had, just wasn't sure if there's any point in producing extra query payload in case e.g. 99% of the goals won't have any custom props

priv/repo/migrations/20251209120138_goals_custom_props.exs
aerosol: ah yes, good catch. irrelevant when it comes to SQL produced I think 7d215aee22

lib/plausible/stats/sql/query_builder.ex
aerosol: Pushed

aerosol: https://github.com/plausible/analytics/pull/5936/commits/6c0d9ff01e8362e0a4d2cd50b0ba5403c848e93c

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
aerosol: 6effcc5812

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
aerosol: 81aff078ae

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_test.exs
aerosol: f05215325e

PR #5940, opened 2025-12-10
Migration: goal custom props + revamped unique constraints 

aerosol:
Cherry picked from #5936


aerosol:
ugh, changing migration only will make the existing tests fail. @ukutaht shall we disable those tests first?

aerosol:
migration went ok on staging

apata (approved):

PR #5941, opened 2025-12-10
Fix broken v1 filters redirect for pages containing special characters

apata:
### Changes

- Fixed direct dashboard links that use legacy dashboard filters containing URL encoded special characters (e.g. character `ê` in the legacy filter `?page=%C3%AA`)

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/util/url-search-params-v1.ts
+++ b/assets/js/dashboard/util/url-search-params-v1.ts
@@ -40,21 +40,22 @@ const LEGACY_URL_PARAMETERS = {
   exit_page: null
 }
 
-function isV1(searchRecord: Record<string, unknown>): boolean {
-  return Object.keys(searchRecord).some(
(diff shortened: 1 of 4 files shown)


assets/js/dashboard/util/url-search-params-v1.ts
```diff
@@ -40,21 +40,22 @@ const LEGACY_URL_PARAMETERS = {
...
+  }
+  return false
 }
 
-function parseSearchRecord(
-  searchRecord: Record<string, unknown>
-): Record<string, unknown> {
-  const searchRecordEntries = Object.entries(searchRecord)
+function parseSearch(searchString: string): Record<string, unknown> {
+  const searchParams = new URLSearchParams(searchString)
```
apata, line 53: The fix is based on how it used to work mid July 2024: https://github.com/plausible/analytics/blob/5a58b5cb99cf8ed16e5753f33ff14ad3f1d84f19/assets/js/dashboard/query.js#L99

PR #5944, opened 2025-12-11
Reapply+bugfix: goals with custom props (#5936)

aerosol:
Reapply #5936 with some anti-nil defences

### DIFF
--- a/priv/repo/migrations/20251211110619_goals_custom_props_default.exs
+++ b/priv/repo/migrations/20251211110619_goals_custom_props_default.exs
@@ -2,8 +2,23 @@ defmodule Plausible.Repo.Migrations.GoalsCustomPropsDefault do
   use Ecto.Migration
 
   def change do
+    drop(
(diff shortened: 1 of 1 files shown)


ukutaht:
Is there any meaningful difference between custom props being `nil` vs `%{}`?

I see that new goals will default to empty map when created via Elixir, but the same default is not set in postgres. I don't know, just having the same state represented in two different ways creates the need to account for both which seems unnecessary detail to have to deal with. Why don't we set the field to `custom_props JSONB NOT NULL DEFAULT '{}'::jsonb` and only deal with one representation of "no props configured" in the code?

Sorry I missed this in review as well :(

aerosol:
I think the difference is how  unique constraints are firing. I dislike nullable columns too. Let me try to run this on top of another migration then

aerosol:
@ukutaht done, but we have to push https://github.com/plausible/analytics/pull/5945 first

aerosol:
Done via #5946

PR #5947, opened 2025-12-13
Upgrade clickhouse server

cnkk:


Makefile
apata:
```suggestion
CH_FLAGS ?= --detach -p 8123:8123 -p 9000:9000 --ulimit nofile=262144:262144 --name plausible_clickhouse --env CLICKHOUSE_SKIP_USER_SETUP=1
```
Skipping user setup is needed here as well

PR #5952, opened 2025-12-15
Expose goals w/ custom props through Sites/Plugins API

aerosol:
This PR implements goals provisioning/indexing with (optional) custom props to Plugins API and Sites API. Plugins API OpenAPI schema will be updated automatically (preview: https://gist.github.com/aerosol/b0df1c0ceecda6a3b0121372c75ea4ee). Sites API docs change pending once the feature is fully rolled out (i.e. with the UI).

TODO:
 - [x] verify if custom props feature is available (?)

### DIFF
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -402,8 +402,6 @@ defmodule Plausible.Goals do
   end
 
   defp custom_props_goals_access_check(site, changeset) do
-    # leaky, even though Goals and Props are separate features,
-    # in case of goals equipped with custom props, we need to check this
     if map_size(Changeset.get_field(changeset, :custom_props)) > 0 do
       site = Plausible.Repo.preload(site, :team)
       Plausible.Billing.Feature.Props.check_availability(site.team)


lib/plausible_web/plugins/api/schemas/goal/create_request/custom_event.ex
```diff
@@ -20,7 +20,13 @@ defmodule PlausibleWeb.Plugins.API.Schemas.Goal.CreateRequest.CustomEvent do
...
         required: [:event_name],
         properties: %{
-          event_name: %Schema{type: :string}
+          event_name: %Schema{type: :string},
+          custom_props: %Schema{
+            type: :object,
+            description: "Custom properties (max 3, string keys and values)",
+            additionalProperties: %Schema{type: :string},
+            maxProperties: 3
+          }
```
apata: question, non-blocking: is it possible not to repeat this in the three schemas?

test/plausible_web/plugins/api/controllers/goals_test.exs
```diff
@@ -110,6 +110,147 @@ defmodule PlausibleWeb.Plugins.API.Controllers.GoalsTest do
...
+
+      [goal] = Plausible.Goals.for_site(site)
+      assert goal.event_name == "Signup"
+      assert goal.custom_props == %{"tier" => "premium", "source" => "landing"}
+
+      first_goal = List.first(resp["goals"])
+      assert first_goal["goal"]["custom_props"] == %{"tier" => "premium", "source" => "landing"}
+    end
+
+    test "creates a goal with empty custom_props by default", %{
```
apata: nitpick, non-blocking: it seems this case could be put into the existing `test "creates a custom event goal"`. We could then use assert_matches to assert the exact shape of the response.

apata:
@aerosol Regarding the todo, are we OK with some besides us already inserting goals like this?

aerosol:
> @aerosol Regarding the todo, are we OK with some besides us already inserting goals like this?

In general it's not the end of the world if some do, those goals are handled at the query level. It's just the feature isn't exposed in the settings UI just yet.

apata (approved):

lib/plausible_web/plugins/api/schemas/goal/create_request/custom_event.ex
aerosol: fe60bcc4f3

test/plausible_web/plugins/api/controllers/goals_test.exs
aerosol: https://github.com/plausible/analytics/pull/5952/commits/6488f2c3ff50ad0b6ea2a25753c79352cfaff617

apata: Hmm, there's still two tests that are quite similar to each other, one is "creates a custom event goal with empty custom_props by default" and the other is "creates a custom event goal". In the other, we assert the name only currently, but if we asserted empty custom props as well, we wouldn't need the test "creates a custom event goal with empty custom_props by default"

aerosol:
Added custom props feature checks via https://github.com/plausible/analytics/pull/5952/commits/6434f62317b6a2cdc20aa81e69aa991e1b214860

test/plausible_web/plugins/api/controllers/goals_test.exs
aerosol: Cleaned up that test module, should make sense now. Could you please look again? :eagle:

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
```diff
@@ -413,6 +414,12 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
 
         H.bad_request(conn, message)
 
+      {:error, :upgrade_required} ->
+        H.payment_required(
+          conn,
+          "Custom Properties is part of the Plausible Business plan. To get access to this feature, please upgrade your account."
```
apata: question, non-blocking: why did we use a different message from `"Your current subscription plan does not include Custom Properties"`?

lib/plausible/goals/goals.ex
```diff
@@ -395,6 +401,17 @@ defmodule Plausible.Goals do
     end
   end
 
+  defp custom_props_goals_access_check(site, changeset) do
+    # leaky, even though Goals and Props are separate features,
+    # in case of goals equipped with custom props, we need to check this
```
apata: nitpick, non-blocking: Unclear what is leaky and if still leaky

aerosol: leaky abstraction, we now have a hybrid feature, consisting of two `Billing.Feature`s. I dislike that, but it makes sense from CS point of view.

extra/lib/plausible_web/controllers/api/external_sites_controller.ex
aerosol: Ah, good catch, copied that message from a random grep instead of looking at module at hand :clown_face: ba94c46b8b

test/plausible_web/controllers/api/external_sites_controller_test.exs
```diff
@@ -665,6 +663,102 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
...
+      test "custom_props defaults to empty map if not provided", %{conn: conn, site: site} do
+        conn =
+          put(conn, "/api/v1/sites/goals", %{
+            site_id: site.domain,
+            goal_type: "event",
+            event_name: "Signup"
+          })
+
+        res = json_response(conn, 200)
+        assert res["custom_props"] == %{}
```
apata, line 706: nitpick, non-blocking: today, custom props on a goal is a new thing and this looks like an important test, but I suspect a month from now we'll be wondering why we don't test the default value of this attribute under existing tests "can add a goal as event to a site", "can add a goal as page to a site". We can use assert_matches to completely assert what happens, with some examples on what I personally like in PlausibleWeb.Api.ExternalSitesControllerSitesCrudApiTest

test/support/teams/test.ex
```diff
@@ -248,21 +248,35 @@ defmodule Plausible.Teams.Test do
     user
   end
 
-  def subscribe_to_growth_plan(user, attrs \\ []) do
+  def subscribe_to_growth_plan(user_or_team, attes \\ [])
```
apata:
```suggestion
  def subscribe_to_growth_plan(user_or_team, attrs \\ [])
```

aerosol: oof, thanks

lib/plausible/goals/goals.ex
apata: Thanks! I understand what you mean, it's potentially some debt for the future. Could we not wire it up so it's flat on the BE `F.Goals, F.Props, F.GoalsWithCustomProps`, but still maintain only the two F.Goals, F.Props toggles for the CS? We'd hard-code it to be added into the features list at the earliest opportunity when both F.Goals and F.Props are true.

test/plausible_web/controllers/api/external_sites_controller_test.exs
aerosol: [aa20327](https://github.com/plausible/analytics/pull/5952/commits/aa2032755f8f7a3ff8023a6cd7db3dca6c7072fd)

test/plausible_web/controllers/api/external_sites_controller_test.exs
```diff
@@ -461,12 +461,17 @@ defmodule PlausibleWeb.Api.ExternalSitesControllerTest do
...
 
-        res = json_response(conn, 200)
+        resp = json_response(conn, 200)
 
-        assert res["goal_type"] == "event"
-        assert res["display_name"] == "Signup"
-        assert res["event_name"] == "Signup"
-        assert res["domain"] == site.domain
+        assert_matches ^strict_map(%{
+                         "custom_props" => %{},
```
apata: It will still match all kinds of custom_props unless we do 
```
        assert_matches ^strict_map(%{
                         "custom_props" => ^strict_map(%{}),
```

lib/plausible/goals/goals.ex
aerosol: > We'd hard-code it to be added into the features list at the earliest opportunity when both F.Goals and F.Props are true.

Can you be more specific? The "features list" (`Feature.list/0`) is used in multiple places where we don't want this hybrid feature to be included.

aerosol: I think we could do:

```diff
diff --git a/lib/plausible/billing/feature.ex b/lib/plausible/billing/feature.ex
index 69554d638e..a13e5d4af2 100644
--- a/lib/plausible/billing/feature.ex
+++ b/lib/plausible/billing/feature.ex
@@ -254,3 +254,22 @@ defmodule Plausible.Billing.Feature.Teams do
     end
   end
 end
+
+defmodule Plausible.Billing.Feature.GoalsWithCustomProps do
+  @moduledoc "This feature is a hybrid state consisting of two separate features and is not toggleable on its own"
+  use Plausible.Billing.Feature,
+    name: :goals_with_custom_props,
+    display_name: "Goals With Custom Properties"
+
+  @impl true
+  def check_availability(team_or_nil) do
+    allowed_features = Plausible.Teams.Billing.allowed_features_for(team_or_nil)
+
+    if Plausible.Billing.Feature.Goals in allowed_features and
+         Plausible.Billing.Feature.Props in allowed_features do
+      :ok
+    else
+      {:error, :upgrade_required}
+    end
+  end
+end
diff --git a/lib/plausible/goals/goals.ex b/lib/plausible/goals/goals.ex
index 8d92b301d7..d48ff2d77f 100644
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -402,11 +402,9 @@ defmodule Plausible.Goals do
   end
 
   defp custom_props_goals_access_check(site, changeset) do
-    # leaky, even though Goals and Props are separate features,
-    # in case of goals equipped with custom props, we need to check this
     if map_size(Changeset.get_field(changeset, :custom_props)) > 0 do
       site = Plausible.Repo.preload(site, :team)
-      Plausible.Billing.Feature.Props.check_availability(site.team)
+      Plausible.Billing.Feature.GoalsWithCustomProps.check_availability(site.team)
     else
       :ok
     end
```

but I'm not sure if there are any hidden consequences of this

test/plausible_web/controllers/api/external_sites_controller_test.exs
aerosol: [00f1322](https://github.com/plausible/analytics/pull/5952/commits/00f132298276debfdebff5a731eccbbd51f10870)

test/plausible_web/plugins/api/controllers/goals_test.exs
apata: Yep, it definitely makes sense now and it made sense before as well. I think the idea that I was trying to get across was similar to https://github.com/plausible/analytics/pull/5952#discussion_r2622273216. That we could save on test count by asserting a bit more in existing tests: https://github.com/plausible/analytics/compare/goals-custom-props-plugins-sites-apis...example-123?expand=1 
I don't want to enforce this preference though, just wanted to share it

aerosol: No, we really don't need to do that in Plugins API tests, since a lot of schema enforcement comes with `open_api_spex` for free. And you have interactive docs at http://localhost:8000/api/plugins/spec/openapi. Try returning a payload not conforming to the schema and that test will break.

lib/plausible/goals/goals.ex
aerosol: Edited the diff above. I'd like @ukutaht to chime in

apata: > Can you be more specific? The "features list" (Feature.list/0) is used in multiple places where we don't want this hybrid feature to be included


Oof, I didn't consider that at all. That diff looks good though

aerosol: pushed https://github.com/plausible/analytics/pull/5952/commits/3d5d7cf4c89355269898a57e8e20d55fa6f31734

ukutaht: Sorry for not replying earlier.

Hmm from a billing perspective it's only Props that we are interested in. Goals are always available, they are never feature-gated by subscription plan. Goals only exists as a 'Feature' to make it work with the toggling abstraction.

The way I intuit the system, I liked the original diff more. In the new diff this line:

` if Plausible.Billing.Feature.Goals in allowed_features and Plausible.Billing.Feature.Props in allowed_features do`

has the potential to confuse the reader because it looks like goals can be missing from the 'allowed_features' list which is not true.

To me the leakyness comes from the fact that in order to toggle Goals we lump it in with other billing features when in fact it has nothing to do with billing.

ukutaht: Another reason Goals are in the billing module is because they're listed in the upgrade/downgrade page as part of Feature.list().

But Goals might as well be listed as static plaintext on the change-plan page and be completely excluded from any dynamic billing context. Listing them is not actually conditional in any way like other billing features.

This reinforces my view that it's the conflation of billing with toggleability that is causing weirdness, not that goals-with-props is a new kind of hybrid feature from a billing perspective.

aerosol: Thanks @ukutaht - reverted.

PR #5954, opened 2025-12-16
Allow deep linking to password protected shared links

apata:
### Changes

Adds return to param for deep linking to password protected shared links
 
### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/stats_controller_test.exs
+++ b/test/plausible_web/controllers/stats_controller_test.exs
@@ -1667,7 +1667,8 @@ defmodule PlausibleWeb.StatsControllerTest do
         %{password: "password"}
       )
 
-    assert redirected_to(conn, 302) == "/share/#{URI.encode_www_form(site.domain)}#{deep_path}?auth=#{link.slug}&#{filters}"
+    assert redirected_to(conn, 302) ==
+             "/share/#{URI.encode_www_form(site.domain)}#{deep_path}?auth=#{link.slug}&#{filters}"
   end
 
   test "return_to from query_params is discarded", %{conn: conn} do
@@ -1697,7 +1698,8 @@ defmodule PlausibleWeb.StatsControllerTest do
         %{password: "password"}
       )
 
-    assert redirected_to(conn, 302) == "/share/#{URI.encode_www_form(site.domain)}/pages?auth=#{link.slug}"
+    assert redirected_to(conn, 302) ==
+             "/share/#{URI.encode_www_form(site.domain)}/pages?auth=#{link.slug}"
   end
 
   test "return_to doesn't allow navigating out of dashboard context", %{conn: conn} do
@@ -1730,7 +1732,8 @@ defmodule PlausibleWeb.StatsControllerTest do
         %{password: "password"}
       )
 
-    assert redirected_to(conn, 302) == "/share/#{URI.encode_www_form(site.domain)}#{cleaned_deep_path}?auth=#{link.slug}&theme=dark"
+    assert redirected_to(conn, 302) ==
+             "/share/#{URI.encode_www_form(site.domain)}#{cleaned_deep_path}?auth=#{link.slug}&theme=dark"
   end
 
   describe "dogfood tracking" do


test/plausible_web/uri_encoding_test.exs
```diff
@@ -0,0 +1,16 @@
+defmodule PlausibleWeb.URIEncodingTest do
+  use ExUnit.Case, async: true
+
+  for {value, expected} <- [
+        {"/:dashboard/settings", "/:dashboard/settings"},
+        {"&", "%26"},
+        {"=", "%3D"},
+        {",", "%2C"},
+        # should be {"hello world", "hello%20world"}, is
+        {"hello world", "hello+world"},
```
apata: @RobertJoonas I wanted to bring this to your attention: according to this test, DashboardQuerySerializer isn't yet quite the same as the JS one. It's serializing space as `+` not the URI encoded `%20`. Dashboard ATM parses `+` value as `+`.

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -370,12 +394,47 @@ defmodule PlausibleWeb.StatsController do
...
+      String.split(trimmed_return_to_value, "/")
+    else
+      []
+    end
+  end
+
+  defp get_rest_of_query_string(conn)
+       when is_nil(conn.query_string) or
+              (is_binary(conn.query_string) and byte_size(conn.query_string)) == 0,
+       do: ""
```
zoldar:
```suggestion
  defp get_rest_of_query_string(conn) when conn.query_string in [nil, ""], do: ""
```

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -370,12 +394,47 @@ defmodule PlausibleWeb.StatsController do
...
 
-  def qs_appendix(conn), do: "&#{conn.query_string}"
+  defp get_rest_of_query_string(conn), do: "&#{conn.query_string}"
```
zoldar, line 430:
```suggestion
  defp get_rest_of_query_string(conn), do: conn.query_string
```

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -370,12 +394,47 @@ defmodule PlausibleWeb.StatsController do
...
-      when is_nil(conn.query_string) or
-             (is_binary(conn.query_string) and byte_size(conn.query_string)) == 0,
-      do: ""
+  defp serialize_star_path_as_query_string_fragment(star_path) do
+    if length(star_path) > 0 do
+      # make the path start with a /
+      # to be able to reject values that don't start with a /
+      serialized_value =
+        "/#{Enum.join(star_path, "/")}"
+        |> PlausibleWeb.URIEncoding.uri_encode_permissive()
```
zoldar: To my understanding, this is the only reason for workarounds of URL encoding, right?

I'm also hesitant about this, given it's purely for aesthetics? If the shared link is embedded in any 3rd party app where, for instance templating/rendering engine does any sort of sanitization/validation, there is a higher chance that something is going to break?

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -370,12 +394,47 @@ defmodule PlausibleWeb.StatsController do
...
+  defp parse_star_path(conn) do
+    return_to_value = conn.query_params["return_to"]
+
+    if not is_nil(return_to_value) and String.starts_with?(return_to_value, "/") do
+      # get rid of the / character that we added
+      "/" <> trimmed_return_to_value = return_to_value
+      String.split(trimmed_return_to_value, "/")
+    else
+      []
+    end
```
zoldar:
```suggestion
    case conn.query_params["return_to"] do
      "/" <> return_to -> String.split(return_to, "/")
      _ -> []
    end
```

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -349,12 +360,25 @@ defmodule PlausibleWeb.StatsController do
       if Plausible.Auth.Password.match?(password, shared_link.password_hash) do
         token = Plausible.Auth.Token.sign_shared_link(slug)
 
+        star_path = parse_star_path(conn)
+
+        query_string_fragment =
+          get_rest_of_query_string(conn)
```
zoldar, line 369: Given how `get_rest_of_query_string` works,

```
(conn.query_string || "")
```

would be equivalent, and easier to follow IMHO

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -349,12 +360,25 @@ defmodule PlausibleWeb.StatsController do
...
+          |> omit_from_query_string("return_to")
+          |> omit_from_query_string("auth")
```
zoldar, line 373: `drop_query_param` perhaps?

lib/plausible_web/controllers/stats_controller.ex
```diff
@@ -349,12 +360,25 @@ defmodule PlausibleWeb.StatsController do
       if Plausible.Auth.Password.match?(password, shared_link.password_hash) do
         token = Plausible.Auth.Token.sign_shared_link(slug)
 
+        star_path = parse_star_path(conn)
```
zoldar, line 364: Passing `conn.query_params["return_to"]` would be easier to follow IMHO

test/plausible_web/controllers/stats_controller_test.exs
```diff
@@ -1578,7 +1578,7 @@ defmodule PlausibleWeb.StatsControllerTest do
         )
 
       conn = post(conn, "/share/#{link.slug}/authenticate", %{password: "password"})
-      assert redirected_to(conn, 302) == "/share/#{site.domain}?auth=#{link.slug}"
+      assert redirected_to(conn, 302) == "/share/#{site.domain}/?auth=#{link.slug}"
```
ukutaht, line 1581: This looks strange to me, why the extra slash?

apata: True, it's a bit strange! 

Seems like the [autogenerated route helpers](https://hexdocs.pm/phoenix/Phoenix.Router.html#module-helpers-deprecated) (which are deprecated) work like this under the hood. I didn't see the point in fighting them.

```
iex(16)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :shared_link, "hi.world", [])
"/share/hi.world/"
iex(17)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :stats, "hi.world", [])
"/hi.world/"
```

ukutaht: Fair, I think it comes from catch-all addition

lib/plausible_web/controllers/stats_controller.ex
apata: Thanks! I considered it, but we're operating on query string not the query params map, so omit seemed more appropriate after all.

lib/plausible_web/controllers/stats_controller.ex
apata: This had to stay in, I've tried to add comments on why in https://github.com/plausible/analytics/pull/5954/commits/c300fa9cb0e155ffdd019ee243dcc6e8a36c4ec2

lib/plausible_web/controllers/stats_controller.ex
apata: Thanks for pointing it out! It was not for URL aesthetics, it was code aesthetics (wanted to match on `"/" <> foo`), which actually turned out not to be necessary. I've made sure it's encoded appropriately now and removed the unnecessary module.

lib/plausible_web/controllers/stats_controller.ex
apata: Thanks, it does look a little weird! I wanted `parse_...` and `serialize_...` to operate with the same type of input to be clearly related and opposites of each other, that's why I'm using the conn as the arg.

lib/plausible_web/controllers/stats_controller.ex
zoldar: Right, I've missed `&` 🙈

PR #5955, opened 2025-12-16
Add comparisons to DashboardQueryParser

RobertJoonas:
### Changes

Main functional objectives:

1) Add the ability to parse comparison params from the dashboard URL search params into ParsedQueryParams. And serialize it back to a query string afterwards
2) Introduce an `add_or_replace_filter` function into `ParsedQueryParams` (to be used for building dashboard filter links)

This PR also includes some major refactoring work:

1) Introduce a QueryInclude structure (replacing the simple map used before)
2) Change how comparison params are represented under ParsedQueryParams
3) Get rid of the query-internal-test endpoint and move/rebuild several tests, building and running the query inside the test directly (instead of going through an API layer)

### Reasoning of 3)

The API v2 project was focused on minimizing the gap between public and internal Stats APIs using the same schema for both (with an optional `"only :internal"` flag). That would have been great if we would have built a more powerful React frontend on top of API v2.

However, with the recent plans to migrate the dashboard to LiveView, we only need to construct queries in Elixir code, and therefore, the "internal API schema" will become irrelevant and the long term plan is to get rid of it. This PR is the first step towards that but there's still a way to go with that.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query_include.ex
+++ b/lib/plausible/stats/query_include.ex
@@ -11,7 +11,7 @@ defmodule Plausible.Stats.QueryInclude do
             legacy_time_on_page_cutoff: nil
 
   @type date_range_tuple() :: {:date_range, Date.t(), Date.t()}
-  @type datetime_range_tuple() :: {:date_range, DateTime.t(), DateTime.t()}
+  @type datetime_range_tuple() :: {:datetime_range, DateTime.t(), DateTime.t()}
 
   @type t() :: %__MODULE__{
           imports: boolean(),


lib/plausible/stats/api_query_parser.ex
```diff
@@ -275,35 +261,43 @@ defmodule Plausible.Stats.ApiQueryParser do
   defp parse_order_direction(entry), do: {:error, "Invalid order_by entry '#{i(entry)}'."}
 
   def parse_include(include) when is_map(include) do
-    with {:ok, include} <- atomize_include_keys(include),
-         {:ok, include} <- parse_comparison_date_range(include) do
-      {:ok, Map.merge(@default_include, include)}
+    parsed_include_params_or_error =
+      include
+      |> Map.to_list()
```
zoldar: I think this is redundant as maps are already enumerable.

lib/plausible/stats/query_include.ex
```diff
@@ -0,0 +1,27 @@
...
+            imports_meta: false,
+            time_labels: false,
+            total_rows: false,
+            trim_relative_date_range: false,
+            compare: nil,
+            compare_match_day_of_week: false,
+            legacy_time_on_page_cutoff: nil
+
+  @type date_range_tuple() :: {:date_range, Date.t(), Date.t()}
+  @type datetime_range_tuple() :: {:date_range, DateTime.t(), DateTime.t()}
```
zoldar: Weird, isn't that supposed to be `:datetime_range`? 🤔

lib/plausible/stats/api_query_parser.ex
RobertJoonas: https://github.com/plausible/analytics/pull/5955/commits/49362e01149b9c02eb4e3858c7b84cd768ce50cd

lib/plausible/stats/query_include.ex
RobertJoonas: Oh, yes. That's a typo. Fixed in https://github.com/plausible/analytics/pull/5955/commits/7ff654b5c20c150f37ac403ca518c993658cb2f9

Issue #5901, opened 2025-11-22
Exception for new tracking code


hatamiarash7:
### Past Issues Searched

- [x] I have searched open and closed issues to make sure that the bug has not yet been reported

### Issue is a Bug Report

- [x] This is a bug report and not a feature request, nor asking for self-hosted support

### Using official Plausible Cloud hosting or self-hosting?

Self-hosting

### Describe the bug

For newly defined sites, the new dedicated tracking script does not work, and Plausible is crashing:

### Expected behavior

Old tracking scripts are ok when migrating to the new one. But newly defined sites are not.

### Screenshots

<img width="2398" height="484" alt="Image" src="https://github.com/user-attachments/assets/e5c79265-4e78-4355-8e7a-c153a5321a6c" />

<img width="688" height="357" alt="Image" src="https://github.com/user-attachments/assets/72500b1f-2254-48a0-95da-1c55b326fc41" />

### Environment

```markdown
- OS: Ubuntu 22.04.4 LTS
- Browser: Firefox
- Browser Version: 145.0.1 (64-bit)
- Plausible Version: v3.1.0
```

ItalyPaleAle:
I am having the same issue, except for me it's newly-created sites that don't work, but old ones do. Restarting the container seems to fix the issue too.

apata:
Thanks for your report @hatamiarash7! Thanks for sharing your experience @ItalyPaleAle! 

I'm looking into this right now and it doesn't reproduce on my locally running CE 3.1.0 instance. 
Can you share the full server error (omitting anything sensitive)? 
What are the steps you use to create a new site?

shalzz:
Hi, here's the server log I have from when trying to load the script in browser:
This server error occurred for me after upgrading to the 3.1.0 release but only for the first site I created after the upgrade, any subsequent sites created on plausible, even for the same domain, works fine with serving the script. 

```
plausible-1            | 07:28:01.511 [error] #PID<0.81964.0> running PlausibleWeb.Endpoint (connection #PID<0.81963.0>, stream id 1) terminated
plausible-1            | Server: <redacted>:80 (http)
plausible-1            | Request: GET /js/pa-4H954SisddMuXrtxBMSRL.js
plausible-1            | ** (exit) an exception was raised:
[...]

PR #5956, opened 2025-12-16
Fix CE tracker script cache issue

apata:
### Changes

Fixes https://github.com/plausible/analytics/issues/5901
Separates EE and CE script caches.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
+++ b/extra/lib/plausible_web/controllers/api/external_sites_controller.ex
@@ -98,7 +98,8 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
               display_name: goal.display_name,
               goal_type: Goal.type(goal),
               event_name: goal.event_name,
-              page_path: goal.page_path
+              page_path: goal.page_path,
+              custom_props: goal.custom_props
             }
           end),
         meta: pagination_meta(page.metadata)
@@ -412,6 +413,12 @@ defmodule PlausibleWeb.Api.ExternalSitesController do
         message = Enum.map_join(changeset.errors, ", ", &translate_error/1)
 
         H.bad_request(conn, message)
+
+      {:error, :upgrade_required} ->
+        H.payment_required(
+          conn,
+          "Your current subscription plan does not include Custom Properties"
+        )
 
       e ->
         H.bad_request(conn, "Something went wrong: #{inspect(e)}")
--- a/lib/plausible/goal.ex
+++ b/lib/plausible/goal.ex
@@ -54,13 +54,7 @@ defmodule Plausible.Goal do
     |> validate_event_name_and_page_path()
     |> validate_page_path_for_scroll_goal()
     |> maybe_put_display_name()
-    |> validate_change(:custom_props, fn :custom_props, custom_props ->
-      if map_size(custom_props) > @max_custom_props_per_goal do
-        [custom_props: "use at most #{@max_custom_props_per_goal} properties per goal"]
-      else
-        []
-      end
-    end)
+    |> validate_change(:custom_props, &validate_custom_props/2)
     |> unique_constraint(:display_name, name: :goals_display_name_unique)
     |> unique_constraint(:event_name, name: :goals_event_config_unique)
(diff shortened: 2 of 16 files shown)


cnkk:
Not 100% sure about why the test does not fail, but there's a difference
https://github.com/plausible/analytics/blob/c0397da66329d598a34f8d8c72607a97ce2b193d/test/plausible_web/tracker_script_cache_test.exs#L11 
it does not contain the `ets_options`

vs.
https://github.com/plausible/analytics/blob/c0397da66329d598a34f8d8c72607a97ce2b193d/lib/plausible/application.ex#L167

compared to e.g.
https://github.com/plausible/analytics/blob/c0397da66329d598a34f8d8c72607a97ce2b193d/test/plausible/shield/hostname_rule_cache_test.exs#L16
https://github.com/plausible/analytics/blob/c0397da66329d598a34f8d8c72607a97ce2b193d/lib/plausible/application.ex#L139

aerosol:
@cnkk spot on

test/plausible/cache_test.exs
```diff
@@ -124,6 +124,16 @@ defmodule Plausible.CacheTest do
                {ExampleCache.get("item1", cache_name: test, force?: true) == :item1, :ok}
              end)
     end
+
+    test "broadcast_put works as expected when called two times for a string value", %{test: test} do
```
aerosol: pointless to test broadcast put, if what you mean is put. 

but also since this behaviour depends on ets options, I'm not sure what's the point again. are we testing if ets work? are we testing if concache works? are we testing if erlang's rpc works?

we don't provide any instructions for self-hosters to run a cluster in which they can broadcast function calls over rpc.

test/plausible_web/tracker_script_cache_test.exs
```diff
@@ -8,7 +8,14 @@ defmodule PlausibleWeb.TrackerScriptCacheTest do
     test "cache caches tracker scripts by id", %{test: test} do
       {:ok, _} =
         Supervisor.start_link(
-          [{TrackerScriptCache, [cache_name: test, child_id: :test_cache_tracker_script]}],
+          [
+            {TrackerScriptCache,
+             [
+               cache_name: test,
+               child_id: :test_cache_tracker_script,
+               ets_options: [:bag, read_concurrency: true]
```
aerosol: you only want :bag here, unless :bag is not what you want - if you care about overwriting keys, don't use `:bag` in application startup. `read_concurrency` is completely pointless to keep here.

test/plausible_web/tracker_script_cache_test.exs
```diff
@@ -36,6 +43,64 @@ defmodule PlausibleWeb.TrackerScriptCacheTest do
...
+        TrackerScriptCache.cache_content(%TrackerScriptConfiguration{
+          updated_config
+          | file_downloads: false
+        }),
+        cache_opts
+      )
+
+      on_ce do
+        # wait for broadcast put to take effect
+        assert eventually(fn ->
```
aerosol: just use `Adapter.put`. there's nothing special about broadcast_put, nothing of relevance - what it does is tested elsewhere.

test/plausible_web/tracker_script_cache_test.exs
apata: Thanks! It's `:bag` with `read_concurrency: true` because I want the test setup to have parity with the real implementation. I don't understand yet if :bag is appropriate for the real implementation.

aerosol: Read concurrency is irrelevant...

test/plausible_web/tracker_script_cache_test.exs
apata: Thanks! We've discussed this when it was merged as well: I wanted broadcast_put to avoid more code branching between CE and EE. I can refactor to put directly once we figure out what's going wrong with put.

test/plausible/cache_test.exs
apata: This was just to make sure that if all those things configured as they are in the test, it works as expected.

PR #5957, opened 2025-12-17
Fix permissions for guests

ukutaht:
### Changes

https://github.com/plausible/analytics/pull/5874/commits/beefacdb80357ba40075814ed28297adab111b8f introduced an unintended change to permissions. Before that commit guest editors were able to view and edit funnels, props and shared links. This PR reverts back to the original behaviour by using `@site_team` over `@current_team` in settings tiles that are rendered in site settings.

See also [bug report](https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/9330602066).

### Tests

### Changelog
Not adding since I don't think this was released in CE (?)

### Dark mode


aerosol (approved):

aerosol:
#5970 should help

ukutaht:
@aerosol Thanks. I'm still struggling with postgres connection issues in CI. Do you have any pointers where to look based on your experience?

aerosol:
Oh no. Anytime we spawn a process basically (and where tests doing that directly or indirectly run in async mode). So look for either Task.start and such and/or LiveViews receiving messages - if the process queries Postgres, it's likely to do that after the main test process finishes, resulting with disconnection log.

aerosol:
@ukutaht I'm getting that locally as well for this branch specifically, so must be something here. I'll have a look tomorrow.

aerosol:
@ukutaht I'm glad we finally have a reproducible one. I find the connection sandboxing quite hard to grok.

```
mix test test/plausible_web/live/funnel_settings_test.exs --warnings-as-errors --seed 78639
```

This logs the errors consistently for me. With `--trace` even. It becomes clear that some two tests are involved.

Some, but not necessarily the ones, that are printed right before the error is logged (otherwise that'd suggest "premium feature notice renders" is at fault - but that one does static render, no LV involved!). It's weird, I thought that turning that whole test to `async: false` will help dealing with the Sandbox ownership and wind down each case properly. 

It seems that, the difference between `@current_team` and `@site_team` is subscription preload (the former has it already, the latter asks for it during feature availability check), and the error is related to LiveView async rendering. 

What's tricky here - the funnel settings LV is spawning funnel form LV (not a Live Compnent). So somehow, a LV sub-process made with `live_render` is still alive shortly after the test finishes, asks for feature availability check, no preload exists, so postgres connection is used, but the connection owner is down already.

So we have two ways of solving it:

1. Ensure subscription is preloaded (with `last_susbscription_query()`) at `Sites.get_for_user!`. If we do that, would be good to do the same in `Sites.get_for_user` (and remove the unnecessary preload at `ExternalSitesController.find_site` then). It's not a bad idea in general, but won't prevent us from hitting similar issues elsewhere.

2. At first I started looking at https://github.com/phoenixframework/phoenix_ecto/blob/main/lib/phoenix_ecto/sql/sandbox.ex - but this seems applicable for Wallaby-style acceptance tests. In our case we don't use websocket connections directly, but probably still need something similar when it comes to sandbox ownership management. But that I haven't figured out yet, will keep you posted.

Summoning @zoldar for a dive :D

image

aerosol:
Seems like this makes things better:

```diff
diff --git a/extra/lib/plausible_web/live/funnel_settings/form.ex b/extra/lib/plausible_web/live/funnel_settings/form.ex
index e273bd1400..5a4b6115bf 100644
--- a/extra/lib/plausible_web/live/funnel_settings/form.ex
+++ b/extra/lib/plausible_web/live/funnel_settings/form.ex
@@ -226,7 +226,7 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
     step_ids = List.delete(socket.assigns.step_ids, idx)
     selections_made = drop_selection(socket.assigns.selections_made, idx)
 
-    send(self(), :evaluate_funnel)
+    if connected?(socket), do: send(self(), :evaluate_funnel)
 
     {:noreply,
      assign(socket, step_ids: step_ids, selections_made: selections_made, funnel_modified?: true)}
@@ -293,7 +293,7 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
   def handle_info({:selection_made, %{submit_value: goal_id, by: combo_box}}, socket) do
     selections_made = store_selection(socket.assigns, combo_box, goal_id)
 
-    send(self(), :evaluate_funnel)
+    if connected?(socket), do: send(self(), :evaluate_funnel)
 
     {:noreply,
      assign(socket,
diff --git a/lib/plausible/sites.ex b/lib/plausible/sites.ex
index b1343a16cb..74ab3ab215 100644
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -467,8 +467,7 @@ defmodule Plausible.Sites do
         |> Repo.one!()
       end
 
-    # Repo.preload(site, team: [subscription: Plausible.Teams.last_subscription_query()])
-    Repo.preload(site, :team)
+    Repo.preload(site, team: [subscription: Plausible.Teams.last_subscription_query()])
   end
 
   def get_for_user(user, domain, opts \\ []) do
```

but additionally, I'd also make that preload in `get_for_user`, for parity

aerosol:
I give up. https://github.com/plausible/analytics/pull/6008

PR #5958, opened 2025-12-17
Update PostgreSQL image version to 18 for non CE

cnkk:


aerosol (approved):
Awesome

aerosol:
@cnkk shall we modify github action runners to use 18 only btw?

PR #5960, opened 2025-12-17
Ensure `put_many/2` erases existing keys from `:bag` type ets caches

aerosol:
It's a fun one, discovered thanks to @apata's curiosity!

Some caches utilize `:bag` table type - so we can store multiple values under one key (example is hostname shield rules). Those are reduced behind the scenes, and multiple values are properly returned.

When a cache refresh cycle comes, during the merge phase, we first insert all the new items via `put_many/2`, then look if any of the old items are stale so they're optionally removed. Unfortunately that wasn't enough for `:bag` caches - since the deduplication was performed on key-basis for performance reasons (we try not to iterate fully over potentially large tables).

This patch ensures that `put_many/2` becomes aware of table types allowing duplicate keys and wipes given keys before insertion.

### DIFF
--- a/lib/plausible/cache/adapter.ex
+++ b/lib/plausible/cache/adapter.ex
@@ -102,9 +102,11 @@ defmodule Plausible.Cache.Adapter do
     |> Enum.group_by(fn {key, _} -> {key, get_name(cache_name, key)} end)
     |> Enum.each(fn {{key, full_cache_name}, items} ->
       ets = ConCache.ets(full_cache_name)
+
       if :ets.info(ets, :type) in [:bag, :duplicate_bag] do
         :ets.delete(ets, key)
       end
+
       true = :ets.insert(ets, items)
     end)
 
--- a/test/plausible/cache/adapter_test.exs
+++ b/test/plausible/cache/adapter_test.exs
@@ -44,7 +44,7 @@ defmodule Plausible.Cache.AdapterTest do
 
       {:ok, _} =
         Supervisor.start_link(
-          Adapter.child_specs(name, name, [ets_options: [:bag]]),
+          Adapter.child_specs(name, name, ets_options: [:bag]),
           strategy: :one_for_one,
           name: :"cache_supervisor_#{test}"
         )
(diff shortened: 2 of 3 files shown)


lib/plausible/cache/adapter.ex
```diff
@@ -99,9 +99,15 @@ defmodule Plausible.Cache.Adapter do
...
   def put_many(cache_name, items) when is_list(items) do
     items
-    |> Enum.group_by(fn {key, _} -> get_name(cache_name, key) end)
-    |> Enum.each(fn {full_cache_name, items} ->
-      true = :ets.insert(ConCache.ets(full_cache_name), items)
+    |> Enum.group_by(fn {key, _} -> {key, get_name(cache_name, key)} end)
+    |> Enum.each(fn {{key, full_cache_name}, items} ->
+      ets = ConCache.ets(full_cache_name)
+
+      if :ets.info(ets, :type) in [:bag, :duplicate_bag] do
```
aerosol, line 106: having second thoughts if altering `put_many/2` is the right approach; we can do this in `merge_items` alternatively, so that the expectations for `put` and `put_many` are equal.

lib/plausible/cache/adapter.ex
```diff
@@ -99,9 +99,15 @@ defmodule Plausible.Cache.Adapter do
   @spec put_many(atom(), [any()]) :: :ok
   def put_many(cache_name, items) when is_list(items) do
     items
-    |> Enum.group_by(fn {key, _} -> get_name(cache_name, key) end)
-    |> Enum.each(fn {full_cache_name, items} ->
-      true = :ets.insert(ConCache.ets(full_cache_name), items)
+    |> Enum.group_by(fn {key, _} -> {key, get_name(cache_name, key)} end)
```
apata, line 102: I'm a bit worried that previously the count of groups was at most equal to partition count. Now it's going to be up to item count. Will :ets.insert be performant if the item count is large?

aerosol:
WIP still

PR #5962, opened 2025-12-18
Improve reports layout

sanne-san:
### Changes

- Create new `ReportLayout` and `ReportHeader` components to ensure consistent layout across reports
- Remove title from report header, and place tabs in the top left corner
- Move the details button to the top right corner, rather than the bottom of the report, to reduce vertical space and improve readability. Reduce it to a simple icon with a 'View details' tooltip.
- Update the Search terms report to use the new `ReportLayout` and `ReportHeader` components.
- Update dashboard to use a simple grid layout, simplifying the html markup and css.
- Change 'Screen sizes' title to 'Devices' in the devices report.
- Change 'Top pages' title to 'Conversion pages' in the pages report, whenever a conversion goal filter is applied.
- Change 'Goals' tab label to the name of the special goal whenever filtered by special goal (e.g. 'File download').
- Add 'last 30min' pill to the behaviour report when on realtime dashboard, rather than displaying this in the title.
- Remove the Combobox input from the custom properties report, and use the same tab dropdown pattern as used for funnels and UTM campaigns.
- Increased the number of items displayed by default before pagination is applied from `5` to `10` in the tab dropdown.
- Activate search input when tab dropdown is opened (if search input is present).
- Improve map and graph colors.
- Improve map tooltip typography and fix regression of tooltip being clipped on the edges of the box.
- Improve typography of the dashboard top bar.
- Add icons for `Ultra-wide` and `(not set)` device sizes.

### Tests

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/more-link.js
+++ b/assets/js/dashboard/stats/more-link.js
@@ -20,7 +20,7 @@ function detailsIcon() {
   )
 }
 
-export default function MoreLink({ linkProps, state, onClick = undefined }) {
+export default function MoreLink({ linkProps, state }) {
   const portalRef = useRef(null)
 
   useEffect(() => {
@@ -47,11 +47,7 @@ export default function MoreLink({ linkProps, state, onClick = undefined }) {
 
   return (
     <Tooltip info="View details" containerRef={portalRef}>
-      <AppNavigationLink
-        {...linkProps}
-        className={baseClassName}
-        onClick={onClick}
-      >
+      <AppNavigationLink {...linkProps} className={baseClassName}>
         {icon}
       </AppNavigationLink>
     </Tooltip>
--- a/assets/js/dashboard/stats/behaviours/conversions.js
+++ b/assets/js/dashboard/stats/behaviours/conversions.js
@@ -6,7 +6,6 @@ import * as metrics from '../reports/metrics'
 import ListReport from '../reports/list'
 import { useSiteContext } from '../../site-context'
 import { useQueryContext } from '../../query-context'
-import { conversionsRoute } from '../../router'
 
 export default function Conversions({ afterFetchData, onGoalFilterClick }) {
   const site = useSiteContext()
@@ -50,10 +49,6 @@ export default function Conversions({ afterFetchData, onGoalFilterClick }) {
       keyLabel="Goal"
       onClick={onGoalFilterClick}
       metrics={chooseMetrics()}
-      detailsLinkProps={{
-        path: conversionsRoute.path,
-        search: (search) => search
-      }}
       color="bg-red-50 group-hover/row:bg-red-100"
       colMinWidth={90}
     />
--- a/assets/js/dashboard/stats/behaviours/goal-conversions.js
+++ b/assets/js/dashboard/stats/behaviours/goal-conversions.js
@@ -11,7 +11,6 @@ import {
 } from '../../util/filters'
 import { useSiteContext } from '../../site-context'
 import { useQueryContext } from '../../query-context'
-import { customPropsRoute } from '../../router'
 
 export const SPECIAL_GOALS = {
   404: { title: '404 Pages', prop: 'path' },
@@ -87,11 +86,6 @@ function SpecialPropBreakdown({ prop, afterFetchData }) {
       getFilterInfo={getFilterInfo}
       keyLabel={prop}
       metrics={chooseMetrics()}
-      detailsLinkProps={{
-        path: customPropsRoute.path,
-        params: { propKey: url.maybeEncodeRouteParam(prop) },
-        search: (search) => search
-      }}
       getExternalLinkUrl={getExternalLinkUrlFactory()}
       color="bg-red-50"
       colMinWidth={90}
--- a/assets/js/dashboard/stats/behaviours/props.js
+++ b/assets/js/dashboard/stats/behaviours/props.js
@@ -6,7 +6,6 @@ import * as url from '../../util/url'
 import { EVENT_PROPS_PREFIX, hasConversionGoalFilter } from '../../util/filters'
 import { useQueryContext } from '../../query-context'
 import { useSiteContext } from '../../site-context'
-import { customPropsRoute } from '../../router'
 
 export default function Properties({ propKey, afterFetchData }) {
   const { query } = useQueryContext()
@@ -47,11 +46,6 @@ export default function Properties({ propKey, afterFetchData }) {
         getFilterInfo={getFilterInfo}
         keyLabel={propKey}
         metrics={chooseMetrics()}
-        detailsLinkProps={{
-          path: customPropsRoute.path,
-          params: { propKey },
-          search: (search) => search
-        }}
         color="bg-red-50 group-hover/row:bg-red-100"
         colMinWidth={90}
       />
--- a/assets/js/dashboard/stats/locations/index.js
+++ b/assets/js/dashboard/stats/locations/index.js
@@ -19,8 +19,9 @@ import { ReportLayout } from '../reports/report-layout'
 import { ReportHeader } from '../reports/report-header'
 import { TabButton, TabWrapper } from '../../components/tabs'
 import MoreLink from '../more-link'
+import { MoreLinkState } from '../more-link-state'
 
-function Countries({ query, site, onClick, afterFetchData, onListUpdate }) {
+function Countries({ query, site, onClick, afterFetchData }) {
   function fetchData() {
     return api.get(apiPath(site, '/countries'), query, { limit: 9 })
   }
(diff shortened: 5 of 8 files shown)


assets/js/dashboard/hooks/use-more-link-data.tsx
```diff
@@ -0,0 +1,31 @@
+import { useCallback, useState } from 'react'
+import { AppNavigationLinkProps } from '../navigation/use-app-navigate'
+
+export function useMoreLinkData() {
```
apata: issue, blocking: this hook duplicates state ([creates redundant state](https://react.dev/learn/choosing-the-state-structure#:~:text=Avoid%20redundant%20state.%20If%20you%20can%20calculate%20some%20information%20from%20the%20component%E2%80%99s%20props%20or%20its%20existing%20state%20variables%20during%20rendering%2C%20you%20should%20not%20put%20that%20information%20into%20that%20component%E2%80%99s%20state.)), which is probably the reason we need 
`// eslint-disable-next-line react-hooks/exhaustive-deps` where it's used.
To fix, we can get whether the list is loading and list data from hook props.

aerosol:
When I filter by goal, the percentage column becomes static, is that intended? If so, what's the reasoning there? Why do they have to slide in on hover in the unfiltered view? 

If I use the mouse cursor to just point where I'm looking at, it feels distracting and makes me wonder what can I do to keep the calm experience from "filter by goal" view.

assets/js/dashboard/components/tabs.tsx
```diff
@@ -148,24 +159,30 @@ const Items = ({
     popover.items.classNames.hoverLink
   )
 
+  useEffect(() => {
+    if (searchable && showSearch && searchRef.current) {
+      const timeoutId = setTimeout(() => {
```
RobertJoonas, line 164: Q: Why are we focusing the search in 100ms and not instantly?

assets/js/dashboard/stats/devices/index.js
```diff
@@ -94,11 +98,12 @@ function Browsers({ afterFetchData }) {
         path: browsersRoute.path,
         search: (search) => search
       }}
+      onListUpdate={onListUpdate}
```
RobertJoonas: I think we can simplify things up a bit.

1. Here (and in other call sites of ListReport) we can avoid introducing `onListUpdate` as a new prop and just include its behaviour in `afterFetchData`
2. Avoid the new `useMoreLinkData` hook

The component structure looks something like this (example of Devices, but it's pretty much the same everywhere else):

* Devices
  * **MoreLink**
  * Browsers
    * **ListReport**
  * OperatingSystems
  * ...

The `MoreLink` component doesn't actually need the `list` of data in order to render itself. It just needs to know:

1) whether or not to render at all
2) whether or not it should actually be a clickable link
3) the linkProps to construct an `AppNavigationLink`

The first two can be achieved via the `afterFetchData` hook. For the third one, we need to move `detailsLinkProps` construction from `ListReport` props into Devices (with the value depending on the currently selected tab).

Here is more or less how I imagine it could work:

```js
// in devices/index.js
const [moreLinkVisible, setMoreLinkVisible] = useState(true)
const [moreLinkClickable, setMoreLinkClickable] = useState(false)

// the function already exists there
function afterFetchData(apiResponse) {
  setLoading(false)
  setSkipImportedReason(apiResponse.skip_imported_reason)

  // this needs to be added
  if (response.results.length > 0) {
    setMoreLinkClickable(true)
  } else {
    setMoreLinkVisible(false)
  }
}

function moreLinkProps() {
  switch (mode) {
    case 'browser':
      return {
        path: browsersRoute.path,
        search: (search) => search
      }
    case 'os':
      return {
        path: operatingSystemsRoute.path,
        search: (search) => search
      }
    // etc...
  }
}

return (
  // ...
  <MoreLink
    visible={moreLinkVisible}
    clickable={moreLinkClickable}
    linkProps={moreLinkProps()}
  >
)
```

sanne-san:
> When I filter by goal, the percentage column becomes static, is that intended? If so, what's the reasoning there? Why do they have to slide in on hover in the unfiltered view?
> 
> If I use the mouse cursor to just point where I'm looking at, it feels distracting and makes me wonder what can I do to keep the calm experience from "filter by goal" view.

@aerosol this is part of a different PR/feature, you can find the full discussion [here](https://3.basecamp.com/5308029/buckets/44692882/card_tables/cards/9256707083). Feel free to add your feedback as we've been going back and forth on this decision. The reason for the percentage not being permanently visible is to keep the dashboard default clean, and percentage is only a "supporting" metric. When filtered by goal, the `CR` shows, which isn't just directly correlated with `Conversions` in the same way `Percentage` is to `Visitors`. It's an important metric to see when filtering by goal hence the decision to keep it visible permanently.

assets/js/dashboard/stats/devices/index.js
sanne-san: Thank you so much for the detailed feedback! 🙏 I've updated this, and also changed the `visible` and `clickable` states to an enum for better state management. 

Also centered the CTA – good catch!

assets/js/dashboard/components/tabs.tsx
sanne-san: The delay is to account for the [animation](https://github.com/plausible/analytics/blob/c11d47a11f634ea8e4466672696bc986240d5d6e/assets/js/dashboard/components/popover.tsx#L6-L9) of the popover.

assets/js/dashboard/hooks/use-more-link-data.tsx
sanne-san: Thanks for spotting this 🙏 Fixed as per @RobertJoonas's suggestion.

assets/js/dashboard/nav-menu/segments/searchable-segments-section.tsx
```diff
@@ -90,7 +90,7 @@ export const SearchableSegmentsSection = ({
         )}
       </div>
 
-      <div className="max-h-[210px] overflow-y-scroll">
+      <div className="max-h-[210px] overflow-y-auto">
```
apata:
```suggestion
      <div className="max-h-[228px] overflow-y-auto">
```
This little change allows initial 5 items and "Show more" to fit without scrollbar. They used to fit in 210px before but the height of the items has changed with the improvements to the dashboard.

assets/js/dashboard/stats/more-link.js
```diff
@@ -19,20 +20,40 @@ function detailsIcon() {
...
-        <AppNavigationLink
-          {...linkProps}
-          className="leading-snug font-bold text-sm text-gray-500 dark:text-gray-400 hover:text-gray-600 dark:hover:text-gray-300 transition-colors duration-150 tracking-wide"
-          onClick={onClick}
-        >
-          {detailsIcon()}
-          DETAILS
-        </AppNavigationLink>
-      </div>
+      <Tooltip info="View details" containerRef={portalRef}>
```
apata, line 42: issue, blocking: on iOS Chrome (on real phone), this tooltip ends up staying active over the modal that opens

Image

assets/js/dashboard/hooks/use-more-link-data.tsx
apata: Thanks! 

It definitely works, although it still duplicates list data and loading state. The reason duplicated state is not good is because then there's a risk that the original source state and the duplicated state get out of sync. Out of sync state can cause UI glitches, sometimes pretty severe ones. However, it's not a blocker when taking into consideration the fact that we're getting rid of React soonish. 

I think I can see why you needed to do it: `MoreLink` is _higher_ in the component tree than the component that holds state with query results and query loading status. `afterFetchData` is also a workaround for that fact. 

Just to share my view on this, not requesting you make any changes, I think the most appropriate solution would have been to lift the logic with query results and loading state logic higher up (to the level that renders MoreLink or above) and pass that state down, rather than to bubble it up from below.

assets/js/dashboard/nav-menu/segments/searchable-segments-section.tsx
sanne-san: Great catch! 🙏

assets/js/dashboard/hooks/use-more-link-data.tsx
sanne-san: Yeah I see what you mean, that would be better indeed.  I'll leave it up to you whether you think we should invest the time? (I think I'd need help with that if we wanted to go that route.)

assets/js/dashboard/stats/more-link.js
sanne-san: I fixed the issue by updating the z-index – do you think we should prevent the tooltip from showing altogether when an element is clickable on mobile? I'm not sure if that's overengineering it and possibly causing undesired knock-on effects 🤔

apata: Thanks for sorting it out! I think your quick fix is appropriate. Let's set aside a separate time to polish the dashboard  for the 33% of visitors that view it on mobile devices. In this case, my hope was just to avoid frustrating mobile users arriving at the detailed views. 
Screenshot 2025-12-25 at 12 55 59

apata (approved):

assets/js/dashboard/hooks/use-more-link-data.tsx
apata: Thanks for understanding! No need for any changes right now, resolving!

assets/js/dashboard/stats/devices/index.js
RobertJoonas: @sanne-san thanks for tackling this! I see it was a bit trickier to implement my propsal on the Locations report (due to it being a class component). I've now done that in https://github.com/plausible/analytics/pull/5974, getting rid of `onListUpdate` and `onDataUpdate` for good. The PR also stops passing `detailsLinkProps` to ListReport as they've become redundant. Feel free to merge those changes into this PR as you see fit.

sanne-san: Thank you so much, a lot cleaner this way! 🙏

PR #5964, opened 2025-12-18
DashboardQueryParser: Labels and segments

RobertJoonas:
### Changes

* Parse strings to integers for segment and city filters, before passing them to ApiQueryParser.parse_filters. Makes both segment and city filters work. Adds missing tests.
* When serializing, add location/segment labels where necessary + tests. Can be removed once the whole dashboard is in LiveView, but needed by React during the migration phase.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/devices/index.js
+++ b/assets/js/dashboard/stats/devices/index.js
@@ -438,7 +438,7 @@ export default function Devices() {
   }
 
   return (
-    <div className="group/report overflow-x-hidden">
+    <div className="overflow-x-hidden">
       <div className="flex justify-between w-full">
         <div className="flex gap-x-1">
           <h3 className="font-bold dark:text-gray-100">Devices</h3>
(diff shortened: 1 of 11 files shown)


lib/plausible/stats/dashboard_query_serializer.ex
```diff
@@ -6,26 +6,26 @@ defmodule Plausible.Stats.DashboardQuerySerializer do
 
   alias Plausible.Stats.{ParsedQueryParams, DashboardQueryParser, QueryInclude}
 
-  def serialize(%ParsedQueryParams{} = params) do
+  def serialize(%ParsedQueryParams{} = params, segments) do
```
zoldar: Maybe let's make `segments` optional and default to `[]`?

RobertJoonas: Agreed! Less work to change all the tests later on as well. https://github.com/plausible/analytics/pull/5964/commits/c08d8285efbd621559440bc742b12506575c7773

PR #5969, opened 2025-12-22
Remove internal query api schema

RobertJoonas:
### Changes

This PR gets rid of the `only :internal` comment in `query-api-schema.json`. From now on, any query that needs to be built "internally" can use the `QueryBuilder.build` function, feeding it already parsed query params.

_Side note: Legacy query construction (`Query.from/2`) still works the same way. It doesn't make sense to focus on that at this stage, since `Api.StatsController` actions using `Query.from` will all be replaced with a LiveView dashboard at some point. After `Api.StatsController` is deleted, we can potentially clean up `Query.from` by introducing an `LegacyApiQueryParser`. But all of that is out of scope for now._

Here's a high-level overview of what's happening in this PR. It's a good idea to review commit by commit to follow along better.

* ApiQueryParser doesn't accept the `schema` argument anymore - it always works with the public API schema from now on.
* All query validation tests that are using `:internal` schema argument, are moved from `Query.ParseAndBuildTest` -> `QueryBuilderTest`. Some of the tests only focusing on the "parsing" part were simply removed.
* All ad-hoc query building around the codebase now happens via `QueryBuilder.build`. This involved quite a bit of manual work. I've also refactored tests to use `QueryBuilder` instead of the legacy `Query.from` (which will hopefully be removed at some point).

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/segments/segment.ex
+++ b/lib/plausible/segments/segment.ex
@@ -109,6 +109,12 @@ defmodule Plausible.Segments.Segment do
 
   @spec validate_segment_data(Plausible.Site.t(), map(), boolean()) ::
           :ok | {:error, {:invalid_filters, String.t()}}
+  @doc """
+  Checks whether the filters in segment_data are valid by attempting to build
+  a `%Query{}` struct with them. The `metrics` and `input_date_range` values
+  do not matter as long as they don't affect filters being the only field
+  affecting query validity.
+  """
   def validate_segment_data(%Plausible.Site{} = site, %{"filters" => filters}, restricted_depth?) do
     with {:ok, parsed_filters} <- ApiQueryParser.parse_filters(filters),
          {:ok, _} <-


lib/plausible/stats/goal_suggestions.ex
```diff
@@ -39,16 +39,11 @@ defmodule Plausible.Stats.GoalSuggestions do
...
-          "date_range" => [Date.to_iso8601(from_date), Date.to_iso8601(to_date)],
-          "metrics" => ["pageviews"],
-          "include" => %{"imports" => true}
-        }
-      )
+      QueryBuilder.build!(site, %ParsedQueryParams{
+        input_date_range: {:date_range, from_date, to_date},
+        metrics: [:pageviews],
+        include: %QueryInclude{imports: true}
+      })
```
zoldar, line 46: I wonder if, purely for convenience, could we allow using keyword lists here as an option, which could be implicitly wrapped in ParsedQueryParams (and would crash if there's an unexpected key), so that we could write it as:

```elixir
QueryBuilder.build!(
  input_date_range: {:date_range, from_date, to_date},
  metrics: [:pageviews],
  include: [imports: true]
)
```

lib/plausible/segments/segment.ex
```diff
@@ -102,77 +103,27 @@ defmodule Plausible.Segments.Segment do
...
-           build_naive_query_from_segment_data(site, filters),
+  @spec validate_segment_data(Plausible.Site.t(), map(), boolean()) ::
+          :ok | {:error, {:invalid_filters, String.t()}}
+  def validate_segment_data(%Plausible.Site{} = site, %{"filters" => filters}, restricted_depth?) do
+    with {:ok, parsed_filters} <- ApiQueryParser.parse_filters(filters),
+         {:ok, _} <-
+           QueryBuilder.build(site, %ParsedQueryParams{
+             metrics: [:visitors],
+             input_date_range: {:last_n_days, 7},
+             filters: parsed_filters
```
apata, line 124: question, non-blocking: Was it obvious why we're building specifically 7d visitors query to validate a segment? If no, we should probably keep the comment.

lib/plausible/segments/segment.ex
```diff
@@ -102,77 +103,27 @@ defmodule Plausible.Segments.Segment do
...
-  def reformat_filters_errors(message) do
-    lines = String.split(message, "\n")
-
-    if Enum.all?(lines, fn line ->
-         String.starts_with?(line, "#/filters/") or String.starts_with?(line, "Invalid filters.")
-       end) do
-      {:error, {:invalid_filters, message}}
-    else
-      {:error, message}
+      {:error, message} -> {:error, {:invalid_filters, message}}
```
apata, line 129: issue, blocking: It's possible that the `QueryBuilder.build` call gives errors about something other than filters. We can't remap to `{:error, {:invalid_filters, message}}` since it might not be the filters at fault, it might actually be an issue with the fake query (e.g. _this metric isn't queryable with these filters_). In that case, we'd ideally raise an error so we could patch the fake query.

RobertJoonas: I see what you mean. My thinking was to just assume that the `:visitors` metric doesn't affect the filter validity in any way. It currently doesn't, and I can't think of any future filter that would become incompatible with this metric either. Same with `input_date_range` -- I don't think there's any way the date range could affect filters validity.

Even with the addition of such a filter (that doesn't allow querying the `:visitors` metric), the worst case scenario is that a customer gets an "invalid filters" error when trying to create/update a segment.

>  In that case, we'd ideally raise an error so we could patch the fake query.

That would also result in the same customer experience. The only difference is that we'd get a Sentry error instead of having to hear about broken behaviour from customers. I understand that's a benefit, but IMHO it feels more like future-proofing than a blocker to this PR. With the way things stand currently, it's impossible for the error message to be about anything else other than filters.

How about this: We merge this PR for now, and introduce a new `QueryBuilderError` struct with both `code` and `message` fields in a follow-up, which will make this future-proofing check a bit more convenient. We have discussed this idea with @zoldar previously and have agreed that we'll implement it when there's actual need to consume the internal error codes.

lib/plausible/segments/segment.ex
RobertJoonas: Added a comment in https://github.com/plausible/analytics/pull/5969/commits/5ff99052ed71fc58b30506e9f77c597027472f34

apata: Good comment!

lib/plausible/stats/goal_suggestions.ex
RobertJoonas: Yeah, that sounds reasonable. I'd do that in a follow-up PR though.

lib/plausible/segments/segment.ex
apata: Thanks for the in-depth reply! 

I also couldn't think of a filters combination that wasn't compatible with visitors (that's why there was no test), but there's so many possible combinations and we're constantly upgrading querying, so it seemed worthwhile to specifically handle filters errors only.  

It wasn't future-proofing: it was having cases according to the possible output of the `with` statement. 

However, that error string parsing looked a bit hacky, so I can see why you wanted to get rid of it. Happy to hear that you are thinking of adding a code field in the future, that'd be much easier to match! While we don't have the code field, why can't the string parsing stay though? I wouldn't ask you to invent it, but it was already in place and even had doctests. 

Anyways, you're right that it's not a blocker.

RobertJoonas: The thing that confused me was not crashing at the spot when first encountering a non-filters-related error. There's no reason to carry a different error structure forward only for it to crash later on with a "No case clause matching" error. Once we have the error codes for internal consumption, I'm thinking to do something like this:

```elixir
{:error, %{code: :invalid_filters, message: message}} -> {:error, {:invalid_filters, message}}
_ -> raise "Segment validation failed with an error from QueryBuilder that was not filters related"
```

>  I wouldn't ask you to invent it, but it was already in place and even had doctests.

It was. Of course it wouldn't hurt to leave it, but I just don't see any point to add it back in, since it's currently (and in the foreseeable future) impossible to hit a non-filters-related query builder error here.

PR #5971, opened 2025-12-22
QueryBuilder follow-up

RobertJoonas:
### Changes

This PR addresses two review comments from https://github.com/plausible/analytics/pull/5969:

* Bring back the behaviour that on segment validation, when query-building fails, we make sure the error is related to filters. If it's something else, we'll crash.
  * Instead of matching on strings, we'll now use a `%QueryError{}` struct with an internal `:invalid_filters` code.
* Build a convenience form of `QueryBuilder.build` that accepts a keyword list of params and turns them into a struct automatically.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - A visitor percentage breakdown is now shown on all reports, both on the dashboard and in the detailed breakdown
+- Shared links can now be limited to a particular segment of the data
 
 ### Removed
 
--- a/assets/js/dashboard.tsx
+++ b/assets/js/dashboard.tsx
@@ -7,7 +7,7 @@ import { createAppRouter } from './dashboard/router'
 import ErrorBoundary from './dashboard/error/error-boundary'
 import * as api from './dashboard/api'
 import * as timer from './dashboard/util/realtime-update-timer'
-import { redirectForLegacyParams } from './dashboard/util/url-search-params'
+import { maybeDoFERedirect } from './dashboard/util/url-search-params'
 import SiteContextProvider, {
   parseSiteFromDataset
 } from './dashboard/site-context'
@@ -19,6 +19,8 @@ import {
   SomethingWentWrongMessage
 } from './dashboard/error/something-went-wrong'
 import {
+  getLimitedToSegment,
+  parseLimitedToSegmentId,
   parsePreloadedSegments,
   SegmentsContextProvider
 } from './dashboard/filtering/segments-context'
@@ -39,8 +41,15 @@ if (container && container.dataset) {
       api.setSharedLinkAuth(sharedLinkAuth)
     }
 
+    const limitedToSegmentId = parseLimitedToSegmentId(container.dataset)
+    const preloadedSegments = parsePreloadedSegments(container.dataset)
+    const limitedToSegment = getLimitedToSegment(
+      limitedToSegmentId,
+      preloadedSegments
+    )
+
     try {
-      redirectForLegacyParams(window.location, window.history)
+      maybeDoFERedirect(window.location, window.history, limitedToSegment)
     } catch (e) {
       console.error('Error redirecting in a backwards compatible way', e)
     }
@@ -83,7 +92,8 @@ if (container && container.dataset) {
               }
             >
               <SegmentsContextProvider
-                preloadedSegments={parsePreloadedSegments(container.dataset)}
+                limitedToSegment={limitedToSegment}
+                preloadedSegments={preloadedSegments}
               >
                 <RouterProvider router={router} />
               </SegmentsContextProvider>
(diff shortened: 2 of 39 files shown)


lib/plausible/stats/query_error.ex
```diff
@@ -0,0 +1,4 @@
+defmodule Plausible.Stats.QueryError do
```
zoldar, line 1: Minor nit: file name does not match module name

extra/lib/plausible_web/live/funnel_settings/form.ex
```diff
@@ -10,7 +10,7 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
 
   import PlausibleWeb.Live.Components.Form
   alias Plausible.{Goals, Funnels}
-  alias Plausible.Stats.{QueryBuilder, ParsedQueryParams}
+  alias Plausible.Stats.{QueryBuilder}
```
zoldar: Cosmetic nit: braces are redundant in this and other similar occurrences.

lib/plausible/stats/parsed_query_params.ex
```diff
@@ -15,6 +15,8 @@ defmodule Plausible.Stats.ParsedQueryParams do
             pagination: nil,
             include: %Plausible.Stats.QueryInclude{}
 
+  @type t() :: %__MODULE__{}
```
zoldar: Is that some leftover from an attempt to add typespecs? (I'm not pushing for them, just asking!) 😅

lib/plausible/segments/segment.ex
```diff
@@ -126,7 +126,14 @@ defmodule Plausible.Segments.Segment do
...
     else
-      {:error, message} -> {:error, {:invalid_filters, message}}
+      {:error, %QueryError{code: :invalid_filters, message: message}} ->
+        {:error, {:invalid_filters, message}}
+
+      {:error, :deep_filters_not_supported} ->
+        {:error, {:invalid_filters, "Invalid filters. Deep filters are not supported."}}
+
+      {:error, message} ->
+        raise "Segment validation failed unexpectedly: #{inspect(message)}"
```
zoldar: I don't see any branch that could trigger this clause? Maybe just let it crash then? 🤔

apata: FWIW I am in support of this `raise`. It's connected with https://github.com/plausible/analytics/pull/5969#discussion_r2640372388

If there were no branches that could match this, wouldn't credo alert us with "...can never match"?

apata (approved):
Neatly done!

lib/plausible/segments/segment.ex
RobertJoonas: Yeah, it's currently impossible to reach this clause because building a query with `metrics: [:visitors]` and `input_date_range: {:last_n_days, 7}` is always valid. If an error emerges from `QueryBuilder`, it's 100% filters.

> Maybe just let it crash then?

Oh, of course. Simply removing the fallback `{:error, message}` clause would also make it crash. The Sentry message will be different but the stacktrace will tell us nicely that a "no case clause error" came from `validate_segment_data/3`. https://github.com/plausible/analytics/pull/5971/commits/6833bdb93bbe72413e8e96ae9afdcb4d5190a658

RobertJoonas: > If there were no branches that could match this, wouldn't credo alert us with "...can never match"?

Apparently it doesn't. Perhaps because the error messages are coming not only from `QueryBuilder`, but from deeper functions like (e.g. `TableDecider`) too. Maybe it would be different if it would all be typespecced but I'm no expert here. 🤷🏼‍♂️

lib/plausible/stats/query_error.ex
RobertJoonas: Nice spot! Thanks! https://github.com/plausible/analytics/pull/5971/commits/ca3ef3cc286f52935b1ffbdfdcd546cefa752721

extra/lib/plausible_web/live/funnel_settings/form.ex
RobertJoonas: Thanks! https://github.com/plausible/analytics/pull/5971/commits/feb98dbb9323f92fa789a4d0ee416258b97dfbc1

lib/plausible/stats/parsed_query_params.ex
RobertJoonas: > Is that some leftover from an attempt to add typespecs?

Hehe, yes... And then I decided - not now. 😅 Removed it for now in https://github.com/plausible/analytics/pull/5971/commits/feb98dbb9323f92fa789a4d0ee416258b97dfbc1

lib/plausible/segments/segment.ex
zoldar: Dialyzer is pretty optimistic in its error checking and by its very nature won't prove that code is error-free. A chapter on "Success Typing" in https://learnyousomeerlang.com/dialyzer gives a pretty good background on that.

PR #5975, opened 2025-12-29
Live.Dashboard.Pages continued

RobertJoonas:
### Changes

* Use Stats.query instead of the legacy Stats.breakdown for fetching pages
* Move metric labelling logic into the query API itself to keep the LiveView code dumb about that. This works with a `query.include.dashboard_metric_labels` flag that optionally writes a `meta.metric_labels` map into the response.
* Consider user preferences in DashboardQueryParser

#### About user preferences

My initial idea was that an empty URL getting parsed into `ParsedQueryParams`, and then serialized back into a query string shouldn’t add stuff into the URL due to defaults or localStorage preferences. E.g.:

1. user loads dashboard with empty query string
2. DashboardQueryParser defaults input_date_range to `{:last_n_days, 28}` (or another valid period from localStorage)
3. In the Top Pages report, every page entry will be a `filterLink` serializing `ParsedQueryParams` into a new, updated query string with the filter added. Clicking it will take the user to `?period=28d&f=…` (`period=28d` param appears in the URL out of the blue - shouldn't happen).

In addition to the `period` param, there are two more query params that would get parsed into the "global" dashboard query state: `comparison` and `match_day_of_week`. So basically, if the user has the following preferences saved in localStorage:

* comparison=previous_period
* match_day_of_week=false
* period=year

Then DashboardQueryParser.parse("") |> DashboardQuerySerializer.serialize() might even result in `?comparison=previous_period&match_day_of_week=false&period=year`.

I wanted to avoid this, but I'm now leaning more towards accepting it because:

1) (most importantly) although stuff will appear in the URL, it doesn't disrupt anyone's dashboard experience as long as the URL params are handled equivalently between React and LiveView.
2) Parsing the global query and applying defaults happens in a single place. Otherwise, e.g. date-picker and pages would both have to apply user preferences on their own in order to know what's the real dashboard period.
3) We would need to allow `compare_match_day_of_week` to be `nil` (instead of enforcing boolean) in `query.include`

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/live/dashboard/pages_test.exs
+++ b/test/plausible_web/live/dashboard/pages_test.exs
@@ -108,7 +108,7 @@ defmodule PlausibleWeb.Live.Dashboard.PagesTest do
                |> get_report_list(@entry_pages_report_list)
 
       assert report_list_as_table(report_list, 3, 2) == [
-               ["Page", "Current visitors"],
+               ["Entry page", "Current visitors"],
                ["/two", "2"],
                ["/one", "1"]
              ]


lib/plausible/stats/dashboard_query_parser.ex
```diff
@@ -28,48 +29,77 @@ defmodule Plausible.Stats.DashboardQueryParser do
...
+  @valid_period_shorthand_keys Map.keys(@valid_period_shorthands)
+
+  @valid_comparison_shorthands %{
+    "previous_period" => :previous_period,
+    "year_over_year" => :year_over_year
+  }
+
+  @valid_comparison_shorthand_keys Map.keys(@valid_comparison_shorthands)
+
+  def parse(query_string, site, user_prefs) when is_binary(query_string) do
```
apata, line 55: question, non-blocking: Do we need to give up the simpler API of `parse(query_string, defaults \\ %{})`? The new API `parse(query_string, site, user_prefs)` gives this function a bunch more complexity and its tests require building a site. What if we extracted the part that gets the defaults based on stored prefs and tested it separately?

lib/plausible/stats/parsed_query_params.ex
```diff
@@ -44,4 +44,10 @@ defmodule Plausible.Stats.ParsedQueryParams do
 
     struct!(parsed_query_params, filters: new_filters)
   end
+
+  def conversion_goal_filter?(%__MODULE__{filters: filters}) do
```
apata, line 48: nitpick, non-blocking: While this definitely works, it seems to duplicate the logic that we use elsewhere
```
Plausible.Stats.Filters.filtering_on_dimension?(query, "event:goal",
          max_depth: 0,
          behavioral_filters: :ignore
        )
```

test/support/dashboard_test_utils.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Plausible.DashboardTestUtils do
+  @moduledoc false
+
+  import Plausible.Test.Support.HTML
+
+  def get_in_report_list(%LazyHTML{} = report_list, opts) do
```
apata: Cool utility! It's a bit hard to understand at first glance though, maybe could use a spec?

test/plausible_web/live/dashboard/pages_test.exs
```diff
@@ -0,0 +1,261 @@
...
+
+      assert get_in_report_list(report_list, :key_label) =~ "Exit page"
+      assert get_in_report_list(report_list, metric_label: 0) =~ "Conversions"
+      assert get_in_report_list(report_list, metric_label: 1) =~ "CR"
+
+      assert get_in_report_list(report_list, item_name: 0) =~ "/two"
+      assert get_in_report_list(report_list, item: 0, metric: 0) =~ "1"
+      assert get_in_report_list(report_list, item: 0, metric: 1) =~ "33.33%"
+
+      refute get_in_report_list(report_list, item_name: 1)
```
apata: suggestion, non-blocking: This is pretty powerful, but I think it would be a clearer test if we had an HTML utility that gathered the full textcontent of the report and formatted it a little bit, so we could assert with something like
```
assert(report_list) = ~S"""
Exit page | Conversions | CR
/two | 1 | 33.33%
"""
```
That would take care of asserts and refutes with a single step, also has the benefit of looking like a table.

lib/plausible/stats/query_include.ex
```diff
@@ -8,7 +8,8 @@ defmodule Plausible.Stats.QueryInclude do
             trim_relative_date_range: false,
             compare: nil,
             compare_match_day_of_week: false,
-            legacy_time_on_page_cutoff: nil
+            legacy_time_on_page_cutoff: nil,
+            dashboard_metric_labels: false
```
apata, line 12: suggestion, non-blocking: Since metrics are ordered in the response and the dashboard knows what it requested (choose_metrics), I think we should provide / map the labels there. Then this struct will only contain fields that influence the actual database query.

lib/plausible_web/live/components/dashboard/report_list.ex
```diff
@@ -28,49 +29,57 @@ defmodule PlausibleWeb.Components.Dashboard.ReportList do
...
-      metrics = assigns.metrics.result
-      meta = assigns.meta.result
-      skip_imported_reason = assigns.skip_imported_reason.result
+      %QueryResult{results: results, meta: meta, query: query} = assigns.query_result.result
+
+      # NOTE: Consider a `query.include` flag like `dashboard_style_response` to return
+      # metric values per key in a map to make this a bit easier. Currently, we need to
+      # fetch metrics by indices. For simplicity, we assume that `:visitors` is always
+      # the first metric.
+      :visitors = List.first(query[:metrics])
```
apata: suggestion, non-blocking: I personally think the list of values is actually preferable to a map, especially for these tabular reports. The less remapping there is, the better.

lib/plausible_web/live/dashboard/pages.ex
```diff
@@ -155,97 +104,40 @@ defmodule PlausibleWeb.Live.Dashboard.Pages do
...
+          pagination: @pagination
+        )
 
-    %{results: results, meta: meta} = Stats.breakdown(site, query, metrics, @pagination_params)
+      query = QueryBuilder.build!(site, params)
 
-    pages =
-      results
-      |> transform_keys(%{entry_page: :name})
+      %QueryResult{} = query_result = Stats.query(site, query)
```
apata, line 124: Nice! It's great that we can finally use the new API. 

Locally, I put the old and new pages report side by side and there are a few differences though, specifically 
1) large number formatting
2) CR value
3) some minor font weight differences and button alignment

Image

lib/plausible/stats/dashboard_query_parser.ex
RobertJoonas: This annoyed me a bit too and I tried to avoid it. There were a couple of reason why it still ended up with these arguments:

1. Need to *parse* user preferences too. E.g. from localStorage we get `"28d"` not `{:last_n_days, 28}`. So DashboardQueryParser would still need to expose that parsing function if the user prefs parsing would happen elsewhere.
2. When the stored value is invalid, we need to fall back to site default, but only when the value in URL params is invalid too. That means that we can't just pass a single argument of defaults. We would have to pass `defaults` and `fallback_period`.

So overall, it felt cleaner to let DashboardQueryParser grow a bit in complexity, but keep all the logic in one place rather than scattering it around and having all tests for it in one place.

lib/plausible/stats/parsed_query_params.ex
RobertJoonas: Good point. Updated now in https://github.com/plausible/analytics/pull/5975/commits/796a655bc81e89b47d2a6edc3319ed42e1efb6e7

lib/plausible/stats/query_include.ex
RobertJoonas: I don't see anything wrong with keeping logic unrelated to the ClickHouse SQL query in `Stats.query`. There's also `include.time_labels` which has nothing to do with the DB query either.

However, I agree that the mapping part is unnecessary. I think you reviewed before I pushed this commit: https://github.com/plausible/analytics/pull/5975/commits/9f7695b3200ab69ea29bb17f958e31c07bbb60c0. I've changed `meta.metric_labels` to be a returned as a list instead of map. That way the metric labels can just be used directly, like metric keys.

There's another edge case that will be a bit easier to handle this way. Revenue metrics are sometimes dropped silently (when mixing currencies in goal filters). So even though a LiveView report always asks for them, they might not be returned.

lib/plausible_web/live/components/dashboard/report_list.ex
RobertJoonas: Yup, that's what I did in https://github.com/plausible/analytics/pull/5975/commits/9f7695b3200ab69ea29bb17f958e31c07bbb60c0 :)

lib/plausible/stats/parsed_query_params.ex
RobertJoonas: On second thought, I added back the `ParsedQueryParams.conversion_goal_filter?` function, but made it call `Stats.Filters.filtering_on_dimension` under the hood. We will use that function all around the dashboard so will be more convenient this way. https://github.com/plausible/analytics/pull/5975/commits/9340e3a88e0467a666556fcda3481a74882228d8

test/support/dashboard_test_utils.ex
RobertJoonas: I changed this according to your other suggestion (returning a table-like format). A typespec sounds unnecessary  with the new signature. I've added a comment though. https://github.com/plausible/analytics/pull/5975/commits/3240cc3d61508178861672ffdd8af9c065fad845

test/plausible_web/live/dashboard/pages_test.exs
RobertJoonas: I like the suggestion! https://github.com/plausible/analytics/pull/5975/commits/fa44c4d64d42ef8c093592fc63dc8d6597f3dfdc (Although a 2D list seems like a cleaner way to assert on the table values in tests) 😀 

I also added some tests targeting list report directly.

lib/plausible_web/live/dashboard/pages.ex
RobertJoonas: Thanks for checking. We'll revisit this once https://github.com/plausible/analytics/pull/5962 is live :)

apata: Yep, happy to see the design mature iteratively, what's the reason for the different CR value though? It's 100% on the left (non-LV) and 0.95% on the right (LV)

lib/plausible/stats/query_include.ex
apata: Hmm, it's actually weird that time labels don't go through the database. It's very database adjacent though.

I see your point about it making an edge case easier to handle, I disagree that it's worth the tradeoff (of making Query and its result more complex). 

I suspect we'll be tempted to put a lot of configuration in QueryInclude. I think we should do so only for things that can't be easily handled another way. Metric labels can be. Still, it's not a blocker for merge.

lib/plausible/stats/dashboard_query_parser.ex
apata: Ah, glad to hear that the same concerns crossed your mind! I don't mind this getting merged, but I personally like smaller modules and to have functions with as few responsibilities as possible. I find them easier to understand, unit test, share in the codebase.

As I see it,
1) getting and parsing stored preferences,
2) getting default preferences 

are separate from 

3) parsing url search params.

Those are three different responsibilities. 

There's a fourth responsibility, to mix these three to arrive at what slice the user desires to see. This would contain business rules such as certain periods from URL search params being treated as more important than the stored comparison value. 

If both 1 and 3 need to share the string <-> tuple parser, that just means we should define an underlying query period serializer / parser that has the sole responsibility of handling particular Elixir tuples and their string representations. We could easily unit test it (e.g. with a doctest) for all the possible periods.

test/plausible_web/live/dashboard/pages_test.exs
apata: You're right, the list is way better! Thanks!

apata (approved):

lib/plausible/stats/dashboard_query_parser.ex
RobertJoonas: The separation you mentioned sounds reasonable. It's likely that we'll revisit this during the migration project.

lib/plausible/stats/query_include.ex
RobertJoonas: I wouldn't say this adds much complexity into the query function. A bit more code and a new input/response field, sure, but it doesn't make it more difficult to reason about.

But I see your concern because the `%Query{}` struct is quite messy and it's quite terrifying to look at it. It needs some cleaning up for sure, and it will happen in this project. There are some legacy fields that are not used anymore (`interval`, `sample_threshold`, `include_imported`, `legacy_breakdown`, maybe more...). And we also have a bunch of stuff there that gets slapped onto the query object somewhere along the pipeline for internal use only (`site_id`, `site_native_stats_start_at`, `consolidated_site_ids`, `sql_join_type`, etc...). All of those fields could be nested under a `query.assigns` field.

lib/plausible_web/live/dashboard/pages.ex
RobertJoonas: Ah nice catch! It's the number formatter. We can ignore that for now.

apata: @RobertJoonas I don't think it's the number formatter. The current dashboard sometimes queries for group conversion rate https://github.com/plausible/analytics/blob/708c89d656f4ddc1938caf450f32076d958badf5/lib/plausible/stats/breakdown.ex#L99
Is that remap in the LV version as well?

lib/plausible/stats/query_include.ex
apata: > A bit more code and a new input/response field, sure, but it doesn't make it more difficult to reason about.

Whether something is easy or hard to reason about is quite hard to discuss. 
I'd prefer we be specific. 
Complexity is a bit better, though we can be even more specific (more/less code, more/less code paths, more/less fields). 
If we add a feature/refactor, I suspect the question is not whether we want more complexity, it's where we want it. 

My argument is that we don't want it in `QueryInclude`, here's why:

What's the responsibility of the Query module? As I see it, it's to represent a request for data, and must contain exactly what is needed to get that data _from the database_. `QueryInclude` is a set of options regarding that request for data. Much like filters, dimensions, metrics. 

If we keep to this responsibility, we can say that as long as one doesn't touch `Query` and `QueryInclude`, there's no risk of regression regarding requests for data. This is the benefit that I'm trying to sell. 

`QueryResult` serializes the data. Let's say its responsibility is also sometimes to get human readable names for the metrics queried (_which IMO is debatable, maybe it should be a separate step before QueryResult_). We don't have to pass this option through `Query.include`. It can be an argument of the function `QueryResult.from`.

Regarding cleanup, I'm not sure that `assigns` is the right way to go: that's a bit of an opaque misc drawer. The struct fields are at least explicit. I look at them this way: "these are the fields necessary to get data, they get populated somewhere, maybe, and when they do, they influence the result".

lib/plausible_web/live/dashboard/pages.ex
RobertJoonas: Ah, right. I mistakenly assumed it was 95.5% and the formatter to blame, but you're right. I will fix that in the next PR.

lib/plausible/stats/query_include.ex
RobertJoonas: I understand your point about separating responsibilities but I'm not seeing how this new include field could cause us any pain in the future. The benefit of not having to worry about metric mapping in every LiveView stats report module sounds far more real. That said though, the query refactoring work is still ahead and we can consider it further down the line. 

> Regarding cleanup, I'm not sure that assigns is the right way to go: that's a bit of an opaque misc drawer. The struct fields are at least explicit. I look at them this way: "these are the fields necessary to get data, they get populated somewhere, maybe, and when they do, they influence the result".

It could also be a nested struct with the fields explicitly defined with default values. The point would be to separate the input fields from those that are attached to the query automatically.

apata: > It could also be a nested struct with the fields explicitly defined with default values. The point would be to separate the input fields from those that are attached to the query automatically.

@RobertJoonas, that overall point is very good! 

> I understand your point about separating responsibilities but I'm not seeing how this new include field could cause us any pain in the future. 

Thanks for considering it! This one field is definitely innocent and good code. It just got me thinking on why we need to do it this way. I believe by doing it a little bit differently, we can have the benefits that both of us brought up (shared code between LiveView stats reports, more singular responsibility for QueryInclude). To illustrate what I had in mind, I put together this https://github.com/plausible/analytics/pull/5983.

PR #5977, opened 2026-01-03
Goals with custom props (settings+filtering UI integration)

aerosol:
### Changes

This PR implements full capability of managing goals with custom props via site settings + filtering the dashboard. 

The PR builds upon https://github.com/plausible/analytics/pull/5978, https://github.com/plausible/analytics/pull/5944, https://github.com/plausible/analytics/pull/5946, https://github.com/plausible/analytics/pull/5952

Billing states, such as Props feature unavailable are handled with appropriate CTAs.

Preview env: https://pr-5977.review.plausible.io/

Documentation update + changelog entry pending. 

Additionally, we'll prohibit editing existing revenue goals currency.

Sample screenshots:

https://github.com/user-attachments/assets/1fc96730-e16e-4de6-a040-fcf31b7bc35a

image

image

https://github.com/user-attachments/assets/818ca315-a146-4c2d-a1f2-39b7696496a8

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/goals.ex
+++ b/lib/plausible/stats/goals.ex
@@ -16,10 +16,7 @@ defmodule Plausible.Stats.Goals do
          Filters.filtering_on_dimension?(filters, "event:goal") do
       site = Plausible.Repo.preload(site, :team)
       props_available? = Plausible.Billing.Feature.Props.check_availability(site.team) == :ok
-
-      goals =
-        site
-        |> Plausible.Goals.for_site(include_goals_with_custom_props?: props_available?)
+      goals = Plausible.Goals.for_site(site, include_goals_with_custom_props?: props_available?)
 
       %{
         # When grouping by event:goal, later pipeline needs to know which goals match filters exactly.


test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_equivalence_test.exs
```diff
@@ -0,0 +1,329 @@
...
+          "site_id" => site.domain,
+          "date_range" => "all",
+          "metrics" => ["visitors", "visits", "events"],
+          "dimensions" => ["visit:country"],
+          "filters" => [
+            ["is", "event:goal", ["Visit /checkout-no-props"]],
+            ["is", "event:props:currency", ["EUR"]],
+            ["is", "event:props:country", ["DE"]]
+          ]
+        })
```
apata, line 251: suggestion, non-blocking: I like these parity tests! However, why aren't the page pathnames the same (both acting on the same "/checkout")? It seems like if they were, populate_stats could be twice as short.

test/plausible/stats/goal_suggestions_test.exs
```diff
@@ -102,4 +102,418 @@ defmodule Plausible.Stats.GoalSuggestionsTest do
...
+      ])
+
+      # "plan" appears 3 times, "amount" 1 time, "referrer" 1 time, "button_id" 1 time
+      suggestions = GoalSuggestions.suggest_custom_property_names(site, "")
+      assert "plan" == hd(suggestions)
+      assert length(suggestions) == 4
+      assert "plan" in suggestions
+      assert "amount" in suggestions
+      assert "referrer" in suggestions
+      assert "button_id" in suggestions
```
apata: question, non-blocking: Is the sort not stable? I think we should make it stable, to be able to assert exactly what we expect the response to be: `assert [a,b,c,d] = suggestions`. If the sort is not stable, there's UX impact as well, right? 

- `"" -> a,b,c,d`
- `"a" -> a`
- `"" -> a,c,b,d` <--- may seem like a glitch to the user

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_equivalence_test.exs
aerosol: be7b5f049b

test/plausible/stats/goal_suggestions_test.exs
aerosol: Ordering should be stable. bcd5b3a006

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_equivalence_test.exs
apata: Hmm, that commit changes what was tested. Now it's testing `custom props goal A` == `custom props goal A + its own prop filters`. 

It's not a bad test this way, but what I meant was to have two goals, "Visit /a" and "Visit /a, with prop-key=prop-value", acting on the same events. 

Looking again, all the tests in this module have a similar setup where the underlying events are doubled. What's the benefit of testing equivalence this way?

lib/plausible/stats/goals.ex
```diff
@@ -14,7 +14,12 @@ defmodule Plausible.Stats.Goals do
   def preload_needed_goals(site, dimensions, filters) do
     if Enum.member?(dimensions, "event:goal") or
          Filters.filtering_on_dimension?(filters, "event:goal") do
-      goals = Plausible.Goals.for_site(site)
+      site = Plausible.Repo.preload(site, :team)
+      props_available? = Plausible.Billing.Feature.Props.check_availability(site.team) == :ok
+
+      goals =
+        site
+        |> Plausible.Goals.for_site(include_goals_with_custom_props?: props_available?)
```
apata: nitpick, non-blocking: I think the error emitted to Stats API users may be a bit misleading if we do it this way. It will say that the requested goal doesn't exist, even though it does, just is unavailable.
```
{
  "error": "Invalid filters. The goal `...` is not configured for this site. Find out how to configure goals here: https://plausible.io/docs/stats-api#filtering-by-goals"
}
```

test/plausible_web/controllers/api/external_stats_controller/query_goal_custom_props_equivalence_test.exs
aerosol: sorry, you're right 3b6388f3eb

lib/plausible/stats/goals.ex
aerosol: We can update the documentation once this is live and state that goals with custom props might become unavailable in Stats API, if subscription does not cover props. Settings web UI will make that clear. cc @metmarkosaric

lib/plausible_web/live/goal_settings/form.ex
```diff
@@ -419,6 +514,68 @@ defmodule PlausibleWeb.Live.GoalSettings.Form do
...
+          </.pill>
+        </.link>
+      </div>
+      <.tooltip enabled?={not @has_access_to_props?} centered?={true}>
+        <:tooltip_content>
+          To get access to this feature,
+          <PlausibleWeb.Components.Billing.upgrade_call_to_action
+            current_user={@current_user}
+            current_team={@site_team}
+            link_class="!inline !text-indigo-400 hover:!text-indigo-300"
```
ukutaht, line 550: I was trying to see what these classes do but don't see a visible difference when I remove the line. Am I doing something wrong or are these not really doing anything? Here's what it looks to me with this line removed

Image

lib/plausible_web/live/goal_settings/list.ex
```diff
@@ -257,4 +266,61 @@ defmodule PlausibleWeb.Live.GoalSettings.List do
...
+      <.tooltip :if={Plausible.Goal.has_custom_props?(@goal)} centered?={true}>
+        <:tooltip_content>
+          <div class="-mx-1 flex flex-col gap-1 text-xs">
+            <div :for={{key, value} <- @goal.custom_props} class="truncate">
+              <span class="bg-white/20 px-1 py-0.5 rounded-sm">{key}</span>
+              is <span class="bg-white/20 px-1 py-0.5 rounded-sm">{value}</span>
+            </div>
+          </div>
+        </:tooltip_content>
+        <svg
```
ukutaht, line 289: Suggestion: Extract this into a function with a name for the icon. raw svg is not very readable or descriptive as to what is being rendered here

lib/plausible_web/live/goal_settings/form.ex
aerosol: Kindly paging @sanne-san

lib/plausible_web/live/goal_settings/list.ex
aerosol: @ukutaht yep, we've discussed this already and agreed to do it separately. I'll make sure we have a card filed.

aerosol: https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/9532516795

lib/plausible_web/live/goal_settings/form.ex
sanne-san: Yeah this is because the upgrade CTA appears inside a tooltip here. The tooltip is there for extra context if people try to click the toggle. This means that even in light mode, it appears on a dark background and becomes hard to read. These classes make sure it 1) reads as a single line and 2) it's more visible. With and without:

CleanShot 2026-02-02 at 13 52 05@2x

CleanShot 2026-02-02 at 13 53 06@2x

PR #5979, opened 2026-01-05
Make minor improvements to the dashboard

sanne-san:
### Changes

- Hide Details button for funnels tab
- Show `Search` as placeholder for search input on mobile
- Hide keyboard shortcut hints on mobile

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/modals/breakdown-table.tsx
+++ b/assets/js/dashboard/stats/modals/breakdown-table.tsx
@@ -56,7 +56,9 @@ export const BreakdownTable = <TListItem extends { name: string }>({
               searchRef={searchRef}
               onSearch={onSearch}
               className={
-                displayError && status === 'error' ? '[&_input]:pointer-events-none' : ''
+                displayError && status === 'error'
+                  ? '[&_input]:pointer-events-none'
+                  : ''
               }
             />
           )}


assets/js/dashboard/components/search-input.tsx
```diff
@@ -22,6 +22,9 @@ export const SearchInput = ({
   placeholderUnfocused?: string
 }) => {
   const [isFocused, setIsFocused] = useState(false)
+  const isMobile =
+    typeof window !== 'undefined' &&
+    window.matchMedia('(max-width: 768px)').matches
```
apata: issue, non-blocking: It works, but I think it's a quite hacky. 

It doesn't align with the breakpoint query tailwind is providing (`@media (width >= 48rem) {`. Also, it's more accurately `is under or at MD breakpoint` not `isMobile` (doesn't actually detect touch input only). 
It changes the meaning of `placeholderFocused` to `placeholderFocusedOrMobile` and the meaning of `placeholderUnfocused` to `placeholderUnfocusedOnlyDesktop`.
It doesn't automatically get recalculated when the window is resized, so if you turn a screen from horizontal to vertical, whatever was previously calculated remains in effect. 

I think we should do this with CSS and HTML. 

To do it, 

- we'd need to wrap the search `input` with a relative positioning div
- give input the `peer` class and make the placeholder `" "`
- render a `span` next to the input with absolute positioning, and make it hide itself when the input placeholder is not shown `peer-[:not(:placeholder-shown)]:hidden`
- render the placeholder texts with TW breakpoints `md:...` within the `span`

assets/js/dashboard/keybinding.tsx
```diff
@@ -145,7 +145,7 @@ export function KeybindHint({
   return (
     <kbd
       className={classNames(
-        'rounded bg-white dark:bg-gray-800 border border-gray-200 dark:border-gray-600 px-1.5 font-medium text-xs text-gray-400',
+        'hidden md:block rounded bg-white dark:bg-gray-800 border border-gray-200 dark:border-gray-600 px-1.5 font-medium text-xs text-gray-400',
```
apata, line 148: nitpick, non-blocking: the PR is marked as not changing the UI and not needing a changelog, but I think it does both. It's a small change but I think we should list it, something like
```
### Changed
- Keybind hints are hidden on smaller screens
```

apata (approved):

assets/js/dashboard/keybinding.tsx
sanne-san: Ok cool, will do! That is good to know, I haven't been listing small things like this as I thought it was intended for more major changes 👍

apata: Thanks! I think the value in documenting these small UI fixes is that if people are updating their CE installations, they'll know if some change is intentional or not.

sanne-san: Yeah makes sense!

PR #5980, opened 2026-01-05
Change position of graph tooltip on mobile

sanne-san:
### Changes

- Rather than showing the tooltip in a fixed position below the graph, it now follows your finger just like it follows the cursor on desktop.
- The information of the tooltip on mobile is reduced to just the value and the comparison difference.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/graph-tooltip.js
+++ b/assets/js/dashboard/stats/graph/graph-tooltip.js
@@ -164,7 +164,9 @@ export default function GraphTooltip(graphData, metric, query, theme) {
                 <div className="flex flex-row justify-between items-center text-sm">
                   <span className="flex items-center mr-4">
                     <div className="size-2 mr-2 rounded-full bg-gray-500"></div>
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/graph/line-graph.js
```diff
@@ -165,17 +168,45 @@ class LineGraph extends React.Component {
...
+    if (tooltipEl) {
+      tooltipEl.style.display = null
+    }
+    this.repositionTooltip(e)
+  }
+
+  handleTouchMove = (e) => {
+    e.preventDefault()
+    this.repositionTooltip(e)
+  }
```
apata, line 203: issue, blocking: probably this inhibits ability to scroll past the main graph on iOS Chrome

sanne-san:
Closing as it hijacks scrolling on mobile which is not ideal – will need a more elaborate solution.

PR #5981, opened 2026-01-05
Fix tooltips in modals showing behind the modal

sanne-san:
### Changes

- Regression fix: due to a necessary z-index change on the tooltip in a previous commit, any tooltip that would show inside a modal would be hidden behind the modal. This commit ensures that whenever the tooltip is showing inside of a modal, it has a higher z-index.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/util/tooltip.tsx
+++ b/assets/js/dashboard/util/tooltip.tsx
@@ -91,7 +91,7 @@ function TooltipMessage({
       ref={setPopperElement}
       style={popperStyle}
       {...popperAttributes}
-      className="z-[99] px-2 py-1 rounded-sm text-sm text-gray-100 font-medium bg-gray-800 dark:bg-gray-700"
+      className="z-[99] [body:has(.modal.is-open)_&]:z-[1000] px-2 py-1 rounded-sm text-sm text-gray-100 font-medium bg-gray-800 dark:bg-gray-700"
       role="tooltip"
     >
       {children}


assets/js/dashboard/util/tooltip.tsx
```diff
@@ -91,7 +91,7 @@ function TooltipMessage({
       ref={setPopperElement}
       style={popperStyle}
       {...popperAttributes}
-      className="z-[99] px-2 py-1 rounded-sm text-sm text-gray-100 font-medium bg-gray-800 dark:bg-gray-700"
+      className="z-[99] [body:has(.modal.is-open)_&]:z-[1000] px-2 py-1 rounded-sm text-sm text-gray-100 font-medium bg-gray-800 dark:bg-gray-700"
```
apata, line 94: Good as a hotfix, but wouldn't ideally depend on custom modal class.

sanne-san: Yeah feels a bit hacky, wondering how you would approach this?

apata: I haven't validated this theory yet, but I suspect in some cases we portal the tooltip to document.body when we should be portalling it (or rendering it) more locally in the DOM. If true, it might be a bit of effort to fix without regressing in other ways.

sanne-san: Alright, I will merge this for now!

PR #5983, opened 2026-01-05
Avoid passing through dashboard specific flags through full querying stack

apata:
### Changes

The aim of this PR is to demonstrate how we can have different kinds of query results without specifying it in QueryInclude (and thus passing that particular flag through the whole querying stack). 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/stats/query/query_test.exs
+++ b/test/plausible/stats/query/query_test.exs
@@ -127,29 +127,25 @@ defmodule Plausible.Stats.QueryTest do
     end
   end
 
-  describe "include.dashboard_metric_labels" do
+  describe "including metric labels" do
     test "visitors -> Visitors (default)", %{site: site} do
       {:ok, query} =
         QueryBuilder.build(site,
           metrics: [:visitors],
-          input_date_range: :all,
-          include: [dashboard_metric_labels: true]
+          input_date_range: :all
         )
 
-      %Stats.QueryResult{meta: meta} = Stats.query(site, query)
-      assert ["Visitors"] = meta[:metric_labels]
+      assert {_, ["Visitors"]} = Stats.dashboard_query(site, query)
     end
 
     test "visitors -> Current visitors (realtime)", %{site: site} do
       {:ok, query} =
         QueryBuilder.build(site,
           metrics: [:visitors],
-          input_date_range: :realtime,
-          include: [dashboard_metric_labels: true]
+          input_date_range: :realtime
         )
 
-      %Stats.QueryResult{meta: meta} = Stats.query(site, query)
-      assert ["Current visitors"] = meta[:metric_labels]
+      assert {_, ["Current visitors"]} = Stats.dashboard_query(site, query)
     end
 
     test "visitors -> Current visitors (realtime and goal filtered)", %{site: site} do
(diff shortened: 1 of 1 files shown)


lib/plausible/stats.ex
```diff
@@ -8,11 +8,33 @@ defmodule Plausible.Stats do
...
+    QueryResult
   }
 
   def query(site, query) do
-    QueryRunner.run(site, query)
+    site |> QueryRunner.run(query) |> QueryResult.from()
+  end
+
+  def dashboard_query(site, query) do
+    {site |> QueryRunner.run(query) |> QueryResult.from(), get_dashboard_metric_labels(query)}
```
apata, line 20: This would ideally be something like the following: 
```suggestion
    site |> QueryRunner.run(query) |> DashboardQueryResult.from()
```
Most of the code between QueryResult and DashboardQueryResult would be shared.

RobertJoonas: @apata I don't think we should introduce a new `dashboard_query` function. The goal of the API v2 project was to have a shared powerful function making sure that the behaviour between the public and the dashboard API wouldn't diverge (and for there to be less maintenance). Two separate functions is the first step towards giving those benefits up.

Coming back to the core of the discussion for a moment -- I really don't see any issue with `query.include.dashboard_metric_labels`. There's a lot of business logic in `Stats.query` already that is not strictly related to the database query. It's true that it would be quite easy to write this logic elsewhere. I'm just not convinced in the benefit of that.

apata: True, I remember that one of the ideas with Stats API v2 was to make the dashboard a thinner layer on top of it. However, refactoring the dashboard to LiveView is a move away from that idea, not towards. 

Still, LV dashboard queries and API v2 queries shared behavior is of value. I am not proposing we give that up. Is **result format** a part of **query behavior**? I'd say it isn't. 

This draft shows how we can easily extract it, to diverge the result format between LV dashboard queries (result rendered to HTML server side) and Stats v2 API query results (JSON format). 

As a random example why we'd want to diverge: Do we need `meta` to be Jason.OrderedObject for LV, or would it be better if it were a struct?
```sh
iex(1)> Jason.OrderedObject.new(hi: "world")
%Jason.OrderedObject{values: [hi: "world"]}
iex(2)> Jason.OrderedObject.new(hi: "world").hi
** (KeyError) key :hi not found in:

    %Jason.OrderedObject{values: [hi: "world"]}

    iex:2: (file)
```

Looking closely, there are a few places in the codebase where the result is first formatted, then re-formatted with custom functions.

zoldar: @apata In general, you make a good point. Now, though, we are dealing with only two direct clients of that query API with nothing more in sight for the foreseeable future. Internally, while query struct does carry over the relevant option, the logic acting on it is relatively well isolated as an explicit step in the pipeline. Yes, it breaks the separation on a module level but keeps the logic in a closer proximity with less jumping around. At the current stage, the trade-off seems to be worth it. We can re-evaluate this if/when we ever come to a point where we need to introduce further context specific (API vs Dashboard vs any other hypothetical future API client) logic. I'll add a housekeeping card with a reference to your proposal so that we don't loose the sight of it.

NOTE: The OrderedObject wrappings in `QueryResult` seem to be purely for APIv2 documentation playground UX/aesthetics. I'm not sure it's necessary at all (and worth the added noise and inconvenience).

PR #5986, opened 2026-01-07
Another iteration of Dashboard.Live.Pages

RobertJoonas:
### Changes

- [x] Add real external link construction that handles internationalized domain names
- [x] Imported warning bubble (without tooltip for now)
- [x] Query group_conversion_rate for pages to match prod behaviour
- [x] Update the tile component according to recent layout changes
- [x] Fix group hover for items in ReportList
- [x] Replicate React first paint inside the LiveViewPortal that is not yet connected

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/liveview-portal.tsx
+++ b/assets/js/dashboard/components/liveview-portal.tsx
@@ -18,13 +18,13 @@ const MIN_HEIGHT = 356
 type LiveViewPortalProps = {
   id: string
   tabs: { value: string; label: string }[]
-  storedTab: string | null
+  storageKey: string
   className?: string
 }
 
 export const LiveViewPortal = React.memo(
-  function ({ id, tabs, storedTab, className }: LiveViewPortalProps) {
-    const activeTab = storedTab || 'pages'
+  function ({ id, tabs, storageKey, className }: LiveViewPortalProps) {
+    const activeTab = localStorage.getItem(storageKey) || 'pages'
 
     return (
       <div
--- a/assets/js/dashboard/index.tsx
+++ b/assets/js/dashboard/index.tsx
@@ -12,6 +12,7 @@ import { useSiteContext } from './site-context'
 import { hasConversionGoalFilter, isRealTimeDashboard } from './util/filters'
 import { useAppNavigate } from './navigation/use-app-navigate'
 import { parseSearch } from './util/url-search-params'
+import { getDomainScopedStorageKey } from './util/storage'
 
 function DashboardStats({
   importedDataInView,
@@ -68,7 +69,7 @@ function DashboardStats({
             { label: 'Entry pages', value: 'entry-pages' },
             { label: 'Exit pages', value: 'exit-pages' }
           ]}
-          storedTab={localStorage.getItem(`pageTab__${site.domain}`)}
+          storageKey={getDomainScopedStorageKey('pageTab', site.domain)}
           className="w-full h-full border-0 overflow-hidden"
         />
       ) : (


lib/plausible/stats/dashboard/query_parser.ex
zoldar, line 1: For consistency's sake, we could adjust the module name as well. Or even change that to `dashboard/query_parser.ex` and `Plausible.Stats.Dashboard.QueryParser`, respectively. When aliasing, we can always alias `Plausible.Stats.Dashboard` and use `Dashboard.QueryParser` to maintain the context, without redundant repetition of `Dashboard.DashboardQueryParser`. This applies to other renaming instances as well.

assets/js/dashboard/components/liveview-portal.tsx
```diff
@@ -7,29 +7,58 @@
...
         style={{ width: '100%', border: '0', minHeight: MIN_HEIGHT }}
       >
-        <div
-          className="w-full flex flex-col justify-center group-has-[[data-phx-teleported]]:hidden"
-          style={{ minHeight: MIN_HEIGHT }}
-        >
-          <div className="mx-auto loading">
-            <div />
-          </div>
+        <div className={'group-has-[[data-phx-teleported]]:hidden'}>
```
zoldar, line 35: Really nice improvement in UX ✨

lib/plausible_web/live/components/dashboard/report_list.ex
```diff
@@ -145,26 +143,22 @@ defmodule PlausibleWeb.Components.Dashboard.ReportList do
...
-                  params={@params}
-                  filter={[:is, @dimension, [@item_name]]}
-                >
-                  {trim_name(@item_name, @col_min_width)}
-                </Base.filter_link>
-              </span>
+            <div class="flex justify-start items-center gap-x-1.5 px-2 py-1.5 text-sm dark:text-gray-300 relative z-9 break-all w-full">
+              <Base.dashboard_link
+                class="max-w-max w-full flex items-center md:overflow-hidden hover:underline"
+                to={Utils.dashboard_route(@site, @params, filter: [:is, @dimension, [@item_name]])}
```
zoldar, line 158: 👌

lib/plausible/stats/dashboard/utils.ex
```diff
@@ -0,0 +1,47 @@
...
+  def page_external_link_fn_for(site) do
+    with true <- Plausible.Sites.regular?(site),
+         [domain | _] <- String.split(site.domain, "/"),
+         {:ok, domain} <- idna_encode(domain),
+         {:ok, uri} <- URI.new("https://#{domain}/") do
+      fn item ->
+        "https://#{uri.host}#{hd(item.dimensions)}"
+      end
+    else
+      _ -> nil
```
zoldar, line 18: Was about to suggest that we should return  a 1-arity function returning an empty string instead, but `report_row`'s `link_fn` in indeed treated as conditional 😅

lib/plausible/stats/dashboard/utils.ex
```diff
@@ -0,0 +1,47 @@
...
+  Shared utilities by different dashboard reports.
+  """
+
+  alias Plausible.Site
+  alias Plausible.Stats.{DashboardQuerySerializer, ParsedQueryParams}
+
+  def page_external_link_fn_for(site) do
+    with true <- Plausible.Sites.regular?(site),
+         [domain | _] <- String.split(site.domain, "/"),
+         {:ok, domain} <- idna_encode(domain),
```
zoldar, line 12: 👍

zoldar:
Ah, and there's a failing test case for external link fn under CE when running against a consolidated site: https://github.com/plausible/analytics/actions/runs/20778435502/job/59669814176?pr=5986

assets/js/dashboard/components/liveview-portal.tsx
RobertJoonas: Actually this gets a bit tricky when we want to replace "Top Pages" -> "Conversion Pages" when filtering by a goal. As soon as the navigation event (let's say applying the goal filter) happens, the component instantly starts to load (optimistically, relying on the `phx-navigation-loading` class). During that optimistic loading we don't know whether the report title should be Conversion/Top pages. It requires a round-trip to the server to know whether a goal filter is applied or not.

assets/js/dashboard/index.tsx
```diff
@@ -57,6 +58,17 @@ function DashboardStats({
...
+            {
+              label: hasConversionGoalFilter(query)
+                ? 'Conversion pages'
+                : 'Top pages',
+              value: 'pages'
+            },
+            { label: 'Entry pages', value: 'entry-pages' },
+            { label: 'Exit pages', value: 'exit-pages' }
+          ]}
+          storedTab={localStorage.getItem(`pageTab__${site.domain}`)}
```
apata: nitpick, non-blocking: there's a util for getting the domain scoped key and for getting the item in storage.js module

RobertJoonas: Used the utility for getting the storage key. Getting the item needs to rely on localStorage only (not the fallback naive memory store) to be in sync with LiveView.

lib/plausible/stats/dashboard/query_parser.ex
RobertJoonas: Agreed! https://github.com/plausible/analytics/pull/5986/commits/3d5ed50ba84706c27a4a586118169b94353b15a9

assets/js/dashboard/components/liveview-portal.tsx
RobertJoonas: I got the green light to revert the dynamic report title change and keep it static as it was before a couple of days ago. Before we do that though, let's make sure there's a plan to overcome the issue with hooks in a deadview. Not a blocker to this PR anyways.

PR #5989, opened 2026-01-07
Use `assign` instead of `assign_async` for pages breakdown query result

zoldar:
### Changes

This change removes async assignment of query result in LV pages breakdown. As we rely on optimistic loading state, async assign is unnecessary and even counterproductive as it requires distinct handling of loading state (phx loading attribute is not set for the duration of loading it, it would have to be done with a hook).

Removal of async assign has uncovered a bug in navigation hook logic which led to intermittent double execution of "handle_params" event when navigating via LV clicks. This bug got addressed as well.

### Tests

### DIFF
--- a/assets/js/liveview/dashboard_root.js
+++ b/assets/js/liveview/dashboard_root.js
@@ -21,17 +21,15 @@ function navigateWithLoader(url) {
 
 export default buildHook({
   initialize() {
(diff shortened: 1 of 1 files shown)


RobertJoonas (approved):
Nice! 👌

PR #5990, opened 2026-01-07
Add team ID to debug metadata

apata:
### Changes

Adds team ID to debug metadata.

### Tests

### Changelog

### Documentation

### Dark mode


test/plausible_web/controllers/api/external_stats_controller/debug_metadata_test.exs
```diff
@@ -0,0 +1,52 @@
+defmodule PlausibleWeb.Api.ExternalController.DebugMetadataTest do
+  alias Plausible.{IngestRepo, ClickhouseRepo}
+  use PlausibleWeb.ConnCase, async: true
```
zoldar: Perhaps better to be on the safe state and run this and the other test suite with `async: false` ?

apata: Thanks, updated! https://github.com/plausible/analytics/pull/5990/commits/cc640fb3603273e14736ff348a53366380457f90

PR #6001, opened 2026-01-13
Fix feature that allows marking events as non-interactive

apata:
### Changes

After this PR, when an event is built, it allows interactive?: false value to be set if it is set in the API request.
After this PR, we will quietly ignore attempts to override any system event to be non-interactive. This is to avoid unexpected behavior. To do so, this PR refactors and expands `SystemGoals` module to `SystemEvents`.

Before this PR, the event request interactive value was ignored completely, so events built from API requests with `interactive: false` were treated as `interactive: true`. 
Existing tests didn't catch it, because session `is_bounce` related tests create events with the event factory, rather than the ingestion pipeline. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Fixed issue with all non-interactive events being counted as interactive
+
 ## v3.2.0 - 2026-01-16
 
 ### Added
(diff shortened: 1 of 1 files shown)


lib/plausible/clickhouse_event_v2.ex
```diff
@@ -70,7 +70,8 @@ defmodule Plausible.ClickhouseEventV2 do
         :revenue_source_amount,
         :revenue_source_currency,
         :revenue_reporting_amount,
-        :revenue_reporting_currency
+        :revenue_reporting_currency,
+        :interactive?
```
apata, line 74: This here was the root cause. We correctly derived what the user wanted to set the interactive? value to, but we didn't allow it to pass to the event we built. It was ignored by `ClickhouseEventV2.new/1` and the event was built with the default true value. This meant that logic to handle non-interactive events down the line never kicked in.

zoldar (approved):
LGTM. Nice work 👍

PR #6004, opened 2026-01-15
Add dashboard loading states

sanne-san:
### Changes

- Add skeleton as loading state for initial page load.
- Add loading states for tab and datepicker transitions, where data is faded and pulses until new data is loaded. 
- Speed up default Tailwind pulse animation.
- Add smooth bar width transitions for when switching between reports.

### Dark mode

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -109,12 +109,14 @@
   .animate-pulse {
     animation: pulse 1.25s cubic-bezier(0.4, 0, 0.6, 1) infinite !important;
   }
-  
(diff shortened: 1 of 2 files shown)


assets/js/liveview/live_socket.js
```diff
@@ -104,6 +104,18 @@ if (csrfToken && websocketUrl) {
   window.addEventListener('phx:page-loading-start', (_info) => topbar.show())
   window.addEventListener('phx:page-loading-stop', (_info) => topbar.hide())
 
+  const dashboardContainer = document.getElementById('live-dashboard-container')
```
zoldar, line 107: Oh, that's nice 👍 Might help simplify loading state selectors I'm currently using in my branch.

PR #6007, opened 2026-01-16
OpenTelemetry improvements

cnkk:
This PR adds:

*  `trace_id` to logs and to the `log_comment` field in the ClickHouse query log to help us correlate better between logs and traces
* extends traces with `team_id`
* move `Query.trace` to `Plausible.Stats.QueryRunner` to ensure we trace/add metadata to more queries

### DIFF
--- a/lib/plausible/clickhouse_repo.ex
+++ b/lib/plausible/clickhouse_repo.ex
@@ -47,7 +47,7 @@ defmodule Plausible.ClickhouseRepo do
   def prepare_query(_operation, query, opts) do
     {plausible_query, opts} = Keyword.pop(opts, :query)
 
-    trace_id = get_current_trace_id()
+    trace_id = Plausible.OpenTelemetry.current_trace_id()
 
     log_comment_data =
       if plausible_query do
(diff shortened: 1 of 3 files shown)


lib/plausible/open_telemetry/logger.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  @doc false
+  def handle_router_dispatch_start(_event, _measurements, _metadata, _config) do
+    case OpenTelemetry.Tracer.current_span_ctx() do
+      :undefined ->
+        :ok
+
+      span_ctx ->
+        trace_id = OpenTelemetry.Span.trace_id(span_ctx)
+        trace_id_hex = Integer.to_string(trace_id, 16) |> String.downcase()
```
zoldar: It might read a bit better with a pipeline:

```suggestion
        trace_id_hex =
          span_ctx 
          |> OpenTelemetry.Span.trace_id()
          |> Integer.to_string(16) 
          |> String.downcase()
```

Besides, this seems to be exact same logic as `ClickhouseRepo.get_current_trace_id`. I'd apply this refactor there and move the function itself under `Plausible.OpenTelemetry.current_trace_id`. WDYT?

lib/plausible/clickhouse_repo.ex
```diff
@@ -46,7 +46,16 @@ defmodule Plausible.ClickhouseRepo do
...
     {plausible_query, opts} = Keyword.pop(opts, :query)
-    log_comment = if(plausible_query, do: Jason.encode!(plausible_query.debug_metadata), else: "")
+
+    trace_id = get_current_trace_id()
+
+    log_comment_data =
+      case plausible_query do
+        nil -> %{trace_id: trace_id}
+        _ -> Map.put(plausible_query.debug_metadata, :trace_id, trace_id)
+      end
```
zoldar:
```suggestion
      if plausible_query do
        Map.put(plausible_query.debug_metadata, :trace_id, trace_id)
      else
        %{trace_id: trace_id} 
      end
```

PR #6010, opened 2026-01-20
Enforce stats requests limit by team

apata:
### Changes

Makes Stats API rate limits more strict.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/plugs/authorize_public_api_test.exs
+++ b/test/plausible_web/plugs/authorize_public_api_test.exs
@@ -95,7 +95,7 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPITest do
 
         conn =
           conn
-          |> authorize(Map.put(api_key, :key, "123"), api_scope: "stats:read:*")
+          |> authorize(api_key, api_scope: "stats:read:*")
 
         assert conn.halted
         assert json_response(conn, 400)["error"] =~ "Missing site ID."


test/plausible/rate_limit_test.exs
```diff
@@ -64,5 +65,23 @@ defmodule Plausible.RateLimitTest do
       assert {:allow, 3} = RateLimit.check_rate(@table, key, scale, limit)
       assert {:deny, 3} = RateLimit.check_rate(@table, key, scale, limit)
     end
+
+    test "denies any concurrent iterations over the limit" do
```
aerosol, line 69: was there a bug or is that because you don't trust [atomics](https://www.erlang.org/doc/apps/erts/atomics.html)?

test/plausible/rate_limit_test.exs
```diff
@@ -64,5 +65,23 @@ defmodule Plausible.RateLimitTest do
...
+      key = key()
+      scale = :timer.seconds(10)
+      limit = 3
+
+      assert eventually(fn ->
+               results =
+                 1..4
+                 |> Enum.map(fn _ ->
```
aerosol, line 77: you can use `Task.async_stream` which is parallel map effectively

test/plausible_web/plugs/authorize_public_api_test.exs
```diff
@@ -84,308 +70,504 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPITest do
...
-  end
-
-  test "logs a warning on using API key against site with no access", %{conn: conn} do
-    user = new_user(team: [hourly_api_request_limit: 1])
-    site = new_site()
+      assert conn.assigns.current_user.id == legacy_api_key_user.id
+    end
 
-    api_key = insert(:api_key, user: user)
+    test "legacy API key requests are rate limited to hardcoded 600 requests per hour _per user_, but their team's team-scoped keys are on an independent rate limit",
```
aerosol: This fails on CI. I suggest not writing very strict assertions in tests that highly depend on time. This can be reduced to a simple test that checks if rate limit of 1 is enforced. Insisting on covering everything with integration tests results in maintenance burden and poor DX. It does not improve confidence in coverage.

test/plausible/rate_limit_test.exs
apata: Neither, actually. Since the test suite is not aware that the rate limit feature is backed by atomics, I think this is an appropriate test. It ensures that a change in the underlying mechanism doesn't break the feature.

test/plausible_web/plugs/authorize_public_api_test.exs
apata: It fails for CE, which has a different hardcoded limit. 
Because the rate limit time period is 1 hour, I don't think this test is too dependent on time. By that logic, all of these tests are too dependent on time. 
Currently, user-based rate limit is hardcoded, so I can't reduce it to a shorter test, unless I make the user-based limit be configurable per environment.

lib/plausible/auth/api_key.ex
```diff
@@ -62,9 +95,7 @@ defmodule Plausible.Auth.ApiKey do
 
   def process_key(changeset), do: changeset
 
-  defp maybe_put_team(changeset, nil), do: changeset
-
-  defp maybe_put_team(changeset, team) do
+  defp put_team(changeset, team) when not is_nil(team) do
```
zoldar: Maybe move the `not is_nil(team)` guard clause to `changeset` and call `put_assoc` directly there in the pipeline?

test/plausible_web/plugs/authorize_public_api_test.exs
```diff
@@ -6,13 +6,14 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPITest do
...
-      conn
+  setup %{build_conn: build_conn} do
+    build_conn = fn ->
+      build_conn.()
       |> put_private(PlausibleWeb.FirstLaunchPlug, :skip)
       |> bypass_through(PlausibleWeb.Router)
+    end
 
-    {:ok, conn: conn}
+    {:ok, conn: build_conn.(), build_conn: build_conn}
```
aerosol: what's the reason for not recycling? and later on what's the reason for keeping an anonymous function here, instead of a regular function.

lib/plausible/auth/auth.ex
```diff
@@ -239,13 +239,18 @@ defmodule Plausible.Auth do
...
           {:ok, %{api_key: Auth.ApiKey.t(), team: Teams.Team.t() | nil}}
           | {:error, :invalid_api_key | :missing_site_id}
-  def find_api_key(raw_key, opts \\ []) do
-    {team_scope, id} = Keyword.get(opts, :team_by, {nil, nil})
+  def find_api_key(raw_key) do
+    find_api_key(raw_key, nil, nil)
+  end
 
-    find_api_key(raw_key, team_scope, id)
+  @spec find_api_key_for_team_of_site(String.t(), String.t()) ::
```
zoldar:
```suggestion
  @spec find_api_key_for_team_of_site(String.t(), String.t() | nil) ::
```

because it's expected and covered by `{:error, :missing_site_id}`, right?

test/plausible_web/plugs/authorize_public_api_test.exs
zoldar: `recycle` copies only a subset of headers, so maybe it would work there was a wrapped version of it, like:

```elixir
def recycle_preverve_ip(conn) do
  recycle(conn, ~w(accept accept-language authorization x-forwarded-for))
end
```

?

apata: Good question! 

Long story short, I couldn't get 1 test in 1 env to work right with it and decided to start from first principles: Why do we recycle the conn in the first place? It's to use the login cookie in all subsequent requests. 

In these tests, there's no need to preserve a login cookie from the user logging in, so there's no need to recycle. A freshly built conn actually resembles API requests better.

Regarding anon function, it was just a way to reuse logic while keeping DRY. It looked pretty weird and is better without though, I've refactored it in: https://github.com/plausible/analytics/pull/6010/commits/e1e30eec9606dfc5a55ac1e52c153f5c92e560b3

lib/plausible/auth/auth.ex
apata: True! 
I was planning to guard against it, but forgot. I was planning to exit as soon as we see `:site` context API request with no `site_id` param set. It can't be served, whatever the API key.

apata: On further inspection, `:missing_site_id` is not one of the possible errors from this code branch. It's emitted from another part of the verification logic, in `AuthorizePublicAPI` module. 
It's emitted after rate limiting the request. There is some sense in that, but we've already wasted resources hitting the database for a request that can't succeed. 
I think I prefer early exit, with no attempt to validate the API key.

apata: Early exit implemented: https://github.com/plausible/analytics/pull/6010/commits/108f9ac6dd21bf9b536e22b88ca04cee260ff8ee

lib/plausible/auth/api_key.ex
apata: Roger that! https://github.com/plausible/analytics/pull/6010/commits/e76b0a7134ba5c7899b8e1bd1bdb19493a6ffde6

test/plausible_web/plugs/authorize_public_api_test.exs
apata: > `recycle` copies only a subset of headers, so maybe it would work there was a wrapped version of it, like:

We actually don't want to recycle any of the headers for these requests (see my comment above). I made an attempt with `conn |> recycle([])` (_and with Enum.reduce to fix reassigning the conn_) but wasn't able to get it working right.

zoldar: That looks better 👍

zoldar:
The refactor to early exit on missing site has led to a different formatting of error in some cases: https://github.com/plausible/analytics/actions/runs/21177381163/job/60909848710?pr=6010. Other than that. LGTM.

lib/plausible/auth/auth.ex
apata: Early exit necessitates changes in external stats controller, or its tests, or both. It has its own missing site_id logic implemented.

test/plausible/rate_limit_test.exs
apata: Thanks for the suggestion! I tried to, but couldn't get it working fast - this will have to do for the moment.

PR #6011, opened 2026-01-21
Prevent "freeze" of modal during close and open and fix history handling

zoldar:
### Changes

This PR changes two things:

- It removes blocking/freezing behavior when opening and closing modals. I've made another round of manual testing with this. I'm assuming the simulated latency of 400ms is enough of a lower threshold for maintaining acceptable experience (given 2g in Firefox network throttling settings implies ~300ms). It turns out, it's not as bad, provided one keeps in mind that the possible failure modes are:
  - closing and opening the same modal in quick succession may cause the subsequent opening get canceled out by closing event. After URL update settles, next open attempt succeeds
  - closing one modal and opening a different modal in quick succession might result in lack of URL update - it will remain on root ("/") path. In practice this only matters when doing a full page refresh, and, truth be told, I don't see this being a huge pain. The modal can still be closed cleanly and subsequent navigation and open/close cycles of modals work correctly. Also, once has to be _very_ fast to manage to do that even at 400ms. So I think, except for professional Starcraft players with cpms in the hundreds, it should be an acceptable experience.
- It fixes navigation handling for modals when navigating via popstate (browser forward/backward) - now modal is opened and closed in accordance to the updated URL - that was missing

There's still some inconsistency in applying page navigation loading state and popstate navigation does not (always/sometimes) trigger the loading state either, but that's most likely something we can address with further revisions of loading logic.

### DIFF
--- a/assets/js/liveview/dashboard_root.js
+++ b/assets/js/liveview/dashboard_root.js
@@ -15,7 +15,11 @@ const MODAL_ROUTES = {
   '/utm_medium': '#utm-mediums-breakdown-details-modal'
 }
 
-function routeModal(path) {
+function routeModal(uri) {
+  // Domain is dropped from URL prefix, because that's what react-dom-router
+  // expects.
+  const path = '/' + uri.pathname.split('/').slice(2).join('/')
+
   const modalId = MODAL_ROUTES[path]
 
   if (modalId) {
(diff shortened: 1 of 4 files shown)


assets/js/liveview/dashboard_root.js
```diff
@@ -33,17 +37,23 @@ export default buildHook({
     this.portalTargets = Array.from(portals, (p) => p.dataset.phxPortal)
     this.url = window.location.href
 
+    this.addListener('phx:navigate', window, (info) => {
```
RobertJoonas, line 40: Looks like the browser history navigation is still missing some logic. When I navigate back, the forward navigation history entry gets lost. E.g:

* Open a modal
* go back in browser history (modal closes - works nice)
* go forward again in browser history - should open the modal again but cannot do that.

zoldar: Hmm, I could reproduce that once by accident, now I can't. I recall I've seen excessive prima events firing on keyboard action (like hitting escape). Maybe that's the reason? I'll have a look.

zoldar: Yeah, it's definitely off but I'm struggling to fix debug and fix it - the backend modal_close handler isn't always called even though the modal is closed. I haven't sorted out why yet. That + there are some duplicate navigation events pushed to the history. I'll shelve it for now and maybe revise later.

PR #6012, opened 2026-01-21
Refresh stats every 30s in realtime mode

RobertJoonas:
### Changes

In `Live.Dashboard`, initiate an interval timer that keeps updating a `last_realtime_update` assign (with the value of the current monotonic time) every 30 seconds. The value itself is currently unused since simply passing it down to stats reports is enough to trigger re-renders (and hence fetch new data).

However, we might need a way to tell whether a component is being updated due to a realtime tick (no loading state; smooth transition of the report bars) or a query change (go to a loading state). For that, we can compare `last_realtime_update` with `System.monotonic_time(:second)` in the components' update function:

* `is_nil(last_realtime_update)` -> `connected?(socket)` is false or period not realtime
* `last_realtime_udpate ==  System.monotonic_time(:second)` -> the update was triggered via handle_params (query changed)
* `System.monotonic_time(:second) - last_realtime_update >= 30` (where 30 is the refresh interval) -> being updated due to a realtime tick

### DIFF
--- a/lib/plausible_web/live/dashboard.ex
+++ b/lib/plausible_web/live/dashboard.ex
@@ -7,6 +7,7 @@ defmodule PlausibleWeb.Live.Dashboard do
 
   alias Plausible.Repo
   alias Plausible.Stats.{Dashboard, ParsedQueryParams}
+  alias Plausible.Stats.Dashboard.Utils
   alias Plausible.Teams
 
   @realtime_refresh_interval :timer.seconds(30)
@@ -210,6 +211,13 @@ defmodule PlausibleWeb.Live.Dashboard do
     """
   end
 
+  def handle_event("close_modal", _params, socket) do
+    close_url = Utils.dashboard_route(socket.assigns.site, socket.assigns.params)
+    socket = push_patch(socket, to: close_url)
+
+    {:noreply, socket}
+  end
+
   @modals %{
     ["pages"] => "pages-breakdown-details-modal",
     ["entry-pages"] => "entry-pages-breakdown-details-modal",
@@ -220,7 +228,7 @@ defmodule PlausibleWeb.Live.Dashboard do
   }
 
   defp maybe_close_modal(socket, old_path) do
-    if length(old_path) == 1 and socket.assigns.path == [] and Map.has_key?(@modals, old_path) do
+    if socket.assigns.path == [] and Map.has_key?(@modals, old_path) do
       Prima.Modal.push_close(socket, @modals[old_path])
     else
       socket
--- a/assets/js/liveview/dashboard_root.js
+++ b/assets/js/liveview/dashboard_root.js
@@ -15,7 +15,11 @@ const MODAL_ROUTES = {
   '/utm_medium': '#utm-mediums-breakdown-details-modal'
 }
 
-function routeModal(path) {
+function routeModal(uri) {
+  // Domain is dropped from URL prefix, because that's what react-dom-router
+  // expects.
+  const path = '/' + uri.pathname.split('/').slice(2).join('/')
+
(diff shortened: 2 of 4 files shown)


lib/plausible_web/live/dashboard.ex
```diff
@@ -209,4 +222,31 @@ defmodule PlausibleWeb.Live.Dashboard do
...
     end
   end
+
+  defp maybe_cancel_existing_realtime_timer(socket) do
+    case socket.assigns[:realtime_timer_ref] do
+      nil ->
+        socket
+
+      timer_ref ->
+        :timer.cancel(timer_ref)
```
zoldar: Ditto about preference for `Process` functions.

lib/plausible_web/live/dashboard.ex
```diff
@@ -209,4 +222,31 @@ defmodule PlausibleWeb.Live.Dashboard do
...
+        assign(socket, realtime_timer_ref: nil)
+    end
+  end
+
+  defp maybe_assign_realtime_timer(socket) do
+    case socket.assigns[:params] do
+      %ParsedQueryParams{input_date_range: :realtime} ->
+        {:ok, timer_ref} =
+          :timer.send_interval(@realtime_refresh_interval, self(), :refresh_realtime_stats)
```
zoldar: `Process.send_after` and `Process.cancel_timer` are generally more recommended as they wrap around `:erlang.send_after` and `:erlang.cancel_timer`, which are [more efficient](https://www.erlang.org/doc/system/commoncaveats#timer-module).

lib/plausible_web/live/dashboard.ex
```diff
@@ -106,6 +118,7 @@ defmodule PlausibleWeb.Live.Dashboard do
       <.live_component
         module={PlausibleWeb.Live.Dashboard.Pages}
         id="pages-breakdown-component"
+        last_realtime_update={@last_realtime_update}
```
zoldar, line 126: Don't other components querying stats, like modals, require passing this attribute to force live_component's `update` call?

lib/plausible_web/live/dashboard.ex
```diff
@@ -66,10 +68,19 @@ defmodule PlausibleWeb.Live.Dashboard do
 
     socket =
       socket
+      |> assign(last_realtime_update: nil)
```
zoldar, line 72: Shouldn't that `assign` be part of `mount` phase, or assigned with `assign_new`? 🤔

RobertJoonas: The intent was to set it to `nil` in `handle_params` when the period changes from `realtime` to any other period. Then the underlying stats components will be able to tell whether an update is caused by realtime tick by looking only at this timestamp.

lib/plausible_web/live/dashboard.ex
RobertJoonas: Yeah, they do. It didn't occur to me before, but as I just started doing it, I realised that it triggers refreshes to every single DetailsModal on the dashboard, no matter if they're open or not. Anyways, I think for the time being we can focus on realtime solely on the main dashboard and exclude details views.

lib/plausible_web/live/dashboard.ex
RobertJoonas: Nice to know, thanks! 🙏 https://github.com/plausible/analytics/pull/6012/commits/2bad2a5e99ba615cb0357bb856f0faf7ff216bac

RobertJoonas:
> Sorry, should have approved earlier - none of what I've mentioned is a showstopper for spike.

No worries! Nice shout about using Process functions instead of `:timer`. Btw also fixed the realtime refresh interval from 3s to 30s 😄

PR #6013, opened 2026-01-21
Add burst rate limit

apata:
### Changes

- Adds burst ratelimits of 60 requests per 10s.

It's not as user-friendly as counting concurrent queries by team (hundreds of 50ms queries within 10 seconds should be fair game and easily handled by the system), but it's simpler. 

### Tests

### Changelog
- [x] This PR does not make a user-facing change (CE limit is astronomical)

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/external_stats_controller/auth_test.exs
+++ b/test/plausible_web/controllers/api/external_stats_controller/auth_test.exs
@@ -129,7 +129,7 @@ defmodule PlausibleWeb.Api.ExternalStatsController.AuthTest do
     |> get("/api/v1/stats/aggregate", %{"site_id" => site.domain})
     |> assert_error(
       429,
-      "Too many API requests. Your API key is limited to 3 requests per hour."
+      "Too many API requests. The limit is 3 per hour. Please contact us to request more capacity."
     )
   end
 


lib/plausible_web/plugs/authorize_public_api.ex
```diff
@@ -54,6 +54,7 @@ defmodule PlausibleWeb.Plugs.AuthorizePublicAPI do
     with {:ok, token} <- get_bearer_token(conn),
          {:ok, api_key, limit_key, hourly_limit} <- find_api_key(conn, token, context),
          :ok <- check_api_key_rate_limit(limit_key, hourly_limit),
+         :ok <- check_api_key_burst_limit(limit_key),
```
zoldar, line 57: Shouldn't the limit key for burst rate check be different? Each call bumps the same atomic counter as in the case of the hourly limit - which would mean it will record the rate twice as high as it actually is. Or at least that's my understanding given what I see in `Plausible.RateLimit`.

apata: It’s a valid question. The full key is a tuple that includes the interval, which is different for the two limits.

PR #6018, opened 2026-01-22
Limit ClickhouseRepo queries max_execution_time

apata:
### Changes

Limits `ClickhouseRepo` queries `max_execution_time` (a https://clickhouse.com/docs/operations/settings/query-complexity) in case `cancel_http_readonly_queries_on_client_close` doesn't take effect.

Had to refactor the cases where the repo is used with infinite timeout (exports job, a cleaning job).

Tried to unset `max_execution_time` in per-query settings (https://clickhouse.com/docs/operations/settings/query-level#order-of-priority), but those don't work when read-only is set (CH server does not allow it).

### Tests
- [x] Manually tested

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -624,12 +624,13 @@ ch_transport_opts =
 config :plausible, Plausible.ClickhouseRepo,
   queue_target: 500,
   queue_interval: 2000,
+  timeout: 15_000,
   url: ch_db_url,
   transport_opts: ch_transport_opts,
   settings: [
     readonly: 1,
     join_algorithm: "direct,parallel_hash,hash",
-    # stops queries when :timeout option from above reached
+    # stops queries when :timeout ClickhouseRepo connection :timeout value reached
     cancel_http_readonly_queries_on_client_close: 1,
     # stops queries when they will likely take over 20s
     # NB! when :timeout is overridden to be over 20s,


config/runtime.exs
```diff
@@ -629,7 +629,13 @@ config :plausible, Plausible.ClickhouseRepo,
   settings: [
     readonly: 1,
     join_algorithm: "direct,parallel_hash,hash",
-    cancel_http_readonly_queries_on_client_close: 1
+    # stops queries when :timeout option from above reached
```
zoldar:
```suggestion
    # stops queries when :timeout option from below reached
```

apata: Thanks! 
It's not great wording on my part. I meant to say that this option is connected with :timeout option (max duration for the app<->db HTTP connection). I had defined it to be 15_000 ms explicitly a few lines of code above (even though the default is 15_000), but forgot to include it in the commit.

apata: Updated https://github.com/plausible/analytics/pull/6018/commits/56cbe68903d958139706fb3e1912c8a26db87632

apata:
> Ugh, I thought I approved this already 🙈

You did! I added a commit that I thought might be controversial (explicitly setting timeout value to its default value) so I re-requested it :)

PR #6024, opened 2026-01-26
Bump `ch` to 0.7.1 (via bumping `ecto_ch`) for improved DB disconnect handling

zoldar:
`ch` uses new capability added to `db_connection` (https://github.com/plausible/ch/pull/292) for graceful handling of DB disconnects.


zoldar:
@apata Hopefully less or virtually no context-less errors like "Ch.Connection (#PID<0.5417.0>) disconnected: ** (Mint.TransportError) socket closed" in our logs. I'm doing the same for persistor, where it seems to cause more issues.

PR #6034, opened 2026-01-28
Reduce calls triggered by /api/plugins/v1/capabilities

cnkk:
I noticed some super slow queries to that endpoint, which queried about 20x subscriptions and enterprise_plans
This reduces the amount of calls necessary.


apata (approved):
Good find!

PR #6035, opened 2026-01-28
Move Top Stats to new internal stats endpoint

RobertJoonas:
### Changes

This PR introduces a new `POST /api/stats/:domain/query` endpoint and makes the Top Stats use it. The endpoint will eventually replace all the Api.StatsController actions such as `main_graph`, `sources`, `pages`, etc.

* To call the new endpoint, there's a new `stats` function in `api.ts`. It takes a `StatsQuery` object which aims to mimic the query params required to construct `%ParsedQueryParams{}` on the backend.
* In order to build the `StatsQuery` object, there's also a new function called `createStatsQuery` which takes `DashboardState` and `ReportParams` objects as arguments. ReportParams contains the information about a specific report, e.g.: dimensions and metrics.

While this PR does not intend to change any behaviour, there are some functional changes with this move:

1. `bounce_rate` is now shown in page-filtered Top Stats, even when imports are included in the view. Since bounce rate from imported data is unavailable in this view, there's an appropriate warning, like with scroll depth. (This was actually already handled by https://github.com/plausible/analytics/pull/6073, but listing it here for better visibility as well)
2. `time_on_page` is now shown in page-filtered Top Stats, even when imports are included in the view. There's a specific exception where the metric is dropped on the backend though -- when legacy time on page is queried in aggregate from imported data. That's because the functionality in the `Legacy.TimeOnPage` module does not exist. This was likely the reason why time on page was simply cut out of page-filtered Top Stats when imports are included.
3.  Top Stats time labels (the ones rendered only in comparison mode) do not span into the future anymore. E.g. if today is Feb 2nd, then:

Before:

image

After:

image

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible/stats/dashboard/query_parser_test.exs
+++ b/test/plausible/stats/dashboard/query_parser_test.exs
@@ -251,10 +251,8 @@ defmodule Plausible.Stats.Dashboard.QueryParserTest do
     end
 
     test "now can't be fixed externally" do
-      {:ok, parsed} =
-        parse("?now=2026-02-17T10:08:52.272894Z", build(:site), %{
-          "now" => "2026-02-17T10:08:52.272894Z"
-        })
+      params = Map.merge(@base_params, %{"now" => "2026-02-17T10:08:52.272894Z"})
+      {:ok, parsed} = parse(params)
 
       assert parsed.now == nil
     end


lib/plausible_web/router.ex
```diff
@@ -269,6 +269,7 @@ defmodule PlausibleWeb.Router do
       end
 
       scope private: %{allow_consolidated_views: true} do
+        post "/:domain/query", StatsController, :query
```
apata, line 282: I know this is a WIP, but with a POST endpoint, we give up the ability to cache query responses client-side and in CDN. I think there's a lot to be gained in app responsiveness and scalability by even very basic caching - we shouldn't give that up.

RobertJoonas: This did cross my mind, but I deemed it unlikely that we ever want to browser/CDN cache dashboard stats requests because:

1) Stats requests depend not only on the query parameters, but server-side state: site goals, props, imported data, feature flags, permissions, etc. Sounds like quite a challenge to invalidate this cache.
2) JS memory caching (e.g. via TanStack Query that we're already using in the details modals) gives us better control over freshness and invalidation without relying on HTTP caches. Clicking around the dashboard (filters/period/with_imported) and landing on the same query -> cache hit. Refresh the entire dashboard -> everything is fresh.

apata: Hmm, but how often do site goals, props, imported data, feature flags, permissions change? 

Imagine 100 users accessing a public dashboard within one minute (e.g. a tweet with a link starts trending).

Do we want to make the 100 * 6 Clickhouse queries to populate the dashboard for each of them individually? What's the real harm in serving most of them let's say up to 2 minutes old data? It would make a difference for our servers, enabling us to scale up without adding more resources.

ukutaht: @apata raises good points. Caching has good potential for sure to reduce pressure on Clickhouse.

On the other hand, POST makes sense here because we need to be able to send a JSON payload. Making it a GET requires some sort of url serialization (bringing in jsonurl dependency elixir-side, using base64 or something like this). That can get a bit messy in terms of debuggability/logging.

Given that this is an internal endpoint that we can freely change in the future, I'm in favour of going with the simplest and most transparent option now (POST with json body). If we want to add http/cdn level caching in the future we can change it to a GET with url serialization. This is based on my assumption that changing it in the future is not a big job, limited to API client code in React and a single controller action on the backend.

RobertJoonas: Thanks both! Let's keep the door open for caching then.

> Making it a GET requires some sort of url serialization (bringing in jsonurl dependency elixir-side, using base64 or something like this)

Actually, in the current API calls it's pretty simple. E.g. filters are JSON-stringified and percent-encoded -- all the backend needs to do is `JSON.decode` them. Thinking of it, the same could be applied to the whole query params object. We could just send it to the backend like this:

`?q={period: '28d', filters: [...], dimensions: [...], order_by: [...]}` -> percent encode

So yeah, I think sticking with POST until we're in actual need of GET is a good idea and I don't think encoding will become a hustle.

assets/js/dashboard/stats/graph/fetch-top-stats.ts
```diff
@@ -0,0 +1,202 @@
...
+function constructTopStatsQuery(
+  dashboardState: DashboardState,
+  metrics: MetricDef[]
+): StatsQuery {
+  const reportParams: ReportParams = {
+    metrics: metrics.map((m) => m.key),
+    include: { imports_meta: true }
+  }
+
+  const adjustedDashboardState = { ...dashboardState }
```
apata: This stands out as a bit risky pattern to me. 

I know it's established in the repository (_we do this with queries in the backend sometimes, removing a filter or two and running the query again_), but I'd really prefer we not have something we call adjusted dashboard state.

Some fields in dashboard state are already derived with many steps (e.g. comparison depends on url params, local storage, whether segment is expanded or not). 

By adjusting the value like this freehand, we're creating potentially illegal dashboard state and passing it along. Further down the line, it's indistinguishable from the actual dashboard state. Is it the actual state or has it been adjusted? 

Just to clarify, we can remap whatever we need from dashboard state to make stats queries. It's just that after the remap, it's not dashboard state any more, it's a whole new thing. I suggest if such remapping is needed, that it be applied centrally and always, e.g. as a step in `createStatsQuery`. 

As an example, `createStatsQuery` would then depend on the actual single dashboard state + extra options (e.g. `shouldUseRealRealtimePeriod`). 
```js
function createStatsQuery(dashboardState, {reportParams, shouldUseRealRealtimePeriod}) {
  const period = // ...period remapped depending on dashboard state and the option specified for this particular query...
}
```

RobertJoonas: Agreed! Makes sense not to have `adjustedDashboardState`. Unnecessary too.

What I currently did to get rid of it though was to simply construct the query with the default params and then apply the necessary changes to the `StatsQuery` object inline. E.g.:

```javascript
const statsQuery = createStatsQuery(dashboardState, reportParams)

if (...) {
  statsQuery.include.compare = ComparisonMode.previous_period
}

if (...) {
  statsQuery.date_range = DashboardPeriod.realtime_30m
}
```

These alterations are very much ad-hoc, so I'd prefer to not extend the `createStatsQuery` function just for these. Perhaps if there will be something more repetitive.

lib/plausible/stats/dashboard/query_parser.ex
```diff
@@ -53,153 +14,107 @@ defmodule Plausible.Stats.Dashboard.QueryParser do
...
-  defp parse_match_day_of_week(_params, _user_prefs), do: true
-
-  defp parse_filters(query_string) do
-    with {:ok, filters} <- decode_filters(query_string) do
-      Plausible.Stats.ApiQueryParser.parse_filters(filters)
+  defp parse_include(params) do
+    with {:ok, compare} <- parse_include_compare(params["include"]) do
+      {:ok,
+       %QueryInclude{
+         imports: params["include"]["imports"] !== false,
```
zoldar: Strict equality check only matters when comparing integers against floats (https://hexdocs.pm/elixir/1.19.5/Kernel.html#===/2). Using `==`/`!=` should be fine enough.

Besides, this will evaluate to `true` when `params["include"]["imports"]` evaluates to `nil`, for instance. Is that intended?

lib/plausible/stats/query_result.ex
```diff
@@ -43,35 +43,22 @@ defmodule Plausible.Stats.QueryResult do
...
-            to_iso8601(query.utc_time_range.last, query.timezone)
-          ],
-          filters: query.filters,
-          dimensions: query.dimensions,
-          order_by: query.order_by |> Enum.map(&Tuple.to_list/1),
-          include: include(query) |> Map.filter(fn {_key, val} -> val end),
-          pagination: query.pagination
-        )
+      meta: meta(runner) |> Jason.OrderedObject.new(),
+      query: query(runner) |> Jason.OrderedObject.new()
```
zoldar, line 51: More of a side note - but perhaps I should have pulled the trigger on removal of these clunky serializations in https://github.com/plausible/analytics/pull/5994. 🙈  They seem to be purely for aesthetics in docs playground.

lib/plausible/stats/query_result.ex
```diff
@@ -140,6 +101,31 @@ defmodule Plausible.Stats.QueryResult do
     end
   end
 
+  defp query(%QueryRunner{site: site, main_query: query}) do
+    [
+      {:site_id, site.domain},
```
zoldar: `site_id: site.domain,` should work fine here syntax-wise, and would be a bit less noisy.

lib/plausible/stats/dashboard/query_parser.ex
RobertJoonas: > Using ==/!= should be fine enough

🤦🏻‍♂️ Always happens to me when working on JS and Elixir in parallel. Fixed in https://github.com/plausible/analytics/pull/6035/commits/19f3bf9bcc51e9ee83943a161df3f84261c1f1d7.

Also, good shout about evaluating to `true` on invalid/missing values. When we first built the `Dashboard.QueryParser` (with a LiveView dashboard in mind) there was a requirement to provide defaults matching the current dashboard behaviour.

However now that there's a `StatsQuery` on the React end, I think it makes sense for that `StatsQuery` to define default values instead. (And make `Dashboard.QueryParser` dumb about defaults). I did that in the same commit (19f3bf9bcc51e9ee83943a161df3f84261c1f1d7)

lib/plausible/stats/query_result.ex
RobertJoonas: Yeah, it's a bit annoying for sure, but I'm not *that* bothered by it. But if we can find a way to get rid of it, all fine by me! 😄

lib/plausible/stats/query_result.ex
RobertJoonas: https://github.com/plausible/analytics/pull/6035/commits/06c15fa941d723dfc1d5657ff92613f84a5f8cca

assets/js/dashboard/stats/graph/fetch-top-stats.test.ts
```diff
@@ -0,0 +1,251 @@
...
+  /** input dashboard state & site state */
+  Pick<DashboardState, 'filters' | 'period'> &
+    Partial<{ site?: Pick<PlausibleSite, 'revenueGoals'> }>,
+  /** expected metrics */
+  MetricDef[],
+  /** expected queries */
+  [StatsQuery, null | StatsQuery]
+]
+
+const cases: TestCase[] = [
```
zoldar, line 49: That might be a matter of preference but IMHO these tests would be much more readable if they were written explicitly case by case - the boilerplate is pretty much minimal and the whole looping and generation of test cases obfuscates the tests, at least for me. I had to spend a good while deciphering destructuring at the bottom, scrolling back and forth, to make sense of it.

zoldar:
@RobertJoonas I have rebased this branch against the (pending) e2e test suite and I have stumbled on one issue:

image

When switching to realtime, it appears that top stats metrics IDs for current visitors and visitors from the last 30 minutes have the same ID ("visitors"). Before the change, "Current visitors" metrics has a distinct ID of "current_visitors".

assets/js/dashboard/stats/graph/fetch-top-stats.test.ts
RobertJoonas: @zoldar. I was thinking to refactor these tests to call fetchTopStats directly (instead of testing both `topStatsQueries` and `chooseMetrics` separately). It will use `mock-api.ts`. With that, I would also separate all test cases. However, me and @apata are currently pairing on main graph related stuff and would like to merge this. I'll refactor the tests as a follow-up.

RobertJoonas:
> @RobertJoonas I have rebased this branch against the (pending) e2e test suite and I have stumbled on one issue:

@zoldar, I have addressed this with a hotfix for now. https://github.com/plausible/analytics/pull/6035/commits/abcdabc5b2aa62766fef4fc20d352b3400ddd0df.

Planning to drop the separation of topStatsQuery and currentVisitorsQuery in a follow-up. The current visitors could just be an integer value in the formatted Top Stats structure. E.g: `{topStats: [...], currentVisitors: integer | null}`

zoldar:
@RobertJoonas nice SHA :D

image

PR #6037, opened 2026-01-29
Fix map on Chrome

apata:
### Changes

Fixes bug on Chrome with map countries not being clickable and staying highlighted. 

A bit of a refactor was needed, because the stratagem to bring the country to the front with appendChild was the problem. It was called too many times and caused event listeners for mouseout / mousemove to not fire. 

I rewrote it so only one country can be highlighted at a time using a non-interactive single overlay path. It's updated to match the shape of the country being hovered.

### Tests
- [x] Manually tested

### Changelog

### Documentation

### Dark mode


aerosol (approved):
double checked that it does fix the issue 🚀

PR #6038, opened 2026-01-29
Implement basics for E2E tests driven by playwright

zoldar:
### Changes

This PR implements basis for running end-to-end dashboard tests using playwright.

- A distinct `e2e_test` MIX environment is introduced for running the tests
- The end-to-end tests live under ./e2e directory, with playwright and all dependencies installed directly there
- A test specific seeds file is put under `./priv/repo/e2e_seeds.exs` - it might be subject to change if we decide to change out approach to setting up system state for tests - it's being discussed internally
- End-to-end tests are run as a part of CI pipeline; screenshots for failed tests uploaded to GH artifacts

The tests can be run locally in two ways:
- `mix test.e2e` - standard mode outputting test report in the terminal
- `mix test.e2e.ui` - interactive mode running in a dedicated window, with ability to run tests and review the outputs selectively

When run on CI, the tests are sharded (just two shards for now). 

Failed test report for a single shard:

https://github.com/plausible/analytics/actions/runs/21710287043/job/62611828894

Full test report summary available under `merge-sharded-e2e-test-report` job (after expanding "Merge into list report"):

https://github.com/plausible/analytics/actions/runs/21710287043/job/62612031552

Screenshots from failed tests on CI are available as downloadable artifacts from CI worflow summary page, like this one: 

https://github.com/plausible/analytics/actions/runs/21710287043

image

While we could work towards fancier report output, I'd opt for keeping things simple for now and not adding en explicit report the the PR thread just yet, to reduce noise.

There are only a couple example tests for now. Follow-up work will involve expanding it into a more comprehensive set of test scenarios.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -215,7 +215,7 @@ jobs:
       - name: Install E2E Playwright Browsers
         if: steps.playwright-cache.outputs.cache-hit != 'true'
         working-directory: ./e2e
-        run: npx playwright install --with-deps=chromium
+        run: npx playwright install --with-deps chromium
 
       - name: Run E2E Playwright tests
         run: npm --prefix ./e2e test -- --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }} --reporter=blob


.github/workflows/elixir.yml
```diff
@@ -113,6 +113,167 @@ jobs:
...
+            tracker:
+              - 'tracker/**'
+
+      - name: Check if tracker and verifier are built already
+        run: |
+          if [ -f priv/tracker/js/plausible-web.js ] && [ -f priv/tracker/installation_support/verifier.js ]; then
+            echo "HAS_BUILT_TRACKER=true" >> $GITHUB_ENV
+          else
+            echo "HAS_BUILT_TRACKER=false" >> $GITHUB_ENV
+          fi
```
apata: question, non-blocking: is building the tracker conditionally necessary? shouldn't it be built always and the result cached if its expensive?

zoldar: You are probably right, that's copypasta from tracker CI config which is unnecessary in this context.

.github/workflows/elixir.yml
```diff
@@ -113,6 +113,167 @@ jobs:
...
+          path: e2e/blob-report
+          retention-days: 1
+
+      - name: Upload E2E screenshots of failed tests
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: e2e-screenshots-${{ matrix.shardIndex }}
+          path: e2e/output
+          retention-days: 1
```
apata: suggestion, non-blocking: I think we should skip this job in the initial version. I think there's some risk with it exposing more than we intend

zoldar: Alright 👍 I'm not sure what we could leak there though? 🤔 Either way, I'll remove it from the first iteration.

zoldar: https://github.com/plausible/analytics/pull/6038/commits/2b5a4f4fbee8f3c1c30ed85e02d93cccceb8e80b

.github/workflows/elixir.yml
zoldar: https://github.com/plausible/analytics/pull/6038/commits/0564e8a1d6dd6dfdaecc2310799a08333fc9d517

PR #6039, opened 2026-01-30
Do not access `location` during import of tracker module

toofishes:
### Changes

After making this change, I am able to use server-side generation of a Vue-based website using vite-ssg and the node package, with a static import of the analytics code so it gets inlined and packed into the main JS bundle.

#### Initialize currentEngagementURL to undefined, not `location.href`

For use in a server-side rendered environment, it is important we don't access browser variables before `init` or `track` are called, that might not exist in Node/Bun/etc.

There isn't much reason to set this variable to location.href prior to `postPageviewTrack` being called, as we won't use it until engagements are registered, in which case, we will have set the URL from the payload.

This is the current error when importing (but not running) the code in Node:

```
Welcome to Node.js v25.4.0.
Type ".help" for more information.
> import('@plausible-analytics/tracker/plausible.js')
Promise {
  <pending>,
  Symbol(async_id_symbol): 103,
  Symbol(trigger_async_id_symbol): 6
}
> Uncaught ReferenceError: location is not defined
```

After this change, the import is successful:

```
Welcome to Node.js v25.4.0.
Type ".help" for more information.
> await import('@plausible-analytics/tracker')
[Module: null prototype] {
  DEFAULT_FILE_TYPES: [
    'pdf', 'xlsx', 'docx', 'txt',
...
    'wma', 'dmg'
  ],
  init: [Function: init],
  track: [Function: track]
}
```

### Tests

### Changelog

I didn't update the tracker changelog- this seems like something done at release time, when I read the tracker-specific docs.

### Documentation

### Dark mode

### DIFF
--- a/tracker/src/engagement.js
+++ b/tracker/src/engagement.js
@@ -9,7 +9,7 @@ import { sendRequest } from './networking'
 var listeningOnEngagement = false
 
 var currentEngagementIgnored
-var currentEngagementURL = location.href
+var currentEngagementURL
 var currentEngagementProps = {}
 var currentEngagementMaxScrollDepth = -1
 


toofishes:
I'm not sure if I'm able to get the `tracker-script-update` CI check to successfully run - it appears to be needing a token. Not sure if this is something broken specifically because I'm PRing from a fork, or if it is broken on the master branch.

toofishes:
I updated this PR after the module export changes (#6158) went in. Any chance it can get looked at?

gurpreetatwal:
@apata  / @cnkk Anything that can be done to get this PR reviewed?

apata:
@toofishes Thanks for your contribution!

It is a real issue, as the package readme (https://www.npmjs.com/package/@plausible-analytics/tracker) is supportive of SSR, yet it doesn't work with `vite-ssg`. Does it work with other SSR solutions? Why aren't more people complaining of this?

tracker/src/engagement.js
```diff
@@ -9,7 +9,7 @@ import { sendRequest } from './networking'
 var listeningOnEngagement = false
 
 var currentEngagementIgnored
-var currentEngagementURL = location.href
+var currentEngagementURL = ''
```
apata: What happens if this is left undefined?

toofishes: That is probably fine as well. I wasn't sure whether to prefer the pattern of `currentEngagementProps`/`currentEngagementMaxScrollDepth` (initializing to something of the same type) or `currentEngagementIgnored` (just leaving undefined). Would you like me to update the PR and change it?

toofishes:
> @toofishes Thanks for your contribution!
> 
> It is a real issue, as the package readme (https://www.npmjs.com/package/@plausible-analytics/tracker) is supportive of SSR, yet it doesn't work with `vite-ssg`. Does it work with other SSR solutions? Why aren't more people complaining of this?

I was also a bit surprised to be reporting it! Likely either use of jsdom/happy-dom whether on purpose or accidentally, deferred loading, or conditionally excluding it from SSR?

tracker/src/engagement.js
apata: Roger, I think that was a reasonable approach! 

On my end, the main concern is what will happen if this value ends up being sent to the API (for whatever future reason). 

I removed the `listeningOnEngagement` condition in `prePageviewTrack` (simulating us screwing up in the future) and tested `undefined`, `null` and `''` values for `currentEngagementURL`. In all cases, tests started failing (great!). 

The `u` value in the payload was `undefined`, `null` or `''` respectively. When undefined or null, the api/event endpoint rejects the event with 400. When it was `''`, the endpoint returns 202, but drops it due to not finding a matching session (`x-plausible-dropped: 1`). 

I think it's better to see the error 400 tracker-side, so I'd prefer it to be uninitialized. If you agree, please update the PR.

toofishes:
PR updated to simply leave it undefined by default.

apata:
@toofishes Sorry to have you jumping through hoops, but can you do the CLA thing again? I couldn't take over your branch / PR due to "Allow maintainer edits" not being set to true, so I cherry picked it and had to fix some workflows as well. https://github.com/plausible/analytics/pull/6573#issuecomment-5180470803

toofishes:
All good, and I believe the CLA stuff should be taken care of now. Your PR looks good to me!

ukutaht:
Superceded by https://github.com/plausible/analytics/pull/6573

PR #6045, opened 2026-02-02
fix: hide views_per_visit when event-only filters are applied

### LINKED ISSUE #5804: Incorrect 'Views per visit' calculation when using event-based filters
### Past Issues Searched

### Issue is a Bug Report

Self-hosting

### Describe the bug

The "Views per visit" metric appears to be calculated incorrectly when an event-level filter (such as a custom property) is applied. While the "Total Pageviews" and "Total Visits" metrics seem to correctly reflect the filtered data, the derived "Views per visit" metric is inflated and misleading.

I suspect the calculation is using the total pageviews from the entire session for any session that matches the filter, instead of using only the pageviews that actually match the filter.

I have attached two screenshots to illustrate this:

imagen.png: The main dashboard without any filters. The calculation is correct: 202,000÷78,800≈2.56.

Filter.png: The dashboard with a filter Property Website is youth applied. The calculation shown is 5.66, but the expected result should be 4,400÷1,700≈2.59.

**Screenshots (see below):**

1. No Filter (Correct Calculation)

2. With Property Website is youth Filter (Incorrect Calculation)

**Detailed Analysis**

The core of the issue lies in the data being used for the numerator in the "Views per visit" calculation when a filter is active.

Total Pageviews (4.4k): This number is correct. It counts only the pageview events that match the filter condition (Property Website is youth).

Total Visits (1.7k): This number is also correct. It counts every unique visit (session) in which at least one pageview event matched the filter condition.

Views per visit (5.66): This is where the error occurs. The calculation being performed seems to be dividing the total number of pageviews from those 1.7k visits by the visit count (1.7k), rather than using the filtered pageview count of 4.4k.

The current, incorrect calculation is:
Views per visit=Number of matching visitsTotal pageviews from all matching visits​

The expected, logical calculation should be:
Filtered VisitsFiltered Pageviews​=1,7004,400​≈2.59

**Illustrative Example**

Consider a single visit consisting of 10 total pageviews.

Only 2 of these pageviews have the property Website is youth.

The other 8 do not.

When the filter is applied, Plausible correctly identifies this visit as matching and counts it:

Total Pageviews: 2 (the ones that match the filter)

Total Visits: 1 (because the session contains at least one matching pageview)

Views per visit: To calculate this, the system appears to take the matching visit and sum up all its associated pageviews (10), not just the 2 filtered ones. This results in a value of 10 for this single visit.

When scaled across thousands of visits, this inflates the numerator and produces the misleading 5.66 figure instead of the expected 2.59.

**Root Cause in the Code**

The provided source code confirms this behavior. The metric is calculated from the pre-aggregated sessions table (s).

defmodule Plausible.Stats.SQL.Expression
Elixir

```Elixir
def session_metric(:views_per_visit, _query) do
  wrap_alias([s], %{
    views_per_visit:
      fragment(
        "greatest(ifNotFinite(round(sum(? * ?) / sum(?), 2), 0), 0)",
        s.sign,
        s.pageviews,  # <-- THIS IS THE PROBLEM
        s.sign
      ),
    __internal_visits: fragment("toUInt32(greatest(sum(sign), 0))")
  })
end
```

The SQL logic is effectively sum(s.pageviews) / sum(s.sign).

sum(s.sign) correctly counts the total number of visits that match the filter.

However, s.pageviews is a column in the sessions table that stores the total number of pageviews for the entire session, regardless of any event-level filters.

When the query filters for sessions that contain a specific event property, it correctly selects the right sessions. But it then sums the entire s.pageviews value for those selected sessions, leading to the incorrect calculation.

In summary, this is a logical error in how derived metrics are calculated on pre-aggregated session data when event-level filters are applied. The calculation should use the filtered event count, not the total event count from the session.

### Expected behavior

The expected, logical calculation should be:
Filtered VisitsFiltered Pageviews​=1,7004,400​≈2.59

### Screenshots

Without filter

Image

With filter

Image

### Environment

```markdown
- OS:
- Browser:
- Browser Version:
```

Sayan-CtrlZ:
Fixes #5804

When event-level filters (event:name, event:goal, event:props:*) are applied, the views_per_visit metric was incorrectly calculated because it uses session-level pageview counts instead of filtered event counts.

This change hides the views_per_visit metric from the dashboard top stats when event-only filters are present, consistent with:
- The External Stats API which already blocks this combination
- How the dashboard handles event:page filters

Added tests for event:name, event:props, and event:goal filter cases.

### Changes

- Added `has_event_only_filter?/1` helper function in `stats_controller.ex` to detect event-only filters
- Modified `fetch_other_top_stats/2` to exclude `views_per_visit` when event-only filters are present
- Added 3 new tests in `top_stats_test.exs` to verify the behavior

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 
 - Fixed issue with all non-interactive events being counted as interactive
 - Fixed incorrect 'Views per visit' calculation when event-only filters (event:name, event:goal, event:props) are applied
+- Fixed countries map countries staying highlighted on Chrome
 
 ## v3.2.0 - 2026-01-16
 
--- a/assets/js/dashboard/stats/locations/map.tsx
+++ b/assets/js/dashboard/stats/locations/map.tsx
@@ -20,6 +20,7 @@ import { apiPath } from '../../util/url'
 import { MIN_HEIGHT } from '../reports/list'
 import { MapTooltip } from './map-tooltip'
 import { GeolocationNotice } from './geolocation-notice'
+import { DashboardQuery } from '../../query'
 
 const width = 475
 const height = 335
@@ -31,6 +32,16 @@ type CountryData = {
   code: string
 }
 type WorldJsonCountryData = { properties: { name: string; a3: string } }
+
+function getMetricLabel(query: DashboardQuery) {
+  if (hasConversionGoalFilter(query)) {
+    return { singular: 'Conversion', plural: 'Conversions' }
+  }
+  if (isRealTimeDashboard(query)) {
+    return { singular: 'Current visitor', plural: 'Current visitors' }
+  }
+  return { singular: 'Visitor', plural: 'Visitors' }
+}
 
 const WorldMap = ({
   onCountrySelect,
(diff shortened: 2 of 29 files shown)


ukutaht:
Thank you for the PR. As [explained](https://github.com/plausible/analytics/issues/5804#issuecomment-3873420532) in the issue we currently want to keep things as-is.

PR #6049, opened 2026-02-04
Fix mixed goals conversions crash

aerosol:
### Changes

When goals with custom properties are mixed with goals without custom properties, ecto_ch only checks the first element of an array to infer its type. When goals are ordered such that those without custom props come first in the array, the custom_props_keys array looks like: `[[],[],[],['url']]` which results with `Array(Nothing)`, which causes ClickHouse to fail with:

```
Serialization is not implemented for type Nothing: value [[],[],[],['url']] cannot be parsed as Array(Array(Nothing))
```

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/prom_ex/striped_peep.ex
+++ b/lib/plausible/prom_ex/striped_peep.ex
@@ -7,7 +7,7 @@ defmodule Plausible.PromEx.StripedPeep do
 
   @impl true
   def scrape(name) do
-    Peep.get_all_metrics(name)
+    (Peep.get_all_metrics(name) || [])
     |> Peep.Prometheus.export()
     |> IO.iodata_to_binary()
   end


aerosol:
@ruslandoga :wave: solution in this PR is trying to work around https://github.com/plausible/ecto_ch/blob/9a517fe9eff14c1cfae6fc8e0c6ebfe372b032e6/lib/ecto/adapters/clickhouse/connection.ex#L1234 

The desperate fix would be something like:

```diff
   defmacro event_goal_join(goal_join_data) do
     quote do
@@ -467,8 +474,8 @@ defmodule Plausible.Stats.SQL.Expression do
             ?,
             ?,
             ?,
-            ?,
-            ?
+            arraySlice(?, 2),
+            arraySlice(?, 2)
           )
         )
         """,
@@ -481,8 +488,14 @@ defmodule Plausible.Stats.SQL.Expression do
         type(^unquote(goal_join_data).event_names_by_type, {:array, :string}),
         type(^unquote(goal_join_data).scroll_thresholds, {:array, :integer}),
         type(^unquote(goal_join_data).indices, {:array, :integer}),
-        type(^unquote(goal_join_data).custom_props_keys, {:array, {:array, :string}}),
-        type(^unquote(goal_join_data).custom_props_values, {:array, {:array, :string}})
+        type(
+          ^[["__TRICK_ECTO_CH__"] | unquote(goal_join_data).custom_props_keys],
+          {:array, {:array, :string}}
+        ),
+        type(
+          ^[["__TRICK_ECTO_CH__"] | unquote(goal_join_data).custom_props_values],
+          {:array, {:array, :string}}
+        )
       )
     end
   end
```

Do you think you know how to properly fix this at source?

ruslandoga:
👋 @aerosol 

I think it's time to do that TODO.

I'd like to open a PR with something like this, but maybe a bit cleaner:

```elixir
defp param_type([]), do: "Array(Nothing)"
  
defp param_type([v | vs]) do
  param_type = param_type(v)
    
  if param_type == "Array(Nothing)" do
    param_type(vs)
  else
    ["Array(", param_type, ?)]      
  end
end
```

aerosol:
@ukutaht awful workaround implemented to unblock customers. Tracking ecto_ch at https://github.com/plausible/ecto_ch/issues/262

PR #6054, opened 2026-02-09
Setup state directly from playwright in E2E tests

zoldar:
### Changes

This PR changes how E2E tests state setup is handled. Instead of using a single seed, each test explicitly sets up system state by executing a sequence of operations via UI. One exception is populating stats. In that case, a request is made to a dedicated endpoint with a JSON payload describing events to add for ingestion for a given site. The endpoint also ensures that site's stats start date is updated so that the populated stats are accounted for.

- When writing the setup helpers, a couple accessibility issues got fixed along the way. Playwright's role based locators heavily rely on the notion of [accessible names](https://developer.mozilla.org/en-US/docs/Glossary/Accessible_name) for looking up elements on the page in a flexible manner
- Initially, I wanted to use test fixtures as described in https://playwright.dev/docs/test-fixtures. I have tried (and failed) to make it work in https://github.com/plausible/analytics/pull/6054/commits/c16c77cf620104f78233c94ed5222b186aa8af32. The examples in the documentation seem to be inconsistent at times and my TS-fu is not great. Eventually, I have decided to go with simple helper functions. We don't gain much from the fixtures - I'd even say that explicit helpers work better as the setup is more explicit and not hidden behind a magic "test" wrapper
- One caveat I got bitten by - when navigating to an LV-driven site, it's strongly advisably [to wait for LV to connect](https://github.com/plausible/analytics/pull/6054/changes#diff-8b637c36d60c6df2fcebb83e1b1fe743e6024053053816275bcdd64cd1b5c71dR19). Otherwise, playwright might start filling up the forms before websocket connection is established and inputs get reset in the middle - and playwright tries to submit an incomplete form.
- A peculiar quirk came up when [removing seeds](https://github.com/plausible/analytics/pull/6054/commits/928ceaf754381d52cf64bc783c325e047d588df6). At first, tests started failing for me, because without `mix run`, the consecutive `mix phx.server` was running under "dev" `MIX_ENV`. Once I've fixed it, I have noticed that the tests execute noticeably slower - by about 1-2 seconds on average (5.5-6.5s instead of 3.5-5s). It turned out that even adding a simple noop mix run task speeds things up. I'm really not sure what's the exact reason for this (something to do with application startup? All tasks within alias are executed in a single mix do command internally). Either way, I'm leaving the noop task there for now.

The tests are noticeably slower than in the case of seed driven ones but the difference is not staggering. On CI it's 12-15 seconds vs  9-11 seconds for the current (however small) sample of tests per shard. We'll see how that works out in practice as the test suite grows.

### Tests

### DIFF
--- a/e2e/tests/general.spec.ts
+++ b/e2e/tests/general.spec.ts
@@ -131,7 +131,7 @@ test('tab selection user preferences are preserved across reloads', async ({
     domain
   )
 
-  await expect(currentTab).toEqual('entry-pages')
+  expect(currentTab).toEqual('entry-pages')
 
   await page.getByRole('button', { name: 'Exit pages' }).click()
 
@@ -142,7 +142,7 @@ test('tab selection user preferences are preserved across reloads', async ({
     domain
   )
 
-  await expect(currentTab).toEqual('exit-pages')
+  expect(currentTab).toEqual('exit-pages')
 })
 
 test('back navigation closes the modal', async ({ page, request, baseURL }) => {


e2e/tests/dashboard-fixtures.ts
```diff
@@ -0,0 +1,175 @@
...
+    await this.page.goto("/activate");
+
+    await this.page.locator("input[name=code]").fill(code);
+
+    await this.page.getByRole("button", { name: "Activate" }).click();
+
+    await expect(
+      this.page.getByRole("button", { name: "Install Plausible" }),
+    ).toBeVisible();
+  }
```
apata: _suggestion, non-blocking:_ 
Since we don't store any state in the class, I think it might be better to declare all these as independent functions, outside a class. Not just for brevity (page.foo vs this.page.foo), but also not to make these functions interdependent (better to have "pure" functions). 

A test would look something like this then

```ts
import { login, register, addSite, makeSitePublic, populateStats } from "./fixtures"
  
test("pages report renders", ({ page, request }) => {
  // setup
  const user = register({ page }) // creates unique user
  login({ page, user })
  const site = addSite({ page, user })
  populateStats({ page, request, site }, [])
  // actual test
  await page.goTo(...)
  await expect...
})
```

e2e/tests/dashboard-fixtures.ts
```diff
@@ -0,0 +1,175 @@
...
+export class AuthPage {
+  constructor(
+    public readonly page: Page,
+    public readonly request: Request,
+  ) {}
+
+  async register(user: User) {
+    await this.page.goto("/register");
+
+    await expect(this.page.locator(".phx-connected")).toHaveCount(1);
```
apata: suggestion, non-blocking: extract to `function expectLiveViewConnected({page}) {...}`

apata:
>  explicit helpers work better as the setup is more explicit and not hidden behind a magic "test" wrapper

I agree, I think you picked a good approach. After all, we might want to set up N sites and N users for a particular test.

apata (approved):

e2e/tests/general.spec.ts
```diff
@@ -1,19 +1,43 @@
 import { test, expect } from "@playwright/test";
+import { setupSite } from "./dashboard-fixtures.ts";
```
apata: suggestion, non-blocking: Just ./fixtures.ts seems a bit more accurate, since they go through the rest of the app as well, not just React dashboard

apata: Or maybe have 
- `./fixtures/user.ts` actions in account menu / team menu
- `./fixtures/site.ts` actions in site settings menu / on dashboard

e2e/tests/dashboard-fixtures.ts
```diff
@@ -0,0 +1,175 @@
...
+      (e) =>
+        e.to[0][0] === user.name &&
+        e.subject.indexOf("is your Plausible email verification code") > -1,
+    );
+
+    expect(emails.length).toEqual(1);
+
+    const [code] = emails[0].subject.split(" ");
+
+    await this.page.goto("/activate");
```
apata: nitpick, non-blocking: The browser is already at this page (asserted on L38), the test is about 400ms faster locally upon removing this.

e2e/tests/dashboard-fixtures.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    await this.page
+      .getByRole("form", { name: "Make stats publicly available" })
+      .getByRole("button")
+      .click();
+
+    await expect(this.page.locator("body")).toContainText("are now public");
+  }
+
+  async populateStats(domain: string, events: object[]) {
```
apata: suggestion, non-blocking: In the future, let's declare the interface/type for event, because we know exactly the shape. `object` is very general thing in Javascript and isn't recommended for use by Typescript lint rules usually (`typeof null === 'object'`)

zoldar: Got it, thanks. I will do that for a subset of commonly used fields (and extend it as needed). I initially copped out at the prospect of speccing the whole event struct 😅

zoldar:
@apata Thanks for all-round great feedback 🙌

e2e/tests/dashboard-fixtures.ts
zoldar: Ah, I was using `/sent-emails` endpoint with `page` before and left it, good spot

e2e/tests/dashboard-fixtures.ts
zoldar: https://github.com/plausible/analytics/pull/6054/commits/b1ed41e075cb8dd7959ff09d5a7917849b95aba7

I have left site setup helper though, as a single site will be used in majority of tests, so it's pretty convenient.

e2e/tests/dashboard-fixtures.ts
zoldar: https://github.com/plausible/analytics/pull/6054/commits/455ad5959e7e8afa02abd62907a614fdc4dd0db0

e2e/tests/general.spec.ts
zoldar: Done in https://github.com/plausible/analytics/pull/6054/commits/b1ed41e075cb8dd7959ff09d5a7917849b95aba7 - leaving everything in `/fixtures.ts` for now.

e2e/tests/dashboard-fixtures.ts
zoldar: Removed in https://github.com/plausible/analytics/pull/6054/commits/7aa3255e77ea8e6598c6b7e086d3bc3ed48cad7c

e2e/tests/dashboard-fixtures.ts
zoldar: Basic type added in https://github.com/plausible/analytics/pull/6054/commits/ba94b97d1d29ea28d33afb7231ca7fcfca2003e5

zoldar:
@apata Thanks again for all the nits and suggestions. In addition to your feedback, I have also configured prettier and eslint, with settings similar to tracker for consistency, addressing feedback from both after that.

e2e/eslint.config.mjs
```diff
@@ -0,0 +1,46 @@
...
+        'error',
+        {
+          allowShortCircuit: true
+        }
+      ]
+    }
+  },
+  // config for tests of tracker scripts and installation support scripts
+  {
+    files: ['test/**/*.js', 'test/**/*.ts'],
```
apata: issue, non-blocking: this particular section matches files in `test/` folder, which e2e package doesn't have. Probably should be files: `['tests/**/*.ts']`

e2e/tests/dashboard-fixtures.ts
apata: > I have left site setup helper though, as a single site will be used in majority of tests, so it's pretty convenient.

True! Good idea

apata:
> In addition to your feedback, I have also configured prettier and eslint, with settings similar to tracker for consistency, addressing feedback from both after that.

Nice! I think it's a good idea. Is it enforced in CI yet? If not, no biggie, can be done in future PRs as well.

zoldar:
@apata I don't see it enforced in tracker CI either so I didn't push for. But we can add that in future updates, sure.

e2e/eslint.config.mjs
zoldar: Weirdly, it worked on contents of tests, regardless of this, huh. Will fix though.

zoldar: https://github.com/plausible/analytics/pull/6054/commits/e1db0089e48cdb4c34d516f183b6f788f08d254c - and yeah, seems like not _all_ rules have been enforced without this pattern fixed, hah, https://github.com/plausible/analytics/pull/6054/commits/932be2a6bd1cddcf50a7c2a02987eeec42119161

PR #6056, opened 2026-02-10
Add usage notifications for billing limits

sanne-san:
### Changes

- As a first step towards improving the subscription settings page, this commit adds usage notifications for billing limits. These are not yet being displayed in the UI.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/billing/notice.ex
```diff
@@ -265,6 +265,240 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+  1. Dashboard locked (grace period expired after 2 cycles exceeded)
+  2. Trial ended (no subscription after trial expires - blocks all access)
+  3. Traffic exceeded for 2 consecutive cycles
+  4. Traffic exceeded for 1 cycle
+  5. Pageview limit approaching (90%+) - takes precedence over site/member limits
+  6. Site and team member limits both reached
+  7. Site limit reached
+  8. Team member limit reached
+  """
+  def determine_notification_type(
```
RobertJoonas: I like the idea of a single function determining the priority of usage notifications. Makes it easier to display a single notice at a time which makes perfect sense to not overwhelm customers! 👏

Codewise, seeing a few possible improvements here:

1. Moving the function
 
I think this function could move elsewhere, so that this module, located at `lib/plausible_web/components` only defines actual components.

A suitable place would perhaps be the `Plausible.Billing.Quota` module, since it's already dealing with usage and limits. The name could also change, perhaps to `usage_notification_type`?

2. We should try to cut down on the number of arguments to make the function signature more readable. Especially in tests.

My proposal is to keep only two arguments: `usage_notification_type(team, usage)`

Every argument is derivable from `team` (see the `subscription` function in `PlausibleWeb.SettingsController`), but the reason I would keep `usage` separate is being able to test it more easily. Generating actual usage in tests is cumbersome.

Note that the `usage` argument would become a map with all the different usage aspects included in it. There's the `Plausible.Teams.Billing.quota_usage` function that returns everything, including `team_members`, `sites` and `monthly_pageviews`.

So in `PlausibleWeb.SettingsController`, I imagine we could be passing only the following assigns to the `subscription.html.heex` template:

```
def subscription(conn, _params) do
    team = conn.assigns.current_team
    subscription = Teams.Billing.get_subscription(team)
    usage = Teams.Billing.quota_usage(team)

    render(conn, :subscription,
      layout: {PlausibleWeb.LayoutView, :settings},
      team: team,
      subscription: subscription,
      usage: usage
    )
  end
```

lib/plausible_web/components/billing/notice.ex
```diff
@@ -265,6 +265,240 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+        site_limit,
+        team_member_usage,
+        team_member_limit,
+        subscription
+      ) do
+    cond do
+      Plausible.Teams.GracePeriod.expired?(team) ->
+        :dashboard_locked
+
+      Plausible.Teams.on_trial?(team) == false and is_nil(subscription) ->
```
RobertJoonas:
```suggestion
      not Plausible.Teams.on_trial?(team) and is_nil(subscription) ->
```

lib/plausible_web/components/billing/notice.ex
```diff
@@ -265,6 +265,240 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+        :traffic_exceeded_sustained
+
+      is_map_key(pageview_usage, :last_cycle) and
+        pageview_usage.last_cycle.total > pageview_limit and
+          pageview_limit != :unlimited ->
+        :traffic_exceeded_last_cycle
+
+      pageview_limit != :unlimited and is_map_key(pageview_usage, :current_cycle) and
+          pageview_usage.current_cycle.total >= pageview_limit * 0.9 ->
+        :pageview_approaching_limit
```
RobertJoonas: Seeing an opportunity to 1) extract some predicate functions and 2) merge these three `cond` branches into a single one that handles billing cycles in a separate function.

Also, the pageview limit is only `:unlimited` for trials, so instead of checking whether the pageview_limit is unlimited, we could be checking whether the team is on trial, with the already existing predicate function. Here's what I'm thinking:

```elixir
cond do
  # ...
  not Plausible.Teams.on_trial?(team) and has_billing_cycles?(usage.monthly_pageviews) ->
    pageview_cycle_usage_notification_type(usage.monthly_pageviews, monthly_pageview_limit)
  # ...
end

defp pageview_cycle_usage_notification_type(usage, limit) do
  exceeded = Plausible.Billing.Quota.exceeded_cycles(usage, limit)

  cond do
    :penultimate_cycle in exceeded and :last_cycle_exceeded ->
      :traffic_exceeded_sustained
    :last_cycle in exceeded ->
      :traffic_exceeded_last_cycle
    usage.current_cycle >= limit * 0.9 ->
      :pageview_approaching_limit
  end
end
```

The `has_billing_cycles?/1` function could be a public predicate function in the `Plausible.Billing.Quota` module, and could also be used in the `exceeds_monthly_pageview_limit?` function (where the behaviour depends on whether the usage is in billing cycles or last 30 days). Using predicates like that is a good practice as it'll give us a better overview in the code about where this specific logic matters.

About the `pageview_cycle_usage_notification_type` function -- I think we should opt to use the actual limit checking function that also takes the `allowance_margins` into account (10% by default). E.g. currently, in the situation where both penultimate and last cycles are exceeded by less than 10% (e.g. 10,150/10,000 and 10,999/10,000), the `exceeds_last_two_usage_cycles?` would return false, while usage.last_cycle > 10,000 would return true. That said though, I'm not fully sure how to handle these margins.

lib/plausible_web/components/billing/notice.ex
```diff
@@ -265,6 +265,240 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+
+      site_usage >= site_limit and site_limit != :unlimited and
+        team_member_usage >= team_member_limit and team_member_limit != :unlimited ->
+        :limits_reached_combined
```
RobertJoonas: WDYT about renaming this to `:site_and_team_member_limit_reached`? Currently, "combined" makes it sound a bit like pageview limit is reached too.

sanne-san: Yes, good idea!

lib/plausible/billing/qouta/quota.ex
```diff
@@ -225,13 +247,16 @@ defmodule Plausible.Billing.Quota do
   end
 
   defp pageview_cycle_usage_notification_type(usage, limit) do
-    exceeded = exceeded_cycles(usage, limit)
+    last_exceeded? = is_map_key(usage, :last_cycle) and usage.last_cycle.total > limit
```
RobertJoonas: I think it would be best to keep using the exceeded cycles function here, just in order to link it to that context so that whenever there's a change related to limit checking, it brings the attention to the subscription settings page as well (because checking limits is relevant here).

My proposal would be to extend the `exceeded_cycles` function with a third argument `opts \\ []` (empty list by default), and then evaulate the actual limit for checking as:

```elixir
limit =
  if Keyword.get(opts, :with_margin, true) do
    Limits.pageview_limit_with_margin(allowed_volume)
  else
    limit
  end
```

And then in the call sites (such as our `pageview_cycle_usage_notification_type`) we could do:

```elixir
exceeded = exceeded_cycles(usage, limit, with_margin: false)
```

lib/plausible/billing/qouta/quota.ex
```diff
@@ -163,4 +161,109 @@ defmodule Plausible.Billing.Quota do
...
+    site_usage = usage.sites
+    team_member_usage = usage.team_members
+
+    pageview_notification =
+      if not Plausible.Teams.on_trial?(team) and has_billing_cycles?(pageview_usage) do
+        pageview_cycle_usage_notification_type(pageview_usage, pageview_limit)
+      end
+
+    cond do
+      Plausible.Teams.GracePeriod.expired?(team) ->
```
RobertJoonas: I think this would be the most suitable condition for the :dashboard_locked notification:

```suggestion
      Plausible.Teams.locked?(team) or Plausible.Teams.GracePeriod.expired?(team) ->
```

`locked?` also takes care of enterprise plans that have been locked out manually. It's also mostly true for expired grace periods, but it's safer to keep both of these conditions here since 1) both mean that the user shouldn't have access, and 2) CheckUsage (the worker setting grace periods) vs LockSites (the worker actually locking sites) are different workers.

lib/plausible/billing/qouta/quota.ex
```diff
@@ -163,4 +161,109 @@ defmodule Plausible.Billing.Quota do
...
+        :dashboard_locked
+
+      not Plausible.Teams.on_trial?(team) and is_nil(subscription) ->
+        :trial_ended
+
+      Plausible.Teams.GracePeriod.active?(team) ->
```
RobertJoonas, line 237: As discussed on Basecamp, this case needs to be split in two:

```elixir
GracePeriod.manual_lock_active?(team) ->
  :manual_lock_grace_period_active

GracePeriod.active?(team) ->
  :grace_period_active
```

In the :manual_lock_grace_period_active case, we should end up displaying this message:

> **Title**: Upgrade required due to sustained higher traffic
> **Body**: To ensure uninterrupted access to your stats, please upgrade to a plan that fits your current usage.


NOTE: The `GracePeriod.manual_lock_active?` function doesn't exist yet, but we should add it.

lib/plausible/billing/qouta/quota.ex
sanne-san: Yeah, good idea 👍

lib/plausible_web/components/billing/notice.ex
```diff
@@ -265,6 +265,213 @@ defmodule PlausibleWeb.Components.Billing.Notice do
...
+  end
+
+  def usage_notification(%{type: :grace_period_active, team: team} = assigns) do
+    days_left = Plausible.Teams.GracePeriod.days_left(team)
+
+    deadline_text =
+      case days_left do
+        1 -> "within the next day"
+        n -> "within the next #{n} days"
+      end
```
RobertJoonas: Hmm, I think instead of this (and the new function) we can actually use the already existing `PlausibleWeb.LayoutView.grace_period_end(team)`

sanne-san: Yeah I was wondering that but the formatting is a bit different (`today` and `tomorrow`). Am I OK to change that? It seems like it isn't actually being used anywhere?

RobertJoonas: Oh, you're right. The function I suggested is not used anywhere anymore since https://github.com/plausible/analytics/pull/4652. But we can adjust it and start to use it, sure.

Actually, `grace_period_end` is in UTC and we compare it with `Date.utc_today()`, so "today" could actually mean "tomorrow" and vice versa, because the person viewing the message might be in a different timezone. I think in order to be as accurate as possible, we should display the difference in hours instead of days when `days_left < 2`. WDYT?

Wrote some code here: https://github.com/plausible/analytics/pull/6067. Feel free to merge :)

sanne-san: Good idea to show hours closer to the deadline 👌 thank you!

lib/plausible/teams/grace_period.ex
```diff
@@ -99,4 +111,40 @@ defmodule Plausible.Teams.GracePeriod do
   def expired?(team) do
     if team && team.grace_period, do: !active?(team), else: false
   end
+
+  @spec expires_in(Teams.Team.t() | nil) :: {non_neg_integer(), :days | :hours} | nil
+  @doc """
+  Returns a tuple representing either the days (if hours_left < 48) or days left
```
RobertJoonas: My own typo I think 😅
```suggestion
  Returns a tuple representing either the hours (if hours_left < 48) or days left
```

lib/plausible/billing/qouta/quota.ex
```diff
@@ -163,4 +166,110 @@ defmodule Plausible.Billing.Quota do
...
+      end
+
+    cond do
+      Plausible.Teams.locked?(team) or Plausible.Teams.GracePeriod.expired?(team) ->
+        :dashboard_locked
+
+      not Plausible.Teams.on_trial?(team) and is_nil(subscription) ->
+        :trial_ended
+
+      Plausible.Teams.GracePeriod.manual_lock_active?(team) ->
```
RobertJoonas, line 234: Sorry, I totally missed this before, but in this situation, the `manual_lock` grace period can also be set due to exceeding the sites limit.

One option would be to branch out even further here, (in a separate function perhaps):

```elixir
Plausible.Teams.GracePeriod.manual_lock_active?(team) ->
  enterprise_over_limits_notification_type(usage, pageview_limit)
```

And then the function:

```elixir
defp enterprise_over_limits_notification_type(usage, pageview_limit) do
  if Quota.exceeds_last_two_usage_cycles?(usage.monthly_pageviews, pageview_limit) do
    :enterprise_traffic_exceeded
  else
    :enterprise_sites_exceeded
  end
end
```

Another option - stick to the `:manual_lock_grace_period_active` notification type, but render a different message to the user:

> Title: You've outgrown your custom plan
> Body: We will contact you via email to discuss the terms of a new plan that would accommodate your usage.

I personally prefer option 2 as it's simpler - the notice doesn't need to say anything about which limit it was (sites or pageviews) that they exceeded.

Not exactly sure about the copy though. Perhaps @metmarkosaric can help here again. The situation we're discussing is when an enterprise plan exceeds either pageview or site limit (i.e. from the moment that the internal email notification gets sent to hello@plausible.io). And this message would be displayed until they've actually went ahead and subscribed to the new plan that accommodates the usage.

sanne-san: Yeah I like option 2 as well – it keeps it simpler and also I think this copy is actually better for this use case than what we have now 👌

Does it then make sense that there's no CTA for this message? Will they always receive an email with a button that goes straight to the upgrade page?

RobertJoonas: I guess the CTA could be something like: "Please check your inbox if you've already received an email from us. If not, feel free to reach out yourself to discuss the terms of your new plan".

sanne-san: I've updated the copy to be: 

> Title: You've outgrown your custom plan
> Body: We'll contact you by email to discuss an updated custom plan based on your current usage.

metmarkosaric: It’s a bit of a complicated situation. The proposed copy doesn’t seem to fit the situation when we’ve already created a new custom plan for them and they can just click to upgrade

And that’s what happens in most cases. they don’t actually need to contact us or speak with us. We get an alert, set up a new plan automatically and tell them about it. There may be a couple of hours delay between we get the alert and set up a new plan and that’s where they could contact us but that’s about it 

Plus in some situations such as when a customer adds site 50 out of a 50 limit we don’t take any action other than to remove the grace period in the CRM for them without requiring them to upgrade. That also removes the alert. As in we just tell the to upgrade if they want to add site 51 

So an alert like this sounds more appropriate for most situations : 

Title: You’ve exceeded your plan
Body: Upgrade to a new custom plan that matches your current usage.

PR #6059, opened 2026-02-10
A major renaming: DashboardQuery -> DashboardState

RobertJoonas:
### Changes

This PR is a pure refactor dealing only with renaming. The diff is quite enormous so I decided to open a separate PR just for that.

Renaming query -> dashboardState globally in the frontend makes sense because:

* the "dashboard state" that is derived from url search params, doesn't yet represent a query. In order to actually define a "query" dimensions and metrics need to be provided
* dashboardState also includes fields that are irrelevant when "querying". E.g. `resolvedFilters` (used for flattening segment filters so the FE could be aware of what filters are applied), and `labels` (for rendering country, city, region, segment names in filter labels)
* An actual `StatsQuery` object is about to be introduced with https://github.com/plausible/analytics/pull/6035, so this is to avoid a naming clash.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/dashboard-time-periods.ts
+++ b/assets/js/dashboard/dashboard-time-periods.ts
@@ -334,7 +334,8 @@ export const getDatePeriodGroups = ({
             period: DashboardPeriod['24h'],
             keybindHint: 'H'
           }),
-          isActive: ({ dashboardState }) => dashboardState.period === DashboardPeriod['24h'],
+          isActive: ({ dashboardState }) =>
+            dashboardState.period === DashboardPeriod['24h'],
           onEvent
         }
       ],


assets/js/dashboard/stats/graph/date-formatter.js
```diff
@@ -92,10 +92,10 @@ const factory = {
  * while other intervals require dates to be displayed.
  * @param {Object} config - Configuration object for determining formatter.
  *
- * @param {string} config.interval - The interval of the query, e.g. `minute`, `hour`
+ * @param {string} config.interval - The interval of the dashboardState, e.g. `minute`, `hour`
  * @param {boolean} config.longForm - Whether the formatted result should be in long or
  * short form.
- * @param {string} config.period - The period of the query, e.g. `12mo`, `day`
+ * @param {string} config.period - The period of the dashboardState, e.g. `12mo`, `day`
```
apata: meganitpick, non-blocking: in this context, the original is more appropriate I think

assets/js/dashboard/stats/graph/line-graph.js
```diff
@@ -111,42 +114,45 @@ class LineGraph extends React.Component {
...
                     longForm: false,
-                    period: query.period,
+                    period: dashboardState.period,
                     shouldShowYear
                   })(this.getLabelForValue(val))
 
                   // Returns a combination of date and hour. This is because
                   // small intervals like hour may return multiple days
-                  // depending on the query period.
+                  // depending on the dashboardState period.
```
apata: meganitpick, non-blocking: the original one is more appropriate inside the comment

apata (approved):
Wow, we do use the word query in a lot of places. Great job getting them all replaced successfully! 

Regarding the new name, I can't think of a better one, but I think internally we should be careful with it for future features. Everything the user sees on the dashboard is part of its state, but not everything should be part of DashboardState and DashboardStateContext. Otherwise, we may trigger unnecessary re-renders that in turn make the page slow for the users.

assets/js/dashboard/stats/graph/date-formatter.js
RobertJoonas: https://github.com/plausible/analytics/pull/6059/commits/1231843658384f53e9defdf7e20c7c4e7db54f65

assets/js/dashboard/stats/graph/line-graph.js
RobertJoonas: https://github.com/plausible/analytics/pull/6059/commits/1231843658384f53e9defdf7e20c7c4e7db54f65

RobertJoonas:
> Regarding the new name, I can't think of a better one, but I think internally we should be careful with it for future features. Everything the user sees on the dashboard is part of its state, but not everything should be part of DashboardState and DashboardStateContext. Otherwise, we may trigger unnecessary re-renders that in turn make the page slow for the users.

@apata, Fair point about being careful!

PR #6062, opened 2026-02-11
Add 24h time range to dashboard and Stats API v2

aerosol:
### Changes

This PR adds "Last 24h" time range to dashboard picker, so we can reflect the view that's been historically presented on /sites.
Additionally, Stats V2 API support for `24h` period is added.

image

### Tests

### Changelog

### Documentation
- [x] [Docs](https://github.com/plausible/docs) have been updated: https://github.com/plausible/docs/pull/659

### Dark mode

### DIFF
--- a/extra/lib/plausible/stats/consolidated_view.ex
+++ b/extra/lib/plausible/stats/consolidated_view.ex
@@ -1,11 +1,11 @@
 defmodule Plausible.Stats.ConsolidatedView do
   alias Plausible.{Site, Stats}
-  alias Plausible.Stats.{ParsedQueryParams, QueryBuilder, QueryInclude}
+  alias Plausible.Stats.QueryBuilder
   require Logger
 
   @spec overview_24h(Site.t(), NaiveDateTime.t()) :: map()
   def overview_24h(%Site{consolidated: true} = view, now \\ NaiveDateTime.utc_now()) do
-    stats = query_24h_stats(view, now)
+    stats = query_24h_stats(view)
     intervals = query_24h_intervals(view, now)
 
     Map.merge(stats, intervals)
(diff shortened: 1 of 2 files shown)


lib/plausible/stats/query_builder.ex
```diff
@@ -100,6 +100,13 @@ defmodule Plausible.Stats.QueryBuilder do
     DateTimeRange.new!(relative_date, relative_date, site.timezone)
   end
 
+  defp build_datetime_range(:"24h", site, _relative_date, now) do
+    to = now |> DateTime.shift_zone!(site.timezone) |> DateTime.shift_zone!("Etc/UTC")
```
ukutaht: Are these shifts necessary? Seems like we start with utc now and end up with utc now

aerosol: absolutely, https://github.com/plausible/analytics/pull/6062/commits/4d8b94097444ac19f10e8aad6e951d8708973507

lib/plausible/stats/comparisons.ex
```diff
@@ -53,13 +53,18 @@ defmodule Plausible.Stats.Comparisons do
...
 
         _ ->
-          comparison_date_range = get_comparison_date_range(source_query)
-
-          DateTimeRange.new!(
-            comparison_date_range.first,
-            comparison_date_range.last,
-            source_query.timezone
-          )
+          # For 24h period, work directly with datetime ranges to preserve time precision
```
ukutaht, line 56: I remember you had to hand-roll 24h comparisons for consolidated views on /sites page as well. As a follow-up can check if it's possible to run those through QueryBuilder now as well.

aerosol: Good memory! :older_man: 
Fixing `now` is a bit janky, but hey: 58fed9a398

PR #6073, opened 2026-02-16
Include bounce rate in top stats with page filter and imported data + fix 500

RobertJoonas:
### Changes

This PR prepares for moving Top Stats to a new `/stats/:domain/query` endpoint. Currently, `bounce_rate` and `time_on_page` are excluded from Top Stats when imported data is included. This PR brings them in (while still staying on the old API endpoint).

I had hoped to cut it out of the scope of this project and not delve deeper into "why" the metrics were excluded in the first place. However, this would have meant extending the Query API with the ability to drop those metrics if `query.include_imported && page_filter? && dimensions == []`. It felt like a step in the wrong direction since imported data shouldn't affect native metrics being queried. On top of that, this behaviour would've also leaked into the public API.

So that's why I decided to dig a bit deeper and go in the right direction instead -- if a metric is only available from native data, we should just render a `metric_warning` (like we do with scroll depth):

image

Also, as it turned out, there's no technical limitation for including imported `time_on_page` in page-filtered top stats either.

#### Pagination optimisation fix

In addition to the above, this PR also fixes a bug with the pagination optimisation when joining imported tables. To reproduce the bug:

1) Open a dashboard with imported data in view
2) Open the Entry Pages > Details view
3) Click on bounce rate to order by it -> 500 internal server error

To fix the bug, for divisible metrics (e.g. `bounce_rate`, `views_per_visit`, `visit_duration`), we need to select the actual metric that is ordered by (i.e. the division product), as well as the separate fields to calculate the merged value later on.

For `visit_duration` the optimisation worked accidentally -- the value that we were ordering by was not the actual metric, but rather the "total visit duration". I fixed that as well, ordering by the division product rather than the total value, which should reduce the inaccuracy of the optimisation. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/.github/workflows/build-private-images-ghcr.yml
+++ b/.github/workflows/build-private-images-ghcr.yml
@@ -72,9 +72,11 @@ jobs:
       - name: Get first line and Co-Authored-By lines of the commit message
         if: ${{ success() && github.ref == 'refs/heads/master' }}
         id: commitmsg
+        env:
+          COMMIT_MSG: ${{ github.event.head_commit.message }}
         run: |
-          first_line=$(echo "${{ github.event.head_commit.message }}" | head -n1)
-          co_authors=$(echo "${{ github.event.head_commit.message }}" | grep -h 'Co-authored-by:' | sort -u | cut -d: -f2- | paste -sd, -)
+          first_line=$(printf '%s\n' "$COMMIT_MSG" | head -n1 | xargs)
+          co_authors=$(printf '%s\n' "$COMMIT_MSG" | grep -h 'Co-authored-by:' | sort -u | cut -d: -f2- | paste -sd, - | xargs)
           {
             echo "first_line=$first_line"
             echo "co_authors=$co_authors"


lib/plausible/stats/imported/sql/expression.ex
```diff
@@ -72,51 +72,95 @@ defmodule Plausible.Stats.Imported.SQL.Expression do
...
+      bounce_rate: fragment("any(0)"),
+      bounces: fragment("any(0)"),
+      __internal_visits: fragment("any(0)")
+    })
   end
 
   defp select_metric(:bounce_rate, "imported_exit_pages", _query) do
-    wrap_alias([i], %{bounces: sum(i.bounces), __internal_visits: sum(i.exits)})
+    wrap_alias([i], %{
+      bounce_rate: fragment("ifNotFinite(round(? / ?), 0)", sum(i.bounces), sum(i.exits)),
```
zoldar: Given this expression is used for ordering, I _think_ this might lead to invalid order as `bounces / exits` is supposed to be a ratio between 0-1 and rounding will force it into 0 or 1. Given it's meant to be a percentage, shouldn't we multiply it by 100 before rounding?

lib/plausible/stats/imported/sql/expression.ex
```diff
@@ -72,51 +72,95 @@ defmodule Plausible.Stats.Imported.SQL.Expression do
...
+      bounces: sum(i.bounces),
+      __internal_visits: sum(i.exits)
+    })
   end
 
   defp select_metric(:bounce_rate, "imported_entry_pages", _query) do
-    wrap_alias([i], %{bounces: sum(i.bounces), __internal_visits: sum(i.entrances)})
+    wrap_alias([i], %{
+      bounce_rate: fragment("ifNotFinite(round(? / ?), 0)", sum(i.bounces), sum(i.entrances)),
```
zoldar: Ditto about rounding issue.

lib/plausible/stats/imported/sql/expression.ex
```diff
@@ -72,51 +72,95 @@ defmodule Plausible.Stats.Imported.SQL.Expression do
...
+      bounces: sum(i.bounces),
+      __internal_visits: sum(i.entrances)
+    })
   end
 
   defp select_metric(:bounce_rate, _table, _query) do
-    wrap_alias([i], %{bounces: sum(i.bounces), __internal_visits: sum(i.visits)})
+    wrap_alias([i], %{
+      bounce_rate: fragment("ifNotFinite(round(? / ?), 0)", sum(i.bounces), sum(i.visits)),
```
zoldar: Ditto about rounding issue

lib/plausible/stats/imported/sql/expression.ex
RobertJoonas: Nice catch! Forgot to divide by 100 indeed.

That aside, I just realised something that makes this optimisation incorrect. It is not applicable when `event:page` dimension or filter is used (in other words, when the imported table is `imported_pages`). Why? Because we are selecting bounce rate as a constant 0 from `imported_pages` which means there's no order to rely on. With millions of different page paths it has a high risk of excluding rows with the JOIN that are relevant for computing other metrics.

I feel like this requires some deeper thinking (and tests!) which goes too far from my current assignment, so what I did was to stop applying this optimisation for bounce rate. https://github.com/plausible/analytics/pull/6073/commits/88d580877c20a18703404a5a6b56643435e52f69

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -402,9 +402,6 @@ defmodule PlausibleWeb.Api.StatsController do
 
     metrics =
       cond do
```
zoldar, line 404: Credo complains about this `cond` (it enforces `if` when there are 2 or less conditional branches).

PR #6075, opened 2026-02-17
Load site sparklines in parallel

aerosol:
### Changes

In order to provide foundation for sortable site index, this PR ensures sparkline plots (on site index + consolidated view, and CRM) use common querying interface now.
By nature, it does not allow querying sites in bulk, so we resort to executing parallel queries, as safely as possible. Because separate processes are spawned to checkout pool connections, we can no longer rely on fixing `now` in tests using process dictionary, therefore explicit options to query parsing and building functions have been added. For external controller tests, we will fix `now` via `%Conn{}`'s private assigns instead.

Tip of the iceberg still -- we'll have to make sparklines ready for varying ranges and intervals, essentially mirroring basic dashboard plotting capabilities. Leaving "24h" static for now, just so the PR doesn't grow too much. I did take the opportunity to rename "hourly_stats" though.

To be seen if using default parallelism isn't to aggressive from UX perspective, but 24h range alone may not reveal any problems just yet.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/templates/auth/activate.html.heex
+++ b/lib/plausible_web/templates/auth/activate.html.heex
@@ -17,13 +17,13 @@
   </:title>
 
   <:subtitle :if={@has_email_code?}>
-    <p class="truncate">
+    <p class="break-all">
       Please enter the 4-digit code we sent to <b>{@conn.assigns[:current_user].email}</b>
     </p>
   </:subtitle>
 
   <:subtitle :if={!@has_email_code?}>
-    <p class="truncate">
+    <p class="break-all">
       A 4-digit activation code will be sent to <b>{@conn.assigns[:current_user].email}</b>
     </p>
   </:subtitle>


extra/lib/plausible_web/live/customer_support/team/components/consolidated_views.ex
```diff
@@ -12,21 +12,21 @@ defmodule PlausibleWeb.CustomerSupport.Team.Components.ConsolidatedViews do
...
+    sparkline_intervals =
       with true <- connected?(socket),
-           {:ok, hourly_stats} <- Stats.ConsolidatedView.safe_overview_24h(consolidated_view) do
-        hourly_stats.intervals
+           {:ok, sparkline} <- Stats.Sparkline.safe_overview_24h(consolidated_view) do
+        sparkline.intervals
       else
         _ ->
-          Stats.ConsolidatedView.empty_24h_intervals()
+          Stats.Sparkline.empty_24h_intervals()
```
zoldar, line 21: We could change this function to always return `[..., %{interval: _, visitors: _}]` instead of tuples. Basically every spot where it's used makes that transformation in the end.

lib/plausible/stats/api_query_parser.ex
```diff
@@ -40,7 +40,8 @@ defmodule Plausible.Stats.ApiQueryParser do
          dimensions: dimensions,
          order_by: order_by,
          pagination: pagination,
-         include: include
+         include: include,
+         now: Keyword.get(opts, :now)
```
zoldar, line 44: @RobertJoonas we talked about this yesterday when I was showing you E2E tests 😄 - though exposing this all the way to the frontend would add a lot of extra mess I'm afraid, so working with real time is probably still better...

lib/plausible/stats/query.ex
```diff
@@ -48,19 +48,38 @@ defmodule Plausible.Stats.Query do
 
   @type t :: %__MODULE__{}
 
+  def parse_and_build(
```
zoldar: Maybe consider moving `debug_metadata` inside `opts` ?

lib/plausible/stats/sparkline.ex
```diff
@@ -1,29 +1,52 @@
...
+  The sparkline graphs are used in site index and in CRM.
+  """
+
   alias Plausible.{Site, Stats}
   alias Plausible.Stats.QueryBuilder
   require Logger
 
+  @spec parallel_overview([Site.t()], NaiveDateTime.t()) :: map()
+  def parallel_overview(sites, now \\ NaiveDateTime.utc_now()) when is_list(sites) do
+    Task.async_stream(
```
zoldar, line 13: Perhaps consider setting `max_concurrency` to something like 4 instead of the default to reduce pool usage? So that a single dashboard with a whole bunch of expensive sites won't hog it.

extra/lib/plausible_web/live/customer_support/team/components/consolidated_views.ex
aerosol: good call

aerosol: https://github.com/plausible/analytics/pull/6075/commits/c95fb654a12c9f48a8de58525f07f17f555c9244

aerosol: sorry, b181a23f1b

lib/plausible/stats/query.ex
aerosol: done via 24277ab17d

lib/plausible/stats/sparkline.ex
aerosol: done

aerosol:
> Looks good to me. A couple comments:
> 1. I think we should use the [built-in](https://plausible.io/docs/stats-api#include.time_labels) `include.time_labels` over manually building the 24h interval labels.

Ooh nice, totally missed that

> 2. Would really like to see some error handling added to the UI for Clickhouse queries.

Do you think we can do something sensible here apart from periodically retrying?

ukutaht:
> Do you think we can do something sensible here apart from periodically retrying?

I guess I just mean that keeping the UI in in loading state forever is
a) incorrect
b) harder to debug (is it just taking a long time or did the request fail?)

IMO it would be better to have a UI state for 'loading failed'.

zoldar:
@ukutaht @aerosol  Just a note that we had a similar situation in the past where we started reporting errors on the dashboard (instead of leaving "infinite" spinners) and the user experience suffered greatly and we reverted it. Perhaps some middle ground like leaving spinners on with some small warning icon in the corner and a tooltip on hover like "this operation is taking longer than usual, please try [reloading](reloadlink)" would work better.

aerosol:
> @ukutaht @aerosol Just a note that we had a similar situation in the past where we started reporting errors on the dashboard (instead of leaving "infinite" spinners) and the user experience suffered greatly and we reverted it. Perhaps some middle ground like leaving spinners on with some small warning icon in the corner and a tooltip on hover like "this operation is taking longer than usual, please try [reloading](reloadlink)" would work better.

Thanks for bringing that up @zoldar. Good points, let's decide what to do in a follow-up. In the meantime I pushed 97e809e049 and b802847622 to gain some clarity.

PR #6076, opened 2026-02-17
Improve subscription settings layout

sanne-san:
### Changes

- Break up settings into three tiles; "Current plan", "Monthly usage", "Site and team usage".
- Improve overview of current plan and clearer buttons to update billing details or change plan/upgrade.
- Show trial ended notification for free trials that have ended.
- Move "Cancel plan" button outside of the tiles, to the bottom of the page.
- Remove "Upgrade" button at the bottom, and instead conditionally change the main button text from "Change plan" to "Upgrade".
- Add red variant to pill component, to be used for "Paused" subscription status.

### Tests

### Changelog

### Documentation
- [x] This change does not need a documentation update (will be updated once all changes to the subscription settings page have been made)

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Keybind hints are hidden on smaller screens
+- Subscription settings page design is updated
 
 ### Fixed
 


lib/plausible_web/templates/settings/subscription.html.heex
```diff
@@ -1,109 +1,141 @@
...
-          <div class="py-2 text-xl font-medium dark:text-gray-100">
-            {PlausibleWeb.BillingView.present_currency(@subscription.currency_code)}{@subscription.next_bill_amount}
-          </div>
-          <.styled_link :if={@subscription.update_url} href={@subscription.update_url}>
-            Update billing info
-          </.styled_link>
-        <% else %>
-          <div class="py-2 text-xl font-medium dark:text-gray-100">---</div>
-        <% end %>
+  <%= if Plausible.Teams.on_trial?(@current_team) do %>
```
ukutaht: So the new logic relies on the `@current_team` assign but there is one edge case.

When a user signs up, we only create a team for them when they create their first site not when they sign up. This means there's a moment in time where a user does not have a team and `@current_team` is unresolved. The user can skip the onboarding flow and access the subscription page directly before creating a site. In that case since `@current_team` is `nil`, they will see this:

Image

When in fact their trial has not started yet.

lib/plausible_web/views/settings_view.ex
```diff
@@ -4,12 +4,51 @@ defmodule PlausibleWeb.SettingsView do
...
+  def subscription_pill_color(Subscription.Status.past_due()), do: :yellow
+  def subscription_pill_color(Subscription.Status.paused()), do: :red
+  def subscription_pill_color(Subscription.Status.deleted()), do: :red
+  def subscription_pill_color(_), do: :gray
+
+  def trial_button_label(team) do
+    if Plausible.Teams.Billing.enterprise_configured?(team) do
+      "Upgrade"
+    else
+      "Choose a plan &rarr;"
```
ukutaht: This does not render as intended:

Image

lib/plausible_web/templates/settings/subscription.html.heex
```diff
@@ -1,109 +1,141 @@
...
+          />
+          <div class="flex flex-col sm:flex-row justify-between items-start gap-4">
+            <div class="flex flex-col gap-1">
+              <div class="flex items-center gap-2 flex-wrap">
+                <% plan = Plausible.Billing.Plans.get_subscription_plan(@subscription)
 
-        <%= if @subscription && @subscription.next_bill_date && Plausible.Billing.Subscription.Status.in?(@subscription, [Plausible.Billing.Subscription.Status.active(), Plausible.Billing.Subscription.Status.past_due()]) do %>
-          <div class="py-2 text-xl font-medium dark:text-gray-100">
-            {Calendar.strftime(@subscription.next_bill_date, "%b %-d, %Y")}
+                plan_name =
```
ukutaht: This is a lot of logic for template code. Extracting a helper should help clean it up.

lib/plausible_web/templates/settings/subscription.html.heex
RobertJoonas: Good catch @ukutaht !

@sanne-san I guess we could just extend the conditional here to render the same content as for trials if the user hasn't started the trial yet.

```suggestion
  <%= is_nil(@current_team) or if Plausible.Teams.on_trial?(@current_team) do %>
```

And perhaps instead of the "days left" pill we could render a message saying that "Your 30-day trial will start when you add your first site".

JIC, I would like @zoldar to chime in here - is there another "team is nil" case that we should be careful about? 🤔

sanne-san: @ukutaht good catch! I'm just wondering what they _should_ be seeing in this case? The on-trial state without a `X days left` badge, maybe? Or a separate state that says something like "Create your first site to start your free trial"?

Given this is quite an edge case I'm inclined to go with the former. Wdyt?

zoldar: In practice, `current_team = nil` should be expected in following situations:

- user registers but exits from wizard flow before creating their first site (what @ukutaht mentioned)
- user with one or more non-guest memberships switches to "My personal sites" while not owning any sites and not having personal team yet
- user registers as a guest from an invitation
- user leaves all teams they are a non-guest member in while not having their personal team
- user deletes their only team (after setting it up) while not being a member on any other team - which is possible only if they don't have any active subscription on that team

Other than that, I _think_ `current_team` should be present 😅

sanne-san: Oh – I only just saw your response after refreshing the page @RobertJoonas! 😄 I was literally just trying to answer the question you asked @zoldar lol. Solution sounds good to me!

RobertJoonas (approved):
Nice work! 👏

PR #6077, opened 2026-02-17
Extend E2E dashboard tests

zoldar:
### Changes

This PR extends previously setup E2E test suite to cover major extent of dashboard interactions.

We don't cover all the possible states, mostly focusing on "happy" paths.

The changes in the frontend code are limited to:

- adding "data-testid" attributes (and one instance of extra "data-" attribute for determining state of tab more easily)
- adding aria labels/labelledby where it makes sense

The missing coverage any cases of greater significance can be added in follow-ups as this PR is already huge. We might add more tests for less usual interaction paths but we will have to strike a balance to avoid inflating the test suite and slowing down the CI pipeline too much.

The top level tests number is kept relatively low (currently between 40-50) as state setup is usually the expensive part of the process. We might consider replacing frontend-centric fixtures setting up user and site via forms on the page with dedicated e2e endpoints in the future to reduce the time spent on it if that becomes too much of a burden (we might still keep the existing fixtures and use them only for a couple general tests to keep covering areas outside the dashboard by extension). We make extensive use of `test.step` to cram more test cases in each test. This means there are effectively more test cases than the report implies.

~~__NOTE__: CI is currently failing due to unrelated tests failing (they fail across other branches too). I'll have a look at them independently.~~ Fixed!

### Tests
- [x] Automated tests _definitely_ have been added

### DIFF
--- a/e2e/tests/fixtures.ts
+++ b/e2e/tests/fixtures.ts
@@ -279,7 +279,9 @@ export async function addCustomGoal({
   }
 
   if (currency) {
-    await page.locator('button[aria-labelledby="enable-revenue-tracking"]').click()
+    await page
+      .locator('button[aria-labelledby="enable-revenue-tracking"]')
+      .click()
     const currencyInput = page.locator('input[id^=currency_input_]')
     await currencyInput.fill(currency)
     await page.locator(`a[phx-value-submit-value="${currency}"]`).click()


assets/js/dashboard/nav-menu/segments/segment-menu.tsx
```diff
@@ -86,6 +86,7 @@ export const SegmentMenu = () => {
               )}
             >
               <ChevronDownIcon
+                data-testid="segmentmenu"
```
apata: meganitpick, non-blocking: this stood out compared to the rest, can we do "segment-menu"?

e2e/tests/test-utils.ts
```diff
@@ -8,3 +8,63 @@ export async function expectLiveViewConnected(page: Page) {
 export function randomID() {
   return Math.random().toString(16).slice(2)
 }
+
+export const tabButton = (page, label) =>
```
apata: suggestion, non-blocking: It's quite handy to have types for utilities. That way if the utility is refactored, TS errors will point to code that needs to be refactored as well. It's also good when updating deps: when an accessor like `getByTestId` is deprecated etc, we'd get errors helping us update code.

e2e/tests/fixtures.ts
```diff
@@ -12,13 +12,48 @@ type EventTimestamp =
   | { minutesAgo: number }
   | { hoursAgo: number }
   | { daysAgo: number }
+  | string
+
+type EventDate = { daysAgo: number } | string
```
apata, line 17: nitpick, non-blocking: this is possibly missing some options { hoursAgo: number } and the others that are defined in e2e_controller.ex

zoldar: Date supports only hoursAgo and ISO string format. Timestamp supports more and it's accounted for in type definitions above.

e2e/tests/test-utils.ts
apata: I gave it a go with https://github.com/plausible/analytics/pull/6138, WDYT?

e2e/tests/fixtures.ts
```diff
@@ -152,24 +226,268 @@ export async function populateStats({
...
+  await expectLiveViewConnected(page)
+
+  await page.getByRole('button', { name: 'Add property' }).click()
+
+  await expect(
+    page.getByRole('heading', { name: `Add property for ${domain}` })
+  ).toBeVisible()
+
+  const propInput = page.locator('input#prop_input')
+  propInput.fill(name)
```
apata: nitpick, non-blocking: missing await

e2e/tests/fixtures.ts
```diff
@@ -152,24 +226,268 @@ export async function populateStats({
...
+  await expect(nameInput).toHaveAttribute('value', name)
+
+  if (displayName) {
+    await page
+      .locator('input#custom_event_display_name_input')
+      .fill(displayName)
+  }
+
+  if (currency) {
+    page.locator('button[aria-labelledby="enable-revenue-tracking"]').click()
```
apata: nitpick, non-blocking: missing await

e2e/tests/fixtures.ts
zoldar: >  non-blocking:

I see what you did there :joy:

e2e/tests/fixtures.ts
zoldar: Yeah, definitely something to fix, thanks!

e2e/tests/test-utils.ts
zoldar: This looks awesome! Thanks!

e2e/tests/fixtures.ts
zoldar: https://github.com/plausible/analytics/pull/6077/commits/24c82583db703cff36d1556ecce8246ddba2d520

e2e/tests/fixtures.ts
zoldar: https://github.com/plausible/analytics/pull/6077/commits/24c82583db703cff36d1556ecce8246ddba2d520

assets/js/dashboard/nav-menu/segments/segment-menu.tsx
zoldar: https://github.com/plausible/analytics/pull/6077/commits/aafe042414e503cd628a4576cff4cf7535308f28

PR #6084, opened 2026-02-19
Removing a typo

metmarkosaric:
I think this should be able to solve this https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/9582538418


aerosol (approved):

zoldar:
We could move the stop/dot inside upgrade call outside the link underline too

PR #6085, opened 2026-02-19
Wire up progress bars for pageview and plan limits

sanne-san:
### Changes

- Replace the tab-based billing cycle selector with a stacked collapsible layout.
- The current cycle is always shown; last and penultimate cycles are only shown when the current or last cycle has exceeded the plan limit.
- Also replaces `usage_progress_row` (which combined a progress bar with a title/count row) with a standalone `usage_progress_bar` component, and removes the now-unused `usage_and_limits_table` and `usage_and_limits_row` components.
- The site and team member usage section in the subscription template is updated to use the new progress bar with inline labels.

### Tests

### Changelog

### Documentation
- [x] This change does not need a documentation update (will be done once all subscription settings page changes have been completed)

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/settings_controller_test.exs
+++ b/test/plausible_web/controllers/settings_controller_test.exs
@@ -447,8 +447,8 @@ defmodule PlausibleWeb.SettingsControllerTest do
         |> get(Routes.settings_path(conn, :subscription))
         |> html_response(200)
 
-      assert html =~ "Owned sites"
-      assert html =~ "/ 50"
+      sites_usage_text = text_of_element(html, "[data-test-id='sites-usage']")
+      assert sites_usage_text =~ "1 / 50"
     end
 
     @tag :ee_only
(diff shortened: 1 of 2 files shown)


test/plausible_web/controllers/settings_controller_test.exs
```diff
@@ -519,39 +442,39 @@ defmodule PlausibleWeb.SettingsControllerTest do
...
-      site_usage_row_text =
+      html =
         conn
         |> get(Routes.settings_path(conn, :subscription))
         |> html_response(200)
-        |> text_of_element("#site-usage-row")
 
-      assert site_usage_row_text =~ "Owned sites 1 / 50"
+      assert html =~ "Owned sites"
+      assert html =~ "/ 50"
```
RobertJoonas: I tried breaking the sites usage rendering and there were no other tests covering it, so we should put the `1` back here (same for team member limit I think).

I also tried putting it back, but it looks like some changes created a newline character between `1` and `/ 50`. That's okay. We can still test the existence of that usage number by giving the span a `data-test-id` and then using something like:

```elixir
sites_usage_text = Plausible.Test.Support.HTML.text_of_element("[data-test-id='sites-used']")
assert site_usage_text =~ "1 / 50"
```

PR #6086, opened 2026-02-19
fix: singular views per visit

ukutaht:
### Changes

When a top stat has value of 1, we adjust the stat label to remove pluralization.

Currently it unconditionally removes the last character which means `views per visit` is rendered as `views per visi`:
Screenshot 2026-02-19 at 16 48 54

This looks broken immediately after onboarding. This PR fixes it so we remove the `s` character from the end only if present:

Screenshot 2026-02-19 at 16 48 46


zoldar (approved):
LGTM, formatting linter is complaining though.

PR #6092, opened 2026-02-23
Move invoices into subscription settings

sanne-san:
### Changes

The invoices tab is moved into the subscription settings as part of a larger overhaul of the subscription settings page. This places invoices where users naturally expect to find them and removes ambiguity about where billing details can be managed.
- Remove the dedicated /billing/invoices route, controller action, and template
- Remove the Invoices sidebar link (was conditionally shown when a subscription existed)
- Add invoices tile directly to the subscription settings page, hidden when there is no subscription
- Inline invoice fetching into the subscription controller action
- Fix table component padding for tables without a thead
- Add gap spacing to funnel and props settings lists (needed after removing the table padding)

### Tests

### Changelog

### Documentation
- [x] This change does not need a documentation update (docs will be updated when all changes around subscription settings are completed)

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,6 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Keybind hints are hidden on smaller screens
-- Subscription settings page design is updated
-- Invoices are now shown directly on the Subscription settings page instead of a separate tab
 
 ### Fixed
 


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 
 - Keybind hints are hidden on smaller screens
 - Subscription settings page design is updated
+- Invoices are now shown directly on the Subscription settings page instead of a separate tab
```
aerosol: We can skip this, changelog is for self-hosters and there's no billing available in CE.

sanne-san: Ah! Good point, thanks 👍

PR #6095, opened 2026-02-25
Redirect `/billing/invoices` to subscription settings

sanne-san:
### Changes

- The invoices page was recently removed and invoices settings are now part of the subscription settings page. This commit redirects to the subscription settings page.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/settings_controller.ex
+++ b/lib/plausible_web/controllers/settings_controller.ex
@@ -13,7 +13,7 @@ defmodule PlausibleWeb.SettingsController do
        when action in [:update_team_name]
 
   plug Plausible.Plugs.AuthorizeTeamAccess,
-       [:owner, :billing] when action in [:subscription, :redirect_invoices]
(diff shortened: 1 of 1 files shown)


lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -13,7 +13,7 @@ defmodule PlausibleWeb.SettingsController do
        when action in [:update_team_name]
 
   plug Plausible.Plugs.AuthorizeTeamAccess,
-       [:owner, :billing] when action in [:subscription]
+       [:owner, :billing] when action in [:subscription, :redirect_invoices]
```
aerosol: I think the redirect target should have authorization in place, so we don't have to authorize the redirect itself?

sanne-san: Yeah good point, reverted that line 👍

PR #6097, opened 2026-02-25
Add per-site pageview breakdown to pageview usage section

sanne-san:
### Changes
- When a team has between 2 and 10 sites, the expandable pageview usage section now shows a breakdown of pageviews and custom events per site.
- Teams with a single site or more than 10 sites see the existing aggregate-only view unchanged.
- The breakdown section is expanded by default when no per-site breakdown is present. It is collapsed by default when a per-site breakdown is present, or when previous cycles are also present.

### Tests

### Changelog
- [x] This PR does not make a user-facing change (not relevant to self-hosters)

### Documentation
- [x] This change does not need a documentation update (Docs will be updated once all subscription settings changes have been implemented)

### Dark mode

### DIFF
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -483,43 +483,34 @@ defmodule Plausible.Teams.Billing do
       Plausible.Stats.Clickhouse.per_site_usage_breakdown(owned_site_ids, date_range)
 
     active_site_ids = Enum.map(per_site_usage_breakdown, fn {site_id, _, _} -> site_id end)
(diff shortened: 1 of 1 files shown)


aerosol (commented):
~~I have some minor reservations with regards to querying code I'll post shortly, but main concern here is do we want to do that for teams having thousands of sites? IMHO our best best is to limit at some viewable maximum (10 maybe?) and order by top descending.~~ nevermind, it's done, I have to wake up, sorry :sweat_smile:

PR #6103, opened 2026-02-26
Remove unnecessary `ok_to_display?` check from consolidated view template

ukutaht:
### Changes

On the `/sites` page, in order to render a consolidated view, the template checks `ConsolidatedView.consolidated_view_ok_to_display?(@current_team)`.

As far as I can tell, this is not necessary because `Sites.init_consolidated_view_assigns/2` calls `ConsolidatedView.enable/1`, which already checks all the cases that `consolidated_view_ok_to_display?` guards against. The `is_nil(@consolidated_view)` conditional in the template is enough. If the view is not OK to display, the `@consolidated_view` assign never gets set. Checking again in the template is redundant and causes 3 extra database queries on both dead and live render.

### DIFF
--- a/lib/plausible_web/live/sites.ex
+++ b/lib/plausible_web/live/sites.ex
@@ -200,10 +200,7 @@ defmodule PlausibleWeb.Live.Sites do
             current_team={@current_team}
           />
           <.consolidated_view_card
-            :if={
-              not @searching? and not is_nil(@consolidated_view) and
-                consolidated_view_ok_to_display?(@current_team)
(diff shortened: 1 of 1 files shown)


ukutaht:
👍 There are two callsites remaining:
1. Email reports
2. Stats controller

I *think* in both cases the only thing they are really guarding is `Plausible.Billing.Feature.ConsolidatedView.check_availability(team) == :ok`. The other checks seem superfluous but I'm not 100% sure.

PR #6117, opened 2026-03-02
Add dashboard links to monthly pageview usage breakdown

sanne-san:
### Changes

- On the subscription settings page, each billing period's "Total billable pageviews" row and each individual site in the breakdown now show an external link icon that deep-links to the dashboard filtered to that exact billing cycle date range.
- For the total link, the consolidated view domain is preferred when available; otherwise the site's own domain is used when the team has exactly one site.

https://github.com/user-attachments/assets/0fe59d2f-49e8-4842-80fa-5cb0b2f67a08

### Tests

### Changelog
- [x] This PR does not make a user-facing change (not relevant to self-hosters)

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/settings_controller.ex
+++ b/lib/plausible_web/controllers/settings_controller.ex
@@ -154,10 +154,8 @@ defmodule PlausibleWeb.SettingsController do
 
     total_pageview_usage_domain =
       if site_usage == 1 do
-        case Plausible.Teams.owned_sites(team, 1) do
-          [site | _] -> site.domain
-          _ -> nil
-        end
+        [site] = Plausible.Teams.owned_sites(team)
+        site.domain
       else
         on_ee(do: team && get_consolidated_view_domain(team), else: nil)
       end


lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -152,6 +152,16 @@ defmodule PlausibleWeb.SettingsController do
...
 
+    site_domain =
+      if site_usage == 1 do
+        case Plausible.Teams.owned_sites(team, 1) do
+          [site | _] -> site.domain
+          _ -> nil
+        end
+      end
+
+    consolidated_view_domain = on_ee(do: team && get_consolidated_view_domain(team), else: nil)
```
RobertJoonas: What do you think about making this a single assign like `total_pageview_usage_domain` which is evaluated as:

```elixir
# Used to render a dashboard link to view the monthly
# pageview usage of a specific billing cycle. Is `nil` if the
# team has more than one site but no consolidated view.
total_pageview_usage_domain =
  if site_usage == 1  do
    site.domain
  else
    on_ee(do: team && get_consolidated_view_domain(team), else: nil)
  end
```

Naming it this way would make it more obvious why this information is passed into the template.

Another idea that just occurred to me -- if the team has only a single site, then this information is already available in `usage.monthly_pageviews.current_cycle.sites`, right? Perhaps we don't need to pass any single site domain here at all? We could just check whether the usage per site is a list with only a single site. If it is, it means that the `total_link` will be of that site.

lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -162,7 +172,10 @@ defmodule PlausibleWeb.SettingsController do
       site_limit: Teams.Billing.site_limit(team),
       team_member_limit: Teams.Billing.team_member_limit(team),
       team_member_usage: team_member_usage,
-      notification_type: notification_type
+      notification_type: notification_type,
+      consolidated_view_domain: consolidated_view_domain,
+      site_domain: site_domain,
+      team_identifier: team && team.identifier
```
RobertJoonas: Since `current_team` is already available in assigns, we could avoid this extra `team_identifier` assign here. In the `subscription.html` template we could just retrieve the value as:

```elixir
<PlausibleWeb.Components.Billing.render_monthly_pageview_usage
  ...
  team_identifier={@current_team && @current_team.identifier}
/>
```

PS: I think I'm missing some context here, but why do we even need the team identifier for constructing the dashboard links?

lib/plausible_web/components/billing/billing.ex
RobertJoonas: The `@usage.sites` confused me a bit because it looks a lot like we're talking about the "how many sites does this team have" usage. Would it make sense to rename this field to `per_site`?

lib/plausible_web/controllers/settings_controller.ex
sanne-san: No you're right, it's not needed as the user is already on the correct team's settings page, so the correct team is already active in their session.

lib/plausible_web/components/billing/billing.ex
sanne-san: Good point 👌

lib/plausible_web/controllers/settings_controller.ex
sanne-san: Yeah, much clearer 🙏 

As for your idea, I think this won't work as `per_site_usage` returns `[]` whenever a team has one site. To make this work we'd have to change that, which would introduce extra DB queries, and we'd still need to hide the breakdown from the UI anyway. Not sure it's worth it.

lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -152,16 +152,16 @@ defmodule PlausibleWeb.SettingsController do
 
     notification_type = Plausible.Billing.Quota.usage_notification_type(team, usage)
 
-    site_domain =
+    total_pageview_usage_domain =
       if site_usage == 1 do
         case Plausible.Teams.owned_sites(team, 1) do
           [site | _] -> site.domain
           _ -> nil
         end
```
RobertJoonas: Sorry I failed to notice this before. I don't think there's any way that `owned_sites` could return more than one site, especially when we're applying a limit of 1 (which I think we shouldn't do either).

My suggestion:

```elixir
[site] = Plausible.Teams.owned_sites(team)
site.domain
```

`Plausible.Teams.owned_sites` can indeed return `[]` when `team` is `nil`, but from `site_usage == 1`  we know that that's not the case.

The pattern matching with `[site] = ...` is a sanity check -- if `owned_sites` returns something other than a singleton, we let it crash.

PR #6119, opened 2026-03-03
Cache visitor graph results locally

apata:
### Changes

This PR refactors top stats and the main graph to cache results locally. This makes the stats and the main graph load instantly for views that the user has already visited in the session. 
The results are cached until the next time the page is refreshed (normal, not hard refresh), or the page is closed and reopened. 
Realtime results are updated any time that period is selected and then at regular intervals. 

Other changes
- Pulls selected interval and selected metric to component state and passes them to subcomponents as props. This makes React aware of these values, which in turn makes the component updates more predictable.
- Fixes the reserved area jump from a stale height guess if the user rotates their screen to landscape mode and refreshes the page
-  Removes unused sampling notice

### Tests
- [x] Manually tested

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/line-graph.js
+++ b/assets/js/dashboard/stats/graph/line-graph.js
@@ -49,10 +49,7 @@ class LineGraph extends React.Component {
     return function (val, _index, _ticks) {
       if (this.getLabelForValue(val) == '__blank__') return ''
 
-      if (
-        graphData.interval === 'hour' &&
-        dashboardState.period !== 'day'
-      ) {
+      if (graphData.interval === 'hour' && dashboardState.period !== 'day') {
         const date = dateFormatter({
           interval: 'day',
           longForm: false,


assets/js/dashboard/stats/graph/line-graph.js
```diff
@@ -5,7 +5,6 @@ import Chart from 'chart.js/auto'
 import GraphTooltip from './graph-tooltip'
 import { buildDataSet, METRIC_LABELS, hasMultipleYears } from './graph-util'
 import dateFormatter from './date-formatter'
-import FadeIn from '../../fade-in'
```
RobertJoonas, line 8: Could we keep this? Comparing with master, the graph appearing with a tiny fade-in makes it noticeably smoother.

Or does it not play nice with the instant "cached stats render" transition?

assets/js/dashboard/stats/graph/line-graph.js
```diff
@@ -264,20 +263,22 @@ class LineGraph extends React.Component {
...
     })
 
-    return (
-      <FadeIn show={graphData}>
-        <div className="relative h-96 print:h-auto print:pb-8 w-full z-0">
-          <canvas id="main-graph-canvas" className={canvasClass}></canvas>
-        </div>
-      </FadeIn>
-    )
+    return <canvas id="main-graph-canvas" className={canvasClass} />
```
RobertJoonas: Not sure if I'm commenting on the correct line, but just noticed on staging that the padding below the time labels (x-axis of the graph) was removed. Let’s add back that empty space.

assets/js/dashboard/stats/graph/visitor-graph.tsx
```diff
@@ -0,0 +1,295 @@
...
+  })
+
+  // select metric that exists
+  useEffect(() => {
+    if (topStatsQuery.data) {
+      const availableMetrics = topStatsQuery.data.topStats
+        .filter((stat) => stat.graphable)
+        .map((stat) => stat.metric)
+
+      setSelectedMetric((currentlySelectedMetric) => {
```
RobertJoonas, line 107: Just curious: what's the reason for passing this kind of function? Why not just this:

```js
if (currentlySelectedMetric && availableMetrics.includes(currentlySelectedMetric)) {
  setSelectedMetric(currentlySelectedMetric)
} else {
  setSelectedMetric(availableMetrics[0])
}
```

RobertJoonas (commented):
visitor-graph looks a lot cleaner, and caching is definitely a good step for us. Nice work! 🙌

Everything looks good to me codewise. Noticed a couple of UI things that should be fixed before merging though. There's one comment inline about bottom margin, but there's also this:

https://github.com/user-attachments/assets/b498860b-2435-4105-8f9f-acd87be06316

Making two requests is expected in realtime -- one is for current visitors. But what's not expected is the flicker when the stats update. Hope you can see it in the recording.

assets/js/dashboard/stats/graph/visitor-graph.tsx
```diff
@@ -0,0 +1,295 @@
...
+  useEffect(() => {
+    if (validIntervals(site, dashboardState).includes(selectedInterval)) {
+      storeInterval(dashboardState.period, site.domain, selectedInterval)
+    }
+  }, [dashboardState, site, selectedInterval])
+
+  const isRealtime = dashboardState.period === 'realtime'
+  const refetchTopStats = topStatsQuery.refetch
+
+  useEffect(() => {
```
ukutaht, line 162: Seems like only top stats are re-fetched in realtime view. I tested by staying on the realtime view for a few minutes and the main graph does not change after multiple ticks.

assets/js/dashboard/stats/graph/visitor-graph.tsx
```diff
@@ -0,0 +1,295 @@
...
+  )
+
+  const topStatsQuery = useQuery({
+    queryKey: ['top-stats', { dashboardState }] as const,
+    queryFn: async ({ queryKey }) => {
+      const [_, { dashboardState }] = queryKey
+      return await fetchTopStats(site, dashboardState)
+    },
+    placeholderData: (previousData) => previousData,
+    staleTime: Infinity
```
ukutaht: I think this might be too aggressive for when the tab is left open for a long time and the user returns. Something like 30 minutes would get most of the benefit but also make sure the data isn't too stale, what do you think?

assets/js/dashboard/stats/graph/line-graph.js
apata: Thanks! The `show` prop was always truthy for it, so I removed it. Will try to add something like it back in.

assets/js/dashboard/stats/graph/line-graph.js
apata: Oh, nice catch! I did change around which components have margins and made sure there was no jumpiness during load, but missed this one. In this PR, the chart is 4 tailwind units taller. Will fix

assets/js/dashboard/stats/graph/visitor-graph.tsx
apata: Good question. `selectedMetric` state is a reactive value. It should then become a dependency of the `useEffect` hook, because the hook makes use of that reactive value. 

Because the effect changes it sometimes to a new value, the effect's dependencies changed, and the effect would run again immediately. On the second run, it would probably not change the value again (because the effect is idempotent: given the same dependencies, yields the same value), so there wouldn't be an infinite loop, but still inefficient.

If we'd leave it out of the dependency array (and ignore the linter), sometimes the effect might not run when it should: a subtle bug in our code. https://react.dev/learn/removing-effect-dependencies#why-is-suppressing-the-dependency-linter-so-dangerous

assets/js/dashboard/stats/graph/visitor-graph.tsx
apata: Thanks, on it! Also, I noticed that the main-graph is sometimes requested with disallowed interval. Working on that too.

assets/js/dashboard/stats/graph/visitor-graph.tsx
apata: Thanks for pointing it out! Because refreshing throws out all stored data at once, we want the user to refresh as little as possible. That means we need to set really unaggressive cache times, as appropriate by period.

Of the periods that include the present moment, I think probably the shorter the range, the shorter the staleTime should be. For year to date, it can be hours. For month to date, an hour. Last 7 days, 15min, Today, 5min. Realtime data should be no more stale than the last updated ticker shows.

Historical ranges can have much longer cache times, let's say 10h, enough to last one working day. By historical I mean anything that isn't supposed to include the present moment.

apata: Something like this https://github.com/plausible/analytics/pull/6119/changes/eab7baa9c77f08e7389b3c94c4e8dcfc0c3fbfd7#diff-6ebff7af37ed5b145ec3ca60c9db1593545938d3d71b5ed7be93a7c8d64dc7fd

No doubt it could be further improved (take into account native stats begin and comparison date range), but I hope it's good enough for starting out.

ukutaht: Agree with the idea to make it dynamic based on the time range - nice touch!

PR #6106, opened 2026-02-26
Reduce imported opts queries

ukutaht:
Currently a call to `Query.build` will result in reduntant `Repo.preload(site, :completed_imports)` calls:
* 2 times for queries with no comparisons
* 3 times for queries with comparisons

This PR makes two changes to preloading `completed_imports`:
1. Do not preload at all if we are skipping imports anyways due to `:unsupported_interval` or `:unsupported_query`. The biggest win here is that for daily stats (`interval == "hour"`) we will do 0 preloads as opposed to 2 or 3 postgres roundtrips.
2. If preloading is necessary, hoist the preload a bit higher in the call tree so it does not need to be fetched multiple times

### /sites sparklines

I looked into this because the separate sparkline graph queries on `/sites` currently result in 5 identical preloads per site card: 3 preloads for `query_24h_stats` and 2 for `query_24h_intervals`. With default page_size of 24 this will result in 120 postgres queries per page load when one per site would suffice. 

With this PR, this will cut it down from 5 identical redundant queries per site to 0 because with `interval == "hour"` the preloads are not needed. For longer time ranges in works by @aerosol, it will be 2 queries per site. It would be possible to cut it down to 1 per site at the cost of some complexity. For that the `Sparklines.overview_24h` function would have to figure out whether imports are supported for the query and if so, run the preload before calling `query_24h_stats` and `query_24h_intervals`.

However, this would require exposing some of the Query internals and I felt like it isn't worth it at the moment. One duplicated (surely cached on postgres side) query per site is not terrible.

PR #6124, opened 2026-03-04
24h sort by traffic in CRM

aerosol:
This PR implements in-memory sortable site index (discussed with @ukutaht) in CRM.

Initially we aimed to sort by traffic over varying ranges, but dynamic timezone casting per each site individually and no sampling applicable made us hold off for now. Data required to sort and paginate coming from two different sources led me to believe the least-intuitive form of pagination is going to work best here, weighing on testing complexity and raw performance given the specifics of our production database. 

The existing code change was kept to a minimum, just so we can get a feeling on how responsive it becomes for large teams over a websocket.

The PR is large, but it's mostly test coverage.
Benchmark code that I keep tinkering with on the side, and wiring up user-facing site index will be done in follow-up PRs, if the idea poses no immediate threats.

### DIFF
--- a/extra/lib/plausible_web/live/customer_support/team/components/sites.ex
+++ b/extra/lib/plausible_web/live/customer_support/team/components/sites.ex
@@ -50,16 +50,18 @@ defmodule PlausibleWeb.CustomerSupport.Team.Components.Sites do
   def handle_event("sort", %{"by" => by}, socket) do
     sort_by = parse_sort_by(by)
     current_state = socket.assigns.index_state
+    current_sort_by = current_state.sort_by
 
     sort_direction =
-      if current_state.sort_by == sort_by do
-        flip_direction(current_state.sort_direction)
-      else
-        if sort_by == :traffic do
+      case sort_by do
+        ^current_sort_by ->
+          flip_direction(current_state.sort_direction)
+
+        :traffic ->
           :desc
-        else
+
+        :alnum ->
           :asc
-        end
       end
 
     new_state = Index.sort(current_state, sort_by: sort_by, sort_direction: sort_direction)


aerosol:
Oh and would be nice to include https://github.com/plausible/analytics/pull/6106 first, including the patch I submitted against that PR

test/plausible_web/live/customer_support/teams_test.exs
```diff
@@ -912,5 +1041,13 @@ defmodule PlausibleWeb.Live.CustomerSupport.TeamsTest do
         for i <- 1..3, do: assert(text =~ "Entry (#{i})")
       end
     end
+
+    defp extract_domains(html) do
+      html
+      |> Floki.find("tbody tr td:first-child a")
```
ukutaht: Let's use LazyHTML over Floki please. LazyHTML is the modern blessed library that ships with new Liveview versions. We only have a few places left to refactor and then Floki can be removed.

assets/js/app.js
```diff
@@ -1,4 +1,6 @@
 import './polyfills/closest'
+
+window.addEventListener('scroll-to-top', () => window.scrollTo(0, 0))
```
ukutaht: Placing this in between import statements is odd. Is that intentional?

lib/plausible_web/live/components/pagination.ex
```diff
@@ -59,7 +60,8 @@ defmodule PlausibleWeb.Live.Components.Pagination do
 
     ~H"""
     <.link
-      navigate={@uri}
+      patch={@uri}
```
ukutaht, line 63: 👍

lib/plausible_web/live/components/pagination.ex
```diff
@@ -59,7 +60,8 @@ defmodule PlausibleWeb.Live.Components.Pagination do
...
+      phx-click={@active? && JS.dispatch("scroll-to-top")}
```
ukutaht, line 64: I recently looked at how Liveview handles scroll position. This was one of the issues I noticed as well.

I can see this becoming a pretty common utility in which case we might want to encapsulate it better somehow. Not really needed for now, just making a note.

test/plausible_web/live/customer_support/teams_test.exs
ukutaht: https://github.com/plausible/analytics/pull/6125

aerosol: ah I think I've tried it via `Plausible.Test.Support.HTML` helpers but couldn't get it return the same order as in the document

aerosol: I'm going to have a fresh look tomorrow

assets/js/app.js
aerosol: no, random, I was trying to get something/anything to work :sweat_smile:

lib/plausible_web/live/components/pagination.ex
aerosol: We'll definitely use it in main sites index indeed

assets/js/app.js
ukutaht: Fair enough :)

Given that this is functionally only relevant on liveview pages it would make sense to move to live_socket.js so it's only included on liveviews.

test/plausible_web/live/customer_support/teams_test.exs
aerosol: https://github.com/plausible/analytics/pull/6124/commits/d1c7de91e44b05029ba2ec32ed18f57b9292b16f

assets/js/app.js
aerosol: 4c13840dd9

test/plausible_web/live/customer_support/teams_test.exs
aerosol: https://github.com/plausible/analytics/pull/6124/commits/9b7009a6f1eb461bca1e6b82b3df6cbf59c2e264

PR #6133, opened 2026-03-05
Indicate pin status in crm site index

aerosol:


lib/plausible_web/components/icons.ex
```diff
@@ -4,6 +4,26 @@ defmodule PlausibleWeb.Components.Icons do
   """
   use Phoenix.Component
 
+  attr(:rest, :global)
+  attr(:filled, :boolean, default: false)
+
+  def pin_icon(assigns) do
```
ukutaht, line 10: 👌

aerosol:
I'll include this in #6143

PR #6135, opened 2026-03-06
Workloads: Create workloads migration

apata:
### Changes

Contains migration to introduce workloads. Once the migration is run, all queries executed on the cluster, unless explicitly set otherwise, will use the new "default" workload, which is slightly more restrictive than our current query settings.

### Tests

### Changelog

### Documentation

### Dark mode


zoldar (approved):
Per https://clickhouse.com/docs/operations/workload-scheduling#cpu_scheduling and https://clickhouse.com/docs/operations/workload-scheduling#threads_vs_cpus both, scheduling and throttling requires `cpu_slot_preemption` to be set to `true` to work. Otherwise it is ignored. Do we account for that already? Just making sure.

apata:
Closing, not needed right now.

PR #6137, opened 2026-03-07
Skip reading engagements when not necessary

ukutaht:
Adds `WHERE name != 'engagement'` all events_v2 queries that don't need engagement data. That is queries without `time_on_page` or `scroll_depth` metrics, or scroll goals included in dimensions/filters. 

Since the `name` field is included in the events_v2 sort key, engagement events cluster together in storage, allowing ClickHouse to efficiently skip granules that contain only engagement rows when they are not needed.

I don't like that the most natural place for this was in the `where_builder.ex` module. But I didn't find a much better place for this either. Naturally I'd like it to be in the `QueryOptimizer` given the name of that module. But it only operates on the `Query` struct and I didn't want to add another field to that struct.

IMO the query pipeline is overdue for a refactoring where the `Query` struct is stripped down to represent user intent and site context. And all hints about SQL generation, which tables to query, JOIN types, and optimizations like this would be added to an intermediate `ExecutionPlan` struct. But this is a separate topic from the optimization itself.

### DIFF
--- a/extra/lib/plausible/stats/funnel.ex
+++ b/extra/lib/plausible/stats/funnel.ex
@@ -28,6 +28,10 @@ defmodule Plausible.Stats.Funnel do
   end
 
   def funnel(_site, query, %Funnel{} = funnel) do
+    funnel_goals = Enum.map(funnel.steps, & &1.goal)
+
+    query = %{query | preloaded_goals: %{all: funnel_goals}}
+
     funnel_data =
       query
       |> Base.base_event_query()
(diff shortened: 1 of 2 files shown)


ukutaht:
Trying to check if it has some effect on queries on preview env but getting "DB::Exception: Received from localhost:9000. DB::Exception: Too many simultaneous queries. Maximum: 150. (TOO_MANY_SIMULTANEOUS_QUERIES) (version 25.11.5.8 (official build))"

@cnkk do we have too many preview environments at the same time?

PR #6143, opened 2026-03-09
Use Sites.Index on /sites

aerosol:
### Changes

This PR switches `/sites` listing to use `Sites.Index` from https://github.com/plausible/analytics/pull/6124

Scrivener dependency gets removed, defaults to traffic sort descending.

@sanne-san whenever you have time, please have a look at the commits I mentioned you on. 

I'll add another PR on top of it, where per user/team sort criteria preferences will be remembered on each setting.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/aggregate.ex
+++ b/lib/plausible/stats/aggregate.ex
@@ -11,7 +11,8 @@ defmodule Plausible.Stats.Aggregate do
   def aggregate(site, query, metrics) do
     query =
       query
-      |> Query.set(metrics: metrics, remove_unavailable_revenue_metrics: true)
+      |> Query.set(metrics: metrics)
+      |> Query.set_include(:drop_unavailable_revenue_metrics, true)
       |> QueryOptimizer.optimize()
 
     %QueryResult{results: [entry], meta: meta} = QueryRunner.run(site, query)
--- a/lib/plausible/stats/breakdown.ex
+++ b/lib/plausible/stats/breakdown.ex
@@ -32,9 +32,9 @@ defmodule Plausible.Stats.Breakdown do
         filters: query.filters ++ dimension_filters(dimension),
         pagination: %{limit: limit, offset: (page - 1) * limit},
         # Allow pageview and event metrics to be queried off of sessions table
-        legacy_breakdown: true,
-        remove_unavailable_revenue_metrics: true
+        legacy_breakdown: true
       )
+      |> Query.set_include(:drop_unavailable_revenue_metrics, true)
 
     %QueryResult{results: results, meta: meta} = QueryRunner.run(site, query_with_metrics)
 
--- a/lib/plausible/stats/dashboard/query_parser.ex
+++ b/lib/plausible/stats/dashboard/query_parser.ex
@@ -80,7 +80,8 @@ defmodule Plausible.Stats.Dashboard.QueryParser do
          compare_match_day_of_week: params["include"]["compare_match_day_of_week"] == true,
          time_labels: params["include"]["time_labels"] == true,
          trim_relative_date_range: true,
-         drop_unavailable_time_on_page: true
+         drop_unavailable_time_on_page: true,
+         drop_unavailable_revenue_metrics: true
        }}
     end
   end
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -24,7 +24,6 @@ defmodule Plausible.Stats.Query do
             # Revenue metric specific metadata
             revenue_currencies: %{},
             revenue_warning: nil,
-            remove_unavailable_revenue_metrics: false,
             site_id: nil,
             consolidated_site_ids: nil,
             site_native_stats_start_at: nil,
--- a/lib/plausible/stats/query_builder.ex
+++ b/lib/plausible/stats/query_builder.ex
@@ -234,7 +234,8 @@ defmodule Plausible.Stats.QueryBuilder do
     end
 
     defp validate_revenue_metrics_access(site, query) do
-      if Revenue.requested?(query.metrics) and not Revenue.available?(site) do
+      if Revenue.requested?(query.metrics) and not Revenue.available?(site) and
+           not query.include.drop_unavailable_revenue_metrics do
         {:error,
          %QueryError{
            code: :feature_access,
--- a/lib/plausible/stats/query_include.ex
+++ b/lib/plausible/stats/query_include.ex
@@ -9,7 +9,8 @@ defmodule Plausible.Stats.QueryInclude do
             compare: nil,
             compare_match_day_of_week: false,
             legacy_time_on_page_cutoff: nil,
-            drop_unavailable_time_on_page: false
+            drop_unavailable_time_on_page: false,
+            drop_unavailable_revenue_metrics: false
 
   @type date_range_tuple() :: {:date_range, Date.t(), Date.t()}
   @type datetime_range_tuple() :: {:datetime_range, DateTime.t(), DateTime.t()}
@@ -24,6 +25,7 @@ defmodule Plausible.Stats.QueryInclude do
             nil | :previous_period | :year_over_year | date_range_tuple() | datetime_range_tuple(),
           compare_match_day_of_week: boolean(),
           legacy_time_on_page_cutoff: any(),
-          drop_unavailable_time_on_page: boolean()
+          drop_unavailable_time_on_page: boolean(),
+          drop_unavailable_revenue_metrics: boolean()
         }
 end
--- a/lib/plausible/stats/query_optimizer.ex
+++ b/lib/plausible/stats/query_optimizer.ex
@@ -205,7 +205,8 @@ defmodule Plausible.Stats.QueryOptimizer do
 
   on_ee do
     defp remove_revenue_metrics_if_unavailable(query) do
-      if query.remove_unavailable_revenue_metrics and map_size(query.revenue_currencies) == 0 do
+      if query.include.drop_unavailable_revenue_metrics and
+           map_size(query.revenue_currencies) == 0 do
         Query.set(query, metrics: query.metrics -- Plausible.Stats.Goal.Revenue.revenue_metrics())
       else
         query
--- a/lib/plausible/stats/timeseries.ex
+++ b/lib/plausible/stats/timeseries.ex
@@ -23,9 +23,9 @@ defmodule Plausible.Stats.Timeseries do
       |> Query.set(
         metrics: transform_metrics(metrics, %{conversion_rate: :group_conversion_rate}),
         dimensions: [time_dimension(query)],
-        order_by: [{time_dimension(query), :asc}],
-        remove_unavailable_revenue_metrics: true
+        order_by: [{time_dimension(query), :asc}]
       )
+      |> Query.set_include(:drop_unavailable_revenue_metrics, true)
       |> QueryOptimizer.optimize()
 
     comparison_query =
--- a/lib/plausible_web/controllers/helpers.ex
+++ b/lib/plausible_web/controllers/helpers.ex
@@ -4,6 +4,7 @@ defmodule PlausibleWeb.ControllerHelpers do
 
   def render_error(conn, status, message) do
     conn
+    |> put_root_layout(false)
     |> put_status(status)
     |> put_view(PlausibleWeb.ErrorView)
     |> render("#{status}.html", message: message, layout: error_layout())
@@ -11,6 +12,7 @@ defmodule PlausibleWeb.ControllerHelpers do
 
   def render_error(conn, status) do
     conn
+    |> put_root_layout(false)
     |> put_status(status)
     |> put_view(PlausibleWeb.ErrorView)
     |> render("#{status}.html", layout: error_layout())
--- a/test/plausible_web/controllers/api/stats_controller/internal_query_api_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/internal_query_api_test.exs
@@ -10,6 +10,35 @@ defmodule PlausibleWeb.Api.StatsController.InternalQueryApiTest do
   setup [:create_user, :log_in, :create_site]
 
   describe "aggregates (e.g. top stats)" do
+    on_ee do
+      test "drops revenue metrics when they are not allowed", %{
+        conn: conn,
+        user: user,
+        site: site
+      } do
+        {:ok, team} = Plausible.Teams.get_or_create(user)
+        insert(:goal, site: site, event_name: "Purchase", currency: "EUR")
+        subscribe_to_growth_plan(team)
+
+        params = %{
+          "date_range" => "day",
+          "filters" => [["is", "event:goal", ["Purchase"]]],
+          "metrics" => [
+            "visitors",
+            "events",
+            "conversion_rate",
+            "total_revenue",
+            "average_revenue"
+          ]
+        }
+
+        %{"results" => results, "query" => query} = do_query_success(conn, site, params)
+
+        assert query["metrics"] == ["visitors", "events", "conversion_rate"]
+        assert results == [%{"dimensions" => [], "metrics" => [0, 0, 0.0]}]
+      end
+    end
+
     test "returns empty metrics when no data", %{conn: conn, site: site} do
       requested_metrics = [
         "visitors",


ukutaht:
Haven't done a full code review but some visual stuff I noticed (cc @sanne-san):

1. We should provide UI feedback immediately when dropdown item is clicked and page is loading.  The most minimal thing that would work: https://github.com/plausible/analytics/commit/0f129a6473b13ece412d90c5dcaf1934b284ec56. Try with `socket.enableLatencySim(500)`.

2. The search input has 38px height, sort dropdown 42px and "Add Site" button 40px. I think there are multiple issues here. From a cursory look we've mostly standardized on 40px inputs and buttons. But `<.filter_bar>` is using a custom input that has a different size. Buttons have their own issue: the `secondary` and `danger` button styles are 2px higher compared to `primary` style because they use broders. I think it's an overall consistency issue but just very noticeable here since we're mixing three different sizes side-by-side.

3. I'm struggling to propose something concrete about this but I'm thinking maybe there's something better than using _"Visitors, high to low"_, _"Name, A-Z"_ as the dropdown copy. Maybe up/down arrow icons or something as a visual indicator would be better. Not sure, more of an open-ended thought to play around with this.

sanne-san:
@ukutaht thanks for noticing these – I had a bit of a rushed look yesterday, my bad. Fixed all the issues. On #3, I changed it to "Most visitors" and "Fewest visitors" for now, as we also agreed upon in the card. It's less ambiguous than arrows, but if we'll add more options (pageviews, conversion, etc.) we might have to switch to arrows anyway. Wdyt? 

I also switched from using the button css to using dedicated css for the dropdown. It's largely the same but diverges in a few ways (e.g. if the dropdown has a width that is wider than its content, it was now centering all content rather than content on the left and chevron on the right). I'm not sure but I think decoupling these styles altogether makes sense? I also ensured that the dropdown menu min-width equals the width of the trigger. It works, but not sure it's the right pattern.

aerosol:
From my point of view it's ready, pinning (+CRM), migration and persistence of choice included. Review(s) appreciated @plausible/developers

ukutaht:
@sanne-san looks good!

> I also switched from using the button css to using dedicated css for the dropdown. It's largely the same but diverges in a few ways (e.g. if the dropdown has a width that is wider than its content, it was now centering all content rather than content on the left and chevron on the right). I'm not sure but I think decoupling these styles altogether makes sense?

makes sense to me

> also ensured that the dropdown menu min-width equals the width of the trigger. It works, but not sure it's the right pattern.

The implementation itself is sound but the way it's patching Prima is quite hacky and error-prone IMO. Matching the trigger size is a common pattern for dropdowns so I'm thinking it would be best to add it as an option to Prima itself.

Not a blocker for this PR

ukutaht:
@sanne-san I've added this to Prima now. Upgrade to 0.2.6 to get access to it. See example "Match Trigger Width" [here](https://prima-lingering-feather-539.fly.dev/dropdown).

lib/plausible_web/live/sites.ex
```diff
@@ -18,14 +19,37 @@ defmodule PlausibleWeb.Live.Sites do
...
+        sort_direction: sort_direction
+      )
+
     socket =
       socket
       |> assign(:uri, uri)
       |> assign(:sparklines, %{})
-      |> assign(:filter_text, String.trim(params["filter_text"] || ""))
+      |> assign(:filter_text, filter_text)
+      |> assign(:sort_by, sort_by)
```
ukutaht: One thing that sticks out to me is treating `sort_by` and `sort_direction` as independent states. They feel so closely coupled that I think representing them as a single data structure (tuple or struct) would be a more meaningful representation.

Not just in assigns but also in the database, so accessing and updating can be done via single DB query. If we make it possible in the future to change them independently, it would still always happen in a context where both are available in code so we can update the whole structure.

lib/plausible_web/live/sites.ex
```diff
@@ -84,11 +129,15 @@ defmodule PlausibleWeb.Live.Sites do
   end
 
   def render(assigns) do
-    assigns = assign(assigns, :searching?, String.trim(assigns.filter_text) != "")
+    assigns =
```
ukutaht: Note for myself: assigning in `render` function feels very smelly to me. Check what's going on here.

lib/plausible_web/live/sites.ex
```diff
@@ -685,13 +752,58 @@ defmodule PlausibleWeb.Live.Sites do
...
+  ]
+
+  def sort_dropdown(assigns) do
+    current_label =
+      Enum.find_value(@sort_options, fn {label, {sort_by, direction}} ->
+        if sort_by == assigns.sort_by and direction == assigns.sort_direction,
+          do: label
+      end)
+
+    assigns = assign(assigns, sort_options: @sort_options, current_sort_label: current_label)
```
ukutaht: Again assigning in `render` function. Check why

lib/plausible_web/live/sites.ex
```diff
@@ -18,14 +19,37 @@ defmodule PlausibleWeb.Live.Sites do
...
+        {nil, nil} ->
+          load_sort_preference(user, team)
+
+        {sort_by, sort_direction} ->
+          {sanitize_sort_by(sort_by), sanitize_sort_direction(sort_direction)}
+      end
+
+    index_state =
+      SitesIndex.build(user,
+        filter_by_domain: filter_text,
```
ukutaht: Storing `filter_text` in both liveview state and SitesIndex state feels wrong to me. Instead of trusting a single source of truth it requires synchronization (`SitesIndex.update_state`).

From my experience with stateful UIs (=React), storing the same state in multiple places is a common but avoidable footgun. It just opens the possibility for a category of bugs that otherwise would not be possible.

The state is already present in liveview so we should be able to pass it as a param to `SitesIndex.paginate` and remove the `filter_by_domain` state from `SitesIndex`.

Same could be said about sorting state. Liveview can own that and only pass it when necessary: `build()` for initial sorting and `sort()` for updating.

lib/plausible_web/live/sites.ex
```diff
@@ -898,6 +1042,32 @@ defmodule PlausibleWeb.Live.Sites do
...
+      socket.assigns.current_user,
+      socket.assigns.current_team,
+      sort_by,
+      sort_direction
+    )
+
+    socket
+    |> assign(:sort_by, sort_by)
+    |> assign(:sort_direction, sort_direction)
+    |> assign(:uri, uri)
```
ukutaht: Note: I'll check if we can avoid keeping `uri` in the state. I feel like it's not necessary. The more idiomatic way seems to be something like:

```elixir
defp set_sort(socket, sort_by, sort_direction) do
    params =
      socket.assigns.params
      |> Map.put("sort_by", to_string(sort_by))
      |> Map.put("sort_direction", to_string(sort_direction))
      |> Map.delete("page")

    push_patch(socket, to: Routes.sites_path(:index, params))
  end
```

lib/plausible_web/live/sites.ex
```diff
@@ -84,11 +129,15 @@ defmodule PlausibleWeb.Live.Sites do
   end
 
   def render(assigns) do
-    assigns = assign(assigns, :searching?, String.trim(assigns.filter_text) != "")
+    assigns =
+      assigns
+      |> assign(:searching?, String.trim(assigns.filter_text) != "")
+      |> assign_new(:sort_by, fn -> :traffic end)
```
ukutaht: Are these necessary? Seems like it's here to apply defaults but I feel like it would be cleaner for `mount()` and `handle_params()` to be responsible for managing this kind of defaulting behaviour so template doesn't have to worry about missing data.

lib/plausible_web/live/sites.ex
aerosol: Right, it was my initial thought too. But what if we add ranges to the mix? Would you still think about it as a single data structure?

lib/plausible_web/live/sites.ex
aerosol: > Same could be said about sorting state. Liveview can own that and only pass it when necessary: build() for initial sorting and sort() for updating.

Disagree on that one though. Previous sort state we should keep track of, because it allows us to make decision on how to sort - e.g. if only order changes, we'll flip it and won't recompute anything. I get the urge to decouple for the sake of it, but the reason we keep it in the UI is different than why we keep it in the index internally.

aerosol:
@ukutaht added a series of commits addressing your remarks, including a new migration for storing sort preferences. Upon approval I'll extract to separate PR. I've also simplified LV/index interactions and state maintenance. Let me know if that looks any better please.

lib/plausible/sites/index.ex
```diff
@@ -71,21 +75,24 @@ defmodule Plausible.Sites.Index do
   @doc """
   Builds an `Index.State` for `user` by running all necessary queries
   """
-  @spec build(Auth.User.t(), [list_opt()]) :: State.t()
-  def build(user, opts \\ []) do
+  @spec build(Auth.User.t(), [index_option()] | map()) :: State.t()
+  def build(user, opts \\ [])
+
+  def build(user, opts) when is_map(opts) do
+    build(user, Keyword.new(opts))
```
zoldar, line 89: Is it safe to assume that keyword opts passed from the outside can be assumed be well-formed while map implies passing parameters from user input? Then, perhaps, it would make sense for the `is_map` guarded clauses of both, `build` and `paginate`, to apply all the casting and then pass the constructed keyword list back to the other clause.

Then:

- the type definition for keyword opts can be tightened (so _in theory_, dialyzer should reject misuse, like passing string/nil for a number in keyword list)
- casting is better separated from the actual logic

lib/plausible/teams/sites.ex
```diff
@@ -155,4 +57,36 @@ defmodule Plausible.Teams.Sites do
...
+                site: s
+              }
+            ]
+        }
+      )
+      |> Repo.all()
+
+    # Restore the caller-supplied order
+    by_id = Map.new(rows, &{&1.id, &1})
+    Enum.flat_map(site_ids, fn id -> List.wrap(Map.get(by_id, id)) end)
```
zoldar:
```suggestion
        select: {s.id , %{
          s
          | pinned_at: selected_as(up.pinned_at, :pinned_at),
            memberships: [
              %{
                role: type(u.role, ^@role_type),
                site_id: s.id,
                site: s
              }
            ]
        }}
      )
      |> Repo.all()
      |> Map.new()

    # Restore the caller-supplied order
    Enum.flat_map(site_ids, fn id -> List.wrap(Map.get(rows, id)) end)
```

lib/plausible_web/components/generic.ex
```diff
@@ -1201,7 +1201,7 @@ defmodule PlausibleWeb.Components.Generic do
             />
 
             <Heroicons.backspace
-              :if={String.trim(@filter_text) != ""}
+              :if={String.trim(@filter_text || "") != ""}
```
zoldar: Huh, there's a default and it still can come as `nil`?

lib/plausible_web/live/sites.ex
```diff
@@ -17,28 +18,55 @@ defmodule PlausibleWeb.Live.Sites do
...
-      ("/sites?" <> URI.encode_query(Map.take(params, ["filter_text"])))
-      |> URI.new!()
+    uri_params = sanitize_uri_params(params)
+
+    index_options =
+      params
+      |> get_index_options(user, team)
+      |> Map.merge(%{
+        team: team
+      })
```
zoldar:
```suggestion
      |> Map.put(:team, team)
```

lib/plausible_web/live/sites.ex
```diff
@@ -56,12 +84,12 @@ defmodule PlausibleWeb.Live.Sites do
           sites: sites,
           current_team: current_team,
           has_sites?: has_sites?,
-          filter_text: filter_text
+          uri_params: uri_params
```
zoldar: Wouldn't it make more sense to assign to `filter_text` from `uri_params` (sanitizing it too) before `then` block in the pipeline? Or is skipping that intentional? I assume updates to `filter_text` in input events sanitize it already (or at least do not allow it to be `nil`).

lib/plausible_web/live/sites.ex
```diff
@@ -84,11 +112,11 @@ defmodule PlausibleWeb.Live.Sites do
   end
 
   def render(assigns) do
-    assigns = assign(assigns, :searching?, String.trim(assigns.filter_text) != "")
+    assigns = assign(assigns, :searching?, assigns.uri_params["filter_text"] not in ["", nil])
```
zoldar: Ditto about keeping `assigns.filter_text` up to date in `handle_params`

lib/plausible_web/live/sites.ex
```diff
@@ -501,7 +537,10 @@ defmodule PlausibleWeb.Live.Sites do
         )
       }
     >
-      <.unstyled_link href={"/#{URI.encode_www_form(@site.domain)}"} class="block">
+      <.unstyled_link
+        href={"/#{URI.encode_www_form(@site.domain)}"}
```
zoldar:
```suggestion
        href={Routes.stats_path(@socket, :stats, @site.domain, [])}
```

lib/plausible_web/live/sites.ex
```diff
@@ -684,14 +725,60 @@ defmodule PlausibleWeb.Live.Sites do
...
+  attr(:sort_by, :atom, default: :traffic)
+  attr(:sort_direction, :atom, default: :desc)
 
   def search_form(assigns) do
     ~H"""
     <.filter_bar filter_text={@filter_text} placeholder="Search Sites"></.filter_bar>
     """
   end
 
+  @sort_options %{
```
zoldar: Would it make sense to turn that into a list of tuples so that dropdown item order is predictable?

lib/plausible_web/live/sites.ex
```diff
@@ -858,44 +970,46 @@ defmodule PlausibleWeb.Live.Sites do
...
-    uri = socket.assigns.uri
-
-    uri_params =
-      uri.query
-      |> URI.decode_query()
-      |> Map.put("filter_text", filter_text)
-      |> URI.encode_query()
-
-    uri = %{uri | query: uri_params}
+    uri_params = Map.put(socket.assigns.uri_params, "filter_text", String.trim(filter_text))
```
zoldar: Ok, I can see where getting rid of  `filter_text` from assigns is coming from. However, I still think it would be good to keep it? It would help improve the clarity of other uses where constant extractions from `uri_params` adds a lot of noise IMHO.

lib/plausible_web/live/sites.ex
```diff
@@ -954,4 +1068,62 @@ defmodule PlausibleWeb.Live.Sites do
...
+    else
+      _ -> @default_sort
+    end
+  end
+
+  defp save_sort_preference(_user, nil, _sort_by, _sort_direction), do: :ok
+
+  defp save_sort_preference(user, team, sort_by, sort_direction) do
+    with {:ok, membership} <- Teams.Memberships.get_team_membership(team, user) do
+      Teams.Memberships.set_preference(membership, :sort_index_options, %{
```
zoldar, line 1088: Maybe that could be an embed which could also be reused for sanitizing the input from params? Feel free to ignore if you think that'd be on overkill.

lib/plausible/teams/sites.ex
aerosol: 9bb1c8338e

lib/plausible_web/components/generic.ex
aerosol: Yeah, if we pass it

lib/plausible_web/live/sites.ex
aerosol: Yeah, it used to be that, I forgot about keys order being undefined, will do

lib/plausible_web/live/sites.ex
aerosol: alright

aerosol: lmk if this is what you meant be175d4660

lib/plausible_web/live/sites.ex
aerosol: lmk if this is what you meant be175d4660

lib/plausible_web/components/generic.ex
aerosol: no longer the case: be175d4660

lib/plausible_web/live/sites.ex
aerosol: be175d4660

lib/plausible_web/live/sites.ex
zoldar: yep!

lib/plausible_web/live/sites.ex
aerosol: aeb7d740b3

lib/plausible_web/live/sites.ex
aerosol: 35b5fe126e

lib/plausible/sites/index.ex
aerosol: Map is basically passed from the database, and this is why I've added that clause, so we can match on keys/values rather than rely on strict ordering.

zoldar: WDYT about defining an embed for that map then, like suggested in https://github.com/plausible/analytics/pull/6143#discussion_r2939561261?

lib/plausible_web/live/sites.ex
ukutaht: Fair enough, make sense.

lib/plausible_web/live/sites.ex
aerosol: Something like https://github.com/plausible/analytics/pull/6143/changes/cc607657acfc585f0eff383246e31c61eb5f5307? Should we use the struct everywhere then, regardless if it comes from the db?

aerosol: hold on, I found the way to structure it nicely :hourglass:

aerosol: https://github.com/plausible/analytics/pull/6143/commits/bdd0dff5919f5df8e61c5ec90b1fb71c64c40706

zoldar: Nice!

PR #6144, opened 2026-03-09
Skip meta.key/meta.value reads when no goal has custom props

ukutaht:
### Changes

I noticed that many slow production queries are caused by unnecessary reading of `meta.key` and `meta.value` columns for `event:goal` breakdowns.

This PR adds an `optimized_conversions` feature flag that enables an optimized SQL path for event:goal queries. When enabled and none of the goals in the query have custom property filters, the query uses a new `event_goal_join_no_props` fragment that omits all references to `meta.key` and `meta.value`. This avoids reading these columns for the query. For some of the biggest conversion reports I observed ~80% less bytes read and around the same speedup in terms of query duration (comparing existing queries with hand-written alternatives that omit reading meta fields).

This doesn't help if site has any goals with custom properties. Future possibility is to explore adding a bloom filter to `meta.key` field to speed up those queries with custom props as well.

Adding it with a feature flag so I can spot-check against some production sites before enabling for all.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/goals.ex
+++ b/lib/plausible/stats/goals.ex
@@ -71,8 +71,7 @@ defmodule Plausible.Stats.Goals do
           page_regexes: [String.t()],
           scroll_thresholds: [non_neg_integer()],
           custom_props_keys: [[String.t()]],
-          custom_props_values: [[String.t()]],
-          has_custom_props: boolean()
+          custom_props_values: [[String.t()]]
         }
 
   @doc """
@@ -81,8 +80,6 @@ defmodule Plausible.Stats.Goals do
   @spec goal_join_data(Plausible.Stats.Query.t()) :: goal_join_data()
   def goal_join_data(query) do
     goals = query.preloaded_goals.matching_toplevel_filters
-
-    has_custom_props = Enum.any?(goals, &Plausible.Goal.has_custom_props?/1)
 
     goals
     |> Enum.with_index(1)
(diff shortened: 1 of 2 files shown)


aerosol (approved):
look like we were close to that at one point: https://github.com/plausible/analytics/pull/5936#discussion_r2605992155

lib/plausible/stats/goals.ex
```diff
@@ -71,7 +71,8 @@ defmodule Plausible.Stats.Goals do
           page_regexes: [String.t()],
           scroll_thresholds: [non_neg_integer()],
           custom_props_keys: [[String.t()]],
-          custom_props_values: [[String.t()]]
+          custom_props_values: [[String.t()]],
+          has_custom_props: boolean()
```
aerosol: Couldn't we check custom_props_keys/custom_props_values instead?

ukutaht:
> look like we were close to that at one point: https://github.com/plausible/analytics/pull/5936#discussion_r2605992155

Yeah, you were right to worry about the extra scanning of `meta` fields. It's surprisingly expensive.

lib/plausible/stats/goals.ex
ukutaht: Good point. https://github.com/plausible/analytics/pull/6144/changes/4fb152eea6cced571ce2faa689f0ab1b1da121bb

PR #6145, opened 2026-03-09
Optimize name filter

ukutaht:
New version of https://github.com/plausible/analytics/pull/6137, this PR also includes name filter for queries that involve a goal and adds a feature flag for production testing.

Adds a `WHERE name` filter to queries based on two conditons:
* If goals involved, use `WHERE name IN (?)`
* If it's safe to skip engagement events, add `WHERE name != 'engagements'`
* Otherwise, scan all events

Since the name field is included in the events_v2 sort key, events cluster together in storage by `name` field, allowing ClickHouse to efficiently skip granules if event names are included in `WHERE`.

### DIFF
--- a/lib/plausible/stats/query_builder.ex
+++ b/lib/plausible/stats/query_builder.ex
@@ -183,7 +183,8 @@ defmodule Plausible.Stats.QueryBuilder do
       revenue_warning: revenue_warning,
       revenue_currencies: revenue_currencies,
       debug_metadata: debug_metadata,
-      optimize_name_filter?: FunWithFlags.enabled?(:optimize_name_filter, for: site)
+      optimize_name_filter?: FunWithFlags.enabled?(:optimize_name_filter, for: site),
+      optimized_conversions: FunWithFlags.enabled?(:optimized_conversions, for: site)
     )
   end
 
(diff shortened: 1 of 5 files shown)


lib/plausible/stats/query_builder.ex
```diff
@@ -182,7 +182,8 @@ defmodule Plausible.Stats.QueryBuilder do
       preloaded_goals: preloaded_goals,
       revenue_warning: revenue_warning,
       revenue_currencies: revenue_currencies,
-      debug_metadata: debug_metadata
+      debug_metadata: debug_metadata,
+      optimize_name_filter: FunWithFlags.enabled?(:optimize_name_filter, for: site)
```
aerosol: Any reason not to use the `?`-suffix convention?

ukutaht:
> We could consider adding {:ok, true} = FunWithFlags.enable(:optimize_name_filter) to test_helper.exs, before going live with flag removed, so hopefully it gets picked up by e2e tests too.

Hmm the reason I didn't do it this way is because the default code path will still be running for most users. I suppose the most appropriate thing would be to run tests both with the flag enabled and without it but I'm not sure if that's feasible or worth the extra time. Checking locally, all tests pass. I'm inclined to continue as-is and remove the flag quickly after some production checks.

lib/plausible/stats/query_builder.ex
ukutaht: https://github.com/plausible/analytics/pull/6145/changes/d1b351774ddad88d78aa3f74c6bd9f3e87110836

PR #6152, opened 2026-03-10
Move main graph to new API v2 endpoint

RobertJoonas:
### Changes

- [x] Implement the required API v2 logic to support everything that main-graph needs 
- [x] A major rewrite of main_graph_test.exs - populate_stats and other tests setup remained the same, but the request is now made against the new endpoint and the assertions on the returned data structures have changed significantly.
- [x] When `include.time_labels` is set, also include `meta.time_label_result_indices` (right next to `meta.time_labels` themselves). This helps to avoid searching for rows corresponding to a time label on the client side.
- [x] Support returning comparisons in the API v2 response format. This was tricky to figure out. See below for details.
- [x] Fix a known bug in API v2 (see commit with description 4923839c5291cf941be6db8984f66358412ec3eb)
- [x] Remove the main graph endpoint
- [x] Remove comparisons logic from legacy timeseries (see commit with description 7ad120a2e0ff2362a27927f6b14ff0cfbdd9b25d)
- [ ] Wire up the UI properly
- [ ] Update changelog

###  New comparison response format for timeseries

Unlike comparisons with empty dimensions or non-time dimensions, the number of dimension groups returned for a time dimension query is _not_ guaranteed to match between original and comparison results. The simplest example: [a custom period comparison](https://plausible.io/plausible.io?period=28d&comparison=custom&compare_from=2025-11-01&compare_to=2025-12-31).

In order for the response to indicate which original group corresponds to which comparison group, we had two options:

**1. Stick with the current response structure where a comparison map is merged into each row of results**

This was the initial solution that this PR tried to implement. The reason this approach seemed better was keeping a single response structure of comparisons. For non-time dimension queries, this format makes perfect sense and is easy to read.

However, the following cons have contributed to choosing option 2 instead:

* Need to zip `results` and `comparison_results` padding the shorter result set with `nil`s if necessary.
* Unlike in a regular (without comparisons) timeseries response, construct rows with empty metrics in both lists where necessary. This is because ClickHouse does not return time buckets that didn't have any data in them.
* Both of the above contribute to a heavier payload
* The main-graph component would have to implement 2 different ways of rendering the graph - iterating over meta.time_labels (when not comparing) and iterating over `results` (when comparing)

**2. Return `comparison_results` as a separate response field when a time dimension is involved**

Even though a separate form of returning comparisons seems inconsistent , it makes much more sense for comparison results to be separated in timeseries as it alleviates all the cons mentioned in option 1.

### Tests

### Changelog
- [ ] Needs a changelog update

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query_include.ex
+++ b/lib/plausible/stats/query_include.ex
@@ -6,7 +6,7 @@ defmodule Plausible.Stats.QueryInclude do
             time_labels: false,
             # `time_label_result_indices` is a convenience for our main graph component. It
             # is not yet ready for a public API release because it should also account for
-            # breakdowns by mutliple dimensions (time + non-time). Also, at this point it is
+            # breakdowns by multiple dimensions (time + non-time). Also, at this point it is
             # still unclear whether `time_labels` will stay in the public API or not.
             time_label_result_indices: false,
             present_index: false,


RobertJoonas:
Closing this in favor of a cleaner rebased version: https://github.com/plausible/analytics/pull/6173.

PR #6154, opened 2026-03-11
Enable WHERE name optimization for all queries, remove feature flag

ukutaht:
Follow-up to https://github.com/plausible/analytics/pull/6145

Confirmed the optimization is working in production with spot checks.

### DIFF
--- a/lib/plausible/stats/query.ex
+++ b/lib/plausible/stats/query.ex
@@ -31,8 +31,7 @@ defmodule Plausible.Stats.Query do
             # Contains information to determine how to combine legacy and new time on page metrics
             time_on_page_data: %{},
             sql_join_type: :left,
-            smear_session_metrics: false,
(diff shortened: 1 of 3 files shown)


ukutaht:
Results from one test. `visit:source` breakdown with `event:props` filter from a large client. Two identical queries with the only difference being the presence of `WHERE name != 'engagement'` filter:

before: 65.25 GiB read bytes, query duration 3,423ms (SelectedMarks':10936, 'FunctionExecute':144,842)
after: 35.20 GiB read bytes, query duration 1,383ms ('SelectedMarks':5919, 'FunctionExecute':84,162)

This optimization will only affect larger sites (due to granule size, smaller sites will not have sufficient clustering in storage) but the improvements should be pretty broad across many queries that scan the events table.

aerosol (approved):
🚀

PR #6155, opened 2026-03-11
Raise errors in Otel/Sentry

cnkk:


config/runtime.exs
```diff
@@ -626,7 +626,7 @@ ch_transport_opts =
 config :plausible, Plausible.ClickhouseRepo,
   queue_target: 500,
   queue_interval: 2000,
-  timeout: 15_000,
+  timeout: 1_000,
```
aerosol, line 629: err don't think we can afford that

lib/plausible/stats/query_runner.ex
```diff
@@ -111,9 +114,37 @@ defmodule Plausible.Stats.QueryRunner do
...
   defp execute_query(query, site) do
-    query
-    |> SQL.QueryBuilder.build(site)
-    |> ClickhouseRepo.all(query: query)
+    try do
+      query
+      |> SQL.QueryBuilder.build(site)
+      |> ClickhouseRepo.all(query: query)
+    rescue
+      e in [Ch.Error, DBConnection.ConnectionError, Mint.TransportError] ->
```
aerosol, line 122: Don't we have all that in `log_comment`? 

ConnectionError AFAIU may be unrelated to the query itself, but a failing checkout for example caused by the query previously made - see https://github.com/plausible/ch/pull/271

PR #6159, opened 2026-03-12
Refactor main graph to d3.js

apata:
Refactors line graph to Typescript, from chart.js library to d3.js library. Depends on the new query backend. https://github.com/plausible/analytics/pull/6173

The rationale for switching the library:

- we need d3.js for the complicated world map (can't be done with chart.js)
- having two charting libraries increases the size of dashboard javascript and therefore makes the website take longer to load

- [x] Test side-by-side
- [x] Remove console.log of "effect running"
- [x] Delete line-graph.js (-301)
- [x] Delete graph-util.js (-127)
- [x] Delete graph-tooltip.js (-216)
- [x] Delete date-formatter.js (-115)

### Changes
- Tooltip is aligned to top left of finger always, but constrained within the chart left to right
- Mobile interactivity (drag left and right to see tooltip, release finger to zoom in on currently shown period)
- Stops x tick texts from rotating, adds subtle x ticks

### Fixes
- Fixes issue with invalid x axis labels on minute resolution charts
- Fixes issue with ambiguous labels in tooltip when comparing year over year

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/util/date.js
```diff
@@ -3,6 +3,14 @@ import utc from 'dayjs/plugin/utc'
...
 dayjs.extend(utc)
 
+const browserDateFormat = Intl.DateTimeFormat(navigator.language, {
+  hour: 'numeric'
+})
+
+export function is12HourClock() {
+  return browserDateFormat.resolvedOptions().hour12
+}
+
```
apata, line 13: Copied from date-formatter.js that was deleted.

assets/js/dashboard/components/graph-tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, [children, className, minWidth])
+
+  return (
+    <Transition
+      as={React.Fragment}
+      appear
+      show
+      // enter delay on mobile is needed to prevent the tooltip from entering when the user starts to y-pan
+      // but the y-pan is not yet certain
+      enter={isTouchDevice ? 'transition-opacity duration-0 delay-150' : ''}
```
RobertJoonas, line 46: This is a nice improvement on mobile! Holding a finger a bit longer on the graph is quite intuitive and it's much better to not let any tooltips show up to just disturb the experience.

The only thing that's bothering me a bit is that text outside the graph container (precisely the Channels tab text) gets selected when I tap-and-hold inside the graph container, which in turn makes the OS render a magnifying glass for text selection right next to my finger (instead of the actual tooltip). When I move my finger a bit, it displays the tooltip though.

^ something to fix another day I guess. Nice work!

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -0,0 +1,75 @@
...
+  long: string
+  currency: string
+}
+
+export type ResultItem = {
+  dimensions: [string] // one item
+  metrics: MetricValues // one item
+}
+
+export type MetricValues = [null] | [number] | [RevenueMetricValue]
```
RobertJoonas: Could we also export a MetricValue type here? We might even want to use a single type across the whole dashboard in the future, but for now we could just export it from here:

```suggestion
export type MetricValue = null | number | RevenueMetricValue
export type MetricValues = [MetricValue]
```

Then the inlined `RevenueMetricValue | number | null` types in `main-graph.tsx` and `main-graph-data.ts` could be replaced with `MetricValue` as well.

assets/js/dashboard/stats/graph/main-graph.tsx
```diff
@@ -0,0 +1,644 @@
...
+      data
+    })
+
+    const gradients = [primaryGradient, secondaryGradient]
+
+    let yMax = 1
+
+    // can't be done in a single pass with remapAndFillData
+    // because we need the xLabels formatting parameters to be known
+    const remappedDataInGraphFormat = remappedData.map(
```
RobertJoonas, line 112: If I'm reading this right, the only reason for not being able to find `yMax` in a single pass with `remapAndFillData` is knowing how to evaluate `shouldShowDate` and `shouldShowYear`, which come from first/last time label from main/comparison results.

Since `meta.time_labels` are the source of truth of x-labels, couldn't we just do:

```
mainFirstLabel = meta.time_labels[0]
mainComparisonLabel = meta.comparison_time_labels[0]
```

(and same for the last labels of both)?

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  const statsQuery = createStatsQuery(dashboardState, reportParams)
+
+  if (isRealTimeDashboard(dashboardState)) {
+    statsQuery.date_range = DashboardPeriod.realtime_30m
+  }
+
+  statsQuery.include.present_index = true
```
RobertJoonas: This is unused now, right? Not pushing to get rid of it yet, but just wanted to flag it. Let's see if the new dotted line definition works out first...

assets/js/dashboard/stats/graph/main-graph-data.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ */
+export const remapAndFillData = ({
+  data,
+  getNumericValue,
+  getValue,
+  getChange
+}: {
+  data: MainGraphResponse
+  getNumericValue: (metrics: RevenueMetricValue | number | null) => number
```
RobertJoonas: This is one of those cases (but there are more) where we could replace the inlined type with a proper MetricValue. Also, `metricValue` would be a better argument name in this function too:

```suggestion
  getNumericValue: (metricValue: MetricValue) => number
```

assets/js/dashboard/stats/graph/main-graph-data.ts
```diff
@@ -0,0 +1,244 @@
...
+        partialTimeLabels
+      }: {
+        timeLabel: string
+        indexOfResult: number | null
+        results: Array<ResultItem | null>
+        partialTimeLabels: string[]
+      }): SeriesValue => {
+        const isPartial = partialTimeLabels.find((l) => l === timeLabel)
+          ? true
+          : false
```
RobertJoonas: I think this would do the trick too?

```suggestion
        const isPartial = partialTimeLabels.includes(timeLabel)
```

assets/js/dashboard/stats/graph/main-graph-data.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  return Math.round(((value - comparisonValue) / comparisonValue) * 100)
+}
+
+export const REVENUE_METRICS = ['average_revenue', 'total_revenue']
+
+export type LineSegment = {
+  startIndexInclusive: number
+  stopIndexExclusive: number
```
RobertJoonas, line 147: Non-blocking suggestion (feel free to ignore): Maybe use shorter names here. E.g.: `fromIndex` and `toIndex` without specifying whether the index is inclusive or exclusive?

assets/js/dashboard/stats/graph/main-graph-data.ts
```diff
@@ -0,0 +1,244 @@
...
+  type: 'full' | 'partial'
+}
+
+/**
+ * Creates segments from points of main series.
```
RobertJoonas:
```suggestion
 * Creates segments from points of a series.
```

Used for comparisons too.

assets/js/dashboard/stats/graph/main-graph-data.ts
```diff
@@ -0,0 +1,244 @@
...
+ * When a point of data is partial, all lines to and from it must be partial lines.
+ * (If that partial point moves, the lines to and from it move.)
+ * A full line is drawn only between two or more continuous full periods.
+ * No line is drawn from or to gaps in the data.
+ */
+export function getLineSegments(data: SeriesValue[]): LineSegment[] {
+  return data.reduce((segments: LineSegment[], curr, i) => {
```
RobertJoonas, line 159: Using `reduce` here makes it really difficult (at least for me) to understand this function. Had a long pause reading it. How about a simple `for i` loop?

I think it can also be simplified due to not needing to account for gaps appearing in the middle of the series anymore. Meaning that we can just break out of the loop on the first encounter of a `!isDefined` datapoint.

assets/js/dashboard/components/graph.tsx
```diff
@@ -0,0 +1,698 @@
...
+  buildAxis: (
+    marginLeft: number,
+    chartAreaWidth: number
+  ) => d3.Selection<SVGGElement, unknown, null, undefined>
+  marginLeft: number
+  chartAreaWidth: number
+  width: number
+  marginRight: number
+  minClientX: number
+  maxAttempts?: number
```
RobertJoonas: I think this doesn't need to be an argument? It should always be a fixed value of 2, right? First try to render the y axis labels, and if it didn't fit, do it a second time with a correct width that fits all labels.

assets/js/dashboard/components/graph.tsx
```diff
@@ -0,0 +1,698 @@
...
+        )
+      )
+
+    svg
+      .on(
+        'pointermove',
+        (event) => {
+          if (typeof onPointerMove !== 'function') {
+            return
+          }
```
RobertJoonas: This appears to be redundant because GraphProps doesn't allow it to not be a function.

assets/js/dashboard/components/graph.tsx
```diff
@@ -0,0 +1,698 @@
...
+    .attr('offset', '0%')
+    .attr('stop-color', stopTop.color)
+    .attr('stop-opacity', stopTop.opacity)
+
+  grad
+    .append('stop')
+    .attr('offset', '100%')
+    .attr('stop-color', stopBottom.color)
+    .attr('stop-opacity', stopBottom.opacity)
+  return id
```
RobertJoonas: The return value is unused. Could make this function return `void`.

assets/js/dashboard/components/graph.tsx
```diff
@@ -0,0 +1,698 @@
...
+): Set<string> => {
+  const maxXTicks = Math.min(bucketCount, 8)
+  const result = new Set<string>()
+  new Array(maxXTicks).fill(null).forEach((_v, i) => {
+    const tickValues = scale.ticks(maxXTicks - i)
+    if (tickValues.every(isWholeNumber)) {
+      result.add(JSON.stringify(tickValues))
+    }
+  })
+  return result
```
RobertJoonas: Let's replace with a simple `for i` loop?

```js
for (let i = 0; i < maxXTicks; i++) {
  const tickValues = scale.ticks(maxXTicks - i)
  if (tickValues.every(isWholeNumber)) {
    result.add(JSON.stringify(tickValues))
  }
}
```

assets/js/dashboard/components/graph.tsx
apata: You're right! Fixed

assets/js/dashboard/stats/graph/main-graph-data.ts
apata: It is pretty difficult to read, but I'd prefer not to rewrite this. It's already a rewrite of a previous simple for i loop ;) It's nicely covered with tests, I'm pretty confident in it.

assets/js/dashboard/stats/graph/main-graph.tsx
apata: That's way better, thanks! Fixed

assets/js/dashboard/stats/graph/fetch-main-graph.ts
apata: Good shout! Updated all places

assets/js/dashboard/components/graph-tooltip.tsx
apata: Thanks for having a look on mobile as well! I agree that it could be much improved, something I hope to look into with Annotations.

It was quite difficult to get these critical interactions working: 

1. user can see tooltip, 
2. user can move finger to move tooltip, 
3. user can quick zoom to period, 
4. **user can pan vertically past the chart.** 

To get the last thing as native as possible, I had to make the event listener passive. 
Because the event listener is passive, the browser will do it's own thing as well, which for longer presses is "select something" or "magnifying glass".

assets/js/dashboard/components/graph.tsx
apata: That's way better, thanks! I went with cheeky subtractive loop though

assets/js/dashboard/stats/graph/main-graph-data.ts
apata: Thanks! I personally like the descriptive variable names because they help to verify these calculations:
```ts
type LineSegment = {
  startIndexInclusive: number
  stopIndexExclusive: number
  // ...
}
              const atOrOverStart =
                line.startIndexInclusive !== undefined
                  ? i >= line.startIndexInclusive
                  : true
              const beforeEnd =
                line.stopIndexExclusive !== undefined
                  ? i < line.stopIndexExclusive
                  : true
```
vs 
```ts
type LineSegment = {
  start: number
  end: number
  // ...
}
              const atOrOverStart =
                line.start !== undefined
                  ? i >= line.start
                  : true
              const beforeEnd =
                line.end !== undefined
                  ? i < line.end
                  : true
```

RobertJoonas (approved):

PR #6161, opened 2026-03-13
Fix limited view issue

apata:
### Changes

- Fixes issue with top stats not loading for shared links limited to segment

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/api/stats_controller.ex
+++ b/lib/plausible_web/controllers/api/stats_controller.ex
@@ -1671,6 +1671,10 @@ defmodule PlausibleWeb.Api.StatsController do
     filters
   end
 
+  defp get_filters_param(_params) do
+    nil
+  end
+
   defp ensure_expected_segment_filter_present(
          filters,
          expected_segment_id


lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -1674,7 +1689,7 @@ defmodule PlausibleWeb.Api.StatsController do
     end
   end
 
-  defp ensure_expected_segment_filter_present(_params, _expected_segment_id) do
+  defp ensure_expected_segment_filter_present(_filters, _expected_segment_id) do
```
zoldar, line 1692: It's strange dialyzer isn't complaining that this clause is not reachable 🤔

zoldar: Scratch that, there's still a guard.

RobertJoonas: If you get into the third function clause of `get_filters_param` then we do end up in this clause. There's the `is_list(filters)` guard on the first one.

PR #6170, opened 2026-03-17
Move subscription settings page content into a LiveView

sanne-san:
### Changes

Prepares the page to host the choose_plan modal as a LiveComponent in a follow-up PR. Pure refactor, no UX changes.
- Extract subscription page content into a new SubscriptionSettings LiveView
- Move data loading out of the controller and into the LiveView
- Embed the LiveView in the existing settings layout via live_render

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/subscription_settings.ex
+++ b/lib/plausible_web/live/subscription_settings.ex
@@ -46,9 +46,8 @@ defmodule PlausibleWeb.Live.SubscriptionSettings do
       |> assign(:team_member_usage, team_member_usage)
       |> assign(:notification_type, notification_type)
       |> assign(:total_pageview_usage_domain, total_pageview_usage_domain)
-      |> assign(:current_path, "/settings/billing/subscription")
 
-    {:ok, socket, layout: {PlausibleWeb.LayoutView, :settings}}
+    {:ok, socket}
   end
 
   def render(assigns) do
--- a/lib/plausible_web/live/settings_context.ex
+++ b/lib/plausible_web/live/settings_context.ex
@@ -1,0 +1,16 @@
+defmodule PlausibleWeb.Live.SettingsContext do
+  @moduledoc false
+
+  import Phoenix.LiveView
+  import Phoenix.Component
+
+  def on_mount(_arg, _params, _session, socket) do
+    socket = attach_hook(socket, :get_current_path, :handle_params, &get_current_path/3)
+    {:cont, socket, layout: {PlausibleWeb.LayoutView, :settings}}
+  end
+
+  defp get_current_path(_params, url, socket) do
+    %{path: current_path} = URI.parse(url)
+    {:cont, assign(socket, :current_path, current_path)}
+  end
+end
--- a/lib/plausible_web/router.ex
+++ b/lib/plausible_web/router.ex
@@ -501,8 +501,10 @@ defmodule PlausibleWeb.Router do
     post "/security/email", SettingsController, :update_email
     post "/security/password", SettingsController, :update_password
 
-    scope alias: Live, assigns: %{connect_live_socket: true} do
-      live "/billing/subscription", SubscriptionSettings, :subscription, as: :settings
+    live_session :settings, on_mount: PlausibleWeb.Live.SettingsContext do
+      scope alias: Live, assigns: %{connect_live_socket: true} do
+        live "/billing/subscription", SubscriptionSettings, :subscription, as: :settings
+      end
     end
 
     get "/billing/invoices", SettingsController, :redirect_invoices


lib/plausible_web/templates/settings/subscription.html.heex
```diff
@@ -1,242 +1 @@
...
-        theme="secondary"
-        class="text-yellow-600 dark:text-yellow-400"
-        mt?={false}
-      >
-        [DEV ONLY] Change status
-      </.button_link>
-    <% end %>
-  </div>
-<% end %>
+{live_render(@conn, PlausibleWeb.Live.SubscriptionSettings, id: "subscription-settings")}
```
ukutaht: Can render the liveview directly from `Router`. The controller action and `subscription.html` template can then be removed.

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -0,0 +1,317 @@
+defmodule PlausibleWeb.Live.SubscriptionSettings do
+  @moduledoc """
+  LiveView for the subscription settings page.
+  """
+  use PlausibleWeb, :live_view
+  use Plausible
+
+  import PlausibleWeb.SettingsView,
```
ukutaht: View helpers are considered legacy since Phoenix added function components.I think it makes sense to modernize these helpers as well as part of this PR.

Either by inlining these helpers to this LiveView module or creating a separate components module with helpers.

lib/plausible_web/live/components/form.ex
```diff
@@ -428,7 +428,7 @@ defmodule PlausibleWeb.Live.Components.Form do
     end)
   end
 
-  attr :conn, Plug.Conn, required: true
+  attr :conn, :any, default: nil
```
zoldar:
```suggestion
  attr :conn, :map, default: %{}
```
?

lib/plausible_web/live/components/form.ex
```diff
@@ -439,7 +439,7 @@ defmodule PlausibleWeb.Live.Components.Form do
     assigns = assign(assigns, :options, flatten_options(options))
 
     ~H"""
-    <.form for={@conn} class="lg:hidden py-4" data-testid="mobile-nav-dropdown">
+    <.form for={@conn || %{}} class="lg:hidden py-4" data-testid="mobile-nav-dropdown">
```
zoldar:
```suggestion
    <.form for={@conn} class="lg:hidden py-4" data-testid="mobile-nav-dropdown">
```

given attribute has a default.

lib/plausible_web/components/billing/billing.ex
```diff
@@ -447,4 +447,61 @@ defmodule PlausibleWeb.Components.Billing do
         """
     end
   end
+
+  def present_plan_name(%Plausible.Billing.Plan{kind: kind}),
```
zoldar: As these functions aren't really components, perhaps we could move them under `PlausibleWeb.Components.Billing.Helpers` ?

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -0,0 +1,318 @@
...
+    <% end %>
+    """
+  end
+
+  on_ee do
+    defp consolidated_view_domain(team) do
+      case Plausible.ConsolidatedView.get(team) do
+        nil -> nil
+        view -> if Plausible.ConsolidatedView.ok_to_display?(team), do: view.domain
+      end
```
zoldar: I think that might be a bit more comprehensible:

```suggestion
      view = Plausible.ConsolidatedView.get(team)
      
      if not is_nil(view) and Plausible.ConsolidatedView.ok_to_display?(team) do
        view.domain
      end
```

lib/plausible_web/templates/layout/settings.html.heex
```diff
@@ -1,4 +1,6 @@
-<% options = account_settings_sidebar(@conn) %>
+<% conn = assigns[:conn] %>
+<% current_path = if conn, do: Path.join(conn.path_info), else: @current_path %>
+<% options = account_settings_sidebar(conn || assigns) %>
```
zoldar:
```suggestion
<% options = account_settings_sidebar(assigns) %>
```

should be fine for both I think. Then there's no need for separate clauses in `account_settings_sidebar`

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -0,0 +1,318 @@
...
+      |> assign(:invoices, invoices)
+      |> assign(:pageview_limit, Teams.Billing.monthly_pageview_limit(subscription))
+      |> assign(:pageview_usage, pageview_usage)
+      |> assign(:site_usage, site_usage)
+      |> assign(:site_limit, Teams.Billing.site_limit(team))
+      |> assign(:team_member_limit, Teams.Billing.team_member_limit(team))
+      |> assign(:team_member_usage, team_member_usage)
+      |> assign(:notification_type, notification_type)
+      |> assign(:total_pageview_usage_domain, total_pageview_usage_domain)
+      |> assign(:current_path, "settings/billing/subscription")
```
zoldar: I'd consider introducing a simple Plug which adds `current_path` under assigns and run it for all settings routes where sidebar is used.

Something like:

```elixir
defmodule PlausibleWeb.Plugs.CurrentPath do
  def init(_) do
    []
  end

  def call(conn, _) do
    assign(conn, :current_path, conn.path_info)
  end
end
```

and then in the router:

```elixir
  scope "/settings", PlausibleWeb do
    pipe_through [:browser, :csrf, PlausibleWeb.RequireAccountPlug, PlausibleWeb.Plugs.CurrentPath]
    ...
  end
```

With that, we can avoid all the branching in the layout template.

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -0,0 +1,318 @@
+defmodule PlausibleWeb.Live.SubscriptionSettings do
+  @moduledoc """
+  LiveView for the subscription settings page.
+  """
+  use PlausibleWeb, :live_view
+  use Plausible
+
+  import PlausibleWeb.Components.Billing,
```
zoldar: Given these will be moved to a separate `Helpers` module, `only` could be skipped.

lib/plausible_web/live/components/form.ex
zoldar: On a second thought, we don't need `conn` here anyway as we determine value using `selected_fn`, we can simply pass an empty `map` here and remove `settings` layout's reliance on `conn` entirely (provided `CurrentPath` plug is used, mentioned elsewhere).

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -0,0 +1,310 @@
...
+      |> assign(:invoices, invoices)
+      |> assign(:pageview_limit, Teams.Billing.monthly_pageview_limit(subscription))
+      |> assign(:pageview_usage, pageview_usage)
+      |> assign(:site_usage, site_usage)
+      |> assign(:site_limit, Teams.Billing.site_limit(team))
+      |> assign(:team_member_limit, Teams.Billing.team_member_limit(team))
+      |> assign(:team_member_usage, team_member_usage)
+      |> assign(:notification_type, notification_type)
+      |> assign(:total_pageview_usage_domain, total_pageview_usage_domain)
+      |> assign(:current_path, "settings/billing/subscription")
```
zoldar: This assign should be no longer needed - plug should provide it in the context already 😄

lib/plausible_web/plugs/current_path.ex
```diff
@@ -0,0 +1,6 @@
+defmodule PlausibleWeb.Plugs.CurrentPath do
+  @moduledoc false
+
+  def init(_), do: []
+  def call(conn, _), do: Plug.Conn.assign(conn, :current_path, Path.join(conn.path_info))
```
zoldar: Is `Path.join` used here for some specific reason? 🤔

lib/plausible_web/views/layout_view.ex
```diff
@@ -95,9 +95,17 @@ defmodule PlausibleWeb.LayoutView do
...
-  def account_settings_sidebar(conn) do
-    current_team = conn.assigns[:current_team]
-    current_team_role = conn.assigns[:current_team_role]
+  def account_settings_sidebar(%Plug.Conn{} = conn) do
+    account_settings_sidebar(%{
+      current_team: conn.assigns[:current_team],
+      current_team_role: conn.assigns[:current_team_role],
+      current_user: conn.assigns[:current_user]
+    })
+  end
```
zoldar: I think we don't need that clause anymore

lib/plausible_web/views/layout_view.ex
```diff
@@ -110,7 +118,7 @@ defmodule PlausibleWeb.LayoutView do
           if(not Teams.setup?(current_team),
             do: %{key: "API keys", value: "api-keys", icon: :api_keys}
           ),
-          if(Plausible.Users.type(conn.assigns.current_user) == :standard,
+          if(Plausible.Users.type(assigns[:current_user]) == :standard,
```
zoldar: I think it's safe to assume that user account should be always available in the context of rendering account settings sidebar:

```
assigns.current_user
```

lib/plausible_web/live/subscription_settings.ex
sanne-san: Tried removing it but it breaks; the plug runs on the HTTP request, but the LiveView uses a WebSocket which doesn't inherit it.

lib/plausible_web/plugs/current_path.ex
sanne-san: No good reason, fixed!

zoldar: Ah, I forgot `path_info` is a list, good call using `request_path` instead 😅

lib/plausible_web/live/subscription_settings.ex
zoldar: Okay, my wrong, for some reason I started thinking assigns from conn carry over to LV, which obviously is not the case argh. There will be more LVs, for each setting screen, I presume. We can still make something executed commonly for all LVs though. There's a mechanism for that, similar to plugs but specifically for LV: https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#on_mount/1

Ideally, we'd wrap this and all future LV routes for settings in something like this:

```elixir
  live_session :settings, on_mount: PlausibleWeb.Live.SettingsContext do
    scope alias: Live, assigns: %{connect_live_socket: true} do
      live "/billing/subscription", SubscriptionSettings, :subscription, as: :settings
    end
  end
```

https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.Router.html#live_session/3 brings a number of additional benefits:

- `settings` layout is consistently applied, so we can skip setting it in every LV's `mount` (more on that below)
- navigating between settings LVs can be potentially made more seamless by using `navigate` redirects which avoid full page reloads

I'm not sure if scope alias will work in this case - you might need to provide a full module path in `live` path definition, like `PlausibleWeb.Live.SubscriptionSettings`, but maybe not.

As for the context hook module, it's unfortunately tad more hairy than plug because we have to extract URI and need a [lifecycle hook](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#attach_hook/4) for that:

```elixir
defmodule PlausibleWeb.Live.SettingsContext do
  import Phoenix.LiveView

  def on_mount(_arg, _params, _session, socket) do
    socket = attach_hook(socket, :get_current_path, :handle_params, &get_current_path/3)
    {:cont, socket, layout: {PlausibleWeb.Layouts, :settings}}
  end

  defp get_current_path(_params, url, socket) do
    %{path: request_path} = URI.parse(url)

    {:cont, assign(socket, :current_path, current_path}
  end
end
```

The above handles both, setting the layout and setting the `current_path` and potentially other assigns that might be shared in the context of settings LVs.

_NOTE:_ I have not tested that code. If you stumble upon any issues implementing it and get stuck, please shout and we can try to sort this out.

lib/plausible_web/views/layout_view.ex
```diff
@@ -95,9 +95,9 @@ defmodule PlausibleWeb.LayoutView do
     |> Enum.reject(&is_nil/1)
   end
 
-  def account_settings_sidebar(conn) do
-    current_team = conn.assigns[:current_team]
-    current_team_role = conn.assigns[:current_team_role]
+  def account_settings_sidebar(assigns) do
+    current_team = assigns.current_team
+    current_team_role = assigns.current_team_role
```
zoldar, line 100: These assigns are indeed always set if `AuthPlug` is present in the pipeline - we just need to keep in mind that they both might be `nil`. But we should be good here 👍

zoldar (approved):
Great work 🙌

sanne-san:
@zoldar thanks for all the improvements!

PR #6171, opened 2026-03-17
Exclude revenue metrics when goal filter doesn't match a revenue goal

ukutaht:
### Changes

Attempted fix for https://sentry.plausible.io/organizations/plausible/issues/2145/?environment=prod&project=2&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=user

It seems that it's somehow possible to get the UI to a state where the main graph is queried with `metrics=total_revenue&filters=[]`. The total_revenue metric should only be requested with a revenue goal. This causes a crash in the backend (it probably shouldn't crash on the backend either) but I wanted to address the reason it happens in the first place.

To be honest I wasn't able to exactly re-produce the Sentry error. I was able to trigger the same crash by filtering by a revenue goal, choosing the total_revenue for main graph, and then changing the URL manually to a different non-revenue goal. But this triggers the main graph request with the total_revenue metric with the new goal in filters that I switched to. The Sentry error has `filters=[]`, I'm not sure how to get to that state.

Regardless, this does fix a related issue and I have hopes it will also fix the original problem.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/fetch-top-stats.test.ts
+++ b/assets/js/dashboard/stats/graph/fetch-top-stats.test.ts
@@ -10,7 +10,10 @@ import { remapToApiFilters } from '../../util/filters'
 import { chooseMetrics, MetricDef, topStatsQueries } from './fetch-top-stats'
 
 const aGoalFilter = ['is', 'goal', ['any goal']] as Filter
-const aRevenueGoal = { display_name: 'a revenue goal', currency: 'USD' } as const
+const aRevenueGoal = {
+  display_name: 'a revenue goal',
+  currency: 'USD'
+} as const
 const aRevenueGoalFilter = ['is', 'goal', ['a revenue goal']] as Filter
 const aPageFilter = ['is', 'page', ['/any/page']] as Filter
 const aPeriodNotRealtime = DashboardPeriod['28d']
--- a/assets/js/dashboard/stats/graph/fetch-top-stats.ts
+++ b/assets/js/dashboard/stats/graph/fetch-top-stats.ts
@@ -62,13 +62,12 @@ export function chooseMetrics(
   site: Pick<PlausibleSite, 'revenueGoals'>,
   dashboardState: DashboardState
 ): MetricDef[] {
-  const revenueMetrics: MetricDef[] =
-    revenueAvailable(dashboardState, site)
-      ? [
-          { key: 'total_revenue', label: 'Total revenue' },
-          { key: 'average_revenue', label: 'Average revenue' }
-        ]
-      : []
+  const revenueMetrics: MetricDef[] = revenueAvailable(dashboardState, site)
+    ? [
+        { key: 'total_revenue', label: 'Total revenue' },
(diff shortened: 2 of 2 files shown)


RobertJoonas (commented):
Thanks for this @ukutaht!

I'm pretty sure this surfaced with https://github.com/plausible/analytics/pull/6119. Here's how it happens:

* apply a revenue goal filter
* select average_revenue as the graph metric
* remove filters

The outcome of these three steps is not a crash, but just a failed request in the network tab with a 500 response. #6119 made a change that started fetching the graph and top stats at the same time, using the stored metric for the graph. The goal was to save time when possible. If the main graph request errors, it will be refetched after Top Stats is loaded and the graph metric validated.

Before that PR, it was just fetch_top_stats -> validate_metric -> fetch_graph, which made sure that revenue metrics were never queried without a goal filter.

About the change in this PR: although my initial thinking was that we can just request revenue metrics with *any goal filter*, and let the backend drop them (we're dropping them anyway when mixed currencies are used), this is correct by all means. To fix the Sentry error though, we should probably add a backend validation in the legacy timeseries module.

EDIT: A validation would fit better in `Api.StatsController.main_graph`. Note that with the switch to the new API, it will resolve by itself.

ukutaht:
@RobertJoonas Thanks for explaining!

> EDIT: A validation would fit better in Api.StatsController.main_graph. Note that with the switch to the new API, it will resolve by itself.

Based on this I'll just close the PR.

ukutaht:
Is it not possible to request the main graph with the correct metric in the first place? I noticed that when I get to the state you described:

apply a revenue goal filter
select average_revenue as the graph metric
remove filters

It not only fires the wrong request the first time, but also on every subsequent refresh of the page.

RobertJoonas:
> Is it not possible to request the main graph with the correct metric in the first place? I noticed that when I get to the state you described.

We can definitely improve this.

Normally, every metric that appears in Top Stats, can also be clicked on to display it on the graph. Now that the frontend is actually the one deciding the metrics, we could check if the `topStatsMetrics` that we have chosen based on the `dashboardState` include the `storedMetric`. If not, fall back to default -- `visitors`, which is available no matter the `dashboardState`.

There are currently two cases where backend *drops* metrics:

1. time_on_page is dropped when imported data is included with legacy time on page in the queried range
2. revenue metrics are dropped when multiple currencies are mixed.

In these cases, we should expect the initial graph request to fail. However, these cases are not common at all, so the behaviour of just refetching after top stats have arrived sounds like a pretty good solution to improve the UX.

It didn't occur to me before, but if we go this way and use top stats to validate the selection of the initial load graph metric, then the change you made in this PR is actually required -- we don't want `average_revenue` to be a valid graph metric with a non-revenue goal filter.

> It not only fires the wrong request the first time, but also on every subsequent refresh of the page.

That's not good 😄 We should make sure the stored metric is updated in that case.

PR #6158, opened 2026-03-11
Fix npm package resolution by adding exports field to package.json

alexgleason:
### Changes

Add `exports` field to the npm package's `package.json`. The package currently only declares `module` (no `main` or `exports`), which causes modern bundlers like Vite and SvelteKit to fail with:

> Failed to resolve entry for package "@plausible-analytics/tracker". The package may have incorrect main/module/exports specified in its package.json.

The `module` field is kept for backward compatibility with older bundlers.

Fixes #5879
See also: https://publint.dev/@plausible-analytics/tracker@0.4.4

### Tests
- [x] This PR does not require tests

### Changelog
- [x] Entry has been added to changelog

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #6174, opened 2026-03-18
[Tracker] Use ResizeObserver instead of polling for document height

### LINKED ISSUE #5727: Tracking script causes forced reflow of the page
### Past Issues Searched

### Issue is a Bug Report

### Describe the bug

I noticed the following performance issue in Google PageSpeed insights for my site: https://pagespeed.web.dev/analysis/https-bonforfait-fr-forfait-mobile/cb3p4e98kp?utm_source=search_console&form_factor=mobile&hl=en

Image

This is caused by the the Plausible tracking script: 

https://github.com/plausible/analytics/blob/1531386b767c6827216ab63d87fa447c46d37bf7/tracker/src/engagement.js#L129-L136

This code is used in the `getDocumentHeight` function to track scrolling position.

### Expected behavior

I'd like to avoid performance this penalty.

Maybe it is possible to wrap the `getDocumentHeight` function in a `requestAnimationFrame`? And to cache its value?

An alternative would be having the option to disable scroll tracking.

### Screenshots

_No response_

### Environment

```markdown

```

ukutaht:
### Changes

Replaces document height polling with `ResizeObserver` to avoid forced reflows.

Fixes https://github.com/plausible/analytics/issues/5727

`ResizeObserver` is a fairly modern API. Should be OK for the main script since overall it doesn't have worse support than fetch-with-keepalive. The only gap I was able to find is Edge versions less than 79 do not support `ResizeObserver` while fetch support goes back longer on Edge. So we will lose pre-2020 Edge versions but they have negligible usage worldwide. It would be good to have some kind of feature matrix for tracker script and check against caniuse data when we update it.

Instead of keeping the old code as fallback for `compat` code, I've disabled engagement tracking completely in `compat` mode. Without fetch-with-keepalive engagement data isn't reliable anyways. But it should probably be checked with sites that actually use the `compat` mode.

Disabling engagements in `compat` mode also allowed me to remove some fallbacks in `getCurrentScrollDepthPx` function which are not relevant in modern browsers (window.scrollY and window.innerHeight are way better supported than fetch and ResizeObserver, the fallbacks were only relevant for IE).

- [x] All existing tests pass
- [x] Add changelog entry
- [ ] Post-merge: release NPM tracker. Include this improvement as well: https://github.com/plausible/analytics/pull/6158

### Verification

On a local page with plausible installed, I profiled a scripted journey where page is loaded and scrolled after waiting a half a second. The old code produced 27 document reflows, with this PR there were 3 reflows. None of the remaining 3 were triggered by our script, they were all internal chrome events.

### DIFF
--- a/tracker/src/engagement.js
+++ b/tracker/src/engagement.js
@@ -151,23 +151,11 @@ export function init() {
   currentDocumentHeight = getDocumentHeight()
   maxScrollDepthPx = getCurrentScrollDepthPx()
 
-  window.addEventListener('load', function () {
+  new ResizeObserver(function () {
     currentDocumentHeight = getDocumentHeight()
-
-    // Update the document height again after every 200ms during the
-    // next 3 seconds. This makes sure dynamically loaded content is
-    // also accounted for.
-    var count = 0
-    var interval = setInterval(function () {
-      currentDocumentHeight = getDocumentHeight()
-      if (++count === 15) {
-        clearInterval(interval)
-      }
-    }, 200)
-  })
+  }).observe(document.documentElement)
 
   document.addEventListener('scroll', function () {
-    currentDocumentHeight = getDocumentHeight()
     var currentScrollDepthPx = getCurrentScrollDepthPx()
 
     if (currentScrollDepthPx > maxScrollDepthPx) {
--- a/tracker/src/engagement.js
+++ b/tracker/src/engagement.js
@@ -137,10 +137,8 @@ function getDocumentHeight() {
 }
 
 function getCurrentScrollDepthPx() {
-  var body = document.body || {}
-  var el = document.documentElement || {}
-  var viewportHeight = window.innerHeight || el.clientHeight || 0
-  var scrollTop = window.scrollY || el.scrollTop || body.scrollTop || 0
+  var viewportHeight = window.innerHeight
+  var scrollTop = window.scrollY
 
   return currentDocumentHeight <= viewportHeight
     ? currentDocumentHeight
--- a/tracker/src/engagement.js
+++ b/tracker/src/engagement.js
@@ -137,6 +137,16 @@ function getDocumentHeight() {
 }
 
 function getCurrentScrollDepthPx() {
+  if (COMPILE_COMPAT) {
+    var el = document.documentElement || {}
+    var body = document.body || {}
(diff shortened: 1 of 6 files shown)


tracker/src/engagement.js
```diff
@@ -151,23 +149,11 @@ export function init() {
...
-      currentDocumentHeight = getDocumentHeight()
-      if (++count === 15) {
-        clearInterval(interval)
-      }
-    }, 200)
-  })
+  }).observe(document.documentElement)
 
   document.addEventListener('scroll', function () {
-    currentDocumentHeight = getDocumentHeight()
```
ukutaht, line 170: No need to recompute document height on every scroll event. If the document height changes at any point, `ResizeObserver` will keep it updated.

tracker/src/engagement.js
```diff
@@ -151,23 +149,11 @@ export function init() {
...
 
-  window.addEventListener('load', function () {
+  new ResizeObserver(function () {
     currentDocumentHeight = getDocumentHeight()
-
-    // Update the document height again after every 200ms during the
-    // next 3 seconds. This makes sure dynamically loaded content is
-    // also accounted for.
-    var count = 0
-    var interval = setInterval(function () {
```
ukutaht, line 161: No need to poll the document for height changes during initial load. Any time DOM is loaded and updated, `ResizeObserver` will keep `currentDocumentHeight` variable in sync.

tracker/src/engagement.js
```diff
@@ -137,10 +137,8 @@ function getDocumentHeight() {
 }
 
 function getCurrentScrollDepthPx() {
-  var body = document.body || {}
-  var el = document.documentElement || {}
-  var viewportHeight = window.innerHeight || el.clientHeight || 0
```
ukutaht, line 142: By removing IE support for engagement tracking, these fallbacks are not necessary since `window.scrollY` and `window.innerHeight` are well-supported:

https://caniuse.com/mdn-api_window_innerheight
https://caniuse.com/mdn-api_window_scrolly

apata: suggestion, non-blocking: Can we keep 0 fallback? It doesn't cost much weight-wise and then the math won't crash entirely if for some reason these values are not truthy. Positive numbers are truthy.

apata (approved):

PR #6176, opened 2026-03-19
Get rid of old /top-stats endpoint

RobertJoonas:
### Changes

Could have been done sooner (with the actual endpoint switch) but I mistakenly thought it's needed for Dashboard CSV export. Which it's not.

The `top_stats_test.exs` probably gives the query API v2 a good amount of extra coverage. This PR modernizes that file to keep that coverage:

* Query the new endpoint instead, where applicable.
* Tests such as those asserting on the metric selection based on query state, or a returned `graph_metric` key were removed
* Instead of asserting on `with_imported_switch_info`, we now assert on `query.skip_imports_reason` and `query.imports_included` in all the same scenarios. 
* Some more basic tests, such as "counts unique visitors" or "counts pageviews" were also removed.

Rewriting the tests uncovered two issues:

1. [unrelated to top stats migration] - when scroll depth is queried alone, and with imported data included, the native query aggregation doesn't work (out of scope here, added a ticket on basecamp)
2. time on page is returned as `nil` (instead of 0 as before) when it cannot be calculated. For users it means that Top Stats will simply hide the metric. I've updated the test to assert on `nil`.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/authorization_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/authorization_test.exs
@@ -2,7 +2,10 @@ defmodule PlausibleWeb.Api.StatsController.AuthorizationTest do
   use PlausibleWeb.ConnCase, async: true
 
   defp query_visitors(conn, site, params) do
-    params = Keyword.merge([date_range: "day", metrics: ["visitors"]], params)
+    params =
+      [date_range: "day", metrics: ["visitors"], filters: []]
+      |> Keyword.merge(params)
+
     post(conn, Routes.stats_path(conn, :query, site.domain), params)
   end
 


test/plausible_web/controllers/api/stats_controller/authorization_test.exs
```diff
@@ -1,6 +1,16 @@
...
   use PlausibleWeb.ConnCase, async: true
 
+  defp call_internal_query_endpoint(conn, site, q, opts \\ []) do
+    params = %{
+      "date_range" => "day",
+      "metrics" => ["visitors"],
+      "filters" => Keyword.get(opts, :filters, [])
+    }
+
+    post(conn, "/api/stats/#{site.domain}/query?#{q}", params)
```
aerosol: It'd be better to build the path using `stats_path` helper.

Technically there's no difference between `q` and `opts`, so we could simplify the signature even:

```elixir
  defp call_internal_query_endpoint(conn, site, params) do
    params =
      Keyword.merge(
        [
          date_range: "day",
          metrics: ["visitors"]
        ],
        params
      )

    post(conn, Routes.stats_path(conn, :query, site.domain), params)
  end
```

and use it without interpolation:

```elixir
    conn =
      call_internal_query_endpoint(conn, site,
        auth: link.slug,
        filters: [["is", "segment", [segment.id + 1]]]
      )
```

aerosol (approved):
One comment inline if you don't mind @RobertJoonas

test/plausible_web/controllers/api/stats_controller/authorization_test.exs
aerosol: (renaming it to `query_visitors` wouldn't hurt either)

RobertJoonas: https://github.com/plausible/analytics/pull/6176/commits/9d44835489cabeac1285f686557ed8ec6cc349d1

PR #6177, opened 2026-03-19
Update ua_inspector

cnkk:


zoldar:
Perhaps worth to consider bumping it together with elixir library? https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/7654102067 for reference

aerosol:
@cnkk https://github.com/plausible/analytics/pull/6183

PR #6184, opened 2026-03-23
Add ability to skip site members and invitations during transfer

zoldar:
### Changes

This PR adds an option to accept site transfer without transferring any associated guest memberships and invitations. The option is suggested only if accepting user's plan cannot accommodate additional team members from the transferred site. The main purpose of this feature is to ease transferring the sites in cases where the site is carried over to somebody on a single-member starter plan.

image

image

Following refactorings were done along the way:

- `Billing.Quota.ensure_within_plan_limits` accept `skip_*` options in a consistent format for all limits (as all of them are used in various contexts.
- `Invitations.accept_transfer_ownership` was refactored to reduce complexity and improve clarity of the logic, especially with added branching

### Tests

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/components/generic.ex
+++ b/lib/plausible_web/components/generic.ex
@@ -252,7 +252,7 @@ defmodule PlausibleWeb.Components.Generic do
               </div>
             </div>
           </div>
-          <div :if={@actions != []} class="relative shrink-0 flex gap-2 z-10">
+          <div :if={@actions != []} class="shrink-0 flex gap-2">
             {render_slot(@actions)}
           </div>
         </div>
--- a/lib/plausible_web/components/prima_dropdown.ex
+++ b/lib/plausible_web/components/prima_dropdown.ex
@@ -59,7 +59,7 @@ defmodule PlausibleWeb.Components.PrimaDropdown do
       id={@id}
       placement="bottom-end"
       match_trigger_width={true}
-      class="bg-white rounded-md shadow-lg ring-1 ring-black/5 focus:outline-none p-1.5 dark:bg-gray-800"
+      class="bg-white rounded-md shadow-lg ring-1 ring-black/5 focus:outline-none p-1.5 dark:bg-gray-800 relative z-10"
     >
       {render_slot(@inner_block)}
     </Dropdown.dropdown_menu>


zoldar:
@plausible/developers Bump

test/plausible_web/controllers/invitation_controller_test.exs
```diff
@@ -186,6 +186,37 @@ defmodule PlausibleWeb.Site.InvitationControllerTest do
       assert Phoenix.Flash.get(conn.assigns.flash, :error) =~
                "Plan limits exceeded: site limit."
     end
+
+    test "succeeds despite exceedeing members limit if parameter provided", %{
```
ukutaht:
```suggestion
    test "succeeds despite exceeding members limit if parameter provided", %{
```

lib/plausible_web/components/generic.ex
```diff
@@ -251,7 +252,7 @@ defmodule PlausibleWeb.Components.Generic do
               </div>
             </div>
           </div>
-          <div :if={@actions != []} class="shrink-0 flex gap-2">
+          <div :if={@actions != []} class="relative shrink-0 flex gap-2 z-10">
```
ukutaht: What does `relative` do here? I'm not seeing any effect at the moment.

As for `z-10` I think setting the z-index should go on the dropdown menu directly instead of the notice component.

Currently PrimaDropdown [sets](https://github.com/plausible/analytics/blob/7a8b57423330a4578bbbf79654cc0bd428b18358/lib/plausible_web/components/prima_dropdown.ex#L81) `z-50` for dropdown items. I think this could be moved to `<.dropdown_menu` to ensure the menus appear above the rest of the page.

Longer term I want to use portals for dropdowns so z-index is not needed

zoldar: image

Good spot 😅

zoldar: Image

Fixed in https://github.com/plausible/analytics/pull/6184/commits/cd44fde3e8a7e4d09c0b7be9c53beae93828f20b and checked across Safari, Firefox and Vivaldi on MacOS.

`relative` is necessary because `z-index` only works for positioned elements, as described in https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/z-index

PR #6187, opened 2026-03-23
Bugfix: adjustments to querying time:minutes and time:hours dimensions for goal conversion

aerosol:
### Changes

Disclaimer: my confidence about this solution is close to zero :sweat_smile:   

https://sentry.plausible.io/organizations/plausible/issues/1101/?environment=prod&project=2&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream - when querying timeseries with 
 ```time:minute` interval and a goal filter, the app would crash with `** (Ecto.QueryError) lib/plausible/stats/sql/special_metrics.ex:114: field `visitors` does not exist in subquery in query```. 

For `time:minute`, session visitor counts are split into a separate "smeared" subquery. The conversion rate calculation then tries to read visitors from the events subquery - but visitors is no longer there, because table decider moves it to another query. The fix is to skip that split entirely when `group_conversion_rate` is in the query. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Fixed main graph crash for time:minute dimension and goal conversion filter
 - Fixed Stats API timeseries returning time buckets falling outside the queried range
 - Fixed issue with all non-interactive events being counted as interactive
 - Fixed countries map countries staying highlighted on Chrome


lib/plausible/stats/table_decider.ex
```diff
@@ -110,9 +110,15 @@ defmodule Plausible.Stats.TableDecider do
...
   #   See `time_slots` usage in `Plausible.Stats.SQL.Expression` to understand how this is done.
+  #
+  # Smearing must be suppressed when :group_conversion_rate is in the query
+  # or when query.bypass_session_smearing? is true (used by group_conversion_rate
+  # totals subquery).
   @smearable_metrics [:visitors, :visits]
   defp smear_session_metrics({:sessions, metrics} = value, query) do
-    if "time:minute" in query.dimensions or "time:hour" in query.dimensions do
+    if ("time:minute" in query.dimensions or "time:hour" in query.dimensions) and
+         :group_conversion_rate not in query.metrics and
```
zoldar, line 120: Given that `bypass_session_smearing?` is set to `true` on adding `group_conversion_rate` metric, is this condition necessary? 🤔

aerosol:
To be fleshed out further, I have no idea what I'm doing and @ukutaht is pointing at different direction. Closing

PR #6188, opened 2026-03-23
Export BEAM processes to OTLP metrics endpoint

cnkk:
Vibed dashboard: https://plausible.grafana.net/d/beam-process-metrics/beam-process-metrics?orgId=1&from=now-1h&to=now&timezone=browser&refresh=30s&var-interval_s=10&var-top_n=20&var-process=$__all&var-namespace=plausible-app-6188

about pinned deps `opentelemetry_*_experimental`:

> the OTel metrics API isn't published yet
> override: true is required because opentelemetry_exporter transitively depends on the 0.5.x Hex stubs, which don't contain the metrics modules (otel_meter, otel_exporter_metrics_otlp, etc.).


cnkk:
@coderabbitai review

PR #6193, opened 2026-03-23
Test a different smearing fix

ukutaht:


ukutaht:
@aerosol when you're back, please give this a review

PR #6196, opened 2026-03-24
Start recording subscription changes in audit trail

zoldar:
### Changes

This PR instruments all spots where the subscription is either created or updated with an audit trail. The entries will appear under "Audit" tab in our CRM. They can also serve as a better basis to gather aggregate statistics of changes in subscriptions (upgrades, downgrades, etc.).

### Tests


zoldar:
We are not going to do it this way, because audit trail might then contain duplicate entries due to webhook retries. This in turn leads to unreliable data. It might be ok for aggregate, statistical queries but not for any sort of book keeping.

PR #6204, opened 2026-03-26
Render top referrers and search terms within the context of sources tabs

zoldar:
### Changes

This PR reorganizes sources breakdown tabs so that "Top referrers" and "Search terms" (Google Search keywords) are always rendered in place of "Sources" instead of replacing the breakdown tile completely with a single tab.

I took this opportunity to add e2e coverage for search terms breakdown using mocked API response (which got updated to reflect actual real payload by the way).

### Tests

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/sources/index.js
+++ b/assets/js/dashboard/stats/sources/index.js
@@ -289,15 +289,13 @@ export default function Sources() {
         : null
     }
 
-    switch (currentTab) {
-      case 'channels':
-        return {
-          path: channelsRoute.path,
-          search: (search) => search
-        }
-      case 'all':
-      default:
-        return sourceMoreLinkProps()
+    if (currentTab === 'channels') {
+      return {
+        path: channelsRoute.path,
+        search: (search) => search
+      }
+    } else {
+      return sourceMoreLinkProps()
     }
   }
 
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/sources/index.js
```diff
@@ -1,25 +1,381 @@
...
+      }[currentTab]
+      return route
+        ? {
+            path: route.path,
+            search: (search) => search
+          }
+        : null
+    }
+
+    switch (currentTab) {
```
RobertJoonas: Since there are only two different outcomes here, how about a regular if-else?

assets/js/dashboard/stats/sources/index.js
```diff
@@ -1,25 +1,381 @@
...
+      return <AllSources afterFetchData={afterFetchData} />
+    }
+  }
+
+  function renderContent() {
+    if (Object.keys(UTM_TAGS).includes(currentTab)) {
+      return <UTMSources tab={currentTab} afterFetchData={afterFetchData} />
+    }
+
+    switch (currentTab) {
```
RobertJoonas: Ditto about maybe preferring if-else.

zoldar: https://github.com/plausible/analytics/pull/6204/commits/f8074bc6ee2ed140bf282e2642f163c32713be07

assets/js/dashboard/stats/sources/index.js
zoldar: https://github.com/plausible/analytics/pull/6204/commits/f8074bc6ee2ed140bf282e2642f163c32713be07

PR #6205, opened 2026-03-26
Make comparisons to today consider only stats in the same exact time range

zoldar:
### Changes

This PR changes behavior when determining time range for "today" when enabling comparisons. Before it, the past time range compared against always included a full day (corrected by site's timezone). Now it only includes the exact time range from midnight relative to the current hour.

The new behavior can be tested live on preview app at https://pr-6205.review.plausible.io/

image

Technical stuff follows...

I have modified input date range parsing logic in both new and legacy query builders to always make an exception for "day" period when it matches today corrected for site's timezone. When the condition is met:

- the `last` timestamp of the datetime range is set to `query.now`
- the comparison time range is computed by shifting exact time range, like it's done for the `24h` period instead of generating a time range for the whole day period

Time related logic got refactored very slightly to reduce repetition and allow mixing dates and datetimes when building `DateTimeRange`.

### Tests

### Changelog

### DIFF
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -136,26 +136,19 @@ defmodule Plausible.Stats.Comparisons do
   defp get_comparison_datetime_range(
          %Query{
            input_date_range: input_range,
-           include: %{compare: :previous_period, compare_match_day_of_week: true}
+           include: %{compare: :previous_period} = include
          } = source_query
        )
        when input_range in [:"24h", :day] do
-    days_back = 7
-    comparison_start = DateTime.shift(source_query.utc_time_range.first, day: -days_back)
-    comparison_end = DateTime.shift(source_query.utc_time_range.last, day: -days_back)
+    offset =
+      if include.compare_match_day_of_week do
+        [day: -7]
+      else
+        [hour: -24]
+      end
 
-    DateTimeRange.new!(comparison_start, comparison_end)
-  end
-
-  defp get_comparison_datetime_range(
-         %Query{
-           input_date_range: input_range,
-           include: %{compare: :previous_period}
-         } = source_query
-       )
-       when input_range in [:"24h", :day] do
-    comparison_start = DateTime.shift(source_query.utc_time_range.first, hour: -24)
-    comparison_end = DateTime.shift(source_query.utc_time_range.last, hour: -24)
+    comparison_start = DateTime.shift(source_query.utc_time_range.first, offset)
+    comparison_end = DateTime.shift(source_query.utc_time_range.last, offset)
 
     DateTimeRange.new!(comparison_start, comparison_end)
   end


lib/plausible/stats/comparisons.ex
```diff
@@ -120,13 +132,14 @@ defmodule Plausible.Stats.Comparisons do
     end
   end
 
-  # For 24h period, shift the datetime range directly to preserve time precision
+  # For 24h and today periods, shift the datetime range directly to preserve time precision
   defp get_comparison_datetime_range(
          %Query{
-           input_date_range: :"24h",
+           input_date_range: input_range,
```
RobertJoonas, line 138: Nitpick suggestion to DRY the code a bit: we could merge two function clauses into one by evaluating a shift variable within the function:

```elixir
shift = if compare_match_day_of_week, do: [day: -7], else: [hour: -24]
```

lib/plausible/stats/comparisons.ex
```diff
@@ -120,13 +132,14 @@ defmodule Plausible.Stats.Comparisons do
     end
   end
 
-  # For 24h period, shift the datetime range directly to preserve time precision
+  # For 24h and today periods, shift the datetime range directly to preserve time precision
```
apata, line 135: question, non-blocking: Did you consider creating a new :today period and remapping :day to it when appropriate? Because "today" gets special treatment, it might lead to clearer code overall. Probably more code though :D

lib/plausible/stats/comparisons.ex
zoldar: https://github.com/plausible/analytics/pull/6205/commits/6d5313123ea0793b43e61067b65e5a3f7c6d8658

zoldar:
> This also makes me wonder if we could get rid of the trim_relative_date_range query option in a follow-up...

@RobertJoonas yeah, that has crossed my mind at one point but then I forgot about it 😅  Good observation. I'll add a follow-up task for that.

lib/plausible/stats/comparisons.ex
zoldar: I have not. However, looking at it now, I think adding this period would add a lot more complexity in other parts of logic working with the query. Also, we'd have to probably maintain support for the "old" way of defining period to support existing external links. Not worth it IMHO.

lib/plausible/stats/comparisons.ex
```diff
@@ -120,38 +132,34 @@ defmodule Plausible.Stats.Comparisons do
...
-
-    DateTimeRange.new!(comparison_start, comparison_end)
-  end
+       )
+       when input_range in [:"24h", :day] do
+    offset =
+      if include.compare_match_day_of_week do
+        [day: -7]
+      else
+        [hour: -24]
```
apata, line 147: question, non-blocking: does -24h work as expected during changeover from DST?

zoldar: I'm not sure what's expected, but AFAIK the conversion happens later on via `DateTimeRange.new!` https://github.com/plausible/analytics/blob/sources-referrers-google-preserve-other-tabs/lib/plausible/stats/datetime_range.ex#L23-L39. In case of gap or ambiguity (for instance, due to DST change), the lower bound always picks the time after, while the upper bound picks the time before the gap/DST change.

PR #6211, opened 2026-03-31
Allow revenue metrics to be queried with session dimensions

ukutaht:
### Changes

Allows revenue metrics to be queries with session dimensions. Prepares for refactoring entry page breakdown to APIv2 code.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/table_decider.ex
+++ b/lib/plausible/stats/table_decider.ex
@@ -14,15 +14,15 @@ defmodule Plausible.Stats.TableDecider do
   @revenue_metrics on_ee(do: Plausible.Stats.Goal.Revenue.revenue_metrics(), else: [])
 
   def events_join_sessions?(query) do
-    filter_session_dims =
+    session_dims_in_filters? =
       query.filters
       |> dimensions_used_in_filters()
       |> Enum.any?(&(dimension_partitioner(query, &1) == :session))
 
-    group_session_dims =
+    session_dims? =
       Enum.any?(query.dimensions, &(dimension_partitioner(query, &1) == :session))
 
-    filter_session_dims or group_session_dims
+    session_dims? or session_dims_in_filters?
   end
 
   def sessions_join_events?(query) do


lib/plausible/stats/table_decider.ex
```diff
@@ -4,15 +4,25 @@ defmodule Plausible.Stats.TableDecider do
...
   def events_join_sessions?(query) do
-    query.filters
-    |> dimensions_used_in_filters()
-    |> Enum.any?(&(dimension_partitioner(query, &1) == :session))
+    filter_session_dims =
+      query.filters
+      |> dimensions_used_in_filters()
+      |> Enum.any?(&(dimension_partitioner(query, &1) == :session))
+
+    group_session_dims =
```
ukutaht: This is needed not for the dashboard but for API. In the dashboard we always add a `[:is_not, "visit:entry_page", ""]` filter to queries so checking filters is enough. We can't expect that there's always a filter present in API queries.

lib/plausible/stats/table_decider.ex
```diff
@@ -4,15 +4,25 @@ defmodule Plausible.Stats.TableDecider do
...
 
   alias Plausible.Stats.{Query, QueryError}
 
+  @revenue_metrics on_ee(do: Plausible.Stats.Goal.Revenue.revenue_metrics(), else: [])
+
   def events_join_sessions?(query) do
-    query.filters
-    |> dimensions_used_in_filters()
-    |> Enum.any?(&(dimension_partitioner(query, &1) == :session))
+    filter_session_dims =
```
zoldar: `session_dims_in_filters?`

lib/plausible/stats/table_decider.ex
zoldar: `session_dims?`

zoldar: Would it make sense to extend the check the same way `sessions_join_events?` ?

lib/plausible/stats/table_decider.ex
ukutaht: https://github.com/plausible/analytics/pull/6211/changes/16eb2be646ba0635a954a78df8f6df30f04091e6

lib/plausible/stats/table_decider.ex
ukutaht: https://github.com/plausible/analytics/pull/6211/changes/16eb2be646ba0635a954a78df8f6df30f04091e6

ukutaht:
> Would it make sense to extend the check the same way sessions_join_events? ?

@zoldar no, the line on https://github.com/plausible/analytics/blob/16eb2be646ba0635a954a78df8f6df30f04091e6/lib/plausible/stats/table_decider.ex#L56 guards against session metrics being requested for event dimensions so adding the same extension to `sessions_join_events` would be a no-op.

I had to change `events_join_sessions?` because we now allow revenue metrics (limited set of event metrics) being requested with session dimensions like `entry_page` or `source`.

zoldar:
@ukutaht got it. If that would ever change and there would be similar exclusions the other way around, the logic would have to be altered anyway and doing that up front would only muddle it up.

PR #6218, opened 2026-04-02
Increase worker count for e2e tests

cnkk:


zoldar:
@cnkk just for clarity: you want to fire up 4 partitions, each with 3 workers instead of 1, right?

zoldar:
I wonder if we could further gain from reducing the number of pratitions and bumping the number of workers even more? The setup time is pretty much dominating the actual test run time now.

PR #6237, opened 2026-04-08
Strict order funnels

aerosol:
### Changes

This PR requires #6238 to be merged first.
It adds the ability to mark funnels strict, meaning that intervening steps will stop processing.

image

Copy and UI control per @sanne-san's suggestions

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible/funnels.ex
```diff
@@ -13,10 +13,12 @@ defmodule Plausible.Funnels do
 
   import Ecto.Query
 
-  @spec create(Plausible.Site.t(), String.t(), [map()]) ::
+  @spec create(Plausible.Site.t(), String.t(), [map()], boolean()) ::
           {:ok, Funnel.t()}
           | {:error, Ecto.Changeset.t() | :invalid_funnel_size | :upgrade_required}
-  def create(site, name, steps)
+  def create(site, name, steps, strict_order? \\ false)
```
zoldar: Perhaps makes sense to add it as an opt? 4 arguments seems like the usual threshold at which we usually do that.

extra/lib/plausible/funnels.ex
```diff
@@ -26,19 +28,24 @@ defmodule Plausible.Funnels do
...
   end
 
   @spec update(Funnel.t(), String.t(), [map()]) ::
           {:ok, Funnel.t()}
           | {:error, Ecto.Changeset.t() | :invalid_funnel_size | :upgrade_required}
-  def update(funnel, name, steps) do
+  @spec update(Funnel.t(), String.t(), [map()], boolean()) ::
+          {:ok, Funnel.t()}
+          | {:error, Ecto.Changeset.t() | :invalid_funnel_size | :upgrade_required}
+  def update(funnel, name, steps), do: update(funnel, name, steps, funnel.strict_order)
```
zoldar: Perhaps a base clause with a default + using opts would make more sense here? Like in the case of create.

extra/lib/plausible/funnels.ex
```diff
@@ -47,27 +54,31 @@ defmodule Plausible.Funnels do
...
-  def edit_changeset(funnel, name, steps) do
-    Funnel.changeset(funnel, %{name: name, steps: steps})
+  def edit_changeset(funnel, name, steps, strict_order? \\ false) do
+    Funnel.changeset(funnel, %{name: name, steps: steps, strict_order: strict_order?})
   end
 
-  @spec ephemeral_definition(Plausible.Site.t(), String.t(), [map()]) :: Funnel.t()
-  def ephemeral_definition(site, name, steps) do
+  @spec ephemeral_definition(Plausible.Site.t(), String.t(), [map()], boolean()) :: Funnel.t()
+  def ephemeral_definition(site, name, steps, strict_order? \\ false) do
```
zoldar: Ditto about opts

extra/lib/plausible_web/live/funnel_settings/form.ex
```diff
@@ -440,12 +468,13 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
   end
 
   defp prepare_socket(socket, site, nil) do
-    form = to_form(Funnels.create_changeset(site, "", []))
+    form = to_form(Funnels.create_changeset(site, "", [], false))
```
zoldar: Perhaps lift the default to a wrapped property in struct, exposed via `Funnel.default_strict_funnel()`? Or `@default_strict_funnel` here directly?

aerosol: This is a non-strict funnel in fact, but I'll just skip that and make the default handled within create_changeset's opts.

aerosol: e2d1d21574

extra/lib/plausible/funnels.ex
aerosol: e2d1d21574

extra/lib/plausible/funnels.ex
aerosol: e2d1d21574

extra/lib/plausible/funnels.ex
aerosol: e2d1d21574

extra/lib/plausible/funnels.ex
```diff
@@ -47,27 +49,37 @@ defmodule Plausible.Funnels do
 
       :ok ->
         funnel
-        |> Funnel.changeset(%{name: name, steps: steps})
+        |> edit_changeset(name, steps,
+          strict_order?: Keyword.get(opts, :strict_order?, !!funnel.strict_order)
```
zoldar, line 53: Provided we currently pass sort_order everywhere (?), perhaps better to user either `Keyword.fetch!` or default to a particular state like `false`? This kind of behavior being dug so deep might be super confusing when revisited by someone not familiar with the code (IMHO).

zoldar (approved):

extra/lib/plausible/funnels.ex
aerosol: We can't default to `false` because we're updating an existing funnel here and it's up to the user to either update many properties at once or just some. The fact that we're currently passing strict_order from within the UI always (mainly because the toggle thing isn't a proper form field), doesn't make it required IMHO. Out of scope tests calling `Funnels.update` should not be concerned with that option.

zoldar: JFC, don't mind me, turned a blind eye to double bang 🤦

PR #6242, opened 2026-04-09
Implement basics of funnel exploration logic along with UI prototype

zoldar:
### Changes

This PR introduces first iteration of user journey exploration. It implements forward-only exploration, largely in line with what the earlier PoC (https://github.com/plausible/analytics/pull/6163) did. The difference is that the querying logic is completely dedicated to exploration and is based on recent research we did with plain SQL prototypes.

One important addition is the ability to filter the next step, so exploration is not entirely dependent on only 10 most popular paths.

The UI is adapted from the earlier prototype. While the API provides ability to query for a full funnel with dropoff and dropoff percentage calculations, for the time being, the UI is only showing visitors, still sticking to incremental querying.

The exploration feature is entirely hidden from everyone but superadmins. For superadmins though, it's available for every site.

The main purpose of this PR is to get an early version out for experimentation against a larger number of various real-world cases - to verify he range of possible journeys across very different datasets as well as the performance.

### Tests


lib/plausible/stats/exploration.ex
```diff
@@ -0,0 +1,263 @@
...
+        field(s, ^:"pathname#{count}") == ^step.pathname
+    )
+  end
+
+  defp maybe_search(query, search_term) do
+    case String.trim(search_term) do
+      term when byte_size(term) > 2 ->
+        from(s in query,
+          where:
+            ilike(selected_as(:next_name), ^"%#{term}%") or
```
aerosol, line 209: Pretty cool that you can search for non-pageviews thanks to that

lib/plausible/stats/exploration.ex
```diff
@@ -0,0 +1,263 @@
...
+        if acc.visitors_at_previous, do: acc.visitors_at_previous - current_visitors, else: 0
+
+      dropoff_percentage = percentage(dropoff, acc.visitors_at_previous)
+
+      funnel = [
+        %{
+          step: step,
+          visitors: current_visitors,
+          dropoff: dropoff,
+          dropoff_percentage: dropoff_percentage
```
aerosol, line 234: Not necessary right now, but how hard would it be to include `conversion_rate`, `conversion_rate_step` and `label` here? That way we'd have something compatible with legacy funnels to display.

lib/plausible/stats/exploration.ex
```diff
@@ -0,0 +1,263 @@
...
+        | acc.funnel
+      ]
+
+      %{acc | funnel: funnel, visitors_at_previous: current_visitors}
+    end)
+    |> Map.fetch!(:funnel)
+    |> Enum.reverse()
+  end
+
+  defp percentage(x, y) when x in [0, nil] or y in [0, nil] do
```
aerosol: nitpick: make this a catch-all clause and guard for non-zero integers in the main one?

aerosol (approved):
Great work ✨ 

UI shortcomings are obvious so I won't comment on that and let @sanne-san iterate.

lib/plausible/stats/exploration.ex
zoldar: Yeah, we can do that for sure, it's just a couple more calculations in the reduction. I'll add a note for that.

zoldar: As for label, yeah, this is something worth adding too, to normalize the label on the backend already.

lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6242/commits/9ad25b00d0efb009c0d316548b7119060ad817ca

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -332,6 +335,54 @@ defmodule PlausibleWeb.Api.StatsController do
...
+          }
+        )
+    end
+  end
+
+  defp parse_journey(input) when is_binary(input) do
+    input
+    |> Jason.decode!()
+    |> Enum.map(&parse_journey_step/1)
+    |> Enum.reject(&is_nil/1)
```
ukutaht, line 375: This `Enum.reject` seems to be a no-op currently. `parse_journey_step` will always return a non-nil value or raise if pattern match fails.

lib/plausible/stats/exploration.ex
```diff
@@ -0,0 +1,244 @@
...
+  defp next_steps_query(query, steps, search_term) do
+    next_step_idx = length(steps) + 1
+    q_steps = steps_query(query, next_step_idx)
+
+    next_name = :"name#{next_step_idx}"
+    next_pathname = :"pathname#{next_step_idx}"
+
+    q_next =
+      from(s in subquery(q_steps),
+        # avoid cycling back to the beginning of the exploration
```
ukutaht, line 102: Why prevent cycling to the first step but allow cycling to step 2, step 3, etc?

ukutaht (commented):
Looks good overall 👍

lib/plausible_web/controllers/api/stats_controller.ex
zoldar: Sorry, my bad. Initially I've made it permissive, silently removing invalid steps and later on I've decided it better crash loudly (an explicit error state can be added later on, if necessary, however, for internal API purposes at least, the input should be expected to be well-formed).

I'll clean that up in one of the follow-ups.

lib/plausible/stats/exploration.ex
zoldar: At first I thought that'd follow the logic of `windowFunnel` with `allow_reentry=false`, which is the default. However, taking a closer look at it's definition, it seems to apply to a more specific case which doesn't even occur when building funnel using the current method.

Other than this, it "felt" wrong 😅 But other than that, I have no rational reason to no allow that. If somebody wants to explore cycles, no point in preventing that. I'll drop that condition in a follow-up.

PR #6245, opened 2026-04-13
Improve top bar and main graph stats UI

sanne-san:
### Changes

**Top bar**
- Add consistent styling and spacing for all buttons in the top bar
- Move period arrows into the period picker visually instead of displaying them as separate buttons
- Add selected state to buttons; when dropdown is open, the trigger background is light gray. This is also true when comparison is enabled.
- Change filter button icon from magnifying glass to a filter icon
- Change period picker button icons from chevron on the right to calendar icons on the left
- Make top bar scroll horizontally on mobile

**New options menu (⋮)**
- Move interval picker, export button, imported data toggle and notices out of the graph and into a new options menu in the top bar
- Hide menu in realtime mode
- Render button immediately as a non-interactive placeholder on page load to prevent layout shift
- Change interval picker dropdown to a segmented control
- Change the imported data toggle to an actual toggle button
- Change notice icon + tooltip to an inline notice component
- The `i` keyboard shortcut now cycles through graph intervals rather than opening the dropdown menu – also when the options menu is closed

**Top stats**
- Change selected metric background highlight to a subtle background instead of a coloured underline on the label
- Refine stat labels and values typography and colors
- Ensure change arrows and percentages are shown also when in comparison mode (and removed from tooltip)
- Show "until HH:MM" when "Today" is the comparison date range to indicate a partial day

**Code organisation**
- Introduce a shared React context (DashboardOptionsProvider) to pass graph state (intervals, import settings) up to the top bar without prop drilling
- Move custom icons to a shared icons file
- Extract segmented control, toggle button and notice as reusable components
- Old standalone StatsExport, WithImportedSwitch, and NoticesIcon components removed; functionality consolidated into the options menu

**Before and after:**
https://github.com/user-attachments/assets/ec519314-629f-4034-87d9-9cded458f646

**New options menu:**
CleanShot 2026-04-14 at 11 06 32@2x

### Tests

### Changelog

### Documentation

### Dark mode


ukutaht:
Tested on preview. Feels great to me overall. A couple things I noticed while testing:

**1. Date picker has uneven padding when arrows not shown**

Screenshot 2026-04-14 at 14 07 21

The padding looks good with arrows.

**2. Date picker dropdown has some weirdness**

I cannot close the dropdown by clicking outside of it just above the datepicker: https://cap.so/s/8dg13mgn12zk6bb

Also I cannot replicate this for some reason but once I did manage to get it into this state where the focus ring shows and does not look great

Screenshot 2026-04-14 at 14 08 54

ukutaht:
Don't want to get too much into small matters of taste, but I couldn't help but notice that in the new variant the 'current visitors' text looks noticeably heavier than the site domain:

Screenshot 2026-04-14 at 14 59 25

As opposed to the current version where the domain is heavier than current visitors:

Screenshot 2026-04-14 at 15 00 25

The new version looks a bit strange to me but I can't put my finger on why or justify it too much :) did you choose it deliberately and prefer it this way?

sanne-san:
@ukutaht good catches, fixed them! RE typography of the site switcher and current visitors: it was a choice between consistency and hierarchy, and I opted for more consistency. I was hesitating as well though so I reverted it to the original. I think because the site domain isn't capitalised and the current visitors starts with a number, the latter feels heavier somehow.

assets/js/dashboard/nav-menu/filters-bar.tsx
```diff
@@ -258,20 +256,19 @@ const SeeMoreMenu = ({
           popover.toggleButton.classNames.rounded,
           popover.toggleButton.classNames.shadow,
           'justify-center',
-          'relative group'
+          'relative group',
+          'size-8'
         )}
         style={{
-          height: SEE_MORE_WIDTH_PX,
-          width: SEE_MORE_WIDTH_PX,
```
apata, line 265: nitpick, non-blocking: The reason size was given with style is for the actual size and the size used in calculations not to be out of sync. Now, the sync between `size-8` and `SEE_MORE_WIDTH_PX` has to be managed manually.

assets/js/dashboard/nav-menu/top-bar.tsx
```diff
@@ -46,14 +47,15 @@ function TopBarInner({ showCurrentVisitors }: TopBarProps) {
...
+        ref={leftActionsRef}
+      >
         <SiteSwitcher />
-        {showCurrentVisitors && (
-          <CurrentVisitors tooltipBoundaryRef={leftActionsRef} />
-        )}
+        {showCurrentVisitors && <CurrentVisitors />}
       </div>
-      <div className="flex w-full">
+      <div className="flex min-w-0 flex-1">
```
apata: issue, non-blocking: Now this section can shrink to nothing. Image

assets/js/dashboard/stats/graph/dashboard-options-context.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const DashboardOptionsContext =
+  createContext<DashboardOptionsContextType | null>(null)
+
+export function DashboardOptionsProvider({
+  children
+}: {
+  children: React.ReactNode
+}) {
+  const [options, setOptions] = useState<DashboardOptionsContextValue | null>(
```
apata: issue, non-blocking: This is duplicative state (between visitor-graph and this component), which causes us to need syncing code. I think a better split would be to have

`GraphIntervalContext` - solely responsible for calculating available intervals and the current one, and providing the means to change the current one

`ImportsIncludedContext` - same as above, just for imports, may itself depend on GraphIntervalContext

and to use the exposed context state in the components where they're needed, when fetching the graph and in the new triple dot menu.

assets/js/dashboard/stats/graph/fetch-top-stats.test.ts
```diff
@@ -229,6 +233,30 @@ describe(`${chooseMetrics.name}`, () => {
   )
 })
 
+describe(`${getPartialDayTimeRange.name}`, () => {
```
apata, line 237: suggestion, non-blocking: Let's also test that it handles timezones in the date string (I've not actually seen `2024-01-13T00:00:00`, always something like 
```
        "date_range": [
            "2026-01-19T00:00:00+00:00",
            "2026-04-19T23:59:59+00:00"
        ],
```

assets/js/dashboard/stats/graph/fetch-top-stats.ts
```diff
@@ -193,5 +193,28 @@ function formatTopStatsData(
     ? query.comparison_date_range[1].split('T')[0]
     : null
 
-  return { topStats, meta, from, to, comparingFrom, comparingTo }
+  const timeRange = getPartialDayTimeRange(query.date_range)
+
+  return { topStats, meta, from, to, comparingFrom, comparingTo, timeRange }
+}
+
+// Returns "until HH:MM" when the date range is a partial day (period=day for
```
apata: suggestion, non-blocking: If the comment about the function is exactly above the function, IDEs include it when hovering at a call site. 
```
/** Does something complex, explanation */ 
const f = () => {
//  ..
}
```

assets/js/dashboard/stats/graph/top-stats.js
```diff
@@ -170,31 +172,35 @@ export default function TopStats({
...
-                className="font-bold text-xl text-gray-500 dark:text-gray-400"
+                className="font-semibold text-[1.2rem] text-gray-500/80 dark:text-gray-400"
               >
                 {topStatNumberShort(stat.metric, stat.comparisonValue)}
               </p>
-              <p className="text-xs text-gray-500 dark:text-gray-400">
-                {formatDateRange(site, data.comparingFrom, data.comparingTo)}
+              <p className="text-xs text-gray-500 dark:text-gray-400 font-medium">
+                {data.timeRange
+                  ? `${formatDayShort(parseUTCDate(data.comparingFrom))}, ${data.timeRange}`
```
apata: issue, blocking: This works for Today vs Previous period, but it is incorrect when comparing Today vs Custom Period (pick yesterday from comparison calendar). The comparison query is made for the whole day, so we'd need to calculate and show `comparisonTimeRange` here.

assets/js/dashboard/stats/reports/change-arrow.test.tsx
```diff
@@ -3,56 +3,61 @@ import { render, screen } from '@testing-library/react'
...
 
-it('renders green for positive change', () => {
+it('renders up arrow for positive change', () => {
   render(<ChangeArrow change={1} className="text-xs" metric="visitors" />)
 
   const arrowElement = screen.getByTestId('change-arrow')
 
   expect(arrowElement).toHaveTextContent('↑ 1%')
-  expect(arrowElement.children[0]).toHaveClass('text-green-500')
+  expect(arrowElement.children[0]).toHaveAttribute('data-direction', 'up')
```
apata: question, non-blocking: Why did we get rid of the color asserts? I think they matter because bounce rate down being rendered as green is important. Up/down direction is already tested with `toHaveTextContent('↑ 1%')`. (The arrow icons are mocked with a particular unicode character)

assets/js/dashboard/site-switcher.tsx
```diff
@@ -90,6 +67,27 @@ const getSwitchToSiteURL = (
...
+const SiteSwitcherStatic = () => {
+  const currentSite = useSiteContext()
+
+  return (
+    <div
+      data-testid="site-switcher-static"
+      className={classNames(
+        popover.toggleButton.classNames.rounded,
+        'gap-x-1.5 px-2.5 font-medium text-gray-700 dark:text-gray-100',
+        '!pl-0'
```
apata: question, non-blocking: why padding left with !important? I don't see padding left previously declared in `popover.toggleButton.classNames.rounded`, just on the prev line.

assets/js/dashboard/site-switcher.tsx
```diff
@@ -170,28 +169,28 @@ export const SiteSwitcher = () => {
           <Popover.Button
             ref={buttonRef}
             className={classNames(
-              'flex items-center rounded h-9 text-sm leading-5 font-semibold dark:text-gray-100',
-              'hover:bg-gray-100 dark:hover:bg-gray-800'
+              popover.toggleButton.classNames.rounded,
+              popover.toggleButton.classNames.ghost,
+              '!pl-1.5'
```
apata, line 173: question, non-blocking: Does the switcher not need same left & right padding? E.g. `!px-1.5`

e2e/tests/dashboard/filtering.spec.ts
```diff
@@ -627,10 +640,7 @@ test.describe('location filtering tests', () => {
 
       await applyFilterButton(page).click()
 
-      await page
-        .getByRole('button', { name: 'See 1 more filter and actions' })
-        .click()
-
+      await page.getByRole('button', { name: /See.*more/ }).click()
```
apata, line 643: issue, non-blocking: These tests are running on a constant viewport (1024px). "See N more" depends on that, filter pills style, and other top bar items. 

Because the UI changed, the conditions to wrap to to "See N more" changed, so I can see why you added another filter. However, the comment is misleading and there's some missing asserts now.

In these tests, I think we should definitively check which filters of the added filters are visible without clicking "See more", that "See more" has the correct label, and that all the rest are visible after clicking "See more".

apata:
Great work! Pretty major changes at first glance, but opening the app it feels very familiar and natural, at least to me. With the top stats enhanced time labels issue solved (https://github.com/plausible/analytics/pull/6245#discussion_r3112374399), it's ready to go as far as I'm concerned.

assets/js/dashboard/stats/graph/dashboard-options-context.tsx
sanne-san: Should `ImportsIncludedContext` own its own query, or is a push from `visitor-graph` fine since the imports state comes from the top stats fetch?

apata: Valid question, I think that needs to be pushed.

apata: But its value should only be saved as state in one place, if possible. One can always access the value everywhere using what the context gives.

apata: Also, one more piece of advice on these contexts (or maybe the import one only), we might need to give it three possible states "we don't know yet" (default), "result A", "result B". That way, UI that depends on it can handle "we don't know yet" on its own if needed (e.g. placeholder, spinner, fallback).

ukutaht:
@sanne-san great, thanks! And thanks @apata for reviewing

Does anyone else have an issue with closing the date dropdown by clicking outside of the dropdown just above the date picker?

https://cap.so/s/d33nbhxy2b9nf9d

assets/js/dashboard/stats/reports/change-arrow.test.tsx
sanne-san: I changed `text-red-400` to `text-red-500` and the tests started failing – I generally prefer not to use classes for testing as it's quite subject to change so figured this would be better. But agree in this case testing the color matters, will revert!

assets/js/dashboard/site-switcher.tsx
sanne-san: Good catch, this was a mistake :)

assets/js/dashboard/site-switcher.tsx
sanne-san: This is intentional; reducing the left padding is specifically because it's the start of the bar and it makes the alignment with the rest of the dashboard a bit less awkward.

apata:
> Does anyone else have an issue with closing the date dropdown by clicking outside of the dropdown just above the date picker?
> 
> https://cap.so/s/d33nbhxy2b9nf9d

That happens for me as well, both macOS Chrome and Firefox, but it's present live as well. Looks like a click on the server-rendered header is swallowed so none of the React managed popover menus close properly.

sanne-san:
> That happens for me as well, both macOS Chrome and Firefox, but it's present live as well. Looks like a click on the server-rendered header is swallowed so none of the React managed popover menus close properly.

Yeah hadn't noticed that before, but same.

assets/js/dashboard/stats/reports/change-arrow.test.tsx
apata: True that classname-based tests are generally not a good idea! Let's avoid them as much as possible.

e2e/tests/dashboard/filtering.spec.ts
sanne-san: The problem was actually that locally tests would pass but not in CI and vice versa depending on whether the filter was hidden behind the "See more" button or not. I think it may be a difference in OS used, rendering the fonts differently maybe? 

I went with this approach because the tests themselves aren't necessarily about testing the overflow functionality. Definitively checking which filters are and aren't visible feels fragile, perhaps if we change the viewport to be much narrower we'd be able to reduce variability?

sanne-san:
@apata I implemented all your feedback, thank for the thorough review! The only thing I'm not sure about is what to do with the filtering tests based on [this comment](https://github.com/plausible/analytics/pull/6245#discussion_r3116377328) – lmk what you think!

apata (approved):

e2e/tests/dashboard/filtering.spec.ts
apata: The point of this test was to test that the filter pills get distributed between the bar and the see more menu (with no duplicates), which I don't think is inherently fragile. 

However, I didn't expect that there was a difference in layout on CI, because we're using https://www.npmjs.com/package/jsdom. I thought it would be fake in the same way, no matter what OS node is running it on. You did the right thing to just get it working then! 

At the moment, based on the inconsistency that you discovered and the struggle to make these tests pass in all envs (needs regex and code comments), I think we should actually delete this test file and move it to the e2e tests suite.

apata: Oh damn, it's already an e2e test. Ignore all of the above then! We should strive to increase parity between CI and local then.

PR #6246, opened 2026-04-14
Exploration: don't reload already selected steps

aerosol:
Even with the UI being a temporary prototype, having all the already selected steps reload on subsequent step selection is unnecessary.


zoldar:
👍

zoldar (approved):

PR #6247, opened 2026-04-14
Exploration direction

aerosol:
### Changes

This PR adds support for exploration direction - so the journey can be built top-bottom or bottom-up depending on user preference.
For now every change fires two requests - next steps followed by exploration funnel.
In a follow-up PR an attempt will be made to fire funnel query within single HTTP request and only if necessary (that is, only when a step is selected, not when refreshing suggestions).

image

### DIFF
--- a/assets/js/dashboard/stats/exploration.js
+++ b/assets/js/dashboard/stats/exploration.js
@@ -26,7 +26,7 @@ function stateWithApplicableFilters(dashboardState, steps) {
 }
 
 function toJourney(steps) {
-  steps.map((s) => { name: s.name, pathname: s.pathname })
+  return steps.map((s) => ({ name: s.name, pathname: s.pathname }))
 }
 
 function fetchColumnData(site, dashboardState, steps, filter, direction) {
--- a/assets/js/dashboard/stats/exploration.js
+++ b/assets/js/dashboard/stats/exploration.js
@@ -317,7 +317,9 @@ export function FunnelExploration() {
           </button>
         </div>
       </div>
-      <div className="flex gap-3">
+      <div
+        className={`flex gap-3 ${direction === 'backward' ? 'flex-row-reverse' : ''}`}
+      >
         {Array.from({ length: numColumns }, (_, i) => (
           <ExplorationColumn
             key={i}
--- a/assets/js/dashboard/stats/exploration.js
+++ b/assets/js/dashboard/stats/exploration.js
@@ -64,10 +64,6 @@ function ExplorationColumn({
   const [loading, setLoading] = useState(steps !== null)
   const [results, setResults] = useState([])
   const [filter, setFilter] = useState('')
-  const stepsFingerprint =
-    steps === null
-      ? null
-      : steps.map((step) => `${step.name}:${step.pathname}`).join(';')
 
   const debouncedOnSearchInputChange = useDebounce((event) =>
     setFilter(event.target.value)
@@ -111,7 +107,14 @@ function ExplorationColumn({
       cancelled = true
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [dashboardState, stepsFingerprint, filter, direction, site, selected])
+  }, [
+    dashboardState,
+    steps === null ? null : steps.join('|||'),
+    filter,
+    direction,
+    site,
+    selected
+  ])
 
   const maxVisitors = results.length > 0 ? results[0].visitors : 1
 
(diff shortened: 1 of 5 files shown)


assets/js/dashboard/stats/exploration.js
```diff
@@ -29,21 +38,44 @@ function fetchColumnData(site, dashboardState, steps, filter) {
...
+
+function fetchFunnelData(site, dashboardState, steps, direction) {
+  const stateToUse = stateWithApplicableFilters(dashboardState, steps)
+
+  const journey = []
+  if (steps.length > 0) {
+    for (const s of steps) {
+      journey.push({ name: s.name, pathname: s.pathname })
+    }
+  }
```
zoldar: Looking at it now, I'd replace it with `toJourney(steps)` and then define it as:

```js
function toJourney(steps) {
  steps.map((s) => { name: s.name, pathname: s.pathname })
}
```

Then it can be reused across `fetchFunnelData` and `fetchColumnData`

assets/js/dashboard/stats/exploration.js
```diff
@@ -29,21 +38,44 @@ function fetchColumnData(site, dashboardState, steps, filter) {
...
+  direction
 }) {
   const site = useSiteContext()
   const [loading, setLoading] = useState(steps !== null)
   const [results, setResults] = useState([])
   const [filter, setFilter] = useState('')
+  const stepsFingerprint =
+    steps === null
+      ? null
+      : steps.map((step) => `${step.name}:${step.pathname}`).join(';')
```
zoldar: Originally it wasn't even array but a concatenated string. I wasn't sure why. Now I took a look at https://react.dev/reference/react/useEffect and it seems dependencies are compared using `Object.is`. Arrays are objects, so `Object.is([1, 2, 3], [1, 2, 3])` returns false as those are distinct instances of array. Given steps are not only modified in place over the lifetime of the component (like dashboardState seems to be?) but also replaced with a new instance under some circumstances, I guess we should indeed concatenate this array as well to avoid excessive reruns.

assets/js/dashboard/stats/exploration.js
```diff
@@ -194,22 +250,84 @@ export function FunnelExploration() {
...
+              direction === EXPLORATION_DIRECTIONS.BACKWARD
+                ? 'bg-indigo-500 text-white'
+                : 'bg-white dark:bg-gray-800 text-gray-600 dark:text-gray-300'
+            }`}
+          >
+            Backward
+          </button>
+        </div>
+      </div>
       <div className="flex gap-3">
```
zoldar: When direction is 'BACKWARD`, we could change the flex direction to reverse?

aerosol: Clever https://github.com/plausible/analytics/pull/6247/commits/16e35333fd6c72b2b92f25d0941441973a62191c

assets/js/dashboard/stats/exploration.js
aerosol: e4b41886d8

assets/js/dashboard/stats/exploration.js
```diff
@@ -7,43 +7,67 @@ import { useDashboardStateContext } from '../dashboard-state-context'
...
+  return {
+    ...dashboardState,
+    filters: dashboardState.filters.filter(
+      ([_op, key]) => !PAGE_FILTER_KEYS.includes(key)
+    )
   }
+}
+
+function toJourney(steps) {
+  steps.map((s) => { name: s.name, pathname: s.pathname })
```
zoldar: Sorry 🤦 - this most likely should be:

```suggestion
  steps.map((s) => { return { name: s.name, pathname: s.pathname } })
```

assets/js/dashboard/stats/exploration.js
aerosol: https://github.com/plausible/analytics/pull/6247/commits/b159439fa2a3716267f2f0f8a7824d6c1c83275a

assets/js/dashboard/stats/exploration.js
aerosol: right, do you mean I should change it from this:

```javascript
return steps.map((s) => ({ name: s.name, pathname: s.pathname }))
```

assets/js/dashboard/stats/exploration.js
zoldar: Sorry, scratch that, that was already a string. 🤡 It can be reverted.

zoldar (approved):
Once the excessive concat is removed, it's good to go as a basis for futher iterations 👍

PR #6249, opened 2026-04-14
Treat a sequence of events with same name/path as a single step

zoldar:
### Changes

Sequences of events with the same name and pathname within user session are now treated as a single step when exploring user journeys.

### Tests

### DIFF
--- a/test/plausible/stats/exploration_test.exs
+++ b/test/plausible/stats/exploration_test.exs
@@ -241,5 +241,96 @@ defmodule Plausible.Stats.ExplorationTest do
       assert next_step2.step.pathname == "/login"
       assert next_step2.visitors == 1
     end
+
+    test "treats identical sequence of events as a single step" do
+      site = new_site()
+
+      now = DateTime.utc_now()
+
+      populate_stats(site, [
+        build(:pageview,
+          user_id: 123,
+          pathname: "/home",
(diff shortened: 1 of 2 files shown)


test/plausible/stats/exploration_test.exs
```diff
@@ -241,5 +241,96 @@ defmodule Plausible.Stats.ExplorationTest do
       assert next_step2.step.pathname == "/login"
       assert next_step2.visitors == 1
     end
+
+    test "treats identical sequence of events as a single step" do
```
aerosol, line 245: would it make sense to add a test verifying that same events next to each other, but with different user IDs, are not treated as one?

aerosol (approved):
let's go! 🚀

test/plausible/stats/exploration_test.exs
aerosol, line 245: or a backward direction test for that matter?

zoldar: https://github.com/plausible/analytics/pull/6249/commits/75bc9e78b65cfeec132c9d6088f634d53bfd25f5

zoldar: https://github.com/plausible/analytics/pull/6249/commits/75bc9e78b65cfeec132c9d6088f634d53bfd25f5

PR #6250, opened 2026-04-15
Implement conversion rate and rate step calculation in exploration funnel API

zoldar:
### Changes

This PR adds computation of conversion rate and rate step, the same way it is done for the existing funnels.

Additionally, the conversion rate is used to render bar graphs of the exploration funnel. In the case of bar graphs in the next step, the value is computed relative to the number of visitors in the first step of the funnel, which reflects conversion rate computation.

image

When exploring funnel in backward direction, the bar graph is computed always relative to the max visitors in the current step suggestions.

image

The visitor numbers in the suggestions in backward direction will be addressed in a separate PR.

### Tests


aerosol (approved):
🙇

PR #6256, opened 2026-04-15
Add fixed heights to /sites sparklines

ukutaht:
Adds fixed heights to sparkline graphs on /sites pages. This avoids the layout shift during the transition from loading state to loaded state. Currently the height of the cards change during the transition which is visually a bit annoying.


zoldar (approved):
LGTM, though there's an unrelated credo complaint that needs addressing still.

aerosol (approved):

PR #6257, opened 2026-04-15
Suggest a journey

aerosol:
### Changes

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/exploration.js
+++ b/assets/js/dashboard/stats/exploration.js
@@ -51,6 +51,14 @@ function fetchFunnelData(site, dashboardState, steps, direction) {
   })
 }
 
+function fetchInterestingFunnel(site, dashboardState) {
+  return api.post(
+    url.apiPath(site, '/exploration/interesting-funnel'),
+    dashboardState,
+    {}
+  )
+}
+
 function ExplorationColumn({
   header,
   steps,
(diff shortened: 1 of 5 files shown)


assets/js/dashboard/stats/exploration.js
```diff
@@ -152,18 +160,23 @@ function ExplorationColumn({
...
         </div>
       ) : (
         <ul className="flex flex-col gap-y-0.5 px-1.5 pb-1.5 h-108 overflow-y-auto">
           {(selected
-            ? results.filter(
-                ({ step }) =>
-                  step.name === selected.name &&
-                  step.pathname === selected.pathname
-              )
+            ? (() => {
```
aerosol: I think this is awful, but I don't know how to address this properly. 

The issue is, the component sometimes fetches candidates and sets "selected", and sometimes the result is just provided as an input directly. 

And the problem (I think) is due to my own uneducated addition - https://github.com/plausible/analytics/pull/6246

assets/js/dashboard/stats/exploration.js
```diff
@@ -120,6 +135,22 @@ function ExplorationColumn({
       ? results[0]?.visitors
       : maxVisitors || results[0]?.visitors
```
zoldar: This is completely BTW - this condition can be simplified to:

```js
const stepMaxVisitors = maxVisitors || results[0]?.visitors
```

since the backwards funnel definition has changed - we can always compute the width relative to the start of the funnel.

aerosol: 706b4f7805

zoldar (approved):
LGTM. I only wonder about this ref-based approach (which is cool!) - I haven't seen it used like this anywhere else in our dashboard code. I assume that's because we usually opt for deactivating UI elements that would re-trigger request during loading state and here we keep the UI responsive, allowing to interrupt the suggestion at any point, right?

aerosol:
> LGTM. I only wonder about this ref-based approach (which is cool!) - I haven't seen it used like this anywhere else in our dashboard code. I assume that's because we usually opt for deactivating UI elements that would re-trigger request during loading state and here we keep the UI responsive, allowing to interrupt the suggestion at any point, right?

Yes, deactivating remaining UI controls would work for sure, though "interesting funnel" may take some time to load. So this makes e.g. backward/forward toggle still usable.

sanne-san (approved):
Looks good, I'll think about the best way to show this in the UI but fine to merge for now!

PR #6258, opened 2026-04-16
Fix bad suggestions due to inconsistent ordering in window queries

zoldar:
### Changes

Addresses a bug where exact same path and pathname combination is suggested in a following step.

It was caused by inconsistent ordering in cases where events occurred within the same timestamp second. 

Before the fix:

image

After the fix:

image

### Tests


aerosol (approved):
:bow:

PR #6259, opened 2026-04-16
Modify backward funnel exploration behaviour

zoldar:
### Changes

This PR alters the backward exploration behavior, so that every step before returns the number of visitors who moved from that step to the next step closer to the end. This is in contrast to the current behaviour where visitors are counted as for a forward funnel exploration.

### Tests


lib/plausible/stats/exploration/journey_step.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Plausible.Stats.Exploration.JourneyStep do
```
aerosol: why not Journey.Step anymore? We have Funnel.Step

aerosol (approved):
🧠

lib/plausible/stats/exploration/journey_step.ex
zoldar: That was a byproduct of me screwing around with macros (and eventually giving up on them). I have reverted that change for now https://github.com/plausible/analytics/pull/6259/commits/1c006c7504d23ecdf4eefd2990afe5ba2535301a. If `Journey` ever becomes an actual entity/struct, we can extract that.

PR #6262, opened 2026-04-20
Dashboard Pages v2

RobertJoonas:
> Instructions to reviewer(s):
> 
> I highly recommended to review this PR commit-by-commit -- I've renamed files in separate commits to give the opportunity to see the old and new component diff side by side.

### Changes (code-structure oriented description)

* IndexBreakdown (aka ListReport v2) - for the 9-entry breakdowns on the dashboard index page
  * Caches responses via TanStack `useQuery`
  * Disables FlipMove animations for any update that's not caused by a realtime "tick"
  * Replaces the broken `FadeIn` component with a simple set of Tailwind classes. The animation duration is set to 300ms to make it noticeable.
* DetailsBreakdown (aka BreakdownModal v2) - for details views
  * uses a new `usePaginatedQueryAPI` hook which uses a StatsQuery object in its `queryKey` rather than a DashboardState.
  * `afterFetchData`  props -> `onDataReady` (Same as for IndexBreakdown. Instead of a separate function for next page vs initial page, the single handler now receives all pages of results which will be needed in the conversions report to determine whether revenue metrics should be displayed or not)
  * Decides metric widths on its own, rather than relying on the Metric class.
* Both Index and DetailsBreakdowns now use an `onDataReady` callback instead of `afterFetchData` and `afterFetchNextPage` -- need to fire the callback when response is read from cache too, not only after a real fetch call)
* Avoid extra state for `apiResponse.query` and `apiResponse.meta` by letting TanStack query cache both as part of the response. 
* Both Index and DetailsBreakdowns now use a shared `ColumnConfiguration` for defining columns. In order to generate columns, actual data is needed. Unlike the legacy `ColumnConfiguration`, the new version enforces both `renderLabel` and `renderCell` as functions, therefore leaving the rendering components (BreakdownTable and IndexBreakdownRenderer) dumb about any special cases around dimension/metric values/labels.
* Define metric labelling logic in `metrics.ts`. Note: it is already extended to components that are not yet migrated (e.g. conversions). Currently dead code but will be used soon.
* Move pages / entry-pages / exit-pages to the new components (both detailed and index reports)
* Add external link icon (the one that shows on row hover) to pages/entry-pages/exit-pages details views (currently only available in index-breakdowns, i.e. ListReports).

#### Recap of user facing changes

* Add external link icon (the one that shows on row hover) to pages/entry-pages/exit-pages details views (currently only available in index-breakdowns, i.e. ListReports).
* Improvements to Top Pages, Entry Pages, and Exit Pages on the dashboard index page:
  * Leverage cache to avoid extra fetching and make the UX faster
  * Make the fade-in animation work
  * Disable FlipMove animations when update not caused by a realtime tick -- e.g. currently on production, if you change the dashboard period from 28d -> realtime, FlipMove animation triggers. This PR disables it.

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/components/table.tsx
```diff
@@ -110,106 +31,67 @@ export const ItemRow = <T extends Record<string, string | number | ReactNode>>({
...
+      onClick={handleClick}
     >
-      {columns.map(({ key, width, align, renderValue, renderItem }) => (
-        <TableCell
-          key={`${(pageIndex ?? null) === null ? '' : `page_${pageIndex}_`}row_${rowIndex}_${String(key)}`}
-          className={width}
-          align={align}
+      {columns.map((col) => (
+        <td
+          key={`${col.key}_${rowKey}`}
```
apata, line 38: nitpick, non-blocking: due to limit + offset pagination, the same dimension/page name can actually appear on multiple pages

assets/js/dashboard/stats/modals/pages.js
```diff
@@ -5,12 +5,12 @@ import {
   isRealTimeDashboard
 } from '../../util/filters'
 import { addFilter, revenueAvailable } from '../../dashboard-state'
-import BreakdownModal from './breakdown-modal'
+import BreakdownModal from './breakdown-modal-legacy'
```
apata: question: still needed?

assets/js/dashboard/stats/pages/pages.tsx
```diff
@@ -0,0 +1,108 @@
...
+    (row: QueryResultRow) => url.externalLinkForPage(site, row.dimensions[0]),
+    [site]
+  )
+
+  return (
+    <Modal>
+      <DetailsBreakdown
+        title="Top pages"
+        dimensionLabel="Page url"
+        dimensions={[DIMENSION]}
```
apata, line 99: nitpick, non-blocking: Because we support a single dimension at the moment, maybe we can refactor dimensions => dimension until we get to multi-dimension reports? Otherwise, having dimensions in plural and just one dimensionLabel can be a bit misleading.

assets/js/dashboard/stats/pages/pages.tsx
```diff
@@ -0,0 +1,108 @@
...
+    prefix: 'page',
+    filter: ['is', 'page', [row.dimensions[0]]] as Filter
+  }
+}
+
+function addSearchFilter(statsQuery: StatsQuery, search: string) {
+  return addDimensionSearchFilter(statsQuery, DIMENSION, search)
+}
+
+export function PagesIndex({
```
apata, line 42: nitpick, non-blocking: the modules `pages`, `entry-pages` and `exit-pages` are structurally the same. If we convert all the reports, there'll be over ten of these. 

This is ok and allows us do things like this partial move to the new API, but some lines are duplicated and could be extracted. 
```
{
  onDataReady?: (data: QueryApiResponse) => void
}

addSearchFilter(statsQuery: StatsQuery, search: string)
```

There's several avenues to extract, one simple way is to create the following configuration type. 

```ts
type ReportOpts = {
  title: string
  dimension: string
  dimensionLabel: string
  addSearchFilter: (statsQuery: StatsQuery, search: string) => StatsQuery
  getFilterInfo: (row: QueryResultRow) => {prefix: string, filter: Filter}
  getBriefMetrics: (opts: { isRealtime: boolean, hasConversionGoalFilter: boolean, isRevenueAvailable: boolean }) => Metric[]
  getDetailedMetrics: (opts: { isRealtime: boolean, hasConversionGoalFilter: boolean, isRevenueAvailable: boolean }) => Metric[]
}

// Usage
const pagesOpts: ReportOpts = {
  title: "Top pages",
  dimension: "event:page",
  dimensionLabel: "Page url",
  addSearchFilter: (statsQuery, search) => ...,
  getBriefMetrics: (opts) => ...,
  getDetailedMetrics: (opts) => ..., 
}

const exitPagesOpts: ReportOpts = {
  title: "Exit pages",
  dimension: "visit:exit_page",  
  ...
}
```

Another one that springs to mind is to get rid of the intermediate components of `[Pages|EntryPages|ExitPages]Index` and `[Pages|ExitPages|EntryPages]Details` altogether. `DetailsBreakdown` and `IndexBreakdown` could handle what these components do.

assets/js/dashboard/stats/pages/pages.tsx
RobertJoonas: This will be needed very soon because browser versions is using `dimensions = ["visit:browser", "visit:browser_version"]`

assets/js/dashboard/stats/metrics.test.ts
```diff
@@ -0,0 +1,236 @@
...
+describe('isSortable', () => {
+  it('returns false for total_visitors', () => {
+    expect(isSortable('total_visitors')).toBe(false)
+  })
+
+  it.each(['visitors', 'bounce_rate', 'visit_duration', 'conversion_rate'])(
+    'returns true for %s',
+    (metric) => {
+      expect(isSortable(metric as Parameters<typeof isSortable>[0])).toBe(true)
+    }
```
apata: We can cast earlier (or do `as const` on the list)
```suggestion
  const sortableMetrics: Metric[] = ['visitors', 'bounce_rate', 'visit_duration', 'conversion_rate']
  it.each(sortableMetrics)(
    'returns true for %s',
    (metric) => {
      expect(isSortable(metric)).toBe(true)
    }
```

assets/js/dashboard/stats/metrics.test.ts
```diff
@@ -0,0 +1,236 @@
+import { isSortable, getMetricLabel, getBreakdownMetricLabel } from './metrics'
+
+describe('isSortable', () => {
```
apata: nitpick, non-blocking: In other unit tests with describe blocks for utility functions / utility modules, we use the style ```describe(`${isSortable.name}`, ...```. The advantage is that a function name refactor immediately affects the describe block title as well. Otherwise they tend to get out of sync.

assets/js/dashboard/stats/metrics.ts
```diff
@@ -1,4 +1,12 @@
-import { Metric } from '../../types/query-api'
+import { Metric as PublicApiMetric } from '../../types/query-api'
+
+export type Metric = PublicApiMetric | 'total_visitors' | 'exit_rate'
+
+const NOT_SORTABLE = ['total_visitors']
+
+export const isSortable = (metric: Metric): boolean => {
+  return !NOT_SORTABLE.includes(metric)
+}
```
apata: suggestion, non-blocking: I think it's better to err on the side of the FE forbidding rather than permitting sorting for newly added metrics (`const NOT_SORTABLE` -> `const SORTABLE_METRICS`). I think it's better for the user not to be able to sort something that should be sortable than get an error about invalid sort.

assets/js/dashboard/stats/metrics.ts
```diff
@@ -33,5 +41,125 @@ export const getMetricLabel = (
...
+export const getBreakdownMetricLabel = (
+  metric: Metric,
+  {
+    hasConversionGoalFilter,
+    isRealtime,
+    dimensions
+  }: {
+    hasConversionGoalFilter: boolean
+    isRealtime: boolean
+    dimensions: string[]
```
apata, line 76: nitpick, non-blocking: Because this function currently handles and needs to handle one dimension only (`switch (dimension[0]`), I think it'd be better to have it accept the single `dimension` as argument. (Same comment as https://github.com/plausible/analytics/pull/6262/changes#r3188747066)

assets/js/dashboard/stats/metrics.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      ['pageviews', 'Pageviews'],
+      ['average_revenue', 'Average'],
+      ['total_revenue', 'Revenue']
+    ] as const)('%s -> %s', (metric, expected) => {
+      expect(getBreakdownMetricLabel(metric, { ...defaults, dimensions })).toBe(
+        expected
+      )
+    })
+
+    it('delegates to getMetricLabel for other metrics', () => {
```
apata: nitpick, non-blocking: I think this test name could be a bit more accurate as to what's being verified, because the test doesn't prove that it's delegating. If we wanted to prove that, we should do something like `toBe(getMetricLabel('bounce_rate', {...defaults}))`

assets/js/dashboard/stats/graph/fetch-main-graph.ts
apata: suggestion, non-blocking: Can we unify these properties with the type from the api module?

assets/js/dashboard/stats/pages/pages.tsx
apata: Thanks for explaining, I didn't even think of that! Is it clear already that the API for it is going to have a single dimensionLabel but multiple dimensions?

assets/js/dashboard/stats/modals/details-breakdown.tsx
```diff
@@ -0,0 +1,518 @@
...
+
+    return [
+      {
+        key: 'dimension',
+        renderLabel: () => dimensionLabel,
+        renderCell: (row, isActive) => (
+          <DimensionCell
+            row={row}
+            getFilterInfo={getFilterInfo}
+            getExternalLinkUrl={getExternalLinkUrl}
```
apata, line 179: issue, blocking: I think this (_but could be something else as well_) is causing premature wrapping of page URLs. Left: this PR, right: staging.

Image

assets/js/dashboard/stats/pages/pages.tsx
RobertJoonas: Definitely worth DRYing it a bit. In fact, Sources and Locations modals are already handled like that in the legacy code:

```javascript
// in router.tsx

export const sourcesRoute = {
  path: 'sources',
  element: <SourcesModal currentView="sources" />
}
export const channelsRoute = {
  path: 'channels',
  element: <SourcesModal currentView="channels" />
}
export const utmMediumsRoute = {
  path: 'utm_mediums',
  element: <SourcesModal currentView="utm_mediums" />
}
export const utmSourcesRoute = {
  path: 'utm_sources',
  element: <SourcesModal currentView="utm_sources" />
}
// ...
export const countriesRoute = {
  path: 'countries',
  element: <LocationsModal currentView="countries" />
}
export const regionsRoute = {
  path: 'regions',
  element: <LocationsModal currentView="regions" />
}
export const citiesRoute = {
  path: 'cities',
  element: <LocationsModal currentView="cities" />
}
```

So yeah, we could definitely do the same with pages as well. If that's okay with you though, I would leave that as a follow-up (which I will tackle as a priority after this is merged). This PR is already quite big 😄

assets/js/dashboard/hooks/use-order-by.ts
```diff
@@ -1,15 +1,15 @@
 import { useCallback, useEffect, useMemo, useState } from 'react'
-import { Metric } from '../stats/reports/metrics'
+import { isSortable, Metric } from '../stats/metrics'
 import { getDomainScopedStorageKey, getItem, setItem } from '../util/storage'
 import { useSiteContext } from '../site-context'
-import { ReportInfo } from '../stats/modals/breakdown-modal'
+import { ReportInfo } from '../stats/modals/breakdown-modal-legacy'
```
apata: suggestion, non-blocking: We can refactor the args of the functions in this new module to accept dimensionLabel directly

assets/js/dashboard/hooks/use-order-by.ts
```diff
@@ -73,25 +73,25 @@ export function cycleSortDirection(
   }
 }
 
-export function findOrderIndex(orderBy: OrderBy, metric: Pick<Metric, 'key'>) {
-  return orderBy.findIndex(([metricKey]) => metricKey === metric.key)
+export function findOrderIndex(orderBy: OrderBy, metric: Metric) {
+  return orderBy.findIndex(([m]) => m === metric)
 }
```
apata: suggestion, non-blocking: this wrapper around findIndex isn't needed for the new API, `findIndex` can be called directly

assets/js/dashboard/hooks/use-order-by.ts
```diff
@@ -23,21 +23,21 @@ export function useOrderBy({
...
 }) {
   const [orderBy, setOrderBy] = useState<OrderBy>([])
-  const orderByDictionary: Record<Metric['key'], SortDirection> = useMemo(
+  const orderByDictionary = useMemo(
     () =>
-      orderBy.length
+      (orderBy.length
         ? Object.fromEntries(orderBy)
-        : Object.fromEntries(defaultOrderBy),
+        : Object.fromEntries(defaultOrderBy)) as Record<Metric, SortDirection>,
```
apata: nitpick, non-blocking: Casting (` as ...`) should ideally be avoided. If we declare the const to be of this type (`const orderByDictionary: Record<Metric, SortDirection> = ...`), we get Typescript compile errors when the const isn't set to this value, which most of the times is useful. Casting is a bit of an escape hatch, to be used for situations when the compiler can't possibly know. Why didn't we need to cast before and now do? Is it because `Order = [string, SortDirection]` should be `[Metric, SortDirection]`?

assets/js/dashboard/hooks/use-order-by.ts
```diff
@@ -145,7 +145,7 @@ export function getStoredOrderBy({
     if (
       validateOrderBy(
         parsed,
-        metrics.filter((m) => m.sortable)
+        metrics.filter((m) => isSortable(m))
```
apata:
```suggestion
        metrics.filter(isSortable)
```
This is a bit shorter

assets/js/dashboard/stats/pages/pages.tsx
apata: Yup, totally fine to tackle any time or not at all!

apata (commented):
Great work! I believe it's very close to being mergable, with the only blocker for my approval being the page urls now wrapping to multiple lines in details views. The other things we discussed in sync, like using the queryKey to build the statsQuery or maintaining pages test suite for longer aren't blockers from my POV.

assets/js/dashboard/stats/pages/pages.tsx
RobertJoonas: > Is it clear already that the API for it is going to have a single dimensionLabel but multiple dimensions?

Yes. `dimensionLabel` is FE responsibility and not something the API needs to be concerned about. It will just be "Browser version" for the browser versions breakdown, and the `DimensionCell` renderer will just concatenate the browser name with version.

apata: Roger that on FE responsibility! I personally think there's some value in actually rendering each dimension as a separate column (url, hostname), but that's another discussion altogether.

assets/js/dashboard/hooks/use-order-by.ts
RobertJoonas: > Casting ( as ...) should ideally be avoided. If we declare the const to be of this type (const orderByDictionary: Record<Metric, SortDirection> = ...), we get Typescript compile errors when the const isn't set to this value, which most of the times is useful. Casting is a bit of an escape hatch, to be used for situations when the compiler can't possibly know.

Makes sense! Thanks for explaining!

> Why didn't we need to cast before and now do? Is it because Order = [string, SortDirection] should be [Metric, SortDirection]?

Good question. Yeah I believe it was related to the Metric type but to be honest I don't remember why I added it at some point.

However, I already changed the `Order` type to be `[string, SortDirection]` rather than `[Metric, SortDirection]` (that's because we need to order by dimensions themselves too and there's no explicit `Dimension` type).

I guess since useOrderBy is only dealing with metric ordering, the correct approach would be to have two separate types:

* `OrderBy = [string, SortDirection][]` in `stats-query.ts`
* `MetricOrderBy = [Metric, SortDirection][]` in `use-order-by.ts`

Happy to tackle that as a follow-up if you think that makes sense. For now I've simply removed the `as ...` cast. https://github.com/plausible/analytics/pull/6262/changes/2f64b3a7e95c948efbeab36bb4a1a82fe2f576e6

assets/js/dashboard/stats/modals/details-breakdown.tsx
RobertJoonas: Good catch! That's because the external link icon (that shows on hover) was added. I've now increased the dimension cell width which gives it a bit more horizontal space. https://github.com/plausible/analytics/pull/6262/commits/d19018ef0e0d45164bf88052c3518dfa945a2138

assets/js/dashboard/stats/metrics.test.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6262/commits/e2819727a90f556b937975a74913b0de21a83c12

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6262/commits/7716c108628f965fbdae3c4cd5c4261ec1975808

assets/js/dashboard/hooks/use-order-by.ts
RobertJoonas: Much cleaner. Thanks! https://github.com/plausible/analytics/pull/6262/commits/04e8502d03f6d896d3e65836081685a8b36a7e4f

assets/js/dashboard/hooks/use-order-by.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6262/commits/fb6a759a7b6b03df257d61013ca1098569a498e0

assets/js/dashboard/hooks/use-order-by.ts
RobertJoonas: Good catch! Removed the wrapper function and its tests from the new useOrderBy https://github.com/plausible/analytics/pull/6262/commits/1319a2e1f9c604b8fce701ae07a43c167d2139a5

assets/js/dashboard/components/table.tsx
RobertJoonas: https://github.com/plausible/analytics/pull/6262/changes/bc9fde139560554616b06763f2c48597af86e21a

assets/js/dashboard/stats/metrics.test.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6262/changes/3c4f8d4a9a2e97474f499a08f26e38f02ab0b716

assets/js/dashboard/stats/metrics.test.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6262/changes/3c4f8d4a9a2e97474f499a08f26e38f02ab0b716

assets/js/dashboard/stats/metrics.ts
RobertJoonas: Makes sense! https://github.com/plausible/analytics/pull/6262/changes/064ac74f1fdd902c74951062d33166abd8f4f17e

assets/js/dashboard/stats/modals/pages.js
RobertJoonas: Good catch! Not needed.

I've removed it now in https://github.com/plausible/analytics/pull/6262/changes/1093252510255ee6b02b80ab3e51dc52e3763be6.

Since the legacy PagesModal was used in `breakdown-modal.test.tsx`, I've simply updated that file to use BrowsersModal instead. Same thing. We'll need to add a details-breakdown.test.tsx to test the same thing for new breakdowns, but I'll do that in a follow-up.

assets/js/dashboard/components/table.tsx
apata: Good! I would prefer page, then row, then col (_right now it's row, page, col_) order for the key, but it's very minor.

assets/js/dashboard/hooks/use-order-by.ts
apata: I see, I didn't even think of dimension sort. Your idea to split the type is good (and we can also import `Dimensions` from query-api.d.ts in the FE and do `[Metric | Dimensions, SortDirection][]`).

assets/js/dashboard/hooks/use-order-by.ts
apata: I just noticed there's one more instance of the same thing in the same file that can be shortened

apata (approved):

PR #6267, opened 2026-04-20
Record trace metadata when executing funnel and exploration queries

zoldar:
### Changes

This change enables viewing the funnel and exploration queries via /debug/clickhouse endpoint

### DIFF
--- a/extra/lib/plausible/stats/funnel.ex
+++ b/extra/lib/plausible/stats/funnel.ex
@@ -36,6 +36,8 @@ defmodule Plausible.Stats.Funnel do
       |> Query.set(preloaded_goals: %{all: [], matching_toplevel_filters: goals})
       |> Base.base_event_query()
       |> funnel_query(funnel)
+      # We pass the query struct to record query metadata for
+      # the CH debug console.
       |> ClickhouseRepo.all(query: query)
 
     # Funnel definition steps are 1-indexed, if there's index 0 in the resulting query,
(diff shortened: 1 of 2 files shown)


extra/lib/plausible/stats/funnel.ex
```diff
@@ -62,7 +63,7 @@ defmodule Plausible.Stats.Funnel do
      }}
   end
 
-  defp query_funnel(query, funnel_definition) do
+  defp funnel_query(query, funnel_definition) do
```
aerosol, line 68: Bond, James

lib/plausible/stats/exploration.ex
```diff
@@ -60,7 +60,7 @@ defmodule Plausible.Stats.Exploration do
     query
     |> Base.base_event_query()
     |> next_steps_query(journey, search_term, direction)
-    |> ClickhouseRepo.all()
+    |> ClickhouseRepo.all(query: query)
```
aerosol, line 65: Is this only to carry debug metadata? In such case maybe we can explicitly handle direct metadata pass, even if it's retrieved from that initial query, the intent will be much clearer. Otherwise - are we not bothered by the query being modified in the pipeline?

aerosol (changes requested):

lib/plausible/stats/exploration.ex
aerosol: Ah nevermind, just please leave a comment on why are we passing the query.

aerosol (approved):

lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6267/commits/4931cacd152170dbbe4766020890c35a6cb829ca

PR #6268, opened 2026-04-21
Put journey exploration under a single API call

aerosol:
### Changes

This PR introduces new `exploration/next-with-funnel` endpoint. Name subject to change, perhaps just "exploration"?

Funnel can be loaded within the same API call (though still in a separate query), based on `include_funnel` boolean input parameter.

The funnel represents visitor counts for already-confirmed journey steps and is independent of the search filter. Making it optional avoids running queries needlessly - only step selection warrants a funnel refresh.

With a single combined endpoint, the fetch must happen at the level that owns both next and funnel state, so that's `FunnelExploration`. Only one column at a time is "active" (unselected, showing suggestions). All other columns either show a single selected item or nothing. 

Should we completely remove previous endpoints?

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/exploration.js
+++ b/assets/js/dashboard/stats/exploration.js
@@ -29,26 +29,27 @@ function toJourney(steps) {
   return steps.map((s) => ({ name: s.name, pathname: s.pathname }))
 }
 
-function fetchColumnData(site, dashboardState, steps, filter, direction) {
-  // Page filters only apply to the first step — strip them for subsequent columns
+function fetchNextWithFunnel(
+  site,
+  dashboardState,
+  steps,
+  filter,
+  direction,
+  includeFunnel
+) {
   const stateToUse = stateWithApplicableFilters(dashboardState, steps)
   const journey = toJourney(steps)
 
-  return api.post(url.apiPath(site, '/exploration/next'), stateToUse, {
-    journey: JSON.stringify(journey),
-    search_term: filter,
-    direction
-  })
-}
-
-function fetchFunnelData(site, dashboardState, steps, direction) {
-  const stateToUse = stateWithApplicableFilters(dashboardState, steps)
-  const journey = toJourney(steps)
-
-  return api.post(url.apiPath(site, '/exploration/funnel'), stateToUse, {
-    journey: JSON.stringify(journey),
-    direction
-  })
+  return api.post(
+    url.apiPath(site, '/exploration/next-with-funnel'),
+    stateToUse,
+    {
+      journey: JSON.stringify(journey),
+      search_term: filter,
+      direction,
+      include_funnel: includeFunnel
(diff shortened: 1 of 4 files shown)


assets/js/dashboard/stats/exploration.js
```diff
@@ -319,34 +314,68 @@ export function FunnelExploration() {
...
+  // search doesn't affect it.
   useEffect(() => {
-    if (steps.length === 0) {
+    setActiveColumnLoading(true)
+    setActiveColumnResults([])
+
+    const stepsChanged = prevStepsRef.current !== steps
+    const contextChanged =
+      prevDirectionRef.current !== direction ||
+      prevDashboardStateRef.current !== dashboardState
```
zoldar: Those 2 booleans could be merged into a single one to reduce confusion - as both are always checked together with "||" anyway.

lib/plausible_web/controllers/api/stats_controller.ex
```diff
@@ -179,6 +180,32 @@ defmodule PlausibleWeb.Api.StatsController do
...
+         {:ok, next_steps} <- Exploration.next_steps(query, journey, search_term, direction) do
+      funnel =
+        if include_funnel? do
+          case Exploration.journey_funnel(query, journey, direction) do
+            {:ok, funnel_data} -> funnel_data
+            {:error, :empty_journey} -> []
+          end
+        else
+          []
+        end
```
zoldar:
```elixir
...,
{:ok, funnel} <- maybe_include_funnel(include_funnel?, query, journey, direction) do
```

instead perhaps?

zoldar (approved):
LGTM with minor nits.

@sanne-san a heads up about this change. The state management is improved here in that, for instance, clearing the step ensures the filter is properly reset too (+ we avoid excess requests to the backend), but it also does not preserve the suggestions from the previous steps anymore. Additional state might be necessary to keep track of them for rendering, if needed.

assets/js/dashboard/stats/exploration.js
aerosol: https://github.com/plausible/analytics/pull/6268/commits/8ab3dea3095b508b24f42892342d999046eae5de

lib/plausible_web/controllers/api/stats_controller.ex
aerosol: Not sure what you mean - are you suggesting making it a `with` clause and responding with an error on empty journey?

zoldar: You can return en empty (`[]`) journey in case of error like you do now. It's purely about code organisation.

aerosol: 94269cd7be

PR #6269, opened 2026-04-21
Remove redundant handler causing double-fetch in realtime

RobertJoonas:
### Changes

Calling refetch manually in the onTick handler ignores the cached value anyway so there's no need to explicitly invalidate the cached realtime responses. Currently that's just triggering another background refetch.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/visitor-graph.tsx
+++ b/assets/js/dashboard/stats/graph/visitor-graph.tsx
@@ -12,7 +12,6 @@ import { PlausibleSite, useSiteContext } from '../../site-context'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { Metric } from '../../../types/query-api'
 import { DashboardPeriod } from '../../dashboard-time-periods'
-import { DashboardState } from '../../dashboard-state'
 import { nowForSite } from '../../util/date'
 import { getStaleTime } from '../../hooks/api-client'
 import { MainGraph, MainGraphContainer, useMainGraphWidth } from './main-graph'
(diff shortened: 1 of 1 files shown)


assets/js/dashboard/stats/graph/visitor-graph.tsx
```diff
@@ -172,17 +171,6 @@ export default function VisitorGraph({
...
       setIsRealtimeSilentUpdate({ topStats: true, mainGraph: true })
-      queryClient.invalidateQueries({
-        predicate: ({ queryKey }) => {
-          const realtimeTopStatsOrMainGraphQuery =
-            ['top-stats', 'main-graph'].includes(queryKey[0] as string) &&
-            typeof queryKey[1] === 'object' &&
-            (queryKey[1] as { dashboardState?: DashboardState })?.dashboardState
-              ?.period === DashboardPeriod.realtime
-
-          return realtimeTopStatsOrMainGraphQuery
```
apata, line 183: issue, blocking: This is causing double fetching, which is a bug, but something like it is needed to invalidate the cached queries for other realtime metrics on tick, otherwise the green realtime indicator becomes misleading. 

**On realtime dashboard:**
- **00**: Tick, request R_visitors is made for visitors metric
- **20**: user selects Total pageviews, request R_total_pageviews is made and served
- **25**: user selects Visitors, cached R_visitors is still fresh, served to user
- **30**: Tick, R_visitors_2 is fetched and served to user
- **35**: user selects Total pageviews, **cached R_total_pageviews is served, because it is still fresh until 50**. At this point, realtime indicator shows data is 5 seconds old, which it is not. This data is 15 seconds old.

apata: This removes the double fetch without running into the issue above: https://github.com/plausible/analytics/compare/fix-realtime-double-fetch...fix-realtime-double-fetch-invalidate?expand=1

RobertJoonas: Good catch! Fixed now with a `refetchType: 'none'` option, which is just telling invalidateQueries to invalidate, but do not refetch. https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientinvalidatequeries

apata: Nice, I wasn't aware of that option! I like the solution that I shared a little better: less code, less effect dependencies. Did you give that a go?

RobertJoonas: https://github.com/plausible/analytics/pull/6269/commits/97a2631939696ea5f6d93b6074303e7c68c194c3

apata (approved):

PR #6271, opened 2026-04-22
Add implicit wildcard matches to exploration suggestions

zoldar:
### Changes

This PR adds experimental implementation of implicit wildcard pathnames support to exploration funnel.

The wildcard suggestions are automatically generated from underlying suggested transitions. The paths are generated segment by segment. The patterns which have visitors counts exactly equal to their non-wildcard counterparts are eliminated to only show the actual wildcards that group more than one path.

~There's still a very small discrepancy between visitors counts for wildcard patterns between next step suggestions and funnel computation. I'm still not sure where it's coming from (the way the visitors are computed is very differernt, though the outcome _should be_ the same). We can look deeper into this, provided this feature will be needed at all.~

~The idea is for this "prototype" to be merged for actual early testing, so that we get feedback whether it's worth having at all.~

In the mean time, on the basis of this change, we will work on adding support for pageview and custom goals in funnel exploration.

Other changes made within the scope of this PR:

- pathnames are normalised, so that the trailing slashes are removed from them. With this, pathnames like "/:dashboard" and "/:dashboard/" are treated as the same path
- we no longer prepend pageview steps with "Visit " - it's label is a plain path
- implicit wilcards are customised to be different from pageview goal patterns - the way they are computed differs from a pageview goal like "/path*" in that is does not count things like "/path-to-some-other-page".

### Tests

### DIFF
--- a/extra/lib/plausible/stats/exploration.ex
+++ b/extra/lib/plausible/stats/exploration.ex
@@ -411,7 +411,7 @@ defmodule Plausible.Stats.Exploration do
         |> String.trim_trailing("...")
         |> Regex.escape()
 
-      pattern = "^#{escaped}($|/.*)$"
+      pattern = "^#{escaped}(/.*)?$"
 
       dynamic(
         [s],
--- a/extra/lib/plausible/stats/exploration.ex
+++ b/extra/lib/plausible/stats/exploration.ex
@@ -8,20 +8,34 @@ defmodule Plausible.Stats.Exploration do
 
     @type t() :: %__MODULE__{}
 
-    @derive {Jason.Encoder, only: [:name, :pathname, :label]}
-    defstruct [:name, :pathname, :label]
+    @derive {Jason.Encoder, only: [:name, :pathname, :label, :include_subpaths, :subpaths_count]}
+    defstruct name: nil, pathname: nil, label: nil, include_subpaths: false, subpaths_count: 0
 
     @spec from(map()) :: t()
     def from(step) do
-      new(step.name, step.pathname)
+      new(step.name, step.pathname, step.include_subpaths, step.subpaths_count)
     end
 
-    @spec new(String.t(), String.t()) :: t()
-    def new(name, pathname) do
+    @spec new(String.t(), String.t(), boolean(), non_neg_integer()) :: t()
+    def new(name, pathname, include_subpaths \\ false, subpaths_count \\ 0) do
+      label =
+        cond do
+          name != "pageview" ->
+            name <> " " <> pathname
+
+          include_subpaths ->
+            pathname <> " (" <> to_string(subpaths_count) <> " pages)"
+
+          true ->
+            pathname
+        end
+
       %__MODULE__{
-        label: if(name == "pageview", do: "", else: name <> " ") <> pathname,
+        label: label,
         name: name,
-        pathname: pathname
+        pathname: pathname,
+        include_subpaths: include_subpaths,
+        subpaths_count: subpaths_count
       }
     end
   end
@@ -32,7 +46,6 @@ defmodule Plausible.Stats.Exploration do
 
   alias Plausible.ClickhouseRepo
   alias Plausible.Stats.Base
-  alias Plausible.Stats.Filters
   alias Plausible.Stats.Query
 
   @type journey() :: [Journey.Step.t()]
@@ -176,7 +189,9 @@ defmodule Plausible.Stats.Exploration do
         select: %{
           name: selected_as(field(s, ^next_name), :name),
           pathname: selected_as(field(s, ^next_pathname), :pathname),
-          visitors: scale_sample(fragment("uniq(?)", s.user_id))
+          visitors: scale_sample(fragment("uniq(?)", s.user_id)),
+          include_subpaths: type(^false, :boolean),
+          subpaths_count: 0
         },
         group_by: [selected_as(:name), selected_as(:pathname)]
       )
@@ -195,13 +210,13 @@ defmodule Plausible.Stats.Exploration do
         join: pname in fragment(@wildcard_array_join, em.pathname),
         on: true,
         hints: "ARRAY",
-        where: selected_as(:pathname) != "*" and selected_as(:pathname) != "/*",
         where: em.name == "pageview",
+        where: selected_as(:pathname) != "" and selected_as(:pathname) != "/",
         select: %{
           name: em.name,
-          pathname: selected_as(fragment("concat(?, '*')", pname), :pathname),
-          pathname_plain: selected_as(fragment("?", pname), :pathname_plain),
-          visitors: selected_as(fragment("sum(?)", em.visitors), :visitors)
+          pathname: selected_as(fragment("?", pname), :pathname),
+          visitors: selected_as(fragment("sum(?)", em.visitors), :visitors),
+          unique_paths: fragment("uniq(?)", em.pathname)
         },
         group_by: [em.name, fragment("?", pname)]
       )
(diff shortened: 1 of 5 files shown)


extra/lib/plausible/stats/exploration.ex
```diff
@@ -162,80 +183,140 @@ defmodule Plausible.Stats.Exploration do
...
-    end)
+
+    q_wildcard_filtered_matches =
+      from(wm in subquery(q_wildcard_matches),
+        left_join: emx in subquery(q_exact_matches),
+        on: emx.name == wm.name and emx.pathname == wm.pathname,
+        where: wm.unique_paths > 1 and (is_nil(emx.name) or wm.visitors != emx.visitors),
+        select: %{
+          name: wm.name,
+          pathname: fragment("concat(?, '...')", wm.pathname),
```
aerosol: Subject to @sanne-san's judgement, but maybe some form of [ellipsis](https://en.wikipedia.org/wiki/Ellipsis) glyph would be good to render?

extra/lib/plausible/stats/exploration.ex
```diff
@@ -322,17 +408,32 @@ defmodule Plausible.Stats.Exploration do
...
   defp step_condition(step, count) do
-    dynamic(
-      [s],
-      field(s, ^:"name#{count}") == ^step.name and
-        field(s, ^:"pathname#{count}") == ^step.pathname
-    )
+    if step.include_subpaths do
+      escaped =
+        step.pathname
+        |> String.trim_trailing("...")
```
aerosol: ditto ellipsis symbol or some other visual indicator that is less likely to be part of the actual URL - not a blocker right now

Image

extra/lib/plausible/stats/exploration.ex
```diff
@@ -162,80 +183,140 @@ defmodule Plausible.Stats.Exploration do
...
+        join: pname in fragment(@wildcard_array_join, em.pathname),
+        on: true,
+        hints: "ARRAY",
+        where: em.name == "pageview",
+        where: selected_as(:pathname) != "" and selected_as(:pathname) != "/",
+        select: %{
+          name: em.name,
+          pathname: selected_as(fragment("?", pname), :pathname),
+          visitors: selected_as(fragment("sum(?)", em.visitors), :visitors),
+          unique_paths: fragment("uniqExact(?)", em.pathname)
```
aerosol: > The uniqExact function uses more memory than uniq, because the size of the state has unbounded growth as the number of different values increases. Use the uniqExact function if you absolutely need an exact result. Otherwise use the [uniq](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/uniq) function.

Just double checking that was a deliberate decision and the cost of `uniqExact` was evaluated

extra/lib/plausible/stats/exploration.ex
```diff
@@ -162,80 +183,140 @@ defmodule Plausible.Stats.Exploration do
...
-                ),
-                :next_label
-              ),
-            name: selected_as(field(s, ^next_name), :next_name),
-            pathname: selected_as(field(s, ^next_pathname), :next_pathname)
-          },
-          visitors: selected_as(scale_sample(fragment("uniq(?)", s.user_id)), :count)
+          name: selected_as(field(s, ^next_name), :name),
+          pathname: selected_as(field(s, ^next_pathname), :pathname),
+          visitors: scale_sample(fragment("uniqExact(?)", s.user_id)),
```
aerosol: ditto - `uniqExact` conscious use - won't the numbers be slighlty inaccurate anyway, if sampling is applied?

extra/lib/plausible/stats/exploration.ex
```diff
@@ -8,20 +8,34 @@ defmodule Plausible.Stats.Exploration do
...
     @spec from(map()) :: t()
     def from(step) do
-      new(step.name, step.pathname)
+      new(step.name, step.pathname, step.include_subpaths, step.subpaths_count)
     end
 
-    @spec new(String.t(), String.t()) :: t()
-    def new(name, pathname) do
+    @spec new(String.t(), String.t(), boolean(), non_neg_integer()) :: t()
+    def new(name, pathname, include_subpaths \\ false, subpaths_count \\ 0) do
```
aerosol: could add guards for args?

test/plausible/stats/exploration_test.exs
```diff
@@ -814,5 +844,95 @@ defmodule Plausible.Stats.ExplorationTest do
...
+          pathname: "/a...",
+          include_subpaths: true,
+          subpaths_count: 4
+        }
+      ]
+
+      assert {:ok, [step1]} = Exploration.journey_funnel(query, journey)
+
+      assert step1.step.label == "/a... (4 pages)"
+      assert step1.visitors == 10
```
aerosol: but the true number of uniqe visitors is 5, are we overcounting for the sake of consistency between steps and funnel?

extra/lib/plausible/stats/exploration.ex
```diff
@@ -162,80 +183,140 @@ defmodule Plausible.Stats.Exploration do
     next_name = :"name#{next_step_idx}"
```
aerosol, line 196: let's add a maximum steps length guard here, to prevent atom overflow vector from being exploited

extra/lib/plausible/stats/exploration.ex
```diff
@@ -322,17 +408,32 @@ defmodule Plausible.Stats.Exploration do
   end
 
   defp step_condition(step, count) do
```
aerosol: let's add a maximum guard here for count to prevent atom build up

lib/plausible_web/controllers/api/stats_controller.ex
aerosol, line 222: Not related to this PR in particular but let's add some limit enforcement here with regards to how many journey steps are even parsed (we could expose and reference it through `Exploration` module and fix up the hardcoded 20's elsewhere). Limit breach could be treated with either silent drop via `Enum.take` or bad request throw.

aerosol (approved):
Overall, incredible work. Approving since everything's private anyway.

extra/lib/plausible/stats/exploration.ex
zoldar: Yes, this is a debliberate decision. Without it, the inaccuracies start to appear at pretty small time windows at counts and it makes for a bad experience IMHO.

extra/lib/plausible/stats/exploration.ex
zoldar: See https://github.com/plausible/analytics/pull/6271#discussion_r3145862645

test/plausible/stats/exploration_test.exs
zoldar: Okay, good question. My assumption was that subpaths under the wilcard are still treated as belonging to distinct journeys, even if made by the same user.

And yeah, that's evidently different from, for instance, pageview goals like "VIsit /page*" in the standard funnels.

I'll have to check how feasible it is to truly count unique visitors treating the pattern/wildcard as the same journey despite different concrete paths underneath.

extra/lib/plausible/stats/exploration.ex
zoldar: Changed suffix to ellipsis glyph in https://github.com/plausible/analytics/pull/6271/commits/ea428a5b6782c42a7eef65b80f8d429fb3e53f1b, however, for the font we use at least, the difference is hard to tell 😅

extra/lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6271/commits/ea428a5b6782c42a7eef65b80f8d429fb3e53f1b

extra/lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6271/commits/8be4bfb163c8c85c23bcce1009450fa0ad85cd7a

lib/plausible_web/controllers/api/stats_controller.ex
zoldar: I've added guards and explicit error handling on API level for too long journeys in https://github.com/plausible/analytics/pull/6271/commits/7a6426d55806ee3b6726f0c24527c21d4ff940fc. The frontend is still unaware of that limitation though. I'll make a card for tackling it in a follow-up.

extra/lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6271/commits/7a6426d55806ee3b6726f0c24527c21d4ff940fc

extra/lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6271/commits/7a6426d55806ee3b6726f0c24527c21d4ff940fc

test/plausible/stats/exploration_test.exs
zoldar: We have decided to modify the aggregation so that wildcards always count unique user visits.

PR #6273, opened 2026-04-22
Exploration: lazily load 2-step funnel on first render

aerosol:
### Changes

To make the exploration more encouraging to play with, we'll be preloading some funnel guesswork in "Explore" section by default. Defensively, we'll also query a smaller number of step candidates while building it.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -52,11 +52,11 @@ function fetchNextWithFunnel(
   )
 }
 
-function fetchSuggestedJourney(site, dashboardState) {
+function fetchInterestingFunnel(site, dashboardState) {
   return api.post(
     url.apiPath(site, '/exploration/interesting-funnel'),
     dashboardState,
-    {}
+    { max_steps: 2, max_candidates: 6 }
   )
 }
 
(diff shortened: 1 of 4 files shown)


assets/js/dashboard/extra/exploration.js
```diff
@@ -336,14 +292,65 @@ export function FunnelExploration() {
     prevDirectionRef.current = direction
     prevDashboardStateRef.current = dashboardState
 
-    const includeFunnel = journeyChanged && steps.length > 0
+    let cancelled = false
+
+    if (!preloadFiredRef.current) {
```
zoldar, line 297: Given the preloaded funnel should is supposed to be only run once on loading the exploration component, perhaps it would make sense to move it to a separate `useEffect(.., [])` (with empty deps)? It then also wouldn't be necessary to track whether it was fired with an extra ref.

assets/js/dashboard/extra/exploration.js
```diff
@@ -52,11 +52,11 @@ function fetchNextWithFunnel(
   )
 }
 
-function fetchSuggestedJourney(site, dashboardState) {
+function fetchInterestingFunnel(site, dashboardState) {
   return api.post(
     url.apiPath(site, '/exploration/interesting-funnel'),
     dashboardState,
-    {}
+    { max_steps: 2, max_candidates: 6 }
```
zoldar, line 59: JFTR, at least for the current implementation(s) of the exploration, whether we fetch 6 or 20 candidates does not really affect query performance that much.

zoldar (approved):
Nice work civilising the API with those opts.

assets/js/dashboard/extra/exploration.js
aerosol: ooh didn't know it works like that, checking

assets/js/dashboard/extra/exploration.js
aerosol: even for extreme sites?

zoldar: Yeah, I think so. The majority of work is done in the 2 innermost queries where ordering must happen on the whole resultset. Also, sort of counter-intuitively, the longer the funnel, the more performant the query because more of the transitions are filtered out early. Selecting from the sorted set is a tiny factor (that is unless you fetch by the thousands where the payload size/serialization and deserialization overhead might be more relevant)

assets/js/dashboard/extra/exploration.js
zoldar: Yeah, scratch that, this is probably better off to be left as is. However, maybe swapping ref for a straight boolean, like in https://react.dev/learn/you-might-not-need-an-effect#initializing-the-application

aerosol: Module-level boolean won't work because it'll break the state when switching tabs (e.g. Exploration -> Goals -> Exploration). On first render it's fine, but when you're back at this tab again, the remount is done based on useState defaults (empty steps, empty funnel) withj the preload bool already set to true

PR #6274, opened 2026-04-22
Cache breakdown responses in details modals

RobertJoonas:
### Changes

* Improve getStaleTime logic (and add tests):
  * 7d, 28d, 6mo, 12mo are now always considered historical as they exclude the current day
  * keep the same stale time for realtime (i.e. 30s)
  * keep the same stale time for historical periods (i.e. 12h)
  * for ongoing periods (those including `now`), give a much shorter stale time of 5min if 'day' (or shorter) graph interval is supported, or 15min otherwise. Use the `validIntervals` logic for determining that, which also takes comparison period into account.
* plug staleTime into `usePaginatedGetAPI`, enabling cache for all breakdown modals, including Google Search Terms. Note that the current prod behaviour serves results from cache as well, but due to `staleTime: 0` those results are instantly considered stale and a background re-fetch is triggered. Therefore, in terms of UX, nothing changes, other than not flashing a loading spinner on cache-hit.
* add two tests for breakdown modal cache actually working

### Tests

### Changelog
- [x] Does not need a changelog update

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/filtering/segments.ts
+++ b/assets/js/dashboard/filtering/segments.ts
@@ -34,9 +34,9 @@ export type SavedSegment = {
   id: number
   name: string
   type: SegmentType
-  /** datetime in site timezone, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
   inserted_at: string
-  /** datetime in site timezone, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
   updated_at: string
 } & SegmentOwnership
 
--- a/assets/js/dashboard/filtering/segments.ts
+++ b/assets/js/dashboard/filtering/segments.ts
@@ -34,9 +34,9 @@ export type SavedSegment = {
   id: number
   name: string
   type: SegmentType
-  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26T10:00:00 */
   inserted_at: string
-  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26T10:00:00 */
   updated_at: string
 } & SegmentOwnership
 
--- a/assets/js/dashboard/hooks/api-client.test.ts
+++ b/assets/js/dashboard/hooks/api-client.test.ts
@@ -1,6 +1,5 @@
-import { DEFAULT_SITE } from '../../../test-utils/app-context-providers'
 import { ComparisonMode, DashboardPeriod } from '../dashboard-time-periods'
-import { formatISO, nowForSite } from '../util/date'
+import { formatISO, utcNow } from '../util/date'
 import {
   CACHE_TTL_HISTORICAL,
   CACHE_TTL_LONG_ONGOING,
@@ -9,12 +8,18 @@ import {
   getStaleTime
 } from './api-client'
 
-const site = DEFAULT_SITE
-const today = nowForSite(site)
-const yesterday = nowForSite(site).subtract(1, 'day')
+const today = utcNow()
+const yesterday = utcNow().subtract(1, 'day')
 
 const noComparison = { comparison: null, compare_from: null, compare_to: null }
-const base = { site, date: null, from: null, to: null, ...noComparison }
+const base = {
+  siteStatsBegin: '',
+  siteTimezoneOffset: 0,
+  date: null,
+  from: null,
+  to: null,
+  ...noComparison
+}
 
 describe(`${getStaleTime.name}`, () => {
   describe('realtime periods', () => {
@@ -100,14 +105,10 @@ describe(`${getStaleTime.name}`, () => {
     })
 
     it('for all time period when stats begin recently', () => {
-      const siteWithRecentStats = {
-        ...DEFAULT_SITE,
-        statsBegin: formatISO(yesterday)
-      }
       expect(
         getStaleTime({
           ...base,
-          site: siteWithRecentStats,
+          siteStatsBegin: formatISO(yesterday),
           period: DashboardPeriod.all
         })
       ).toBe(CACHE_TTL_SHORT_ONGOING)
@@ -140,11 +141,10 @@ describe(`${getStaleTime.name}`, () => {
     })
 
     it('for all time period when stats begin over 12 months ago', () => {
-      const siteWithOldStats = { ...DEFAULT_SITE, statsBegin: '2020-01-01' }
       expect(
         getStaleTime({
           ...base,
-          site: siteWithOldStats,
+          siteStatsBegin: '2020-01-01',
           period: DashboardPeriod.all
         })
       ).toBe(CACHE_TTL_LONG_ONGOING)
--- a/assets/js/dashboard/hooks/api-client.test.ts
+++ b/assets/js/dashboard/hooks/api-client.test.ts
@@ -82,7 +82,7 @@ describe(`${getStaleTime.name}`, () => {
       ).toBe(CACHE_TTL_SHORT_ONGOING)
     })
 
-    it('for period=year and date=yesterday', () => {
+    it('for period=year and date=today', () => {
       expect(
         getStaleTime({ ...base, period: DashboardPeriod.year, date: today })
       ).toBe(CACHE_TTL_SHORT_ONGOING)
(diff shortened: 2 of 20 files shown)


assets/js/dashboard/hooks/api-client.test.ts
```diff
@@ -0,0 +1,153 @@
...
+const site = DEFAULT_SITE
+const today = nowForSite(site)
+const yesterday = nowForSite(site).subtract(1, 'day')
+
+const noComparison = { comparison: null, compare_from: null, compare_to: null }
+const base = { site, date: null, from: null, to: null, ...noComparison }
+
+describe(`${getStaleTime.name}`, () => {
+  describe('realtime periods', () => {
+    it('for realtime', () => {
```
apata: meganitpick, non-blocking: Can we refactor `it(` to `test(`? `it` works better if there's a verb right after

assets/js/dashboard/hooks/api-client.test.ts
```diff
@@ -0,0 +1,153 @@
...
+import { formatISO, nowForSite } from '../util/date'
+import {
+  CACHE_TTL_HISTORICAL,
+  CACHE_TTL_LONG_ONGOING,
+  CACHE_TTL_REALTIME,
+  CACHE_TTL_SHORT_ONGOING,
+  getStaleTime
+} from './api-client'
+
+const site = DEFAULT_SITE
```
apata: suggestion, non-blocking: If we refactor `getStaleTime` not to depend on arg named `site` but directly on the two properties from the site, it makes the test depend on less things. That way, we'd not need to import or fake a full site here, just the properties `offset` and `statsBegin`.

assets/js/dashboard/hooks/api-client.ts
```diff
@@ -108,59 +125,40 @@ export const cleanToPageOne = <
...
-    case DashboardPeriod.all:
-    default:
-      // err on the side of less caching,
-      // to avoid the user refresheshing
-      return 15 * 60 * 1000
+  const availableIntervals = validIntervals(
+    Object.fromEntries(
+      Object.entries(props).filter(([k, _v]) => k !== 'date')
+    ) as GetIntervalProps
+  )
```
apata: issue, blocking: The cleaner way to exclude `date` would be to destructure `const {date, ...rest} = props`, `const availableIntervals = validIntervals(rest)`, but we don't actually have to do it, because the function ignores the date prop if present. It doesn't make use of it. Checking locally, the line `const availableIntervals = validIntervals(props)` is valid according to typescript compiler.

assets/js/dashboard/stats/graph/graph-interval-context.tsx
```diff
@@ -21,9 +21,10 @@ export function GraphIntervalProvider({
 }) {
   const site = useSiteContext()
   const { dashboardState } = useDashboardStateContext()
+  const intervalStorageKey = `interval__${dashboardState.period}__${site.domain}`
 
   const { selectedInterval, onIntervalClick, availableIntervals } =
-    useStoredInterval({
+    useStoredInterval(intervalStorageKey, {
```
apata: question, non-blocking: Why can't the hook contain the logic to derive the storage key? Seems totally within its limited responsibilities.

assets/js/dashboard/hooks/api-client.test.ts
RobertJoonas: Makes sense. https://github.com/plausible/analytics/pull/6274/commits/bb48cee20c6bcdb7166f6e3b1ec6e338244981e4

assets/js/dashboard/hooks/api-client.test.ts
RobertJoonas: Agreed. I've flattened `site` into `siteTimezoneOffset` and `siteStatsBegin` in https://github.com/plausible/analytics/pull/6274/commits/efc0da0322af7ab815a2aacc509281d944fd2d25.

Refactoring this, I also found a bug in `validIntervalsForAllTimePeriod` function with `site.statsBegin` parsing. (Introduced that myself just recently when working with intervals.) it's currently using the `dateForSite` utility which is meant to take a naive UTC datetime and shift it according to the offset to return the local date. The `site.statsBegin` is a date string (not datetime) in site timezone already -- using `dateForSite` can make it a different date, but it should never do that. I've fixed that with:

```js
parseUTCDate(siteStatsBegin).utcOffset(siteTimezoneOffset / 60, true)
```

where the `true` argument is `keepLocalTime`. Also added test coverage in `intervals.test.ts`.

Btw, digging deeper into where `dateForSite` is called, I also noticed an inaccurate comment -- segment timestamps are not local, but naive UTC. Fixed that as well in https://github.com/plausible/analytics/pull/6274/commits/a53c1bda717272426ab2ee5948519f434cfe8d89

assets/js/dashboard/hooks/api-client.ts
RobertJoonas: Much cleaner. Thanks! https://github.com/plausible/analytics/pull/6274/commits/1308651a67dc005a605a3f4a167d22aba4082b7f

assets/js/dashboard/stats/graph/graph-interval-context.tsx
RobertJoonas: Changed that as well in https://github.com/plausible/analytics/pull/6274/commits/efc0da0322af7ab815a2aacc509281d944fd2d25

assets/js/dashboard/filtering/segments.ts
```diff
@@ -34,9 +34,9 @@ export type SavedSegment = {
   id: number
   name: string
   type: SegmentType
-  /** datetime in site timezone, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
   inserted_at: string
-  /** datetime in site timezone, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
```
apata:
```suggestion
  /** naive UTC timestamp, example 2025-02-26T10:00:00 */
```

assets/js/dashboard/filtering/segments.ts
```diff
@@ -34,9 +34,9 @@ export type SavedSegment = {
   id: number
   name: string
   type: SegmentType
-  /** datetime in site timezone, example 2025-02-26 10:00:00 */
+  /** naive UTC timestamp, example 2025-02-26 10:00:00 */
```
apata:
```suggestion
  /** naive UTC timestamp, example 2025-02-26T10:00:00 */
```

assets/js/dashboard/hooks/api-client.test.ts
apata: Well spotted about the segments! 
Why do we parse a local date such as '2026-04-27' (for a site with stats beginning today) with `parseUTCDate`? That seems misleading.

RobertJoonas: Tbh I didn't find a cleaner way of writing it. For parsing a date from string, we have the following options:

* simply inline `dayjs(dateString)` - parses the Dayjs object in local browser timezone. 
* the already existing `parseNaiveDate(dateString)` utility - sounds correct at callsite, but if you look at the implementation, it's just calling `dayjs(dateString)` under the hood
* parseUTCDate (the one I went with)

In all cases, the initial timezone information is discarded with the second step -- `.utcOffset(siteTimezoneOffset / 60, true)`. So functionally there's no difference. However, I understand your confusion. It does read a bit strange.

I personally dislike the `dayjs()` approach because converting Dayjs objects into local browser timezone is something we never want. We should only be working with either UTC or site.timezone. And currently, there's no real "naive" datetime representation in the project.

Speaking of which, the `parseNaiveDate` function is wrong and misleading:

```js
> parseNaiveDate('2021-07-01').format()
2021-07-01T00:00:00+01:00
> parseNaiveDate('2021-07-01T00:00:00-05:00')
2021-07-01T06:00:00+01:00
```

assets/js/dashboard/filtering/segments.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6274/commits/bbabfcdf35db1af12a5fb39ff93566ca2ebcf9f2

assets/js/dashboard/filtering/segments.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6274/commits/bbabfcdf35db1af12a5fb39ff93566ca2ebcf9f2

apata (approved):

assets/js/dashboard/hooks/api-client.test.ts
apata: Thanks for looking deeper into it! I also can't see a way forward that isn't weird.

PR #6276, opened 2026-04-27
Improve explore tab UI

sanne-san:
### Changes

- Keep showing the full candidate list when a step is selected
- Show conversion rate for each step
- Update title
- Add total conversion rate and dropoff rate
- Add refresh button
- Add path connectors
- Change starting/end point toggle into an inline dropdown
- Add color coding for pageviews vs custom events
- Remove "Visit" prefix from pageview steps
- Update columns and steps visually

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -778,7 +778,7 @@ export function FunnelExploration() {
               )}
             </button>
           ) : (
-            <Tooltip info="Deselect all">
+            <Tooltip info={<span className="whitespace-nowrap">Deselect all</span>}>
               <button
                 onClick={handleReset}
                 className={`${popover.toggleButton.classNames.rounded} ${popover.toggleButton.classNames.outline} justify-center !h-7 px-1.5`}
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -9,8 +9,8 @@ import {
   numberShortFormatter,
   numberLongFormatter
 } from '../util/number-formatter'
-import { RefreshIcon } from '../components/icons'
-import { ChevronUpDownIcon } from '@heroicons/react/20/solid'
+import { RefreshIcon, CursorIcon } from '../components/icons'
+import { ChevronUpDownIcon, ChevronRightIcon } from '@heroicons/react/20/solid'
 import { popover } from '../components/popover'
 
 const PAGE_FILTER_KEYS = ['page', 'entry_page', 'exit_page']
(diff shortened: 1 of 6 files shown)


aerosol:
The visitor counts are being dragged from one step to another before ending up with final value.

https://github.com/user-attachments/assets/b20094b0-5093-493c-b7dc-9271e723bcd1

assets/js/dashboard/extra/exploration.js
```diff
@@ -277,23 +547,64 @@ export function FunnelExploration() {
...
+    setProvisionalFunnelEntries({})
+    setFrozenColumnResults({})
+  }
+
   function handleDirectionSelect(nextDirection) {
     if (nextDirection === direction) return
 
+    journeyVersionRef.current++
     setDirection(nextDirection)
-    setSteps(steps.toReversed())
```
aerosol, line 290: Is this deliberate to start over instead of reversing the journey?

assets/js/dashboard/extra/exploration.js
```diff
@@ -321,8 +632,34 @@ export function FunnelExploration() {
...
             setFunnel(response)
+            // Backfill candidate lists for each preloaded (selected) column
+            // so the user can immediately switch options without first
+            // having to clear and re-search. We capture the current journey
+            // version so any of these fetches that resolve after the user
+            // has navigated away can be safely ignored.
+            const journeyVersion = journeyVersionRef.current
+            preloadedSteps.forEach((_, idx) => {
+              const prefix = preloadedSteps.slice(0, idx)
+              fetchNextWithFunnel(
```
aerosol: @zoldar this is making essentially N+1 calls to `next-with-funnel`

Image

aerosol:
The paths go bonkers if you change e.g. query range on an already selected journey

image

assets/js/dashboard/extra/exploration.js
```diff
@@ -77,6 +81,187 @@ function isSameStep(step, otherStep) {
...
+        clipY,
+        clipHeight
+      })
+    }
+
+    recalculate()
+
+    const observer = new ResizeObserver(recalculate)
+    observer.observe(container)
+    window.addEventListener('resize', recalculate)
```
aerosol: why is this even here? makes no difference

assets/js/dashboard/extra/exploration.js
sanne-san: Yes this is deliberate, I think it's better UX to start with a clean slate when switching direction.

zoldar (approved):
We have fixed the more nagging issues, some of them with no too pretty workarounds, but the UI is functional. There's still a bit errant behavior on date range change where the suggestions are not properly cleared. We will address this later on. The component logic will undergo a major revamp when moving to Typescript anyway.

sanne-san:
> We have fixed the more nagging issues, some of them with no too pretty workarounds, but the UI is functional. There's still a bit errant behavior on date range change where the suggestions are not properly cleared. We will address this later on. The component logic will undergo a major revamp when moving to Typescript anyway.

Sounds good, thanks @aerosol @zoldar for getting this over the line 🙏

PR #6284, opened 2026-04-28
Make exploration implicit wildcard togglable

zoldar:
### Changes

- Make implicit wildcard pathnames optional in next step suggestions
- Add a feature flag for selectively disabling it on site level
- Stop filtering out "/" pathname from suggestions

### Tests


aerosol (approved):
🪨

PR #6286, opened 2026-04-28
Pass segment timestamps in site timezone

apata:
### Changes

The frontend doesn't have enough information to render the UTC timestamps sent to it correctly. 

Site timezone offset isn't enough, because the offset is dynamic. It depends on the time. 

For example, for a site with the timezone Tallinn/Estonia, the UTC datetime `2026-01-01T00:00:00` must be rendered as `2026-01-01 02:00:00` whereas `2026-04-01T00:00:00` must be rendered as `2026-04-01 03:00:00`. 

This problem manifests with segment related timestamps: if the user on the Tallinn/Estonia site creates a segment in January and looks at it in April, the shown timestamp will be off by one hour, because we use the offset right now.

This PR fixes it by converting the dates on the backend. 
The alternative would be to provide the timezone locale to the FE and use that. The issue with that is the tz database falling out of sync between the BE and the FE. Also, we already convert some dates on the backend (in queries with the time dimension). 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to this project will be documented in this file.
 - Fixed comparison tooltip in the top pages report missing date labels
 - Fixed top bar not scrolling horizontally on mobile
 - Fixed incline/decline percentages not showing in top stats in comparison mode
+- Fixed issue with timestamps being rendered incorrectly in segment menus and modals for some timezones
 
 ## v3.2.0 - 2026-01-16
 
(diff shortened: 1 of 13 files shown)


RobertJoonas (commented):
Looks good to me! But could we also add a test to actually assert on the timestamps being shifted into site timezone when passed to the frontend?

ukutaht (approved):

RobertJoonas (approved):

PR #6293, opened 2026-04-29
Add highlighted vertical line to graph (with tooltip sticky to the line)

apata:
### Changes

- Adds indicator line for selected x-value on graph.
- Makes tooltip anchor to the right* of the indicator line, aligned by top edge to top of graph.
  - (*) when it fits to the right. Otherwise to the left. If it fits neither right or left, it is clamped to the right edge of the graph.
- Touch device tooltip change:
On touch devices, a tap summons the tooltip at tapped x-value (_pre-existing logic_). When the tooltip has been summoned, dragging on the graph horizontally moves the tooltip (**new functionality in this PR**). Panning vertically dismisses the tooltip (_pre-existing logic_). Dragging horizontally without the tooltip being summoned has no effect (_pre-existing logic_). 

There's a pre-existing bug that clicks on menus don't dismiss the tooltip on mobile: this will be addressed separately.

### Tests
- [x] Manually tested on multiple browsers and devices

### Changelog

### Documentation

### Dark mode
- [x] Tested in both dark and light mode

### DIFF
--- a/assets/js/dashboard/stats/graph/main-graph.tsx
+++ b/assets/js/dashboard/stats/graph/main-graph.tsx
@@ -101,6 +101,16 @@ export const MainGraph = ({
     setTooltip(initialTooltipState)
   }, [data])
 
+  useEffect(() => {
+    const onPointerCancel = (e: PointerEvent) => {
+      if (e.pointerType === 'touch') {
+        setTooltip(initialTooltipState)
+      }
+    }
+    window.addEventListener('pointercancel', onPointerCancel)
+    return () => window.removeEventListener('pointercancel', onPointerCancel)
+  }, [])
+
   const {
     remappedData,
     yMax,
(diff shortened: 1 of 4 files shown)


ukutaht (approved):
Looks good to me

assets/js/dashboard/stats/graph/main-graph.tsx
```diff
@@ -249,22 +263,40 @@ export const MainGraph = ({
...
+          } else if (
+            now - panGestureStartTimeRef.current >=
+            HORIZONTAL_PAN_DELAY_MS
+          ) {
+            setTooltip({
+              selectedIndex: closestPoint.index,
+              x: closestPoint.x,
+              y: 0,
+              persistent: true
+            })
```
RobertJoonas, line 280: Nitpick (non-blocking): Looks like the y position of the tooltip is always a fixed 0 (and I think that makes perfect sense with the vertical line!) Did you consider removing the y value from the state?

RobertJoonas (approved):
The UX feels amazing and looks good code-wise too. Nice work! 🙌

assets/js/dashboard/stats/graph/main-graph.tsx
apata: Thanks and well spotted! I did consider it, but then decided to keep the (x, y) positioning API. With just x, it's slightly less self-explanatory. However, now that you pointed it out, I can see how it may be confusing and a source of bugs. Suggestion applied! https://github.com/plausible/analytics/pull/6293/commits/c8d8fb0a5c798cda6839de65b9e9ef1e43bfa010

PR #6294, opened 2026-04-29
Adjust mobile tooltip layout to make it fit on smaller screens

sanne-san:


apata:
Closing because we opted to run with the bigger text tooltip on mobile after all.

PR #6298, opened 2026-04-30
Standardize all filter and segment modals styles

sanne-san:
### Changes

- Add a shared ModalLayout/ModalFooter to give every modal the same header, body spacing, and button alignment
- Add a shared Button component (matching the Phoenix variant) and switch all modal buttons to it for consistent sizing and theming
- Align footer buttons to the right in all modal footers
- Restructure the segment details modal: fixed "Segment details" title, with the segment name, type, and authorship shown inline in the body
- Update the delete segment modal with a clearer title and a confirmation message
- Make filter pills better visible inside modals with a gray background
- Add a Cancel button to the filter modal when no filters are applied
- Refine button theme borders so ghost, yellow, and icon variants line up at the same height as bordered buttons, and fix the visible border on disabled primary buttons

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/button.tsx
+++ b/assets/js/dashboard/components/button.tsx
@@ -3,7 +3,8 @@ import classNames from 'classnames'
 
 /**
  * Themes and sizes are kept in sync with the Phoenix `button` component in
- * `lib/plausible_web/components/generic.ex`. Update both together.
+ * `lib/plausible_web/components/generic.ex`. The actual Tailwind classes live
+ * in `assets/css/app.css` (.btn-base, .btn-{sm,md}, .btn-theme-*).
  */
 
 export type ButtonTheme =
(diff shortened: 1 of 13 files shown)


assets/js/dashboard/components/button.tsx
```diff
@@ -0,0 +1,69 @@
...
+  yellow:
+    'border border-yellow-600/90 bg-yellow-600/90 text-white hover:bg-yellow-600 focus-visible:outline-yellow-600 disabled:border-yellow-400/60 disabled:bg-yellow-400/60 disabled:dark:border-yellow-600/30 disabled:dark:bg-yellow-600/30 disabled:dark:text-white/35',
+  danger:
+    'border border-gray-300 dark:border-gray-800 text-red-600 bg-white dark:bg-gray-800 hover:text-red-700 dark:hover:text-red-400 dark:text-red-500 active:text-red-800 disabled:text-red-700/40 disabled:hover:shadow-none dark:disabled:text-red-500/35 dark:disabled:bg-gray-800',
+  ghost:
+    'border border-transparent text-gray-700 dark:text-gray-300 hover:text-gray-900 dark:hover:text-gray-100 hover:bg-gray-100 dark:hover:bg-gray-800 hover:border-gray-100 dark:hover:border-gray-800 disabled:text-gray-500 disabled:dark:text-gray-600 disabled:hover:bg-transparent disabled:hover:border-transparent',
+  icon: 'border border-transparent text-gray-400 dark:text-gray-400 hover:text-gray-900 dark:hover:text-gray-100'
+}
```
apata: suggestion, non-blocking: if these must be in sync between the server and the FE, we can declare custom classes https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes. I believe it's the one appropriate use case for them.

assets/js/dashboard/segments/segment-modals.tsx
```diff
@@ -24,8 +24,12 @@ import { ApiError, get } from '../api'
 import { ErrorPanel } from '../components/error-panel'
 import { useSegmentsContext } from '../filtering/segments-context'
 import { Role, UserContextValue, useUserContext } from '../user-context'
-import { removeFilterButtonClassname } from '../components/remove-filter-button'
 import { useSiteContext } from '../site-context'
+import { Button, buttonClassName } from '../components/button'
+
+const inModalPillClassName = '!shadow-none !bg-gray-100 dark:!bg-gray-750'
```
apata: nitpick, non-blocking: It's a deliberate choice that the pill looks the same in both places it's rendered. The idea was that the user is always given the same information in the same way: if it looks the same, it is the same. 
It is harder to distinguish on the modal's background, true, but to me that's a lesser evil.

apata (approved):
Vast improvement, good job!

assets/js/dashboard/components/button.tsx
sanne-san: Good idea, done!

assets/js/dashboard/segments/segment-modals.tsx
sanne-san: Not sure I agree it's the lesser evil but I reverted it for now. There are plans to improve filtering and segments in the future so let's just leave it as-is til then 👍

PR #6299, opened 2026-04-30
Add custom and pageview goals to exploration suggestions

zoldar:
### Changes

This PR makes exploration UI account for custom and pageview goals in suggestions and exploration funnel queries.

The goals that match the event or pageview exactly, replace it. Pageview goals using a pattern are always rendered as they can neither match exact event or pageview, nor can thay match a wildcard entry, which has a different definition (it applies only to subpaths delimited "/" and not just by prefix).

~The current implementation introduces a significant performance regression. It will be addressed by a subsequent refactoring - either as a part of this PR or in a follow-up.~ Performance seems to be on an acceptable enough level, however testing against higher traffic sites will give a better picture.

image

### Tests

### DIFF
--- a/extra/lib/plausible/stats/exploration.ex
+++ b/extra/lib/plausible/stats/exploration.ex
@@ -479,98 +479,32 @@ defmodule Plausible.Stats.Exploration do
   # single prefix (their exact pathname), so they naturally get
   # subpaths_count = 1 and are only emitted as exact rows.
   @wildcard_array_join """
-  if(? = 'pageview' and ? = '', arrayFold(
+  if(? = 'pageview', arrayFold(
     acc, x -> arrayPushBack(acc, concat(acc[-1], '/', x)), 
     arraySlice(splitByChar('/', ?) AS split_pathname, 2), 
     arraySlice(split_pathname, 1, 1)), [?])
   """
 
-  @goal_pathname_condition """
-  if(? != '', match(?, ?), ? = ?)
-  """
-
-  defp combined_wildcard_query(q_matches, goals, true = _include_wildcard?) do
-    goal_values =
-      Enum.map(goals, fn g ->
-        pathname = g.page_path || ""
-
-        regex_pathname =
-          if String.contains?(pathname, "*") do
-            Filters.Utils.page_regex(pathname)
-          else
-            ""
-          end
-
-        %{
-          label: g.display_name,
-          name: g.event_name || "pageview",
-          pathname: pathname,
-          regex_pathname: regex_pathname
-        }
-      end)
-
-    types = %{label: :string, name: :string, pathname: :string, regex_pathname: :string}
-
+  defp combined_wildcard_query(q_matches, true = _include_wildcard?) do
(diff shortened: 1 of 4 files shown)


extra/lib/plausible/stats/exploration.ex
```diff
@@ -238,7 +255,21 @@ defmodule Plausible.Stats.Exploration do
   defp normalize_pathname("/"), do: "/"
   defp normalize_pathname(pathname), do: String.trim_trailing(pathname, "/")
 
-  defp next_steps_query(query, steps, search_term, direction, max_candidates, include_wildcard?)
+  defp filter_eligible_goals(goals) do
+    Enum.filter(goals, fn g ->
+      is_nil(g.currency) and g.scroll_threshold == -1 and g.custom_props == %{}
+    end)
```
aerosol:
```suggestion
    Enum.reject(goals, fn g ->
      Plausible.Goal.Revenue.revenue?(g) or g.scroll_threshold > -1 or Plausible.Goal.has_custom_props?(g)
    end)
```

extra/lib/plausible/stats/exploration.ex
```diff
@@ -342,6 +342,81 @@ defmodule Plausible.Stats.Exploration do
...
+          pathname: g.page_path || ""
+        }
+      end)
+
+    types = %{name: :string, pathname: :string}
+
+    from m in subquery(query),
+      left_join: g in values(to_exclude, types),
+      on: g.name == m.name and g.pathname == m.pathname,
+      where: g.name == "" or m.includes_subpaths
```
aerosol, line 418: when is ever `g.name` an empty string? Shouldn't it be `is_nil(g.name)`? Or is that an equivalent in ClickHouse? 🤔

extra/lib/plausible/stats/exploration.ex
```diff
@@ -8,16 +8,22 @@ defmodule Plausible.Stats.Exploration do
...
+              subpaths_count: 0,
+              is_goal: false
 
     @spec from(map()) :: t()
     def from(step) do
-      new(step.name, step.pathname, step.includes_subpaths, step.subpaths_count)
+      new(step.name, step.pathname, step.includes_subpaths, step.subpaths_count, step.is_goal)
     end
 
     @spec new(String.t(), String.t(), boolean(), non_neg_integer()) :: t()
```
aerosol: update typespecs with the 5th argument?

aerosol (approved):

extra/lib/plausible/stats/exploration.ex
zoldar: When joining "values" with ad-hoc String type, the left join without a match results in this field defaulting to empty string and not null. At least that's what I observe.

extra/lib/plausible/stats/exploration.ex
zoldar: https://github.com/plausible/analytics/pull/6299/commits/f3e9d56bd0f89d7a7ce67047aa95832698f10a60

PR #6300, opened 2026-04-30
Fix match_day_of_week for year_over_year comparisons in day and 24h periods

RobertJoonas:
### Changes

Currently, the `Last 24h` and `today` periods are not respecting `match_day_of_week` in `year_over_year` comparison mode.

For `Last 24h` it was never implemented. For the `today` period it used to work before https://github.com/plausible/analytics/pull/6205 because it didn't get the special "datetime" treatment (therefore, it fell into the default `get_comparison_date_range(%Query{include: %{compare: :year_over_year}} = source_query)` function clause, and got `maybe_match_day_of_week` called on the result).

This PR implements the missing piece and adds tests.

### Tests

### Changelog
- [x] Does not require a changelog update -- both last 24 hours period and today comparison matching the exact number of hours are part of the current release. 

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/comparisons.ex
+++ b/lib/plausible/stats/comparisons.ex
@@ -156,14 +156,28 @@ defmodule Plausible.Stats.Comparisons do
   defp get_comparison_datetime_range(
          %Query{
            input_date_range: input_range,
-           include: %{compare: :year_over_year}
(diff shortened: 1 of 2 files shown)


zoldar (approved):
LGTM ✨

PR #6317, opened 2026-05-06
Exclude selected wilcard pattern matches from subsequent step suggestions

zoldar:
### Changes

This PR ensures that if either w pageview goal pattern or implicit wildcard were selected in the previous step, all matching entries are excluded from the suggestions for the next step.

It also fixes a crash bug due to empty list of goals to exclude from the combined query.

### Tests


aerosol (approved):

aerosol:
Much nicer this way!

PR #6319, opened 2026-05-07
Upgrade available only to owners and billing roles

aerosol:
### Changes

The prominent "Upgrade" button rendered on every page, should be only visible to `:owner` and `:billing` roles in case of set up teams. Similarly, the subscription settings LV should be only accessible directly only for these roles. No such restrictions apply to personal teams.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/controllers/settings_controller.ex
+++ b/lib/plausible_web/controllers/settings_controller.ex
@@ -13,9 +13,6 @@ defmodule PlausibleWeb.SettingsController do
        when action in [:update_team_name]
 
   plug Plausible.Plugs.AuthorizeTeamAccess,
-       [:owner, :billing] when action in [:subscription]
-
-  plug Plausible.Plugs.AuthorizeTeamAccess,
        [:owner]
        when action in [
               :team_danger_zone,
--- a/lib/plausible_web/live/subscription_settings.ex
+++ b/lib/plausible_web/live/subscription_settings.ex
@@ -5,6 +5,8 @@ defmodule PlausibleWeb.Live.SubscriptionSettings do
   use PlausibleWeb, :live_view
   use Plausible
 
+  require Plausible.Billing
+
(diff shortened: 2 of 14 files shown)


lib/plausible_web/controllers/settings_controller.ex
```diff
@@ -12,9 +12,6 @@ defmodule PlausibleWeb.SettingsController do
        [:owner, :admin]
        when action in [:update_team_name]
 
-  plug Plausible.Plugs.AuthorizeTeamAccess,
-       [:owner, :billing] when action in [:subscription]
```
aerosol, line 16: dead code

lib/plausible_web/live/subscription_settings.ex
```diff
@@ -12,42 +12,47 @@ defmodule PlausibleWeb.Live.SubscriptionSettings do
...
-    pageview_usage = Teams.Billing.monthly_pageview_usage(team)
-    site_usage = Teams.Billing.site_usage(team)
-    team_member_usage = Teams.Billing.team_member_usage(team)
 
-    usage = %{
-      monthly_pageviews: pageview_usage,
-      sites: site_usage,
-      team_members: team_member_usage
-    }
+    if Teams.setup?(team) && socket.assigns.current_team_role not in [:owner, :billing] do
```
zoldar: Perhaps we could at least extract the billing-eligible roles (`[:owner, :billing]`) and put them behind something like `Plausible.Billing.allowed_roles()` or `Plausible.Auth.allowed_billing_roles()`?

zoldar (approved):

lib/plausible_web/live/subscription_settings.ex
aerosol: https://github.com/plausible/analytics/pull/6319/commits/ab3fca4f975595c92b33337a5503d8f9584cf6a9

PR #6327, opened 2026-05-07
Add non-catch-all version of dashboard route

ukutaht:
Currently dashboard links are generated with trailing slash across the application. This PR adds a route without the catch-all part which affects the router helpers so the link is generated without the trailing slash as shown here:

```
iex(6)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :stats, "plausible.io", [])
"/plausible.io/"
iex(7)> r PlausibleWeb.Router
{:reloaded, [PlausibleWeb.Router, PlausibleWeb.Router.Helpers]}
iex(8)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :stats, "plausible.io", [])
"/plausible.io"
```

The trailing slash appears in live demo stats, this is an attempt to suppress it.
Screenshot 2026-05-07 at 14 15 05

### DIFF
--- a/lib/plausible_web/router.ex
+++ b/lib/plausible_web/router.ex
@@ -734,6 +734,7 @@ defmodule PlausibleWeb.Router do
       put "/:domain/settings", SiteController, :update_settings
 
       get "/:domain/export", StatsController, :csv_export
+      get "/:domain", StatsController, :stats
       get "/:domain/*path", StatsController, :stats
     end
   end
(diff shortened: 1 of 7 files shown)


apata (approved):

aerosol:
Nice catch 😁
If only we could get rid of the trailing wildcard too... I recall it's what is preventing us from using compile time route helpers

ukutaht:
>  If only we could get rid of the trailing wildcard too... I recall it's what is preventing us from using compile time route helpers

Yeah I think the problem is that with our routing structure *any* route is technically defined so the new verified routes would provide no validation at all. Like if we use this route:

`~r"/blah/blah/foo/bar"`

it would still match the router entry


`get "/:dashboard/*path"`

and pass the compile-time validation

PR #6331, opened 2026-05-08
Two fixes related to main graph and revenue metrics

RobertJoonas:
### Changes

* Code refactor: organize `QueryBuilder.do_build` a bit better, using a pipeline with steps rather than a long function.
* Fix 1: querying revenue metrics while including imported data resulted in invalid SQL due to the imported pagination optimization
* Fix 2: optimistic graph fetch 500 -- stored graph metric is a revenue metric -> goal filter gets removed -> optimistic attempt to fetch the graph -> revenue metrics dropped -> empty metrics -> invalid SQL.

### Tests

### Changelog
- [x] No update needed. Both bugs surfaced with main graph v2 which still hasn't made it to a CE release

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/parsed_query_params.ex
+++ b/lib/plausible/stats/parsed_query_params.ex
@@ -1,6 +1,8 @@
 defmodule Plausible.Stats.ParsedQueryParams do
   @moduledoc false
 
+  alias Plausible.Stats.Query
+
   defstruct input_date_range: nil,
             # `relative_date` is a convenience currently exclusive to the internal
             # dashboard API for constructing datetime ranges. It adds the ability
@@ -19,6 +21,11 @@ defmodule Plausible.Stats.ParsedQueryParams do
             # configured for the site. Missing goals simply match nothing in SQL, matching
             # the behaviour of legacy endpoints like top_stats.
             skip_goal_existence_check: false
+
+  def to_query!(%__MODULE__{} = parsed_query_params) do
+    query_fields = (Query.__struct__() |> Map.keys()) -- [:__struct__]
+    struct!(%Query{}, Map.take(parsed_query_params, query_fields))
+  end
 
   def new!(params) when is_map(params) do
     struct!(__MODULE__, Map.to_list(params))
--- a/lib/plausible/stats/parsed_query_params.ex
+++ b/lib/plausible/stats/parsed_query_params.ex
@@ -23,7 +23,7 @@ defmodule Plausible.Stats.ParsedQueryParams do
             skip_goal_existence_check: false
 
   def to_query!(%__MODULE__{} = parsed_query_params) do
-    query_fields = (Query.__struct__() |> Map.keys()) -- [:__struct__]
+    query_fields = %Query{} |> Map.from_struct() |> Map.keys()
     struct!(%Query{}, Map.take(parsed_query_params, query_fields))
   end
 
--- a/lib/plausible/stats/query_builder.ex
+++ b/lib/plausible/stats/query_builder.ex
@@ -30,7 +30,7 @@ defmodule Plausible.Stats.QueryBuilder do
 
   def build(site, %ParsedQueryParams{} = parsed_query_params, debug_metadata) do
     with {:ok, parsed_query_params} <- resolve_segments_in_filters(parsed_query_params, site),
-         query = do_build(parsed_query_params, site, debug_metadata),
+         {:ok, query} <- do_build(parsed_query_params, site, debug_metadata),
          :ok <- validate_order_by(query),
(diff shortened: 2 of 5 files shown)


lib/plausible/stats/query_builder.ex
```diff
@@ -150,49 +150,62 @@ defmodule Plausible.Stats.QueryBuilder do
   end
 
   defp do_build(parsed_query_params, site, debug_metadata) do
-    now = parsed_query_params.now || DateTime.utc_now(:second)
+    {%{relative_date: relative_date}, query_fields} =
+      Map.split(parsed_query_params, [:relative_date, :skip_goal_existence_check])
+
+    struct!(%Query{}, Map.to_list(query_fields))
```
aerosol: much nicer like this, you could start the pipeline with `%Query{}` too. There's also no need for `Map.to_list/1`, the signature is `struct!(module() | struct(), Enumerable.t()) :: struct()`

lib/plausible/stats/query_builder.ex
```diff
@@ -150,49 +150,62 @@ defmodule Plausible.Stats.QueryBuilder do
   end
 
   defp do_build(parsed_query_params, site, debug_metadata) do
-    now = parsed_query_params.now || DateTime.utc_now(:second)
+    {%{relative_date: relative_date}, query_fields} =
```
aerosol: Does it effectively mean we discard `skip_goal_existence_check`? Unclear what's the reasoning behind it. Not sure why do we split at all, if we only care about `relative_date`. Can you elaborate?

lib/plausible/stats/query_builder.ex
```diff
@@ -252,10 +250,29 @@ defmodule Plausible.Stats.QueryBuilder do
         :ok
       end
     end
+
+    defp maybe_drop_revenue_metrics(query) do
+      if query.include.drop_unavailable_revenue_metrics and
+           map_size(query.revenue_currencies) == 0 do
```
aerosol: maybe pattern match against that in function head/guard?

lib/plausible/stats/query_builder.ex
aerosol: If the intention is to rewrite `ParsedQueryParams` into `Query` struct, whenever a key exists on both ends we could do something like:

```elixir
Map.take(parsed_query_params,  Map.keys(Query.__struct__())
```

and then refer to `parsed_query_params.relative_date`

aerosol: And perhaps move this to `ParsedQueryParams.to_query()`?

lib/plausible/stats/query_builder.ex
RobertJoonas: We split to get a map with direct query fields. Not everything in `ParsedQueryParams` will make it into the actual query struct. Some fields are just required to build a query.

`relative_date` is used for constructing a `utc_time_range` which is why this function needs it. As for `skip_goal_existence_check` that's used elsewhere in the `QueryBuilder.build` pipeline. This function simply ignores it indeed.

lib/plausible/stats/query_builder.ex
RobertJoonas: We're Map.splitting a struct so `query_fields` is a struct too. Removing `Map.to_list` results in `(Protocol.UndefinedError) protocol Enumerable not implemented for Plausible.Stats.ParsedQueryParams (a struct).`

lib/plausible/stats/query_builder.ex
RobertJoonas: https://github.com/plausible/analytics/pull/6331/commits/ac9d0a557c4f21a356cbb793317865b96dc25617

lib/plausible/stats/query_builder.ex
RobertJoonas: For some reason didn't see the other 2 comments before. But your suggested approach sounds good. I've extracted `ParsedQueryParams.to_query!/1` that finds the common query fields and builds the `%Query{}` struct.

aerosol (approved):

lib/plausible/stats/parsed_query_params.ex
```diff
@@ -20,6 +22,11 @@ defmodule Plausible.Stats.ParsedQueryParams do
             # the behaviour of legacy endpoints like top_stats.
             skip_goal_existence_check: false
 
+  def to_query!(%__MODULE__{} = parsed_query_params) do
+    query_fields = (Query.__struct__() |> Map.keys()) -- [:__struct__]
```
aerosol: `%Query{} |> Map.from_struct() |> Map.keys()`

PR #6340, opened 2026-05-11
Return explicit journey end event

zoldar:
### Changes

This PR implements explicit listing of journey end. It's currently displayed like any other item among the suggestions, subject to the same sorting order and limit of number of displayed entries (if it's small enough, it won't make it to the list).

The visual representation might be improved in follow-up work (dedicated icon? some behavior on click?).

image

### Tests

### DIFF
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -104,7 +104,7 @@ function stepsToJourneyParam(steps) {
   )
 }
 
-function parseFetchedResults(results) {
+function maybeEmptyResults(results) {
   if (
     results.length === 0 ||
     (results.length === 1 && results[0].step.name === JOURNEY_END_EVENT)
@@ -820,7 +820,7 @@ function useExplorationData(site, dashboardState, inViewport) {
                 if (!isStale())
                   setState((prev) => ({
                     ...prev,
-                    activeResults: parseFetchedResults(r?.next ?? []),
+                    activeResults: maybeEmptyResults(r?.next ?? []),
                     rateLimited: false
                   }))
               })
@@ -865,7 +865,7 @@ function useExplorationData(site, dashboardState, inViewport) {
               if (!isStale())
                 setState((prev) => ({
                   ...prev,
-                  activeResults: parseFetchedResults(r?.next ?? []),
+                  activeResults: maybeEmptyResults(r?.next ?? []),
                   rateLimited: false
                 }))
             })
@@ -916,7 +916,7 @@ function useExplorationData(site, dashboardState, inViewport) {
         setState((prev) => {
           const next = {
             ...prev,
-            activeResults: parseFetchedResults(response?.next ?? []),
+            activeResults: maybeEmptyResults(response?.next ?? []),
             rateLimited: false
           }
           if (includeFunnel) {
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -105,11 +105,9 @@ function stepsToJourneyParam(steps) {
 }
 
 function parseFetchedResults(results) {
-  if (results.length === 0) {
-    return []
-  } else if (
-    results.length === 1 &&
-    results[0].step.name === JOURNEY_END_EVENT
+  if (
+    results.length === 0 ||
+    (results.length === 1 && results[0].step.name === JOURNEY_END_EVENT)
   ) {
     return []
   } else {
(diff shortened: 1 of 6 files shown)


assets/js/dashboard/extra/exploration.js
```diff
@@ -102,6 +104,19 @@ function stepsToJourneyParam(steps) {
   )
 }
 
+function parseFetchedResults(results) {
```
aerosol: this doesn't do what it says it does, though I'm not sure what would be a better name. `maybeEmptyResults`?

Could use an or in place of else if for returning the empty array?

assets/js/dashboard/extra/exploration.js
```diff
@@ -354,7 +369,9 @@ function CandidateCard({
   colIndex,
   onSelect
 }) {
-  const isCustomEvent = step.name !== 'pageview'
+  const isJourneyEnd = step.name === JOURNEY_END_EVENT
+  const isCustomEvent =
+    step.name !== 'pageview' && step.name !== JOURNEY_END_EVENT
```
aerosol, line 374: funny, should we prevent it from being created as a goal? https://github.com/plausible/analytics/blob/exploration-no-further-action/lib/plausible/goal.ex#L149-L162

assets/js/dashboard/extra/exploration.js
```diff
@@ -22,6 +22,8 @@ import { ChevronUpDownIcon, ChevronRightIcon } from '@heroicons/react/20/solid'
 import { FlagIcon, MagnifyingGlassIcon } from '@heroicons/react/24/outline'
 import { popover } from '../components/popover'
 
+const JOURNEY_END_EVENT = '__journey_end__'
```
aerosol, line 25: Similarly to journey max length, I wonder if we should expose this kind of metadata through the API instead of keeping redundant definitions.

extra/lib/plausible/stats/journey/step.ex
```diff
@@ -12,6 +12,15 @@ defmodule Plausible.Stats.Exploration.Journey.Step do
             subpaths_count: 0,
             is_goal: false
 
+  @journey_end_event "__journey_end__"
+  @journey_end_label "no further action"
```
aerosol, line 16: should it be searchable within the step box? 🙈

aerosol (approved):
Approach is fine, but we'll have to tweak it visually most likely once @sanne-san is back.

assets/js/dashboard/extra/exploration.js
zoldar: > Could use an or in place of else if for returning the empty array?

Sorry, I'm a bit lost here, what do you mean?

assets/js/dashboard/extra/exploration.js
zoldar: Indeed, it would make sense to make it reserved 👍

assets/js/dashboard/extra/exploration.js
zoldar: Will do, thanks!

extra/lib/plausible/stats/journey/step.ex
zoldar: Huh, it should be, but for reason it isn't. I'll look into it, thanks.

assets/js/dashboard/extra/exploration.js
aerosol: Great job me formatting text! I meant something like

`if (results.length === 0 || (results.length === 1 && results[0].step.name === JOURNEY_END_EVENT)) { return [] } else { return results }`

zoldar: https://github.com/plausible/analytics/pull/6340/commits/32a4321904e6adf5be3e0752a368d29a625eeb8b and https://github.com/plausible/analytics/pull/6340/commits/c48ae5985431f4b5dcd90e16fe8c99770c393059

assets/js/dashboard/extra/exploration.js
zoldar: https://github.com/plausible/analytics/pull/6340/commits/1809220977b149882fb5f439921c8f0989d56718

assets/js/dashboard/extra/exploration.js
zoldar: On a second thought, I'll address that in a follow-up, perhaps together with max length enforcement.

extra/lib/plausible/stats/journey/step.ex
zoldar: https://github.com/plausible/analytics/pull/6340/commits/6845756f0b1ac450144497e77d185699d866b9aa

PR #6341, opened 2026-05-11
Fix realtime top stats in limited to segment shared link

RobertJoonas:
### Changes

Currently the realtime Top Stats are broken in shared links that are limited to a segment. Precisely, the "current visitors" query is what the backend rejects -- `StatsController` enforces a segment filter via a plug, but the FE simply strips all filters from this query (including the enforced segment one).

Previous behaviour was to skip the plug and allow the current visitors query without the "limited to segment" filter, therefore returning data that is completely out of scope for the given dashboard. Instead of sticking to that behaviour and building an additional flag and ignore it in the new `/query` endpoint too, I've made the new current visitors metric respect the segment filter. It now removes all other filters **except** the limited segment one.

Note: This is a quick and dirty fix. I've made some good progress with improving the API v2 interface on the frontend. Planning to have currentVisitors apiState available via a new context (so that the response could be used wherever necessary, i.e. in top stats or the top-bar). That allows us to keep top-stats itself to a single query, with a solid Tanstack QueryKey, etc...

The tests will therefore change too.

### Tests
- [x] In upcoming PR...

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/fetch-top-stats.ts
+++ b/assets/js/dashboard/stats/graph/fetch-top-stats.ts
@@ -12,12 +12,15 @@ import { createStatsQuery, ReportParams, StatsQuery } from '../../stats-query'
 import {
   hasConversionGoalFilter,
   hasPageFilter,
-  isRealTimeDashboard
+  isRealTimeDashboard,
+  remapToApiFilters
 } from '../../util/filters'
+import { isSegmentFilter } from '../../filtering/segments'
 
 export function topStatsQueries(
   dashboardState: DashboardState,
-  metrics: Metric[]
+  metrics: Metric[],
+  limitedToSegmentId: number | null
 ): [StatsQuery, StatsQuery | null] {
   let currentVisitorsQuery = null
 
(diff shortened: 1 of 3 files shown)


assets/js/dashboard/stats/graph/fetch-top-stats.ts
```diff
@@ -26,7 +29,16 @@ export function topStatsQueries(
       metrics: ['visitors']
     })
 
-    currentVisitorsQuery.filters = []
+    const filters = limitedToSegmentId
```
apata, line 32: issue, blocking: This changes the meaning of "Current visitors" top stats block depending on whether the user accesses the dashboard using shared link limited to segment or normally, with the same segment applied.

RobertJoonas: Yeah, the decision was to keep the current behaviour where current visitors ignore all filters (including enforced segment). That's now done in https://github.com/plausible/analytics/pull/6351, where I've switched current visitors back to the dedicated `/current-visitors` endpoint. Closing this PR in favour of 6351.

PR #6349, opened 2026-05-13
Avoid reload of the current column on deselecting entry

zoldar:
### Changes

This PR prevents flicker of the step column when an entry in it is deselected. The frozen version of the deselected column is preserved and displayed until the step and funnel request completes.

### DIFF
--- a/assets/js/dashboard/extra/exploration.js
+++ b/assets/js/dashboard/extra/exploration.js
@@ -578,7 +578,7 @@ function ExplorationColumn({
           </span>
         )}
 
-        {showSearch && (
+        {!headerConversionRate && showSearch && (
           <input
             data-testid="search-input"
             type="text"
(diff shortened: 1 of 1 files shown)


aerosol:
https://github.com/user-attachments/assets/4f78cf44-1e89-4e61-800b-80f9993c8d71

Doesn't work too well when aggressively triggering rate limit, not sure if critical enough to spend time on.

aerosol (approved):

zoldar:
> Doesn't work too well when aggressively triggering rate limit, not sure if critical enough to spend time on.

Should work better after https://github.com/plausible/analytics/pull/6349/commits/8745d0af89cfc6b443f93a6089ebab792f9522c0

I have also very slightly simplified the monstrous condition https://github.com/plausible/analytics/pull/6349/commits/0cd2e8e2c7f78d67994bd096bfa2648e52ea2ef3. All that will be reworked in a follow-up refactor, _after_ I prepare an e2e test harness 🙈

PR #6350, opened 2026-05-13
Enforce UI maximum of 20 steps per journey

aerosol:
The limit is kept at the backend too.

It's not ideal we maintain redundant maximums, but also it felt too much in here to handle a server-side
generated error, since technically the limit is known up-front.

For now, it follows the `__journey_end_event__`
pattern.


zoldar (approved):
Nice ✨

PR #6351, opened 2026-05-13
Refactor FE API v2 interface + current visitors bugfix

RobertJoonas:
### Changes

_For reviewers: commit-by-commit highly recommended_

Makes the API interface more solid on the frontend. Smaller improvements should be pretty self-explanatory and easy to understand when skimming through the PR commit by commit.

There were also some bigger API v2 interface design changes though.

#### 1. Introduce `use-query-api.ts`

It's a module containing two hooks:

* `useQueryApi` (new)
  * handles realtime refetch logic (returns `{apiState, isRealtimeSilentUpdate}`)
  * built-in cache TTL logic (nothing the caller should be concerned about anymore)
  * can be used by every report that targets the /query API, except for details views
* `useSearchAndPaginateQueryApi` (partially existing as usePaginatedQueryApi)
  * added search functionality, which is why the name change -- if there's a non-empty search string (that's also included in the query key as it affects query uniqueness) a `['contains', dimensions[0], [searchString]]` filter gets added to  the constructed statsQuery.

Both hooks now operate with a queryKey of a new unified type -- `StatsReportQueryKey`.

#### 2. The bug fix

To be able to switch to the new hook in top stats, the "current visitors" query had to be extracted somewhere else. We're using current visitors in two places, so it makes sense for them to be fetched in a single context. Doing that, I also discovered a bug in shared links limited to a segment, where realtime top stats fetch current visitors from the `/query` API, enforcing the segment filter -- 400 returned while should allow current visitors to ignore even the enforced segment filter. This PR fixes that by revertingg back to the old `/current-visitors` endpoint.

### Tests
- [x] Tests have been adjusted. More coverage to be added in upcoming PRs

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/graph/main-graph.tsx
```diff
@@ -91,8 +85,8 @@ export const MainGraph = ({
   const { selectedIndex } = tooltip
   const panGestureStartTimeRef = useRef<number | null>(null)
   const metric = data.query.metrics[0] as Metric
-  const interval = data.interval
-  const period = data.period
+  const interval = data.query.dimensions[0].split(':')[1] as Interval
```
apata: suggestion, non-blocking: this seems like a good responsibility for shared FE utility or just plain map from BE interval to FE interval

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -1,17 +1,77 @@
...
+  return { apiState, isRealtimeSilentUpdate }
+}
+
+function getMainGraphQuery(queryKey: StatsReportQueryKey): StatsQuery {
+  const [_reportId, keyOpts] = queryKey
+  const { dashboardState, reportParams } = keyOpts
+  const statsQuery = createStatsQuery(dashboardState, reportParams)
+
+  if (isRealTimeDashboard(dashboardState)) {
+    statsQuery.date_range = DashboardPeriod.realtime_30m
```
apata: suggestion, non-blocking: Modifying objects in place creates code paths that are harder to follow, for me anyway.
```suggestion
    return { ...statsQuery, date_range: DashboardPeriod.realtime_30m }
```

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -1,17 +1,77 @@
...
+  metric: Metric | null,
+  interval: Interval
+): {
+  apiState: UseQueryResult<MainGraphResponse>
+  isRealtimeSilentUpdate: boolean
+} {
+  const site = useSiteContext()
+  const { dashboardState } = useDashboardStateContext()
+
+  const mainGraphQueryKey = useMemo((): StatsReportQueryKey => {
```
apata: question, non-blocking: What's the reason to memo this?

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -1,17 +1,77 @@
...
+    return [
+      'main-graph',
+      {
+        dashboardState,
+        reportParams: {
+          // Should default to visitors if metric is null? Currently possibly invalid
+          // query with `metrics: [null]` which will never run due to `enabled: false`
```
apata: Reading this as a question to reviewer (and it's a valid question!), I say it's better to fail with an invalid query in those cases. Reasoning: less work for the BE.

assets/js/dashboard/hooks/use-query-api.ts
```diff
@@ -0,0 +1,200 @@
...
+    queryKey: statsReportQueryKey,
+    enabled,
+    queryFn: async ({ queryKey }) => {
+      const [_, keyOpts] = queryKey as StatsReportQueryKey
+      const response = await stats<TResponse>(site, getStatsQuery(queryKey))
+      return withExtraContext(response, keyOpts.dashboardState)
+    },
+    placeholderData: (previousData) => previousData,
+    staleTime: ({ queryKey }) => {
+      const [_, keyOpts] = queryKey as StatsReportQueryKey
```
apata: nitpick, non-blocking: casting with `as` shouldn't be needed here, TS should automatically infer it by virtue of the impressive type logic of tanstack-query (which is one of its major selling points, actually). There's a few other potentially superfluous casts like that.

assets/js/dashboard/stats/current-visitors.js
```diff
@@ -1,39 +1,17 @@
...
-      document.removeEventListener('tick', updateCount)
-    }
-  }, [updateCount])
-
-  useEffect(() => {
-    updateCount()
-  }, [dashboardState, updateCount])
-
-  if (currentVisitors !== null && dashboardState.filters.length === 0) {
+  if (currentVisitors !== null) {
```
apata, line 14: question, non-blocking: why is the guard against dashboardState.filters.length === 0 not needed any more?

assets/js/dashboard/api.ts
```diff
@@ -37,10 +38,21 @@ export type QueryResultRow = {
...
+  isRealtime: boolean
+  hasConversionGoalFilter: boolean
+}
+
 export type QueryApiResponse = {
   query: QueryResultQuery
   meta: QueryResultMeta
   results: QueryResultRow[]
+  extraContext: ExtraContext
 }
```
apata, line 56: issue, non-blocking: I think `useQuery` has API for this so we don't actually need to hand roll it. Check the description for `meta` at https://tanstack.com/query/latest/docs/framework/react/reference/useQuery. I didn't try it yet locally, but I think we should

apata (approved):
I think it's going in a good direction. Maybe the useQuery wrapper could be thinner, but that's just my personal preference. I appreciate the refactor to make greater use of query-api-schema.json!

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: Yeah, I framed the question a bit wrong there. What I actually had in mind was setting a default selectedMetric when there's nothing stored, which could end up improving the UX significantly.

Current behaviour: when there's no stored metric, set the initial `selectedMetric` to `null`. That means, we do not attempt to fetch the graph until Top Stats have loaded. Top Stats give us the "available metrics" from which we take the first one (currently, it's always `visitors`). Then we refetch the graph.

What it could be: when there's no stored metric, set the initial `selectedMetric` to `visitors`. Then we can fetch top stats and main graph at the same time, which makes the report appear for the user ~twice as fast.

assets/js/dashboard/stats/current-visitors.js
RobertJoonas: I moved this into the context provider. It now only fetches current visitors if `isRealTimeDashboard(dashboardState) || dashboardState.filters.length === 0`. If this condition is not met, the query is not executed and `CurrentVisitorsContext.Provider` provides the value as `null`. Therefore, the additional filter length check in the CurrentVisitors component is redundant.

Forgot to mention this in the PR description but this is actually a pretty important fix. Currently on production, if there are filters applied and `period != realtime`, we still fetch current visitors. To see the behaviour on prod, go to `https://plausible.io/plausible.io?f=is,page,/:dashboard&period=28d` and monitor the network tab until a realtime tick. You can see a redundant `GET /current-visitors` request even though the value never gets rendered. This PR fixes that.

apata: Thanks for clarifying! Should it be the responsibility of current visitors context to toggle this top bar component on and off? Is data === null the best signal for that?

assets/js/dashboard/stats/graph/fetch-main-graph.ts
apata: Oh I see! That seems like a great improvement indeed. I suggest we do it in a separate PR though

assets/js/dashboard/api.ts
RobertJoonas: I don't think that solves the problem with rendering previousData. As soon as `queryKey` changes, `meta` does too. In other words, it will always represent the `dashboardState` of the current (new) query, rather than placeholderData.

assets/js/dashboard/stats/graph/main-graph.tsx
RobertJoonas: https://github.com/plausible/analytics/pull/6351/commits/f395656046fb9ee106f1b1a3cb972a1608a357df

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: Looks cleaner indeed! https://github.com/plausible/analytics/pull/6351/commits/b0d50346e59a7693e85a2448b9c98c5fb67b99f9

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: Good shout. Not really needed here. https://github.com/plausible/analytics/pull/6351/commits/47380b467cc07872881e0fdac1f8b3471e96a74a

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: Agreed about doing that separately. I've moved the comment to visitor-graph, on top of the `selectedMetric` state, and made the comment a bit more accurate. https://github.com/plausible/analytics/pull/6351/commits/3b02b1f0a33b9d38117eee80137d1d8f8379b0b4

assets/js/dashboard/hooks/use-query-api.ts
RobertJoonas: Nice catch! https://github.com/plausible/analytics/pull/6351/commits/a8efa583d5150586f669c76ae7be6de3d450f2f3

assets/js/dashboard/stats/current-visitors.js
RobertJoonas: IMHO completely fine to let the context decide in what situation the value is needed to be fetched. It's not only up to the context to decide whether to render it though. E.g. `TopBar` also takes a `showCurrentVisitors` prop which is false on a realtime dashboard. And top stats only renders it in realtime.

PR #6352, opened 2026-05-14
Serve exploration config to the front-end

aerosol:


assets/js/dashboard/site-context.tsx
```diff
@@ -9,6 +9,11 @@ export function parseSiteFromDataset(dataset: DOMStringMap): PlausibleSite {
     funnelsAvailable: dataset.funnelsAvailable === 'true',
     propsAvailable: dataset.propsAvailable === 'true',
     explorationAvailable: dataset.explorationAvailable === 'true',
+    explorationJourneyEndEvent: dataset.explorationJourneyEndEvent ?? '',
+    explorationMaxJourneySteps: parseInt(
+      dataset.explorationMaxJourneySteps ?? '0',
+      10
+    ),
```
zoldar: AFAIU, the fallbacks are here mostly to avoid issues during rolling release right? We might want to make those properties asserted as non-null with "!" later on.

assets/js/dashboard/extra/exploration.js
```diff
@@ -1030,7 +1036,9 @@ function useExplorationData(site, dashboardState, inViewport) {
     state.steps,
     state.activeFilter,
     inViewport,
-    retryCount
+    retryCount,
+    JOURNEY_END_EVENT,
+    MAX_JOURNEY_STEPS
```
zoldar: Are they really necessary as effect dependencies? 🤔

assets/js/dashboard/extra/exploration.js
```diff
@@ -370,6 +367,7 @@ function CandidateCard({
   colIndex,
   onSelect
 }) {
+  const { explorationJourneyEndEvent: JOURNEY_END_EVENT } = useSiteContext()
```
zoldar: IMHO it's better to use `journeyEndEvent` and `maxJourneySteps`, event if those variables extracted from context are constants. Uppercase convention is AFAIK expected for locally declarted constants and that might throw someone reading the code off and wonder where it's coming from.

zoldar (approved):

assets/js/dashboard/site-context.tsx
aerosol: 553a1f2607

assets/js/dashboard/extra/exploration.js
aerosol: 4bd86fca83, thanks

assets/js/dashboard/extra/exploration.js
aerosol: 1f2ac3d6d0

PR #6367, opened 2026-05-18
Fix graph conversion rate

RobertJoonas:
### Changes

Fix regression introduced in https://github.com/plausible/analytics/pull/6351. The refactor missed the `conversion_rate` -> `group_conversion_rate` override. Graph should always request the latter.

Trying to get this in as a quick fix but working on tests as we speak.

### Tests
- [x] Upcoming PR

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/graph/fetch-main-graph.ts
+++ b/assets/js/dashboard/stats/graph/fetch-main-graph.ts
@@ -1,8 +1,7 @@
 import { Metric } from '../metrics'
-import { DashboardState } from '../../dashboard-state'
 import { DashboardPeriod } from '../../dashboard-time-periods'
-import { PlausibleSite, useSiteContext } from '../../site-context'
-import { createStatsQuery, ReportParams, StatsQuery } from '../../stats-query'
+import { useSiteContext } from '../../site-context'
+import { createStatsQuery, StatsQuery } from '../../stats-query'
 import { isRealTimeDashboard } from '../../util/filters'
 import { MetricValue } from '../../api'
 import * as api from '../../api'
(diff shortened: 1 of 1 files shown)


ukutaht (approved):

assets/js/dashboard/stats/graph/fetch-main-graph.ts
```diff
@@ -21,12 +20,15 @@ export function useMainGraphQuery(
...
+  const metricToQuery =
+    metric === 'conversion_rate' ? 'group_conversion_rate' : metric
+
   const mainGraphQueryKey: StatsReportQueryKey = [
     'main-graph',
     {
       dashboardState,
       reportParams: {
-        metrics: [metric!],
+        metrics: [metricToQuery!],
```
zoldar, line 31: Out of curiosity - if the metric is asserted to be non-null, why is the function argument's signature marking it as nullable? 🤔

zoldar (approved):

assets/js/dashboard/stats/graph/fetch-main-graph.ts
RobertJoonas: The bang is currently just silencing TypeScript there. The metric can actually be `null` if there's no stored metric. I'm planning to change that in a follow-up though. There's a comment in visitor-graph.tsx explaining what it could be:

```
// Possible future improvement -- currently, if there's no stored metric,
// the graph fetch doesn't run until Top Stats are loaded. That's because
// Top Stats tell us which metrics are available for the graph. However,
// as things stand today, the `visitors` metric is always available and
// could become the default selectedMetric, making it possible to fetch
// the graph instantly.
```

PR #6387, opened 2026-05-21
Make "No fruther action" clickable

zoldar:
### Changes

"No further action" entries can be now clicked like any other suggestion. Conversion stats are calculated and displayed for them as well.

### Tests

image


aerosol (approved):
Clever in a good way

PR #6393, opened 2026-05-25
Exploration/billing integration

aerosol:
### Changes

This PR renames "Funnels" billing feature to "Funnels and user journeys".
The exploration panel is no longer available for super-admins on all sites, but only on sites where the related subscription enables the Funnels feature. 
Deploying this is essentially go-live at this point.

~~Awaiting CTA tweaks from @sanne-san, then I'll rebase clean.~~

### DIFF
--- a/assets/js/dashboard/stats/behaviours/index.js
+++ b/assets/js/dashboard/stats/behaviours/index.js
@@ -32,7 +32,7 @@ import { getSpecialGoal, isPageViewGoal, isSpecialGoal } from '../../util/goals'
 
 /*global BUILD_EXTRA*/
 /*global require*/
-function maybeRequire() {
+function maybeRequireFunnels() {
   if (BUILD_EXTRA) {
     // eslint-disable-next-line @typescript-eslint/no-require-imports
     return require('../../extra/funnel')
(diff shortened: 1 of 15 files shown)


assets/js/dashboard/stats/behaviours/index.js
```diff
@@ -537,7 +576,7 @@ function Behaviours({ importedDataInView, setMode, mode }) {
                   Funnels
                 </TabButton>
               ))}
-            {!site.isConsolidatedView && site.explorationAvailable && (
+            {!site.isConsolidatedView && isEnabled(Mode.EXPLORATION) && (
```
zoldar, line 579: Perhaps we should also check for `FunnelExploration` being set, like for funnel at https://github.com/plausible/analytics/pull/6393/changes#diff-e368685087d921e325f4a03fe91dbf05740fca7faa373792c50a2884befb43ffR556 ? CE tests do not seem to be complaining, but they may be not catching that.

lib/plausible/teams/billing.ex
```diff
@@ -674,7 +674,8 @@ defmodule Plausible.Teams.Billing do
   end
 
   def allowed_features_for(team) do
-    team = Teams.with_subscription(team)
+    team =
+      Teams.with_subscription(team)
```
zoldar: Some accidental reformat?

assets/js/dashboard/stats/behaviours/index.js
aerosol: This is handled, container is pointing the user to cloud in such case

zoldar: Scratch that, there's a CE-focused banner inside that view so lack of improt check here is intentional. All good.

zoldar (approved):

PR #6404, opened 2026-05-26
Duplicate legacy, and add new sources/browsers/os-s/conversions tests

RobertJoonas:
### Changes

* Duplicate all `Api.StatsController` tests into new legacy files (e.g. `BrowsersTest` -> `BrowsersLegacyTest`)
* Modify the existing tests to target the new /query API.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/test/plausible_web/controllers/api/stats_controller/conversions_test.exs
+++ b/test/plausible_web/controllers/api/stats_controller/conversions_test.exs
@@ -3,6 +3,21 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
 
   @user_id Enum.random(1000..9999)
 
+  defp query_conversions(conn, site, opts) do
+    params = %{
+      "dimensions" => Keyword.get(opts, :dimensions, ["event:goal"]),
+      "date_range" => Keyword.get(opts, :date_range, "all"),
+      "filters" => Keyword.get(opts, :filters, []),
+      "metrics" => Keyword.get(opts, :metrics, ["visitors", "events", "conversion_rate"]),
+      "include" => Keyword.get(opts, :include, nil),
+      "order_by" => Keyword.get(opts, :order_by, nil)
+    }
+
+    conn
+    |> post("/api/stats/#{site.domain}/query", params)
+    |> json_response(200)
+  end
+
   describe "GET /api/stats/:domain/conversions" do
     setup [:create_user, :log_in, :create_site]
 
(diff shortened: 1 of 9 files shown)


test/plausible_web/controllers/api/stats_controller/conversions_test.exs
```diff
@@ -1389,19 +1242,12 @@ defmodule PlausibleWeb.Api.StatsController.ConversionsTest do
...
-          "/api/stats/#{site.domain}/conversions?period=day"
-        )
+      response = query_conversions(conn, site, date_range: "day", include: %{"imports" => true})
 
-      assert json_response(conn, 200)["results"] == [
-               %{
-                 "name" => "Visit /blog**",
-                 "visitors" => 2,
-                 "events" => 2,
-                 "conversion_rate" => 50
```
apata, line 1403: question, non-blocking: why does the result change? If I'm reading it correctly, the original test wanted to ensure that imported data doesn't influence the result when it's not queried

test/plausible_web/controllers/api/stats_controller/sources_test.exs
```diff
@@ -2098,6 +2006,13 @@ defmodule PlausibleWeb.Api.StatsController.SourcesTest do
       ])
 
       populate_stats(site, [
+        build(:imported_visitors,
```
apata, line 2009: question, non-blocking: why do we have to build imported_visitors here?

apata (approved):

test/plausible_web/controllers/api/stats_controller/conversions_test.exs
RobertJoonas: That test description clearly didn't match with what the test did, so I interpreted it as a typo by the test author. It's saying "with imported data" and IMO the main logic under test here is that glob patterns are handled well in imported queries.

excluding imported data when it's not requested is already tested in query_parser and in a few other Api.StatsController tests. E.g.: test/plausible_web/controllers/api/stats_controller/pages_test.exs:1324 (test called `returns top pages by visitors with imported data`) is doing two requests, first without, and then with imports included.

I assume it's what this test intended to do too, but the second request was never added. I believe it's fine to only test the with imported case here though.

test/plausible_web/controllers/api/stats_controller/sources_test.exs
RobertJoonas: One "imported visitor" always ends up in all the `imported_*` tables. For a sources breakdown, the table from which we get `visitors` (per source) from is `imported_sources`. But for getting the `total_visitors` (percentage metric denominator), the `imported_visitors` table is queried.

The reason it isn't needed in legacy tests is that the legacy API automatically applies a "not empty" filter for a bunch of dimensions. See lib/plausible/stats/breakdown.ex:141. With the new API, this will be up to the client to do.

And now, getting to the deepest level of this explanation 😄 -- the `utm_content != ""` filter makes the `total_visitors` query select from `imported_sources` rather than `imported_visitors` (because it needs to filter).

PR #6408, opened 2026-05-27
Exploration UI refactor

zoldar:
### Changes

This PR attempts to do some gradual refactoring improvements to exploration UI logic.

So far, following changes were made:

- conversion rate type in funnel and provisional funnel types is now always a string, for consistency as that's how we get it from the backend to avoid float related inaccuracies
- the rounded percentage helper got removed with a more general round number formatter
- the selected suggestion is now passed as a single object instead of a collection of fields/arguments
- ensure `stepMaxVisitors` is always a numeric value to avoid potential division by zero (TS did not catch that, strangely)
- always user `Number` for casting strings to numeric values (instead of `parseFloat`)

There's definitely potential for more but that's at least some small start.

@apata if you have any free cycles, I'd really appreciate your eyes on this. I'd be keen on going over the exploration UI code together if you'd like to, to provide any missing context or unclear intent.

### DIFF
--- a/assets/js/dashboard/util/number-formatter.ts
+++ b/assets/js/dashboard/util/number-formatter.ts
@@ -67,13 +67,17 @@ export function durationFormatter(duration: number): string {
   }
 }
 
+export function roundedNumberFormatter(number: number): string {
+  if (Math.abs(number) > 0 && Math.abs(number) < 0.1) {
+    return number.toFixed(2)
+  } else {
+    return number.toFixed(1).replace(/\.0$/, '')
+  }
+}
+
 export function percentageFormatter(number: number | null): string {
   if (typeof number === 'number') {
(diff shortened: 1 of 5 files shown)


assets/js/dashboard/util/number-formatter.ts
```diff
@@ -67,13 +67,17 @@ export function durationFormatter(duration: number): string {
...
+}
+
 export function percentageFormatter(number: number | null): string {
   if (typeof number === 'number') {
-    if (Math.abs(number) > 0 && Math.abs(number) < 0.1) {
-      return number.toFixed(2) + '%'
-    } else {
-      return number.toFixed(1).replace(/\.0$/, '') + '%'
-    }
+    return roundedNumberFormatter(number) + '%'
```
aerosol, line 80: 👍

apata (approved):

PR #6411, opened 2026-05-28
FE sources migration

RobertJoonas:
### Changes

Migrate all sources reports (utms, sources, referrer-drilldown, and search terms) to v2 breakdown components (both index and details).

While this is supposed to be a refactor, some functional improvements happened naturally:

* Top referrers now displays slide-in percentage metric in IndexBreakdown
* All sources index tabs (including search terms) now cache results (moved to TanStack useQuery)
* Google search terms index style is now consistent with the rest of the dashboard

Also, fixes a bug -- exit and entry pages used to filter out empty results, but stopped doing that with the pages migration. This PR adds the same implicit "not empty" filters for all the dimensions that the backend currently does.

### Tests

### Changelog

### Documentation

### Dark mode
- [ ] The UI has been tested both in dark and light mod


assets/js/dashboard/stats/sources/search-terms.tsx
```diff
@@ -69,140 +66,92 @@ function ConfigureSearchTermsCTA({
...
+        key: 'visitors',
+        renderLabel: () => 'Visitors',
+        renderCell: (item, _isActive) => (
+          <span
+            className="font-medium font-medium text-sm block text-gray-800 dark:text-gray-200"
+            data-testid="metric-value"
+          >
+            {numberShortFormatter(item.visitors)}
+          </span>
+        ),
```
apata, line 119: nitpick, non-blocking: these can easily get out of sync with other index-breakdowns, can we export or generalize the components we use in index breakdown?

assets/js/dashboard/hooks/use-query-api.ts
apata, line 179: suggestion, non-blocking: If we go this route, how to include search can be the responsibility of `getStatsQuery` to avoid editing the stats query on too many levels.

assets/js/dashboard/stats/sources/index.tsx
```diff
@@ -0,0 +1,381 @@
...
+    return (
+      <IndexBreakdown
+        metrics={metrics}
+        dimensions={reportConfig.dimensions}
+        dimensionLabel={reportConfig.dimensionLabel}
+        DimensionElement={DimensionElement}
+        onDataReady={setCurrentQueryApiData}
+        // Possible improvement: getFilterInfo can be dropped from IndexBreakdown props
+        // and be given directly to DimensionCell element (which would also provide the
+        // default function)
```
apata: Good shout about the passthrough prop! Though I would prefer no default to be set and for each report to explicitly define it, importing a particular function if it's shared between multiple reports.

apata (approved):

assets/js/dashboard/stats/sources/search-terms.tsx
RobertJoonas: Good point! For Bar, I've made it accept children, and included the common classnames in a "children wrapper" `<div>`. For metric value rendering I've extracted a wrapper component. https://github.com/plausible/analytics/pull/6411/commits/9df56ed1f3e76cba58c2b165dc571513e35be5d9

assets/js/dashboard/hooks/use-query-api.ts
RobertJoonas: I personally don't see anything wrong with adding query filters in multiple places. The filters are simply joined with an AND operator on the backend.

assets/js/dashboard/stats/sources/index.tsx
RobertJoonas: I've done that in https://github.com/plausible/analytics/pull/6411/commits/e8c250c3e818eed5067e8eb20e9130aab33ed554

PR #6422, opened 2026-06-02
Update channels/source resolution

aerosol:
https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/8966826432

This PR makes multiple changes to source/channel resolution:

   - Introduces AI Assistants channel where popular choices are grouped
   - Bluesky & Mastodon added as Organic Social sources (+ some extra mastodon instances)
   - "Twitter" becomes "X (Twitter)"
   - Copilot hostnames map to "Microsoft Copilot" 
   - "Threads" to Organic Social
   - Fixed favicon resolution for Bluesky, Mastodon, Kagi, AI Assistants, X (Twitter) and Linktree
   - Kagi classified as Organic Search
 

---

- On top of that, source resolution has been rewritten to favour custom sources in the first place. This allows us to categorize "gemini.google.com" before it becomes classified as "Google".
- RefInspector results arte normalized via `canonical/1` call
- Fixed an import gap: GA4 calls `from_referrer/1` so imported referrers are normalized identically to live ingestion, e.g. `t.co` -> "X (Twitter)"

Controller tests were compacted for readability and diffstat karma.

This PR contains migrations, will be split once review is concluded.


lib/plausible/ingestion/acquisition.ex
```diff
@@ -30,13 +31,23 @@ defmodule Plausible.Ingestion.Acquisition do
...
+    {"mastodon", "SOURCE_CATEGORY_SOCIAL"},
+    {"chatgpt", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"chatgpt.com", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"claude", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"phind", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"deepseek", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"copilot", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"grok", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"google gemini", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"perplexity", "SOURCE_CATEGORY_AI_ASSISTANTS"},
```
zoldar, line 49: Not strictly related to this PR but: To reduce a risk of a missed typo causing headaches, perhaps we could consider putting category string labels in module attributes?

lib/plausible/ingestion/acquisition.ex
```diff
@@ -30,13 +31,24 @@ defmodule Plausible.Ingestion.Acquisition do
...
-    {"chatgpt.com", "SOURCE_CATEGORY_SEARCH"},
     {"temu.com", "SOURCE_CATEGORY_SHOPPING"},
     {"discord", "SOURCE_CATEGORY_SOCIAL"},
     {"sogou", "SOURCE_CATEGORY_SEARCH"},
-    {"microsoft teams", "SOURCE_CATEGORY_SOCIAL"}
+    {"microsoft teams", "SOURCE_CATEGORY_SOCIAL"},
+    {"bluesky", "SOURCE_CATEGORY_SOCIAL"},
+    {"mastodon", "SOURCE_CATEGORY_SOCIAL"},
+    {"chatgpt", "SOURCE_CATEGORY_AI_ASSISTANTS"},
+    {"chatgpt.com", "SOURCE_CATEGORY_AI_ASSISTANTS"},
```
ukutaht: I *think* this is a dead entry because the referrer `chatgpt.com` will be transformed to `ChatGPT` source before channel lookups.

priv/custom_sources.json
```diff
@@ -1,12 +1,42 @@
...
+  "mas.to":"Mastodon",
+  "mastodon.online":"Mastodon",
+  "mastodon.social":"Mastodon",
+  "mastodon.world":"Mastodon",
+  "mathstodon.xyz":"Mastodon",
+  "mstdn.jp":"Mastodon",
+  "mstdn.social":"Mastodon",
+  "scholar.social":"Mastodon",
+  "sigmoid.social":"Mastodon",
+  "social.tchncs.de":"Mastodon",
```
ukutaht, line 31: Another source from our live demo that should resolve to Mastodon is `android-app://org.joinmastodon.android`

ukutaht (approved):

priv/custom_sources.json
```diff
@@ -1,12 +1,42 @@
 {
+  "bsky.app":"Bluesky",
+  "go.bsky.app":"Bluesky",
   "android-app://com.reddit.frontpage":"Reddit",
   "baidu.com":"Baidu",
+  "chatgpt.com":"ChatGPT",
+  "chat.openai.com":"ChatGPT",
+  "claude.ai":"Claude",
+  "copilot.microsoft.com":"Microsoft Copilot",
+  "copilot.com":"Microsoft Copilot",
```
zoldar, line 10: plain `copilot` source seems to be missing here (it's in the remap script)?

priv/custom_sources.json
```diff
@@ -40,10 +70,15 @@
   "gads": "Google",
   "google ads": "Google",
   "adwords": "Google",
-  "twitter-ads": "Twitter",
+  "x.com":"X (Twitter)",
+  "mobile.twitter.com": "X (Twitter)",
```
zoldar, line 76: this one seems to be missing from the remap migration for a change?

priv/custom_sources.json
```diff
@@ -40,10 +70,15 @@
...
+  "twitter-ads": "X (Twitter)",
+  "twitter": "X (Twitter)",
   "tiktokads": "TikTok",
   "tik.tok": "TikTok",
   "perplexity": "Perplexity",
+  "chatgpt": "ChatGPT",
+  "claude": "Claude",
```
zoldar: I dont't see those 2 in the remap migration either

lib/plausible/ingestion/acquisition.ex
aerosol: I think you're right, but I find it quite hard to build the mental model around the orchestration of this. BTW would you agree that we should deploy code first, then migrations?

aerosol: e1ef7aba8b

priv/custom_sources.json
aerosol: ede1bc5dea

priv/custom_sources.json
aerosol: ff99b7970f

priv/custom_sources.json
aerosol: :disappointed_relieved: thanks c195d96b6f

priv/custom_sources.json
aerosol: I think those are redundant by now, I'll remove them

aerosol: e9f39ba9f8

lib/plausible/ingestion/acquisition.ex
aerosol: They're used in sql eex :sweat_smile: - how would you make them available there for compile-time validation?

priv/ingest_repo/migrations/20260601000001_remap_sources_v3.exs
cnkk, line 1: Can this be batched into a single ALTER TABLE command? this way we have only one mutation and one column rewrite (per table)

aerosol: https://github.com/plausible/analytics/pull/6422/commits/ac77a891154f3b7251ba5577e554bfefb84e15ad

lib/plausible/ingestion/acquisition.ex
zoldar: They could be put under module attribute as a map of `%{social: "SOURCE_CATEGORY_SOCIAL", ...}` and then exposed via a function, which should be usable during compile time with no issues I think?

aerosol: OK, since this is currently mostly convenience, I'll follow-up in a separate change

aerosol:
Shipped via #6430 and #6431

PR #6423, opened 2026-06-02
Remove the `/team/select` page

sanne-san:
### Changes

- Topbar dropdown now lists all teams (scrolls after ~5 to be sure, though no users have more than 5 teams yet)
- Empty-state "Go to team sites" now has a dropdown button that routes directly instead of via a separate page
- Deleted the route, controller, template, and related tests

### Tests

### Changelog

### Documentation

### Dark mode


zoldar:
Probably not a big deal, but I wonder if it would make sense to move the selected team to the top of the list to ensure it's always in view?

image

sanne-san:
> Probably not a big deal, but I wonder if it would make sense to move the selected team to the top of the list to ensure it's always in view?

Yeah that's a great idea, I'll give that a try!

zoldar (approved):
✨

PR #6434, opened 2026-06-08
Fix issue in e2e tests with clicking tab dropdown before data is loaded

apata:
### Changes

Fixes issue in e2e tests with clicking a tab with dropdown before the data in the dropdown has been fetched.


zoldar (approved):
👍 ❤️

PR #6436, opened 2026-06-08
Dashboard csv export v2

RobertJoonas:
### Changes

Add v2 dashboard CSV export under a feature flag called `dashboard_csv_export_v2`. When enabled, a v2 export button is rendered in the top bar options menu.

> Note: the export is still missing custom props and conversions reports.

Unlike the old export, the v2 `isExporting` state lives in React (rather than a cookie). The new flow is:

* click the button to trigger `POST api/stats/:domain/export`
* await a successful `application/zip` response
* programmatically create an `<a>` element with a download action pointing to an `ObjectUrl` and click it

The new `Api.StatsController` action now takes a request body with all the report parameters from the frontend (`reports-config.ts`). The intent is to let frontend be responsible for the configuration of all dashboard stats reports, so that we could get rid of all legacy Api.StatsController breakdown actions (`sources`, `pages`, `countries`, etc...)

However, in order to keep user affecting changes to a minimum, backend still does a bunch of metric and dimension label overrides. This is legacy behaviour that should go away sooner or later, so there's no good reason for the frontend to do it at this stage. Something to fix in the future when working on improving the CSV export.

### User facing changes

While this PR is meant to be a refactor, it introduces the following functional changes:

* `event:page` filtered CSV export
  * `visitors.csv` does not contain `view_per_visit` because API v2 prohibits it with a page filter
  * `visitors.csv` replaces `visit_duration` with `time_on_page` -- here we could mimic the legacy behaviour and keep fetching `visit_duration`, but I figured it's the right call to update it.
* `event:props:*` or `event:page` filtered CSV export
  * `exit_pages.csv` does not contain `exit_rate` because API v2 prohibits it with when any event filters are used
* CSV filename when dashboard period is not a custom period (for custom periods, the filename remains the same). Reason: frontend has to decide the filename, but the period shorthand is turned into an actual date range on the backend.
  * before: `Plausible export <domain> <from> <to>`
  * after: `Plausible export <domain> <period_shorthand>` (e.g. `period_shorthand = '28d'`)
* Error handling (when `POST /api/stats/:domain/export` responds with 400):
  * before: endless loading spinner
  * after: display an exclamation circle icon with a tooltip saying "Export failed". The export button will appear again after any dashboardState change. (see screenshot below)

image

### Tests

### Changelog
- [x] This PR does not make a user-facing change (yet)

### Documentation

### Dark mode


apata (approved):
I'd prefer more aggressive changes to get rid of the duplication in tests and all the exceptions (regarding dimension and metric names etc), but as a refactor this is good!

PR #6440, opened 2026-06-09
Migrate Conversions, Props reports to /query endpoint

apata:
### Changes

Migrates Conversions and Props reports to /query endpoint

### Tests

### Changelog

### Documentation

### Dark mode


assets/js/dashboard/stats/behaviours/conversions.tsx
```diff
@@ -0,0 +1,89 @@
...
+} from '../../util/filters'
+import { useDashboardStateContext } from '../../dashboard-state-context'
+import { QueryApiResponse, QueryResultRow } from '../../api'
+import { NonTimeDimension } from '../../stats-query'
+import { FilterInfo } from '../../components/drilldown-link'
+
+const BAR_COLOR = 'bg-red-50 group-hover/row:bg-red-100'
+
+type ConversionsProps = {
+  afterFetchData?: (data: QueryApiResponse) => void
```
RobertJoonas: Nit: Let's rename this to onDataReady to be consistent?

assets/js/dashboard/stats/behaviours/index.tsx
```diff
@@ -134,30 +182,37 @@ function Behaviours({ importedDataInView, setMode, mode }) {
...
   const onGoalFilterClick = useCallback(
-    (e) => {
-      const goalName = e.target.innerHTML
-      const isSpecial = isSpecialGoal(goalName)
-      const isPageview = isPageViewGoal(goalName)
+    (goalName: string) => {
+      const isSpecialGoalClick = isSpecialGoal(goalName)
+      // isPageViewGoal currently has no return statement and resolves to
+      // `undefined` at runtime; cast preserves the historical truthiness check.
+      const isPageview = isPageViewGoal(goalName) as unknown as boolean
```
RobertJoonas: Since `isPageViewGoal` never returns anything, we can safely get rid of this `const` I guess. The `!undefined` below currently evaluates to `true` anyway so it's redundant in the `&&` chain.

FTR: I believe the intent was to not switch tab to `Mode.PROPS` if the goal filter being applied is a pageview goal. Looks like that broke at some point then. Maybe accidentally, but I guess it makes sense to change tab even in this case. The props tab will definitely be more interesting than the single goal item :)

assets/js/dashboard/stats/behaviours/index.tsx
```diff
@@ -461,40 +520,45 @@ function Behaviours({ importedDataInView, setMode, mode }) {
...
-  function isRealtime() {
+  function isRealtime(): boolean {
     return dashboardState.period === 'realtime'
   }
 
-  function renderImportedQueryUnsupportedWarning() {
+  function renderImportedQueryUnsupportedWarning(): ReactNode {
     if (mode === Mode.CONVERSIONS) {
       return (
         <ImportedQueryUnsupportedWarning
```
RobertJoonas: Suggestion (non-blocking): What do you think about renaming this component to `FunnelApiImportedWarningBubble`? Otherwise we'd have `ImportedQueryUnsupportedWarning` and `ImportedWarningBubble`, making it hard to understand how they're different.

When I first introduced the new `ImportedWarningBubble` component, I was hoping we could use that here too. Now that I've went down a familiar rabbit hole trying to get rid of `ImportedDataInView`, I've realised that there's no obvious solution on the table for the "Funnels" and "Explore" tabs.

For custom props and conversions however, we could use `ImportedWarningBubble`. For conversions, it'd only be a matter of passing the whole API response. For props, we'd have to also extend the component with a `message?` option.

Then the `FunnelApiImportedWarningBubble` would only depend on `importedDataInView` and nothing else.

Might be worth adding a module-level comment there too:

> Renders an imported warning bubble for the "Funnels" and "Explore" tabs. Currently, while the funnel and exploration queries silently ignore everything related to imports, we still want to let the user know that imports are not included in those reports. Therefore, we rely on Top Stats response to know whether imported data is in range or not, and if it is, we render the warning bubble.

All that said though, probably best to do this separately in a follow-up

assets/js/dashboard/stats/behaviours/conversions.tsx
```diff
@@ -0,0 +1,89 @@
...
+import {
+  hasConversionGoalFilter,
+  isRealTimeDashboard
+} from '../../util/filters'
+import { useDashboardStateContext } from '../../dashboard-state-context'
+import { QueryApiResponse, QueryResultRow } from '../../api'
+import { NonTimeDimension } from '../../stats-query'
+import { FilterInfo } from '../../components/drilldown-link'
+
+const BAR_COLOR = 'bg-red-50 group-hover/row:bg-red-100'
```
RobertJoonas: Could we export this from here and use in `props.tsx` and `special-goal-prop-breakdown.tsx` too?

assets/js/dashboard/stats/behaviours/conversions.tsx
```diff
@@ -0,0 +1,89 @@
...
+      dimensionLabel={reportConfig.dimensionLabel}
+      alwaysOnFilters={reportConfig.alwaysOnFilters}
+      DimensionElement={DimensionElement}
+      onDataReady={afterFetchData}
+      hideMetricsIfAllNull={['total_revenue', 'average_revenue']}
+    />
+  )
+}
+
+function getGoalsFilterInfo(
```
RobertJoonas: this appears to be an exact duplicate of the same function in `stats/modals/conversions.tsx`. Let's share a single function.

assets/js/dashboard/stats/modals/conversions.tsx
```diff
@@ -0,0 +1,80 @@
...
+      isRealtime: isRealTimeDashboard(dashboardState),
+      isDetailed: true,
+      isRevenueAvailable: false
+    }
+  )
+
+  /*global BUILD_EXTRA*/
+  const metrics = BUILD_EXTRA
+    ? [...baseMetrics, 'average_revenue' as const, 'total_revenue' as const]
+    : baseMetrics
```
RobertJoonas: With the CSV PR merged just now, I believe this could become:


```suggestion
  const metrics = reportConfig.getMetrics({isRevenueAvailable: BUILD_EXTRA})
```

assets/js/dashboard/stats/modals/details-breakdown.tsx
```diff
@@ -60,6 +60,18 @@ type DetailsBreakdownProps = SharedBreakdownReportProps & {
   searchEnabled?: boolean
   onDataReady?: (data: PaginatedData) => void
   DimensionElement: (props: DimensionCellProps) => ReactNode
+  /**
+   * When true (default), `percentage` is shown inline inside the Visitors
+   * cell rather than as its own column. Set to false for reports that want
+   * percentage as a separate breakdown column (e.g. custom properties).
+   */
+  bundlePercentageWithVisitors?: boolean
```
RobertJoonas: Both `bundlePercentWithVisitors` and `hideMetricsIfAllNull` are used in both index and details breakdowns. How about adding these fields into SharedBreakdownReportProps and documenting in a single place?

assets/js/dashboard/stats/modals/details-breakdown.tsx
```diff
@@ -157,14 +173,30 @@ export function DetailsBreakdown({
     [dashboardState, dimensions]
   )
 
+  const columnsHiddenForAllNull = useMemo((): Set<Metric> => {
```
RobertJoonas: Very similar to the one in `index-breakdown.tsx`, with the only exception being the shape of the data. Maybe there's a way to deduplicate it? A hook like `useHiddenMetricColumns`?

assets/js/dashboard/stats/behaviours/props.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const DimensionElement = useCallback(
+    (props: DimensionCellWithBarProps) => {
+      const value = props.row.dimensions[0]
+      return (
+        <DimensionCellWithBar
+          {...props}
+          barClassName={BAR_COLOR}
+          text={value}
+          getFilterInfo={() => ({
```
RobertJoonas: Could we use `makeGetFilterInfo` from `stats/modals/props.tsx` here to deduplicate?

assets/js/dashboard/stats/behaviours/props.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return (
+      <div className="flex-1 flex items-center justify-center font-medium text-gray-500 dark:text-gray-400">
+        No custom properties found
+      </div>
+    )
+  }
+
+  return (
+    <div className="w-full" style={{ minHeight: `${MIN_HEIGHT}px` }}>
+      <IndexBreakdown
```
RobertJoonas, line 75: Something we already talked about but leaving a note JIC: Should pass a bigger column width for desktop. Same in special goal prop breakdown.

assets/js/dashboard/stats/reports/index-breakdown.tsx
```diff
@@ -77,7 +91,9 @@ export function IndexBreakdown({
         dimensions,
         order_by: [
           ['visitors', 'desc'],
-          ...dimensions.map((dim): OrderByEntry => [dim, 'asc'])
+          ...dimensions
+            .filter((dim) => dim !== 'event:goal')
```
RobertJoonas: We're making sure here that we’re not trying to order by `event:goal` which is unsupported. This logic is currently duplicated in both index and details breakdowns. Would be nice to define that limitation in a single place, e.g.: reports-config.ts? Either by setting an explicit `dimensionOrderBy` field for the goals report, or just exporting a `const CANNOT_ORDER_BY_DIMENSIONS`

assets/js/dashboard/stats/reports/reports-config.ts
```diff
@@ -236,5 +237,36 @@ export const BREAKDOWN_REPORTS: Record<
...
+  }
+}
+
+export function customPropsReportConfig(
+  propKey: string
+): BreakdownReportConfig {
+  return {
+    dimensions: [`event:props:${propKey}` as NonTimeDimension],
+    metricsByContext: {
+      realtimeMetrics: ['visitors', 'events'],
```
RobertJoonas: Already talked about it but highlighting again JIC: missing conversion rate in realtime.

RobertJoonas (approved):

PR #6441, opened 2026-06-10
fix: ignore imported data when querying goals with custom props

ukutaht:
### Changes

Fixes [APP-217](https://sentry.plausible.io/organizations/plausible/issues/2246/?environment=prod&project=2&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20meta.value&referrer=issue-stream&sort=date)

Goals with custom props are not considered in the imported data query logic. Filtering or breaking down by such a goal on a site with imported data either:

- crashes — the goal filter emits meta.value[indexOf(meta.key, ...)] conditions against imported_custom_events, which has no meta columns → Ch.Error: UNKNOWN_IDENTIFIER, or
- over-counts — the event:goal breakdown matches imported rows by event name / page path only, attributing all imported events to a prop-restricted goal

Changes

- Imported.Base: skip imported data (with the standard unsupported_query warning) when any goal matching a top-level filter has custom props. Applied to all three table-decision paths, mirroring the existing scroll-goal handling.
- Stats.Goals.goal_join_data/1: emit "" into types and event_names_imports (consumed only by the imported merge) for custom-props goals, so the unfiltered goal breakdown no longer counts imported rows toward them — other goals in the breakdown keep their imported data.

### DIFF
--- a/lib/plausible/stats/goals.ex
+++ b/lib/plausible/stats/goals.ex
@@ -98,11 +98,20 @@ defmodule Plausible.Stats.Goals do
         goal_type = Plausible.Goal.type(goal)
         {prop_keys, prop_values} = Enum.unzip(goal.custom_props)
 
+        # `types` and `event_names_imports` are only used for matching goals
+        # against imported data. Imported tables are aggregated without custom
+        # properties, so goals with custom props must never match - "" ensures
+        # that.
+        queryable_from_imports? = not Plausible.Goal.has_custom_props?(goal)
+
         %{
           indices: [idx | acc.indices],
-          types: [to_string(goal_type) | acc.types],
+          types: [if(queryable_from_imports?, do: to_string(goal_type), else: "") | acc.types],
           # This will contain "" for non-event goals
-          event_names_imports: [to_string(goal.event_name) | acc.event_names_imports],
+          event_names_imports: [
+            if(queryable_from_imports?, do: to_string(goal.event_name), else: "")
+            | acc.event_names_imports
+          ],
           event_names_by_type: [event_name_by_type(goal_type, goal) | acc.event_names_by_type],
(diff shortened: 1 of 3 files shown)


lib/plausible/stats/goals.ex
```diff
@@ -98,11 +98,20 @@ defmodule Plausible.Stats.Goals do
         goal_type = Plausible.Goal.type(goal)
         {prop_keys, prop_values} = Enum.unzip(goal.custom_props)
 
+        # `types` and `event_names_imports` are only used for matching goals
```
zoldar, line 101: I'd consider creating a separate function like `goal_join_imported_data` and return only fields that are relevant in each respective context. As is, this is turning very hairball-like.

lib/plausible/stats/goals.ex
```diff
@@ -98,11 +98,20 @@ defmodule Plausible.Stats.Goals do
...
+        # against imported data. Imported tables are aggregated without custom
+        # properties, so goals with custom props must never match - "" ensures
+        # that.
+        queryable_from_imports? = not Plausible.Goal.has_custom_props?(goal)
+
         %{
           indices: [idx | acc.indices],
-          types: [to_string(goal_type) | acc.types],
+          types: [if(queryable_from_imports?, do: to_string(goal_type), else: "") | acc.types],
```
zoldar, line 109: Provided we split the logic, wouldn't it make more sense to filter out the custom prop goals right at the start?

ukutaht:
@zoldar better late than never :) Is this what you had in mind? https://github.com/plausible/analytics/pull/6441/changes/3101cf26eb7c47ce978ce2ab9726b7d8ff75d21b

zoldar (approved):
Nice ✨

PR #6442, opened 2026-06-10
Add ability to make the FE trigger automatic dashboard refresh

RobertJoonas:
### Changes

Resurrecting what was started in https://github.com/plausible/analytics/pull/5362.

Adds a new plug in the `:internal_stats_api` pipeline that adds an `x-api-version` response header. Incrementing the value will trigger an automatic dashboard refresh.

In order to not trigger repeated dashboard reloads during deployments, the `x-api-version` header gets the value of the minimum version across the cluster. Each node has `@api_version` compiled in. That same value gets passed into the FE via a `<meta>` tag. Without the meta tag in the DOM, the frontend defaults to `EXPECTED_API_VERSION = 0`, which:

* keeps the FE test suite working without having to worry about the api version
* makes sure that dashboard does not refresh when this PR itself is deployed

Whenever the FE version gets ahead of the BE (i.e. load the dashboard from a freshly deployed node while some app servers are still on the old version), we do not refresh the dashboard. Done by comparing version integers. 

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -7,6 +7,9 @@ import { serializeApiFilters } from './util/filters'
 import * as url from './util/url'
 import { MainGraphResponse } from './stats/graph/fetch-main-graph'
 import { CsvExportRequestBody } from './stats/csv-export/csv-export-body'
+import { maybeReloadForApiVersion } from './util/url-search-params'
+
+const EXPECTED_API_VERSION = '0'
 
 let abortController = new AbortController()
 let SHARED_LINK_AUTH: null | string = null
@@ -149,6 +152,12 @@ async function throwApiErrorIfNotOk(response: Response) {
 }
 
 async function handleApiResponse(response: Response) {
+  const apiVersion = response.headers.get('x-api-version')
+
+  if (apiVersion && apiVersion !== EXPECTED_API_VERSION) {
+    maybeReloadForApiVersion(window.location, EXPECTED_API_VERSION)
+  }
+
   await throwApiErrorIfNotOk(response)
   return response.json()
 }
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -149,8 +149,11 @@ async function throwApiErrorIfNotOk(response: Response) {
   }
 }
 
-async function handleApiResponse(response: Response) {
-  maybeReloadForApiVersion(window.location, response.headers)
+async function handleApiResponse(response: Response, opts: Record<'idempotent', boolean> = {idempotent: true}) {
+  if (opts.idempotent) {
+    maybeReloadForApiVersion(window.location, response.headers)
+  }
+
   await throwApiErrorIfNotOk(response)
   return response.json()
 }
@@ -264,5 +267,5 @@ export const mutation = async <
     body: fetchOptions.body,
     signal: abortController.signal
   })
-  return handleApiResponse(response)
+  return handleApiResponse(response, {idempotent: false})
 }
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -149,7 +149,10 @@ async function throwApiErrorIfNotOk(response: Response) {
   }
 }
 
-async function handleApiResponse(response: Response, opts: Record<'idempotent', boolean> = {idempotent: true}) {
+async function handleApiResponse(
+  response: Response,
+  opts: Record<'idempotent', boolean> = { idempotent: true }
+) {
   if (opts.idempotent) {
     maybeReloadForApiVersion(window.location, response.headers)
   }
@@ -267,5 +270,5 @@ export const mutation = async <
     body: fetchOptions.body,
     signal: abortController.signal
   })
-  return handleApiResponse(response, {idempotent: false})
+  return handleApiResponse(response, { idempotent: false })
 }
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -9,11 +9,6 @@ import { MainGraphResponse } from './stats/graph/fetch-main-graph'
 import { CsvExportRequestBody } from './stats/csv-export/csv-export-body'
 import { maybeReloadForApiVersion } from './util/url-search-params'
 
-const EXPECTED_API_VERSION =
-  document
-    .querySelector('meta[name="x-api-version"]')
-    ?.getAttribute('content') ?? '0'
-
 let abortController = new AbortController()
 let SHARED_LINK_AUTH: null | string = null
 
@@ -155,12 +150,7 @@ async function throwApiErrorIfNotOk(response: Response) {
 }
 
 async function handleApiResponse(response: Response) {
-  const currentApiVersion = response.headers?.get('x-api-version')
-
-  if (currentApiVersion && currentApiVersion !== EXPECTED_API_VERSION) {
-    maybeReloadForApiVersion(window.location, currentApiVersion)
-  }
-
+  maybeReloadForApiVersion(window.location, response.headers)
   await throwApiErrorIfNotOk(response)
   return response.json()
 }
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -155,10 +155,10 @@ async function throwApiErrorIfNotOk(response: Response) {
 }
 
 async function handleApiResponse(response: Response) {
-  const apiVersion = response.headers.get('x-api-version')
+  const currentApiVersion = response.headers.get('x-api-version')
 
-  if (apiVersion && apiVersion !== EXPECTED_API_VERSION) {
-    maybeReloadForApiVersion(window.location, EXPECTED_API_VERSION)
+  if (currentApiVersion && currentApiVersion !== EXPECTED_API_VERSION) {
+    maybeReloadForApiVersion(window.location, currentApiVersion)
   }
 
   await throwApiErrorIfNotOk(response)
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -9,7 +9,10 @@ import { MainGraphResponse } from './stats/graph/fetch-main-graph'
 import { CsvExportRequestBody } from './stats/csv-export/csv-export-body'
 import { maybeReloadForApiVersion } from './util/url-search-params'
 
-const EXPECTED_API_VERSION = '0'
+const EXPECTED_API_VERSION =
+  document
+    .querySelector('meta[name="x-api-version"]')
+    ?.getAttribute('content') ?? '0'
 
 let abortController = new AbortController()
 let SHARED_LINK_AUTH: null | string = null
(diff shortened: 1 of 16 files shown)


lib/plausible_web/plugs/internal_stats_api_version.ex
```diff
@@ -0,0 +1,24 @@
...
+  """
+  @behaviour Plug
+  import Plug.Conn
+
+  @impl true
+  def init(opts), do: opts
+
+  @api_version "0"
+
+  def api_version, do: @api_version
```
zoldar: Perhaps it would make sense to set that value in runtime config? The value could be exposed via `Plausible.internal_stats_api_version()`.

lib/plausible_web/plugs/internal_stats_api_version.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  @impl true
+  def call(conn, _opts \\ nil) do
+    put_resp_header(conn, "x-api-version", @api_version)
```
apata: suggestion, non-blocking: let's test that it's added to an internal stats endpoint like `/query`

assets/js/dashboard/api.ts
```diff
@@ -149,6 +155,12 @@ async function throwApiErrorIfNotOk(response: Response) {
 }
 
 async function handleApiResponse(response: Response) {
+  const currentApiVersion = response.headers?.get('x-api-version')
```
apata: nitpick, non-blocking: I think a separate utility like `getCurrentApiVersion` would be appropriate, next to other code related to this feature.

apata: nitpick, non-blocking: when are response headers not defined?

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -19,6 +19,51 @@ const LABEL_URL_PARAM_NAME = 'l'
...
+    return
+  }
+
+  const newSearch = addOrReplaceSearch(
+    windowLocation.search,
+    API_VERSION_RELOAD_PARAM_NAME,
+    currentApiVersion
+  )
+
+  console.log('API version mismatch detected, reloading...')
```
apata: nitpick, non-blocking: the other redirect is notifying with console.warn

assets/js/dashboard/api.ts
```diff
@@ -149,6 +155,12 @@ async function throwApiErrorIfNotOk(response: Response) {
...
+
+  if (currentApiVersion && currentApiVersion !== EXPECTED_API_VERSION) {
```
apata: suggestion, non-blocking: if we move the checks into `maybeReloadForApiVersion`, it can be better unit tested. The args for it can be windowLocation and responseHeaders

assets/js/dashboard/api.ts
```diff
@@ -149,6 +155,12 @@ async function throwApiErrorIfNotOk(response: Response) {
...
+    maybeReloadForApiVersion(window.location, currentApiVersion)
```
apata: issue, non-blocking: I think we should reload on idempotent requests only, not when updating segments etc.

assets/js/dashboard/stats/modals/filter-modal-row.js
```diff
@@ -7,11 +7,11 @@ import Combobox from '../../components/combobox'
 
 import {
   FILTER_OPERATIONS,
-  fetchSuggestions,
   isFreeChoiceFilterOperation,
   getLabel,
   formattedFilters
 } from '../../util/filters'
+import { fetchSuggestions } from '../../util/fetch-suggestions'
```
apata, line 14: question, blocking: is this change connected with the new feature?

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -19,6 +19,51 @@ const LABEL_URL_PARAM_NAME = 'l'
...
+/**
+ * Navigates to the current URL with `api_version_reloaded=<currentApiVersion>`
+ * appended, using `location.replace` so the pre-reload entry is not kept in
+ * browser history. Returns early without navigating when the param is already
+ * present, which prevents an infinite reload loop when the versions are
+ * permanently out of sync.
+ *
+ * BE: lib/plausible_web/plugs/internal_stats_api_version.ex
+ */
+export function maybeReloadForApiVersion(
```
apata, line 47: issue, blocking: I think this must be unit tested quite extensively.

The FE makes multiple requests for data concurrently. The requests may hit different instances of the BE. During rolling deploys, the BE instances may report different API versions. What should happen? 

I suggest we spec it out with a jest test suite that includes the concurrent requests during rolling deploys scenario.

I suspect we may need to set a flag that we're about to update to have deterministic behavior, e.g. at module level, `let reloadingDueToApiVersionChange`

assets/js/dashboard/util/url-search-params.ts
```diff
@@ -19,6 +19,51 @@ const LABEL_URL_PARAM_NAME = 'l'
...
+    currentApiVersion
+  )
+
+  console.log('API version mismatch detected, reloading...')
+  windowLocation.replace(
+    `${windowLocation.pathname}${newSearch}${windowLocation.hash}`
+  )
+}
+
+function addOrReplaceSearch(
```
apata: issue, blocking: This regex-based function is not encoding the value, so as a utility with such a general name, it's actually quite risky. Can we use the utils `parseSearch` and `stringifySearch` to create the new link?

lib/plausible_web/plugs/internal_stats_api_version.ex
RobertJoonas: @zoldar could you elaborate on the benefits you have in mind here?

assets/js/dashboard/api.ts
RobertJoonas: In Jest tests. I figured the tests shouldn't have to worry about providing empty response headers just in order to make this work.

lib/plausible_web/plugs/internal_stats_api_version.ex
zoldar: Okay, on a second thought, runtime config does not make sense as it's something that only changes together with code. Version in `Plausible.internal_stats_api_version()` could be nice for discoverability but I have no strong feelings about it. Feel free to ignore.

assets/js/dashboard/util/url-search-params.ts
RobertJoonas: > During rolling deploys, the BE instances may report different API versions. What should happen?

Uff, yeah good catch! I feel like this is now turning into an infrastructure problem. So basically, a dashboard reload is useless until there's old code still around on at least one app server. We definitely don't want the dashboard to keep reloading itself until our deployment finishes.

I guess the version bump should happen separate from the code change then. I'm now thinking about storing the version in Postgres and bumping it via remote shell once all deployments have finished, maybe?

assets/js/dashboard/stats/modals/filter-modal-row.js
RobertJoonas: Kind of. In `api.ts`, importing from `url-search-params.ts` created a circular dependency:

* `filters.js` imports from `api.ts`
* `api.ts` imports from `url-search-params.ts` (added in this PR)
* `url-search-params.ts` imports from `url-search-params-v1.ts`
* `url-search-params-v1.ts` imports from `filters.js`

This gets rid of it by eliminating the first -- make `filters.js` not rely on `api.ts`

lib/plausible_web/plugs/internal_stats_api_version.ex
RobertJoonas: https://github.com/plausible/analytics/pull/6442/commits/e6db30cdab01d5c1c6c20d74c093fe24369ca9c2

assets/js/dashboard/api.ts
RobertJoonas: Makes sense. Even more so because we also need to compare versions as integers on the frontend. The function I landed on is:

```javascript
function getCurrentApiVersion(responseHeaders: Headers): number | null {
  const versionString = responseHeaders?.get('x-api-version')
  return versionString ? parseInt(versionString, 10) : null
}
```

assets/js/dashboard/util/url-search-params.ts
RobertJoonas: Updated to `console.warn` now.

assets/js/dashboard/api.ts
RobertJoonas: Agreed. Did that now, and added unit tests as well.

assets/js/dashboard/api.ts
RobertJoonas: Nice catch! Did that in https://github.com/plausible/analytics/pull/6442/commits/80f3c803208cd7f1e0c3ea28afa60987f24f6f33

assets/js/dashboard/util/url-search-params.ts
RobertJoonas: I've landed on `:rpc.multicall` to get the effective API version as the minimum across the cluster, falling back to the current node version on non-cluster setups or when the multicall fails. the "effective version" value is cached for 30s.

Basically, the idea is that when we start rolling out new code, the effective version will be the old one for as long as there's at least one app node still at that version. This guarantees that all nodes have updated once we trigger the automatic refresh.

However, the client can still (re)load the dashboard manually while deployment is in progress, meaning they'd get the new FE code with requests served by the old BE. Although there's a version mismatch there, we don't reload the page because the frontend API version is ahead. Basically the frontend is in the correct state, and just waiting for the backend to update.

assets/js/dashboard/util/url-search-params.ts
RobertJoonas: https://github.com/plausible/analytics/pull/6442/commits/5cf4782348ad33c1025e8869e7ec55aea82fac4b

lib/plausible/internal_stats_api_version.ex
```diff
@@ -0,0 +1,66 @@
...
+  @api_version 0
+
+  @refresh_interval :timer.seconds(30)
+
+  @spec api_version() :: non_neg_integer()
+  def api_version, do: @api_version
+
+  @spec effective_version() :: non_neg_integer()
+  def effective_version() do
+    GenServer.call(__MODULE__, :get)
```
zoldar: Although it's highly unlikely this process might go bad or become a bottleneck, I'd consider persisting the effective version in ETS instead and attempting to fetch it here with a graceful fallback, just in case. It could be done in a way similar to Salts https://github.com/plausible/analytics/blob/master/lib/plausible/session/salts.ex#L10-L27

test/test_helper.exs
```diff
@@ -3,6 +3,7 @@ if not Enum.empty?(Path.wildcard("lib/**/*_test.exs")) do
 end
 
 {:ok, _} = Application.ensure_all_started(:ex_machina)
+Plausible.InternalStatsApiVersion.start_link()
```
zoldar: This should be redundant as the app supervision tree is brought up during tests?

zoldar (approved):
:sparkles:

test/test_helper.exs
RobertJoonas: https://github.com/plausible/analytics/pull/6442/commits/6c0965508b26ee65819bf471a3924183c7ec355e

PR #6444, opened 2026-06-11
Fix issue with dropdown interactions in e2e tests

apata:
### Changes

In local development, I've noticed two tests in behaviours.spec.ts being flaky. This should fix it.


zoldar (approved):
❤️

PR #6445, opened 2026-06-11
Refactor to pass around metrics with their labels

apata:
### Changes

Refactors dashboard to pass around metrics around the app not by key only, but by their full name. 

Solves the nuisance of picking the right metrics for a particular context and then needing to generate the labels for them after the fact.

We know when we prepare the query for the exit pages report that in this context, visitors means unique exits.

### Tests

### Changelog

### Documentation

### Dark mode


apata:
`VISITORS` and `VISITORS_AS_UNIQUE_CONVERSIONS` are effectively unique metrics from the perspective of the FE, even though both are queried from the BE with the same metric key: `visitors`. 

This PR attempted to codify it, packing the uniqueness indicator (label) and the metric key together, but the approach I picked has drawbacks. 

E.g. `PERCENTAGE`, `PERCENTAGE_AS_PERCENT_SIGN` are not unique metrics from the perspective of the FE, they have the same meaning. With these two, in one special case, we just want to render the `%` sign as column header, a UI concern that should be the responsibility of that particular UI element.

I think there's another uniqueness indicator that can be used to solve this, something like `{metric: "visitors", alias: "visitors_as_unique_conversions"}`. It would allow us to pick the right label for the metrics without needing to derive whether the metric is unique all over again. 

Until we agree on what the uniqueness indicator is, closing this PR.

PR #6453, opened 2026-06-17
Fix report out of view issue in e2e tests

apata:
### Changes

There was some flakiness with tests of reports that are wrapped by LazyLoader. 
It's a component that enables the reports and the queries driving those reports only when the report is in view.
 
Scrolling the tab buttons into view occasionally scrolled the page exactly to the bottom of the tab buttons, so not a single px of the report itself was in view. Consequently, the report was never enabled, no queries were made and the assertions failed.

The fix is to scroll more, not just to the tab buttons, but a few px more. Rather than working with a certain px count, I decided to add an empty h=0 div to the bottom of each report that we can scroll to.

To click a button, it doesn't have to be in view, playwright scrolls to it automatically (https://playwright.dev/docs/input#mouse-click:~:text=scroll%20the%20element%20into%20view), so there's no harm done by changing the existing scrollIntoView calls.

### DIFF
--- a/assets/js/dashboard/stats/reports/report-layout.tsx
+++ b/assets/js/dashboard/stats/reports/report-layout.tsx
@@ -19,6 +19,7 @@ export function ReportLayout({
       )}
     >
       {children}
+      <div data-testid="report-end"></div>
     </div>
   )
 }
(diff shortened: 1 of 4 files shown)


zoldar (approved):
👍 thanks for resolving this

PR #6454, opened 2026-06-18
Handle :EXIT messages in ingestion WriteBuffer

### LINKED ISSUE #6382: FunctionClauseError in WriteBuffer.handle_info/2 on {:EXIT, port, :normal}
### Past Issues Searched

### Issue is a Bug Report

Self-hosting

### Describe the bug

We are seeing intermittent write buffer crashes on self-hosted Plausible `v3.2.0`.

## Error
```text
[notice] Flushing Elixir.Plausible.Session.WriteBuffer buffer before shutdown...
[error] GenServer Plausible.Session.WriteBuffer terminating
** (FunctionClauseError) no function clause matching in Plausible.Ingestion.WriteBuffer.handle_info/2
(plausible 0.0.1) lib/plausible/ingestion/write_buffer.ex:63:
Plausible.Ingestion.WriteBuffer.handle_info({:EXIT, #Port<...>, :normal}, %{name: Plausible.Session.WriteBuffer, ...})
Last message: {:EXIT, #Port<...>, :normal}
```
The Plausible pod stays up, but the write buffer process crashes and restarts.

This looks like an unhandled `{:EXIT, port, :normal}` message in `handle_info/2`.

I have also seen separate ClickHouse connection messages (Ch.Connection disconnected: socket closed), but this seems to be a distinct bug: even if a connection/client event is the trigger, the immediate failure is the FunctionClauseError.

**Questions**
- Is this a known issue?
- Can buffered writes be lost when this happens?

### Expected behavior

FunctionClauseError should not be happen.

### Screenshots

_No response_

### Environment

```markdown
- Plausible v3.2.0
- self-hosted on Kubernetes
- ClickHouse backend
```

SAY-5:
### Changes

`Plausible.Ingestion.WriteBuffer` calls `Process.flag(:trap_exit, true)` so `terminate/2` can flush the buffer on shutdown, but `handle_info/2` only matched `:tick`. When a linked port or process (the ClickHouse connection socket) exits, the GenServer receives `{:EXIT, port, :normal}` and crashes with a `FunctionClauseError`, then restarts. This adds a `handle_info/2` clause that ignores `:EXIT` messages so the buffer keeps running. Fixes #6382.


lib/plausible/ingestion/write_buffer.ex
```diff
@@ -66,6 +66,13 @@ defmodule Plausible.Ingestion.WriteBuffer do
     {:noreply, %{state | buffer: [], buffer_size: 0, timer: timer}}
   end
 
+  def handle_info({:EXIT, _from, _reason}, state) do
+    # We trap exits so terminate/2 can flush on shutdown. Linked ports and
```
aerosol: This comment is unnecessary, let's log the reason instead (`warning` level is fine).

test/plausible/ingestion/write_buffer_test.exs
```diff
@@ -0,0 +1,36 @@
...
+  @opts WriteBuffer.compile_time_prepare(Plausible.ClickhouseEventV2)
+
+  test "keeps running when a linked process exits while trapping exits" do
+    opts =
+      @opts
+      |> Map.take([:header, :insert_sql, :insert_opts])
+      |> Map.to_list()
+      |> Keyword.put(:name, :"write_buffer_exit_#{System.unique_integer([:positive])}")
+      # large interval so the empty buffer never flushes to ClickHouse during the test
+      |> Keyword.put(:flush_interval_ms, 60_000)
```
aerosol: What do you think about making this a bit more readable? Since `compile_time_prepare` result is a map, let's build a keyword list verbatim and only fetch necessary fields via the dot-notation access.

Something like:

```elixir
test "...", %{test: test} do
opts = [
  header: @compile_time_opts.header,
  insert_sql: @compile_time_opts.insert_sql,
  insert_opts: @compile_time_opts.insert_opts,
  name: test, # unique atom is at our disposal already
  flush_interval_ms: 1_000_000_000 # no need to ever flush automatically on slow CI
]
```

aerosol (changes requested):
One last round of changes if you don't mind and we're good to go

lib/plausible/ingestion/write_buffer.ex
aerosol: Bonus points for asserting the log entry in test 🤗

aerosol (approved):

PR #6455, opened 2026-06-18
Dashboard csv export v2 continued

RobertJoonas:
### Changes

Add `custom_props.csv` and `conversions.csv` to the v2 CSV export.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/dashboard/csv_export.ex
```diff
@@ -211,4 +242,80 @@ defmodule Plausible.Stats.Dashboard.CsvExport do
...
+    with {:ok, parsed} <- QueryParser.parse(raw_params),
+         {:ok, query} <- QueryBuilder.build(site, parsed, debug_metadata) do
+      {:ok, Query.set(query, order_by: [{:visitors, :desc}, {"event:prop_key", :asc}])}
+    end
+  end
+
+  defp get_custom_prop_keys(site, query) do
+    case Plausible.Stats.Filters.get_toplevel_filter(query, "event:props:") do
+      [_op, "event:props:" <> key | _rest] ->
+        [key]
```
apata, line 264: issue, non-blocking (buggy on prod as well): For a dashboard filter like `["is_not", "event:props:theme", ["system"]]`, it focuses the exported data to exactly that prop.

lib/plausible/stats/dashboard/csv_export.ex
```diff
@@ -211,4 +242,80 @@ defmodule Plausible.Stats.Dashboard.CsvExport do
...
+
+      _ ->
+        %QueryResult{results: results} = QueryRunner.run(site, query)
+
+        Enum.map(results, fn r -> hd(r.dimensions) end)
```
apata, line 269: issue, blocking: I think we should limit the amount of prop keys we query to prevent too many concurrent queries being run on CH in the next step.

lib/plausible/stats/dashboard/csv_export.ex
```diff
@@ -211,4 +242,80 @@ defmodule Plausible.Stats.Dashboard.CsvExport do
...
+        {:cont, {:ok, elem(acc, 1) ++ [{prop_key, query}]}}
+      else
+        error_tuple -> {:halt, error_tuple}
+      end
+    end)
+  end
+
+  defp get_custom_props_csv(_site, []), do: nil
+
+  defp get_custom_props_csv(site, prop_value_queries_by_prop_key) do
```
apata, line 298: suggestion, non-blocking: This function has a lot of responsibilities so it's a bit hard to read. I think it could be split along these lines

- format header
- run queries
- format individual query result rows to array that matches header shape
- flatten output
- write to csv using a particular API

assets/js/dashboard/stats/csv-export/csv-export-body.ts
```diff
@@ -114,6 +119,10 @@ export function createCsvExportRequestBody(
           ? PAGE_FILTERED_VISITORS_CSV_METRICS
           : DEFAULT_VISITORS_CSV_METRICS
     },
+    'custom_props.csv': {
+      dimensions: ['event:props:*'],
+      metrics: getCustomPropsMetrics({ hasConversionGoalFilter: isGoalFilter })
```
apata: suggestion, non-blocking: The function `getCustomPropsMetrics` depends on `isRevenueAvailable`, which isn't passed. 

It seems we actually want to say that in the CSV report, revenue is not available (`isRevenueAvailable: false`), or do we? With MetricContext allowing undefined, I can't actually tell what the intent is for this CSV report.

I think we shouldn't allow undefined in MetricContext. 

Shorter args list at call sites is not worth losing explicitly declared intent at call sites.

lib/plausible/stats/dashboard/csv_export.ex
RobertJoonas: Yeah, buggy indeed. One of the many things wrong with the current CSV export. I think it's fine to consider this out of scope here though.

assets/js/dashboard/stats/csv-export/csv-export-body.ts
RobertJoonas: > It seems we actually want to say that in the CSV report, revenue is not available (isRevenueAvailable: false), or do we? With MetricContext allowing undefined, I can't actually tell what the intent is for this CSV report.

Yeah, the intent is that revenue is not available. No specific reason other than sticking to legacy behaviour.

> I think we shouldn't allow undefined in MetricContext.

IMHO it doesn't (currently) hurt readability because the behaviour between `undefined` and `false` is equivalent. Everything in the `MetricContext` that is irrelevant from the caller's perspective can simply be left unspecified and get interpreted as false, which, at least to me, is quite intuitive.

This would change with the addition of a context variable that would have a default value of true, but that's not the case, and I doubt it ever will be?

That said though, I've written out `isRevenueAvailable: false` to make this specific case explicit. https://github.com/plausible/analytics/pull/6455/commits/1632261e8c2d1c16337fc74a49521c967d57a58c.

lib/plausible/stats/dashboard/csv_export.ex
RobertJoonas: I've split it into smaller functions here: https://github.com/plausible/analytics/pull/6455/commits/8f251f145860dd7742162774c4ba32d6c3a7c114

lib/plausible/stats/dashboard/csv_export.ex
RobertJoonas: Yeah, makes sense. The query is currently unlimited but the queries are also executed synchronously. I've set a limit of 25 prop keys now in https://github.com/plausible/analytics/pull/6455/commits/be7997c27c9294fa2302c54c00eba28f34dc8cf1, which should be fine.

apata (approved):

PR #6460, opened 2026-06-23
Add `current_plan` as custom property to track

sanne-san:
### Changes

- Having a `current_plan` custom property allows us to see an accurate conversion rate for plan-specific features, such as User journeys.
- The prop is called `current_plan` to distinguish it from the `plan` custom property used to track CTA clicks on the landing page.

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/dogfood.ex
```diff
@@ -68,11 +68,28 @@ defmodule PlausibleWeb.Dogfood do
 
   defp location_override(_), do: nil
 
-  defp custom_properties(%{current_user: user}) when is_map(user) do
-    %{logged_in: true, theme: user.theme}
+  defp custom_properties(%{current_user: user} = assigns) when is_map(user) do
+    %{
+      logged_in: true,
+      theme: user.theme,
+      current_plan: current_plan(assigns[:current_team])
```
aerosol, line 75: Would be better to put that in assigns before dashboard is rendered and dogfooding triggered, via some plug preferably, so we don't do redundant queries over and over. And are we OK with publicly exposing that data?

sanne-san: Yes, confirmed with @ukutaht on exposing this data 👌

aerosol (approved):

aerosol (changes requested):
Sorry for misguiding you. Since `Dogfood` module is only called once and all its functions operate on assigns, and, as it turns out, `AuthPlug` is already preloading subscription to `current_team` we can do what you've initially had except we don't need to preload the subscription anymore. `current_plan` could match on `(%Team{subscription: s} = team) when not is_nil(s)` to fetch the plan. The extra query for enterprise plans is fine.

aerosol (approved):

PR #6462, opened 2026-06-25
Better error message change domain

aerosol:
Two UX improvements for domain change:

1. Owning one.example.com and two.example.com, it's now possible to swap the names via triple rename. This used to be impossible due to domain_changed_from constraint clash.
2. Better error message if trying to change site's domain to an already owned one

2026-06-25_14-56-45_vice

Migration will be extracted once review is concluded.


priv/repo/migrations/20260625000000_allow_same_team_domain_swap.exs
```diff
@@ -0,0 +1,37 @@
+defmodule Plausible.Repo.Migrations.AllowSameTeamDomainSwap do
+  use Ecto.Migration
+
+  def up do
+    execute("""
+    CREATE OR REPLACE FUNCTION check_domain() RETURNS TRIGGER AS $$
+    BEGIN
+      IF EXISTS (
+        SELECT 1 FROM sites
+         WHERE (NEW.domain = domain_changed_from AND NEW.id != id AND (team_id IS NULL OR NEW.team_id IS NULL OR team_id != NEW.team_id))
```
aerosol, line 10: Technically we take a good care to prevent `site.team_id` from ever being `NULL` (wondering if we can now use regular DB constraint to guarantee that), but the additional check here shouldn't hurt.

RobertJoonas (approved):

priv/repo/migrations/20260625000000_allow_same_team_domain_swap.exs
zoldar: Yeah, that probably was on teams post-release cleanup list (and _maybe_ blocked by syncing with CE) and we forgot about it eventually 🙈

zoldar (approved):

PR #6463, opened 2026-06-26
Bugfix: sites ellipsis menu and allowed roles

aerosol:
Site settings should be only seen by editors, admins and owners.
The only option for billing users should be the capability of pinning sites via the ellipsis menu. Both were broken.

### DIFF
--- a/lib/plausible_web/live/sites.ex
+++ b/lib/plausible_web/live/sites.ex
@@ -573,7 +573,7 @@ defmodule PlausibleWeb.Live.Sites do
       </.unstyled_link>
 
       <div class="absolute right-1 top-3.5">
-        <.ellipsis_menu site={@site} can_manage?={List.first(@site.memberships).role != :viewer} />
+        <.ellipsis_menu site={@site} can_manage?={List.first(@site.memberships).role in [:owner, :admin, :editor]} />
       </div>
     </li>
     """
--- a/lib/plausible_web/live/sites.ex
+++ b/lib/plausible_web/live/sites.ex
@@ -573,7 +573,10 @@ defmodule PlausibleWeb.Live.Sites do
       </.unstyled_link>
 
       <div class="absolute right-1 top-3.5">
-        <.ellipsis_menu site={@site} can_manage?={List.first(@site.memberships).role in [:owner, :admin, :editor]} />
+        <.ellipsis_menu
+          site={@site}
+          can_manage?={List.first(@site.memberships).role in [:owner, :admin, :editor]}
+        />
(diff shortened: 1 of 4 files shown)


test/plausible/sites_test.exs
```diff
@@ -268,4 +268,29 @@ defmodule Plausible.SitesTest do
       assert Sites.get_for_user!(user, domain, include_consolidated?: true)
     end
   end
+
+  describe "toggle_pin/2" do
+    test "billing role team member can pin a site" do
+      owner = new_user()
+      site = new_site(owner: owner)
+      team = site.team |> Plausible.Teams.complete_setup()
+      billing_user = add_member(team, role: :billing)
```
apata, line 277: suggestion, non-blocking: I think we should loop over all roles that should have this capability (quite easy with `add_site_guest_or_team_member` util) and assert that they have pin/unpin capability. Right now, reading this spec file as a whole, it appears pinning should be limited to billing role

lib/plausible_web/live/sites.ex
```diff
@@ -573,7 +573,10 @@ defmodule PlausibleWeb.Live.Sites do
       </.unstyled_link>
 
       <div class="absolute right-1 top-3.5">
-        <.ellipsis_menu site={@site} can_manage?={List.first(@site.memberships).role != :viewer} />
+        <.ellipsis_menu
+          site={@site}
+          can_manage?={List.first(@site.memberships).role in [:owner, :admin, :editor]}
```
apata, line 578: suggestion, non-blocking: Relying on List.first to access current user seems like a potential source of capabilities bugs. I think we should have a more direct way of accessing current user role.

test/plausible_web/live/sites_test.exs
```diff
@@ -705,6 +705,78 @@ defmodule PlausibleWeb.Live.SitesTest do
...
+      {:ok, _lv, html} = live(conn, "/sites?__team=#{team.identifier}")
+
+      settings_selector =
+        ~s|li[data-domain="#{site.domain}"] a[href$="/settings/general"]|
+
+      refute element_exists?(html, settings_selector)
+    end
+  end
+
+  describe "ellipsis menu settings visibility" do
```
apata, line 745: nitpick, non-blocking: This describe block doesn't cover site guests.

lib/plausible_web/live/sites.ex
aerosol: It has been always a list of one

apata (approved):

PR #6468, opened 2026-06-29
Prevent special goal renames

aerosol:
This PR locks special goal renames so they don't lose their front-end meaning.
In case of goals whose display names fall within the special set, but their corresponding event names don't, we allow the rename in case a manual status regain is needed.

Ref https://app.basecamp.com/5308029/buckets/36789884/card_tables/cards/9834018285
Ref #6464

### DIFF
--- a/test/plausible/goals_test.exs
+++ b/test/plausible/goals_test.exs
@@ -520,6 +520,79 @@ defmodule Plausible.GoalsTest do
            }
   end
 
+  test "update/2 prevents renaming event_name of a special goal" do
+    site = new_site()
+
+    for event_name <- Plausible.Goal.special_goals() do
+      {:ok, goal} = Goals.create(site, %{"event_name" => event_name})
+
+      assert {:error, changeset} =
+               Goals.update(goal, %{"event_name" => "Renamed #{event_name}"})
+
+      assert {"cannot be changed for an automated goal", _} = changeset.errors[:event_name]
+    end
+  end
+
+  test "update/2 prevents renaming display_name of a special goal to a non-canonical value" do
+    site = new_site()
+
+    for event_name <- Plausible.Goal.special_goals() do
+      {:ok, goal} = Goals.create(site, %{"event_name" => event_name})
+
+      assert {:error, changeset} =
+               Goals.update(goal, %{"display_name" => "Renamed #{event_name}"})
+
+      assert {"cannot be changed for an automated goal", _} = changeset.errors[:display_name]
+    end
+  end
+
+  test "update/2 allows restoring display_name of a special goal to its canonical value" do
+    site = new_site()
+    {:ok, goal} = Goals.create(site, %{"event_name" => "File Download"})
+
+    # Simulate a previously broken goal by directly updating the DB
+    Plausible.Repo.update_all(
+      Ecto.Query.where(Plausible.Goal, id: ^goal.id),
+      set: [display_name: "My File Downloads"]
+    )
+
+    broken_goal = Plausible.Repo.reload!(goal)
+    assert broken_goal.display_name == "My File Downloads"
+
+    assert {:ok, fixed} = Goals.update(broken_goal, %{"display_name" => "File Download"})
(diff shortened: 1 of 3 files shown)


test/plausible/goals_test.exs
```diff
@@ -520,6 +520,79 @@ defmodule Plausible.GoalsTest do
...
+
+    assert {:ok, updated} =
+             Goals.update(goal, %{
+               "event_name" => "File Download",
+               "display_name" => "File Download",
+               "custom_props" => %{"0" => "path"}
+             })
+
+    assert updated.custom_props == %{"0" => "path"}
+  end
```
apata, line 578: question, non-blocking: I know it's allowed right now, but now that we're looking into special goals, should we keep allowing this? Special goals with some user-defined properties seem likely to bite us in the future.

test/plausible/goals_test.exs
```diff
@@ -520,6 +520,79 @@ defmodule Plausible.GoalsTest do
...
+
+  test "update/2 allows updating non-name fields of a special goal" do
+    site = new_site()
+    {:ok, goal} = Goals.create(site, %{"event_name" => "File Download"})
+
+    assert {:ok, updated} =
+             Goals.update(goal, %{
+               "event_name" => "File Download",
+               "display_name" => "File Download",
+               "custom_props" => %{"0" => "path"}
```
apata: meganitpick, non-blocking: `custom_props` here could be more realistic `%{"path" => "/portfolio.pdf"}` or similar

test/plausible/goals_test.exs
aerosol: bite us how?

apata: Let's say we want to make File Download dashboard component more special. Instead of relying on the system-defined setup for the goal that's the same for everyone, we need to take into account variants where the otherwise system-defined goal has arbitrary custom props defined.

test/plausible_web/live/goal_settings/form_test.exs
```diff
@@ -332,6 +332,38 @@ defmodule PlausibleWeb.Live.GoalSettings.FormTest do
...
+      lv |> element(~s/button#edit-goal-#{goal.id}/) |> render_click()
+
+      lv
+      |> element("#goals-form-modalseq0 form")
+      |> render_submit(%{goal: %{event_name: "File Download", display_name: "My Downloads"}})
+
+      assert render(lv) =~ "cannot be changed for an automated goal"
+    end
+
     test "renders error when goal is invalid", %{conn: conn, site: site} do
```
apata, line 367: meganitpick, non-blocking: I see now that there's several ways for the goal to be invalid, this test could be renamed to something like `"renders error when renaming goal to be the same as an existing goal name"`

apata (approved):

test/plausible/goals_test.exs
aerosol: Or ignore any unrecognizable configuration. Nothing prevents you from sending ANY goal data. There is no way to tell whether this is legitimate event from the tracker, matching an auto-created goal or just random payload that happens to be named "File Download" with any custom props attached. We are relying on display names to recognize special goals, nothing is guaranteed in such setup.

test/plausible_web/live/goal_settings/form_test.exs
aerosol: IMHO at the form level, it does not matter, we only care if the changeset error shows up, whatever that might be. The test is fine. Lower level tests can go into details.

test/plausible/goals_test.exs
aerosol: d52c8b20e8

test/plausible/goals_test.exs
aerosol: Also, special goals with custom props work fine, granted there's data. If you use the seeded db and attach `logged_in: true` to "Outbound Link: Click", the report will render narrowed down to that. I don't see why should we prevent that based on pure speculation.

PR #6466, opened 2026-06-29
Code cleanup after successful dashboard to api v2 migration

RobertJoonas:
### Changes

Depends on https://github.com/plausible/analytics/pull/6455. Cleans up old FE and BE code after a successful dashboard to Stats API v2 migration.

### Tests
- [x] Automated tests have been added

### Changelog
- [x] This PR does not make a user-facing change

### Documentation
- [x] This change does not need a documentation update

### Dark mode
- [x] This PR does not change the UI

PR #6471, opened 2026-06-30
A few more refactoring cleanups

RobertJoonas:
### Changes

Some follow-ups to https://github.com/plausible/analytics/pull/6466:

* The `Stats.{Aggregate, Breakdown, Timeseries}` modules now only exist because of the public legacy API. Add them under a Legacy module scope.
* Cleaning up the query struct:
  * Remove the `interval` field
  * Remove the default `20_000_000` value of `sample_threshold`. The field itself is still in use, but the struct default never takes effect because query builders always call `Sampling.put_threshold`
* Get rid of `internal_query_api_test.exs` by moving the tests into `top_stats_test.exs`

### Tests

### Changelog

### Documentation

### Dark mode


aerosol (approved):
Nice. That sampling threshold value was quite misleading to keep.

PR #6474, opened 2026-06-30
Extract building query date ranges

apata:
### Changes

Pure refactor. This extracts the logic to build query date ranges. Needed to query other data in the time period of the dashboard, like annotations, without building the full stats query.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/stats/query_builder.ex
+++ b/lib/plausible/stats/query_builder.ex
@@ -15,10 +15,9 @@ defmodule Plausible.Stats.QueryBuilder do
     TableDecider,
     DateTimeRange,
     QueryError,
-    QueryInclude
+    QueryInclude,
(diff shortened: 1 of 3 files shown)


aerosol (approved):
Just making sure if it's intentional to make all the previously private functions public now?

apata:
> Just making sure if it's intentional to make all the previously private functions public now?

@aerosol, valid question! I really wanted to doctest these pretty vital functions (query period wrong -> stats wrong). It's trivial to do so with the new function signature not taking site, just what the function actually depends on. It's inconvenient to need to make them public to do so, but in my mind it's a lesser evil. They do stand on their own as well, so it's not really dangerous to make them public.

PR #6479, opened 2026-07-02
Update auth pages visually

sanne-san:
### Changes

- Create new auth layout context that looks visually cleaner, and hides the header and footer
- Transform all auth pages to use the new context. This includes:
  - Login page
  - Signup page
  - Password reset page
  - Password reset request page
  - Password reset request success page
  - Password reset error page
  - Activate account page
  - Verify 2FA page
  - Verify 2FA recovery code page
  - SSO login page
  - SSO provision notice page
  - SSO provision issue page
  - Invitation expired page
- Tighten copy across auth pages to be more consistent and concise
- Remove password conformation field from signup page
- Extract password field to a new component with visibility toggle
- Extract Plausible logo to a new component
- Ensure first field on any page is in focus

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/components/two_factor.ex
+++ b/lib/plausible_web/components/two_factor.ex
@@ -25,51 +25,28 @@ defmodule PlausibleWeb.Components.TwoFactor do
   attr :field, :any, required: true
   attr :class, :string, default: ""
   attr :show_button?, :boolean, default: true
+  attr :autofocus?, :boolean, default: false
 
   def verify_2fa_input(assigns) do
-    input_class =
-      "font-mono tracking-[0.5em] w-36 pl-5 font-medium shadow-sm focus:ring-indigo-500 focus:border-indigo-500 block border-gray-300 dark:border-gray-500 dark:text-gray-200 dark:bg-gray-900 rounded-l-md"
-
-    input_class =
-      if assigns.show_button? do
-        input_class
-      else
-        [input_class, "rounded-r-md"]
-      end
-
-    assigns = assign(assigns, :input_class, input_class)
     assigns = assign(assigns, :field, assigns[:form][assigns[:field]])
 
     ~H"""
-    <div class={[@class, "flex items-center"]}>
-      <input
-        type="text"
-        name={@field.name}
-        value={@field.value}
-        autocomplete="off"
-        class={@input_class}
-        oninput={
-          if @show_button? do
-            "this.value=this.value.replace(/[^0-9]/g, ''); if (this.value.length >= 6) document.getElementById('#{@id}').focus()"
-          else
-            "this.value=this.value.replace(/[^0-9]/g, '');"
-          end
-        }
-        onclick="this.select();"
+    <div class={[@class, "flex flex-col gap-y-6 items-center"]}>
+      <.otp_input
+        field={@field}
+        length={6}
         oninvalid={@show_button? && "document.getElementById('#{@id}').disabled = false"}
-        maxlength="6"
-        placeholder="••••••"
-        required="required"
+        autofocus={@autofocus? && "autofocus"}
       />
       <.button
         :if={@show_button?}
         type="submit"
         id={@id}
         mt?={false}
-        class="rounded-l-none [&>span.label-enabled]:block [&>span.label-disabled]:hidden [&[disabled]>span.label-enabled]:hidden [&[disabled]>span.label-disabled]:block"
+        class="w-full [&>span.label-enabled]:block [&>span.label-disabled]:hidden [&[disabled]>span.label-enabled]:hidden [&[disabled]>span.label-disabled]:block"
       >
         <span class="label-enabled pointer-events-none">
-          Verify &rarr;
+          Verify
         </span>
 
         <span class="label-disabled">
--- a/lib/plausible_web/components/two_factor.ex
+++ b/lib/plausible_web/components/two_factor.ex
@@ -26,18 +26,24 @@ defmodule PlausibleWeb.Components.TwoFactor do
   attr :class, :string, default: ""
   attr :show_button?, :boolean, default: true
   attr :autofocus?, :boolean, default: false
+  attr :error, :string, default: nil
 
   def verify_2fa_input(assigns) do
     assigns = assign(assigns, :field, assigns[:form][assigns[:field]])
 
     ~H"""
     <div class={[@class, "flex flex-col gap-y-6 items-center"]}>
-      <.otp_input
-        field={@field}
-        length={6}
-        oninvalid={@show_button? && "document.getElementById('#{@id}').disabled = false"}
-        autofocus={@autofocus? && "autofocus"}
-      />
+      <div class="flex flex-col gap-y-2 w-full">
+        <.otp_input
+          field={@field}
+          length={6}
+          oninvalid={@show_button? && "document.getElementById('#{@id}').disabled = false"}
+          autofocus={@autofocus? && "autofocus"}
+        />
+        <p :if={@error} class="text-xs text-red-500 text-center">
+          {@error}
+        </p>
+      </div>
       <.button
         :if={@show_button?}
         type="submit"
--- a/lib/plausible_web/controllers/auth_controller.ex
+++ b/lib/plausible_web/controllers/auth_controller.ex
@@ -325,7 +325,7 @@ defmodule PlausibleWeb.AuthController do
       |> UserAuth.log_in_user(user, redirect_path)
     else
       {:error, :wrong_password} ->
-        Auth.log_failed_login_attempt("Incorrect password for #{email}")
+        Auth.log_failed_login_attempt("wrong password for #{email}")
 
         conn
         |> put_flash(:login_error, "Incorrect email or password. Please try again.")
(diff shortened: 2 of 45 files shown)


lib/plausible_web/templates/auth/activate.html.heex
```diff
@@ -1,107 +1,81 @@
...
-        if the problem persists
-      </:item>
-      <:item :if={ce?()}>
-        Ask on our
-        <.styled_link href="https://github.com/plausible/analytics/discussions" new_tab={true}>
-          community-supported forum
+    <div class="mt-6 flex flex-col gap-y-2 text-sm text-gray-500 dark:text-gray-400 text-center">
+      <p>
+        Didn't receive it?
+        <.styled_link href="/activate/request-code" method="post">Resend code</.styled_link>
```
apata, line 36: issue, blocking: I don't see feedback on clicking this. In the previous implementation, there was a flash. 

Success!
Activation code was sent to user1234@example.com

In this implementation, nothing happens visually.

lib/plausible_web/live/components/form.ex
```diff
@@ -252,8 +257,114 @@ defmodule PlausibleWeb.Live.Components.Form do
...
+  attr(:rest, :global, include: ~w(autofocus oninvalid))
+
+  def otp_input(assigns) do
+    oninput_js =
+      "this.value=this.value.replace(/[^0-9]/g, '');" <>
+        if assigns.autosubmit? do
+          " if (this.value.length >= #{assigns.length}) this.form.requestSubmit();"
+        else
+          ""
+        end
```
apata: nitpick, non-blocking: With the autosubmit feature in Activate account feature, it's quite easy to hit the rate limiter, which leads to a frustrating "Rate limited. Go to homepage (actually lands you back on Activate account feature)" experience. Can we treat autosubmitting activation codes as a separate extra feature from the bulk of the changes?

lib/plausible_web/templates/auth/password_reset_request_form.html.heex
```diff
@@ -1,31 +1,46 @@
...
-  </:title>
-
-  <:subtitle>
-    Enter your email so we can send a password reset link
-  </:subtitle>
-
-  <.form :let={f} for={@conn} action={Routes.auth_path(@conn, :password_reset_request)}>
-    <div class="my-4 mt-8">
-      <.input type="email" field={f[:email]} placeholder="user@example.com" />
+<div class="w-full max-w-md mx-auto mt-10 pb-16 px-4">
```
apata: question, non-blocking: can we find a way not to repeat the container class? some sort of generic wrapper?

lib/plausible_web/templates/auth/activate.html.heex
sanne-san: Good catch! They were rendered in the header which is hidden in the new auth layout. Fixed 👍

lib/plausible_web/templates/auth/password_reset_request_form.html.heex
sanne-san: Yes, good idea 👌

lib/plausible_web/live/components/form.ex
sanne-san: @apata I reverted to the original behaviour, and I also changed the flow to an inline message instead of a separate page for a less frustrating experience:

CleanShot 2026-07-06 at 17 09 27@2x

test/plausible_web/controllers/auth_controller/logs_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule PlausibleWeb.AuthController.LogsTest do
           post(conn, "/login", email: user.email, password: "wrong")
         end)
 
-      assert logs =~ "[warning] [login] wrong password for #{user.email}"
```
aerosol, line 29: Is this necessary?

test/plausible_web/controllers/auth_controller_test.exs
```diff
@@ -479,20 +486,22 @@ defmodule PlausibleWeb.AuthControllerTest do
...
+            body = html_response(conn, 200)
 
-            {conn.status == 429, conn}
+            {body =~ "Too many attempts", conn}
           end,
           500
         )
 
-      assert html_response(response, 429) =~ "Too many activation attempts"
+      assert html_response(response, 200) =~
```
aerosol, line 496: What's the reason for changing from 429 to 200?

lib/plausible_web/components/two_factor.ex
```diff
@@ -25,51 +25,28 @@ defmodule PlausibleWeb.Components.TwoFactor do
...
-        class={@input_class}
-        oninput={
-          if @show_button? do
-            "this.value=this.value.replace(/[^0-9]/g, ''); if (this.value.length >= 6) document.getElementById('#{@id}').focus()"
-          else
-            "this.value=this.value.replace(/[^0-9]/g, '');"
-          end
-        }
-        onclick="this.select();"
+    <div class={[@class, "flex flex-col gap-y-6 items-center"]}>
```
aerosol, line 35: This produces broken input length in settings

Image

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -116,19 +129,23 @@ defmodule PlausibleWeb.AuthController do
         end
 
       {:error, :incorrect} ->
-        render(conn, "activate.html",
-          error: "Incorrect activation code",
+        render_auth_page(conn, "activate.html",
+          error: "That code didn't work. Please try again.",
```
aerosol, line 133: This is a bit inconsistent - we're rendering wrong activation code error inline, under the input, but now wrong 2FA code at login is indicated by the flash. Same with wrong recovery code.

lib/plausible_web/templates/auth/activate.html.heex
```diff
@@ -1,107 +1,81 @@
...
-        >
-          Delete this account
-        </.styled_link>
-        and start over
-      </:item>
-    </.focus_list>
-  </:footer>
-</.focus_box>
+      <% else %>
+        <.styled_link method="delete" href="/me?redirect=/register">change email</.styled_link>
```
aerosol, line 77: are we deliberately dropping the confirmation dialog?

lib/plausible_web/controllers/auth_controller.ex
```diff
@@ -277,19 +325,19 @@ defmodule PlausibleWeb.AuthController do
       |> UserAuth.log_in_user(user, redirect_path)
     else
       {:error, :wrong_password} ->
-        Auth.log_failed_login_attempt("wrong password for #{email}")
+        Auth.log_failed_login_attempt("Incorrect password for #{email}")
```
RobertJoonas: This message is not exposed to the user so I would revert this change. There might be a Grafana graph somewhere that depends on the word "wrong".

lib/plausible_web/templates/auth/login_form.html.heex
```diff
@@ -1,75 +1,77 @@
...
-            Routes.sso_path(@conn, :login_form,
-              return_to: @conn.params["return_to"],
-              prefer: "manual"
-            )
-          }>
-            Sign in here
-          </.styled_link>
-          instead.
+        <%= if ee?() do %>
+          <%= on_ee do %>
```
RobertJoonas: I think the outer if block can be removed here.

lib/plausible_web/router.ex
```diff
@@ -411,23 +411,32 @@ defmodule PlausibleWeb.Router do
...
     pipe_through [:browser, :csrf]
 
-    scope alias: Live, assigns: %{connect_live_socket: true} do
+    scope alias: Live,
+          assigns: %{
+            connect_live_socket: true,
+            hide_header?: true,
+            hide_footer?: true,
+            disable_global_notices?: true
+          } do
```
RobertJoonas: Since the new AuthLayoutContext is responsible for putting the `:auth` layout, what do you think if we move these layout-related assigns there too?

```elixir
def on_mount(_arg, _params, _session, socket) do
  socket =
    assign(socket,
      hide_header?: true,
      hide_footer?: true,
      disable_global_notices?: true
    )

  {:cont, socket, layout: {PlausibleWeb.LayoutView, :auth}}
end
```

test/plausible_web/controllers/auth_controller/logs_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule PlausibleWeb.AuthController.LogsTest do
...
+      assert logs =~ "[warning] [login] Incorrect password for #{user.email}"
```
RobertJoonas: Ditto: this log is for internal use. Best to revert.

RobertJoonas (commented):
Nice work! 👏 Happy to see much less classname duplication and more extracted reusable components :)

lib/plausible_web/templates/auth/activate.html.heex
sanne-san: Yeah deliberate, at this point in the flow it's not really a destructive action. Clicking 'change email' and then being asked 'Deleting your account cannot be reversed. Are you sure?' sounds quite alarming for what's essentially 'let me fix a typo/use a different email'.

test/plausible_web/controllers/auth_controller_test.exs
sanne-san: @aerosol we switched the rate-limit from a full-page 429 error to an inline error on the `/activate` page itself for better UX (see [this convo](https://github.com/plausible/analytics/pull/6479#discussion_r3529313323)). Since inline errors are rendered as a normal form render rather than an error page, I thought the status is 200 instead of 429. Would you prefer if we keep 429 instead?

RobertJoonas (approved):

PR #6480, opened 2026-07-02
Dashboard loading state UX improvements

RobertJoonas:
### Changes

* Enable the fade-in animation when switching between `IndexBreakdown` tabs (e.g. Sources -> Channels) -- done by passing a `key={selectedTabKey}` to the IndexBreakdown components
  * Note: the animation does not trigger when changing dashboard state to something that has already been visited (i.e. the results are cached)
* `WorldMap` component improvements
  * initital state (data is fetching) -- render loading spinner instead of blank world map without colors
  * wrap with LazyLoader like any other IndexBreakdown (fetches data only after scrolled into view)
  * add fade-in animation when transitioning from loading -> data (or showing cached results after tab change) -- basically do the same as IndexBreakdowns
  * code organization: separate SVG rendering into a new component (world-map-svg.tsx) that assumes the presence of data

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/stats/locations/map.tsx
+++ b/assets/js/dashboard/stats/locations/map.tsx
@@ -8,7 +8,6 @@ import {
 import { useAppNavigate } from '../../navigation/use-app-navigate'
 import { useSiteContext } from '../../site-context'
 import { useDashboardStateContext } from '../../dashboard-state-context'
-import { UIMode, useTheme } from '../../theme-context'
 import { MIN_HEIGHT } from '../reports/index-breakdown'
 import { GeolocationNotice } from './geolocation-notice'
 import { DashboardState } from '../../dashboard-state'
@@ -20,13 +19,7 @@ import {
   BreakdownReportKey
 } from '../reports/reports-config'
 import LazyLoader from '../../components/lazy-loader'
-import {
-  CountryData,
-  MetricLabel,
-  WorldMapSvg,
-  MAP_CONTAINER_WIDTH,
-  MAP_CONTAINER_HEIGHT
-} from './world-map-svg'
+import { CountryData, MetricLabel, WorldMapSvg } from './world-map-svg'
 
 function getMetricLabel(dashboardState: DashboardState): MetricLabel {
   if (hasConversionGoalFilter(dashboardState)) {
@@ -46,7 +39,6 @@ const WorldMap = ({
   onDataReady: (response: QueryApiResponse) => void
 }) => {
   const navigate = useAppNavigate()
-  const { mode } = useTheme()
   const site = useSiteContext()
   const { dashboardState } = useDashboardStateContext()
   const [visible, setVisible] = useState(false)
@@ -75,7 +67,10 @@ const WorldMap = ({
     ],
     { enabled: visible }
   )
-  const { data } = apiState
+  const { data, isPending, isPlaceholderData, isError } = apiState
+
+  const isLoading =
+    isPending || isError || (isPlaceholderData && !isRealtimeSilentUpdate)
 
   useEffect(() => {
     if (data) {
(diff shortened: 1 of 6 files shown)


assets/js/dashboard/stats/locations/world-map-svg.tsx
```diff
@@ -0,0 +1,229 @@
...
+export type CountryData = {
+  alpha_3: string
+  name: string
+  visitors: number
+  code: string
+}
+
+export type MetricLabel = { singular: string; plural: string }
+
+export function WorldMapSvg({
```
apata, line 25: question: Why must the world map flash in and out whenever the user is doing something on the dashboard? Is it to indicate we're actually working on their new selection? Can we find a cheaper, less disruptive way to indicate that? By cheaper, I mean less work for the browser. By less disruptive, I mean something that's changing less things on the dashboard at once. The map itself flashing in and out is, in my view, a step backward.

As I see it, the world map with country outlines is constant (_at least in dashboard timescales_). The previous implementation took pains to respect that and to just fade paint according to the current data. 

With dummy.site where the autogenerated visitors are spread between very few locations, this fade wasn't very noticeable, but on real sites, it seems the variety of visitor locations guarantees a noticeable change in the map when filters are changed.

RobertJoonas: > Why must the world map flash in and out whenever the user is doing something on the dashboard? Is it to indicate we're actually working on their new selection?

Yeah, and also just trying to make the UX consistent. It feels strange when you apply a filter and the whole dashboard reacts to that change with loading spinners all around (including the graph container), but the map stays intact. Sometimes, there's not even any notable change in the colors and the report just looks frozen.

> Can we find a cheaper, less disruptive way to indicate that? By cheaper, I mean less work for the browser. By less disruptive, I mean something that's changing less things on the dashboard at once. The map itself flashing in and out is, in my view, a step backward.

Yeah, I see how it feels a bit like stepping backwards. But perhaps for the sake of consistency, it's acceptable for the time being?

Letting the SVGs (both map and the main graph) stay mounted throughout the loading state is definitely an idea worth pursuing -- I can see the potential to make the loading state look really nice, but only if we get rid of the loading spinners altogether. For the rest of the dashboard (i.e. IndexBreakdowns) we should probably run some animations from old (placeholder) data to new, fresh data. There's also the question of the initial load where we'd most likely have to introduce some silhouettes. I believe @sanne-san had something sketched out already.

Doing all that sounds quite tricky and I don't think this will be worked on any time soon. Therefore proposing this small step backwards in pursuit of consistency.

Re: browser performance, I believe it used to work like this in the past for a long time without any issues. Do you see it as a blocker @apata?

apata: Disruptive is subjective and I couldn't prove the performance difference, so I shouldn't say it's a blocker. It's something that fully caught my attention though.  

(_I tried to benchmark this env vs staging with browser performance APIs but failed: the envs didn't have as much as parity as I expected. There were differences in first paint performance, which this PR shouldn't much affect_). 

Even though I don't have concrete data to show, I just want us to keep in mind that any time we do relayout or fade or anything fancier (like scaling up / down animation), it's the user's device doing work. We want the work to be justified and proportional to the benefits we hope to achieve. 

Also, the more we do, the more we are exposed to glitching from bugs in browsers. Maybe a minor matter, but it's an argument in favor of simplicity.

> Re: browser performance, I believe it used to work like this in the past for a long time without any issues.

It seems "worked in the past without any issues" argument can be used to justify *not* changing the map loading logic as well. To me, it's a matter of choosing what we (or what we believe our users will) value more, slightly more consistent from report box to box, or slightly less visually disruptive.

On the subject of consistency, there's the time axis to consider as well. The map has worked without flashing in and out for years. Changing it (_in any way_) will cause an inconsistency with on the time axis. I suspect quite a few users don't actually like things changing.

assets/js/dashboard/stats/locations/map.tsx
```diff
@@ -115,221 +105,94 @@ const WorldMap = ({
...
+}: {
+  isPending: boolean
+  isPlaceholderData: boolean
+  isError: boolean
+  isRealtimeSilentUpdate: boolean
+  maxValue: number
+  dataByAlpha3Code: Map<string, CountryData>
+  metricLabel: MetricLabel
+  mode: UIMode
+  onCountryClick: (country: CountryData) => void
```
apata: nitpick, non-blocking: These 5 props pass through and the others are from API state to calculate whether to show loading screen. Can it be avoided by removing this intermediate component? 
Also, mode can be accessed from context directly in WorldMapSVG, it's not used elsewhere.

assets/js/dashboard/stats/locations/map.tsx
```diff
@@ -115,221 +105,94 @@ const WorldMap = ({
...
+          <div />
+        </div>
+      ) : (
+        <div
+          className="relative flex justify-center items-center mt-4 w-full"
+          style={{
+            height: MAP_CONTAINER_HEIGHT,
+            maxWidth: MAP_CONTAINER_WIDTH
+          }}
+        >
```
apata: suggestion, non-blocking: Because this wrapper is integral to rendering the map and tooltip correctly, I'd wrap it in WorldMapSvg

apata (approved):

assets/js/dashboard/stats/locations/map.tsx
RobertJoonas: https://github.com/plausible/analytics/pull/6480/commits/af4522660f1f072386ed2ef4d69693c89e09ba50

assets/js/dashboard/stats/locations/map.tsx
RobertJoonas: Much cleaner, thanks! https://github.com/plausible/analytics/pull/6480/commits/af4522660f1f072386ed2ef4d69693c89e09ba50

PR #6481, opened 2026-07-02
Annotations: FE

apata:
### Changes

Adds the rest of Annotations feature. Depends on the BE PR.

### Tests

### Changelog

### Documentation
- [ ] https://github.com/plausible/docs/pull/733 has been updated and merged

### Dark mode

### TODO after merge
- [ ] Merge website pricing plans update https://github.com/plausible/website/pull/1108


assets/js/dashboard/annotations/annotation-list-items.tsx
```diff
@@ -0,0 +1,60 @@
...
+)
+
+export const AnnotationItemRow = ({ children }: { children: ReactNode }) => (
+  <div className="group flex flex-row gap-x-2">
+    <VerticalBar />
+    {children}
+  </div>
+)
+
+export const AnnotationAttributionLine = ({
```
RobertJoonas: Minor nit: the word "attribution" threw me off a bit when first reading it. How about "signature" instead?

assets/js/dashboard/annotations/annotations-modals.tsx
```diff
@@ -0,0 +1,379 @@
...
+    const time = formatTime(date, {
+      use12HourClock: is12HourClock(),
+      includeMinutes: true
+    })
+    return `${formatDay(date)} at ${time}`
+  }
+  return formatDay(date)
+}
+
+const NOTE_RECOMMENDED_MAX_LENGTH = 250
```
RobertJoonas: Why the word "recommended"? Same question about the UI label

```suggestion
const NOTE_MAX_LENGTH = 250
```

assets/js/dashboard/annotations/annotations-modals.tsx
```diff
@@ -0,0 +1,379 @@
...
+}: AnnotationModalProps &
+  ApiRequestProps & {
+    initialDatetime: AnnotationPayload['datetime']
+    initialGranularity: AnnotationPayload['granularity']
+    initialType: AnnotationPayload['type']
+  } & {
+    onSave: (input: AnnotationPayload) => void
+  }) => {
+  const defaultNote = ''
+  const [note, setNote] = useState(defaultNote)
```
RobertJoonas: This makes it look like the default value is needed elsewhere. Let's inline?

```suggestion
  const [note, setNote] = useState('')
```

assets/js/dashboard/annotations/annotations-modals.tsx
```diff
@@ -0,0 +1,379 @@
...
+        </Button>
+        <SaveButton
+          disabled={
+            status === 'pending' || overLimit || disabledMessage !== null
+          }
+          onSave={() => {
+            const trimmedNote = note.trim()
+            const saveableNote = trimmedNote.length
+              ? trimmedNote
+              : notePlaceholder
```
RobertJoonas: Shouldn't we prohibit creating an empty note rather than saving the note with `notePlaceholder`? When you open the "Add note" modal and hit save directly, it creates this note:

Image

assets/js/dashboard/annotations/annotations-modals.tsx
```diff
@@ -0,0 +1,379 @@
...
+}
+
+interface AnnotationModalProps {
+  user: UserContextValue
+  siteAnnotationsAvailable: boolean
+  onClose: () => void
+  notePlaceholder: string
+}
+
+export const CreateAnnotationModal = ({
```
RobertJoonas, line 64: I think `notePlaceHolder` and `initialType` shouldn't be props, but rather fixed values defined as module level `const`s. They do make sense as props in the "Edit note" modal, but not in the "Add note" one.

assets/js/dashboard/annotations/annotations-modals.tsx
```diff
@@ -0,0 +1,379 @@
...
+  siteAnnotationsAvailable,
+  user,
+  notePlaceholder,
+  status,
+  error,
+  reset
+}: AnnotationModalProps &
+  ApiRequestProps & {
+    onSave: (input: Pick<Annotation, 'id' | 'note' | 'type'>) => void
+    onDelete?: (annotation: Annotation) => void
```
RobertJoonas: Looks like this function is always provided, so we could enforce this prop and remove the `typeof onDelete === 'function'` check below?

assets/js/dashboard/annotations/annotations.test.ts
```diff
@@ -0,0 +1,330 @@
...
+          type: AnnotationType.personal,
+          user: loggedInUser(role),
+          siteAnnotationsAvailable: false
+        })
+      ).toBe(true)
+    }
+  )
+
+  it.each([[AnnotationType.personal], [AnnotationType.site]])(
+    'forbids the public role from adding %s annotations',
```
RobertJoonas, line 124: Should we test the same for super admins?

assets/js/dashboard/annotations/annotations.ts
```diff
@@ -0,0 +1,235 @@
...
+const ROLES_WITH_MAYBE_SITE_ANNOTATIONS = [Role.admin, Role.editor, Role.owner]
+const ROLES_WITH_PERSONAL_ANNOTATIONS = [
+  Role.billing,
+  Role.viewer,
+  Role.admin,
+  Role.editor,
+  Role.owner
+]
+
+export function canEditAnnotation({
```
RobertJoonas, line 90: The whole function body is very similar to canAddAnnotation. Maybe extract a private function like `hasPermissions({type, user, siteAnnotationsAvailable})` to not repeat this logic?

assets/js/dashboard/stats/graph/main-graph.tsx
```diff
@@ -71,31 +84,72 @@ const initialTooltipState: TooltipState = {
...
+    const onClickOutside = (event: MouseEvent) => {
+      if (!tooltipRef.current?.contains(event.target as Node)) {
+        setTooltip(initialTooltipState)
+      }
+    }
+    if (tooltip.persistent && isTouchDevice === false) {
+      document.addEventListener('click', onClickOutside)
+    } else {
+      document.removeEventListener('click', onClickOutside)
+    }
```
RobertJoonas, line 138: The else branch should be unnecessary because the return function always cleans the eventListener up.

assets/js/dashboard/stats/graph/main-graph.tsx
```diff
@@ -398,41 +499,147 @@ export const MainGraph = ({
...
+          isTouchDevice={isTouchDevice}
+          closeTooltip={closeTooltip}
+        />
+      )}
+      {hasActions && (
+        <div className="flex flex-row gap-x-2 mt-2">
+          {!!annotationDatetime && canAddAnnotation && (
+            <Button
+              size="xs"
+              className="flex-1 bg-gray-600/70 border-gray-600/70 hover:bg-gray-600 hover:border-gray-600"
```
RobertJoonas: The classname between the two buttons should be exactly the same so perhaps let's extract that into a variable?

RobertJoonas (approved):

assets/js/dashboard/stats/graph/main-graph.tsx
apata: Thanks for pointing this out! You're right, I'm not taking into account how effects work:

https://react.dev/reference/react/useEffect#parameters
> setup: The function with your Effect’s logic. Your setup function may also optionally return a cleanup function. When your [component commits](https://react.dev/learn/render-and-commit#step-3-react-commits-changes-to-the-dom), React will run your setup function. After every commit with changed dependencies, React will first run the cleanup function (if you provided it) with the old values, and then run your setup function with the new values. After your component is removed from the DOM, React will run your cleanup function.

assets/js/dashboard/annotations/annotations.test.ts
apata: Valid question! Super admin is a bit of problematic role right now. To be honest, I'm not sure that value should reach the dashboard at all. Let's look at it as a separate issue?

assets/js/dashboard/annotations/annotations-modals.tsx
apata: Agreed, fixed! It's a leftover from a time when note placeholder was dynamic, something like "Note on May 15th (<the actual date of note to be>)" rather than generic "E.g. ...". 

That dynamic placeholder allowed quickly injecting markers without needing to think of a name, which was nice.

assets/js/dashboard/annotations/annotations-modals.tsx
apata: Thanks, fixed, explanation here! https://github.com/plausible/analytics/pull/6481#discussion_r3569367463

assets/js/dashboard/annotations/annotations-modals.tsx
apata: Valid question! I think it was to soften the tone. Removing

assets/js/dashboard/annotations/annotation-list-items.tsx
apata: That's a good suggestion, but for now I went with `AnnotationAuthorship` to be on par with Segments.

assets/js/dashboard/stats/graph/main-graph.tsx
apata: Hmm, true that it is duplicative... AFAIK TW recommends against variables though. They suggest to generate repetitive items with map, or extract to component, or to use multicursor when editing. I can do the map thing, but it creates this situation:
```tsx
      {hasActions && (
        <div className="flex flex-row gap-x-2 mt-2">
          {[
            !!annotationDatetime &&
              canAddAnnotation && {
                text: 'Add note',
                onClick: () => {
                  closeTooltip()
                  setModal({
                    type: 'create-annotation',
                    annotation: {
                      type: AnnotationType.personal,
                      datetime: annotationDatetime,
                      granularity: getAnnotationGranularity(interval)
                    }
                  })
                }
              },
            !!zoomDate && {
              text: `View ${interval}`,
              onClick: () => onZoomToPeriod(zoomDate)
            }
          ]
            .filter((b) => !!b)
            .map(({ text, onClick }) => (
              <Button
                key={text}
                size="xs"
                className="flex-1 bg-gray-600/70 border-gray-600/70 hover:bg-gray-600 hover:border-gray-600"
                onClick={onClick}
              >
                {text}
              </Button>
            ))}
        </div>
      )}

```

RobertJoonas: Makes sense. Maybe extracting a simple button component would be the cleanest approach then?

```js
function TooltipActionButton(props) {
  const className = "flex-1 bg-gray-600/70 border-gray-600/70 hover:bg-gray-600 hover:border-gray-600"
  return <Button className={className} {...props}>
}
```

Fine for me to leave the duplicate classes though.

PR #6485, opened 2026-07-05
Add PR checking via deepsec

cnkk:


.deepsec/data/analytics/tech.json
```diff
@@ -0,0 +1,14 @@
...
+    "elixir",
+    "github-actions",
+    "phoenix"
+  ],
+  "sentinels": [
+    "Dockerfile",
+    "mix.exs"
+  ],
+  "detectedAt": "2026-06-09T18:53:23.061Z",
+  "rootPath": "/Users/cenk/plausible/analytics"
```
zoldar: Uhm, maybe that should be redacted or corrected?

cnkk: added to `.gitignore` and removed:)

.deepsec/matchers/auth-flow-entrypoint.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export const authFlowEntrypoint: MatcherPlugin = {
+  slug: "plausible-auth-flow-entrypoint",
+  description:
+    "Login, registration, 2FA, password reset, session, invitation, and membership service boundaries",
+  noiseTier: "normal",
+  filePatterns: [
+    "lib/plausible_web/controllers/auth_controller.ex",
+    "lib/plausible_web/controllers/invitation_controller.ex",
+    "lib/plausible_web/live/register_form.ex",
```
apata, line 22: issue, non-blocking: All the instruction files are too much to validate for correctness in a reasonable amount of time. I'm sure they're better than what I could produce manually, but I think ideally what we commit to the repo (_and that we thus commit to maintaining_) should be able to be checked for correctness.

Long term I hope we find a way to reduce the duplication between the instructions and the actual app structure.

apata (approved):

PR #6489, opened 2026-07-07
Fix dropdown clipping in filter and segment modals

sanne-san:
### Changes

- Add `grow` mode to `Modal` that lets the panel grow with content and defers scroll to the outer overlay
- `ModalLayout` opts in, so filter and segment modals stop clipping combobox popovers
- Breakdown modals keep the bounded behavior as inner content scrolls
- Remove dead `viewport`/`handleResize`/`DEFAULT_WIDTH`/`getStyle`/`allowScroll` plumbing

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/components/modal-layout.tsx
+++ b/assets/js/dashboard/components/modal-layout.tsx
@@ -7,21 +7,15 @@ export function ModalLayout({
   title,
   onClose,
   children,
-  maxWidth = '460px',
-  allowScroll = true
+  maxWidth = '460px'
 }: {
   title: ReactNode
   onClose: () => void
   children: ReactNode
   maxWidth?: string
-  allowScroll?: boolean
 }) {
   return (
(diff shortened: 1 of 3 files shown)


aerosol (approved):

apata (changes requested):
Unfortunately, this introduces a new clipping issue in breakdown modals when using a phone in landscape mode

Image

sanne-san:
> Unfortunately, this introduces a new clipping issue in breakdown modals when using a phone in landscape mode

@apata good catch, I missed this because it only happens when the screen width is > 768px. It was caused by the min-height on the modal's inner div, fixed [here](https://github.com/plausible/analytics/pull/6489/changes/4c968e7d8cb06400192135740e4cf3776b566278).

apata (approved):

PR #6490, opened 2026-07-07
Fix period arrow keys hijacking custom-range calendar

sanne-san:
### Changes
- On periods with `< >` controls (Today, Month to Date, Last Month, Year to Date, specific day/month/year), the document-level ArrowLeft/ArrowRight listener kept firing while the calendar popover was open, silently shifting the period and closing the calendar.
- Suppress that listener whenever a flatpickr calendar is mounted, via a new `isDateRangeCalendarOpen` predicate on `Keybind.shouldIgnoreWhen`.
- Added a regression test covering both directions: arrows still shift the period when the calendar is closed, and no longer do so while it's open.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/nav-menu/query-periods/dashboard-dates.test.tsx
+++ b/assets/js/dashboard/nav-menu/query-periods/dashboard-dates.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen, waitFor } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { TestContextProviders } from '../../../../test-utils/app-context-providers'
 import { stringifySearch } from '../../util/url-search-params'
@@ -171,9 +171,7 @@ test('arrow keys shift the period when the custom-range calendar is closed, but 
   expect(screen.getByText('July 2024')).toBeVisible()
 
   await userEvent.keyboard('c')
-  await waitFor(() =>
-    expect(document.querySelector('.flatpickr-calendar')).not.toBeNull()
-  )
+  expect(document.querySelector('.flatpickr-calendar')).not.toBeNull()
 
   await userEvent.keyboard('{ArrowLeft}')
   await userEvent.keyboard('{ArrowLeft}')
--- a/assets/js/dashboard/nav-menu/query-periods/dashboard-dates.test.tsx
+++ b/assets/js/dashboard/nav-menu/query-periods/dashboard-dates.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen } from '@testing-library/react'
+import { render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { TestContextProviders } from '../../../../test-utils/app-context-providers'
 import { stringifySearch } from '../../util/url-search-params'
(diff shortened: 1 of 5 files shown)


assets/js/dashboard/nav-menu/query-periods/dashboard-dates.test.tsx
```diff
@@ -150,6 +150,38 @@ test.each([
...
+  })
+
+  expect(screen.getByText('August 2024')).toBeVisible()
+  expect(document.querySelector('.flatpickr-calendar')).toBeNull()
+
+  await userEvent.keyboard('{ArrowLeft}')
+  expect(screen.getByText('July 2024')).toBeVisible()
+
+  await userEvent.keyboard('c')
+  await waitFor(() =>
```
RobertJoonas: This should work fine without the `waitFor` wrapper. Everything that happens on keypress is synchronous.

assets/js/dashboard/nav-menu/query-periods/move-period-arrows.tsx
```diff
@@ -31,11 +36,22 @@ const ArrowKeybind = ({
     [site, dashboardState, keyboardKey]
   )
 
+  const handler = useCallback(() => {
+    navigate({ search })
+  }, [navigate, search])
+
   return (
-    <NavigateKeybind
+    <Keybind
```
RobertJoonas: Adding ignore conditions to `NavigateKeybind` sounds like something we might want to reuse later on, so we could extend the component with a new prop like `shouldAlsoIgnoreWhen`:

```js
export function NavigateKeybind({
  keyboardKey,
  type,
  navigateProps,
  shouldAlsoIgnoreWhen = []
}: {
  keyboardKey: string
  type: KeyboardEventType
  navigateProps: AppNavigationTarget
  shouldAlsoIgnoreWhen?: Array<(event: KeyboardEvent) => boolean>
}) {
  const navigate = useAppNavigate()
  const handler = useCallback(() => {
    navigate({ ...navigateProps })
  }, [navigateProps, navigate])

  return (
    <Keybind
      keyboardKey={keyboardKey}
      type={type}
      handler={handler}
      shouldIgnoreWhen={[isModifierPressed, isTyping, ...shouldAlsoIgnoreWhen]}
      targetRef="document"
    />
  )
}
```

So here, we could still keep calling `NavigateKeybind`, just passing `shouldAlsoIgnoreWhen = [isDateRangeCalendarOpen]` as an extra prop.

RobertJoonas (approved):
Tested on staging and it does what intended, so approving 👍

I feel like there's another side to actually making it work though -- the calendar doesn't automatically gain focus. For example, if I hit the down arrow right after opening the calendar, it'll start scrolling down the page (browser default). So even though left/right arrow keys are not affecting the rest of the page (which is correct of course), the keypresses are not registered by the calendar itself either. This is the behaviour I've experienced on Chrome and Safari at least.

sanne-san:
> Tested on staging and it does what intended, so approving 👍
> 
> I feel like there's another side to actually making it work though -- the calendar doesn't automatically gain focus. For example, if I hit the down arrow right after opening the calendar, it'll start scrolling down the page (browser default). So even though left/right arrow keys are not affecting the rest of the page (which is correct of course), the keypresses are not registered by the calendar itself either. This is the behaviour I've experienced on Chrome and Safari at least.

@RobertJoonas agree – I actually tried to fix this too but didn't succeed. Since the bug report was about this specifically, I thought it was worth shipping this as-is so that someone else can look into that separately. I will implement your feedback so this can go out and add a card for that as a separate issue!

assets/js/dashboard/nav-menu/query-periods/move-period-arrows.tsx
sanne-san: Yeah I considered this but wasn't sure which approach was better. Will go with this, thanks!

PR #6491, opened 2026-07-08
Implement support for replaying events

zoldar:
### Changes

This PR implements support for replaying events via existing ingestion pipeline. Although the current implementation is entirely contained in `analytics` repository, the eventual solution will be split among `analytics` and `persistor`.

The plan for release is following:
- ClickHouse database migration - merge, release and run; the only thing it does is renaming already existing columns in our cluster, which were added when we were recovering lost events from persistor some time ago
- Persistor logic changes - extend the ClickHouse schemas for event and session with `replay_session_id` field and use a different cache key for replayed events as well in `CacheStore`; the change will be merged and released as well
- Finally, analytics pipeline changes - a preview app will be created first to test recovery using an already setup staging instance of ingestvcr service; merging the PR will be held off until headers sanitisation will be implemented in our infra to prevent sending replay headers from outside our system

#### Recording

A simplified diagram of how recording is integrated into the event ingestion pipeline:

image

When the event payload reach Ingress proxy, the haproxy dispatches its body and headers to the agent process listening on a tcp socket. The agent deserialises the payload, assigns a random ID to the event and asynchronously persists the event in the store while immediately returning the newly assigned ID to the proxy. The proxy appends it as `X-Replay-Event-ID` header. As of now, it's not used and is ignored down the pipeline.

The analytics ingest pipeline processes the event normally without any additional processing and so does persistor, after which the event ends up batched and inserted into ClickHouse DB.

#### Replaying

image

When replaying, each payload will go through the same ingress proxy but it won't be sent to the recording agent on the basis of replay header being already present. The replay utility appends 3 following headers when sending the payloads:

- `X-Replay-Event-ID` (ignored for now, like in the case of recording)
- `X-Replay-Session-ID`
- `X-Replay-Time`

Once the payload reaches ingestion pipeline, it's decoded in the request building phase. There, for replayed event, the event timestamp is replaced with the one provided in `X-Replay-Time` and `X-Replay-Session-ID` is put under `request.replay_session_id`. The replay session ID is then carried over to event attributes.

The `put_salts` step is crucial in ensuring that replayed events are processed in complete isolation from other events even if some user sessions are still valid from the time of recording. Without that, the existing sessions will get corrupted. Processing replayed events in isolation is a trade-off we have to make - there's no feasible way to retroactively adjust existing sessions, especially somewhere in the middle.

The modified `generate_user_id` routine used in `put_user_id` and `register_session` steps is crucial in ensuring that replayed events are processed in complete isolation from other events even if some user sessions are still valid from the time of recording. Without that, the existing sessions will get corrupted. Processing replayed events in isolation is a trade-off we have to make - there's no feasible way to retroactively adjust existing sessions, especially somewhere in the middle. The isolation is achieved by appending `replay_session_id` to the hashed value if it's set.

The last step is session management in persistor. Replayed sessions are stored and fetched using distinct shape of cache key, including replay session ID. Although risk of user ID collisions is low, it's still better to be on the safe side - that's one more level of ensuring isolation.

Lastly, when a new replayed session is provisioned, its replay session ID attribute is populated accordingly, like in the case of event.

Recording replay session ID allows relatively easy "rollback" of a botched replay and starting over again. Also, two different replays (with different session IDs), even if overlapping, are easy to tell apart and clean up properly as well. An interrupted session can also be safely resumed (the replay utility allows to provide an existing session ID; otherwise it generates a unique one on every run).

### Tests


extra/lib/plausible/session/computed_salts.ex
```diff
@@ -0,0 +1,66 @@
+defmodule Plausible.Session.ComputedSalts do
```
aerosol: Minor but the "Computed" bit doesn't sound too meaningful, they're all "computed" in a way. How about renaming it to ReplaySalts?

lib/plausible/ingestion/request.ex
```diff
@@ -131,6 +135,53 @@ defmodule Plausible.Ingestion.Request do
...
+        else
+          changeset
+        end
+      else
+        changeset
+      end
+    end
+
+    defp to_integer(s) when is_binary(s) do
+      case Integer.parse(s) do
```
aerosol: Should be fine to do String.to_integer - that input is naively trusted by contract and it's OK to crash if violated

aerosol (approved):
Looks good :shipit:

lib/plausible/ingestion/request.ex
zoldar: https://github.com/plausible/analytics/pull/6491/commits/de96d3560b1d3c5f0ce369838cd0995382d9b6dc

extra/lib/plausible/session/computed_salts.ex
zoldar: https://github.com/plausible/analytics/pull/6491/commits/c39d69831ed58982238630c777629392b3944d53

ukutaht:
Thanks for the writeup @zoldar! Based on the description do I understand correctly that `X-Replay-Session-Id` is constant for all events within the same replay run and generated randomly per replay run?

zoldar:
> Thanks for the writeup @zoldar! Based on the description do I understand correctly that X-Replay-Session-Id is constant for all events within the same replay run and generated randomly per replay run?

Correct. That is, unless the session ID is explicitly provided when running replay. This enables resuming interrupted session.

ukutaht:
Thanks! I don't fully understand why a separate salts mechanism is needed here. From what I can tell its purpose is to ensure that user_ids from live ingestion vs different replay runs do not clash. But can that not be achieved by including the replay_session_id as an input to the user_id hash function?

zoldar:
@ukutaht That's what I basically did initially when spiking it. However, after some more research, I've concluded this is not going to cut it. In [SipHash](https://en.wikipedia.org/wiki/SipHash), providing a secret key with good entropy is crucial. `:crypto.strong_rand_bytes` meets those requirements. `replay_session_id` is neither secret (it's persisted verbatim together with event and session) nor it has a good entropy (`rand.UInt64` is used for generating it and the [docs](https://pkg.go.dev/math/rand#pkg-overview) explicitly state it's pseudo random and not suitable for security-sensitive applications).

Using `Plug.Crypto.KeyGenerator` adds certain degree of complexity, but I think that in this particular case it's warranted, as ensuring the stats are properly anonymised is one of our core features.

ukutaht:
So I was thinking of something like:
1. Live ingestion is generated with `siphash(salt, user_agent <> ip <> domain)`
2. Replay ingestion is generated with `siphash(salt, user_agent <> ip <> domain <> replay_session_id)` where the salt is the same salt as normal

Is this what you had spiked as well? I understand that a randomly generated session_id is not suitable for cryptographic purposes but that's the job of the salting mechanism we already have. The salt is generated with `:crypto.strong_rand_bytes` and discarded which provides the anonymity. The string in the second argument is purely provided to manage uniqueness. It's fine to include low-entropy inputs in the second string argument since it is not the part that is responsible for rendering keys irreversible.

I do see the need if you had spiked a version where the `replay_session_id` takes place of the salt for replayed events:

`siphash(replay_session_id, user_agent <> ip <> domain)`

In this version I agree the session_id is not strong enough. But then I would ask why not move the session_id to the second argument and keep the usual salt in the first.

zoldar:
@ukutaht 🤦 Of course at first I've spiked the dumb version and somehow completely missed the possibility of altering hashed value in `generate_user_id` instead. Yeah, that will make it _a bit_ simpler.

zoldar:
Ah, okay, I think that has crossed my mind actually but I've discarded the idea because salts could be cycled in the middle of replay. That's a non-issue though. I don't know, either way. That. Was. Dumb. argh

zoldar:
Simplified with https://github.com/plausible/analytics/pull/6491/commits/a0579e5bbd0ffeb9b3d18eb99effbaa902509d0f and https://github.com/plausible/analytics/pull/6491/commits/5ef89186bdc8c1e9e11fbc9ececcbf9cee6835cd. Thanks.

ukutaht (approved):

ukutaht:
@zoldar nice, thanks 👍 

> I've discarded the idea because salts could be cycled in the middle of replay. That's a non-issue though

Hmm yeah the ingestion pipeline is designed in theory to carry sessions over during salt rotation but I'm not sure how well it works in practice tbh.

PR #6498, opened 2026-07-09
CRM: Trial prospects worker

ukutaht:
### Changes

Context: Marko has wanted the CRM to be able to surface trial prospects better so we can support teams with high potential better during the trial.

This PR adds a postgres table and a background worker to score trial prospects by potential MRR based on actual usage. UI will be added in a follow-up PR. The UI will support ranking trialing teams by MRR, traffic, and signup date to keep an eye on active trials along with their potential if they choose to upgrade.

The background worker scores the whole trial population (= teams with no subscription and up to 30 days after trial expiry) every night. On each run previous scores are overwritten and stale entries removed from the table.

Migration will be extracted once PR is approved.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/clickhouse.ex
```diff
@@ -76,6 +80,76 @@ defmodule Plausible.Stats.Clickhouse do
...
+              ^last_complete_day,
+              ^@window_days
+            )
+        }
+      )
+
+    ClickhouseRepo.one(
+      from(t in subquery(aggregates),
+        select: %{
+          first_data_day: t.first_data_day,
```
apata, line 138: According to my manual testing, this comes back as `1970-01-01 00:00:00` if there's no events for any of the sites.

extra/lib/plausible/workers/score_trial_prospects.ex
```diff
@@ -0,0 +1,91 @@
...
+      %{events_in_window: 0} -> []
+      traffic -> [build_row(team, traffic, site_ids)]
+    end
+  end
+
+  defp build_row(team, traffic, site_ids) do
+    score =
+      TrialProspects.score(
+        traffic.estimated_monthly,
+        Teams.Billing.features_usage(team, site_ids),
```
apata, line 69: This one line starts 10 postgres queries, so I'm a bit worried about postgres stress at the time of this worker: the more teams are created, the more stress there'll be.

lib/plausible/stats/clickhouse.ex
apata: Looking deeper, I see lines where events_in_window=0 get ignored down the line, so it won't matter.

extra/lib/plausible/workers/score_trial_prospects.ex
```diff
@@ -0,0 +1,91 @@
...
+  @impl Oban.Worker
+  def perform(_job) do
+    rows =
+      Date.utc_today()
+      |> trial_population()
+      |> Enum.flat_map(&score_team/1)
+
+    persist(rows)
+    :ok
+  end
```
apata, line 26: If any of the numerous postgres queries run as part of the job fails, the effort spent on the whole batch is wasted. Could we create separate jobs to score and insert / replace each team?

extra/lib/plausible/workers/score_trial_prospects.ex
apata: Correction, the more teams there are to score, the longer the period of stress will be: the intensity will remain constant due to sequential approach.

apata (approved):
We discussed implementing this on a separate BI platform, but just to try something out first, I think it's worth it.

extra/lib/plausible/customer_support/trial_prospects.ex
```diff
@@ -0,0 +1,131 @@
...
+  @spec plan_kind([module()], non_neg_integer(), non_neg_integer()) ::
+          {:starter | :growth | :business, [String.t()]}
+  defp plan_kind(feature_modules, site_count, member_count) do
+    used = Enum.map(feature_modules, & &1.name())
+
+    feature_k = feature_kind(used)
+    site_k = site_count_kind(site_count)
+    member_k = member_count_kind(member_count)
+
+    kind = feature_k |> higher_kind(site_k) |> higher_kind(member_k)
```
zoldar, line 74:
```suggestion
    kind = Enum.max_by([feature_k, site_k, member_k], &@kind_rank[&1])
```

extra/lib/plausible/customer_support/trial_prospect.ex
```diff
@@ -0,0 +1,54 @@
...
+    field :forced_by, {:array, :string}, default: []
+    field :pageview_limit, :integer
+    field :over_top_tier, :boolean, default: false
+    field :estimated_mrr, :integer
+    field :computed_at, :utc_datetime
+
+    timestamps()
+  end
+
+  def changeset(prospect, attrs) do
```
zoldar: Given `changeset` is only needed for tests now, I'd use `Ecto.Changeset.change` there instead and remove it.

extra/lib/plausible/workers/score_trial_prospects.ex
```diff
@@ -0,0 +1,91 @@
...
+  defp build_row(team, traffic, site_ids) do
+    score =
+      TrialProspects.score(
+        traffic.estimated_monthly,
+        Teams.Billing.features_usage(team, site_ids),
+        length(site_ids),
+        Teams.Billing.team_member_usage(team)
+      )
+
+    now = DateTime.utc_now() |> DateTime.truncate(:second)
```
zoldar:
```suggestion
    now = DateTime.utc_now(:second)
```

zoldar (approved):
Left some optional suggestions

extra/lib/plausible/workers/score_trial_prospects.ex
ukutaht: IMO sequential is better because it means the job runs longer but it will never occupy more than 1 database connection. Parallel processing would be faster but also create more db pressure while it's running.

Not too worried about occupying a single db connection at 4am UTC, it would be a problem if the job runs too long.

ukutaht: If the job turns out to be too slow I can look into reducing the amount of postgres roundtrips in `Billing.feature_usage`

extra/lib/plausible/workers/score_trial_prospects.ex
ukutaht: Good point overall but feel like separating the jobs is a bit overkill. This functionality is not critical at all and failures have a low cost.

What do you think about limiting the blast radius so that in case something goes wrong during scoring we skip that team, log it to Sentry, and continue with the batch?

apata: Yes, that's a reasonable alternative. Maybe a little retry with delay for each as well?

extra/lib/plausible/customer_support/trial_prospect.ex
ukutaht: https://github.com/plausible/analytics/pull/6498/changes/85a27ebebf1b5f554c7b679aadc4808ca77be18e

PR #6501, opened 2026-07-12
Validate `page` param in Stats API breakdown endpoint

### LINKED ISSUE #6500: Stats API breakdown endpoint returns 500 on non-integer `page` parameter
### Description

The public Stats API `GET /api/v1/stats/breakdown` endpoint crashes with a **500 Internal Server Error** when the `page` query parameter is not a valid positive integer.

The parameter is passed directly into `String.to_integer/1`:

https://github.com/plausible/analytics/blob/master/lib/plausible_web/controllers/api/external_stats_controller.ex#L44

```elixir
page = String.to_integer(Map.get(params, "page", "1"))
```

`String.to_integer/1` raises `ArgumentError` on any non-numeric input, so the request fails with a 500 instead of a helpful client error.

### Steps to reproduce

```
GET /api/v1/stats/breakdown?site_id=<site>&property=event:page&page=foo
```

Also triggered by `page=1.5`, `page=` (empty), etc.

### Expected behavior

A clean **400 Bad Request** with a descriptive error message — consistent with the sibling `limit` parameter a few lines below, which is already validated via `Integer.parse/1` (`validate_or_default_limit/1`).

### Actual behavior

`500 Internal Server Error` (and a reported exception).

### Proposed fix

Validate `page` the same way `limit` is validated, returning a 400 for invalid or non-positive values. Happy to open a PR.

nafey7:
## Changes

The public Stats API `GET /api/v1/stats/breakdown` endpoint passed the `page` query parameter straight into `String.to_integer/1`, which raises `ArgumentError` on any non-numeric input — resulting in a **500 Internal Server Error**.

This PR validates `page` the same way the sibling `limit` parameter is already validated (`validate_or_default_limit/1`): invalid or non-positive values now return a clean **400** with a descriptive message.

```elixir
defp validate_or_default_page(%{"page" => page}) do
  case Integer.parse(page) do
    {page, ""} when page > 0 -> {:ok, page}
    _ -> {:error, "Please provide page as a number greater than 0."}
  end
end

defp validate_or_default_page(_), do: {:ok, 1}
```

## Tests

Added tests to `breakdown_test.exs` covering non-integer, negative, and zero `page` values (mirroring the existing `limit` validation tests).

## Notes

- Behavior for valid `page` values (including the default of `1` when omitted) is unchanged.
- CHANGELOG updated under a new **Fixed** section.

Fixes #6500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ All notable changes to this project will be documented in this file.
 - Always compare against the same time range in comparisons with "Today"
 - Added vertical indicator line to graph to make it easier to see what's hovered / selected
 
+### Fixed
+
+- Stats API `/api/v1/stats/breakdown` now returns a `400` error for an invalid `page` parameter instead of crashing with a `500`
+
 ### Removed
 
 - Removed the standalone team switcher page; team switching is now done from the topbar dropdown only
--- a/lib/plausible_web/controllers/api/external_stats_controller.ex
+++ b/lib/plausible_web/controllers/api/external_stats_controller.ex
@@ -40,9 +40,8 @@ defmodule PlausibleWeb.Api.ExternalStatsController do
          :ok <- validate_filters(site, query.filters),
          {:ok, metrics} <- parse_and_validate_metrics(params, query),
          {:ok, limit} <- validate_or_default_limit(params),
+         {:ok, page} <- validate_or_default_page(params),
          :ok <- ensure_custom_props_access(site, query) do
-      page = String.to_integer(Map.get(params, "page", "1"))
-
       %{results: results, meta: meta} =
         Legacy.Breakdown.breakdown(site, query, metrics, {limit, page})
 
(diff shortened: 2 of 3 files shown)


ukutaht (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ All notable changes to this project will be documented in this file.
 - Always compare against the same time range in comparisons with "Today"
 - Added vertical indicator line to graph to make it easier to see what's hovered / selected
 
+### Fixed
```
zoldar, line 21: Thanks for contributing. One small nit: there's a "Fixed" section for unreleased changes already at https://github.com/plausible/analytics/pull/6501/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR43

ukutaht:
Closing in favor of https://github.com/plausible/analytics/pull/6501

PR #6503, opened 2026-07-13
Fix manual lock grace period bug

RobertJoonas:
### Changes

The bug is caused by `Plausible.GracePeriod.active?/1` unconditionally returning `true` for any `manual_lock` grace_period. It should return `false` when the manual lock grace period has been terminated manually via CS.

The bug is pretty tricky, so I'll do my best explaining what happens under the hood:

* An enterprise customer has not upgraded in time and our customer support decides to hit the "Lock" button on their team
* This action: 1) locks the team instantly and 2) sets their (manual lock) `grace_period.is_over = true`
* At the next UTC midnight, `site_locker.ex` runs -- checks every team and ends up calling `Plausible.Teams.Billing.check_needs_to_upgrade/2`
* ❗️`check_needs_to_upgrade` ends up returning `:no_upgrade_needed` because the `Teams.GracePeriod.expired?(team)` cond branch wrongly returns `false`
* `site_locker.ex` unlocks the site but keeps the grace period in place  

At this point, we're already in a messed up state because on one hand, the team is unlocked, but on the other, it has a manually terminated grace period. Now we get to rendering `usage_notification` for the team in the team settings billing section:

* `usage_notification_type` tries to find the type of the notification to render
* From these three cond branches (see screenshot below), we incorrectly end up in the third one because:
  * the team is not locked (as mentioned above)
  * `Teams.GracePeriod.expired?(team)` returns false (also mentioned above)
  * `manual_lock_active?` returns false as well -- this is correct because `is_over` being `true` means it's not active
  * Finally, `GracePeriod.active?` returns true (the root bug)

image

* the `usage_notification` component then tries to render "days left" but no case clause matches `nil` -- a `manual_lock` grace period has `end_date = nil`

Expected behaviour:

* The manually locked team stays locked after site_locker runs (unless the usage has decreased enough to unlock)
* usage_notification renders `:dashboard_locked`

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/teams/grace_period.ex
+++ b/lib/plausible/teams/grace_period.ex
@@ -85,8 +85,8 @@ defmodule Plausible.Teams.GracePeriod do
     Date.diff(end_date, Date.utc_today()) >= 0
   end
 
-  def active?(%{grace_period: %__MODULE__{manual_lock: true}}) do
-    true
+  def active?(%{grace_period: %__MODULE__{manual_lock: true}} = team) do
+    manual_lock_active?(team)
   end
 
   def active?(_team), do: false
--- a/lib/plausible/teams/grace_period.ex
+++ b/lib/plausible/teams/grace_period.ex
@@ -81,12 +81,12 @@ defmodule Plausible.Teams.GracePeriod do
   """
   def active?(team)
 
-  def active?(%{grace_period: %__MODULE__{end_date: %Date{} = end_date}}) do
-    Date.diff(end_date, Date.utc_today()) >= 0
-  end
-
   def active?(%{grace_period: %__MODULE__{manual_lock: true}} = team) do
     manual_lock_active?(team)
+  end
+
+  def active?(%{grace_period: %__MODULE__{end_date: %Date{} = end_date}}) do
+    Date.diff(end_date, Date.utc_today()) >= 0
   end
 
   def active?(_team), do: false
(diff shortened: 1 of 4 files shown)


lib/plausible/teams/grace_period.ex
```diff
@@ -85,8 +85,8 @@ defmodule Plausible.Teams.GracePeriod do
     Date.diff(end_date, Date.utc_today()) >= 0
   end
 
-  def active?(%{grace_period: %__MODULE__{manual_lock: true}}) do
-    true
+  def active?(%{grace_period: %__MODULE__{manual_lock: true}} = team) do
```
zoldar, line 84: Would it make sense to move that clause to the top, so it always takes precedence over anything else? 🤔

zoldar (approved):
Nice work explaining all the context of this tricky bug 👏

lib/plausible/teams/grace_period.ex
RobertJoonas: Yes, I think it would. Although it wouldn't make a functional difference (because end date being a `%Date{}` and `manual_lock` being `true` are be mutually exclusive), a manual lock grace period getting an end date by mistake sounds more probable than a grace period randomly becoming `manual_lock: true`. https://github.com/plausible/analytics/pull/6503/commits/84b994363e8d50a75b78271fd36e8bc2fb48762e

PR #6506, opened 2026-07-15
Fix sample rate calculator estimated traffic logic

apata:
### Changes

This PR fixes how we estimate traffic to calculate the sample rate for the query. 

There were two issues. The first was noticed by a user, the second by me when I was fixing the first. 

## 1) Overestimating events to scan for longer queries

We were overestimating the events the query would scan when the query start date was significantly before site stats start date, e.g. date range is 1970 to 2026, but stats actually start on 2025. 

Previously, the calculator when run for this query would have estimated traffic to be 56 years worth of last month's average daily traffic.

Now, the calculator takes into account that stats start from 2025, so for this query, the estimated traffic is 1 years worth of last month's average daily traffic.

## 2) Underestimating events to scan for newer sites with high ingestion rate

We were underestimating the events the query would scan when the site's ingest counters have accumulated over less than 30 days. 

Previously, a query for a new site ingesting 10M events per day for 3 days would not have been sampled, but a query for an older site ingesting 10M events per day for the last 30 days would have been. Expected is that both are sampled with the same rate.

### Tests

### Changelog
- [x] This doesn't make a CE-user facing change

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/stats/sampling.ex
+++ b/extra/lib/plausible/stats/sampling.ex
@@ -1,9 +1,24 @@
 defmodule Plausible.Stats.Sampling do
   @moduledoc """
-  Sampling related functions
+  This module contains the logic to calculate the sample fraction to use for a particular query.
+
+  To do so, it
+  1) estimates how much traffic the site has per day
+  2) how many days of traffic does the query cover
+  3) calculates the fraction so as not to scan more than 10M events
+
+  ### Example calculation
+  traffic = 3M / day
+  query covers = 14 days
+  events in query = 3M / day * 14 days = 42M
+
+  10M = 42M * fraction
+  ->
+  fraction = 10M / 42M
+  fraction = 0.24
   """
+
   @default_sample_threshold 10_000_000
-  @sampling_window_days 30
 
   import Ecto.Query
 
@@ -70,19 +85,7 @@ defmodule Plausible.Stats.Sampling do
     do: :no_sampling
 
   def fractional_sample_rate(traffic_30_day, query) do
-    date_range = Query.date_range(query)
-    stats_start = query.site_native_stats_start_at
-
-    # A site's traffic before their stats begin date isn't queried,
-    # so it shouldn't figure in the traffic estimation
-    start_of_range_with_stats =
-      if stats_start && Date.before?(date_range.first, stats_start) do
-        stats_start
-      else
-        date_range.first
-      end
-
-    duration = Date.diff(date_range.last, start_of_range_with_stats)
+    duration = queried_duration_days(query)
 
     estimated_traffic = estimate_traffic(traffic_30_day, duration, query)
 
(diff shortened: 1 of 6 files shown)


apata:
There's one more buggy scenario that I noticed: when the end date is far in the future, we'll overestimate events to scan and use an inappropriate sample rate. 

A query from the year 2026 to the year 2100 would estimate events to scan to be ~75 years worth of last 30 month average event count, while the actual events to scan would be half a year's worth, up to now. Can't have events in the future.

extra/lib/plausible/stats/sampling.ex
```diff
@@ -70,7 +71,18 @@ defmodule Plausible.Stats.Sampling do
 
   def fractional_sample_rate(traffic_30_day, query) do
     date_range = Query.date_range(query)
-    duration = Date.diff(date_range.last, date_range.first)
+    stats_start = query.site_native_stats_start_at
+
+    # A site's traffic before their stats begin date isn't queried,
+    # so it shouldn't figure in the traffic estimation
+    start_of_range_with_stats =
```
ukutaht: We have an existing function that clamps the query's datetime range based on the actual start date of native data [here](https://github.com/plausible/analytics/blob/d5af396464c2bd3d844f0b54647c7488f98edd45/lib/plausible/stats/time.ex#L8-L20). Could it be used here to keep unified logic and code path?

extra/lib/plausible/stats/sampling.ex
```diff
@@ -92,11 +104,23 @@ defmodule Plausible.Stats.Sampling do
...
+         now: %DateTime{} = now,
+         site_native_stats_start_at: %NaiveDateTime{} = at
+       }) do
+    now
+    |> Date.diff(at)
+    |> min(@sampling_window_days)
+    |> max(1)
+  end
+
+  defp days_of_traffic_in_window(%Query{}), do: @sampling_window_days
```
zoldar: Under what circumstances is this clause expected to match? AFAIK `native_stats_start_at` is always filled by default and has a non-null constraint on it

apata: Well spotted! It was leftover from when I used `stats_start_at()`, rather than `native_stats_start_at`. Removing.

Though there's no guarantee that the non-null site field reaches the `Query`, where the field is `nil` by default. Still, your point stands: if that happens, it's better to crash.

extra/lib/plausible/stats/sampling.ex
apata: Thanks! I made use of it now in https://github.com/plausible/analytics/pull/6506/commits/4354b921096be468c7a79aca91d3d9eca5642a0d. In the commit I also fix another bug concerning duration estimates, which were off by one day for the style of queries we use.

ukutaht (approved):

PR #6507, opened 2026-07-15
Fix wrong asserts in exploration test filtered by Firefox scenario

apata:
### Changes

The asserts were wrong for one test step. I've verified that there's indeed one visitor from Firefox in that step. They passed in some cases because the test runs fast enough to evaluate on the column's previous values, which aren't cleared up instantly on clicking the filter.

### DIFF
--- a/e2e/tests/dashboard/exploration.spec.ts
+++ b/e2e/tests/dashboard/exploration.spec.ts
@@ -1143,7 +1143,7 @@ test('change filters during a 3-step journey', async ({ page, request }) => {
 
     await expect(
       firstColumn.getByTestId('exploration-row').getByTestId('metric-value')
-    ).toHaveText(['2'])
+    ).toHaveText(['1'])
 
     await expect(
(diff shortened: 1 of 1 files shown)


aerosol (approved):

aerosol:
> they passed in some cases because the test runs fast enough to evaluate on the column's previous values, which aren't cleared up instantly on clicking the filter.

Sorry, I'm not super clear on that one. Would that mean that even if the expected values are correct now, we'll still have some flaky behaviour due to some timing issues?

zoldar:
@aerosol AFAIK, assertions are retried until the desired state is reached. Before the change, the wrongly asserted values did appear in a transient state while the UI was loading/refreshing. Now the assertion is made on correct, eventually expected values - so they should always succeed.

zoldar (approved):

aerosol:
ah, good to know, thanks @zoldar & @apata

PR #6509, opened 2026-07-15
Fix tracking for site installation method

zoldar:
### Changes

Fixes site installation method added in https://github.com/plausible/analytics/pull/6461.

LiveView is hijacking events, so class based custom events won't work (at least not reliably). This PR switches to hook based metrics submission which was originally implemented when moving user registration form to LiveView years ago.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/installation.ex
+++ b/lib/plausible_web/live/installation.ex
@@ -81,6 +81,7 @@ defmodule PlausibleWeb.Live.Installation do
 
     {:ok,
      assign(socket,
+       trigger_submit: false,
        site: site,
        site_created?: params["site_created"] == "true",
        flow: flow
@@ -152,7 +153,12 @@ defmodule PlausibleWeb.Live.Installation do
             />
           <% end %>
 
-          <.form for={@tracker_script_configuration_form.result} phx-submit="submit" class="mt-4">
+          <.form
+            for={@tracker_script_configuration_form.result}
+            phx-submit="submit"
+            phx-trigger-action={@trigger_submit}
+            class="mt-4"
+          >
             <.input
               type="hidden"
               field={@tracker_script_configuration_form.result[:installation_type]}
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/installation.ex
```diff
@@ -152,7 +153,12 @@ defmodule PlausibleWeb.Live.Installation do
             />
           <% end %>
 
-          <.form for={@tracker_script_configuration_form.result} phx-submit="submit" class="mt-4">
+          <.form
```
ukutaht, line 155: I haven't tried it but I wonder if a plain `onsubmit="window.plausible('event_name', props)"` javascript handler would work as well.

Reasoning: We used to have to wait for the POST /api/event request to be completed before submitting the form because we were relying on XMLHTTPRequest for delivery which cancels the request as soon as the page unloads.

Since last autumn we use fetch-with-keepalive for event delivery which means that the request will be completed by the browser even in case the page unloads due to form submit.

The class-based event triggering system does not work reliably because it's implemented such that the event handler is set on the document and it depends on the event bubbling up the document hierarchy. When Liveview submits the form, it suppresses the event propagation (this one I'm least sure of).

There's a couple of if's in there that I haven't confirmed but could be worth checking to simplify this and the register form submit handler as well.

ukutaht: To zoom out even more (not at all relevant for this PR) - for class-based event delivery as well as other auto-triggered events like outbound links and form submissions, we register the event handler at the document level. The reason for this is to make it work in case a form, link, or class has been added dynamically via client-side rendering. But this method does rely on the event being propagated all the way to the document level in the DOM. Liveview forms is a good example where this approach breaks down.

A better way to do this in the tracker would be to add event listeners on each of the tracked elements directly, and deal with dynamic DOM by registering a MutationObserver so we can register/deregister listeners as the DOM is manipulated in real time. This would allow our listeners to be registered on the elements they are tracking directly and therefore be resilient to the JS frameworks hijacking events and suppressing their propagation.

This would be a relatively big change to the tracker behaviour but all things considered desirable in my opinion. It would make the original class based implementation work in this case for example.

zoldar: I have added a card for this improvement, so that it won't get lost completely: https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/10102481064

The `onsubmit` listener approach you have suggested does indeed work: https://github.com/plausible/analytics/pull/6514. I'll close the current PR in favour of that one.

zoldar:
Closing in favour of https://github.com/plausible/analytics/pull/6514

PR #6512, opened 2026-07-16
SSRF protection measures (multiple PRs)

aerosol:
### Changes

Initial IP classification interface - to be built upon in follow-up PRs (merged here already)

### Tests

### Changelog

### Documentation

### Dark mode


mix.exs
```diff
@@ -103,6 +103,7 @@ defmodule Plausible.MixProject do
       {:location, git: "https://github.com/plausible/location.git"},
       {:mox, "~> 1.0", only: [:test, :ce_test, :e2e_test]},
       {:nanoid, "~> 2.1.0"},
+      {:nimble_csv, "~> 1.3"},
```
aerosol, line 106: It's used elsewhere too but was never put up as an explicit dependency

lib/mix/tasks/download_ip_registries.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Mix.Tasks.DownloadIpRegistries do
```
zoldar, line 1: You probably have already thought about it, but we could also add a CI check comparing local and remote CSVs which would at least warn if the files are out of date?

zoldar (approved):
Mad props for figuring out translating ip/mask to patterns/guards 🙇

lib/mix/tasks/download_ip_registries.ex
aerosol: Ah no, that's a good idea but not sure if worth doing that on every CI run. Is there a way to do it e.g. monthly?

zoldar: Uh, one way would be a conditional run based on the current time / calendar + CI cached file signifying that the current run was already completed? That would make CI fail at most once to give a heads up. Just an idea off the top of my head, maybe there's something less hacky for that purpose.

aerosol: Looks like we have cron at our disposal: https://github.com/marketplace/actions/set-cron-schedule

aerosol: Done

cnkk (approved):

PR #6516, opened 2026-07-20
Verification to dashboard

RobertJoonas:
### Changes

Move verification (all flows: `provisioning`, `review`, and `domain_change`) to the dashboard (between the top bar and sources/pages reports). It will remain a LiveView, but will be teleported into the stats-react-container with the help of `<.portal>`. Various UI changes in this PR to make the UI feel more "banner-like".

> _For reviewers: For UI testing, I encourage playing around with it locally -- this PR builds a dev convenience for mocking different verification outcomes. The same code is also used for E2E testing._

In addition to verification, we'll also move the email reports CTA from `stats.html.heex` to React. The reason for that is positioning -- otherwise impossible to render it between top bar and stats reports. 

### New db field and onboarding logic

This PR also includes a migration, adding a new `site.onboarding_status` field which makes all of this possible. There are currently four different statuses possible:

* `:new_site`
  * The site was just created, didn't succeed verification, didn't receive any pageviews yet
  * Means that a "Setup required" pill should be rendered on this site card on the /sites page
  * Any links to this dashboard (such as the one in `site-switcher.js`) should append `?verify_installation=true&flow=provisioning` to the URL to initiate a verification banner upon entering the dashboard.
* `:verification_completed`
  * Advances to this state once a verification success is encountered (no matter through what flow -- `provisioning` / `review` / `domain_change`)
  * Means that "Setup required" won't ever render for this site again, and verification banner won't be triggered automatically for this site anymore. The only way to get verification again is Site Settings > Review Installation
* `:first_pageview`
  * The site has received at least one pageview
  * Advances to this state when:
    * a full dashboard load happens and the site has pageviews (piggybacking on the same condition where `site.stats_start_date` is set for the very first time)
    * the site card on the /sites page is about to display "X visitors in the last 24h" (where X > 0) for the very first time
    * (note that either of the above can happen **before** verification success is encountered, in which case that step is simply skipped. This is also exactly what happens on CE)
  * Similar to `:verification_completed`, we won't automatically trigger verification anymore
  * In this state, the email reports CTA starts appearing to any site member with editor permissions (owner, admin, editor)
* `:completed`
  * Everything related to onboarding is done
  * This is the database default for the `onboarding_status` field -- running the migration will set the new field value to `:completed` for every existing site.
  * As for new sites (after this PR gets deployed), the only way to get to this state will be to dismiss or action (click the `/:domain/settings/email-reports` link) the email reports CTA banner.

### Migration
- [ ] Extract the migration and execute it

### Tests

### Changelog
- [x] Need to update and review the flows on CE

### Documentation

### Dark mode


lib/plausible_web/live/sites.ex
```diff
@@ -552,7 +559,17 @@ defmodule PlausibleWeb.Live.Sites do
...
       <.unstyled_link
-        href={Routes.stats_path(PlausibleWeb.Endpoint, :stats, @site.domain, [])}
+        href={
+          Routes.stats_path(
+            PlausibleWeb.Endpoint,
+            :stats,
+            @site.domain,
+            if(@needs_verification?,
+              do: [verify_installation: true, flow: PlausibleWeb.Flows.provisioning()],
+              else: []
```
apata, line 570: suggestion, non-blocking: Can we avoid the verification being triggerable with users manually entering these URL params? 

I think it'd be better if this server side process started with a user interaction, not navigation. 

Because the redirect is handled by the server, maybe we can pass the fact that the verification banner should happen server-side, in conn.assigns?

If we pass this knowledge server-side, the user refreshing the dashboard while verification is happening would clear the verification banner (_unless the verification is forced on them by server because they don't have stats_start_date_) - they'd have to click browser back and click verify again to see it again.

RobertJoonas: > suggestion, non-blocking: Can we avoid the verification being triggerable with users manually entering these URL params?

Could you elaborate on why? Is this about security, or more about code architecture/cleanliness?

I don't think there's an actual security concern here -- `verify_installation?` already requires an authenticated user with access to that specific site, the check itself is rate-limited, and "re-verification" for old sites is already reachable through Settings -> Review installation anyway. Also, already triggered by navigation (the `/:domain/verification` route), so I don't think doing that based on a query parameter on the dashboard exposes any new vulnerability.

> Because the redirect is handled by the server, maybe we can pass the fact that the verification banner should happen server-side, in conn.assigns?

`conn.assigns` is just a map scoped to a single request/response cycle — it doesn't survive a redirect at all, so it can't do this. `Conn.put_session` could keep the trigger hidden from the user instead (or `Phoenix.Controller.put_flash` to limit it to a single page load).

`put_session` could technically work, but it would get tricky — LiveView sockets can't write to the session directly, so both submit action on the Installation LV (to set it) and the dismiss action on the Verification banner LV (to clear it) would need to go through controller-level round-trips. Not worth the hustle I think.

> If we pass this knowledge server-side, the user refreshing the dashboard while verification is happening would clear the verification banner (unless the verification is forced on them by server because they don't have stats_start_date) - they'd have to click browser back and click verify again to see it again

I don't think limiting verification to a single page load is good UX — I imagine many users would have the intuition to refresh the page to re-trigger a failed verification (instead of clicking "Check again"). So I'd leave verification visible until the user *decides to dismiss it*, OR once verification succeeds and the installation is confirmed working — that's the point at which we'd hide the banner on a refresh.

Also, the "server-side forcing verification" part is exactly what I wanted to avoid — the user should be able to dismiss it and not see it again even after a page refresh.

apata: Thanks for the detailed reply! 

The concern is security. It's in the same broader category that it shouldn't be possible to build links that delete site or user when people click them. Basically, any mutation should need a user interaction. Verification isn't a mutation, but it's close. 

Regarding implementation, thanks for considering it, looks like put_flash would be the way to do it, not ideal but not entirely custom either.

You're right to point out that it's already possible to build links that trigger verification so this PR wouldn't expose anything new, but it makes previously safe-looking links unsafe: https://plausible.io/plausible.io?f=is,page,/blog&f=is,source,Direct%20/%20None&f=is,screen,Desktop&verify_installation=true&flow=domain_change&f=is,os,Windows&f=is,browser,Firefox&f=is,region,DE-HE&f=is,country,DE&l=DE-HE,Hesse&l=DE,Germany

> Also, the "server-side forcing verification" part is exactly what I wanted to avoid — the user should be able to dismiss it and not see it again even after a page refresh.

Agreed, dismiss should dismiss. Should be possible with flash.

Regarding users potentially refreshing the page and losing the server side grant to verification, you're probably right. But then they can press the back button and click whatever they clicked to start verification and it will restart. Some tradeoffs are to be expected.

apata: Anyways, this is non-blocking, just wanted to put this option on the table

lib/plausible/application.ex
```diff
@@ -194,6 +194,11 @@ defmodule Plausible.Application do
         end,
         Plausible.Ingestion.Counters,
         Plausible.Session.Salts,
+        on_ee do
+          if Mix.env() in [:dev, :e2e_test, :test] do
```
ukutaht: This runtime call to Mix.env() will not work in a release.

Better to extract into a module attribute like:

```elixir
  @mock_scenarios? Mix.env() in [:dev, :e2e_test, :test]

# later
on_ee do
  if @mock_scenarios?, do: Plausible.InstallationSupport.Verification.MockScenarios
end
```

assets/js/dashboard/verification/portal.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export const VerificationLiveViewPortal = React.memo(() => {
+  const navigate = useAppNavigate()
+
+  useEffect(() => {
+    function handleVerificationFinished(event: Event) {
+      const { queryParams } = (event as CustomEvent<VerificationFinishedDetail>)
+        .detail
+
+      navigate({
```
ukutaht, line 28: Hmm I'm thinking using `replace: true` here would be more fitting so the path with URL params does not stay in the browser history.

ukutaht (approved):
It was a lot to review so I don't feel like I could pay full attention to everything. Overall looks all good to me.

In general it doesn't feel great to me to have liveview running alongside React like this on the dashboard long-term but I also understand why it was done this way.

lib/plausible/application.ex
RobertJoonas: https://github.com/plausible/analytics/pull/6516/commits/3290d257214ed2a3ed2fb83a2b5898d35e8a9e01

assets/js/dashboard/verification/portal.tsx
RobertJoonas: Agreed. Did that in https://github.com/plausible/analytics/pull/6516/commits/90342eba3c50ba0bdc8ea7ea156e2f9b3e49a55b.

There's still one quirk where if you navigate in React (e.g. change period/filters) then the verification params will stay in those history entries. So for example:

0) verification banner is visible (either loading or failed) and the params are in the URL
1) open the page filter modal, then close it
2) close the verification banner (params are stripped from the URL)
3) hit back in browser history (page filter modal appears)
4) Quirk: the verification params re-appear in the URL as well, refreshing the page now will bring verification back

In practice though, I think it's very unlikely to hurt the UX because:

* although some clicking-around on the dashboard to discover what it can do is expected, going back in browser history when there's nothing meaningful happening on the dashboard yet is probably rare.
* even though the query params might re-appear, the verification will only come back if 1) it didn't succeed before, **and** 2) the dashboard doesn't have any pageviews yet.
* the worst case scenario is: user has to dismiss the banner for a second time.

PR #6521, opened 2026-07-21
Anti SSRF Client - on top of #6521

aerosol:
### Changes

Base branch: ip-tools (#6512)
This PR implements SSRF protection - a Req wrapper meant to be used with arbitrary user-supplied hostnames. 

Split from a bigger chunk of work, incoming PRs will be opened on top.

### Tests
- [x] Automated tests have been added
- [ ] This PR does not require tests

### Changelog
- [ ] Entry has been added to changelog
- [x] This PR does not make a user-facing change

### Documentation
- [ ] [Docs](https://github.com/plausible/docs) have been updated
- [x] This change does not need a documentation update

### Dark mode
- [ ] The UI has been tested both in dark and light mode
- [x] This PR does not change the UI

PR #6522, opened 2026-07-21
SSRF-aware SSO domain provisioning (on top of #6521)

aerosol:
### Changes

Base branch: ip-tools-ssrf-client (#6521)

SSRF-aware SSO domain provsioning verification

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1100,3 +1100,8 @@ unless s3_disabled? do
 end
 
 config :plausible, Plausible.Cache.Adapter, sessions: [partitions: 100]
+
+config :plausible, Plausible.IP.Tools,
+  allow_reserved_ips?:
+    config_env() in [:dev, :ce_dev] and
+      get_bool_from_path_or_env(config_dir, "ALLOW_RESERVED_IPS", false)
--- a/lib/ip/tools.ex
+++ b/lib/ip/tools.ex
@@ -52,18 +52,14 @@ defmodule Plausible.IP.Tools do
   end
 
   def allowed?(ip) when is_tuple(ip) and tuple_size(ip) in [4, 8] do
-    case allow_reserved_ips?() do
-      true ->
-        if reserved?(ip) do
-          Logger.warning(
-            "IP #{inspect(ip)} forcibly allowed due to :allow_reserved_ips? env set. "
-          )
-        end
+    if allow_reserved_ips?() do
+      if reserved?(ip) do
+        Logger.warning("IP #{inspect(ip)} forcibly allowed due to :allow_reserved_ips? env set. ")
+      end
 
-        true
-
-      false ->
-        not reserved?(ip)
+      true
+    else
+      not reserved?(ip)
     end
   end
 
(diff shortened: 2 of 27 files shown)


config/runtime.exs
```diff
@@ -1100,3 +1100,8 @@ unless s3_disabled? do
 end
 
 config :plausible, Plausible.Cache.Adapter, sessions: [partitions: 100]
+
+config :plausible, Plausible.IP.Tools,
+  allow_reserved_ips?:
+    config_env() in [:dev, :ce_dev] and
+      get_bool_from_path_or_env(config_dir, "ALLOW_RESERVED_IPS", false)
```
apata, line 1107: suggestion, non-blocking: Rather than an on-off switch, I suggest we configure a whitelist of reserved IPs. That way, we still exercise the feature in dev

aerosol: My network setup is different than yours, someone's docker is different than those, I don't want to deal with that

apata: In my opinion, needing to adjust the whitelist for my particular dev env is the lesser evil here, compared to disabling the logic completely in dev. To be able to disable it, we create code paths around the feature which may become exploitable in prod.

aerosol: How do you exploit that on prod?

apata: Hmm, that's a valid question, but more appropriate if this were already deployed to prod and we were discussing whether it needs to be patched, and how urgently. 

It's not yet, so IMO the matter to resolve is whether a whitelist config is sufficiently better than an on-off config to justify refactoring this PR. 

My arguments are that a whitelist keeps exercising the feature even in dev (_good for parity_) and that having a code path to toggle it off is riskier for us (we mustn't forget to use the toggled-on path when working on XYZ) than having it always on but respecting a whitelist. The counterargument is that a whitelist may cause dev frustration when setting up a non-standard dev env.

zoldar: Guarding that switch with `config_env` is strong enough measure to make it unexploitable in production. The same way it is for test routes in the router or other parts of logic guarded in a similar fashion.

apata: @zoldar But the true/false config value isn't the guard. `skip_checks?` propagating through the application code is.

By going the whitelist route, there's no need for `skip_checks?` code path: in every env - dev, test, prod - we will go through the all the motions to guard against SSRF, with one of the checks being the whitelist. 

In some envs, the whitelist will have a few values. In prod, no values. 

It's common practice in this repo to have these separate code paths, that's why I'm not requesting changes here, just trying to highlight how there's other ways as well.

lib/ip/tools.ex
```diff
@@ -49,8 +52,24 @@ defmodule Plausible.IP.Tools do
   end
 
   def allowed?(ip) when is_tuple(ip) and tuple_size(ip) in [4, 8] do
-    not reserved?(ip)
+    case allow_reserved_ips?() do
```
zoldar: `if` should be fine here?

config/runtime.exs
zoldar: This still applies a safe default, like in many other similar cases.

The whitelist will IMHO be a PITA to manage, adding complexity for little benefit.

zoldar (approved):

cnkk (approved):

lib/ip/tools.ex
aerosol: ba47ef2598

PR #6529, opened 2026-07-22
Fix flaky special goals test by waiting for verification page to load fully

apata:
### Changes
The e2e test testing special goals was flaky. We proceeded to the dashboard in e2e tests too quick from the verification screen for the automatic async special goal creation logic to run. This adds an await on the verification screen, which allows the special goal creation logic to run.


zoldar (approved):
👍

PR #6540, opened 2026-07-23
Remove limitation that blocked users without owned sites from startin…

sanne-san:
…g a subscription.

### Changes

- The "No sites owned" upgrade gate no longer applies; any team can subscribe regardless of site count.
- Site transfer form now offers an upgrade link when moving to "My personal sites" without an active subscription.
- The reason for doing this is that with the recent improvements to the site transfer flow, we're already detecting at the start whether a site can be moved to My personal sites. It's not possible to initiate a site transfer to My personal sites without a subscription, but it's also not possible to start a subscription without sites; thus creating a dead end.

CleanShot 2026-07-23 at 13 02 02@2x
CleanShot 2026-07-23 at 12 41 16@2x

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ All notable changes to this project will be documented in this file.
 - Update custom range datepicker styles
 - Improved site transfer UI
 - Redesigned authentication pages (register, sign in, 2FA, password reset, account activation)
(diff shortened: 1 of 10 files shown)


zoldar (approved):
✨

PR #6546, opened 2026-07-24
Friendly captcha

ukutaht:
### Changes

Replaces hCaptcha with invisible Friendly Captcha. Due to the 'invisible' part, some frontend complexity is necessarily added.

NB ⚠️ Do not merge. Requires Friendly Captcha credentials to be set up on prod before deploying.

### Tests

This part is manually tested.

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/captcha.ex
```diff
@@ -0,0 +1,105 @@
...
+        This site is protected by
+        <PlausibleWeb.Components.Generic.styled_link href="https://friendlycaptcha.com" new_tab={true}>
+          Friendly Captcha
+        </PlausibleWeb.Components.Generic.styled_link>
+      </p>
+      <script
+        phx-update={if @live?, do: "ignore"}
+        id="frc-captcha-script"
+        type="module"
+        src="https://cdn.jsdelivr.net/npm/@friendlycaptcha/sdk@1/site.min.js"
```
cnkk: wdyt about self hosting instead, see https://developer.friendlycaptcha.com/docs/v2/getting-started/install#using-the-scripts-without-a-cdn-ie-self-hosting

ukutaht: Good point, thanks https://github.com/plausible/analytics/pull/6546/changes/50846887b09c42d120a4312ba2ed712e5efa8584

lib/plausible_web/components/captcha.ex
```diff
@@ -0,0 +1,115 @@
...
+
+  def widget(assigns) do
+    ~H"""
+    <div>
+      <div
+        phx-update={if @live?, do: "ignore"}
+        id="frc-captcha-placeholder"
+        class="frc-captcha hidden mb-2"
+        data-sitekey={PlausibleWeb.Captcha.sitekey()}
+        data-start="auto"
```
cnkk, line 32: I still have to click on "I am human", shouldn't this auto solve and click?

ukutaht: No, this is controlled by the 'widget mode' config on their backend: https://developer.friendlycaptcha.com/docs/v2/sdk/configuration#widget-mode

I have it on one-click now for testing but will switch to smart on prod:

Screenshot 2026-08-04 at 15 49 45

aerosol (approved):
LGTM

lib/plausible_web/live/register_form.ex
```diff
@@ -259,7 +244,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
         %{assigns: %{invitation: %{} = invitation}} = socket
       ) do
     if not PlausibleWeb.Captcha.enabled?() or
-         PlausibleWeb.Captcha.verify(params["h-captcha-response"]) do
+         PlausibleWeb.Captcha.verify(params["frc-captcha-response"]) do
```
RobertJoonas: Nitpick: We could omit the `not PlausibleWeb.Captcha.enabled?()` check here (as password reset does). `verify` returns `true` if not enabled anyway.

Same goes for line 262 in this file (non-invitation register event handler).

lib/plausible_web/live/register_form.ex
```diff
@@ -165,7 +141,16 @@ defmodule PlausibleWeb.Live.RegisterForm do
...
+          <.button
+            id="register"
+            type="submit"
+            class="w-full"
+            mt?={false}
+            x-data={"{ captchaReady: #{not PlausibleWeb.Captcha.enabled?()} }"}
+            x-on:frc-captcha-ready.window="captchaReady = true"
+            x-on:frc-captcha-reset.window="captchaReady = false"
+            x-bind:disabled={"!captchaReady || #{@disable_submit}"}
+          >
```
RobertJoonas, line 153: Currently, a failed submit (for whatever reason, e.g. email already exists) will use up the one-time token and the next submit is guaranteed to fail with a "Please complete the captcha" error. Only a page refresh will fix that, meaning that the user needs to fill out all the fields again.

In LiveViews, when a form submit failure comes back from the server, we should make sure to reset the captcha automatically.

ukutaht: Thanks for catching that! Solved here: https://github.com/plausible/analytics/pull/6546/changes/ceed5657f1a24a18953a35b4d06e6ff212217f7e

lib/plausible_web/live/register_form.ex
ukutaht: Thanks! Done here: https://github.com/plausible/analytics/pull/6546/changes/8bf04322024d638c07733258db5711cc31e8fb5d

RobertJoonas (approved):

PR #6547, opened 2026-07-26
Fix import purge on cleaner failure and reclaim leftover local CSV disk

### LINKED ISSUE #6515: ObanErrorReporter purges imported data when the cleanup worker crashes — silent loss of completed CSV imports
**Version:** CE v3.2.1 (self-hosted, community-edition compose)

**Summary:** `ObanErrorReporter.on_job_exception/1` matches crashed Oban jobs by `queue: "analytics_imports"` + presence of `import_id` in args — it never checks *which worker* crashed. `Plausible.Workers.LocalImportAnalyticsCleaner` runs in that same queue with `import_id` in its args. When the cleaner crashes (in our case: `File.rm!` → `:eacces` on import CSVs owned by a different uid), the reporter misinterprets it as a failed *import*:

- attempts 1–19 → `import_fail_transient/1` → `Plausible.Purge.delete_imported_stats!/1` — **deletes the completed import's ClickHouse data** while `site_imports.status` still reads `completed`
- attempt 20 (discard) → `import_fail/1` → import marked failed + failure email

**Observed impact:** 15 successful local-storage CSV imports had all their data purged ~1 hour after completion (the cleaner's first scheduled run), while dashboards showed the imports as completed for 13 days. The only user-visible signal came two weeks later when the cleaner jobs exhausted retries and each import flipped to failed with a "please try the import again" email.

**Repro sketch:** complete a local CSV import, make the uploaded files un-deletable by the app user (e.g. `chown root` them), wait for the cleaner's scheduled run, observe `imported_*` tables purged for that `import_id` while the import still reads completed.

**Suggested fix:** scope `on_job_exception/1` to `worker: "Plausible.Workers.ImportAnalytics"` (or exclude the cleaner), so cleanup failures can't purge import data. Arguably the cleaner also shouldn't share consequences with importers at all despite sharing the queue.

Happy to provide fuller forensics (query_log timelines, oban job history) if useful.

EmmS21:
## Summary

Fixes [#6515](https://github.com/plausible/analytics/issues/6515): when `LocalImportAnalyticsCleaner` crashed (often `File.rm!` → permission denied on self-hosted volumes), `ObanErrorReporter` treated that like an `ImportAnalytics` failure and purged already-imported ClickHouse stats.

1. **Fix the product bug** — Scope `import_fail` / `import_fail_transient` to `Plausible.Workers.ImportAnalytics` only. Cleanup crashes must not undo a completed import. I considered a separate queue for the cleaner; matching on worker fixes the misclassification with less churn.
2. **Log stuck cleanup** — On the cleaner’s final attempt, log path + `import_id` when delete still fails (ops/debug signal, not user email).

Leftover local CSV disk / volume ownership on CE images is a separate follow-up and belongs in Discussions as an image/ops change, not this bugfix.

## Test plan

- [ ] `mix test test/workers/oban_error_reporter_test.exs test/workers/local_import_analytics_cleaner_test.exs`
- [ ] Confirm ImportAnalytics exceptions still purge / mark failed / email as before
- [ ] Confirm LocalImportAnalyticsCleaner exceptions do **not** purge stats or mark the import failed
- [ ] Local cleaner: deletes when import completed; snoozes while importing; logs only on final undeleteable attempt

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ ENV MIX_ENV=$MIX_ENV
 RUN adduser -S -H -u 999 -G nogroup plausible
 
 RUN apk upgrade --no-cache
-RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates su-exec \
+RUN apk add --no-cache openssl ncurses libstdc++ libgcc ca-certificates \
   && if [ "$MIX_ENV" = "ce" ]; then apk add --no-cache certbot; fi
 
 COPY --from=buildcontainer --chmod=555 /app/_build/${MIX_ENV}/rel/plausible /app
@@ -76,6 +76,7 @@ COPY --chmod=755 ./rel/docker-entrypoint.sh /entrypoint.sh
 # docker container can be started with arbitrary uid
 RUN mkdir -p /var/lib/plausible && chmod ugo+rw -R /var/lib/plausible
 
+USER 999
 WORKDIR /app
 ENV LISTEN_IP=0.0.0.0
 ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ ENV MIX_ENV=$MIX_ENV
 RUN adduser -S -H -u 999 -G nogroup plausible
 
 RUN apk upgrade --no-cache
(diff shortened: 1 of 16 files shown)


EmmS21:
Further discussion:
https://github.com/plausible/analytics/discussions/6548

PR #6555, opened 2026-07-28
Update to node 24

apata:
### Changes

Updates node version to 24 (LTS)

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -242,7 +242,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 24.18.0
+          node-version: 24.17.0
           cache: 'npm'
           cache-dependency-path: e2e/package-lock.json
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


aerosol (approved):

.github/workflows/elixir.yml
```diff
@@ -242,7 +242,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 23.2.0
+          node-version: 24.18.0
```
cnkk: let's use `24.17.0` everywhere instead?
it seems to be the lowest common version. the alpine image we use during build doesn't have `24.18.0` available

apata: You're right to point this out, @cnkk. Thinking deeper on it, here's how I see it. 

The main benefit of using the lowest common version would be to have highest possible parity between as-tested and as-deployed. That's important for our must-pass CI workflows to be meaningful. 

The risk is missing out on patches in envs that potentially would be vulnerable (local and CI machines).

However, the one instance of lower version of Node is used in the build container only, whereas local and CI machines are on the internet, so I'm leaning towards leaving the discrepancy in. WDYT?

cnkk: I think we should still test/run on the same version we're building the assets on for the release. Security wise this is acceptable

PR #6556, opened 2026-07-28
Add script for purging replayed sessions and events by ID and date range

zoldar:
### Changes

This PR implements a helper script for safely deleting replayed events.

It explicitly prevents deletion of normal events and sessions (replay_session_id = 0).

As the date range is strictly limited to 3 days at most, the number of partitions involved in the operation should be low enough to not require further optimisations.

### DIFF
--- a/lib/plausible/data_migration/purge_ingest_replay.ex
+++ b/lib/plausible/data_migration/purge_ingest_replay.ex
@@ -61,7 +61,7 @@ defmodule Plausible.DataMigration.PurgeIngestReplay do
 
     if days_scope < 0 or days_scope > 2 do
       raise ArgumentError,
-            "The period declared by inclusive from/to range cannot be longer than 3 days "
+            "The period declared by inclusive from/to range cannot be longer than 3 days"
     end
 
     %{rows: [[session_count]]} =
--- a/lib/plausible/data_migration/purge_ingest_replay.ex
+++ b/lib/plausible/data_migration/purge_ingest_replay.ex
@@ -1,6 +1,16 @@
 defmodule Plausible.DataMigration.PurgeIngestReplay do
   @moduledoc """
-  Script purging ingest replay
+  Script purging ingest replay.
+
+  The script is used in the ingest replay procedure when a particular replay turns
+  out to be incorrect - for instance, due to too much overlap with normally ingestged
+  events.
+
+  Three options are required:
+
+  - `session_id` - `replay_session_id` to purge, provided as an integer
+  - `from` and `to` - provided as Dates; necessary for reducing scope of database scan; 
+    the date range is inclusive; to scan a single day, `from` and `to` can be the same
   """
 
   alias Plausible.IngestRepo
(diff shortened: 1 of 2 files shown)


lib/plausible/data_migration/purge_ingest_replay.ex
```diff
@@ -0,0 +1,115 @@
...
+    AND toDate(timestamp) >= {$1:Date} 
+    AND toDate(timestamp) <= {$2:Date}
+  """
+
+  def run(opts \\ []) do
+    from = %Date{} = Keyword.fetch!(opts, :from)
+    to = %Date{} = Keyword.fetch!(opts, :to)
+    session_id = String.to_integer(Keyword.fetch!(opts, :session_id))
+
+    days_scope = Date.diff(to, from)
```
apata, line 56:
```
iex(1)> Date.diff(~D[2026-07-28], ~D[2026-07-27])
1
```
This diff yields 1, yet both to and from are inclusive (>=, <=), so these from, to arguments actually target two days. This may cause some confusion what is intended

lib/plausible/data_migration/purge_ingest_replay.ex
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    if days_scope < 0 or days_scope > 2 do
+      raise ArgumentError, "The number of days between the dates must be from 0 to 2 range"
+    end
+
+    %{rows: [[session_count]]} =
+      IngestRepo.query!(
+        @count_query_sessions,
+        [session_id, Date.add(from, -2), to, from, to],
```
apata: I'm a bit confused by this query. Why do we move backwards in time from `from` to filter session start dates? Can a replayed session start earlier than the first event in the replay? Maybe a private function with keyword args would allow the code to explain itself better

zoldar: That was meant as an optimisation because of the lack of indexing of `timestamp` in sessions. However you are right, replayed sessions should start strictly within declared time range - I'll simplify the queries.

zoldar: https://github.com/plausible/analytics/pull/6556/commits/195b978433368b7dc30d022ea45573201a3187d4

lib/plausible/data_migration/purge_ingest_replay.ex
zoldar: I've documented the parameters in https://github.com/plausible/analytics/pull/6556/commits/e33963e5d0421f33a6a7dc5ec3d867fc2faf3037. BTW `session_id` is now accepted as a plain integer om https://github.com/plausible/analytics/pull/6556/commits/96bc3c25983e96063fac33076837f517be26e693 - there was no reason to force it in string form.

apata: It clarifies it a bit, thanks! 
However, it's still a bit of an issue IMO because the PR description states max range is 2 days, but the code allows days_scope's 0, 1, 2, where 0 is means 1 day will be cleaned [~D[2026-07-27], ~D[2026-07-27]], 1 is 2 days cleaned, 2 is 3 days cleaned.

zoldar: I have lined up the PR description, range argument error and module doc to consistently mention it's max of 3 days with date range inclusive

aerosol (approved):

PR #6557, opened 2026-07-28
Improve onboarding pages

sanne-san:
### Changes
- Give the add site and installation pages a plain white layout with a heading, subtitle and a progress indicator at the bottom of the page
- Remove the timezone field from the add site form, and detect it automatically
- Rebuild the installation instructions as numbered steps with click-to-copy code, and hide the extra options during onboarding
- Offer a way out of every step with a secondary button, either skipping ahead or heading back to where you came from
- Drop the separate verification step from the progress indicator, as verification now happens on the dashboard
- Tidy up shared form and UI details along the way: help text under fields, read-only inputs, smaller toggles and an improved loading spinner

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/live/installation/instructions.ex
```diff
@@ -3,87 +3,97 @@ defmodule PlausibleWeb.Live.Installation.Instructions do
...
-    <div class="text-sm my-4 leading-6">
-      Paste this snippet into the <code>&lt;head&gt;</code>
-      section of your site. See our
-      <.styled_link href="https://plausible.io/docs/integration-guides" new_tab={true}>
-        installation guides.
-      </.styled_link>
-      Once done, click the button below to verify your installation.
+      <p class="text-sm text-gray-500 dark:text-gray-400">
+        Need help?
+        <.styled_link href="https://plausible.io/docs/plausible-script">
```
RobertJoonas: Let's keep this opening in a new tab? I think better to leave the installation page open.

lib/plausible_web/live/installation.ex
```diff
@@ -82,7 +82,12 @@ defmodule PlausibleWeb.Live.Installation do
     {:ok,
      assign(socket,
        site: site,
-       flow: flow
+       site_created?: params["site_created"] == "true",
```
RobertJoonas: This is dead code which was removed by moving verification to the dashboard. Missed that during rebase since there was no conflict. Safe to remove.

lib/plausible_web/live/installation.ex
```diff
@@ -197,19 +222,16 @@ defmodule PlausibleWeb.Live.Installation do
...
     </div>
     """
   end
 
   on_ee do
-    defp submit_button_text("manual"), do: "Verify Script installation"
-    defp submit_button_text("wordpress"), do: "Verify WordPress installation"
-    defp submit_button_text("gtm"), do: "Verify Tag Manager installation"
-    defp submit_button_text("npm"), do: "Verify NPM installation"
+    defp submit_button_text, do: "I've installed it"
```
RobertJoonas: Now that this is static, I think a module attribute would be more appropriate. Also, the reason why CE currently renders a different button text there is that there's no verification on CE. Since the button text does not mention "verify ..." anymore, I think it makes sense to also use the same value for EE and CE -- "I've installed it" sounds great!

lib/plausible_web/templates/layout/app.html.heex
```diff
@@ -30,7 +30,11 @@
   </head>
   <body
     class={[
-      "flex flex-col bg-gray-50 dark:bg-gray-950",
+      "flex flex-col dark:bg-gray-950",
+      if(assigns[:white_bg?],
```
RobertJoonas: Does this concern only light mode then? Reading this assign at first site definitely raises an eyebrow. Could we do `assigns[:bg_class]` instead and use the passed value if it exists?

So the default would be `bg-gray-50 dark:bg-gray-950`, and instead of `:white_bg?` we'd pass `bg-white dark:bg-gray-950`. WDYT?

lib/plausible_web/router.ex
```diff
@@ -598,10 +598,16 @@ defmodule PlausibleWeb.Router do
...
-        live "/:domain/installation", Installation, :installation, as: :site
+      live_session :onboarding, on_mount: PlausibleWeb.Live.OnboardingLayoutContext do
+        scope assigns: %{
+                dogfood_page_path: "/:website/installation"
+              } do
+          live "/:domain/installation",
+               Installation,
+               :installation,
+               as: :site,
+               container: {:div, class: "flex-1 flex flex-col"}
```
RobertJoonas, line 609: Is this necessary? Tried removing it and can't tell the difference on the page.

sanne-san: Yeah unfortunately, LiveView always wraps the page in a div, so removing it will result in a div with `display: block` rather than `flex`. Without `flex-1` it doesn't stretch to full height, so the progress bar ends up right below the content instead of at the bottom of the screen on the installation page.

Image

lib/plausible_web/live/installation/instructions.ex
sanne-san: Yeah, the reason I removed it is because in some cases I'd rather not show the icon. I've ran into this before though so I now decoupled the new tab behaviour from displaying the icon instead.

lib/plausible_web/templates/layout/app.html.heex
sanne-san: Good idea 👍

RobertJoonas (approved):

assets/js/liveview/copy-snippet.js
```diff
@@ -0,0 +1,34 @@
+// Copy-to-clipboard behaviour for a readonly textarea holding a code snippet.
```
ukutaht, line 1: Since this is used on a LiveView, using a Liveview hook is preferable to Alpine.js. This should work: https://github.com/plausible/analytics/commit/f43c3c1970cc83d6561ddd77e77d4250895c662d

ukutaht: Using `navigator.clipboard` is also a bit nicer because it doesn't select and scroll the text, resulting in a calmer experience.

PR #6563, opened 2026-07-30
Revert "Add workaround for Clickhouse query condition cache bug"

apata:
### Changes

Reverts https://github.com/plausible/analytics/pull/6562


apata:
Not needed, stable.

PR #6564, opened 2026-07-31
Use Blacksmith checkout with sticky-disk git mirror in Elixir CI

cnkk:
`useblacksmith/checkout` is a drop-in fork of `actions/checkout` that keeps a git mirror on a Blacksmith sticky disk, so subsequent runs only fetch deltas instead of the full history each time.


apata (approved):
Can this make us run credo diff with stale master? Approving and let's see

PR #6573, opened 2026-08-04
Initialize currentEngagementURL to undefined, not location.href, refactor npm publish flow

apata:
Tracker package readme is supportive of SSR, but the package accesses browser-only `location` object on import, not on init / track. 

There isn't a reason to set this variable to location.href prior to `postPageviewTrack` being called, as we won't use it until engagements are registered, in which case, we will have set the URL from the payload.

This PR fixes it. Also, it
- updates tracker package publishing logic to follow https://docs.npmjs.com/trusted-publishers
- uses official playwright image to run tracker tests in

Picks up https://github.com/plausible/analytics/pull/6039, see those comments. 

TODO:
- [x] Configure NPM-side https://docs.npmjs.com/trusted-publishers#managing-trusted-publisher-configurations

### DIFF
--- a/.github/workflows/tracker-script-npm-release.yml
+++ b/.github/workflows/tracker-script-npm-release.yml
@@ -11,6 +11,7 @@ jobs:
     permissions:
       pull-requests: read
       contents: read
+      id-token: write
     if: >-
       ${{ github.event.pull_request.merged == true && (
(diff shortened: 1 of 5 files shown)


toofishes (approved):
Looks good to me! Thanks for tackling this.

ukutaht (approved):

PR #6574, opened 2026-08-05
Move tracking options into general site settings

sanne-san:
### Changes

- Outbound links, file downloads and form submissions can now be turned on and off from site settings, saving as you toggle them, instead of when reviewing installation
- Group the site domain and reporting timezone into a single Site details section, with the save button only active once the timezone changes
- Show the tracking status and a link to the setup instructions alongside the toggles, and list the measurements that need manual setup in a collapsible section
- Share the row, label, section and divider markup so other settings pages can pick up the same layout

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/generic.ex
```diff
@@ -707,6 +708,97 @@ defmodule PlausibleWeb.Components.Generic do
...
+    """
+  end
+
+  attr(:title, :string, required: true)
+  attr(:docs, :string, default: nil)
+  attr(:tooltip, :string, default: nil)
+  attr(:expandable?, :boolean, default: false)
+  slot(:inner_block, required: true)
+
+  def settings_section(%{expandable?: true} = assigns) do
```
RobertJoonas: This module is getting pretty big already. Could we extract all of these new settings-related components into a new module (e.g. `PlausibleWeb.Components.Settings`)? Could also accommodate the already existing `tile` and `settings_tiles`.

lib/plausible_web/templates/site/settings_general.html.heex
```diff
@@ -1,47 +1,49 @@
...
-    <:title>
-      Site domain
-    </:title>
-    <:subtitle>
-      Moving your site to a different domain? We got you!
-    </:subtitle>
-    <.input name="domain" label="Domain" value={@site.domain} disabled width="w-1/2" mt?={false} />
+  <.tile>
+    <:title>Site details</:title>
+    <:subtitle>Manage your site's domain and reporting timezone.</:subtitle>
```
RobertJoonas: The subtitle shouldn't mention domain change for consolidated views.

Image

Perhaps we could change the message into something more general that would not mention the fields that you can change at all? Thinking something like "Manage general site configuration"?

lib/plausible_web/live/tracking_settings.ex
```diff
@@ -0,0 +1,220 @@
...
+  def render(assigns) do
+    ~H"""
+    <div id="tracking-settings-main">
+      <.flash_messages flash={@flash} />
+
+      <.tile docs={if ce?(), do: "plausible-script"}>
+        <:title>Tracking</:title>
+        <:subtitle>Manage how Plausible collects data for this site.</:subtitle>
+
+        <div class="flex flex-col gap-6">
```
RobertJoonas: Seems like this extra div wrapper will always be needed between `<.tile>` and `<.settings_row>` / `<.settings_section>`. Maybe extract that into a `<.settings_rows>` component too? Similar logic as with settings_tiles -- just to avoid the risk of these classes getting out of sync.

lib/plausible_web/templates/site/settings_general.html.heex
```diff
@@ -1,47 +1,49 @@
...
-    <.button_link
-      href={
-        Routes.site_path(@conn, :installation, @site.domain, flow: PlausibleWeb.Flows.review())
-      }
-      mt?={false}
-    >
-      Review installation
-    </.button_link>
+        <.settings_divider />
+      </div>
```
RobertJoonas: We could avoid the `div` with `flex flex-col gap-6` classname here too by doing this:

```elixir
<%= if Plausible.Sites.regular?(@site) do %>
        <.settings_row label="Site domain" docs="change-domain-name">
          <span class="text-gray-800 dark:text-gray-200">{@site.domain}</span>
          <.styled_link
            href={Routes.site_path(@conn, :change_domain, @site.domain)}
            aria-label="Change domain"
            class="text-sm"
          >
            Change
          </.styled_link>
        </.settings_row>
        
        <.settings_divider />
      <% end %>
```

lib/plausible_web/templates/site/settings_general.html.heex
```diff
@@ -1,47 +1,49 @@
...
+            Save
+          </.button>
+        </.settings_row>
+      </.form>
+    </div>
   </.tile>
+
+  <div :if={Plausible.Sites.regular?(@site)}>
+    {live_render(@conn, PlausibleWeb.Live.TrackingSettings, session: %{"domain" => @site.domain})}
+  </div>
```
RobertJoonas: Same here. Could live_render directly without the redundant div wrapper:

```
<%= if Plausible.Sites.regular?(@site) do %>
    {live_render(@conn, PlausibleWeb.Live.TrackingSettings, session: %{"domain" => @site.domain})}
  <% end %>
```

lib/plausible_web/live/tracking_settings.ex
```diff
@@ -0,0 +1,220 @@
...
+              <.toggle_switch
+                id="outbound_links"
+                checked={@tracker_script_configuration.outbound_links}
+                phx-click="toggle"
+                phx-value-field="outbound_links"
+              />
+            </.measurement>
+            <.measurement label="File downloads">
+              <:icon><Heroicons.arrow_down_tray class="size-4.5" /></:icon>
+              <.toggle_switch
```
RobertJoonas, line 78: There's an existing issue with this component. In high-latency scenarios, the user doesn't get instant feedback after clicking the toggle in "server mode".

You can try it locally or even in production by typing `liveSocket.enableLatencySim(300)` in the browser console. In order for the component to render in the final state, a server round-trip is required anyway -- the component should never instantly/optimistically turn on/off without the server confirming it.

I think for any such component we should always consider an intermediate state where the component gets "greyed out" and loses the pointer cursor. LiveView automatically slaps a `phx-click-loading` attribute on the element that was clicked, so we can use that to add the "disabled" classes.

I experimented with it a bit, and this additional class on the `toggle_switch` button made it feel pretty okay:

```elixir
def toggle_switch(assigns) do
    server_mode? = not is_nil(assigns.checked)
    assigns = assign(assigns, :server_mode?, server_mode?)

    ~H"""
    <button
      id={"#{@id}-#{@id_suffix}"}
      class={[
        "h-6",
        if(@disabled, do: "cursor-not-allowed", else: "cursor-pointer"),
        if(@server_mode?, do: "phx-click-loading:cursor-default phx-click-loading:opacity-60"),
      ]}
      ...
      """
end
```

WDYT?

lib/plausible_web/live/tracking_settings.ex
```diff
@@ -0,0 +1,220 @@
...
+  defp assign_installation_status(socket) do
+    status =
+      cond do
+        ce?() -> nil
+        socket.assigns.site.onboarding_status in @completed_onboarding_statuses -> :completed
+        true -> :pending
+      end
+
+    assign(socket, installation_status: status)
+  end
```
RobertJoonas: Let's prevent CE from ever reading `site.onboarding_status` at all here during compile time.

```elixir
on_ee do
  defp installation_status(site) do
    if site.onboarding_status in @completed_onboarding_statuses do
      :completed
    else
      :pending
    end
  end
else
  defp installation_status(_), do: nil
end
```

...And then in mount call this instead:

```elixir
|> assign(:installation_status, installation_status(socket.assigns.site))
```

lib/plausible_web/live/tracking_settings.ex
```diff
@@ -0,0 +1,220 @@
...
+     )
+     |> assign_installation_status()}
+  end
+
+  def render(assigns) do
+    ~H"""
+    <div id="tracking-settings-main">
+      <.flash_messages flash={@flash} />
+
+      <.tile docs={if ce?(), do: "plausible-script"}>
```
RobertJoonas, line 40: The tracking section looks a bit weird on CE:

Image

Perhaps it could say `Installation instructions` on the left, and then the "Review" link wouldn't feel so... alone? 😄

lib/plausible_web/live/tracking_settings.ex
```diff
@@ -0,0 +1,220 @@
...
+      {render_slot(@inner_block)}
+    </div>
+    """
+  end
+
+  attr :href, :string, required: true
+
+  defp learn_more(assigns) do
+    ~H"""
+    <.tooltip centered?={true}>
```
RobertJoonas: I wonder if we could instruct the tooltip not to stay visible even when the mouse is moving into the tooltip itself. Quick video to show you what I mean:

https://github.com/user-attachments/assets/bf103546-1759-4413-b3aa-3ed2ddcdadf9

It's a bit annoying if you accidentally hover over the bottom link and actually want to click the one on the top. Maybe pass `interactive: false` for this specific case and make it disappear, even when the mouse moves directly onto the tooltip itself?

RobertJoonas (approved):
Nice work! 👌

PR #6582, opened 2026-08-10
Refactor auth layout

RobertJoonas:
### Changes

Refactor the auth inner layout to a component rather than a controller/router assigned `layout: {PlausibleWeb.LayoutView, :auth}`.

For reviewers -- the diffstat is that big because I've wrapped whole template files with `<Layouts.auth>`. Recommended: Review without whitespace diff highlighted.

### Tests

### Changelog

### Documentation

### Dark mode


RobertJoonas:
Closing it for the time being. Planning to go a bit deeper with the refactor to end up in a state where we can define `Layout.auth` as something like:

```elixir
def auth(assigns) do
  ~H"""
  <.app header?={false} footer?={false} global_notices?={false} >
    ...
  </.app>
  """
end
```

Meaning that we won't have to duplicate the logic of assigning `hide_header?` / `hide_footer?` / `disable_global_notices?`  between LiveViews and Controllers.

PR #6589, opened 2026-08-10
feat: CRM page listing trial prospects by MRR potential

ukutaht:
Surfaces the `trial_prospects` table in CRM.

### Changes

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible/customer_support/trial_prospects.ex
```diff
@@ -37,6 +43,82 @@ defmodule Plausible.CustomerSupport.TrialProspects do
...
+    from([team: t] in q, preload: [team: {t, owners: ^owners}])
+  end
+
+  defp order_prospects(q, "trial_start", direction) do
+    order_by(q, [team: t], [{^direction, t.inserted_at}])
+  end
+
+  # Over-the-top-tier (Custom/Enterprise) prospects rank first
+  defp order_prospects(q, "mrr", direction) do
+    order_by(q, [p], [{^direction, p.over_top_tier}, {^direction, p.estimated_mrr}])
```
apata, line 119:
```suggestion
    order_by(q, [p], [{^direction, p.estimated_mrr}])
```
Because PG ranks `null` value higher than integer, this simpler order by should give us the enterprise candidates top or bottom depending on sort direction.

apata: If we actually want them on top always, we need `order_by(q, [p], [{:desc, p.over_top_tier}, {^direction, p.estimated_mrr}])` according to my testing.

extra/lib/plausible_web/live/customer_support/trial_prospects.ex
```diff
@@ -0,0 +1,188 @@
...
+    {:noreply, socket}
+  end
+
+  def render(assigns) do
+    ~H"""
+    <Layout.layout show_search={false} flash={@flash}>
+      <div class="flex items-center justify-between">
+        <h2 class="text-xl font-bold sm:text-2xl">🔥 Trial prospects</h2>
+        <p class="text-sm text-gray-500 dark:text-gray-400">
+          {@total_entries} active trials — ranked by estimated MRR potential
```
apata:
```suggestion
          {@total_entries} prospects
```
I suggest to clarify this, because the query also covers recently expired trials. 
I removed ranking string because it may be misleading when table is sorted by something else. 

Alternatively, maybe something like 
`{@total_entries} prospects — ranked by {Phoenix.Naming.humanize(@sort_by)} {Phoenix.Naming.humanize(@sort_direction)}`?

apata (approved):

extra/lib/plausible_web/live/customer_support/trial_prospects.ex
ukutaht: Agree! The copy is not accurate as it stands

extra/lib/plausible/customer_support/trial_prospects.ex
ukutaht: > Because PG ranks null value higher than integer, this simpler order by should give us the enterprise candidates top or bottom depending on sort direction.

Functionally correct, but I feel like keeping `p.over_top_tier` is more expressive about the intent of the query. If we take it out the reader must know ordering of NULL vs integers off the top of their head to understand what the code is doing.

> If we actually want them on top always, we need order_by(q, [p], [{:desc, p.over_top_tier}, {^direction, p.estimated_mrr}]) according to my testing.

When sort order is flipped I think the enterprise ones should be at the bottom.

PR #6591, opened 2026-08-11
Reduce site deletions impact on ClickHouse 

aerosol:
### Changes

Remix of https://github.com/plausible/analytics/pull/6551

https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/10064243324

For reviewers: commit order matters, reading the messages alone should somewhat reflect the thought process. Migration will be extracted to a separate PR once review is concluded.

This PR changes how periodic site cleanup works:

- Pending deletions are stored for background execution as a part of the Postgres site deletion transaction
- Pending deletions are cleaned up once handled by the (now) weekly worker
- Pending deletions store imported/pageview ranges per site, so that data boundary is kept
- Site.Cache entry eviction is broadcasted upon Postgres site deletion so we stop accepting traffic ASAP. Edge case: it is still possible we'll keep some sessions_v2 around if they span over dates (sessions_v2 `timestamp` vs `start`). Classifying this as negligible for now.
- Lightweight deletions are issued where possible
- Dedicated ClickHouse connection is used for deletions (the same that we used to use for user-initiated import cancellations). Its connection pool was bumped from 1 to 2, just in case. This is to avoid extra ingest interference.
- Previously used orphaned site IDs listing is kept as a function, to be executed on prod once for the last time (I thought data migration would've been an overkill). The function effectively back fills pending deletions.
- Even though most of worker operation is asynchronous, basic worker metrics are exposed to the scraper. Again, considered polling ongoing mutations metrics, but there is no good way of monitoring/identifying the mutations only this part of app might've triggered, so leaving ClickHouse monitoring out of scope (for now at least).
- Pending deletions table is somewhat extensible - we list deletions by reason, so distinct ones could be implemented later on.
- For partitioned tables, deletions are enqueued partition by partition - we are operating under the assumption that the app isn't the best place to manage ClickHouse's internal queue.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/stats/clickhouse.ex
```diff
@@ -13,8 +13,10 @@ defmodule Plausible.Stats.Clickhouse do
   alias Plausible.Timezones
   alias Plausible.Stats
 
-  @spec pageview_start_date_local(Plausible.Site.t()) :: Date.t() | nil
-  def pageview_start_date_local(site) do
+  @spec pageview_start_date_local(Plausible.Site.t(), Keyword.t()) :: Date.t() | nil
+  def pageview_start_date_local(site, opts \\ []) do
```
apata: question: This function has a `where e.timestamp >= ^site.native_stats_start_at` condition. It gets persisted to the new deletions table as the stats start date for this site. However, the delete handler actually deletes all stats for the sites in the table using `min(timestamp)`. What's the reason for this difference?

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -31,51 +36,90 @@ defmodule Plausible.Workers.ClickhouseCleanSites do
...
-            timeout: :infinity
-          )
-        end,
-        timeout: :infinity
-      )
-
-    ch_sites = rows |> MapSet.new(fn [site_id] -> site_id end)
-
-    MapSet.difference(ch_sites, pg_sites) |> MapSet.to_list()
+  defp clear_table_via_mutation!(table, site_ids) do
```
apata, line 133: nitpick, non-blocking: The name may mislead the reader on the other functions in this file: LW deletes also end up in `system.mutations`. Maybe explicitly name it `clear_table_with_heavyweight_delete!`?

priv/repo/migrations/20260810060815_pending_stats_deletions.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Plausible.Repo.Migrations.PendingStatsDeletions do
+  use Ecto.Migration
+
+  def change do
+    create table(:pending_stats_deletions) do
+      # site id may not in postgres anymore, so not a reference
+      add :site_id, :integer, null: false
+      add :stats_start_date, :date, null: false
+      add :stats_end_date, :date, null: false
```
apata: question: What do we gain from storing start and end date values for the 0...7 days these records exist? The deletion script doesn't rely on them AFAICT. Is it for manual validation / statistics?

lib/plausible/stats/clickhouse.ex
```diff
@@ -30,7 +32,38 @@ defmodule Plausible.Stats.Clickhouse do
...
+        if use_site_timezone? do
+          Timezones.to_date_in_timezone(datetime, site.timezone)
+        else
+          NaiveDateTime.to_date(datetime)
+        end
+    end
+  end
+
+  @spec pageview_end_date_local(Plausible.Site.t(), Keyword.t()) :: Date.t() | nil
+  def pageview_end_date_local(site, opts \\ []) do
```
apata: nitpick, non-blocking: having _local in the names of these two functions is now a bit weird. Maybe `pageview_end_date` that always returns UTC and we'll convert to site TZ only where necessary? There's not too many call sites, ~10 for the two functions, so a refactor wouldn't yield a massive diff

lib/plausible/pending_stats_deletions.ex
```diff
@@ -0,0 +1,152 @@
...
+
+    %Ch.Result{columns: ["site_id"], rows: rows} =
+      DBConnection.run(
+        ch,
+        fn conn ->
+          Ch.query!(
+            conn,
+            """
+            SELECT site_id FROM events_v2 GROUP BY site_id
+            UNION DISTINCT
```
apata: nitpick, non-blocking: 
This won't catch the orphaned sites that have imports but no events or sessions (there's up to ~200 of these). To clear those as well, we'd have to scan all the imports tables as well. 

Because we put these into MapSet, we don't need to do UNION DISTINCT on CH, we could make a request for every table separately. In that case, every table's query gets retried independently, which looks slightly more durable to me.

apata: Also, regarding the query itself, I think it might be legacy. I've tested `SELECT DISTINCT site_id FROM #{tablename}` and works just as well.

lib/plausible/site/cache.ex
```diff
@@ -60,7 +60,8 @@ defmodule Plausible.Site.Cache do
 
   @impl true
   def get_from_source(domain) do
-    query = from s in base_db_query(), where: s.domain == ^domain
+    query =
+      from s in base_db_query(), where: s.domain == ^domain or s.domain_changed_from == ^domain
```
apata, line 64: question: I didn't see this change explained in commit messages. Is it safe to do?

lib/plausible/site/removal.ex
```diff
@@ -4,16 +4,19 @@ defmodule Plausible.Site.Removal do
...
-  @spec run(Plausible.Site.t()) :: {:ok, map()}
-  def run(site) do
+  @spec run(Plausible.Site.t(), Keyword.t()) :: {:ok, map()}
+  def run(site, opts \\ []) do
     Repo.transaction(fn ->
       site = Repo.preload(site, :team)
 
+      {:ok, pending_stats_deletion} = PendingStatsDeletions.store(site)
+
       result = Repo.delete_all(from(s in Plausible.Site, where: s.domain == ^site.domain))
```
ukutaht, line 20: Since `PendingStatsDeletions` is keyed on `site.id` rather than `site.domain`, it's also worth changing this line to delete via the same `site.id`.

There's a small chance with some domain change shenanigans that things will get weird otherwise.

lib/plausible/pending_stats_deletions.ex
```diff
@@ -0,0 +1,152 @@
...
+
+  alias Plausible.ClickhouseRepo
+  alias Plausible.PendingStatsDeletion
+  alias Plausible.Repo
+  alias Plausible.Site
+  alias Plausible.Sites
+
+  @spec store(Site.t(), atom()) :: {:ok, PendingStatsDeletion.t() | nil}
+  def store(%Site{} = site, reason \\ :user_request) do
+    case Sites.stats_range(site) do
```
ukutaht: This does not account for sites that have reset their stats. When stats are reset, the `site.native_stats_start_at` is bumped while leaving Clickhouse rows untouched. The result is that when we clean data for a site like this, the removal process will also leave those stats untouched.

To properly establish a date range to clean I think it needs to come from Clickhouse rather than relying on postgres pointers which are not guaranteed to match Clickhouse.

lib/plausible/pending_stats_deletions.ex
```diff
@@ -0,0 +1,152 @@
...
+      from(p in PendingStatsDeletion,
+        where: p.reason == ^reason,
+        select: %{
+          site_ids:
+            fragment(
+              "coalesce(array_agg(DISTINCT ? ORDER BY ?), ARRAY[]::integer[])",
+              p.site_id,
+              p.site_id
+            ),
+          stats_start: min(p.stats_start_date),
```
ukutaht: This value propagates to `Clickhouse.partition_ids` to choose the range of partitions to touch. This value also contains imported data whose range is essentially user input and can go very far back.

So if a batch contains imports from 2005, we issue 264 mutations over 132 partitions for native tables that for most of that range have no data at all. They will essentially be no-ops but not free, Clickhouse will still process them as mutations. Better to avoid.

ukutaht: Essentially it's another point towards not using postgres pointers for deriving partition ranges. The postgres pointer includes imported data and forgets about resets so it's not a reliably indicator for what needs to be deleted.

lib/plausible/site/cache.ex
ukutaht: It does not appear to be safe. As with event replay there's a weird edge case if there are two different sites with the same `domain` and `domain_changed_from` values. There is no uniqueness guarantee between them. In a scenario like this, `Gatekeeper.check/2` will raise `Ecto.MultipleResultsError`. We have established this scenario is very unlikely but just wanted to put this out there.

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -31,51 +36,90 @@ defmodule Plausible.Workers.ClickhouseCleanSites do
...
+
+        :ok
     end
+  end
+
+  defp measure_stage(stage, fun) do
+    PlausibleMetrics.measure_duration(telemetry_stage_duration(), fun, %{stage: stage})
+  end
+
+  defp clear_partitioned_table!(table, partition_id, site_ids) do
```
ukutaht, line 117: Perhaps I don't understand the characteristics between Heavy vs Lightweight deletes enough but this is surprising: I thought the benefit of lightweight deletes was that we don't need app-side partitioning because these are lighter. And @cnkk can apply the masks at in a partitioned fashion at a convenient time.

If we do need to orchestrate partitioning app-side anyways, is there any benefit to the two-step process of lightweight deletes?

lib/workers/clickhouse_clean_sites.ex
aerosol: ClickHouse docs refers to it as Delete Mutation, not Heavyweight Delete.

priv/repo/migrations/20260810060815_pending_stats_deletions.exs
aerosol: We don't gain anything, it's convenient to use the date column type, since we aggregate min/max over dates.

lib/plausible/pending_stats_deletions.ex
aerosol: Good call about imports, I'll add that.

I don't understand what do you mean by "every query gets retried". Retried where and how? 

We use MapSet because for some reason it was used before. You could probably do `list1 -- list2` with as much success. Note that this function is temporary. 

> Also, regarding the query itself, I think it might be legacy. I've tested SELECT DISTINCT site_id FROM #{tablename} and works just as well.

What is legacy specifically and how do you recognize it's legacy? SELECT DISTINCT seems to be using more memory, I don't know if that's accidental, but again, I don't think we need to spend time on that since it's a one-off thing that this PR aims to replace.

lib/plausible/site/cache.ex
aerosol: Site.Cache lets you `get` each site by two keys (domain and old domain). This fallback is used during tests and lets you do the same now, though it's never used. I accidentally stumbled upon it, and it's safe (and correct) thing to do.

lib/plausible/stats/clickhouse.ex
aerosol: Sure thing, will do

lib/plausible/pending_stats_deletions.ex
```diff
@@ -0,0 +1,152 @@
...
+  def store(%Site{} = site, reason \\ :user_request) do
+    case Sites.stats_range(site) do
+      %{stats_start_date: nil, stats_end_date: nil} ->
+        {:ok, nil}
+
+      %{stats_start_date: stats_start_date, stats_end_date: stats_end_date} ->
+        Repo.insert(%PendingStatsDeletion{
+          site_id: site.id,
+          stats_start_date: stats_start_date,
+          stats_end_date: stats_end_date,
```
ukutaht: The end date gets fixed when `PendingStatsDeletion` record is created but `Site.Cache` eviction is best effort and async

There is a potential timing bug here where `PendingStatsDeletion` is run close to a month boundary so some remaining events get added to the partition after `stats_end_date`. It would result in some (albeit small amount) of orphaned events.

ukutaht (changes requested):
Inline comments aside, I have some reservations about the overall direction in this PR.

It introduces a worklist (`PendingSiteDeletions`) over calculating the set of site ids to be deleted by comparing ids in postgres vs clickhouse as was done previously. With this design I'm seeing new failure modes and edge cases that did not exist before. There's a potential timing issue with cache eviction. It also creates a synchronization point where site deletions must always add a deletion entry as well. Baking date ranges from postgres into it causes pretty significant errors.

The benefit of the previous design was that however sites were deleted, the `ClickhouseCleanSites` worker would always move towards a stable state where the set of site ids present in clickhouse is equal to the ones in postgres. The new design does not have that property and introduces the possibility (or even likelihood) that the two sets will diverge over time.

The way I see it, the proposed design loses some great properties we used to get for free. The previous contract of "when site is deleted from postgres, it will eventually be sweeped from clickhouse" was better decoupled from the rest of the application, simple to reason about and hard to get wrong.

What's the reasoning for adding the `PendingSiteDeletions` table? What problem is it solving that the previous design didn't?

The core issue with the old design was execution strategy that ended in big and heavy mutations. My current view is that it would be better to keep the old design for deciding which sites to delete, and only touch the execution part of it.

### Notes about execution strategy

~~This PR adds partitioning **and** lightweight deletes. Running a local test for how much space the mutations claim, the result is the following:~~
```
One table, 3 monthly partitions, 9M rows, 74.7 MiB on disk. Deleting one site out of 5,000 — 0.02% of rows:

┌─────────────────────────────────────────┬───────────────────┬───────────────────────────────────┐
│                strategy                 │ extra disk needed │           as % of table           │
├─────────────────────────────────────────┼───────────────────┼───────────────────────────────────┤
│ ALTER … DELETE WHERE (current)          │ 74.8 MiB          │ 100.2% — whole table duplicated   │
├─────────────────────────────────────────┼───────────────────┼───────────────────────────────────┤
│ ALTER … DELETE IN PARTITION ID '202502' │ 25.0 MiB          │ 33.4% — exactly 1 of 3 partitions │
├─────────────────────────────────────────┼───────────────────┼───────────────────────────────────┤
│ DELETE FROM (lightweight)               │ 73.5 KiB          │ 0.096%                            │
└─────────────────────────────────────────┴───────────────────┴───────────────────────────────────┘
```

~~This shows the problem with the current strategy - the whole table needs to be copied to run the mutation. On the other hand, lightweight deletes use negligible disk space. It seems like we should be able to issue a whole-table lightweight delete with no worries about disk space.~~

~~Running heavy mutations in a partitioned manner is also an option since it does reduce disk space needed meaningfully and would fix the prod issue.~~

~~But perhaps the simplest thing to try is to switch the current worker to use lightweight deletes without changing anything else and seeing how it works. Given that it comes with a risk of breaking ingestion it's probably worth verifying in some capacity on prod as well before enabling.~~

@apata got a different result [here](https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/10064243324#__recording_10134758521). Not sure what the truth is on lightweight vs heavy deletes and the need to partition them.

aerosol:
Some good points, especially the stats dates quirk that you both noticed. 
I failed to see the "reset stats" scenario and the fact that we should not estimate partition range based on imported data. 

With regards to partitioning lightweight deletes, my understanding was it's always helpful to chunk work, especially given that ClickHouse _was_ running mutations (on **2025**07) for a partitioned lightweight delete in a test @apata conducted:

image

I also started thinking of ClickHouse queue as of configurable self-managing capacity rather than limitation.

Anyway:

> What's the reasoning for adding the PendingSiteDeletions table? What problem is it solving that the previous design didn't?

The idea was exactly to track sites that need to be deleted, instead of running quite expensive queries that may get even more expensive over the coming years.
In my mind this had the extra benefit of not doing any work at all, if there's nothing to delete. 

Moreover I thought of reusing that table for handling expired trials/abandoned accounts ahead of time (different `reason`). Querying it on a weekly basis could be helpful for monitoring/reporting insight.

> But perhaps the simplest thing to try is to switch the current worker to use lightweight deletes without changing anything else and seeing how it works. 

I like it, we can try that, but:

> The benefit of the previous design was that however sites were deleted, the ClickhouseCleanSites worker would always move towards a stable state where the set of site ids present in clickhouse is equal to the ones in postgres. 

It sounds nice on paper but the previous design wasn't querying `sessions_v2`, just `events_v2` to determine the set of site IDs, which was never stable state - there were some dangling entries left in the former table. And per @apata's [finding](https://github.com/plausible/analytics/pull/6591#discussion_r3758581272), imported tables also contain IDs that don't exists anywhere else. 

The heavy querying I mentioned becomes heavier if we want to address that AND keep the old way of determining the IDs that need to be removed. 

What direction would you like see here?

> Given that it comes with a risk of breaking ingestion it's probably worth verifying in some capacity on prod as well before enabling.

What do you mean by "risk of breaking ingestion" here? The fact that current clean up worker uses `IngestRepo` pool or are you suggesting lightweight deletes alone will affect ingestion somehow?

lib/plausible/site/cache.ex
aerosol: Filed https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/10194198173

lib/plausible/pending_stats_deletions.ex
apata: Agreed that it's not worth spending time on! By legacy I mean I didn't find any mention of it being optimal in CH, but I did find a mention of it being preferred / optimal in PG, thought it was from those days. Thanks for checking the memory, interesting find, so it potentially is more optimal still :D

lib/workers/clickhouse_clean_sites.ex
```diff
@@ -31,51 +36,90 @@ defmodule Plausible.Workers.ClickhouseCleanSites do
...
+    DeletionRepo.query!(
+      "DELETE FROM {$0:Identifier} IN PARTITION ID {$1:String} WHERE site_id IN {$2:Array(UInt64)}",
+      [table, partition_id, site_ids],
+      settings: @settings
```
cnkk, line 121: we need `lightweight_deletes_sync = 0` if we want it run async or at least `= 1` to wait for the current node.

lib/plausible/pending_stats_deletions.ex
cnkk: I think `partition_ids` could be simplified by querying Clickhouse for the partitions
```sql
SELECT DISTINCT _partition_id FROM events_v2 WHERE site_id IN (…)
```

ukutaht:
@aerosol Thanks. Getting to the heart of the discussion on the `PendingStatsDeletions` design

> The idea was exactly to track sites that need to be deleted, instead of running quite expensive queries that may get even more expensive over the coming years.

Right. I hadn't really considered this as an issue because we haven't hit any limits on this as far as I can tell. Artur said the clickhouse [query](https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/10064243324#__recording_10135088039) currently takes 20 seconds. In the medium term I would expect queries like this to get less expensive. As we start clearing inactive customers from the database, the dataset to query each time will be smaller since we have a long history of inactive customers still in both databases. But I see the long-term concern.

On one hand I agree - a 20 second query on clickhouse side feels a bit uncomfortable. Same for reading all site ids (six figure number) from postgres to app memory. On the other hand I don't think these have hit any limits yet so I wonder if it's a 'yagni' type of thing. I'm conflicted about whether the site id querying in the old design is a problem or not without knowing where the actual limits are.

I can definitely see the benefit in the worklist design in that the query heaviness and large `site_id` sets being shuffled around is not something we need to worry about in the future. And I agree with the point about other tables - the old design was leaving orphaned data in tables other than `events_v2` whereas the `PendingStatsDeletion` list is a single signal to sweep all tables which is definitely a nice property.

Expanding the old design for all tables comes with its own set of issues and I'm happy to move forward with the `PendingStatsDeletion` approach proposed here as long as the questions about stats boundaries and partition calculation are resolved.

> I failed to see the "reset stats" scenario and the fact that we should not estimate partition range based on imported data.

I'll add more on these.

> With regards to partitioning lightweight deletes, my understanding was it's always helpful to chunk work, especially given that ClickHouse _was_ running mutations (on **2025**07) for a partitioned lightweight delete in a test [@apata](https://github.com/apata) (https://github.com/apata) conducted:

Let's also discuss this inline

lib/plausible/pending_stats_deletions.ex
ukutaht: Further explanation: assume a site whose clickhouse native data starts from `2026-01-01` and the `native_stats_start_date` is also at `2026-01-01`. No imports. Now take the following steps:
1. Reset stats today (`Plausible.Purge.reset!/1` bumps `Site.native_stats_start_date` to `2026-08-12`)
2. Delete the site
3. Expected result: `PendingStatsDeletion` is inserted with `stats_start_date: 2026-01-01`
4. Actual result: depends on whether any events were ingested after stats were reset
  4a. If no events received: no `PendingStatsDeletion` inserted at all because `Plausible.Sites.stats_range/1` returns `{nil, nil}` 
  4b. If some events received: `PendingStatsDeletion` is inserted with `stats_start_date` being the date of the first event received after `2026-08-12`

Fundamentally the issue is that `Plausible.Stats.Clickhouse.pageview_start_date_local/2` uses `site.native_stats_start_date` as a filter so the start date is clamped to the postgres field. But this postgres field is not a faithful representation of when clickhouse data started. It can be moved by the reset operation `Plausible.Purge.reset/1` and it can also be overwritten in the CRM. There is no postgres field that tracks the actual start date of native clickhouse stats.

lib/plausible/site/cache.ex
ukutaht: Makes sense @aerosol - I missed that it's a test-only change.

The actual production bug, as established, is very unlikely and only limited to sites within a single team (I think?). Doesn't seem priority to me.

lib/plausible/pending_stats_deletions.ex
ukutaht: @aerosol Explaining further: 

Assume a site with native data starting on 2026-01-01 and an import with `start_date` on 2005-01-01.
When this site is deleted, a `PendingStatsDeletion` entry is added with `stats_start_date=2005-01-01`

When `ClickhouseCleanSites` requests a batch of pending deletions with this entry included, the `stats_start_date` for the whole batch becomes `2005-01-01`. This is passed as the start date to `Clickhouse.partition_ids` which generates partition IDs all the way back to 2005, most of which do not exist in our partitioned native tables. Result is that mutations are issued for non-existent partitions in `events_v2` and `sessions_v2` tables.

I'm not sure what the cost of issuing no-op mutations is at the database level. Just flagging that a single site with a long import history would affect the whole batch like this.

> SELECT DISTINCT _partition_id FROM events_v2 WHERE site_id IN (…)

This feels a lot better. Perhaps it would also enable removing `stats_start_date` and `stats_end_date` from `PendingStatsDeletion` entry altogether and also fix the reset stats issue with start date calculation.

aerosol:
@ukutaht @cnkk @apata pushed https://github.com/plausible/analytics/pull/6591/commits/431e6c886931c4a6a240f0ffe0e681f5f435b645 and got rid of wobbly date markers dependency. This is still under assumption we should chunk lightweight deletes by partitions, and it is now using the method suggested by Cenk. I'm not sure about `lightweight_deletes_sync = 0`, it seems as if lightweight delete Artur tried before ran synchronously with the current setting. I'm puzzled.

lib/workers/clickhouse_clean_sites.ex
aerosol: My understanding is without `IN PARTITION ID`, a mutation (lightweight or not) still gets scheduled against every existing part across the entire table's history. The db can skip non-matching parts via primary key, but it still has to open and check each. Idk, but keeping it partitioned doesn't seem harmful - mutations are kept small and independently executable. Perhaps that's better for ClickHouse's internal queue and monitoring as opposed to one big hog? I am totally guessing though.

ukutaht (approved):

PR #6592, opened 2026-08-11
fixes to the enterprise_over_limit internal alert

metmarkosaric:
it's been a while since we touched this alert so I've taken the liberty to get some ai help to bring it up to date with teams, CRM etc. main changes:

- do not send multiple alerts when the same team goes over the limit (right now it sends one email per team owner)
- do not send the alert when someone hits the site limit, only when someone goes above the site limit (right now it sends the alerts and starts grace period when someone hits the limit)
- show what's over the limit and add a link to the team to make it easier to handle the case

### DIFF
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -150,10 +150,10 @@ defmodule Plausible.Teams.Billing do
 
   defp revise_pageview_usage(team, usage_mod) do
     case Plausible.Workers.CheckUsage.check_pageview_usage_two_cycles(team, usage_mod) do
-      {:over_limit, _} ->
+      {:over_limit, _, _} ->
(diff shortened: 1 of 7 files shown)


RobertJoonas:
@metmarkosaric I've adjusted the code and the email template a bit. Everything that you mentioned in the PR description still works. Here's how the email now looks like:

image

metmarkosaric:
That’s great, thank you @RobertJoonas! Let me know if I need to do anything else

RobertJoonas:
> That’s great, thank you @RobertJoonas! Let me know if I need to do anything else

From our side it's ready then. Would just like to get it reviewed by another developer before merging.

ukutaht (approved):

PR #6595, opened 2026-08-12
Fix workers dealing with enterprise plans

RobertJoonas:
### Changes

Two commits, two independent bug fixes:

1. `LockSites` unlocking manually locked enterprise teams

Currently, the `LockSites` worker automatically unlocks sites that have reduced their pageview usage and fallen back within their plan limits. But there's no exception for enterprise plans, where sites limit should matter too. Currently, if an enterprise team gets manually locked (due to not upgrading to a new enterprise plan we've created for them), the SiteLocker undoes that the next midnight, on the condition that their pageviews are OK. Fix: check the sites usage too.

2. `CheckUsage.check_site_usage_for_enterprise` not respecting grandfathered "unlimited" sites

Any accounts created before `2021-05-05` have unlimited sites, but `CheckUsage` does not account for that. I don't think it has ever happened before, but currently we'd get an internal notification saying "this team has outgrown their enterprise plan... sites: 110 / 100", even though they'd see that they have unlimited sites in their team/account settings. Fix: use the global `Plausible.Teams.Billing.site_limit` function to determine the site limit for the team.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -5,11 +5,11 @@ defmodule Plausible.Teams.Billing do
 
   import Ecto.Query
 
-  alias Plausible.Billing.EnterprisePlan
   alias Plausible.Billing.Subscription
   alias Plausible.Billing.Subscriptions
   alias Plausible.Repo
   alias Plausible.Teams
+  alias Plausible.Workers.CheckUsage
 
   alias Plausible.Billing.{EnterprisePlan, Feature, Plan, Plans, Quota}
   alias Plausible.Billing.Feature.{Goals, Props, SitesAPI, StatsAPI, SharedLinks, SSO}
@@ -141,23 +141,29 @@ defmodule Plausible.Teams.Billing do
         {:needs_to_upgrade, :no_active_trial_or_subscription}
 
       Teams.GracePeriod.expired?(team) ->
-        revise_pageview_usage(team, usage_mod)
+        revise_usage(team, usage_mod)
 
       true ->
         :no_upgrade_needed
     end
   end
 
-  defp revise_pageview_usage(team, usage_mod) do
-    case Plausible.Workers.CheckUsage.check_pageview_usage_two_cycles(team, usage_mod) do
-      {:over_limit, _, _} ->
-        {:needs_to_upgrade, :grace_period_ended}
-
-      {:below_limit, _, _} ->
-        Plausible.Teams.remove_grace_period(team)
-        :no_upgrade_needed
+  defp revise_usage(team, usage_mod) do
+    with {:below_limit, _, _} <- CheckUsage.check_pageview_usage_two_cycles(team, usage_mod),
+         :below_limit <- revise_site_usage(team) do
+      Plausible.Teams.remove_grace_period(team)
+      :no_upgrade_needed
+    else
+      _ -> {:needs_to_upgrade, :grace_period_ended}
     end
   end
+
+  defp revise_site_usage(%Teams.Team{enterprise_plan: %EnterprisePlan{}} = team) do
+    {site_usage_status, _, _} = CheckUsage.check_site_usage_for_enterprise(team)
+    site_usage_status
+  end
+
+  defp revise_site_usage(_), do: :below_limit
 
   @doc """
   Enterprise plans are always allowed to add more sites (even when
--- a/lib/plausible/teams/billing.ex
+++ b/lib/plausible/teams/billing.ex
@@ -141,16 +141,16 @@ defmodule Plausible.Teams.Billing do
         {:needs_to_upgrade, :no_active_trial_or_subscription}
 
       Teams.GracePeriod.expired?(team) ->
-        revise_usage(team, usage_mod)
+        revise_usage_for_grace_period_ended(team, usage_mod)
 
       true ->
         :no_upgrade_needed
     end
   end
 
-  defp revise_usage(team, usage_mod) do
+  defp revise_usage_for_grace_period_ended(team, usage_mod) do
     with {:below_limit, _, _} <- CheckUsage.check_pageview_usage_two_cycles(team, usage_mod),
-         :below_limit <- revise_site_usage(team) do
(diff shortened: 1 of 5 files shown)


lib/plausible/teams/billing.ex
```diff
@@ -141,24 +141,30 @@ defmodule Plausible.Teams.Billing do
...
     end
   end
 
+  defp revise_site_usage(%Teams.Team{enterprise_plan: %EnterprisePlan{}} = team) do
+    {site_usage_status, _, _} = CheckUsage.check_site_usage_for_enterprise(team)
+    site_usage_status
+  end
+
+  defp revise_site_usage(_), do: :below_limit
+
```
apata: suggestion, non-blocking: If not an enterprise plan, `:below_limit` does not seem accurate. With those plans, we don't actually care about site usage in this context, so we don't check it. Maybe emit `:not_applicable` here or make this branch conditional some other way?

lib/plausible/teams/billing.ex
```diff
@@ -141,24 +141,30 @@ defmodule Plausible.Teams.Billing do
...
+    with {:below_limit, _, _} <- CheckUsage.check_pageview_usage_two_cycles(team, usage_mod),
+         :below_limit <- revise_site_usage(team) do
+      Plausible.Teams.remove_grace_period(team)
+      :no_upgrade_needed
+    else
+      _ -> {:needs_to_upgrade, :grace_period_ended}
     end
   end
 
+  defp revise_site_usage(%Teams.Team{enterprise_plan: %EnterprisePlan{}} = team) do
```
apata: nitpick, non-blocking: The function name `revise_...` implies some other action besides checking (e.g. changing grace period).

lib/workers/check_usage.ex
```diff
@@ -88,8 +88,8 @@ defmodule Plausible.Workers.CheckUsage do
     :ok
   end
 
-  defp check_site_usage_for_enterprise(subscriber) do
-    limit = subscriber.enterprise_plan.site_limit
+  def check_site_usage_for_enterprise(subscriber) do
+    limit = Teams.Billing.site_limit(subscriber)
```
apata, line 92: issue, blocking: I think we should trust the value in `enterprise_plan`. As I see it, these plans are rare and each one has been handcrafted and negotiated: `team.inserted_at` shouldn't figure in these. Otherwise, an old user/team (from before 2021) converting from any legacy plan to a new enterprise plan, would automatically be considered to have "unlimited sites" despite the site limit selected for the plan in the backoffice.

RobertJoonas: Yeah, that's the intent, their accounts are entitled to unlimited sites forever. I don't think we can ever change that. In their site settings they also see "Sites usage: x / unlimited sites".

lib/plausible/teams/billing.ex
RobertJoonas: I just read the function name again, and the intent is not very obvious indeed. Also a bit bothered by returning `:grace_period_ended` as a reason, without the function name having anything to do with grace periods. How about `check_ended_grace_period_again`? And then `check_ended_grace_period_site_usage` / `check_ended_grace_period_pageview_usage`. A bit long, but accurate at least.

lib/plausible/teams/billing.ex
RobertJoonas: It's technically accurate to say `:below_limit` because regular (non-enterprise) subscriptions are never allowed to go **over** the site limit. Not a fan of introducing another specific reason such as `:not_applicable`, as that would leave the with branch as something like:

```elixir
with site_usage when site_usage in [:below_limit, :not_applicable]
  <- check_ended_grace_period_site_usage(team) do
  ...
end
```

...and make it feel more complicated than it actually is. How about returning just `:ok` instead of `:below_limit`?

```elixir
with :ok <- check_ended_grace_period_site_usage(team) do
  ...
end
```

lib/workers/check_usage.ex
apata: Right, I see your point! I didn't consider what's shown to the user. 

I'm testing setting up a custom plan for a team from 2021-05-01. The plan creation form in CS allows specifying sites. 

There's no indication to the admin that this value will be ignored. 

When upgrading to the plan in the app, the value that the admin entered is presented to the user. (from `def present_enterprise_plan(assigns)`)

After upgrading to the plan, the billing page to the user says sites are unlimited.

I guess my problem is that this change doesn't fully solve the site limit discrepancy between the server, superadmin and user. It changes the server view to match one of the user views. Because the server's current implementation is more defensive, I'd lean towards that.

lib/plausible/teams/billing.ex
apata: I thought something like this 
```elixir
  defp revise_usage(team, usage_mod) do
    with {:below_limit, _, _} <- CheckUsage.check_pageview_usage_two_cycles(team, usage_mod),
         :below_limit <- maybe_check_site_usage(team) do
      Plausible.Teams.remove_grace_period(team)
      :no_upgrade_needed
    #...
  end
    
  defp maybe_check_site_usage(team_with_enterprise_plan), do: #...
  # for normal plans, site limit can't be exceeded, presumed to be within spec
  defp maybe_check_site_usage(_team_with_normal_plan), do: :below_limit
```

lib/plausible/teams/billing.ex
apata: Answered here: https://github.com/plausible/analytics/pull/6595#discussion_r3797369401

lib/workers/check_usage.ex
apata: Discussed offline, this is not blocking any more

apata (approved):

lib/plausible/teams/billing.ex
RobertJoonas: https://github.com/plausible/analytics/pull/6595/commits/58ff79f430f58cfb77caca5a85cc9a664ac4474a

PR #6599, opened 2026-08-14
App layout refactor: Phase 2

RobertJoonas:
### Changes

See [phase 1 PR](https://github.com/plausible/analytics/pull/6598) for full context.

* Add a new `Layouts.app` component: typed attrs (`header?`, `footer?`, `global_notices?`,
  `trial_badge?`, `embedded?`, `load_dashboard_js?`), all defaulting to today's
  shown-by-default behavior. `Layouts.legacy` becomes a thin adapter translating the
  old `hide_header?`/`hide_footer?`/`disable_global_notices?`/`hide_trial_badge?`
  assigns into the new attrs — no existing call sites change.
- `_header`/`_footer`/`_notice`/`_flash` partials converted to real function
  components (`components/layouts/*.html.heex`, via `embed_templates`), dropping
  `@conn` in favor of plain assigns and `PlausibleWeb.Endpoint` for route helpers.
  Renamed the billing-notices partial to `team_notices` to avoid colliding with
  `Generic.notice/1`.
- `LayoutView.home_dest/1` now takes `current_user` directly instead of `conn`.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/layouts/footer.html.heex
```diff
@@ -6,8 +6,8 @@
           <img
             src={
               PlausibleWeb.Router.Helpers.static_path(
-                @conn,
-                logo_path("logo_dark.svg")
+                PlausibleWeb.Endpoint,
```
aerosol, line 9: is there a specific difference between using `Endpoint` and `@conn` here?

lib/plausible_web/components/layouts.ex
```diff
@@ -1,44 +1,95 @@
...
+          type="text/javascript"
+          src={Routes.static_path(PlausibleWeb.Endpoint, "/js/dashboard.js")}
+        >
         </script>
       <% end %>
     </div>
     """
   end
+
+  def legacy(assigns) do
```
aerosol, line 89: For this component and others, shall we enumerate the attrs+defaults for each for better compile-time hygiene?

aerosol (approved):

lib/plausible_web/components/layouts/footer.html.heex
RobertJoonas: AFAIK it doesn't make a difference here, nor anywhere else in this PR where I've moved from `@conn` -> `PlausibleWeb.Endpoint`. It's just to get rid of the conn dependency in header/footer/etc.

lib/plausible_web/components/layouts.ex
RobertJoonas: > For this component

Done in https://github.com/plausible/analytics/pull/6599/commits/886d8643c0d4e1497ce35b3c581ae095d276cd77, where I've also removed some dead code (`:inner_layout`).

> and others

By others I suppose you mean header/footer/etc? In order to define attrs like that we'd have to write them into actual function defs (rather than `.heex` templates). I'm definitely in favour of doing that. Aside from providing compile time validation for mismatching attributes and their types, it would also make it possible to "jump to definition" quickly, instead of searching and opening the file. I'd tackle that in a follow-up after step three though.

aerosol: Sounds great!

PR #6600, opened 2026-08-14
App layout refactor: Phase 3

RobertJoonas:
### Changes

See [phase 1 PR](https://github.com/plausible/analytics/pull/6598) for full context.

Review recommendations:
1) Review commit by commit -- I've split each template migration into a separate commit.
2) Hide whitespace -- whole templates wrapped in `Layouts.auth`

* Add a `legacy_layout?` assign (default `true`) on the shared root layout. When
  `false`, the root renders a plain `@inner_content` instead of going through
  `Layouts.legacy` — the mechanism for a page to fully own its own layout.
- New `Layouts.auth` wraps its content in
  `<.app header?={false} footer?={false} global_notices?={false}>` directly,
  providing the centered logo/heading/subtitle that the new auth pages need.
- Removed `AuthLayoutContext` and `ControllerHelpers.render_auth_page` — call sites
  just pass `legacy_layout?: false` alongside their existing assigns.
- All auth templates wrapped in `<Layouts.auth>`; deleted the old
  `templates/layout/auth.html.heex` secondary layout.

### Tests

### Changelog

### Documentation

### Dark mode


extra/lib/plausible_web/controllers/sso_controller.ex
```diff
@@ -24,9 +24,9 @@ defmodule PlausibleWeb.SSOController do
         redirect(conn, to: Routes.auth_path(conn, :login_form, return_to: params["return_to"]))
 
       _ ->
-        render_auth_page(conn, "login_form.html",
-          autosubmit: params["autosubmit"] != nil,
-          heading: "Sign in with SSO"
```
aerosol, line 29: So sometimes we don't pass the `heading` attr (auth sso views) and sometimes we do (password reset errors) at the controller level, even though they're all handled within `Layout.auth`. Is the idea to only pass it when it's dynamic? 🤔

aerosol (approved):

PR #6603, opened 2026-08-18
Improve funnel layout

sanne-san:
### Changes

- Rebuild the funnel report as a React component with CSS bars, replacing the old Chart.js canvas and its hover-only tooltip
- Give every step its own card and bar, labelled with the visitors who continued and the visitors who dropped off
- Align the funnel header with the Explore header.
- Add `strict` or `flexible` to the funnel header, indicating whether the funnel must be followed in order or not.
- Scroll longer funnels horizontally, four steps at a time
- Show step placeholders while loading, and one short message when a funnel cannot be shown
- Drop the chartjs-plugin-datalabels dependency

### Tests

### Changelog

### Documentation

### Dark mode


assets/package-lock.json
```diff
@@ -21,7 +21,6 @@
         "abortcontroller-polyfill": "^1.7.3",
         "alpinejs": "^3.13.1",
         "chart.js": "^3.3.2",
```
apata: Looks like we can drop this dependency as well!

extra/lib/plausible/stats/funnel.ex
```diff
@@ -56,6 +56,7 @@ defmodule Plausible.Stats.Funnel do
     {:ok,
      %{
        name: funnel.name,
+       strict_order: funnel.strict_order,
```
apata, line 59: issue, blocking: This change doesn't seem to be covered with Elixir tests.

assets/js/dashboard/util/number-formatter.ts
```diff
@@ -82,3 +82,7 @@ export function percentageFormatter(number: number | null): string {
     return '-'
   }
 }
+
+export function rateFormatter(rate: string): string {
+  return percentageFormatter(Number(rate))
+}
```
apata, line 88: suggestion, non-blocking: If these are internal endpoints, we should be able to refactor them to return CR percentage as numbers, so we'd not need this utility

assets/js/dashboard/extra/funnel/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+            className="absolute inset-x-0 flex justify-center translate-y-1/2 transition-[bottom] duration-500 ease-out"
+            style={{
+              bottom: `max(${fill}%, ${LABEL_HALF_HEIGHT_PX + LABEL_FLOOR_GAP_PX}px)`
+            }}
+          >
+            {step !== null && (
+              <div className="flex flex-col gap-1 max-w-full px-2.5 py-1.5 rounded-md text-xs font-medium whitespace-nowrap border border-gray-200 dark:border-gray-750 bg-white dark:bg-gray-825 shadow-sm opacity-100 transition-opacity duration-300 starting:opacity-0">
+                <StepOutcomeRow outcome={step.continued} dropoff={false} />
+                <StepOutcomeRow outcome={step.droppedOff} dropoff={true} />
+              </div>
```
apata: question: This PR replaces on-column box data with another kind of data. The currently boxed data moves to headers. Are we sure the data in headers is prominent enough? 

Image

apata: Wait, I think I understand what's confusing  me actually: it's that the info rendered in the box on the column doesn't add up with the column height itself. It's positioned as the label for the column, but it's not the label, it's extra info about this step. 

Maybe we can  position it in a separate textbox below each column, now that the step name is at the top? With a text comment, e.g. "never entered this funnel: ..."
"dropoff: ..."

assets/js/dashboard/extra/funnel/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+      try {
+        setFunnel(await fetchFunnel())
+        setError(undefined)
+      } catch (error) {
+        setError(error as Error)
+      } finally {
+        setLoading(false)
+      }
+    })()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
apata: suggestion, non-blocking: Since we're rewriting, we can replace the hand-rolled fetch and data/error/loading states with useQuery hook from react-query.

aerosol:
Not even a suggestion, just something that made me pause.

Question: should these

image

correspond somehow visually with:

image

Maybe color-wise?

Also, just double checking if we want to present detailed number in two places:

image

As opposed to saying "3K" somewhere for example.

assets/js/dashboard/util/number-formatter.ts
apata: There's a FE rounding issue when the values are close to each other. 
Screenshot 2026-08-24 at 10 22 10

Data from BE is
```
{
    "name": "Upgrade",
    "steps": [
        {
            "label": "Trial notification upgrade click",
            "visitors": 1000,
            "conversion_rate": "100",
            "conversion_rate_step": "0",
            "dropoff": 0,
            "dropoff_percentage": "0"
        },
        {
            "label": "Purchase Subscription",
            "visitors": 140,
            "conversion_rate": "14",
            "conversion_rate_step": "14",
            "dropoff": 860,
            "dropoff_percentage": "86"
        }
    ],
    "strict_order": true,
    "all_visitors": 8740700,
    "entering_visitors": 1000,
    "entering_visitors_percentage": "0.01",
    "never_entering_visitors": 8739700,
    "never_entering_visitors_percentage": "99.99"
}
```

sanne-san: > suggestion, non-blocking: If these are internal endpoints, we should be able to refactor them to return CR percentage as numbers, so we'd not need this utility

@apata do you think this should be done in a follow-up PR or rather in this one?

sanne-san:
> Not even a suggestion, just something that made me pause.
> 
> Question: should these
> 
> image
> correspond somehow visually with:
> 
> image
> Maybe color-wise?

They are more similar in the current iteration! Not entirely the same though because I think the caps should be reserved for the labels at the top of the report.

assets/js/dashboard/util/number-formatter.ts
sanne-san: > There's a FE rounding issue when the values are close to each other.

Good catch. Raised it for discussion [here](https://3.basecamp.com/5308029/buckets/26383192/card_tables/cards/10190292942#__recording_10231567093), the conclusion is to accept it for now.

aerosol: The reason we use strings is something something json float encoding, depending on the codec used, which is another category of errors IIRC cc @apata

apata: Ah, so @aerosol is saying there's a reason we use strings here. I haven't checked whether the app encoding / decoding percentage as number in the API response has any gotchas. We use it in the top stats and a few other places, so it's probably not too bad, but separate PR / effort is the right call. Thanks for calling me out on it!

sanne-san: Alright, thanks both! 👌

sanne-san: @apata RE rounding error: changed it so that any percentage under 0.1 is paired with a two decimal value (>99.9). This rounding logic also applies to Explore conversion rate which is fine, when a number is so close to 100 it's probably better to be accurate rather than suggest every single user converted.

apata: Great, that works! These little "two values don't add up" issues undermine trust IMO, happy that you found a solution.

assets/js/dashboard/extra/funnel/outcome-box.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  const { box, compact, expanded } = useOutcomeSize(label)
+
+  return (
+    <button
+      ref={box}
+      type="button"
+      aria-label={label}
+      data-open={open || undefined}
+      onClick={() => setOpen((shown) => !shown)}
```
apata, line 169: nitpick, non-blocking: the expanded state of an outcome doesn't reset when switching funnels

apata (approved):

assets/js/dashboard/extra/funnel/outcome-box.tsx
sanne-san: Fixed!

PR #6606, opened 2026-08-18
Distinguish repeated signup submissions

ukutaht:
### Background

PR #6514 changed signup tracking from a LiveView hook to an inline form onsubmit handler.

Before this change, Signup was tracked only after the user record was created successfully. After it, Signup is sent when the browser submits the form, before CAPTCHA verification and server-side validation complete.

Since this change, there is a significant gap between unique conversions and total conversions. This means the same users are re-submitting the form multiple times. The goal of this PR is to understand better if there's some UX friction on the registration form. Could also just be bots caught by captcha. But then why aren't the bots excluded from our tracking? Anyways, this will help narrow things down.

### Changes

Track the previous registration error on Signup events.

Each initial signup submission sends:
{"previous_error": "none"}

A later submission after a rejected attempt sends the previous CAPTCHA or validation error. This makes it possible to identify repeated submissions caused by issues such as CAPTCHA failure or an existing email address.

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible_web/live/register_form.ex
+++ b/lib/plausible_web/live/register_form.ex
@@ -50,6 +50,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
          captcha_error: nil,
          password_strength: Auth.User.password_strength(changeset),
          disable_submit: false,
+         signup_previous_error: nil,
          trigger_submit: false,
          heading: heading,
          subtitle: subtitle
@@ -95,7 +96,7 @@ defmodule PlausibleWeb.Live.RegisterForm do
           id="register-form"
           class="flex flex-col gap-y-6"
           action={Routes.auth_path(@socket, :login)}
-          onsubmit={form_submit_event(@invitation)}
+          onsubmit={form_submit_event(@invitation, @signup_previous_error)}
           phx-change="validate"
           phx-submit="register"
           phx-trigger-action={@trigger_submit}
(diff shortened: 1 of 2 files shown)


lib/plausible_web/live/register_form.ex
```diff
@@ -293,13 +298,23 @@ defmodule PlausibleWeb.Live.RegisterForm do
...
+          )
           |> PlausibleWeb.Components.Captcha.reset()
 
         {:noreply, socket}
     end
   end
 
+  defp signup_error(changeset) do
+    case List.first(changeset.errors) do
+      {_field, {message, _options}} -> message
```
apata: Recording changeset error message in event seems risky. What if it contains sensitive information? Maybe it doesn't right now (`"previous_error": "password too weak"`), but if it does in the future, the damage would be great (`"previous_error": "password 'swordfish' too weak"`)

Could additional logs / metrics help to get to the bottom of this instead? Or generating e2e tests locally while actively dogfooding to see which scenarios have the uniqueness issue.

ukutaht: I do want to use Plausible dogfooding for this as part of a bigger effort to dogfood more, and to make stats like this accessible to non-devs on the team.

The initial pass had a function to categorize the error to an enumerated field to avoid cardinality or sensitive information issues. I removed it because we do not echo the inputs back in error messages and it seems weird to do it. But I take your point, added back in for future-proofing in https://github.com/plausible/analytics/pull/6606/changes/a26a07b961028d4995ed21a5fdbba05e4db4dc84

apata (approved):

PR #6607, opened 2026-08-19
Reduce funnels on special goals delete

aerosol:
Deleting a goal is supposed to go through Plausible.Goals.delete/2, which correctly wipes an entire funnel when deleting one of its goals would drop it below the 2-step minimum. But the four "special goal" helpers never go through it (delete_404/1, delete_form_submissions/1 etc.) - called from sync_goals when tracker configuration changes.

### DIFF
--- a/lib/plausible/goals/goals.ex
+++ b/lib/plausible/goals/goals.ex
@@ -228,13 +228,16 @@ defmodule Plausible.Goals do
   end
 
   def delete(id, site_id) do
-    goal_query =
+    delete_matching(
       from(g in Goal,
         where: g.id == ^id,
         where: g.site_id == ^site_id
       )
+    )
+  end
 
-    goal_query = on_ee(do: preload(goal_query, funnels: :steps), else: goal_query)
+  defp delete_matching(base_query) do
+    goal_query = on_ee(do: preload(base_query, funnels: :steps), else: base_query)
 
     result =
       Multi.new()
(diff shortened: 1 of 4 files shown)


apata:
Code-wise, there's no issue, but what's worse for the user? Ending up with 0 or 1 step funnel, or a funnel disappearing with no warning or explanation? Rewriting the special goal sync UI to give warning is a bigger job of course, but I'd say it'd be more correct, particularly if there's been no complaints about these broken funnels.

aerosol:
@apata there are other ways to delete a goal or update tracker script configuration (e.g. the plugins API), in which case, yes, we reduce/delete funnels without asking anyone. Funnels can be easily recreated at any time and no data is lost. Anyway, I added the confirmation prompt to the tracker configuration toggle UI.

aerosol:
@apata Oh and to me this is way spookier, the tracking toggles add goals behind the scenes automatically and the goal deletion also happens without a warning. Funnels reduction/deletion should follow that, not make the user doubt whether it's safe to use.

aerosol:
To picture what I mean:

image


Nothing here communicates the goal dependency ever, let alone funnels. This is transparent by design I believe. You make it sound like funnels are something precious you don't want to accidentally lose if you've decided you want to stop tracking 404s. Hard disagree, but the code is there if you prefer it that way.

apata:
Hmm, I didn't think about the other ways funnels may disappear. Thanks for bringing those up! 

These were the scenarios I was looking at:
- Delete 404 goal in `Site > Settings > Goals`, user gets warning dialog about funnels being reduced. 
- Disable 404 tracking in `Site > Settings > General`, user won't get warning dialog. 

The other scenarios (in which there's no warning of funnels being affected):
- `DELETE /api/v1/sites/goals/:id`
- `PUT /api/v1/sites/:site_id` with `tracker_script_configuration` changes
- `DELETE /api/plugins/v1/goals/:id`
- `DELETE /api/plugins/v1/goals`
- `PUT /api/plugins/v1/tracker_script_configuration`

Looking at the changes, with the warning dialog on toggle, my preference would be no warning then, to merge just [24fa65e](https://github.com/plausible/analytics/pull/6607/commits/24fa65eb6d3d3b7124bb6701a3db9c0fb9363f68).

apata (approved):

PR #6608, opened 2026-08-19
Refactor onboarding layout + flow definitions

RobertJoonas:
### Changes

* Make new site form and installation page opt out of legacy layout
* Define <.onboarding> layout as a function component (as opposed to `onboarding.html.heex`)
* Give onboarding layout dynamically depending on `@flow == Flows.register()`
* Drop "steps" from flow definitions (dead code) and use public function `@doc` instead for describing what steps the flows include.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/templates/site/provisioning_new_site.html.heex
```diff
@@ -0,0 +1,20 @@
+<Layouts.app
```
ukutaht, line 1: Previously we had global notices disabled here. Should probably add `global_notices?={false}` unless this change was intentional

ukutaht: Same issue with the `installation.ex` liveview.

ukutaht (approved):

lib/plausible_web/templates/site/provisioning_new_site.html.heex
RobertJoonas: Good catch, thanks! https://github.com/plausible/analytics/pull/6608/commits/29401893a36f4bebc883d2a5d6bbf8a5f2a1a39d

PR #6609, opened 2026-08-19
Remove unvalidated site / team name from emails subject and body

apata:
### Changes

Removes unvalidated site and team names from automated email subject and body. Protects against the server being used to send mass emails.

The drawback is that the messages look more generic. 

I've kept the email of the inviter in the body: this should allow the recipient to verify directly with the inviter without clicking any links in the email. 

### Tests

### Changelog

### Documentation

### Dark mode


apata:
Closing, plan is to use some other measures

PR #6610, opened 2026-08-19
Experimental engaged session switch

ukutaht:
### Background

An experiment to see how it would look like to add a 'engaged sessions only' filter to deal with bot traffic. An engaged session is defined as a session with at least one `engagement` event.

This is intentionally the smallest way to test this in production. It is meant to be easily revertable. If the experiment is a success in that the filter:
* Helps clean up dashboards with spam
* Does not exclude too much real human traffic

Then the plan is to revert this commit and think about how to build this properly. Currently it ignores many considerations including realtime data, imported data, and historical data before `engagement` events were even added.

### Changes

For superadmins only, adds a UI switch that always applies a `[:has_done, [:is, "event:name", ["engagement"]]` filter to all API requests. Does not apply to the current visitors number and when enabled, imported data is inaccessible due to backend rules.

Screenshot 2026-08-19 at 23 12 18

### Tests

### Changelog
- [x] This PR does not make a user-facing change (experiment for super-admins only)

### Documentation

### Dark mode

### DIFF
--- a/assets/js/dashboard/api.ts
+++ b/assets/js/dashboard/api.ts
@@ -111,7 +111,10 @@ export function dashboardStateToParams(
     queryObj.to = formatISO(dashboardState.to)
   }
   if (dashboardState.filters) {
-    queryObj.filters = serializeApiFilters(dashboardState.filters)
+    queryObj.filters = serializeApiFilters(
+      dashboardState.filters,
+      dashboardState.engagedSessionsOnly
+    )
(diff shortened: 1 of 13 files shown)


aerosol (approved):
Interesting idea, let's see how it behaves

ukutaht:
Putting this on ice for now. Exploring alternative solutions to bot issues.

PR #6611, opened 2026-08-20
[Team Deletions] Migration: add team deletion schedule

aerosol:
This isn't final form, but an initial base migration for tracking/audting team deletions life cycle (expired trials, cancelled subscriptions).

Discussion: https://3.basecamp.com/5308029/buckets/44692882/messages/10216872798

### DIFF
--- a/priv/repo/migrations/20260820120000_create_team_deletion_schedules.exs
+++ b/priv/repo/migrations/20260820120000_create_team_deletion_schedules.exs
@@ -18,6 +18,7 @@ defmodule Plausible.Repo.Migrations.CreateTeamDeletionSchedules do
       timestamps()
     end
 
+    create index(:team_deletion_schedules, [:team_id])
     create index(:team_deletion_schedules, [:status])
     create index(:team_deletion_schedules, [:deletion_date])
     create index(:team_deletion_schedules, [:first_notice_due_date])
--- a/priv/repo/migrations/20260820120000_create_team_deletion_schedules.exs
+++ b/priv/repo/migrations/20260820120000_create_team_deletion_schedules.exs
@@ -22,14 +22,14 @@ defmodule Plausible.Repo.Migrations.CreateTeamDeletionSchedules do
     create index(:team_deletion_schedules, [:deletion_date])
     create index(:team_deletion_schedules, [:first_notice_due_date])
 
-    # Only one active (i.e. not yet cancelled/deleted) schedule per team - a team
+    # Only one active (i.e. not yet cancelled/completed) schedule per team - a team
     # can churn, come back, and churn again over time, so this isn't a plain
     # unique index on team_id.
     create index(
              :team_deletion_schedules,
              [:team_id],
              unique: true,
-             where: "status NOT IN ('cancelled', 'deleted')",
+             where: "status NOT IN ('cancelled', 'completed')",
              name: :one_active_schedule_per_team
            )
   end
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20260820120000_create_team_deletion_schedules.exs
```diff
@@ -0,0 +1,36 @@
...
+    create index(:team_deletion_schedules, [:first_notice_due_date])
+
+    # Only one active (i.e. not yet cancelled/completed) schedule per team - a team
+    # can churn, come back, and churn again over time, so this isn't a plain
+    # unique index on team_id.
+    create index(
+             :team_deletion_schedules,
+             [:team_id],
+             unique: true,
+             where: "status NOT IN ('cancelled', 'completed')",
```
ukutaht, line 33: How is `status: 'completed'` different from the row being deleted via cascade from the `teams` table deletion?

aerosol: We don't delete the team, we delete its sites. The team stays and remains still available for upgrade. Does that make sense?

aerosol: We have to support the case when the team subscribes for a month and then cancels only to renew next year

ukutaht: EDIT: Ignore stuff below, I didn't see your reply before posting. I am slightly surprised by that – I was assuming we delete the team. But if that's agreed with @metmarkosaric then all good. 

Thinking about the lifecycle after reading https://github.com/plausible/analytics/pull/6615/changes#r3852893234

When a deletion is executed, I'm assuming the `team` will be deleted and it cascades to this table automatically.

Is there a planned lifecycle state where the team still exists but the deletion schedule is in `completed` state?

If there is a use-case for this state, the `sync_eligible` function does not seem to account for that. In case team exists and deletion schedule is `completed`, it looks like another deletion schedule will be created (or is this intended?)

metmarkosaric: @ukutaht we just planned to delete dashboards and stats. accounts and teams we don't touch. it's mostly sites/stats that affect us in terms of data storage/backups and CS while it seems nice for people to keep access to their account, team setup, subscription history etc. it also makes it easier to come back and start again if/when needed

ukutaht: Hmm but if we only delete sites then won't the churned customer be on a never-ending cycle of deletions?

1. `sync_eligible` finds a churned customer with no active deletion schedule
2. Deletion schedule is created
3. Email notifications go out
4. Sites are deleted
5. Deletion schedule status is changed to `completed`
6. GOTO step 1

or am I missing something that prevents this?

aerosol: Yeah @ukutaht that's a valid observation. I'll address that

aerosol: It's a nice catch overall, I think what we should do is to ensure at least one site exists for a team in order to fulfill the eligibility check. That way already completed deletion sweeps exist as audit entries and don't loop, unless a new site is created (which implies subscription must go through its cycle again). Does that make sense?

aerosol: Pushed https://github.com/plausible/analytics/pull/6615/changes/14f4f0b98bdaf190661d55b05cd9d93074d35fd0

ukutaht (approved):

PR #6615, opened 2026-08-24
[Team Deletions] effective scheduling

aerosol:
On top of #6613 this PR implements the core deletion eligibility logic including the background worker (tested but kept inactive for now).

A daily worker seeks for eligible teams and bulk-inserts `team_deletion_schedules` rows.
Additional Paddle webhooks checks are planted to cancel any pending deletions upon subscription renewal.

Schema/migration change (deleted->completed rename) backported to previous branches in the stack.

### DIFF
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -69,21 +69,7 @@ defmodule Plausible.TeamDeletionSchedules do
     team = Teams.with_subscription(team)
 
     if Subscriptions.active?(team.subscription) do
-      {:ok, count} =
-        Repo.transact(fn ->
-          case active_schedule_for(team.id) do
-            nil ->
-              {:ok, 0}
-
-            schedule ->
-              case cancel(schedule) do
-                {:ok, _} -> {:ok, 1}
-                {:error, _} -> {:ok, 0}
-              end
-          end
-        end)
-
-      count
+      cancel_active_schedule(team.id)
     else
       0
     end
@@ -170,6 +156,25 @@ defmodule Plausible.TeamDeletionSchedules do
 
   defp transition(%TeamDeletionSchedule{status: from}, to, _extra_changes) do
     {:error, {:invalid_transition, from, to}}
+  end
+
+  defp cancel_active_schedule(team_id) do
+    {:ok, count} =
+      Repo.transact(fn ->
+        case active_schedule_for(team_id) do
+          nil -> {:ok, 0}
+          schedule -> {:ok, cancel_count(schedule)}
+        end
+      end)
+
+    count
+  end
+
+  defp cancel_count(schedule) do
+    case cancel(schedule) do
+      {:ok, _} -> 1
+      {:error, _} -> 0
+    end
   end
 
   defp active_schedule_for(team_id) do
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -117,6 +117,7 @@ defmodule Plausible.TeamDeletionSchedules do
   def due_for_deletion(today \\ Date.utc_today()) do
     Repo.all(
       from(sch in TeamDeletionSchedule,
+        inner_join: t in assoc(sch, :team),
         where: sch.status == :reminder_sent,
         where: sch.deletion_date <= ^today,
         preload: [:team]
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -10,6 +10,7 @@ defmodule Plausible.TeamDeletionSchedules do
   alias Plausible.Billing.Subscription
   alias Plausible.Billing.Subscriptions
   alias Plausible.Repo
+  alias Plausible.Site
   alias Plausible.TeamDeletionSchedule
   alias Plausible.Teams
   alias Plausible.Teams.DeletionSchedule
@@ -32,6 +33,7 @@ defmodule Plausible.TeamDeletionSchedules do
           where:
             fragment("coalesce((?->>'manual_lock')::boolean, false)", field(t, :grace_period)) ==
               false,
+          where: exists(from(s in Site.regular(), where: s.team_id == parent_as(:team).id)),
           where:
             (is_nil(s.id) and not is_nil(t.trial_expiry_date) and t.trial_expiry_date < ^today) or
               (not is_nil(s.id) and
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -75,6 +75,57 @@ defmodule Plausible.TeamDeletionSchedules do
     end
   end
 
+  @doc """
+  Schedules due for their first notice - still scheduled and
+  past their first_notice_due_date.
+  """
+  @spec due_for_first_notice(Date.t()) :: [TeamDeletionSchedule.t()]
+  def due_for_first_notice(today \\ Date.utc_today()) do
+    Repo.all(
+      from(sch in TeamDeletionSchedule,
+        where: sch.status == :scheduled,
+        where: sch.first_notice_due_date <= ^today,
+        preload: [team: [:owners, :billing_members]]
+      )
+    )
+  end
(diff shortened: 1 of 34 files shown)


lib/plausible/team_deletion_schedules.ex
```diff
@@ -0,0 +1,122 @@
...
+        from(t in Teams.Team,
+          as: :team,
+          left_lateral_join: s in subquery(Teams.last_subscription_join_query()),
+          on: true,
+          left_join: ep in assoc(t, :enterprise_plan),
+          where: is_nil(ep.id),
+          where:
+            fragment("coalesce((?->>'manual_lock')::boolean, false)", field(t, :grace_period)) ==
+              false,
+          where:
```
ukutaht: I think extracting these two candidate populations and naming them would improve readability. Something like:

```elixir
where: ^is_expired_trial(s, t) or ^is_expired_subscription(s)
```

I followed the `expired_subscription` naming because I saw it somewhere else in this diff. But I think we should also consider `churned_subscription` for naming this population.

"expired" - suggests a passive thing happening on its own
"cancelled" - suggests an explicit action by customer
"churned" - includes both, used heavily in SaaS billing. Some customers cancel actively while for others their card expires and the subscription is paused on their behalf. So I think this fits best.

lib/plausible/team_deletion_schedules.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  @doc """
+  Finds newly eligible teams (expired trial/churned subscription) and
+  schedules each for the deletion pipeline. Reactivation is handled
+  via billing (Paddle) handlers.
+  """
+  @spec sync_eligible(Date.t()) :: non_neg_integer()
+  def sync_eligible(today \\ Date.utc_today()) do
+    candidates =
+      Repo.all(
```
ukutaht, line 27: Worth wrapping candidate selection and schedule insertion in a transaction to avoid races

lib/plausible/team_deletion_schedules.ex
```diff
@@ -0,0 +1,122 @@
...
+          where: is_nil(ep.id),
+          where:
+            fragment("coalesce((?->>'manual_lock')::boolean, false)", field(t, :grace_period)) ==
+              false,
+          where:
+            (is_nil(s.id) and not is_nil(t.trial_expiry_date) and t.trial_expiry_date < ^today) or
+              (not is_nil(s.id) and
+                 s.status in [^Subscription.Status.deleted(), ^Subscription.Status.paused()] and
+                 s.paddle_plan_id != "free_10k" and s.next_bill_date < ^today),
+          where:
```
ukutaht, line 36: Here we leave the existing `DeletionSchedule` rows alone even if the team state has changed since it was created. Once the `DeletionSchedule` row has been inserted, it's still possible for CS to extend someone's trial, add an enterprise plan or add a manual lock. So the pending deletion schedules will get out of sync with team billing states.

I think it's fine for them to get out of sync as long as the email and deletion worker re-check everything and reconcile the deletion table as needed. I assume you've already thought of this but I just wanted to double check :)

lib/plausible/team_deletion_schedules.ex
```diff
@@ -0,0 +1,122 @@
...
+    candidates =
+      Repo.all(
+        from(t in Teams.Team,
+          as: :team,
+          left_lateral_join: s in subquery(Teams.last_subscription_join_query()),
+          on: true,
+          left_join: ep in assoc(t, :enterprise_plan),
+          where: is_nil(ep.id),
+          where:
+            fragment("coalesce((?->>'manual_lock')::boolean, false)", field(t, :grace_period)) ==
```
ukutaht: This is slightly mysterious to me at the moment (grace periods are not easy!). Just leaving a note for myself to review this in more depth. I'll post the comments I have so far because I'm taking a little break right now :)

ukutaht: It looks to me like the point of this is to exclude teams that have a manual grace period applied.

Manual grace periods are applied to enterprise plans and teams with enterprise plans are already excluded from candidates on line 31 above. What does the manual lock condition do that the enterprise plan exclusion doesn't? Or do we have cases where a manual lock is applied to a non-enterprise subscriber?

lib/plausible/team_deletion_schedules.ex
aerosol: I only thought of deletion snooze, not the fact that CS would want to extend the trial. Why would anyone want to add a lock though? (to a team that implicitly doesn't accept traffic anymore?). Having a look if there are other team states we need to watch out for. This is starting to cry for some event-driven fsm transitions.

ukutaht (approved):

lib/plausible/team_deletion_schedules.ex
aerosol: Man, I never remember how those grace periods/locks work. You're right, they're only applied to enterprise plans and enterprise plans cannot be deleted afterwards AFAIU

aerosol: a0b362d975

lib/plausible/team_deletion_schedules.ex
aerosol: a2e2e968b6

lib/plausible/team_deletion_schedules.ex
aerosol: Can't wrap my head around this yet, leaving it up for discussion at https://3.basecamp.com/5308029/buckets/44692882/messages/10216872798#__recording_10236982367

lib/plausible/team_deletion_schedules.ex
aerosol: Err, can you describe an example race you're thinking of preventing here?

ukutaht: 1.   Candidates are selected for deletion
 2.  Webhook arrives, rendering one team ineligible for deletion
 3.  Deletion schedules are inserted including for the ineligible team

Or

 1.   Candidates are selected for deletion
 2.  One of the candidate teams is deleted
 3.  Deletion schedule insert_all fails due to foreign key constraint

I'm not sure if any of these would result in something bad happening but using a transaction here just feels like good hygene anyways :)

aerosol: First scenario cannot be prevented with a transaction unfortunately. The result is always a team with an active subscription and a scheduled deletion row that hasn't been cancelled. 

Webhook cancellation runs in a separate already commiting transaction and no isolation level on sync_eligible's own transaction affects what a concurrently commiting transaction does. 

Same if we did explicit each-team row locking with `SELECT FOR UPDATE`, the webhook would block for as long as the oban job's lock is held, but then.. the result is the same. 

Correct me if I'm wrong.

I'll implement another `cancel_for_team` call/check before sending anything out and advancing the schedule status. 

Second scenario is perfectly fine, worst case we'll crash and retry (it's an oban job really).

ukutaht: Yeah, fair point I was expecting too much from transactions :) no further comments from me

lib/plausible/team_deletion_schedules.ex
```diff
@@ -0,0 +1,138 @@
...
+      Repo.insert_all(TeamDeletionSchedule, rows,
+        on_conflict: :nothing,
+        conflict_target: {:unsafe_fragment, terminal_statuses_index_predicate()}
+      )
+
+    count
+  end
+
+  @doc """
+  Cancels any pending deletion schedule for a team
```
zoldar: Documenting the meaning of the returned value and perhaps a rought context it might be used in would be helpful 🙏

aerosol: Alright, will do!

lib/plausible/team_deletion_schedules.ex
aerosol: Done in #6630

lib/plausible/team_deletion_schedules.ex
aerosol: It's now quite literal, so no need for that anymore.

PR #6617, opened 2026-08-24
Limit team name length and content

apata:
### Changes

Limits team name length and forbids names like `buy stuff at https://spam.example.com`. Depends on #6616 

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/teams/team.ex
```diff
@@ -133,6 +172,23 @@ defmodule Plausible.Teams.Team do
     change(team, trial_expiry_date: Date.utc_today() |> Date.shift(day: -1))
   end
 
+  defp validate_name(changeset) do
+    changeset
+    |> validate_length(:name, max: @max_name_length)
+    |> validate_length(:name, count: :bytes, max: @max_name_bytes)
+    |> validate_change(:name, fn :name, name ->
```
aerosol, line 179: validate_format?

lib/plausible/teams/team.ex
```diff
@@ -21,6 +21,13 @@ defmodule Plausible.Teams.Team do
   @trial_accept_traffic_until_offset_days 14
   @subscription_accept_traffic_until_offset_days 30
 
+  @max_name_length 50
+  # `teams.name` is a varchar(255), which counts code points. Capping bytes
```
aerosol, line 25: Is there a reason to keep `varchar(255)`? Can't we migrate to `varchar(50)` and get everything trimmed automatically?

Otherwise, what's the plan for existing long team names?

apata: There's so few team names longer than 50 characters that out of those, the spammy ones can be deleted manually and the legit ones can stay. 

The only reason to limit length to 50 is to make spamming less attractive, and non-spammy long names are not actually a problem. Column migration seems heavy. How would we present that change to CE users?

lib/plausible/teams/team.ex
apata: Thanks, didn't know about that utility. Checked and we could use it, but I'd prefer not to. 

The pattern `~r/^(?!.*:\/\/)/s` scales linearly (some patterns don't, including another one I tried before), but it's still 400x times slower according to local tests than String.contains?. 

The input to this is uncontrolled and the change / format validator runs even if the change is already determined to be over max length, so expensive pattern + long input -> wasted server resources.

ukutaht (approved):

PR #6624, opened 2026-08-27
[Team Deletions] deletion e-mails

aerosol:
### Changes

This PR introduces two new emails: deletion full notice (T-30d, with trial/subscription branching and capped site list) and deletion reminder (T-5d). Each subscription is re-cehcekd before sending to address the reactivation race. 

AcceptTrafficUntil worker's "stats stop tomorrow" e-mail now includes the pending deletion date for steady-state (non-backlog) expired trials, instead of separate e-mail.

The worker is disabled for now.

@metmarkosaric if you could have a final look at the copy please

on top of #6623

### DIFF
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -117,6 +117,7 @@ defmodule Plausible.TeamDeletionSchedules do
   def due_for_deletion(today \\ Date.utc_today()) do
     Repo.all(
       from(sch in TeamDeletionSchedule,
+        inner_join: t in assoc(sch, :team),
         where: sch.status == :reminder_sent,
         where: sch.deletion_date <= ^today,
         preload: [:team]
--- a/lib/plausible/team_deletion_schedules.ex
+++ b/lib/plausible/team_deletion_schedules.ex
@@ -75,6 +75,57 @@ defmodule Plausible.TeamDeletionSchedules do
     end
   end
 
+  @doc """
+  Schedules due for their first notice - still scheduled and
+  past their first_notice_due_date.
+  """
+  @spec due_for_first_notice(Date.t()) :: [TeamDeletionSchedule.t()]
+  def due_for_first_notice(today \\ Date.utc_today()) do
+    Repo.all(
+      from(sch in TeamDeletionSchedule,
+        where: sch.status == :scheduled,
+        where: sch.first_notice_due_date <= ^today,
+        preload: [team: [:owners, :billing_members]]
+      )
+    )
+  end
+
+  @doc """
+  Schedules due for the reminder: first notice already sent.
+  Applies to backlog rows too.
+  """
+  @spec due_for_reminder(Date.t()) :: [TeamDeletionSchedule.t()]
+  def due_for_reminder(today \\ Date.utc_today()) do
+    reminder_threshold = Date.add(today, DeletionSchedule.reminder_before_deletion_days())
+
+    Repo.all(
+      from(sch in TeamDeletionSchedule,
+        where: sch.status == :first_notice_sent,
+        where: sch.deletion_date <= ^reminder_threshold,
+        preload: [team: [:owners, :billing_members]]
+      )
+    )
+  end
+
+  @doc """
+  Pending, non-backlog expired trial schedules for the
+  given team ids, keyed by `team_id`
+  """
+  @spec pending_steady_state_trials_by_team_id([pos_integer()]) :: %{
+          pos_integer() => TeamDeletionSchedule.t()
+        }
+  def pending_steady_state_trials_by_team_id([]), do: %{}
+
+  def pending_steady_state_trials_by_team_id(team_ids) do
+    TeamDeletionSchedule
+    |> where([sch], sch.team_id in ^team_ids)
+    |> where([sch], sch.category == :expired_trial)
+    |> where([sch], sch.status == :scheduled)
+    |> where([sch], sch.is_backlog == false)
+    |> Repo.all()
+    |> Map.new(&{&1.team_id, &1})
+  end
+
   @type transition_result ::
           {:ok, TeamDeletionSchedule.t()} | {:error, {:invalid_transition, atom(), atom()}}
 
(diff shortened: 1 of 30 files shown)


metmarkosaric:
@aerosol thanks, the copy looks good!

lib/plausible/team_deletion_schedules.ex
```diff
@@ -75,6 +75,57 @@ defmodule Plausible.TeamDeletionSchedules do
...
+  def pending_steady_state_trials_by_team_id([]), do: %{}
+
+  def pending_steady_state_trials_by_team_id(team_ids) do
+    TeamDeletionSchedule
+    |> where([sch], sch.team_id in ^team_ids)
+    |> where([sch], sch.category == :expired_trial)
+    |> where([sch], sch.status == :scheduled)
+    |> where([sch], sch.is_backlog == false)
+    |> Repo.all()
+    |> Map.new(&{&1.team_id, &1})
```
zoldar, line 182: Wouldn't that overwrite repeating occurrences of `team_id`? Maybe you meant `Enum.group_by` ?

lib/plausible/team_deletion_schedules.ex
```diff
@@ -75,6 +75,57 @@ defmodule Plausible.TeamDeletionSchedules do
...
+  Schedules due for their first notice - still scheduled and
+  past their first_notice_due_date.
+  """
+  @spec due_for_first_notice(Date.t()) :: [TeamDeletionSchedule.t()]
+  def due_for_first_notice(today \\ Date.utc_today()) do
+    Repo.all(
+      from(sch in TeamDeletionSchedule,
+        where: sch.status == :scheduled,
+        where: sch.first_notice_due_date <= ^today,
+        preload: [team: [:owners, :billing_members]]
```
zoldar: That's a very minor improvement but at least `team` preload could be done from a join.

lib/plausible/team_deletion_schedules.ex
```diff
@@ -75,6 +75,57 @@ defmodule Plausible.TeamDeletionSchedules do
...
+  """
+  @spec due_for_reminder(Date.t()) :: [TeamDeletionSchedule.t()]
+  def due_for_reminder(today \\ Date.utc_today()) do
+    reminder_threshold = Date.add(today, DeletionSchedule.reminder_before_deletion_days())
+
+    Repo.all(
+      from(sch in TeamDeletionSchedule,
+        where: sch.status == :first_notice_sent,
+        where: sch.deletion_date <= ^reminder_threshold,
+        preload: [team: [:owners, :billing_members]]
```
zoldar: Ditto about preloading `team` from a join.

lib/workers/accept_traffic_until_notification.ex
```diff
@@ -48,15 +51,25 @@ defmodule Plausible.Workers.AcceptTrafficUntil do
           group_by: [u.id, t.id]
       )
 
+    pending_trial_schedules_by_team_id =
+      notifications
+      |> Enum.filter(&(&1.deadline == tomorrow))
```
zoldar: Even though we don't do any fancy stuff like mixing calendars, I'd opt for using `Date.compare` instead of struct comparison, just to be safe.

lib/plausible/team_deletion_schedules.ex
zoldar: Sorry, scratch that, I've read the typespec wrong and for some reason thought it's a map of maps 🤦

lib/workers/accept_traffic_until_notification.ex
```diff
@@ -75,6 +88,14 @@ defmodule Plausible.Workers.AcceptTrafficUntil do
       end
     end
 
+    if not dry_run? do
+      pending_trial_schedules_by_team_id
+      |> Map.values()
+      |> Enum.each(fn schedule ->
+        TeamDeletionSchedules.mark_first_notice_sent(schedule, report_if_invalid?: true)
```
zoldar: Why not run it directly after sending the notification in the original loop? 🤔

zoldar (approved):

lib/workers/accept_traffic_until_notification.ex
aerosol: ~~Poor man's fault tolerance. Correct me if I'm wrong: first, we send out a bunch of notifications and store "send_accept_traffic_until_notifications" rows for each, in the same loop, to minimize the risk of sending dupe e-mails in case the job runs again.~~

~~Then in another loop, we'll mark deletions schedules as "first notice sent" so deletion schedules can advance. If any of those DB updates fails, the worker gets restarted - it won't send first notice e-mails again, but it will advance the deletion schedules that weren't touched due to previous crash.~~

~~If we had it in one loop, a crash would restart the worker but there's a chance it would then skip advancing deletion schedules leaving them in stale state forever.~~

aerosol: Nevermind, you were right, will fix that 🫡

aerosol: 70acf506c5 WYT?

lib/workers/accept_traffic_until_notification.ex
aerosol: ee4b089620

lib/plausible/team_deletion_schedules.ex
aerosol: b8fbc39061

lib/plausible/team_deletion_schedules.ex
aerosol: b8fbc39061

lib/workers/accept_traffic_until_notification.ex
```diff
@@ -78,6 +74,43 @@ defmodule Plausible.Workers.AcceptTrafficUntil do
...
+
+        schedule ->
+          # Advance the schedule before composing/sending the email,
+          # so a crash right after this point can never leave it stuck at :scheduled 
+          # while the customer has already been told the deletion date.
+          case TeamDeletionSchedules.mark_first_notice_sent(schedule, report_if_invalid?: true) do
+            {:ok, updated} -> updated.deletion_date
+            {:error, _} -> nil
+          end
+      end
```
zoldar:
```suggestion
      if schedule = Map.get(pending_trial_schedules_by_team_id, notification.team.id) do
        # Advance the schedule before composing/sending the email,
        # so a crash right after this point can never leave it stuck at :scheduled 
        # while the customer has already been told the deletion date.
        case TeamDeletionSchedules.mark_first_notice_sent(schedule, report_if_invalid?: true) do
          {:ok, updated} -> updated.deletion_date
          {:error, _} -> nil
        end
      end
```

might be slightly less noisy?

lib/workers/accept_traffic_until_notification.ex
zoldar: Looks good! Left a suggestion for further slight simplification of expression.

PR #6625, opened 2026-08-27
Prioritize domain exact match when looking up site for user

aerosol:
Ref https://3.basecamp.com/5308029/buckets/36789884/card_tables/cards/10244275997

### DIFF
--- a/lib/plausible/sites.ex
+++ b/lib/plausible/sites.ex
@@ -537,6 +537,8 @@ defmodule Plausible.Sites do
         where: coalesce(gm.role, tm.role) in ^roles,
         where: s.domain == ^domain or s.domain_changed_from == ^domain,
         where: is_nil(gm.id) or gm.site_id == s.id,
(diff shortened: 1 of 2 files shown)


lib/plausible/sites.ex
```diff
@@ -537,6 +537,8 @@ defmodule Plausible.Sites do
         where: coalesce(gm.role, tm.role) in ^roles,
         where: s.domain == ^domain or s.domain_changed_from == ^domain,
```
aerosol, line 538: Turns out this was done to support ongoing Sites API automations during the domain transition period. Doubt there's much value in that either...

zoldar: You mean it would be better to not consider `domain_changed_from` at all when handling requests for Sites API?

zoldar (approved):

lib/plausible/sites.ex
aerosol: Yeah, but let's not make breaking changes

PR #6632, opened 2026-09-01
Add support for comparison in funnels

sanne-san:
### Changes

- Return a comparison funnel and date ranges from the stats API
- Show two bars per step, with conversion-rate change next to the current rate
- Add change arrow to header conversion rate when comparing, with a tooltip showing the comparison conversion rate

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/extra/lib/plausible/stats/funnel.ex
+++ b/extra/lib/plausible/stats/funnel.ex
@@ -43,10 +43,10 @@ defmodule Plausible.Stats.Funnel do
        name: funnel.name,
        strict_order: funnel.strict_order,
        comparison: comparison,
-       date_range: iso_date_range(query.utc_time_range, query.timezone),
+       date_range: tz_date_range(query.utc_time_range, query.timezone),
        comparison_date_range:
          if(query.comparison_utc_time_range,
-           do: iso_date_range(query.comparison_utc_time_range, query.timezone)
+           do: tz_date_range(query.comparison_utc_time_range, query.timezone)
          )
      })}
   end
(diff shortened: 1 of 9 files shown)


extra/lib/plausible/stats/funnel.ex
```diff
@@ -169,4 +189,9 @@ defmodule Plausible.Stats.Funnel do
     |> elem(2)
     |> Enum.reverse()
   end
+
+  defp iso_date_range(utc_time_range, timezone) do
+    range = DateTimeRange.to_date_range(utc_time_range, timezone)
+    [Date.to_iso8601(range.first), Date.to_iso8601(range.last)]
+  end
```
zoldar: Explicit cast to ISO format is not necessary as serialisation handles it. It could be slightly simplified to something like:


```suggestion
  defp tz_date_range(utc_time_range, timezone) do
    range = DateTimeRange.to_date_range(utc_time_range, timezone)
    [range.first, range.last]
  end
```

extra/lib/plausible/stats/funnel.ex
```diff
@@ -29,6 +29,29 @@ defmodule Plausible.Stats.Funnel do
...
+      end
+
+    {:ok,
+     query
+     |> compute(funnel)
+     |> Map.merge(%{
+       name: funnel.name,
+       strict_order: funnel.strict_order,
+       comparison: comparison,
+       date_range: iso_date_range(query.utc_time_range, query.timezone),
```
zoldar:
```suggestion
       date_range: tz_date_range(query.utc_time_range, query.timezone),
```

extra/lib/plausible/stats/funnel.ex
```diff
@@ -29,6 +29,29 @@ defmodule Plausible.Stats.Funnel do
...
+       comparison_date_range:
+         if(query.comparison_utc_time_range,
+           do: iso_date_range(query.comparison_utc_time_range, query.timezone)
```
zoldar:
```suggestion
           do: tz_date_range(query.comparison_utc_time_range, query.timezone)
```

extra/lib/plausible/stats/funnel.ex
zoldar: https://github.com/plausible/analytics/pull/6632/commits/46299d1b7596200b40ea104716410b0dbec4f449

zoldar (approved):

PR #6634, opened 2026-09-02
[Team Deletions] Spread expired/churned backlog across 120d window

aerosol:
On top of #6630 

Ensures the backlog schedules don't go all at once

(CI errors related to CE. This feature is EE-exclusive and will be moved to `extra/` later on in the stack)

PR #6642, opened 2026-09-03
[Team Deletions] Make it ee-exclusive (move to extra/ etc.)

aerosol:
On top of #6634


aerosol:
Will re-submit, everything went wrong here

PR #6646, opened 2026-09-03
Enforce `DISABLE_REGISTRATION` in LiveView too

zoldar:
### Changes

Alternative approach to https://github.com/plausible/analytics/pull/6638. 
- Uses `on_mount` to fully lock out even mounting LiveView in a particular context. 
- Non-existent invitations case is handled by redirecting to a dedicated view instead of rendering in LV directly - redirect prevents proper mount of LV. 
- Handler actions are still guarded to avoid a situation where user with LV already mounted can finish their registration after another user has already registered. 
- Conditional rendering of registration related elements is now using a new function instead of checking runtime config directly. 
- Auth layout is now rendering flash messages (they were not passed to the layout template for some reason)

### Tests

### DIFF
--- a/lib/plausible_web/live/register_form.ex
+++ b/lib/plausible_web/live/register_form.ex
@@ -236,27 +236,49 @@ defmodule PlausibleWeb.Live.RegisterForm do
         %{"user" => _} = params,
         %{assigns: %{invitation: %{} = invitation}} = socket
       ) do
-    if PlausibleWeb.Captcha.verify(params["frc-captcha-response"]) do
-      user =
-        params["user"]
-        |> Map.put("email", invitation.email)
-        |> Auth.User.new()
+    case Plausible.Auth.check_registration_enabled(:invitation) do
+      :ok ->
+        if PlausibleWeb.Captcha.verify(params["frc-captcha-response"]) do
+          user =
+            params["user"]
+            |> Map.put("email", invitation.email)
+            |> Auth.User.new()
 
-      with_team? = invitation.type == :site_transfer
+          with_team? = invitation.type == :site_transfer
 
-      add_user(socket, user, with_team?: with_team?)
-    else
-      {:noreply, captcha_failed(socket)}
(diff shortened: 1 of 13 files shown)


lib/plausible_web/live/register_form.ex
cnkk, line 264: Wouldn't we still need to guard this?
(theory) e.g. when there is no admin account yet
* 2 tabs opened with /register
* user is created on tab 1
* tab 2 would still be able to create a user

lib/plausible_web/templates/page/index.html.heex
```diff
@@ -20,7 +20,10 @@
           Login
         </.styled_link>
       </:item>
-      <:item>
+      <:item :if={
+        Keyword.fetch!(Application.get_env(:plausible, :selfhost), :disable_registration) ==
```
cnkk: better going through `check_registration_enabled` so it is still displayed on first launch as well

lib/plausible_web/live/register_form.ex
zoldar: Yeah, indeed... ugh. I'll make a test case and plug it too.

lib/plausible_web/templates/page/index.html.heex
zoldar: https://github.com/plausible/analytics/pull/6646/commits/afdb397946c9a58c730a39901b60fa0ae9a6d557

lib/plausible_web/live/register_form.ex
zoldar: https://github.com/plausible/analytics/pull/6646/commits/cda34bb263d584da0e70de4828423f0d8f551fd7

zoldar:
@plausible/developers Does anyone recall if there was a specific reason for _not_ passing on flash messages to auth layout, or is it just an accidental omission during one of refactors?

aerosol (approved):
Looks good, tested manually in ce_dev

PR #6650, opened 2026-09-07
Upgrade oban

aerosol:
https://oban.hexdocs.pm/2.23.1/changelog.html#bug-fixes

[Peer] Scope lease renewal to the owning node

Leadership renewal matched the peers row by name alone, so a node that lost its lease renewed whichever row it found and concluded it was still the leader. The belief never corrected, letting a second node run Cron and insert duplicate jobs until restart.

Renewal is now scoped to the row's owner, so a peer that lost the lease demotes at the next election. Failed elections concede leadership rather than retaining it, and no longer crash the peer.


zoldar:
Removed migration because CI won't allow it. The migration will be released _after_ the bump. Given the migration contents (the only enum addition in the schema applies to Oban Pro feature), it's safe to do.

zoldar (approved):

PR #6651, opened 2026-09-07
Migrate Oban tables to schema v14

zoldar:
### Changes

Extracted from https://github.com/plausible/analytics/pull/6650


zoldar:
Hm, no, that won't work. The migration must be run after the bump.

PR #6656, opened 2026-09-08
Implement routes for discovering /mcp and the OAuth endpoints

apata:
### Changes

This adds the routes to discover the `/mcp` and OAuth endpoints. 

What we expect clients to do when wanting to request data at `#{server}/#{path}`:
1) add `/.well-known/oauth-protected-resource/` between `server` and `path` 
2) learn on that path how to auth.
I've removed the fallback route get "/.well-known/oauth-protected-resource" for now. If it turns out to be necessary, can always be added back in.

It also proposes a general `Plausible.Auth.Scopes` module, to be used to centralize scope names, descriptions. Adopting the general module means refactoring some code related to API keys as well. The alternative is to redeclare the same `stats:read:*` etc scopes separately for OAuth. 

In anticipation of more than one resource protected by OAuth (maybe Plugins API as well), this adds the `ProtectedResources` module.

### Tests

### Changelog
- [x] Will be documented together with /mcp release

### Documentation

### Dark mode

### DIFF
--- a/lib/plausible/auth/scopes.ex
+++ b/lib/plausible/auth/scopes.ex
@@ -1,0 +1,12 @@
+defmodule Plausible.Auth.Scopes do
+  @moduledoc """
+  This module enumerates the scopes that any particular
+  authorization grant can be limited to.
+  """
+
+  @stats_read_star "stats:read:*"
+  def stats_read(), do: @stats_read_star
+
+  @sites_read_star "sites:read:*"
+  def sites_read(), do: @sites_read_star
+end
--- a/lib/plausible/auth/scopes.ex
+++ b/lib/plausible/auth/scopes.ex
@@ -4,9 +4,9 @@ defmodule Plausible.Auth.Scopes do
   authorization grant can be limited to.
   """
 
-  @stats_read_star "stats:read:*"
-  def stats_read(), do: @stats_read_star
+  @stats_read_all "stats:read:*"
+  def stats_read(), do: @stats_read_all
 
-  @sites_read_star "sites:read:*"
-  def sites_read(), do: @sites_read_star
+  @sites_read_all "sites:read:*"
+  def sites_read(), do: @sites_read_all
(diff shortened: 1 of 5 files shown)


lib/plausible/auth/scopes.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Plausible.Auth.Scopes do
+  @moduledoc """
+  This module enumerates the scopes that any particular
+  authorization grant can be limited to.
+  """
+
+  @stats_read_star "stats:read:*"
```
zoldar: How about `@stats_read_all`?

lib/plausible/auth/scopes.ex
```diff
@@ -0,0 +1,12 @@
...
+  def stats_read(), do: @stats_read_star
+
+  @sites_read_star "sites:read:*"
```
zoldar: Ditto about name conveying meaning rather than syntax.

lib/plausible_web/controllers/oauth/metadata_controller.ex
```diff
@@ -5,26 +5,42 @@ defmodule PlausibleWeb.OAuth.MetadataController do
...
+    })
   end
 
-  defp not_implemented(conn) do
-    conn
-    |> put_status(501)
-    |> json(%{error: "not_implemented"})
-  end
+  defp authorization_endpoint(), do: "/login/oauth/authorize"
+  defp token_endpoint(), do: "/login/oauth/token"
```
zoldar: Is there a particular reason for which these are functions and not module attributes?

zoldar (approved):

lib/plausible_web/controllers/oauth/metadata_controller.ex
apata: Thanks! There's no particular reason, will swap to module attributes.

PR #6660, opened 2026-09-09
Convert to verified routes

zoldar:
### Changes

This PR makes a sweeping conversion from using router helpers for building paths and urls to [verified routes](https://phoenix.hexdocs.pm/Phoenix.VerifiedRoutes.html).

I have made several passes with an adapted mix task interleaved with manual corrections and several manual passes of manual verification. Still, a chance I have missed something is definitely non-zero. I'd very appreciate your 👀 on this.

Please note I convert largely only existing route helpers calls. I do not convert existing plain-string routes to verified routes. The scope of this PR is already huge anyway. This is something that can be addressed successively in the future.

### Why now?

We have considered moving to verified routes in the past (Phoenix had them for quite some time already and they are treated as a standard now). However, at the time we were evaluating them, they didn't work for us due to specific way our routing is setup. To be specific, we have three routes which amount to a universal catch-all to match on sites' domains:

```elixir
    get "/share/:domain/*path", StatsController, :shared_link

    ...

    delete "/:domain", SiteController, :delete_site
    ...
      get "/:domain", StatsController, :stats
      get "/:domain/*path", StatsController, :stats
```

Because of these routes, every verified route was treated as valid. This defeated the purpose of the feature, so we gave up on it.

In the mean time, a new [option](https://phoenix.hexdocs.pm/Phoenix.Router.html#match/5-options) got implemented for routes, called `warn_on_verify`:

> `:warn_on_verify` - the boolean for whether matches to this route in verified routes should emit a warning, rather than being accepted as verified. It is useful to ignore an otherwise catch-all route definition from being matched when verifying routes. Defaults `false`.

With this mechanism available, we are finally available to fully move to the verified routes.

### How

First thing that had to be done was adding `warn_on_verify: true` to all the routes above.

Next, `PlausibleWeb.VerifiedRoutes` got introduced. It must be `use`d in every context where verified routes are called. In addition to providing the `sigil_p` and a bunch of other standard macros and functions from `Phoenix.VerifiedRoutes`, it also exposes a couple additional functions for building wildcard paths and URLs. They are built using `unverified_path|url`, because those specific cases of wildcard routes are explicitly throwing warnings now.

With that setup in place, full conversion could proceed. For initial (and further passes) I have adapted an existing mix task I have found at https://gist.github.com/andreaseriksson/e454b9244a734310d4ab74d8595f98cd. Eventually after a number of passes, I have somewhat extended it to correctly convert more cases of interpolations and multi-line route expressions: https://gist.github.com/zoldar/8b3dd506cab062318ab09796d4b84df1. It's not ideal and fails on some trickier cases, but it still helped to handle quite substantial amount of work. However, a lot of manual interventions were still necessary due to a number of caveats. Some of those caveats might also be relevant in future use of verified routes - more on those below.

### Caveats

While, the [documentation](https://phoenix.hexdocs.pm/Phoenix.VerifiedRoutes.html) covers the basics, there are some things worth mentioning that I've stumbled upon during conversion

#### Prefer using keyword list for params over a map

This is relevant because, for instance, asserting on redirects in tests relies on it when checking equality. While there is an option that can be set in Phoenix config to always sort params in verified routes, it's incurring additional runtime cost, so at least for now, I've opted no to use it (though on a second thought, it could be set in test env only; regardless, being consistent might have other benefits).

#### When there are multiple routes, with the same verb, pointing at the same exact controller action, only the first one is treated as valid

En example:

```elixir
    delete "/:domain/settings/google-search", SiteController, :delete_google_auth
    delete "/:domain/settings/google-import", SiteController, :delete_google_auth
```

In this case, only the first path can be used as a verified route. That was the only case I've stumbled upon in our code base, but it's something worth keeping in mind.

#### It's no possible to use verified routes outside functions, for instance in module-level macros

This means existing uses of path helpers for building tests via macros, like:

```elixir
 describe "unauthorized calls" do
      for {method, url} <- [
            {:get, Routes.plugins_api_funnels_url(PlausibleWeb.Endpoint, :index)},
            {:get, Routes.plugins_api_funnels_url(PlausibleWeb.Endpoint, :get, 1)},
            {:put, Routes.plugins_api_funnels_url(PlausibleWeb.Endpoint, :create, %{})}
          ] do
        test "unauthorized call: #{method} #{url}", %{conn: conn} do
          conn
          |> unquote(method)(unquote(url))
          |> json_response(401)
          |> assert_schema("UnauthorizedError", spec())
        end
...
```

cannot be converted straight to verified routes - it's going to throw an error during compilation. Verified routes require @endpoint in env context to work properly. A workaround for this is giving up on verification in this particular case, and using `unverified_url` to wrap a path string at runtime:

```elixir
    describe "unauthorized calls" do
      for {method, url} <- [
            {:get, "/api/plugins/v1/funnels"},
            {:get, "/api/plugins/v1/funnels/1"},
            {:put, "/api/plugins/v1/funnels"}
          ] do
        test "unauthorized call: #{method} #{url}", %{conn: conn} do
          conn
          |> unquote(method)(unverified_url(PlausibleWeb.Endpoint, unquote(url)))
          |> json_response(401)
          |> assert_schema("UnauthorizedError", spec())
        end
      end
    end
...
```

#### Verification of static routes is pretty limited

The static routes configured via `statics` option passed to `use Phoenix.VerifiedRoutes` are checked only by prefix (literally, using `String.starts_with?` against the path). This means they match and validate any path that starts with them. So, for instance if we have `css` among statics, both, `/css/style.css` and `/csshelly/boo` are going to be treated as valid without further checking. It's understandable that static paths can't be verified any further because assets can be served from external resources with urls proxied or rewitten but it's a bit of a bummer that it doesn't check it's a complete path component. I'm not sure if the decision is deliberate or is it an accident - I might try to bring it up with phoenix team.

We currently don't have any overlapping prefixes between statics and actual app routes that would cause issues, but it's something we have to pay attention to, because it's easy to silently compromise verification this way.

#### Interpolating a complete query string as a string won't work

The query string is expected as either a map, keyword list (preferred) or potentially it can be interpolated directly with `~p"/some/path?first=#{param1}&second=#{param2}"`. However, passing a full query string as `~p"/some/path?#{query_string}" won't work, because it's url-encoded as a whole.

#### Order of routes matters

Before the change, the order of shared link routes was like following:

```elixir
    get "/share/:domain/*path", StatsController, :shared_link
    post "/share/:slug/authenticate", StatsController, :authenticate_shared_link
```

First, without adding `warn_on_verify` to the wildcard path, all matches that should hit the authenticate route and trigger warning (like `~p"/share/plausible.io/authenticate/whatever"`) were silently accepted, because the wildcard match swallowed it. After adding the option, `authenticate` couldn't be matched because the wildcard path was always throwing a warning. The solution is always putting the wildcard path last:

```elixir
    post "/share/:slug/authenticate", StatsController, :authenticate_shared_link
    get "/share/:domain/*path", StatsController, :shared_link, warn_on_verify: true
```

#### Unverified paths basically do no additional processing of the passed string paths

Something to keep in mind for the future when adding new exceptions. For instance, the `domain` component must be explicitly encoded. Parsing star path in `shared_stats_path` is re-implemented on the basis of what `Phoenix.VerifiedRoutes` does internally at https://github.com/phoenixframework/phoenix/blob/v1.8.14/lib/phoenix/verified_routes.ex#L770-L783


lib/plausible_web/templates/site/settings_integrations.html.heex
```diff
@@ -23,7 +23,7 @@
 
         <.button_link
           theme="danger"
-          href={Routes.site_path(@conn, :delete_google_auth, @site.domain)}
+          href={~p"/#{@site.domain}/settings/google-search"}
```
zoldar, line 26: There are 2 different paths pointing at exact same controller action and verified routes treats only one as valid.

lib/plausible_web/verified_routes.ex
```diff
@@ -0,0 +1,73 @@
...
+      use Phoenix.VerifiedRoutes,
+        router: PlausibleWeb.Router,
+        endpoint: PlausibleWeb.Endpoint,
+        statics: unquote(static_paths)
+
+      import unquote(__MODULE__)
+    end
+  end
+
+  def stats_path(domain, params \\ []) when is_binary(domain) and byte_size(domain) > 0 do
```
aerosol, line 28: There's one potential regression issue with this, but I am not sure about severity, if any:

```
iex(4)> PlausibleWeb.VerifiedRoutes.stats_path("a b.com")
"/a+b.com"
iex(5)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :stats, "a b.com", [])
"/a%20b.com"
```

aerosol (approved):
As far as I can tell it looks good 😅

One potential regression - commented inline.

I think it's fine to yolo merge it and do any follow ups if necessary (should be fairly quickly known).

lib/plausible_web/verified_routes.ex
zoldar: Aaah, good point. We are using `URI.encode_www_form/1` to encode the domain in manually built URLs quite often, but indeed, the router helpers use segment encoding method which I've already reimplemented for adding star paths. I'm applying segment encoding to domain as well, even if to keep things consistent in https://github.com/plausible/analytics/pull/6660/commits/f2db75802aad1a7abe7f1fceaefe649a99f320f6. Now both return the same format:

```elxir
iex(1)> PlausibleWeb.VerifiedRoutes.stats_path("a b.com")
"/a%20b.com"
iex(2)> PlausibleWeb.Router.Helpers.stats_path(PlausibleWeb.Endpoint, :stats, "a b.com", [])
"/a%20b.com"
```

apata:
Nice work! We can pass in plausible_web.ex
```
  def router do
    quote do
      use Phoenix.Router, helpers: false
```

zoldar:
> Nice work! We can pass in plausible_web.ex

@apata Ah, thanks, I've missed that. Done in https://github.com/plausible/analytics/pull/6660/commits/5b417faaf2fd7b1af039abfadb9d4b4db2002c4b

PR #6665, opened 2026-09-09
Implement open funnel with first and last steps required

zoldar:
### Changes

This PR implements support for a flexible funnel, where only first and last step are mandatory, while the rest is optional. It's a special, simplified case of an open funnel.

On the implementation side, another shrunk version of the queried funnel is computed - only with 2 steps - the first and last one. The metrics computed for the last step of the short funnel replace the metrics of the last step in the original funnel. And that's basically it.

The UI for configuring funnels is revised as well - now instead of a strict funnel toggle, a user can switch the funnel to work in one of three modes by setting its type:

- sequential (strict_order: false, first_and_last: false)
- flexible - the new mode (strict_order: false, first_and_last: true)
- strict (strict_order: true, first_and_last: false)

The current implementation treats `funnel_type` as a computed field, abstracting the existing `strict_order` and the new `first_an_last` flags. ~While flags could be replaced with a single enum field, I'm not sure if that's the way to go, yet. Funnel may undergo more changes some time soon and might require further revisions of how we store the configuration.~ While `funnel_type` is a real schema field, it's not actively used at this stage. A backfill task will be run together with the updated logic. Once backfill is run, the changeset and base funnel query will be modified to write to and read from `funnel_type` field directly instead of indirectly changing the boolean flag fields. The flag fields will be removed as the part of the final cleanup.

The only visible change on the dashboard is mostly around the sequential/flexible/strict funnel indication copy shown in place of former strict funnel label.

Luckily, our public API did not expose "strict_order" as an option in funnel creation yet, so there's no problem of dealing with a potential breaking change - yet.

The migration will be extracted and released first once this PR gets approved.

image

image

### Tests

### Dark mode


extra/lib/plausible_web/live/funnel_settings/form.ex
```diff
@@ -80,17 +80,62 @@ defmodule PlausibleWeb.Live.FunnelSettings.Form do
...
-                  phx-click="toggle-strict-order"
-                />
+            <div class="flex flex-col gap-y-1 mt-8">
+              <.label>
+                Funnel type
+              </.label>
+              <div class="flex items-center justify-between gap-4 mt-2">
+                <.input
+                  type="radio"
+                  class="block h-5 w-5 dark:bg-gray-700 border-gray-300 text-indigo-600 focus:ring-indigo-600"
```
apata, line 90: issue, blocking: This class doesn't take effect, can put `h-15` here and nothing happens, it is overridden. Same issue for all instances of `.input type="radio"` component in this file.

extra/lib/plausible/stats/funnel.ex
```diff
@@ -54,9 +60,41 @@ defmodule Plausible.Stats.Funnel do
...
+
+  defp convert_to_first_and_last(funnel) do
+    last_step = List.last(funnel.steps)
+
+    steps =
+      [List.first(funnel.steps), last_step]
+      |> Enum.with_index(1)
+      |> Enum.map(fn {step, idx} -> %{step | step_order: idx} end)
+
+    %{funnel | funnel_type: :sequential, steps: steps}
```
apata, line 92: This super optional suggestion has less mapping, unless the first step sometimes doesn't have the right step order?
```suggestion
    first = List.first(funnel.steps)
    last = %{List.last(funnel.steps) | step_order: 2}

    %{funnel | funnel_type: :sequential, steps: [first, last]}
```

apata, line 92: issue, non-blocking: This temporary funnel has contradictory values: it's `first_and_last?: true, funnel_type: :sequential`. 

It seems the virtual field :funnel_type field creates a whole class of these integrity problems. There's a few others spots where they lurk in wait. 

Can we migrate to a single persisted :funnel_type field for this feature, then migrate out to separate columns again if it indeed turns out we need something else?

- migration: create `funnel_type` column
- deploy code change: when persisting funnels, set funnel_type based on strict_order
- migration: backfill `funnel_type` based on `strict_order`
- deploy code change: this PR, rewires everything to purely `funnel_type` and implements the new feature
- migration: drop `strict_order` column

zoldar: With a fresh eye, yeah, I agree, it's best to switch to the new field completely. However, with the abstracting layer already in place, it can be done simpler:

- add `funnel_type` _in addition_ to existing `strict_order` and newly added `first_and_last`
- turn `funnel_type` into a non-virtual schema field; now everything should work normally when deployed in the current state with https://github.com/plausible/analytics/pull/6665/commits/3b76140455f71ae7416b33b7cd80ec5272537cae and https://github.com/plausible/analytics/pull/6665/commits/0c30a5c3c9da004716a6ac468cf4785582cf8570 applied
- backfill `funnel_type` (data script TBD)
- modify `Funnel.changeset` and `Funnels.base_get_query` to populate and read `funnel_type` directly instead of using flag fields
- remove  `first_and_last` and `strict_order`

zoldar: Backfill task got added and the actual logic got fixed, because update of funnel steps with out-of-sync `funnel_type` was involuntarily changing strict funnels to sequential.

extra/lib/plausible_web/live/funnel_settings/form.ex
apata: @zoldar Did you have a chance to look at this? If we're happy with how the form looks right now, the class can be removed totally

zoldar: @sanne-san It seems that our `.input` radio variant completely ignores `class` attribute and defines styling statically at https://github.com/plausible/analytics/blob/open-funnel-with-first-and-last/lib/plausible_web/live/components/form.ex#L140-L140

I have tried temporarily replacing `class="..."` with `class={@class}`, but, at least for me, the outcome looked visually the same. Are you ok with using `class` as defined internally for radio input?

sanne-san: Yes! Not sure why the classes are there but no need to override anything from the default component here 👍

zoldar: https://github.com/plausible/analytics/pull/6665/commits/0ae7bd08fb5315e545b8d5181fb6d821f75b7a16 removes unused attributes from all uses of this radio input component.

apata (approved):

PR #6666, opened 2026-09-09
New team setup UI

RobertJoonas:
### Changes

This PR builds a new UI for the team setup page that is completely independent from TeamManagement code (which continues to work exactly the same way in team settings). The objective is to make the initial team setup UI cleaner in order to add team creation as an optional onboarding step.

The new UI is a simple form, where:

* name update happens only after hitting "Create team" (no db write on defocus anymore)
* the user can add multiple "email + role" rows before hitting submit
* the number of rows is limited by `team_member_limit`. Hitting the plus icon to add rows while at limit will not add any more rows and will display a "not-allowed" cursor
* invitations are sent only after hitting "Create team" (just like before, but there's no confusing extra "Invite" step anymore that might make it feel like the invitation was already sent)
* a validation error on any input field will flash an error message without resetting the form state itself (e.g. some email is invalid or the team name is reserved/prohibited)
* Any rows with an empty email are ignored

_Note: The handrolled `member-rows.js` hook is quite a lot of code, but as a follow-up, I'm planning to introduce a `Prima.Listbox` component and use it here. So `member-rows.js` should eventually get rid of the hand-crafted keyboard navigation, focusing and blurring the dropdown, etc. We'll also be able to use `Listbox` in the TeamManagement LiveView to make the UX feel more snappy in high-latency scenarios._

### Tests

### Changelog

### Documentation

### Dark mode

### DIFF
--- a/assets/js/liveview/member-rows.js
+++ b/assets/js/liveview/member-rows.js
@@ -1,0 +1,58 @@
+// Instantly adds/removes rows, and selects a role, in the "create team"
+// form's member list - entirely client-side, no server round trip. Row and
+// role state is plain form data (an email input and a hidden role input per
+// row), read once when the form is submitted.
+//
+// Expects a `template[data-row-template]` (the row blueprint, with the
+// literal placeholder `__ROW_ID__` standing in for the row id in its
+// attributes), a `[data-row-list]` container to append/remove rows from, a
+// `[data-add-row]` button, `[data-remove-row]` buttons, and role pickers
+// built from a `details[data-role-picker]` containing a `[data-role-label]`
+// span, a `[data-role-value]` hidden input, and `[data-role-item]` buttons.
+
+const ROW_ID_PLACEHOLDER = '__ROW_ID__'
+
+const capitalize = (s) => s.charAt(0).toUpperCase() + s.slice(1)
+
+export default {
+  mounted() {
+    this.template = this.el.querySelector('template[data-row-template]')
+    this.list = this.el.querySelector('[data-row-list]')
+
+    this.el
+      .querySelector('[data-add-row]')
+      .addEventListener('click', () => this.addRow())
+
+    this.list.addEventListener('click', (e) => {
+      const removeButton = e.target.closest('[data-remove-row]')
+      if (removeButton) return this.removeRow(removeButton)
+
+      const roleItem = e.target.closest('[data-role-item]')
+      if (roleItem) return this.selectRole(roleItem)
+    })
+  },
+
+  addRow() {
+    const rowId =
+      window.crypto?.randomUUID?.() ?? `${Date.now()}-${Math.random()}`
+
+    const html = this.template.innerHTML.replaceAll(ROW_ID_PLACEHOLDER, rowId)
+    const wrapper = document.createElement('div')
+    wrapper.innerHTML = html
+
+    this.list.appendChild(wrapper.firstElementChild)
+  },
+
+  removeRow(button) {
+    button.closest('[data-row]').remove()
+  },
+
+  selectRole(item) {
+    const role = item.dataset.roleItem
+    const row = item.closest('[data-row]')
+
+    row.querySelector('[data-role-value]').value = role
+    row.querySelector('[data-role-label]').textContent = capitalize(role)
+    row.querySelector('[data-role-picker]').removeAttribute('open')
+  }
+}
--- a/assets/js/liveview/member-rows.js
+++ b/assets/js/liveview/member-rows.js
@@ -7,8 +7,10 @@
 // literal placeholder `__ROW_ID__` standing in for the row id in its
 // attributes), a `[data-row-list]` container to append/remove rows from, a
 // `[data-add-row]` button, `[data-remove-row]` buttons, and role pickers
-// built from a `details[data-role-picker]` containing a `[data-role-label]`
-// span, a `[data-role-value]` hidden input, and `[data-role-item]` buttons.
+// built from a `details[data-role-picker]` (listbox-button pattern: a
+// `<summary>` trigger, a `[role=listbox]` container, and `[data-role-item]`
+// `[role=option]` buttons) containing a `[data-role-label]` span and a
+// `[data-role-value]` hidden input.
 
 const ROW_ID_PLACEHOLDER = '__ROW_ID__'
 
@@ -30,6 +32,23 @@ export default {
       const roleItem = e.target.closest('[data-role-item]')
       if (roleItem) return this.selectRole(roleItem)
     })
+
+    // Native <details> only closes on a second click on <summary> - close it
+    // on an outside click too, like any other dropdown.
+    this.handleOutsideClick = (e) => {
+      this.list.querySelectorAll('[data-role-picker][open]').forEach((details) => {
+        if (!details.contains(e.target)) this.closeRolePicker(details)
+      })
+    }
+    document.addEventListener('click', this.handleOutsideClick)
+
+    this.list
+      .querySelectorAll('[data-role-picker]')
+      .forEach((details) => this.wireRolePicker(details))
+  },
+
+  destroyed() {
+    document.removeEventListener('click', this.handleOutsideClick)
   },
 
   addRow() {
@@ -39,8 +58,10 @@ export default {
     const html = this.template.innerHTML.replaceAll(ROW_ID_PLACEHOLDER, rowId)
     const wrapper = document.createElement('div')
     wrapper.innerHTML = html
+    const row = wrapper.firstElementChild
 
-    this.list.appendChild(wrapper.firstElementChild)
+    this.list.appendChild(row)
+    this.wireRolePicker(row.querySelector('[data-role-picker]'))
   },
 
   removeRow(button) {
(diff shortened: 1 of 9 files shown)


assets/js/liveview/member-rows.js
```diff
@@ -0,0 +1,178 @@
...
+
+    row.querySelector('[data-role-value]').value = role
+    row.querySelector('[data-role-label]').textContent = capitalize(role)
+    items.forEach((i) => i.setAttribute('aria-selected', i === item))
+
+    this.closeRolePicker(details)
+    details.querySelector('summary').focus()
+  },
+
+  // Wires up the WAI-ARIA listbox-button keyboard pattern for a role picker:
```
apata, line 102: issue, non-blocking: These handlers don't match the dashboard's top bar popovers from `headless-ui` (https://headlessui.com/react/popover). I think ideally, we should be consistent with that. Popover's handlers do less than this and that's good. Less to go wrong.

assets/js/liveview/member-rows.js
```diff
@@ -0,0 +1,178 @@
+// Instantly adds/removes rows, and selects a role, in the "create team"
+// form's member list - entirely client-side, no server round trip. Row and
+// role state is plain form data (an email input and a hidden role input per
+// row), read once when the form is submitted.
+//
+// Expects a `template[data-row-template]` (the row blueprint, with the
```
apata, line 6: nitpick, non-blocking: This template approach is novel (first in this repo), but the comment is quite hard to read and understand for me. A few changes that might improve it
1) refactor into block comment `/* */`
2) show the expected HTML directly OR link the appropriate Elixir module
```js
/* expects HTML in this shape:
  <template data-row-template>
      ...
  </template>
*/
```
On further thought, since Elixir-side we need to handle empty rows anyway, can we create the max member limit rows Elixir-side and make the `+ -> unhide using CSS` -  `- -> hide using CSS & clear using JS`? Could get rid of the templating logic then.

assets/js/liveview/member-rows.js
```diff
@@ -0,0 +1,178 @@
...
+// from a `details[data-role-picker]` (listbox-button pattern: a `<summary>`
+// trigger, a `[role=listbox]` container, and `[data-role-item]`
+// `[role=option]` buttons) containing a `[data-role-label]` span and a
+// `[data-role-value]` hidden input, and a `data-max-rows` attribute capping
+// how many rows can exist (computed server-side from the plan's team member
+// limit) - the add button shows a not-allowed cursor once that cap is hit.
+
+const ROW_ID_PLACEHOLDER = '__ROW_ID__'
+
+const capitalize = (s) => s.charAt(0).toUpperCase() + s.slice(1)
```
apata, line 19: nitpick, non-blocking: this seems like it could be done once on the BE

lib/plausible/teams/team.ex
```diff
@@ -147,11 +147,21 @@ defmodule Plausible.Teams.Team do
...
+
+    # validate_exclusion/3 only runs its check when the field actually changed
+    # relative to the struct's current value, which would let a freshly
+    # auto-created team (whose name already equals the reserved default) keep
+    # that name simply by resubmitting it unchanged. Check unconditionally.
+    if get_field(changeset, :name) == Plausible.Teams.default_name() do
+      add_error(changeset, :name, "is reserved")
+    else
+      changeset
+    end
```
apata, line 164: suggestion, non-blocking: There's a force_change function in [Ecto](https://ecto.hexdocs.pm/Ecto.Changeset.html#force_change/3). We ended up using it for annotations for a similar situation where I originally had built a workaround.

assets/js/liveview/member-rows.js
RobertJoonas: This consistency sounds nice but it's difficult to achieve and the scope expands far beyond this PR. There's a split between LiveViews and the dashboard, and I just logically followed the LiveView example, aiming for consistency with other Prima components (just pushed a PR that creates this component in Prima: https://github.com/plausible/prima/pull/25).

assets/js/liveview/member-rows.js
RobertJoonas: Yeah, agreed that the comment (as well as the whole module in general) is a bit hairy. I'd like to address these improvements in a follow-up though, if that's OK.

> On further thought, since Elixir-side we need to handle empty rows anyway, can we create the max member limit rows Elixir-side and make the + -> unhide using CSS -  - -> hide using CSS & clear using JS? Could get rid of the templating logic then.

Yeah, was thinking the same actually. And not only for cleaner code, but it's one way of making it possible to use the new Prima component... With the current `<template>` approach, LiveView is unaware of those duplicated rows, and it's impossible to mount hooks on rows that don't exist in the initially mounted template.

assets/js/liveview/member-rows.js
RobertJoonas: Hmm, that would mean sending both `value` and `capitalizedValue` from BE -> FE for each role at the current stage. Once we start using the Prima component here, it will move to the backend and will be done once.

assets/js/liveview/member-rows.js
apata: Roger that on the followup!

assets/js/liveview/member-rows.js
apata: Is the split justified though? From perspective of the user, it's all Plausible. 

I understand the reluctance to rewrite this minor piece of UI again, but this extra complexity (over a headlessui popover) isn't justified for this component, at least I don't think so. 

Who are these keyboard handlers for and what's our confidence that they serve the purpose? 

All custom keyboard handlers risk messing with people that can't use a mouse to navigate, or see what's happening on the screen.

To them, the browser's default Home button behavior may be the escape from out of a messed up cookie banner that's preventing them from completing some essential task on a government website. Close your eyes, turn on screen reading, and try to set up an appointment to renew your passport. Another challenge: set up Plausible. 

I'm not an expert on accessibility, but I do know that it takes expertise to build an accessible component.

Anyways, this is just me advising to take care here and in the future with custom keyboard handlers. It's a non-blocking comment because I can't prove that it's not perfectly accessible.

lib/plausible/teams/team.ex
RobertJoonas: Makes sense to use `force_change`, thanks for the suggestion! Did that in https://github.com/plausible/analytics/pull/6666/commits/3aa293758919e58584fa96057d70ad1495ba5ceb. Also made sure that a blank name change doesn't end up with two errors -- both "name is blank" and "name is reserved".

assets/js/liveview/member-rows.js
ukutaht: @apata When it comes to keyboard navigation, Prima's goal is to follow HeadlessUI as much as possible.

HeadlessUI's `<ListBox>` has the same keyboard shortcuts as we have here, plus some more so it can stand in for a native select field: https://headlessui.com/react/listbox#keyboard-interaction

Since this is a form field with a clear set of values to choose from, but with application-specific UI, I think `ListBox` is the most appropriate component here.

RobertJoonas: > Is the split justified though? From perspective of the user, it's all Plausible.

Agreed, by all means.

> I understand the reluctance to rewrite this minor piece of UI again, but this extra complexity (over a headlessui popover) isn't justified for this component, at least I don't think so.

Hmm, are you suggesting to get rid of this keyboard navigation ability in all Prima components? Because I don't think this listbox input component should behave different than Prima's dropdown menu or combobox -- both of them already support such keyboard navigation.

> Who are these keyboard handlers for and what's our confidence that they serve the purpose?

The keyboard shortcuts work for everyone, although it's probably mostly useful for sighted people to interact with the form "extra fast". I've tried to interact with the form with VoiceOver (macOS built-in screen reader) enabled and my eyes closed. Both ways led me to a successful form completion with everything announced properly on the way:

1) with VoiceOver commands (e.g. control+option+arrow_key to navigate, shift+control+option+down_arrow to enter an element, control+option+space to click, etc)
2) using tab / shift+tab / enter / escape / arrow keys

> I'm not an expert on accessibility, but I do know that it takes expertise to build an accessible component.

I'm not an expert either, but using a real screen reader has given me a fair amount of confidence. I see it as our best effort on the UX and not unnecessary complexity.

apata: Right! Sorry @RobertJoonas @ukutaht, I should have read into headless-ui listbox before I commented. It does match the use case. Thanks for taking the time to explain it.

apata (approved):

lib/plausible_web/live/team_setup.ex
```diff
@@ -53,16 +61,230 @@ defmodule PlausibleWeb.Live.TeamSetup do
...
+        <Heroicons.minus class="size-4" />
+      </button>
+    </div>
+    """
+  end
+
+  def handle_event("create-team", %{"team" => %{"name" => name}} = params, socket) do
+    changeset = Teams.Team.name_changeset(socket.assigns.current_team, %{name: name})
+
+    case Repo.update(changeset) do
```
ukutaht, line 177: Doing the name update and invitations separately can get weird. For example if email validation for one of the rows fails, the team name will still be updated and error shown to the user. Since the whole form is now submitted as a single unit, I think the Repo actions should also succeed or fail together.

RobertJoonas: Agreed. Moved the `Repo.update` call into the `Layout.persist` transaction: https://github.com/plausible/analytics/pull/6666/commits/be0984c558f075cc8d7b1d76f84c4b078732cf28

PR #6671, opened 2026-09-11
Update favicon placeholder for sites

sanne-san:
### Changes

- Add a placeholder with a globe on a grey tile, served for `?placeholder=site`
- Use it in the site switcher and on the sites page, where a missing favicon sits next to real ones; the reports keep the plain link icon
- Restyle the "All sites" icon as the same tile in indigo
- Move both site icons into `components/site-icon.tsx`
- Delete the unused `priv/globe_favicon.svg` and `GlobeIcon`

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/plugs/favicon.ex
```diff
@@ -135,10 +149,19 @@ defmodule PlausibleWeb.Favicon do
   end
 
   defp send_placeholder(conn) do
+    conn = fetch_query_params(conn)
+
+    icon =
+      case conn.query_params["placeholder"] do
+        "site" -> @site_placeholder_icon
+        "site_dark" -> @site_dark_placeholder_icon
+        _ -> @placeholder_icon
```
ukutaht: Is there a reason to keep the old placeholder around?

assets/js/dashboard/components/site-icon.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <img
+      aria-hidden="true"
+      alt=""
+      src={`/favicon/sources/${encodeURIComponent(domain)}?placeholder=${placeholder}`}
+      onError={(e) => {
+        const target = e.target as HTMLImageElement
+        target.onerror = null
+        target.src = `/favicon/sources/placeholder?placeholder=${placeholder}`
```
ukutaht: This URL was pretty confusing to read at first. I think it would be clearer if the query string was `?ui-mode=light|dark` instead of `placeholder=site|site_dark`

lib/plausible_web/plugs/favicon.ex
sanne-san: Yeah, this is the link icon that's still used in the Sources reports. Wasn't sure if we should change that to a globe icon too, but I think keeping the link icon is more fitting (and style wise the line icon fits better with other icons used in other reports).

assets/js/dashboard/components/site-icon.tsx
sanne-san: Good point, changed. Now it's `?placeholder=source|site` for the drawing and `?ui-mode=light|dark` for the theme. I also moved the standalone placeholder URL to `/favicon/placeholders/site`, otherwise it'd read `/favicon/sources/placeholder?placeholder=site`.

ukutaht (approved):

PR #6681, opened 2026-09-16
Implement e-mail sending suppression (core functions)

aerosol:
### Changes

This PR sets a base for e-mail suppression, based on Postmark-proxied bounce/complaint data.
Migration will be extracted to a separate PR once review is concluded.

Continued in: https://github.com/plausible/analytics/pull/6685

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/mailer.ex
```diff
@@ -19,4 +33,16 @@ defmodule Plausible.Mailer do
...
     end
   end
+
+  defp suppressed_recipients(email) do
+    # to, cc/bcc can hold a raw string, a {name, address} tuple, or any
+    # struct implementing `Bamboo.Formatter` (e.g. `Plausible.Auth.User`),
+    # each possibly wrapped in a list.
+    normalized = Bamboo.Mailer.normalize_addresses(email)
+
+    [normalized.to, normalized.cc, normalized.bcc]
```
aerosol: We don't do any carbon copying now AFAIK, but it's going to be nice to have _some_ behavior defined, if we ever do. Maybe some e-mail notifications targeting multiple team members could be wrapped in a single message in the future.

lib/plausible/mailer.ex
```diff
@@ -2,8 +2,22 @@ defmodule Plausible.Mailer do
   use Bamboo.Mailer, otp_app: :plausible
   require Logger
 
-  @spec send(Bamboo.Email.t()) :: :ok | {:error, :unknown_error}
+  @spec send(Bamboo.Email.t()) :: :ok | {:error, :unknown_error | :suppressed}
   def send(email) do
+    case suppressed_recipients(email) do
+      [] ->
```
aerosol, line 9: TODO: we shouldn't suppress priority e-mails ever?

lib/plausible/mailer.ex
aerosol: I'm considering just crashing here if there's more than 1 address

lib/plausible/email_suppression.ex
```diff
@@ -0,0 +1,62 @@
...
+  @spec changeset(t(), map()) :: Ecto.Changeset.t()
+  def changeset(suppression, attrs) do
+    suppression
+    |> cast(attrs, @required ++ @optional)
+    |> validate_required(@required)
+    |> unique_constraint(:email)
+  end
+
+  @spec reactivate_changeset(t(), pos_integer()) :: Ecto.Changeset.t()
+  def reactivate_changeset(suppression, user_id) do
```
zoldar, line 56: Nit, feel free to ignore: To make the API slightly less confusing, the changeset could accept `User` struct and use id field only, internally.

lib/plausible/email_suppressions.ex
```diff
@@ -0,0 +1,110 @@
...
+  @spec reactivate(String.t(), Plausible.Auth.User.t()) ::
+          {:ok, EmailSuppression.t()} | {:error, :not_found | Ecto.Changeset.t()}
+  def reactivate(email, %Plausible.Auth.User{id: user_id}) do
+    case Repo.get_by(EmailSuppression, email: email) do
+      nil ->
+        {:error, :not_found}
+
+      suppression ->
+        suppression
+        |> EmailSuppression.reactivate_changeset(user_id)
```
zoldar, line 97: Reactivation overriding any future postmark updates is intentional, correct?

lib/plausible/mailer.ex
```diff
@@ -1,9 +1,24 @@
...
+        do_send(email)
+
+      suppressed ->
```
zoldar, line 12: Given only a subset of emails got suppressed, would it make sense to still send the email to the ones that weren't?

lib/plausible/mailer.ex
```diff
@@ -19,4 +34,38 @@ defmodule Plausible.Mailer do
...
+      else
+        # to, cc/bcc can hold a raw string, a {name, address} tuple, or any
+        # struct implementing `Bamboo.Formatter` (e.g. `Plausible.Auth.User`),
+        # each possibly wrapped in a list.
+        normalized = Bamboo.Mailer.normalize_addresses(email)
+
+        [normalized.to, normalized.cc, normalized.bcc]
+        |> List.flatten()
+        |> Enum.map(fn {_name, address} -> address end)
+        |> Enum.filter(&Plausible.EmailSuppressions.suppressed?/1)
```
zoldar, line 63: Maybe it would be possible to run that check in one go with a single query?

lib/plausible/mailer.ex
aerosol: See https://github.com/plausible/analytics/pull/6681#discussion_r4022778721 - I think we might just crash if there's >1 recipient right now (AFAIU we never send to more than one at the moment)

lib/plausible/mailer.ex
aerosol: See https://github.com/plausible/analytics/pull/6681#discussion_r4022778721 - I think we might just crash if there's >1 recipient right now (AFAIU we never send to more than one at the moment)

lib/plausible/email_suppressions.ex
aerosol: Yeah that was intentional but it's a good question nevertheless. 
Reactivating like that means we will never deactivate. 

@metmarkosaric once you manually reactivate an address, and then subsequent send results with some error still, should we deactivate the address or keep it active (application-side) until manually deactivated?

metmarkosaric: @aerosol  hmm in that case I think it is best to deactivate if the error again happens to be one of those we don't send emails for. we won't be manually deactivating addresses on our own

lib/plausible/mailer.ex
zoldar: Got it, it's for non-priority comms anyway, so no huge harm there anyway

lib/plausible/mailer.ex
aerosol: https://github.com/plausible/analytics/pull/6681/commits/e181e365cd070f903fe50474a76d2423f9757fed

lib/plausible/email_suppressions.ex
aerosol: 6cad0321c3

zoldar (approved):

PR #6684, opened 2026-09-16
Implement OAuth authorization server core functions

apata:
### Changes

Implements core functions of the OAuth authorization server needed for an MCP endpoint. These are not exposed to the internet yet because the SSRF client needs to be validated to work with a large number of external URLs.

Minor tuning:
- Proposes new team_role/1 function `@spec team_role(team_id: pos_integer(), user_id: pos_integer()) ::
          {:ok, Teams.Membership.role()} | {:error, :not_a_member}`.

- Makes `scopes` field required on grant and code, enforces minimum count of 1

- Enforces limits for challenge and verifier. Previously, one could declare an empty verifier, which defeats the purpose. 

- Removes optional `scopes_supported` field from authorization server metadata endpoint: can't declare it here globally because `scopes_supported` depends on resource.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible/oauth/cimd.ex
```diff
@@ -0,0 +1,233 @@
...
+    path
+    |> String.split("/")
+    |> Enum.any?(fn segment ->
+      (segment |> String.downcase() |> String.replace("%2e", ".")) in [".", ".."]
+    end)
+  end
+
+  defp ssrf_get(url) do
+    opts =
+      [receive_timeout: @fetch_timeout, decode_body: false]
```
zoldar, line 69: For a moment I thought we could check for the size of the payload as the request is running and reject it mid-flight if needed, but that requires streaming + reduction/into shenanigans. A relatively short timeout is enough of a safeguard anyway.

lib/plausible/oauth/cimd.ex
```diff
@@ -0,0 +1,233 @@
...
+  def validate_redirect_uris(_redirect_uris), do: {:error, :missing_redirect_uris}
+
+  defp valid_redirect_uri?(uri)
+       when is_binary(uri) and byte_size(uri) <= @max_redirect_uri_length do
+    case URI.parse(uri) do
+      %URI{fragment: fragment} when not is_nil(fragment) -> false
+      %URI{userinfo: userinfo} when not is_nil(userinfo) -> false
+      %URI{scheme: "https", host: host} -> is_binary(host) and host != ""
+      %URI{scheme: "http", host: host} -> loopback_host?(host)
+      %URI{scheme: scheme} when is_binary(scheme) -> String.contains?(scheme, ".")
```
zoldar, line 151: Is that for cases of scheme-less URIs?

lib/plausible/oauth/cimd.ex
```diff
@@ -0,0 +1,233 @@
...
+
+  # Control characters (C0, DEL, C1) truncate or garble the name when its displayed.
+  # Bidi overrides `U+202A`-`U+202E` and isolates `U+2066`-`U+2069`
+  # reorder the glyphs around them. Therefore, names containing such characters are unreadable and forbidden.
+  defp unreadable?(name) do
+    name
+    |> String.to_charlist()
+    |> Enum.any?(fn codepoint ->
+      codepoint <= 0x1F or codepoint in 0x7F..0x9F or
+        codepoint in 0x202A..0x202E or codepoint in 0x2066..0x2069
```
zoldar, line 183: Is that prompted by the spec? That excludes right-to-left languages like Arabic from use for client names too, right? Just making sure.

lib/plausible/oauth/cimd.ex
```diff
@@ -0,0 +1,233 @@
...
+        loopback_host?(registered_uri.host) and equal_but_for_port?(registered_uri, uri)
+      end)
+  end
+
+  defp loopback_host?(host), do: downcase(host) in ["localhost", "127.0.0.1", "::1"]
+
+  defp equal_but_for_port?(%URI{} = a, %URI{} = b) do
+    downcase(a.scheme) == downcase(b.scheme) and
+      is_nil(a.userinfo) and
+      a.userinfo == b.userinfo and
```
zoldar, line 224:
```suggestion
      is_nil(b.userinfo) and
```

maybe?

lib/plausible/oauth.ex
```diff
@@ -0,0 +1,240 @@
...
+
+  alias Plausible.OAuth.{AuthorizationCode, Grant, PKCE, ProtectedResources, Token}
+  alias Plausible.Teams.Memberships
+
+  @authorization_code_ttl_seconds 600
+  @access_token_ttl_seconds 3600
+  @refresh_token_ttl_seconds 30 * 24 * 3600
+
+  @roles_with_oauth [:owner, :admin, :editor, :billing, :viewer]
+  true = Enum.all?(@roles_with_oauth, &(&1 in Plausible.Teams.Membership.roles()))
```
zoldar, line 25: I'm wondering if it would make sense to implement a convenience for this kind of assertion. Something like:

```elixir
  def roles!(roles) do
    roles_wrapped = List.wrap(roles)

    if Enum.any(roles_wrapped, &(&1 not in @roles)) do
      raise "Unknown roles: #{wrapped_roles -- @roles}"
    end

    roles
  end
```

in `Plausible.Teams.Membership` perhaps.

Then this could be expressed as:

```elixir
@roles_with_oauth Teams.Membership.roles!([:owner, :admin, :editor, :billing, :viewer])
```

lib/plausible/oauth/protected_resources.ex
```diff
@@ -2,11 +2,70 @@ defmodule Plausible.OAuth.ProtectedResources do
...
   """
 
+  @type t() :: %{
+          resource_path: String.t(),
+          scopes_supported: [String.t()]
+        }
+
+  @spec mcp() :: t()
   def mcp(),
     do: %{
```
zoldar, line 16: Why not make it a `Resource` struct inlined here, or a separate, `OAuth.ProtectedResource` (singular) to make aliasing less awkward? Though maybe that's an overkill at this stage. Feel free to ignore.

lib/plausible/oauth/protected_resources.ex
```diff
@@ -2,11 +2,70 @@ defmodule Plausible.OAuth.ProtectedResources do
...
+  Rejects the whole list if it names a scope the resource does not support,
+  which can happen when a scope is withdrawn (or refactored) after the list was stored.
+  """
+  @spec normalize_granted_scopes([String.t()], t()) ::
+          {:ok, [String.t()]} | {:error, :invalid_scope}
+  def normalize_granted_scopes(scopes, resource) when is_list(scopes) and scopes != [] do
+    supported = resource.scopes_supported
+
+    if Enum.all?(scopes, &(&1 in supported)) do
+      {:ok, Enum.filter(supported, &(&1 in scopes))}
```
zoldar, line 64: Maybe I'm missing something but this seems redundant? Could it be just:

```suggestion
      {:ok, scopes}
```
? Though this implies that's rather validation and not normalisation.

lib/plausible/oauth.ex
```diff
@@ -0,0 +1,240 @@
...
+    resource_url = ProtectedResources.get_resource_url(resource)
+
+    query =
+      from(g in Grant,
+        where:
+          g.access_token_hash == ^Token.hash(raw_access) and
+            g.resource == ^resource_url and
+            g.access_token_expires_at > ^now() and
+            is_nil(g.revoked_at),
+        preload: [:user, :team]
```
zoldar, line 203: Not a huge deal, but those two relations could be preloaded from joins to avoid extra roundtrips to DB.

zoldar (approved):
Nice work ✨ Left only some questions and rather optional suggestions.

lib/plausible/oauth/protected_resources.ex
apata: This reorders `scopes` according to the order in the `scopes_supported` as declared on resource. Most likely the order won't matter on our side: resource-side we will check it something like `grant.scopes |> Enum.any(fn s -> s == required_scope end)`. Will cleanup.

lib/plausible/oauth.ex
apata: Hmm, you're right, this function does too many trips: Memberships.team_role() also queries. 

I checked the codebase and we often build ad-hoc queries with `!= :guest`, which is basically what I'm doing here, except I'm favoring a whitelist not blacklist.

Can we do it without redeclaring the `user IS in team WITH role DB relationship` in this module? I don't think this module should be aware of it.

lib/plausible/oauth/cimd.ex
apata: It should reject a scheme like `javascript`, but accept `com.someapp`, a custom url scheme sometimes registered by mobile apps (explained a bit on https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-8.4.3-1). 

It's the reverse of a real domain name, one of the ways people can prove on app stores that the app is authored by people connected with `someapp.com`. Apparently, some popular apps use dotless custom url schemes, `instagram`, `airbnb`, `claude` which would be rejected by this handler (vs `com.openai.chat` which would be accepted). 

Also, apps can use universal urls that begin with `https://`. 

We could drop this clause and reintroduce when we get a related feature request.

lib/plausible/oauth/cimd.ex
apata: Good spot!

lib/plausible/oauth/cimd.ex
apata: Thanks! It's not from spec, just general protective (potentially overprotective) approach against shenanigans. 

`Client So-and-so from https://example.com/identity.json requests your permissions to ...`: we don't want to allow characters in the name to control the surroundings. 

We really want the users to check the URL `client_id` primarily. They should verify that it's a trustworthy value.

I checked just now, this passes locally, so I think we're safe regarding legitimate names in RTL languages. Not 100% sure, but I think these control characters are not usually present when typing out something. 
```
    test "accepts an absent name, non-ASCII text, right to left, and internal spaces" do
      for name <- [
            nil,
            "Claude Code",
            "Café Auswärts",
            # "analysis client" in Japanese
            "分析クライアント",
            # "analysis app 100" in Arabic (RTL)
            "تطبيق تحليل 100",
            String.duplicate("N", 255)
          ] do
        assert :ok = CIMD.validate_client_name(name)
      end
    end
```

lib/plausible/oauth.ex
apata: How about this? I simplified it a bit https://github.com/plausible/analytics/pull/6684/commits/c1c836b51e9e24244d15cbde360f2c21bf988797

lib/plausible/oauth/cimd.ex
apata: Fixed in https://github.com/plausible/analytics/pull/6684/commits/9374829eab3bb3e964c319435aaa2b4ecfeba06a

lib/plausible/oauth/cimd.ex
apata: I start rejecting it in https://github.com/plausible/analytics/pull/6684/commits/9374829eab3bb3e964c319435aaa2b4ecfeba06a

lib/plausible/oauth.ex
zoldar: Very nice

PR #6688, opened 2026-09-17
Clean up credo warnings

zoldar:
### Changes

We've been relying on credo diff for checks because of a long tail of complaints that were hard to address. The tail got so small, it was manageable to do it in one go.

It will also help avoiding issues with reverts where we often got bitten by CI failing due to credo complaints popping up out of the blue.


apata (approved):
Nice

PR #6699, opened 2026-09-22
Team management UI improvements

RobertJoonas:
### Changes

_Still a draft because this PR depends on a new Prima release._

* Create a new `PrimaListbox` component: powered by `Prima.Listbox` and provides common styling
* Use `PrimaListbox` in three places:
  * Role select in the team setup screen
  * Role select in the (single) invite form in team management (Team settings > General)
  * Member-role-switcher (on each existing team member row) in team management
* While this was meant as a refactor, a few functional changes were necessary or made sense to do:
  * Extract the "Remove member" option out of the dropdown, into a trashcan icon to the right of the role-switcher
  * Stop displaying the "Owner" option in both the invite form and in member-role-switcher **if the user doesn't have the permission to invite or change existing roles to "Owner"** (currently an option that always shows but is enabled only for the owner role -- only owners can create other owners)
  * Do not let the user open the member-role-switcher at all if all options are disabled
  * Highlight the currently selected option with a blue checkmark (relying on `data-selected` to make the checkmark appear instantly on the client-side)

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/prima_listbox.ex
```diff
@@ -0,0 +1,130 @@
...
+          assigns.class,
+          assigns.disabled && "pointer-events-none !text-gray-300 dark:!text-gray-600"
+        ]
+      )
+
+    ~H"""
+    <Listbox.listbox_trigger
+      id={@id}
+      class={@computed_class}
+      aria-disabled={@disabled && "true"}
```
ukutaht: From a library design perspective I don't like that we have to worry about `aria-disabled` or `tabindex` here. I don't have a specific proposal yet but I'll play around a bit

PR #6705, opened 2026-09-22
Fix some flaky tests

RobertJoonas:
Only touches a single test file.

In several tests, the sorting criteria was not sufficient to guarantee deterministic ordering of team members across test runs, making assertions like `nth-of-type(1)` unreliable. This commit changes those tests to use `user.email`-based selectors instead. Ordering itself is unit-tested well.

`Layout.sorted_for_display/1` tied on `entry.name` (because every fixture user without an explicit name falls back to "Jane Smith"). Therefore, tests were relying on the ExMachina email sequence in string comparison for a tie-break. It works 99% of the time (i.e. "user-1" comes before "user-2" alphabetically, "user-2" before "user-3", and so on), but in rare cases where the digit length changes, that ordering breaks (i.e.: 10 comes before 9, 100 comes before 99, etc).


zoldar (approved):
❤️

PR #6706, opened 2026-09-22
Unify button and input styling across the app

sanne-san:
### Changes

- Buttons, dropdown and listbox triggers, inputs and selects now get their look from btn-* and input-* classes in app.css. Call sites pick a theme and a size instead of listing the styles themselves.
- Buttons and fields share one size scale of 32, 36 and 40px, so a field lines up with the button beside it. The heights are explicit and no longer depend on line-height.
- The class attribute of the input and dropdown components now adds to the base style instead of replacing it. Replacing was what forced a caller to restate the whole style to change one detail, and it is what produced the copies that this commit removes.
- Drops those copies and the !important overrides they needed: the role trigger, the ellipsis menu, the site people dropdown, the filter bar field, the mobile navigation select, the Google button and the header links.
- Makes an icon-only button a modifier on top of any theme, btn-icon, rather than a theme of its own. Spacing and colour are separate axes again, so an icon button can be red or indigo without a theme for each combination.
- Secondary buttons, danger buttons and fields now share one surface in dark mode, instead of three different greys.

Visual only, no change of behaviour. The custom Alpine dropdown now takes the same theme and size attributes as PrimaDropdown, so its two callers can move over and it can be deleted in a follow-up.

### Tests

### Changelog

### Documentation

### Dark mode


lib/plausible_web/components/layouts/header.html.heex
```diff
@@ -134,39 +134,20 @@
...
-                    Login
-                  </a>
-                </div>
-              </li>
-            </ul>
+            <div
+              class="flex items-center gap-2"
+              x-show="!document.cookie.includes('logged_in=true')"
+            >
+              <.button_link href="/login" theme="ghost" mt?={false}>Login</.button_link>
```
RobertJoonas, line 141: Non-blocking: What do you think about removing the `mt?` attribute? IMO the button shouldn't have to worry about its margins. Doesn't need to happen as part of this PR, but it's just something that's already bothered me before.

lib/plausible_web/components/layouts/header.html.heex
```diff
@@ -134,39 +134,20 @@
...
+              <.button_link href="/register" mt?={false}>Sign up</.button_link>
```
RobertJoonas, line 149: We could avoid these big if branches by using `:if={Plausible.Auth.check_registration_enabled() != :ok}` on the register link directly.

lib/plausible_web/components/generic.ex
```diff
@@ -1010,127 +1031,16 @@ defmodule PlausibleWeb.Components.Generic do
...
   def edit_button(assigns) do
     ~H"""
-    <.icon_button :let={icon_class} href={@href} class={@class} {@rest}>
-      <svg
-        xmlns="http://www.w3.org/2000/svg"
-        fill="none"
-        viewBox="0 0 24 24"
-        class={icon_class}
-      >
+    <button
```
RobertJoonas, line 1039: Could we use `<.button icon?={true}>` here?

lib/plausible_web/components/generic.ex
```diff
@@ -1149,14 +1059,19 @@ defmodule PlausibleWeb.Components.Generic do
   attr :rest, :global, include: ~w(method disabled)
 
   def delete_button(assigns) do
+    assigns =
```
RobertJoonas, line 1062: I might be missing something, but JIC the same question: maybe possible to use `<.button icon?={true}>` here too?

RobertJoonas (commented):
Looks really nice code-wise, nice work! 🙌 I'd like to do a staging sweep of the affected UI elements before leaving an approval though.

lib/plausible_web/components/layouts/header.html.heex
sanne-san: Yep, agree! I'll take care of it in a separate PR 👍

lib/plausible_web/components/generic.ex
sanne-san: Yeah good catch, this was an oversight as I changed how icon styles are being applied afterwards 👌

lib/plausible_web/components/layouts/header.html.heex
```diff
@@ -133,20 +133,19 @@
...
               <li :if={ee?()} id="changelog-notification" class="relative py-2 empty:hidden"></li>
             </ul>
-          <% Plausible.Auth.check_registration_enabled() != :ok -> %>
-            <div
-              class="flex items-center gap-2"
-              x-show="!document.cookie.includes('logged_in=true')"
-            >
-              <.button_link href="/login" theme="ghost" mt?={false}>Login</.button_link>
-            </div>
           <% true -> %>
```
RobertJoonas: Since the `cond` statement now has only two branches, we could turn it into a simple `if`.

